PR #3854, opened 2017-05-02
Temp keys policy onboarding

chenriksson:


src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            await EntitiesContext.SaveChangesAsync();
+
+            return RedirectToAction("Index");
+        }
+
+        private static string[] GetUsernamesFromQuery(string query)
+        {
+            return query.Split(new[] { ',', '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
+                .Select(username => username.Trim()).ToArray();
```
joelverhagen: You should probably filter out empty results after trimming.

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        private IEnumerable<User> FindUsers(string[] usernames)
+        {
+            return EntitiesContext.Users
+                .Where(u => usernames.Any(name => u.Username.Equals(name, StringComparison.OrdinalIgnoreCase)))
```
joelverhagen: This properly becomes `Username IN ('a', 'b', ...)` in SQL? Or is this being evaluated client side?

src/NuGetGallery/Areas/Admin/Views/SecurityPolicy/Index.cshtml
```diff
@@ -0,0 +1,168 @@
+﻿@model SecurityPolicyViewModel
+
+@{
+    ViewBag.Title = "Security Policies";
+}
+
+<section>
```
joelverhagen: Could we get a screenshot?

chenriksson: @joelverhagen sent

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
chenriksson: Evaluated SQL-side, with similar query...
~~~~
SELECT * FROM Users WHERE EXISTS(
  SELECT 1 AS [C1] FROM  (
    SELECT N'name1' AS [C1] FROM  ( SELECT 1 AS X ) AS [SingleRowTable1]
    UNION ALL
    SELECT N'name2' AS [C1] FROM  ( SELECT 1 AS X ) AS [SingleRowTable2]
) AS [UnionAll2] WHERE Users.[Username] = [UnionAll2].[C1])
~~~~

src/NuGetGallery/Areas/Admin/Views/SecurityPolicy/Index.cshtml
```diff
@@ -0,0 +1,168 @@
...
+                    }
+                }
+                
+                this.toggleSelectAll = function (data, e) {
+                    var policyGroup = e.currentTarget.nextSibling.data;
+                    $self.selectAllState[policyGroup](!$self.selectAllState[policyGroup]());
+                    return true;
+                };
+
+                this.generateValue = function (user, policyGroup) {
```
joelverhagen: Do we really have to invent our own serialization here? Can't we just JSON encode?

src/NuGetGallery/Security/UserSecurityPolicyGroup.cs
```diff
@@ -0,0 +1,81 @@
...
+        
+        private static IEnumerable<UserSecurityPolicyGroup> CreateUserPolicyGroups()
+        {
+            yield return new UserSecurityPolicyGroup()
+            {
+                Name = UserSecurityPolicyGroup.SecurePush,
+                Policies = new []
+                {
+                    new UserSecurityPolicy(RequirePackageVerifyScopePolicy.PolicyName),
+                    new UserSecurityPolicy(RequireMinClientVersionForPushPolicy.PolicyName) { Value = "{\"v\":\"4.1.0\"}" }
```
joelverhagen: Let's use JSON API to write this JSON string.

src/NuGetGallery/Security/UserSecurityPolicyGroup.cs
```diff
@@ -0,0 +1,81 @@
...
+                    c.Scopes.Count == 0 ||
+                    c.Scopes.Any(s =>
+                        s.AllowedAction.Equals(NuGetScopes.PackagePush, StringComparison.OrdinalIgnoreCase) ||
+                        s.AllowedAction.Equals(NuGetScopes.PackagePushVersion, StringComparison.OrdinalIgnoreCase)
+                        ))
+                );
+
+            foreach (var key in pushKeys)
+            {
+                key.Expires = DateTime.Now.AddDays(7);
```
joelverhagen: This will fall into the typical expiration notification flow -- so user's might now know reason why this is expiring, right?

joelverhagen: Also, `DateTime.UtcNow`

joelverhagen: This could actually give some keys _more_ time right? Shouldn't we only do this if the `Expires` is greater than 7 days in the future?

src/NuGetGallery/Security/UserSecurityPolicyGroup.cs
```diff
@@ -0,0 +1,81 @@
...
+                {
+                    _instances = CreateUserPolicyGroups().ToList();
+                }
+                return _instances;
+            }
+        }
+        
+        private static IEnumerable<UserSecurityPolicyGroup> CreateUserPolicyGroups()
+        {
+            yield return new UserSecurityPolicyGroup()
```
joelverhagen: If this is all constant, can't we just subclass `UserSecurityPolicyGroup`? This allows us to avoid the `Action<User>` and just make it an overridden method.

src/NuGetGallery/Security/UserSecurityPolicyGroup.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using NuGetGallery.Authentication;
+
+namespace NuGetGallery.Security
+{
+    /// <summary>
+    /// Grouping of one or more user security policies for enrollment.
+    /// </summary>
+    public class UserSecurityPolicyGroup
```
joelverhagen: It seems to me like a "policy group" is a bit redundant. Isn't this just another policy, a la "aggregate policy"? Maybe I'm off here but couldn't a policy either have it's own logic, be an aggregate of other policies, or both? I don't see why we need another type for a grouping when the `UserSecurityPolicy` type could describe it as well.

src/NuGetGallery/Areas/Admin/Views/SecurityPolicy/Index.cshtml
chenriksson: Updated

src/NuGetGallery/Security/UserSecurityPolicyGroup.cs
chenriksson: @joelverhagen @skofman1 per our discussion, will add a new column to UserSecurityPolicies table to track what was enrolled in - which could be a single policy or aggregate.

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -0,0 +1,117 @@
...
+
+namespace NuGetGallery.Areas.Admin.Controllers
+{
+    public class SecurityPolicyController : AdminControllerBase
+    {
+        public IEntitiesContext EntitiesContext { get; }
+
+        public SecurityPolicyController(IEntitiesContext entitiesContext)
+        {
+            EntitiesContext = entitiesContext;
```
skofman1: null check

src/NuGetGallery/Security/UserSecurityPolicyGroup.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        /// <summary>
+        /// Action to take on user enrollment.
+        /// </summary>
+        public Action<User> OnEnroll { get; set; }
+
+        /// <summary>
+        /// Get supported user security policy groups.
+        /// </summary>
+        private static List<UserSecurityPolicyGroup> _instances;
```
skofman1: If you need this class to be a singleton you should use dependency injection and configure it there.

src/NuGetGallery/Security/UserSecurityPolicyGroup.cs
```diff
@@ -0,0 +1,87 @@
...
+        public static List<UserSecurityPolicyGroup> Instances
```
skofman1: consider using Lazy

src/NuGetGallery/Security/UserSecurityPolicyGroup.cs
```diff
@@ -0,0 +1,87 @@
...
+                    RequireMinClientVersionForPushPolicy.CreatePolicy(new NuGetVersion(SecurePushVersion))
+                },
+                OnEnroll = OnEnroll_SecurePush
+            };
+        }
+
+        private static void OnEnroll_SecurePush(User user)
+        {
+            var pushKeys = user.Credentials.Where(c =>
+                c.Type.StartsWith(CredentialTypes.ApiKey.Prefix) &&
```
skofman1: >c.Type.StartsWith(CredentialTypes.ApiKey.Prefi [](start = 16, length = 46)

I think there is a method IsApiKey that has this logic

src/NuGetGallery/Areas/Admin/ViewModels/SecurityPolicySearchResult.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Usernames not found in the database.
+        /// </summary>
+        public IEnumerable<string> UsersNotFound { get; set; }
+    }
+
+    /// <summary>
+    /// Security policy group enrollments for a user.
+    /// </summary>
+    public class SecurityPolicyEnrollments
```
skofman1: >r [](start = 21, length = 1)

one class per file please

src/NuGetGallery/Areas/Admin/Views/SecurityPolicy/Index.cshtml
skofman1: can you send everybody?

---
In reply to: [114229583](https://github.com/NuGet/NuGetGallery/pull/3854#discussion_r114229583) [](ancestors = 114229583)

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -0,0 +1,117 @@
...
+                        var removedPolicies = user.RemovePolicies(policyGroup);
+                        foreach (var p in removedPolicies)
+                        {
+                            EntitiesContext.UserSecurityPolicies.Remove(p);
+                        }
+                    }
+                }
+            }
+
+            await EntitiesContext.SaveChangesAsync();
```
skofman1: should we send a user an e-mail when he is enrolled? Perhaps Anand has ideas.

skofman1:
:shipit:

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
scottbommarito: @anangaur

src/NuGetGallery/Security/UserSecurityPolicyGroup.cs
scottbommarito: @chenriksson make sure that column handles the state of being enrolled in multiple aggregates.

e.g.
I'm enrolled in policy A through groups X and Y. I leave group X. I should still be enrolled in policy A through group Y.

src/NuGetGallery/Security/UserSecurityPolicyGroup.cs
scottbommarito: Agreed. This is really confusing as is...just make this an abstract class with `SecurePushUserSecurityPolicyGroup` a subclass with all of this logic.

tests/NuGetGallery.Facts/Security/UserSecurityPolicyGroupFacts.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace NuGetGallery.Security
+{
+    public class UserSecurityPolicyGroupFacts
+    {
+        [Theory]
+        [InlineData("")]
+        [InlineData("[{\"a\":\"package:push\", \"s\":\"theId\"}]")]
+        [InlineData("[{\"a\":\"package:pushversion\", \"s\":\"theId\"}]")]
+        public void SecurePush_OnEnrollExpiresPushApiKeys(string scopes)
```
scottbommarito: They expire after a week, correct?

```
SecurePush_OnEnrollExpiresPushApiKeysIn1Week
```

tests/NuGetGallery.Facts/Security/UserSecurityPolicyExtensionsFacts.cs
```diff
@@ -0,0 +1,135 @@
...
+using Newtonsoft.Json;
+using System.Web.Mvc;
+
+namespace NuGetGallery.Security
+{
+    public class UserSecurityPolicyExtensionsFacts
+    {
+        [Theory]
+        [InlineData("", "")]
+        [InlineData(null, "")]
```
scottbommarito: should also check `[InlineData("", null)]` just in case

src/NuGetGallery/Security/UserSecurityPolicyExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Linq;
+
+namespace NuGetGallery.Security
+{
+    public static class UserSecurityPolicyExtensions
+    {
+        /// <summary>
+        /// Determine whether two security policies are equivalent.
+        /// </summary>
+        public static bool Matches(this UserSecurityPolicy first, UserSecurityPolicy second)
```
scottbommarito: Why is this not an `Equals` method?

e.g. `UserSecurityPolicy` should implement `IEquatable<UserSecurityPolicy>`

tests/NuGetGallery.Facts/Security/UserSecurityPolicyExtensionsFacts.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        [Theory]
+        [InlineData("[[\"A\",\"\"]]", "[[\"A\",null]]")]
+        [InlineData("[[\"A\",\"B\"],[\"E\",\"\"]]", "[[\"E\",\"\"],[\"A\",\"B\"],[\"C\",\"D\"]]")]
+        public void IsEnrolledReturnsTrue(string groupPolicies, string userPolicies)
+        {
+            // Arrange.
+            var group = new UserSecurityPolicyGroup()
+            {
+                Policies = LoadPolicies(groupPolicies)
```
scottbommarito: If you use `MemberData` instead of `InlineData` you can get rid of the serialization/deserialization.

tests/NuGetGallery.Facts/Security/UserSecurityPolicyExtensionsFacts.cs
```diff
@@ -0,0 +1,135 @@
...
+            LoadPolicies(userPolicies).ToList().ForEach(p => user.SecurityPolicies.Add(p));
+
+            // Act & Assert.
+            Assert.False(user.IsEnrolled(group));
+        }
+
+        [Theory]
+        [InlineData("[[\"A\",\"\"]]")]
+        [InlineData("[[\"A\",\"B\"],[\"E\",\"\"]]")]
+        public void AddPoliciesAddsPolicies(string groupPolicies)
```
scottbommarito: Do we want test cases around
1 - adding policies that are already added
2 - removing policies that are not enrolled

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            return RedirectToAction("Index");
```
scottbommarito: Usually these kind of operations will set a message on the page, e.g. `User x was enrolled in y.`

Do we want to do that here? It could be nontrivial to figure out what was changed otherwise.

chenriksson:
@scottbommarito 

policies

src/NuGetGallery/Areas/Admin/Views/SecurityPolicy/Index.cshtml
chenriksson: Attached to conversation

src/NuGetGallery/Security/SecurePushSubscription.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    public class SecurePushSubscription : IUserSecurityPolicySubscription
+    {
+        public const string SubscriptionName = "SecurePush";
+        public const string MinClientVersion = "4.1.0";
+
+        /// <summary>
+        /// Subscription name.
+        /// </summary>
+        public string Name
```
joelverhagen: >Name [](start = 22, length = 4)

Nit: `=> SubscriptionName`

src/NuGetGallery.Core/Entities/UserSecurityPolicy.cs
```diff
@@ -1,22 +1,41 @@
...
+using System;
 using System.ComponentModel.DataAnnotations;
 
 namespace NuGetGallery
 {
     /// <summary>
     /// User-subscribed security policy.
     /// </summary>
-    public class UserSecurityPolicy : IEntity
+    public class UserSecurityPolicy : IEntity, IEquatable<UserSecurityPolicy>
```
joelverhagen: >IEquatable [](start = 47, length = 10)

IEquatable should override GetHashCode

tests/NuGetGallery.Core.Facts/Entities/UserSecurityPolicyFacts.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        [Theory]
+        [MemberData(nameof(EqualsReturnsTrue_Data))]
+        public void EqualsReturnsTrueForPolicyMatches(UserSecurityPolicy first, UserSecurityPolicy second)
+        {
+            // Act & Assert
+            Assert.True(first.Equals(second));
+        }
+
+        public static IEnumerable<UserSecurityPolicy[]> EqualsReturnsFalse_Data()
```
joelverhagen: >EqualsReturnsFalse_Data [](start = 56, length = 23)

nit make this a property

src/NuGetGallery/Migrations/201705041614287_UserSecurityPolicies_SubscriptionColumn.cs
```diff
@@ -0,0 +1,18 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class UserSecurityPolicies_SubscriptionColumn : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.UserSecurityPolicies", "Subscription", c => c.String(nullable: false, maxLength: 256));
```
joelverhagen: >nullable [](start = 80, length = 8)

What will existing records do with this?

Or are we assuming there is no data?

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -0,0 +1,128 @@
...
+namespace NuGetGallery.Areas.Admin.Controllers
+{
+    /// <summary>
+    /// Controller for the security policy management Admin view.
+    /// </summary>
+    public class SecurityPolicyController : AdminControllerBase
+    {
+        public IEntitiesContext EntitiesContext { get; }
+
+        public ISecurityPolicyService PolicyService { get; }
```
joelverhagen: >PolicyService [](start = 38, length = 13)

nit why public?

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -0,0 +1,128 @@
...
+            var model = new SecurityPolicyViewModel()
+            {
+                SubscriptionNames = PolicyService.UserSubscriptions.Select(s => s.Name)
+            };
+
+            return View(model);
+        }
+
+        [HttpGet]
+        public virtual ActionResult Search(string query)
```
joelverhagen: >Search [](start = 36, length = 6)

thoughts on unit testing this?

joelverhagen: Ideally we could mock the context and policy service

---
In reply to: [115030333](https://github.com/NuGet/NuGetGallery/pull/3854#discussion_r115030333) [](ancestors = 115030333)

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            // Parse query and look for users in the DB.
+            var usernames = GetUsernamesFromQuery(query);
+            var users = FindUsers(usernames);
+            var usersNotFound = usernames
+                .Where(name => !users.Any(u => u.Username.Equals(name, StringComparison.OrdinalIgnoreCase)))
```
joelverhagen: >Where [](start = 17, length = 5)

linq except might be a bit clearer for this

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -52,12 +86,108 @@ public SecurityPolicyResult Evaluate(SecurityPolicyAction action, HttpContextBas
...
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+            if (subscription == null)
+            {
+                throw new ArgumentNullException(nameof(subscription));
+            }
+            
+            if (!IsSubscribed(user, subscription))
```
joelverhagen: >IsSubscribed [](start = 17, length = 12)

This isn't thread safe. Two simultaneous enrolls to the same subscription would result in duplicate policies right? Or is there a optimistic concurrency on the user?

src/NuGetGallery/Security/SecurePushSubscription.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+        
+        public void OnUnsubscribe(User user)
+        {
+        }
+
+        /// <summary>
+        /// Expire API keys with push capability on secure push enrollment.
+        /// </summary>
+        private static void SetPushApiKeysToExpire(User user, int expirationInDays = 7)
```
joelverhagen: >= 7 [](start = 83, length = 3)

why optional parameter? there is only one acceptable value here today right? seems like a private const more correct here

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -14,14 +15,47 @@ namespace NuGetGallery.Security
     /// </summary>
     public class SecurityPolicyService : ISecurityPolicyService
```
joelverhagen: >SecurityPolicyService [](start = 17, length = 21)

I don't see any logging around these actions. Are we thinking that auditing is enough?

joelverhagen:
:clock1:

src/NuGetGallery/Migrations/201705041614287_UserSecurityPolicies_SubscriptionColumn.cs
chenriksson: No data in the table yet.

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// </summary>
+    public class SecurityPolicyController : AdminControllerBase
+    {
+        public IEntitiesContext EntitiesContext { get; }
+
+        public ISecurityPolicyService PolicyService { get; }
+
+        public SecurityPolicyController(IEntitiesContext entitiesContext, ISecurityPolicyService policyService)
+        {
+            if (entitiesContext == null)
```
skofman1: nit: c# 7 introduces throw expressions: https://blogs.msdn.microsoft.com/dotnet/2016/08/24/whats-new-in-csharp-7-0/

src/NuGetGallery/Migrations/201705041614287_UserSecurityPolicies_SubscriptionColumn.cs
joelverhagen: Sounds good. Thanks.

src/NuGetGallery/ExtensionMethods.cs
```diff
@@ -461,8 +461,7 @@ public static Credential GetCurrentApiKeyCredential(this User user, IIdentity id
             var claimsIdentity = identity as ClaimsIdentity;
             var apiKey = claimsIdentity.GetClaimOrDefault(NuGetClaims.ApiKey);
 
-            return string.IsNullOrEmpty(apiKey) ? null
-                : user.Credentials.FirstOrDefault(c => c.Value == apiKey);
+            return user.Credentials.FirstOrDefault(c => c.Value == apiKey);
```
skofman1: What will be the behavior if we authenticated with a credential that is not an API key?

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
chenriksson: I think I tried in an earlier commit and had a CI issue... but will try it.

src/NuGetGallery/Security/SecurityPolicyService.cs
skofman1: I'm afraid that's the case for a lot of our CUD actions... for scoped API keys I added "locks" in the page itself to prevent double clicks from creating duplicate calls.

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
scottbommarito: Make sure you have the changes to `build.ps1` from [this commit](https://github.com/NuGet/NuGetGallery/commit/f86270976da0eaa960c5bc9b69a3ce7857bfa1c6). If it still fails, run it on the build machines queue and not the hosted agent queue.

src/NuGetGallery.Core/Entities/UserSecurityPolicy.cs
chenriksson: done

src/NuGetGallery/Security/SecurityPolicyService.cs
chenriksson: Was planning to do auditing and telemetry together in next PR.

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
chenriksson: @joelverhagen Tests added!

src/NuGetGallery/Security/SecurityPolicyService.cs
chenriksson: The confirmation prompt should help guard against duplicate clicks. I don't think a lock in the gallery (per instance) or optimistic concurrency in the DB (per row) is the solution here.

Now that we have the Subscription column, we could add a unique index (Name, Subscription)... assuming we wouldn't have a subscription contain multiple of the same policies differing in value only. We could silently ignore duplicate inserts (SqlException 2601). Thoughts?

src/NuGetGallery/ExtensionMethods.cs
chenriksson: The apiKey var would be empty/null, and user.Credentials.FirstOrDefault would return null. Same behavior, so the IsNullOrEmpty check is unnecessary.

This is actually a regression from PR #3855, since some test infra uses empty API keys.

src/NuGetGallery/Security/SecurityPolicyService.cs
joelverhagen: `(Name, Subscription)` unique index seems right. That being said, it seems like the "is enrolled?" and unenroll code paths handle duplicates just fine.

PR #3855, opened 2017-05-02
Key creation telemetry

chenriksson:
Per @anangaur, additional telemetry necessary to track temp key usage

### DIFF
--- a/src/NuGetGallery/ExtensionMethods.cs
+++ b/src/NuGetGallery/ExtensionMethods.cs
@@ -453,6 +453,18 @@ namespace NuGetGallery
             return user;
         }
 
+        /// <summary>
+        /// Get the current API key credential, if available.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/ExtensionMethods.cs
```diff
@@ -453,6 +453,18 @@ public static User GetCurrentUser(this IOwinContext self)
             return user;
         }
 
+        /// <summary>
+        /// Get the current API key credential, if available.
+        /// </summary>
+        public static Credential GetCurrentApiKeyCredential(this User user, IIdentity identity)
+        {
+            var claimsIdentity = identity as ClaimsIdentity;
```
joelverhagen: Why do we need extension method here? Could it be on the model itself? Perhaps `IIdentity` isn't available there?

chenriksson: Yes, identity isn't available to the model.

PR #3860, opened 2017-05-03
Support MinClientVersion <= 4.1.0.0

maartenba:
Fixes https://github.com/NuGet/Home/issues/5156

### DIFF
--- a/src/NuGetGallery/Constants.cs
+++ b/src/NuGetGallery/Constants.cs
@@ -21,7 +21,7 @@ namespace NuGetGallery
         public const double AllowedLoginAttempts = 10;
 
         public const int MaxEmailSubjectLength = 255;
-        internal static readonly NuGetVersion MaxSupportedMinClientVersion = new NuGetVersion("3.4.0.0");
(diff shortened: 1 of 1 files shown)


emgarten:
4.1.0 is out already
4.3.0 is available in the CLI preview builds

Why does the gallery limit this? Just to verify the package is sane?

joelverhagen:
Perhaps we should make max exclusive and switch to `5.0.0--`. For just say the major part must be less than or equal to 4. This seems like an overly strict requirement.

skofman1:
Moved the issue to NuGetGallery repo: https://github.com/NuGet/NuGetGallery/issues/3866

maartenba:
> 4.1.0 is out already
> 4.3.0 is available in the CLI preview builds

Unfortunately, there is no 4.1 of any of the client libraries on NuGet.org?

skofman1:
@maartenba , why do we need MinClient version check at all? What was the intent on having it as part of the package validation process? I'm trying to understand if we should have an additional issue to remove this validation all together.

PR #3869, opened 2017-05-03
SecurityPolicies migration fix

### LINKED ISSUE #3865: Update database error during Setup-DevEnvironment.ps1
Got the following error when running Setup-DevEnvironment.ps1

`ERROR: Unable to update database to match the current model because there are pending changes and automatic migration is
 disabled. Either write the pending model changes to a code-based migration or enable automatic migration. Set DbMigrati
onsConfiguration.AutomaticMigrationsEnabled to true to enable automatic migration.`

To find out what model is causing this, I went to Visual Studio package manager console, run Add-Migration to see what will be generated:
```
namespace NuGetGallery.Migrations
{
    using System;
    using System.Data.Entity.Migrations;
    
    public partial class WhatIsMissing : DbMigration
    {
        public override void Up()
        {
            CreateTable(
                "dbo.UserSecurityPolicies",
                c => new
                    {
                        Key = c.Int(nullable: false, identity: true),
                        UserKey = c.Int(nullable: false),
                        Name = c.String(nullable: false),
                        Value = c.String(),
                    })
                .PrimaryKey(t => t.Key)
                .ForeignKey("dbo.Users", t => t.UserKey, cascadeDelete: true)
                .Index(t => t.UserKey);
            
        }
        
        public override void Down()
        {
            DropForeignKey("dbo.UserSecurityPolicies", "UserKey", "dbo.Users");
            DropIndex("dbo.UserSecurityPolicies", new[] { "UserKey" });
            DropTable("dbo.UserSecurityPolicies");
        }
    }
}
```

Looks like UserSecurityPolicies is not considered migrated to the database. Any idea on why?

chenriksson:
Fixes #3865


chenriksson:
Per our discussion, will fix in follow-up migration instead. @skofman1 @ryuyu

Issue #3792, opened 2017-04-17
report abuse form changes


karann-msft:
1.	Request form changes – change the dropdown option to say The package violates a Copyright I own instead of License. 
2.	At the bottom of this form, add the line “You can also send us an email at admin@nuget.org"
3.	Required Statements should only show up when the  "The package violates a Copyright I own" option is selected

skofman1:
Posting the e-mail publicly will allow spammers to use it..

anangaur:
As per discussion with legal, this seems to be a requirement for DMCA requirements.

PR #3873, opened 2017-05-04
https://github.com/NuGet/NuGetGallery/issues/3792

karann-msft:
Based on discussion with legal, updating that description


skofman1:
This issue: https://github.com/NuGet/NuGetGallery/issues/3792
specified 3 changes we need to apply. Lets have them all in the same PR.

PR #3887, opened 2017-05-05
Show full version on package details page

xavierdecoster:
Only display the full version in the title.
Use NormalizedVersion anywhere else.

image

### DIFF
--- a/src/NuGetGallery/ViewModels/PackageViewModel.cs
+++ b/src/NuGetGallery/ViewModels/PackageViewModel.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Linq;
 using System.Collections.Generic;
+using NuGet.Versioning;
 
 namespace NuGetGallery
 {
@@ -10,13 +11,26 @@ namespace NuGetGallery
     {
         private readonly Package _package;
         private string _pendingTitle;
+        private string _fullVersion;
 
         public PackageViewModel(Package package)
         {
             _package = package;
+
+            NuGetVersion nugetVersion;
+            if(NuGetVersion.TryParse(package.Version, out nugetVersion))
+            {
+                _fullVersion = nugetVersion.ToFullString();
+            }
+            else
+            {
+                _fullVersion = string.Empty;
+            }
+
             Version = String.IsNullOrEmpty(package.NormalizedVersion) ?
                 NuGetVersionNormalizer.Normalize(package.Version) :
                 package.NormalizedVersion;
+            
             Description = package.Description;
             ReleaseNotes = package.ReleaseNotes;
             IconUrl = package.IconUrl;
@@ -69,6 +83,7 @@ namespace NuGetGallery
         }
 
         public string Version { get; set; }
+        public string FullVersion => _fullVersion;
 
         public string Title
         {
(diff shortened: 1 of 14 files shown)


joelverhagen:
Why no full version in the version list at the bottom of the page? 

Also, is this full version visible on web UI package upload or package edit?

xavierdecoster:
> Why no full version in the version list at the bottom of the page?

Full Version may be quite large (I've seen people using metadata parts of 50+ characters), so wouldn't want to clutter the table. As the URL for those hyperlink uses NormalizedVersion anyway, I figured it made sense to keep those as is.

Same for the version in the left side bar, don't want it to flow through in other text.

> Also, is this full version visible on web UI package upload or package edit?

That is already the case. Try uploading one ;)

joelverhagen:
Sounds good. Thanks.

src/NuGetGallery/ViewModels/PackageViewModel.cs
```diff
@@ -3,20 +3,34 @@
...
             _package = package;
+
+            NuGetVersion nugetVersion;
+            if(NuGetVersion.TryParse(package.Version, out nugetVersion))
+            {
+                _fullVersion = nugetVersion.ToFullString();
+            }
+            else
+            {
+                _fullVersion = string.Empty;
```
scottbommarito: If this parsing fails, the UI will show no version, which shouldn't be possible.

A couple options:
1 - fall back to what `Version` shows
2 - throw an exception and fail
3 - print `package.Version` directly ([which we currently do](https://github.com/NuGet/NuGetGallery/blob/29b3aa1e57c2f4aa3e990b198d7fc9dd7029d724/src/NuGetGallery.Core/NuGetVersionExtensions.cs#L17))

src/NuGetGallery/ViewModels/PackageViewModel.cs
```diff
@@ -3,20 +3,34 @@
...
         private readonly Package _package;
         private string _pendingTitle;
+        private string _fullVersion;
 
         public PackageViewModel(Package package)
         {
             _package = package;
+
+            NuGetVersion nugetVersion;
+            if(NuGetVersion.TryParse(package.Version, out nugetVersion))
```
scottbommarito: Should we move this to [NuGetVersionExtensions](https://github.com/NuGet/NuGetGallery/blob/29b3aa1e57c2f4aa3e990b198d7fc9dd7029d724/src/NuGetGallery.Core/NuGetVersionExtensions.cs)?

`NuGetFullVersioner`?

src/NuGetGallery/ViewModels/PackageViewModel.cs
xavierdecoster: Addressed with https://github.com/NuGet/NuGetGallery/pull/3887/commits/2767721f09aa40b9063845756dd6250c4f94a147

src/NuGetGallery/ViewModels/PackageViewModel.cs
xavierdecoster: Addressed with https://github.com/NuGet/NuGetGallery/pull/3887/commits/2767721f09aa40b9063845756dd6250c4f94a147

PR #3893, opened 2017-05-08
Highlight semver2 packages on package details view

xavierdecoster:
Client version to be confirmed.

Preview:

image

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -220,7 +220,7 @@
     @if (Model.IsSemVer2)
     {
         <p class="message warning">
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -217,6 +217,14 @@
         </p>
     }
 
+    @if (Model.IsSemVer2)
+    {
+        <p class="message warning">
+            This package is only available for download with NuGet client version 4.3.0 and above. 
+            Reach out to support if you have more questions.
```
joelverhagen: Ideally we would have docs about this, rather having to contact support.

xavierdecoster:
@joelverhagen the message I used here is verbatim message I received from @anangaur (except for the version number, that's to be verified I guess)

PR #3934, opened 2017-05-10
Added index on ([SemVerLevelKey], [IsLatest], [Deleted]) on Packages table

agr:
The 
```
(@p__linq__0 int)SELECT 
    [GroupBy1].[A1] AS [C1]
    FROM ( SELECT 
        COUNT(1) AS [A1]
        FROM [dbo].[Packages] AS [Extent1]
        WHERE ([Extent1].[Deleted] <> 1) AND (([Extent1].[SemVerLevelKey] = @p__linq__0) OR (([Extent1].[SemVerLevelKey] IS NULL) AND (@p__linq__0 IS NULL))) AND ([Extent1].[IsLatest] = 1)
    )  AS [GroupBy1]
```
query now tops as the worst performing query, following the Azure's performance recommendation, creating an index for that query.


joelverhagen:
Please make sure we add a step the deployment issue about running this migration.

src/NuGetGallery/Migrations/201705102016311_AddIndexSemVerLevelKeyDeletedIsLatest.cs
```diff
@@ -0,0 +1,21 @@
...
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddIndexSemVerLevelKeyDeletedIsLatest : DbMigration
+    {
+        private const string IndexName = "nci_Packages_SemVerIsLatestDeleted";
+
+        public override void Up()
+        {
+            Sql($"IF NOT EXISTS(SELECT * FROM sys.indexes WHERE name = '{IndexName}' AND object_id = OBJECT_ID('Packages')) " +
+                $"CREATE NONCLUSTERED INDEX [{IndexName}] ON [dbo].[Packages]([SemVerLevelKey], [IsLatest], [Deleted]) ");
```
xavierdecoster: Misses `WITH (ONLINE = ON)`. You don't want the query to time-out when creating this index on PROD, and instead want to create this index online on the db server.

If you revert this change and cherry pick from `feature/semver2`, you'll get this for free ;)
https://github.com/NuGet/NuGetGallery/pull/3867/files#diff-a713b5aadf42ca59c0c90263d6d72100R10

agr: `WITH(ONLINE=ON)` does not work with SQL Express Local DB...

xavierdecoster: True, which is why, in the past, we always removed that part _after_ applying this migration to PROD.

PR #3937, opened 2017-05-10
Add initial attempt of new front page

joelverhagen:
screencapture-localhost-1494456512560


skofman1:
Please don't call folders "new". Give it a name we don't need to change later, when we delete the old stuff 😆

joelverhagen:
> Please don't call folders "new". Give it a name we don't need to change later, when we delete the old stuff 

Great idea! That will really save a lot of work.

PR #3939, opened 2017-05-11
Fix Focus Visibility

ryuyu:
We were setting outline: none in a few places, which caused many elements to not show when they had focus. Set a :focus style globally, and override where necessary.

@xavierdecoster, @skofman1 Is there a reason this was done in these places? I don't want to mess up any explicit decisions we made.

@joelverhagen

### DIFF
--- a/src/NuGetGallery/Content/Site.css
+++ b/src/NuGetGallery/Content/Site.css
@@ -231,12 +231,16 @@ th {
     color: #ff9600;
 }
 
-.warning.fancy { 
-    color: black;
-    border-color: #f7d3a5;
-    font-size: 1em;
-    font-weight: 500;
-    padding: 5px;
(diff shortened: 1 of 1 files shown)


xavierdecoster:
@ryuyu no idea about past decisions made before I worked on gallery :)

src/NuGetGallery/Content/Site.css
```diff
@@ -231,12 +231,16 @@ th {
     color: #ff9600;
 }
 
-.warning.fancy { 
-    color: black;
-    border-color: #f7d3a5;
-    font-size: 1em;
-    font-weight: 500;
-    padding: 5px;
+    .warning.fancy {
```
joelverhagen: why indent?

PR #3945, opened 2017-05-11
Temp keys telemetry and auditing

chenriksson:
Auditing and telemetry for user security policies (subscription, unsubscription, failed evaluations)


src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -70,16 +80,34 @@ public SecurityPolicyResult Evaluate(SecurityPolicyAction action, HttpContextBas
                 var foundPolicies = user.SecurityPolicies.Where(p => p.Name.Equals(handler.Name, StringComparison.OrdinalIgnoreCase));
                 if (foundPolicies.Any())
                 {
-                    var result = handler.Evaluate(new UserSecurityPolicyContext(httpContext, foundPolicies));
+                    var result = handler.Evaluate(new UserSecurityPolicyEvaluationContext(httpContext, foundPolicies));
                     if (!result.Success)
                     {
+                        Auditing.SaveAuditRecordAsync(new FailedUserSecurityPolicyAuditRecord(
+                            user.Username, GetAuditAction(action), foundPolicies)).Wait();
```
chenriksson: Want to call out this Wait() -- any better suggestions? The issue is that policy evaluation is done from an action filter, but MVC5 filters don't support async.

Alternatives, which I'm not fond of:
-Move policy evaluation inside action methods
-Fire SaveAuditRecordAsync, but don't wait
-Create the audit record here, but write from an async handler or something?

src/NuGetGallery/Security/ISecurityPolicyService.cs
```diff
@@ -14,6 +16,16 @@ namespace NuGetGallery.Security
     public interface ISecurityPolicyService
     {
         /// <summary>
+        /// Auditing for the security policy service.
+        /// </summary>
+        IAuditingService Auditing { get; }
```
joelverhagen: >Auditing [](start = 25, length = 8)

Commented offline about this, dependencies should not be on the interface

joelverhagen:
>     public enum AuditedUserAction

Is this enum serialized somewhere? If so, is it serialized as int or string?

---
Refers to: src/NuGetGallery.Core/Auditing/AuditedUserAction.cs:6 in 9358458. [](commit_id = 93584580d567d9f8ec89910a8b82c905ed639cac, deletion_comment = False)

src/NuGetGallery.Core/Auditing/FailedUserSecurityPolicyAuditRecord.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using NuGetGallery.Auditing.AuditedEntities;
+
+namespace NuGetGallery.Auditing
+{
+    /// <summary>
+    /// Audit record for failed user security policy evaluations.
+    /// </summary>
+    public class FailedUserSecurityPolicyAuditRecord : AuditRecord<AuditedPackageAction>
```
joelverhagen: >FailedUserSecurityPolicyAuditRecord [](start = 17, length = 35)

I'm loving these read-only types. Very nice.

src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
```diff
@@ -16,6 +17,7 @@ public class UserAuditRecord : AuditRecord<AuditedUserAction>
         public CredentialAuditRecord[] Credentials { get; }
         public CredentialAuditRecord[] AffectedCredential { get; }
         public string AffectedEmailAddress { get; }
+        public AuditedUserSecurityPolicy[] Policies { get; }
```
joelverhagen: >AuditedUserSecurityPolicy [](start = 15, length = 25)

Nit, if it's all the same we should be using IEnumerable or IReadOnlyList if you need indexing. Arrays are mutable which is probably not necessary here.

src/NuGetGallery/Security/SecurityPolicyService.cs
skofman1: I would recommend making Evaluate async, and moving the Wait() to the attribute. In this case, if Evaluate is called from a different location in the future (say, a controller), it can be called correctly

---
In reply to: [116026021](https://github.com/NuGet/NuGetGallery/pull/3945#discussion_r116026021) [](ancestors = 116026021)

src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
```diff
@@ -12,7 +12,6 @@ public enum AuditedPackageAction
         Unlist,
         Edit,
         UndoEdit,
-
-        
+        Verify
```
skofman1: Shims will need to be updated as well

src/NuGetGallery/Security/SecurityPolicyService.cs
joelverhagen: Yes agreed. Make as much of the stack async. If the filter is the thing that can't be async, that's the place that should have the `.Result` or `.Wait()`.

---
In reply to: [116067013](https://github.com/NuGet/NuGetGallery/pull/3945#discussion_r116067013) [](ancestors = 116067013,116026021)

joelverhagen:
We should wait till @dtivel fills us in the process of updating our internal auditing implementation. As it stands today, it seems like we should have both PRs ready and reviewed before we merge to `dev`.

joelverhagen:
:clock1:

chenriksson:
@joelverhagen action enum gets serialized as string in audit record. Also added audit record for Verify per our discussion.

This PR is against tempkeys feature branch, but agree I should have Shims PR ready before merging this to dev.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -245,7 +245,7 @@ public virtual Task<ActionResult> GetNuGetExe()
             return (ActionResult)result ?? new EmptyResult();
         }
 
-        private HttpStatusCodeWithBodyResult VerifyPackageKeyInternal(User user, Credential credential, string id, string version)
+        private async Task<HttpStatusCodeWithBodyResult> VerifyPackageKeyInternal(User user, Credential credential, string id, string version)
```
cristinamanum: Consider adding "async" in the method name.

src/NuGetGallery/Filters/ApiAuthorizeAttribute.cs
```diff
@@ -91,7 +91,10 @@ protected override bool AuthorizeCore(HttpContextBase httpContext)
             // If ApiKey authorization succeeds, evaluate any security policies.
             if (authorizeResult && SecurityPolicyAction.HasValue)
             {
-                SecurityPolicyResult = SecurityPolicyService.Evaluate(SecurityPolicyAction.Value, httpContext);
+                var evaluateTask = SecurityPolicyService.EvaluateAsync(SecurityPolicyAction.Value, httpContext);
+                evaluateTask.Wait();
+
+                SecurityPolicyResult = evaluateTask.Result;
```
cristinamanum: .Result will do the wait. There is no need for the line 95.

src/NuGetGallery/Security/SecurePushSubscription.cs
```diff
@@ -41,38 +47,46 @@ public IEnumerable<UserSecurityPolicy> Policies
...
+            var pushKeys = context.User.Credentials.Where(c =>
                 CredentialTypes.IsApiKey(c.Type) &&
                 (
                     c.Scopes.Count == 0 ||
                     c.Scopes.Any(s =>
                         s.AllowedAction.Equals(NuGetScopes.PackagePush, StringComparison.OrdinalIgnoreCase) ||
                         s.AllowedAction.Equals(NuGetScopes.PackagePushVersion, StringComparison.OrdinalIgnoreCase)
                         ))
-                );
+                ).ToList();
```
cristinamanum: Why was the ToList needed?

src/NuGetGallery.Core/Auditing/FailedUserSecurityPolicyAuditRecord.cs
```diff
@@ -0,0 +1,45 @@
...
+                throw new ArgumentException(nameof(policies));
+            }
+
+            Username = username;
+            Policies = policies.Select(p => new AuditedUserSecurityPolicy(p)).ToArray();
+        }
+
+        public override string GetPath()
+        {
+            return Path; // store in <auditpath>/failedusersecuritypolicy/all
```
skofman1: does this path mean the blob storage folder where this audit record is stored? In that case why not save it in the user's folder?

src/NuGetGallery.Core/Auditing/UserAuditAction.cs
```diff
@@ -1,16 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-namespace NuGetGallery.Auditing
-{
-    public enum UserAuditAction
```
skofman1: why delete?

src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
```diff
@@ -55,7 +57,13 @@ public UserAuditRecord(User user, AuditedUserAction action, string affectedEmail
         {
             AffectedEmailAddress = affectedEmailAddress;
         }
-        
+
+        public UserAuditRecord(User user, AuditedUserAction action, IEnumerable<UserSecurityPolicy> policies)
+            : this(user, action, Enumerable.Empty<Credential>())
+        {
+            Policies = policies.Select(p => new AuditedUserSecurityPolicy(p)).ToArray();
```
skofman1: policies are available as part of the user object. Why pass them separately?

skofman1:
:clock1:

src/NuGetGallery.Core/Auditing/UserAuditAction.cs
chenriksson: Was dead code; UserAuditRecord uses [AuditedUserAction](https://github.com/NuGet/NuGetGallery/blob/9efe2950f41fb1028e48d8c72cefe7363f4be678/src/NuGetGallery.Core/Auditing/AuditedUserAction.cs) instead.

Though, I should probably use present tense verbs for actions to match the others.

src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
chenriksson: I should rename AffectedPolicies and add a comment. These only include the policies that are currently being subscribed to or unsubscribed from.

skofman1:
Like was discussed with @joelverhagen and @dtivel , we should have auditing for successful policy validation as well.

.vs/config/applicationhost.config
```diff
@@ -155,7 +155,7 @@
         <sites>
             <site name="NuGet Gallery (nuget.localtest.me)" id="2">
                 <application path="/">
-                    <virtualDirectory path="/" physicalPath="..\..\src\NuGetGallery" />
+                    <virtualDirectory path="/" physicalPath="D:\Nuget\NuGetGallery\src\NuGetGallery" />
```
dtivel: Should this file be checked in?  This path doesn't exist on my machine.

Perhaps add this file to .gitignore.

src/NuGetGallery.Core/Auditing/FailedUserSecurityPolicyAuditRecord.cs
```diff
@@ -0,0 +1,43 @@
...
+            IEnumerable<UserSecurityPolicy> policies)
+            :base(action)
+        {
+            if (string.IsNullOrEmpty(username))
+            {
+                throw new ArgumentNullException(nameof(username));
+            }
+            if (policies == null || policies.Count() == 0)
+            {
+                throw new ArgumentException(nameof(policies));
```
dtivel: This constructor takes a message.  [This other constructor](https://msdn.microsoft.com/en-us/library/sxykka64(v=vs.110).aspx) takes a message and parameter name.

src/NuGetGallery.Core/Auditing/FailedUserSecurityPolicyAuditRecord.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using NuGetGallery.Auditing.AuditedEntities;
+
+namespace NuGetGallery.Auditing
+{
+    /// <summary>
+    /// Audit record for failed user security policy evaluations.
+    /// </summary>
+    public class FailedUserSecurityPolicyAuditRecord : AuditRecord<AuditedSecurityPolicyAction>
```
dtivel: FailedSecurityPolicyEvaluationAuditRecord or FailedUserSecurityPolicyEvaluationAuditRecord?

src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
```diff
@@ -17,6 +18,11 @@ public class UserAuditRecord : AuditRecord<AuditedUserAction>
         public CredentialAuditRecord[] AffectedCredential { get; }
         public string AffectedEmailAddress { get; }
 
+        /// <summary>
+        /// Subset of user policies affected by the action (subscription / unsubscription).
+        /// </summary>
+        public AuditedUserSecurityPolicy[] AffectedPolicies { get; }
```
dtivel: Be sure to update shims.

src/NuGetGallery/Security/ISecurityPolicyService.cs
```diff
@@ -39,6 +39,6 @@ public interface ISecurityPolicyService
         /// <param name="action">Security policy action.</param>
         /// <param name="context">Authorization context.</param>
         /// <returns>Policy result indicating success or failure.</returns>
-        SecurityPolicyResult Evaluate(SecurityPolicyAction action, HttpContextBase httpContext);
+        Task<SecurityPolicyResult> EvaluateAsync(SecurityPolicyAction action, HttpContextBase httpContext);
```
dtivel: /// <returns>A task that represents the asynchronous operation.
        /// The task result (<see cref="Task{TResult}.Result" />) returns a <see cref="SecurityPolicyResult" />
        /// instance.</returns>

src/NuGetGallery/Security/ISecurityPolicyService.cs
```diff
@@ -39,6 +39,6 @@ public interface ISecurityPolicyService
         /// <param name="action">Security policy action.</param>
         /// <param name="context">Authorization context.</param>
```
dtivel: `httpContext`

src/NuGetGallery/Security/SecurePushSubscription.cs
```diff
@@ -41,21 +47,19 @@ public IEnumerable<UserSecurityPolicy> Policies
...
-        public void OnSubscribe(User user)
-        {
-            SetPushApiKeysToExpire(user);
-        }
-        
-        public void OnUnsubscribe(User user)
+        public SecurePushSubscription(IAuditingService auditing, IDiagnosticsService diagnostics)
         {
+            Auditing = auditing ?? throw new ArgumentNullException(nameof(auditing));
+            Diagnostics = diagnostics?.SafeGetSource(nameof(SecurePushSubscription)) ?? throw new ArgumentNullException(nameof(diagnostics));
```
dtivel: Shouldn't throw an `ArgumentNullException` for `diagnostics` if `diagnostics` is not `null` and `SafeGetSource(...)` returns `null`.

Consider changing the parameter type from `IDiagnosticsService` to `IDiagnosticsSource`.

src/NuGetGallery/Security/SecurePushSubscription.cs
```diff
@@ -41,21 +47,19 @@ public IEnumerable<UserSecurityPolicy> Policies
...
         }
 
         /// <summary>
-        /// Expire API keys with push capability on secure push enrollment.
+        /// On subscribe, set API keys with push capability to expire in 30 days.
```
dtivel: `...to expire in <see cref="PushKeysExpirationInDays" /> days.`

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -23,13 +26,20 @@ public class SecurityPolicyService : ISecurityPolicyService
...
         {
         }
 
-        public SecurityPolicyService(IEntitiesContext entitiesContext)
+        public SecurityPolicyService(IEntitiesContext entitiesContext, IAuditingService auditing, IDiagnosticsService diagnostics)
         {
             EntitiesContext = entitiesContext ?? throw new ArgumentNullException(nameof(entitiesContext));
+            Auditing = auditing ?? throw new ArgumentNullException(nameof(auditing));
+            Diagnostics = diagnostics?.SafeGetSource("SecurityPoliceService")
+                ?? throw new ArgumentNullException(nameof(diagnostics));
```
dtivel: Same as before:  don't throw an `ArgumentNullException` for 'diagnostics` if `diagnostics` is not null.

chenriksson:
@skofman1 @joelverhagen Added auditing of successful evaluation as well

src/NuGetGallery/Security/SecurePushSubscription.cs
```diff
@@ -18,6 +21,9 @@ public class SecurePushSubscription : IUserSecurityPolicySubscription
         private const string MinClientVersion = "4.1.0";
         private const int PushKeysExpirationInDays = 30;
 
+        private IAuditingService Auditing { get; }
```
joelverhagen: >IAuditingService [](start = 16, length = 16)

nit: typically these are private readonly fields rather than props

tests/NuGetGallery.Facts/Security/SecurePushSubscriptionFacts.cs
```diff
@@ -40,10 +45,10 @@ public void SubscriptionPolicies()
         [InlineData("")]
         [InlineData("[{\"a\":\"package:push\", \"s\":\"theId\"}]")]
         [InlineData("[{\"a\":\"package:pushversion\", \"s\":\"theId\"}]")]
-        public void OnSubscribeExpiresPushApiKeysIn1Week(string scopes)
+        public async void OnSubscribeExpiresPushApiKeysIn1Week(string scopes)
```
dtivel: Use `async Task` instead of `async void`.

https://msdn.microsoft.com/en-us/magazine/jj991977.aspx

src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
chenriksson: @joelverhagen Left as arrays, which is consistent with other audit records properties. I'd prefer to be consistent for now, and revisit as a separate item since it's general feedback for all auditing. Added this feedback to #3946.

tests/NuGetGallery.Facts/Security/SecurityPolicyServiceFacts.cs
```diff
@@ -56,73 +58,92 @@ public void UserHandlers()
         }
 
         [Fact]
-        public void EvaluateThrowsIfHttpContextNull()
+        public async void EvaluateThrowsIfHttpContextNull()
```
dtivel: As a general rule I like to see test names in the format:

    MemberName_Expectation

For example, this test would be:

    EvaluateAsync_ThrowsForNullHttpContext

Then it is very clear what is being tested and what the expectation is.  Without this and with refactorings over time you can be left with tests where the original intent isn't clear.

The previous test ([UserHandlers](https://github.com/NuGet/NuGetGallery/pull/3945/files#diff-b7375d025bad5bfabb18db49ed816c3dR43)) doesn't follow this format, so the intent of the test isn't stated.  Other tests (like [this one](https://github.com/NuGet/NuGetGallery/pull/3945/files#diff-796ec632e669060b29479461e7585604R62)) don't seem to accurately describe the member under test.  It may, but from a cursory look I didn't see `OnSubscribe` anywhere in the test.  Shouldn't it be `SubscribeAsync_DoesNotExpireNonPushCredentials`?

Can you please review all your test names to ensure they are both accurate and clearly named?  Most of them look pretty good with respect to stating expectations, but the name of the member under test isn't always accurate and some expectations are just missing.

tests/NuGetGallery.Core.Facts/Auditing/UserSecurityPolicyAuditRecordFacts.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace NuGetGallery.Auditing
+{
+    public class UserSecurityPolicyAuditRecordFacts
```
dtivel: Need more tests than just null/empty argument validation.  What about verifying that the constructor initializes properties correctly?  There is [non-trivial code](https://github.com/NuGet/NuGetGallery/pull/3945/files#diff-a132dddf74a5ae6705ca94dce0752f98R40) in the constructor that is testable.

What about testing the `GetPath(...)` override?

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -70,16 +80,37 @@ public SecurityPolicyResult Evaluate(SecurityPolicyAction action, HttpContextBas
...
+
                         return result;
                     }
                 }
             }
+
             return SecurityPolicyResult.SuccessResult;
         }
 
+        private AuditedSecurityPolicyAction GetAuditAction(SecurityPolicyAction policyAction)
```
dtivel: Do you have `EvaluateAsync(...)` tests that cover each case below?

src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
```diff
@@ -55,7 +61,13 @@ public UserAuditRecord(User user, AuditedUserAction action, string affectedEmail
         {
             AffectedEmailAddress = affectedEmailAddress;
         }
-        
+
+        public UserAuditRecord(User user, AuditedUserAction action, IEnumerable<UserSecurityPolicy> affectedPolicies)
```
dtivel: Please add tests for this constructor in UserAuditRecordTests.cs.

tests/NuGetGallery.Facts/Security/SecurePushSubscriptionFacts.cs
```diff
@@ -17,15 +19,18 @@ public class SecurePushSubscriptionFacts
         [Fact]
         public void SubscriptionName()
         {
+            // Arrange.
+            var subscription = CreateSecurityPolicyService().UserSubscriptions.First();
```
dtivel: Would `Single()` be a stronger test?  The same feedback applies to subsequent tests.

src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
joelverhagen: Yup, sounds great.

src/NuGetGallery/Security/SecurePushSubscription.cs
```diff
@@ -65,14 +69,24 @@ private static void SetPushApiKeysToExpire(User user)
                         ))
                 );
 
+            var expires = DateTime.UtcNow.AddDays(PushKeysExpirationInDays);
             foreach (var key in pushKeys)
             {
-                var expires = DateTime.UtcNow.AddDays(PushKeysExpirationInDays);
                 if (!key.Expires.HasValue || key.Expires > expires)
                 {
+                    await _auditing.SaveAuditRecordAsync(
```
loic-sharma: From my understanding, this saves each audit record sequentially. Would it be better to kick off all save audit record tasks and then doing a `Task.WhenAll`?

src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
chenriksson: done

src/NuGetGallery/Security/SecurityPolicyService.cs
chenriksson: yes - reviewed

tests/NuGetGallery.Core.Facts/Auditing/UserSecurityPolicyAuditRecordFacts.cs
chenriksson: added

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -88,7 +88,8 @@ public virtual IEnumerable<IUserSecurityPolicySubscription> UserSubscriptions
                     var result = handler.Evaluate(new UserSecurityPolicyEvaluationContext(httpContext, foundPolicies));
 
                     await Auditing.SaveAuditRecordAsync(new UserSecurityPolicyAuditRecord(
-                        user.Username, GetAuditAction(action), foundPolicies, result.Success, result.ErrorMessage));
+                        user.Username, GetAuditAction(action), foundPolicies, result.Success, result.ErrorMessage))
+                        .ConfigureAwait(false);
```
loic-sharma: Can you explain this? From my understanding, this means you don't need to save the current context, but aren't you using the context's `result` directly after this await?

chenriksson: Mistake... let me correct it.

src/NuGetGallery.Core/Auditing/UserSecurityPolicyAuditRecord.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using NuGetGallery.Auditing.AuditedEntities;
+
+namespace NuGetGallery.Auditing
+{
+    /// <summary>
+    /// Audit record for failed user security policy evaluations.
```
skofman1: failed and succeeded policies evaluation

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -207,6 +207,10 @@ protected override void Load(ContainerBuilder builder)
                 .As<ISecurityPolicyService>()
                 .InstancePerLifetimeScope();
 
+            builder.RegisterType<SecurePushSubscription>()
```
skofman1: any reason not to have it as a SingleInstance() ? Same for SecurityPolicyService. The more we use singleton, the cheaper a request will be.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -254,7 +267,11 @@ private HttpStatusCodeWithBodyResult VerifyPackageKeyInternal(User user, Credent
                 return new HttpStatusCodeWithBodyResult(
                     HttpStatusCode.NotFound, String.Format(CultureInfo.CurrentCulture, Strings.PackageWithIdAndVersionNotFound, id, version));
             }
-            
+
+            // Write an audit record
+            await AuditingService.SaveAuditRecordAsync(
+                new PackageAuditRecord(package, AuditedPackageAction.Verify, "Verified via API."));
```
skofman1: is the string "Verified via API." needed? The action implies that already. Consider removing.

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -182,7 +233,7 @@ private static IEnumerable<UserSecurityPolicyHandler> CreateUserHandlers()
         /// </summary>
         private static IEnumerable<IUserSecurityPolicySubscription> CreateUserSubscriptions()
         {
-            yield return new SecurePushSubscription();
+            yield return DependencyResolver.Current.GetService<SecurePushSubscription>();
```
skofman1: why not use dependency injection?

PR #3950, opened 2017-05-11
Adding recommended index to the Packages table

agr:
Cherry-picked the bf59b12c7e1c13041a31256bdd1b2908c14b648b commit to add the recommended index.


src/NuGetGallery/Migrations/201705031714183_AddIndexSemVerLevelKey.cs
```diff
@@ -0,0 +1,18 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddIndexSemVerLevelKey : DbMigration
+    {
+        public override void Up()
+        {
+            Sql("IF NOT EXISTS(SELECT * FROM sys.indexes WHERE name = 'nci_wi_Packages_SemVerLevelKey' AND object_id = OBJECT_ID('Packages')) CREATE NONCLUSTERED INDEX [nci_wi_Packages_SemVerLevelKey] ON [dbo].[Packages] ([SemVerLevelKey], [IsLatest], [Deleted]) INCLUDE ([PackageRegistrationKey]) WITH (ONLINE = ON)");
```
agr: My concern here is that `WITH (ONLINE = ON)` won't work with the SQL Express LocalDB. Are there known workarounds?

scottbommarito: What do we do when creating our other indexes? Do we have `WITH (ONLINE = ON)` as well?

agr: nope

agr: Recent example: cd091c71cd6b340a617b857f6d6f3566c9b7d268

xavierdecoster: We always used Online=On to apply these in PROD, followed by removing the Online=On statement from the codebase, so that people without SqlServer Enterprise can just run the migrations as-is.

PR #3951, opened 2017-05-11
Make PackageDownloadsDetail and PackageDownloadsByVersion graphs update without refreshing the page

scottbommarito:
This is an accessibility fix that has developed into a user experience improvement in general.

Previously, the [PackageDownloadsDetail](https://www.nuget.org/stats/packages/newtonsoft.json/10.0.1) and [PackageDownloadsByVersion](https://www.nuget.org/stats/packages/newtonsoft.json) pages would refresh every time the selected dimensions were changed. This was obviously bad for accessibility, as the focus would be lost every time the page refreshed, which disorients users.

I have fixed this.

- Previously, the logic to inflate the table of information used by the browser was done in CSHTML, which is created server side. This logic is now done in JS using knockout (see `_PivotTable.cshtml`). We already had knockout in our project so I used the existing version.

- Added a route that returns the statistics report in JSON.

- Clicking upon a checkbox now makes an AJAX call that, on success, reinflates the information table using the statistics report in JSON. On error, it shows a message that asks the user if they would like to try again, which can be clicked to retry the AJAX call.

Tested on all the major browsers (`Firefox`, `Chrome`, `Edge`).

### DIFF
--- a/src/NuGetGallery/Controllers/StatisticsController.cs
+++ b/src/NuGetGallery/Controllers/StatisticsController.cs
@@ -149,7 +149,7 @@ namespace NuGetGallery
         }
 
         //
-        // GET: /stats/package/{id}
+        // GET: /stats/packages/{id}
 
         public virtual async Task<ActionResult> PackageDownloadsByVersion(string id, string[] groupby)
         {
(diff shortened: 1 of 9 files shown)


src/NuGetGallery/Controllers/StatisticsController.cs
```diff
@@ -158,6 +158,63 @@ public StatisticsController(IStatisticsService statisticsService, IAggregateStat
...
+
+        public virtual async Task<JsonResult> PackageDownloadsByVersionReport(string id, string[] groupby)
+        {
+            return Json(await GetPackageDownloadsByVersionReport(id, groupby), JsonRequestBehavior.AllowGet);
+        }
+
+        //
+        // GET: /stats/packages/{id}/{version}
+
+        public virtual async Task<ActionResult> PackageDownloadsDetail(string id, string version, string[] groupby)
```
skofman1: please add UTs for the new endpoints.

src/NuGetGallery/Controllers/StatisticsController.cs
```diff
@@ -158,6 +158,63 @@ public StatisticsController(IStatisticsService statisticsService, IAggregateStat
...
+
+            model.UseD3 = UseD3();
+
+            return View(model);
+        }
+
+        //
+        // GET: /stats/reports/packages/{id}
+
+        public virtual async Task<JsonResult> PackageDownloadsByVersionReport(string id, string[] groupby)
```
skofman1: consider adding basic functional tests to check those endpoints

scottbommarito: Existing functional tests have been adapted to check these new endpoints.

src/NuGetGallery/Controllers/StatisticsController.cs
```diff
@@ -158,6 +158,63 @@ public StatisticsController(IStatisticsService statisticsService, IAggregateStat
...
+
+            return View(model);
+        }
+
+        //
+        // GET: /stats/reports/packages/{id}/{version}
+
+        public virtual async Task<JsonResult> PackageDownloadsDetailReport(string id, string version, string[] groupby)
+        {
+            var report = await GetPackageDownloadsDetailReport(id, version, groupby);
```
chenriksson: Should the new methods also check if (_statisticsService == NullStatisticsService.Instance)?

src/NuGetGallery/Scripts/statsdimensions.js
```diff
@@ -1,5 +1,49 @@
...
+        url: baseUrl + query,
+        type: 'GET',
+        dataType: 'json',
+        success: renderGraphHandler,
+        error: function () {
+            renderGraphHandler(null);
+
+            $('#statistics-retry').click(function () {
+                renderGraph(baseUrl, query);
+            });
```
chenriksson: Does it need to register a new click callback on every failure?

scottbommarito: Yes, `query` can change.

src/NuGetGallery/Controllers/StatisticsController.cs
```diff
@@ -158,6 +158,63 @@ public StatisticsController(IStatisticsService statisticsService, IAggregateStat
                 return new HttpStatusCodeResult(HttpStatusCode.NotFound);
             }
 
+            var report = await GetPackageDownloadsByVersionReport(id, groupby);
+
+            StatisticsPackagesViewModel model = new StatisticsPackagesViewModel();
```
loic-sharma: var?

scottbommarito: was existing code, but i'll fix

src/NuGetGallery/Controllers/StatisticsController.cs
scottbommarito: will add

src/NuGetGallery/Controllers/StatisticsController.cs
scottbommarito: done

PR #3966, opened 2017-05-15
Hijack IsLatest(Stable)Version OData filter when semVerLevel=2.0.0

### LINKED ISSUE #3905: $filter=IsLatestVersion and semVerLevel=2.0.0 yields no results
The client does the following query when using the Browse tab is VS:
```http
GET https://dev.nugettest.org/api/v2/Search()?$filter=IsLatestVersion&searchTerm=''&targetFramework='net46'&includePrerelease=false&$skip=0&$top=26&semVerLevel=2.0.0 HTTP/1.1
Accept: application/atom+xml, application/xml
user-agent: NuGet Client V3/4.3.0 (Microsoft Windows NT 10.0.14393.0, VS Enterprise/15.0)
X-NuGet-Client-Version: 4.3.0
Accept-Language: en-US
Host: dev.nugettest.org
Accept-Encoding: gzip, deflate
```

This yields no results. I think perhaps its a bug with the introduction of `IsLatestSemVer2` and `IsLatestStableSemVer2`.

This results in an empty "browse" tab when using SemVer 2.0.0:
image

xavierdecoster:
Fixes #3905

### DIFF
--- a/src/NuGetGallery/WebApi/QueryResult.cs
+++ b/src/NuGetGallery/WebApi/QueryResult.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.Data.OData;
-using Microsoft.Data.OData.Query;
 using System;
 using System.Globalization;
 using System.Linq;
@@ -15,19 +13,12 @@ using System.Web.Http;
 using System.Web.Http.OData;
 using System.Web.Http.OData.Query;
 using System.Web.Http.Results;
+using Microsoft.Data.OData;
+using Microsoft.Data.OData.Query;
+using NuGetGallery.OData.QueryFilter;
 
 namespace NuGetGallery.WebApi
 {
-    public static class QueryResultDefaults
-    {
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2211:NonConstantFieldsShouldNotBeVisible")]
-        public static ODataQuerySettings DefaultQuerySettings = new ODataQuerySettings()
-        {
-            HandleNullPropagation = HandleNullPropagationOption.False,
-            EnsureStableOrdering = true,
-            EnableConstantParameterization = false
-        };
-    }
 
     public class QueryResult<TModel>
         : IHttpActionResult
@@ -127,8 +118,8 @@ namespace NuGetGallery.WebApi
                 if (queryOptions.Filter != null)
                 {
                     if (_semVerLevelKey != SemVerLevelKey.Unknown
-                        && (string.Equals(queryOptions.Filter.RawValue, "IsLatestVersion", StringComparison.OrdinalIgnoreCase)
-                            || string.Equals(queryOptions.Filter.RawValue, "IsAbsoluteLatestVersion", StringComparison.OrdinalIgnoreCase)))
+                        && (string.Equals(queryOptions.Filter.RawValue, ODataQueryFilter.IsLatestVersion, StringComparison.OrdinalIgnoreCase)
+                            || string.Equals(queryOptions.Filter.RawValue, ODataQueryFilter.IsAbsoluteLatestVersion, StringComparison.OrdinalIgnoreCase)))
                     {
                         // The client uses IsLatestVersion and IsAbsoluteLatestVersion by default,
                         // and just appends semVerLevel=2.0.0 to the query string.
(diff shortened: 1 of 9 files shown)


src/NuGetGallery/WebApi/QueryResult.cs
```diff
@@ -1,19 +1,20 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.Data.OData;
+using Microsoft.Data.OData.Query;
```
chenriksson: nit: move under System namespaces

src/NuGetGallery/WebApi/QueryResult.cs
```diff
@@ -121,7 +126,22 @@ public IHttpActionResult GetInnerResult()
 
                 if (queryOptions.Filter != null)
                 {
-                    queryResults = queryOptions.Filter.ApplyTo(queryResults, _querySettings);
+                    if (_semVerLevelKey != SemVerLevelKey.Unknown
+                        && (string.Equals(queryOptions.Filter.RawValue, "IsLatestVersion", StringComparison.OrdinalIgnoreCase)
+                            || string.Equals(queryOptions.Filter.RawValue, "IsAbsoluteLatestVersion", StringComparison.OrdinalIgnoreCase)))
```
chenriksson: Add constants for these? Looks like these strings are used in SearchAdapter as well.

tests/NuGetGallery.Facts/Controllers/ODataV2CuratedFeedControllerFacts.cs
```diff
@@ -115,52 +115,132 @@ public class ODataV2CuratedFeedControllerFacts
...
+            // Assert
+            Assert.Equal(NonSemVer2Packages.Where(p => !p.IsPrerelease).Count(), searchCount);
+        }
+
+        [Theory]
+        [InlineData("2.0.0")]
+        [InlineData("2.0.1")]
+        [InlineData("3.0.0-alpha")]
+        [InlineData("3.0.0")]
+        public async Task SearchCount_IncludesSemVerV2PackageVersionsWhenSemVerLevel2OrHigher_IncludePrerelease(string semVerLevel)
```
chenriksson: Is there a test for semverlevel < 2 to ensure it resolves as unknown?

src/NuGetGallery/WebApi/QueryResult.cs
```diff
@@ -59,6 +61,9 @@ public QueryResult(ODataQueryOptions<TModel> queryOptions, IQueryable<TModel> qu
             _totalResults = totalResults;
             _generateNextLink = generateNextLink;
 
+            var queryDictionary = HttpUtility.ParseQueryString(queryOptions.Request.RequestUri.Query);
+            _semVerLevelKey = SemVerLevelKey.ForSemVerLevel(queryDictionary["semVerLevel"]);
```
ryuyu: Is this ok if semVerLevel isn't passed? Or are we guaranteed to always pass semVerLevel? Should we do a TryGet instead?

tests/NuGetGallery.Facts/Controllers/ODataV2CuratedFeedControllerFacts.cs
xavierdecoster: Yes, there's `SearchCount_FiltersSemVerV2PackageVersions_IncludePrerelease` and `SearchCount_FiltersSemVerV2PackageVersions_ExcludePrerelease`

src/NuGetGallery/WebApi/QueryResult.cs
xavierdecoster: This is OK with `semVerLevel=null` being passed :) It uses a [`NameValueCollection`](https://msdn.microsoft.com/en-us/library/8d0bzeeb.aspx) under the hood, for which the index property returns `null` if the key is not found (which in turn is handled in the `SemVerLevelKey.ForSemVerLevel`

src/NuGetGallery/WebApi/QueryResult.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/3966/commits/48846ad34658ae0af0ddea292849cb033305209d

src/NuGetGallery/WebApi/QueryResult.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/3966/commits/48846ad34658ae0af0ddea292849cb033305209d

PR #3968, opened 2017-05-15
Fix: "Example glob patterns" button and "close" button are not accessible using keyboard

skofman1:


src/NuGetGallery/Views/Users/ApiKeys.cshtml
```diff
@@ -599,11 +599,13 @@
         $('.boxclose')
                .click(function () {
                    $(this).parent().hide();
+                   return false;
                });
 
         $('.globHelpLink')
             .click(function () {
                 $(this).next().find('.popupbox').show();
+                return false;
```
chenriksson: False prevents propagation? Maybe comment is useful?

skofman1: done

PR #3983, opened 2017-05-17
Fix focus outlining for real

ryuyu:
Allow auto outlining. Fix issue with outline on click.
This gets rid of the orange outline in favour of letting the browser decide, and fixes the issue with seeing the outline when clicking an item.

@joelverhagen @skofman1

### DIFF
--- a/src/NuGetGallery/Content/Site.css
+++ b/src/NuGetGallery/Content/Site.css
@@ -156,6 +156,10 @@ a {
         border: none;
     }
 
+    a:focus {
+        outline: auto;
+    }
+
 blockquote {
     color: #666;
     font-style: italic;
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Content/Site.css
```diff
@@ -987,6 +987,12 @@ fieldset.form {
         border-left: none;
     }
 
+        .form-field input[type="checkbox"]:focus,
```
joelverhagen: I don't see `outline: auto !important;` for `<textarea>`. Is that okay?

skofman1:
🚢

src/NuGetGallery/Content/Site.css
ryuyu: I didn't force outlines for inputs that had other methods of displaying that they were selected.
I believe text fields are considered accessible even without an explicit outline as the blinking cursor indicates focus.

joelverhagen: Sounds good!

Issue #3795, opened 2017-04-17
Force using HTTPS every time for the Gallery


agr:
The nuget.org web site uses HSTS, which presumes that only HTTPS should be used for visiting it, but it does not force HTTPS on the first visit. So if some browser was never used to access nuget.org, then first visit to http://nuget.org will not redirect to https://nuget.org. User would be able to download packages over plain HTTP which is not safe. Only authentication page currently forces HTTPS (and once you get there, you'd be unable to use HTTP again due to HSTS header, so no reason not to redirect to HTTPS for *any* HTTP request)

skofman1:
@agr, please split this into an issue for gallery and an issue for search.

skofman1:
Make sure this is configurable (external users shouldn't be forced to use https)

agr:
Copied from the [ServerCommon#45](https://github.com/NuGet/ServerCommon/pull/45)

Some statistics:
We are getting 2-2.5 million requests over HTTP per day (not HTTPS). Almost all of them are GET requests and will be properly redirected. Also, there are ~5K HEAD requests that also would be properly redirected.

We are getting about 10K POST requests per day, most of the are $batch OData requests from a single IP address that result in HTTP 301, because they are made to nuget.org (and get redirected to www.nuget.org), seems to be not properly processed by whoever sends them. We also get about 20-25 legit POST requests (that result in 2xx response). These would start to produce errors, so clients would have to switch to HTTPS.

We also get the DELETE, OPTIONS, PROPFIND, PUT and TRACE requests, usually less than 10 per day. Most of them either result in 404, or produce the same response as GET, so we can ignore them.

PR #3993, opened 2017-05-19
Forcing HTTPS in Gallery

agr:
#3795 

* Removed the `ForceSslWhenAuthenticatedMiddleware` and `RequireSslAttribute` that are not needed anymore. 
* Added the ForceSslMiddleware from `NuGet.Services.Owin`.
* Added functional tests


src/NuGetGallery/packages.config
```diff
@@ -84,6 +84,7 @@
   <package id="NuGet.Packaging.Core" version="4.3.0-preview1-2507" targetFramework="net46" />
   <package id="NuGet.Protocol" version="4.3.0-preview1-2507" targetFramework="net46" />
   <package id="NuGet.Services.KeyVault" version="1.0.0.0" targetFramework="net46" />
+  <package id="NuGet.Services.Owin" version="2.2.0-develop-6981" targetFramework="net46" />
```
joelverhagen: We should produce a stable version and depend on that instead of a prerelease build.

tests/NuGetGallery.FunctionalTests/Security/HttpToHttpsRedirectTests.cs
```diff
@@ -0,0 +1,118 @@
...
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace NuGetGallery.FunctionalTests.Security
+{
+    public class HttpToHttpsRedirectTests
+    {
+        public static IEnumerable<object[]> UrlsToTest
```
joelverhagen: 👍

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -444,6 +444,9 @@
       <HintPath>..\..\packages\NuGet.Services.KeyVault.1.0.0.0\lib\net45\NuGet.Services.KeyVault.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="NuGet.Services.Owin, Version=2.2.2.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\NuGet.Services.Owin.2.2.2-develop-7285\lib\net452\NuGet.Services.Owin.dll</HintPath>
```
agr: Should use stable release

src/NuGetGallery/Configuration/IAppConfiguration.cs
```diff
@@ -34,6 +34,12 @@ public interface IAppConfiguration
         int SSLPort { get; set; }
 
         /// <summary>
+        /// A string containing a regex for the path(s) exluded from
```
agr: the comment is wrong. Not a regex anymore

chenriksson:
Will Gallery functional test 'LoadTests.HitSearchEndPointDirectly' need updating? Or will functional hit HTTPS directly?

This is the test that failed during INT deployment when it accidentally was pointing at DEV which had the HTTPS redirect.

agr:
Dev had an issue with HTTPS configuration: the host names used for the services were not matching the HTTPS certificates installed on the machines. Not sure if those failures happened before or after it was fixed. But anyway it was deployed to dev and verified that functional and e2e tests pass with those changes.

src/NuGetGallery/Configuration/AppConfiguration.cs
```diff
@@ -28,6 +28,9 @@ public class AppConfiguration : IAppConfiguration
         [DefaultValue(443)]
         public int SSLPort { get; set; }
 
+        [DefaultValue("")]
+        public string ForceSslExclusion { get; set; }
```
skofman1: add a comment

src/NuGetGallery/Configuration/IAppConfiguration.cs
```diff
@@ -34,6 +34,12 @@ public interface IAppConfiguration
         int SSLPort { get; set; }
 
         /// <summary>
+        /// A string containing a path exluded from
+        /// forcing the HTTP to HTTPS redirection.
```
skofman1: Give an example of the expected format

agr:
@joelverhagen , how about `/api/health-probe`? Or `load-balancer-probe`?

joelverhagen:
Great ideas! I think `/api/health-probe` is the best.

src/NuGetGallery/Configuration/AppConfiguration.cs
agr: What's the point of documenting implementation? It would be a copy-paste of the documentation from the interface, so we'd need to keep it updated, and since we are not supposed to use the concrete class directly anywhere besides DI setup, chances of that documentation to be read are minuscule.

PR #3994, opened 2017-05-19
Add log in page to redesign

joelverhagen:
1. Add log in page to redesign. Old sign up page is now at `/users/account/SignUp`.
1. Fixed footer height on phones reported by @skofman1 
1. Fixed horizontal scroll reported by @xavierdecoster 

https://preview-dev.nugettest.org/users/account/LogOn


skofman1:
I logged in successfully. After log-in got redirected to home page, however there was no indication that I'm logged in (home page still show sign up/login options).
So I tried to log-in again, and got to home page with message "already logged-in". However the color of the message blended with the background, and it wasn't noticeable:
image

skofman1:
IMO the input fields are too big.. checkout https://login.live.com for example.
Though this feedback is not specific to the implementation, but to the design itself. @jonwchu ?

joelverhagen:
> however there was no indication that I'm logged in (home page still show sign up/login options).

By design. Phase 1 does not include logged in stuff.

> So I tried to log-in again, and got to home page with message "already logged-in". However the color of the message blended with the background, and it wasn't noticeable:

This is pending transient message design from @jonwchu.

> IMO the input fields are too big.. 

Just too wide? Or too wide and too tall (and text is too big). In other words would making the text field simply more narrow be better?

joelverhagen:
Spoke offline, will add the "alread logged-in" stuff in a follow-up PR.

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -35,11 +35,17 @@
...
+                        @if (!User.Identity.IsAuthenticated)
+                        {
                             var returnUrl = ViewData.ContainsKey(Constants.ReturnUrlViewDataKey) ? (string)ViewData[Constants.ReturnUrlViewDataKey] : Request.RawUrl;
+                            <li class="@(ViewBag.Tab == "Log in" ? "active" : string.Empty)"><a href="@Url.LogOn(returnUrl)" title="Log in to an existing NuGet.org account"><span>Log in</span></a></li>
+                            <li class="@(ViewBag.Tab == "Sign up" ? "active" : string.Empty)"><a href="@Url.SignUp(returnUrl)" title="Sign up for a new NuGet.org account"><span>Sign up</span></a></li>
+                        }
+                        else
+                        {
+                            <li><a href="@Url.Action("Account", "Users", new { area = string.Empty })"><b>@(User.Identity.Name)</b></a></li>
+                            <li><a href="@Url.LogOff()">Log off</a></li>
```
skofman1: "Sign out".
That's what we have in NuGet.org and even gmail :)

joelverhagen: I'll send a mail to @jonwchu about this. I will follow the design for now. Either way it would be "Sign in/Sign out" OR "Log in/Log out".

PR #4031, opened 2017-05-31
Version validation should happen on the full string

### LINKED ISSUE #3916: On package validation failure an actionable error message should be displayed.
The validation package failure will result in 500 and the user does not have any context what the reason may be.

Use the below file and rename it to nupkg.
Try to upload the package on the dev environment.
[NBBLib2 - Copy.zip](https://github.com/NuGet/NuGetGallery/files/987748/NBBLib2.-.Copy.zip)

Result: 500

xavierdecoster:
Fixes #3916

The code was relying on `NuGetVersion.ToString` implementation instead of `NuGetVersion.ToFullString`.


xavierdecoster:
Screenshot of the error message:

image

PR #4032, opened 2017-05-31
Add dotnet core installation instructions

a-h:
dotnet Core doesn't have a Package Manager Console that I'm aware of, so I think it makes sense to show people how to add packages using the `dotnet add package` command line to reduce confusion.


a-h:
No longer required, since someone else has now added this.

screen shot 2017-09-04 at 13 44 53

Issue #4039, opened 2017-06-02
Average download shown incorrectly when its 1.x 


anangaur:
From a tweet: https://twitter.com/plecaillon/status/870322417251602432 
 
![image](https://cloud.githubusercontent.com/assets/14800916/26705813/734fe18e-46ed-11e7-9cee-716c226ca86b.png)

Its 63 days since first publish of the package: https://www.nuget.org/packages/Evolve 
![image](https://cloud.githubusercontent.com/assets/14800916/26705823/817fd7dc-46ed-11e7-91c5-bf6bca0d5499.png)

 It should be 79/63 = ~1 download per day. 

Fix https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml#L43

```
Model.DownloadsPerDay <= 1 ? "<1" : Model.DownloadsPerDay.ToNuGetNumberString() 
```
should be (??)
```
Model.DownloadsPerDay < 1 ? "<1" : Model.DownloadsPerDay.ToNuGetNumberString()
```

xavierdecoster:
Verified in PROD, this now shows `1` instead of `<1`.
https://www.nuget.org/packages/Evolve

PR #4040, opened 2017-06-02
Average download shown incorrectly when its 1.x #4039

xavierdecoster:
Fixes https://github.com/NuGet/NuGetGallery/issues/4039

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -40,7 +40,7 @@
             <p class="stat-label">@(Model.DownloadCount == 1 ? "Download" : "Downloads") of v @Model.Version</p>
         </div>
         <div class="stat">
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -40,7 +40,7 @@
             <p class="stat-label">@(Model.DownloadCount == 1 ? "Download" : "Downloads") of v @Model.Version</p>
         </div>
         <div class="stat">
```
skofman1: Since this line is easy to get wrong, we should have UTs. Can you please move the logic to the ViewModel?

xavierdecoster: Addressed with https://github.com/NuGet/NuGetGallery/pull/4040/commits/760b2ad64509c93fb1c56485241a680545d70c86

PR #4048, opened 2017-06-02
[Redesign]Statistics Overview Page

ryuyu:
Matching Live Page: https://nuget.org/stats

Changed styling of stats overview page to match redesign styles.
Added table of data in charts.
Rewrote charts to more closely match design.
Narrow screen compatible.

Verified tab ordering, focus highlighting.
Adds aria-labels to non-anchor table rows for screen reader accessibility.

Chart itself can be made accessible by keyboard if we want (tested on Chrome, Edge. Does NOT work in IE).
Let me know of any additional styling changes we want to make, or if something needs to change.
Note that chart highlights bar on mouse over. This behaviour can change if we want; I was just keeping parity with what the current chart does.

@microsoftsam @scottbommarito @loic-sharma @joelverhagen 

Screenshot:
image


src/NuGetGallery/Views/Statistics/Index.cshtml
```diff
@@ -2,155 +2,230 @@
...
-    <p>
-        <strong>NOTE: </strong>Download statistics displayed on this page reflect the actual package downloads from the NuGet.org site. This is an indicator of usage. However,
-        many package installs, particularly those of widely depended upon packages, will actually be done from a client machine's local NuGet cache, as a result these numbers should
-        not be taken as exact usage.
-    </p>
-</section>
+<div class="page-statistics-overview">
+    <div class="container">
+        <div class="row page-heading">
+            <div class="col-xs-12">Statistics</div>
```
joelverhagen: Headings should use `<h1>` or similar.

src/NuGetGallery/Views/Statistics/Index.cshtml
```diff
@@ -2,155 +2,230 @@
...
+        </div>
+        <div class="row page-subheading">
+            <div class="col-xs-12">Actual package downloads from the NuGet.org site</div>
```
joelverhagen: Sub headings should use `<h2>` or similar.

src/NuGetGallery/Views/Statistics/Index.cshtml
```diff
@@ -2,155 +2,230 @@
...
-            </thead>
-            <tbody>
-            @{
-                var index = 0;
-                foreach (var item in Model.DownloadPackageVersionsSummary)
-                {
-                    index++;
+                                    <div class="row table-row-data">
+                                        <div class="col-xs-8 text-left table-cell"><a href="@Url.Package(item.PackageId)" title="@item.PackageId">@item.PackageId</a></div>
+                                        <div class="col-xs-4 text-right table-cell"><a href="@Url.StatisticsPackageDownloadByVersion(item.PackageId)">@Model.DisplayDownloads(item.Downloads)</a></div>
```
joelverhagen: Why `<div>` for this tabular data?

joelverhagen: Ideally we would use a `<table>` here.

src/NuGetGallery/Views/Statistics/Index.cshtml
```diff
@@ -2,155 +2,230 @@
...
+                            }
+                            <div class="row">
+                                <div class="col-xs-12 table-row-data">
+                                    <a href="@Url.StatisticsAllPackageVersionDownloads()">More...</a>
+                                </div>
+                            </div>
+                        </div>
+                    </div>
+                </div>
+                                }
```
joelverhagen: nit: this curly seems like its indented weirdly

src/NuGetGallery/Views/Statistics/Index.cshtml
```diff
@@ -2,155 +2,230 @@
...
+        </div>
 
-<p></p>
+        <div class="row">
+            @if (Model.IsNuGetClientVersionAvailable)
+            {
+                <div class="col-md-6 col-xs-12">
+                    <div class="row stats-title-text">
+                        <div class="col-xs-12">NuGet&nbsp;Client&nbsp;Usage&nbsp;(Last&nbsp;6&nbsp;Weeks)</div>
```
joelverhagen: Why `&nbsp;`?

src/NuGetGallery/Views/Statistics/Index.cshtml
```diff
@@ -2,155 +2,230 @@
...
+                                }
+                            }
+                        </div>
+                    </div>
+                </div>
+                                }
+            @if (Model.IsLast6WeeksAvailable)
+            {
+                <div class="col-md-6 col-xs-12">
+                    <div class="row stats-title-text">
```
joelverhagen: Headings should use `<h2>` or similar

src/Bootstrap/less/theme/page-statistics-overview.less
```diff
@@ -0,0 +1,79 @@
+.page-statistics-overview {
+    .page-heading {
+        font-size: 2.5em;
+        margin-top: 30px;
+    }
+
+    .page-subheading {
```
joelverhagen: We should avoid styling headings in a page-specific way. `<h1>` tags are the better.

src/Bootstrap/less/theme/page-statistics-overview.less
```diff
@@ -0,0 +1,79 @@
...
+        font-size: 1em;
```
joelverhagen: What's the point of `1em`?

ryuyu: Excellent question. Was just so i could keep track of relative sizes between things in my head. Cleaned up.

src/NuGetGallery/Views/Statistics/Index.cshtml
```diff
@@ -2,155 +2,230 @@
...
-    <p>
-        <strong>NOTE: </strong>Download statistics displayed on this page reflect the actual package downloads from the NuGet.org site. This is an indicator of usage. However,
-        many package installs, particularly those of widely depended upon packages, will actually be done from a client machine's local NuGet cache, as a result these numbers should
-        not be taken as exact usage.
-    </p>
-</section>
+<div class="page-statistics-overview">
+    <div class="container">
+        <h1 class="page-heading">Statistics</h1>
+        <span class="page-subheading">Actual package downloads from the NuGet.org site</span>
```
joelverhagen: I think this pattern is common across a lot of pages. Can we move this to a common place? Both `page-heading` and `page-subheading`.

microsoftsam:
The padding at the bottom of the page isn't consistent here. (non existent in this page). Let's consider fixing that padding for all pages. ie: settings it on the container. That way it reflects on all pages.

src/NuGetGallery/Views/Statistics/Index.cshtml
```diff
@@ -2,155 +2,230 @@
...
+                                        <td class="text-left table-cell"><a href="@Url.Package(item.PackageId)" title="@item.PackageId">@item.PackageId</a></td>
+                                        <td class="text-right table-cell"><a href="@Url.StatisticsPackageDownloadByVersion(item.PackageId)">@Model.DisplayDownloads(item.Downloads)</a></td>
+                                    </tr>
+                                }
+                            }
+                        </tbody>
+                        <tfoot>
+                            <tr class="row table-row-data">
+                                <td class="table-cell">
+                                    <a href="@Url.StatisticsAllPackageVersionDownloads()">More...</a>
```
microsoftsam: Let's add more info here. Just having the screen reader read more is too far out of context from what More is leading to. This could be "View more package download statistics". 
We can make this only show for screen readers by using class="sr-only". 
so in this case, you could have a ```<span class="sr-only">View more package download statistics</span>``` inside the link and this would only be read by a screen reader.

Referring to Accessibility WCAG Link Purpose (in context). 
For more info:
https://www.youtube.com/watch?v=73EdfgAPw8E&feature=youtu.be&t=8m55s

ryuyu: Added a span with this to both "More..." links.

src/NuGetGallery/Views/Statistics/Index.cshtml
ryuyu: Moved to base.less.

src/NuGetGallery/ViewModels/StatisticsPackagesViewModel.cs
```diff
@@ -76,19 +84,72 @@ public void Update()
...
 
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1822:MarkMembersAsStatic", Justification = "We want to be able to use this easily in the related view.")]
         public string DisplayPercentage(float amount, float total)
         {
             return (amount / total).ToString("P0", CultureInfo.CurrentCulture);
         }
+
+        public string DisplayShortNumber(int number)
+        {
+            var abbreviation = new string[] { "", "k", "M", "B", "q", "Q", "s", "S", "o", "n" };
```
loic-sharma: What are `q`, `Q`, `s`, `S`, `o`, and `n`? Shouldn't `T` for trillion be in there? Also, should the `abbreviation` array be cached to avoid an allocation every method call?

ryuyu: Yes i missed trillion. Yes we should probably cache it. Good call.

PR #4051, opened 2017-06-03
[Redesign]Update global font

ryuyu:
Updates font to match the font used in the specs: Segoe UI > Segoe UI light

@scottbommarito @joelverhagen @jonwchu @loic-sharma @microsoftsam


ryuyu:
After speaking to @loic-sharma, we may want to hold off on this as it appears the specs we have use Segoe UI in some places, and Segoe UI light in other places.

microsoftsam:
Consider using ms-font classes that are in fabric.css to have greater flexibility in what font you use where.

PR #4059, opened 2017-06-06
Update telemetry processors

### LINKED ISSUE #4041: NuGetGallery:  reduce exception telemetry noise
NuGetGallery Application Insights telemetry contains lots of noise in the "exceptions" table.  This noise is characterized by the following properties:

1.  The associated request had an HTTP response code less than 500.
2.  The root cause behind the error is generally out of our control and is either infeasible or undesirable to "fix".

This exception telemetry noise should be converted into trace warning telemetry.  "Real", actionable exceptions should continue to be logged to the "exceptions" table.  "Real" exceptions are either of the following:

1.  exceptions of type `System.Web.HttpException` with an HTTP response code greater than or equal to 500
2.  exceptions of a type other than `System.Web.HttpException`

dtivel:
Resolve #4041.

### DIFF
--- a/src/NuGetGallery/Telemetry/Telemetry.cs
+++ b/src/NuGetGallery/Telemetry/Telemetry.cs
@@ -9,13 +9,18 @@ namespace NuGetGallery
 {
     internal static class Telemetry
     {
-        private static readonly TelemetryClient _telemetryClient = new TelemetryClient();
+        static Telemetry()
+        {
+            TelemetryClient = new TelemetryClient();
+        }
+
+        internal static TelemetryClient TelemetryClient { get; }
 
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Telemetry/Telemetry.cs
```diff
@@ -9,13 +9,18 @@ namespace NuGetGallery
 {
     internal static class Telemetry
     {
-        private static readonly TelemetryClient _telemetryClient = new TelemetryClient();
+        static Telemetry()
+        {
+            TelemetryClient = new TelemetryClient();
+        }
+
+        internal static TelemetryClient TelemetryClient { get; }
```
loic-sharma: Nit: you could initialize the `TelemetryClient` here and remove the static constructor.

PR #4061, opened 2017-06-06
Support for policy propagation

chenriksson:
- Added propagation policy which can be manually subscribed from admin security policies view
- Added 'add owner' confirmation from ManageUsers view, with policy warning if propagation detected
- Modified add owner request email to include policy warning if propagation detected
- Added email to existing owners once new owner is added, with policy warning if propagation detected
- Removed unused queries from ManagePackageOwnersViewModel

Will email various scenarios w/ expected communication to the team shortly for review... since they differ slightly from @anangaur 's experience spec.


src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -59,71 +61,153 @@ public virtual ActionResult GetPackageOwners(string id, string version)
...
+        private Task<JsonResult> ManagePackageOwnerAsync(string id, string username, Func<PackageRegistration, User, User, JsonResult> action)
+        {
+            return ManagePackageOwnerAsync(id, username, (package, user, currentUser) => Task.FromResult(action(package, user, currentUser)));
+        }
+
+        private Task<JsonResult> ManagePackageOwnerAsync(string id, string username, Func<PackageRegistration, User, User, Task<JsonResult>> actionAsync)
         {
+            if (string.IsNullOrEmpty(id))
+            {
+                throw new ArgumentNullException(nameof(id));
```
skofman1: >ArgumentNullException [](start = 26, length = 21)

ArgumentException since if id is empty ArgumentNullException will be confusing

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -59,71 +61,153 @@ public virtual ActionResult GetPackageOwners(string id, string version)
...
+            if (string.IsNullOrEmpty(username))
+            {
+                throw new ArgumentNullException(nameof(username));
+            }
+
             var package = _packageService.FindPackageRegistrationById(id);
             if (package == null)
             {
-                return Json(new { success = false, message = "Package not found" });
+                return Task.FromResult(Json(new { success = false, message = "Package not found." }));
```
skofman1: messages should be in Strings.resx

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -206,10 +206,13 @@ protected override void Load(ContainerBuilder builder)
                 .AsSelf()
                 .As<ISecurityPolicyService>()
                 .InstancePerLifetimeScope();
-
+
```
loic-sharma: Empty white space here.

src/NuGetGallery/Views/Packages/ManagePackageOwners.cshtml
```diff
@@ -80,21 +82,47 @@
...
+                        url: '@Url.Action("GetAddPackageOwnerConfirmation", "JsonApi")?id=' + viewModel.package.id + '&username=' + newUsername,
+                        cache: false,
+                        dataType: 'json',
+                        type: 'GET',
+                        headers: this.headers(),
+                        contentType: 'application/json; charset=utf-8',
+                        success: function (data) {
+                            if (data.success) {
+                                if (data.confirmation) {
+                                    if (!confirm(data.confirmation)) {
```
loic-sharma: Merge this with the previous conditional?

src/NuGetGallery/Security/RequirePackageVerifyScopePolicy.cs
```diff
@@ -11,7 +10,7 @@ namespace NuGetGallery.Security
     /// </summary>
     public class RequirePackageVerifyScopePolicy : UserSecurityPolicyHandler
     {
-        public const string PolicyName = "RequirePackageVerifyScopePolicy";
+        public const string PolicyName = nameof(RequirePackageVerifyScopePolicy);
```
skofman1: 👍

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1022,8 +1026,50 @@ private ActionResult EditFailed(string id, EditPackageRequest formData, Package
...
+                var ownersWithPolicy = package.Owners.Where(o => RequireSecurePushForCoOwnersPolicy.IsSubscribed(o));
+                if (ownersWithPolicy.Any())
+                {
+                    await _securityPolicyService.SubscribeAsync(user, SecurePushSubscription.Name);
+                    policyMessageOwners = ownersWithPolicy.Select(o => o.Username).ToList();
+                    policyMessage = SecurePushMessages.NoticeOfPoliciesSubscribedByPolicyOwner(policyMessageOwners, user);
+                }
+            }
+
             ConfirmOwnershipResult result = await _packageService.ConfirmPackageOwnerAsync(package, user, token);
```
skofman1: token verification might fail. This will happened only in ConfirmPackageOwnerAsync method. With the current code this will be after subscription propagation was applied.

chenriksson: Thanks - good catch!

src/NuGetGallery/Controllers/JsonApiController.cs
chenriksson: done, except for existing and new emails in MessageService

src/NuGetGallery/Security/SecurePushMessages.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <summary>
+        /// Scenario: Confirm addition of 'Microsoft' (propagating) as owner of current user's package.
+        /// </summary>
+        public static string ConfirmationOfPoliciesRequiredByNewPendingCoOwner(User newOwner)
+        {
+            return $@"User '{newOwner.Username}' has the following requirements that will be enforced for all co-owners once the user accepts ownership of this package:
+
+{PolicyDescriptions}
+
+Note this step cannot be easily undone. If you are unsure and/or need more information, please contact support@nuget.org.
```
chenriksson: Need to update to use 'Gallery.GalleryOwner' for support@nuget.org.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1033,6 +1043,74 @@ private ActionResult EditFailed(string id, EditPackageRequest formData, Package
...
+            // subscribe existing owners if new owner has co-owners policy requirement.
+            // note: if another pending owner has the co-owners policy requirement, this user
+            // will be subscribed when the user with the requirement has confirmed their ownership.
+            var policyMessage = string.Empty;
+            var policyMessageOwners = new List<string>();
+            if (RequireSecurePushForCoOwnersPolicy.IsSubscribed(user))
+            {
+                foreach (var owner in package.Owners)
+                {
+                    if (await SubscribeToSecurePushAsync(owner))
```
loic-sharma: Is this likely to fail? If so, how would we find and then add missing secure push policies? Would it be better to move `SubscriptToSecurePushAsync` to the `PackageService`'s `AddPackageOwnerAsync`, and if `SubscriptToSecurePushAsync` fails for any of the owners, then the overall `AddPackageOwnerAsync` operation fails.

skofman1: I agree with Loic here. Consider spliting ConfirmPackageOwnerAsync to two methods: one to evaluate the token and return true/false, and another to actually apply the ownership. HandleNewPackageOwnerAsync can run after verifications of ownership but before actually applying.

---
In reply to: [120957931](https://github.com/NuGet/NuGetGallery/pull/4061#discussion_r120957931) [](ancestors = 120957931)

chenriksson: Can do; was trying to avoid modifying the existing PackageService.ConfirmPackageOwnerAsync which is committed in a separate transaction.

skofman1: we can still keep it in a separate transaction, it's just better to fail adding the owner (and the user can retry that), then to fail adding policies.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -498,20 +504,114 @@ public class TheConfirmOwnerMethod
...
                 var result = await controller.ConfirmOwner("foo", "username", "token");
 
+                // Assert
                 var model = ResultAssert.IsView<PackageOwnerConfirmationModel>(result);
                 Assert.Equal(confirmationResult, model.Result);
                 Assert.Equal("foo", model.PackageId);
             }
+
+            [Fact]
+            public async Task SubscribesOwnersToSecurePushAndSendsEmailIfNewOwnerRequires()
```
scottbommarito: I would suggest putting these inside a `TestContainer` so that these tests are separated from the rest of the `TheConfirmOwnerMethod` tests, because they are specifically related to `RequireSecurePushForCoOwnersPolicy`.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -498,20 +504,114 @@ public class TheConfirmOwnerMethod
...
+                Assert.False(policyMessages.ContainsKey(fakes.User.Username));
+                Assert.Equal(2, policyMessages.Count);
+                Assert.StartsWith("User 'testUser' has the following requirements that are now enforced for your account:",
+                    policyMessages[fakes.Owner.Username]);
+                Assert.StartsWith("User 'testUser' has the following requirements that are now enforced for your account:",
+                    policyMessages[fakes.ShaUser.Username]);
+            }
+
+            [Fact]
+            public async Task SubscribesNewOwnerToSecurePushAndSendsEmailIfOwnerRequires()
```
scottbommarito: We need failure state tests:

- subscribing new owner fails
    - new owner should not be added
    - old owners shouldn't be subscribed if they weren't before
- subscribing old owners fails
    - new owner should not be added
    - new owner and other old owners shouldn't be subscribed if they weren't before

chenriksson:
Still looking into the following:
- #4079 (Admin/SecurityPolicies header checkbox)
- #4081 (Replacement for modal dialog?)
- Retain user search results on /admin policy subscription
- Display secure push info on packages view / manage package owners, for owners/admins

chenriksson:
Screenshot for DisplayPackage.cshtml (if owner or admin only):
displaypackagepolicymessage

Screenshot for ManagePackageOwners.cshtml confirmation (no policies):
manageownersaddconfirmationmessage

Screenshot for ManagePackageOwners.cshtml confirmation (with policies):
manageownersaddconfirmationmessagewithpolicies

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: @loic-sharma @skofman1 @scottbommarito Code was updated to validate the owner request token before applying policy subscription(s). I didn't add logic to unsubscribe should SQL to add owner fail. This is currently hard since there are multiple points of failure: add owner transaction, remove owner request transaction, auditing. Failing more gracefully is probably something we can generally do better, but I think the current behavior here is sufficient for now.

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -69,40 +69,43 @@ public virtual JsonResult Search(string query)
...
                 );
 
-            // Iterate all users and groups to handle both subscribe and unsubscribe.
-            var usernames = GetUsernamesFromQuery(viewModel.UsersQuery);
-            var users = FindUsers(usernames);
-            foreach (var user in users)
+            foreach (var r in subscribeRequests)
             {
-                foreach (var subscription in PolicyService.UserSubscriptions)
+                var user = FindUser(r.Key);
```
agr: Check `r.Key` for null before calling the `FindUser`?
(and another instance few lines below)

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -118,6 +122,25 @@ private AuditedSecurityPolicyAction GetAuditAction(SecurityPolicyAction policyAc
         /// <summary>
         /// Check if a user is subscribed to one or more security policies.
         /// </summary>
+        public bool IsSubscribed(User user, string subscriptionName)
+        {
+            if (string.IsNullOrEmpty(subscriptionName))
+            {
+                throw new ArgumentException(nameof(subscriptionName));
```
agr: `ArgumentNullException`?

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
loic-sharma: Also, `FindUser` will return `null` if the user doesn't exist. Should this case be handled?

src/NuGetGallery/Security/SecurityPolicyService.cs
chenriksson: Per @skofman1 I should throw ArgException for IsNullOrEmpty. I'll update the other APIs that take subscriptionName to throw ArgException for consistency.

src/NuGetGallery/Areas/Admin/Controllers/SecurityPolicyController.cs
```diff
@@ -112,6 +115,12 @@ private static string[] GetUsernamesFromQuery(string query)
                 .Where(username => !string.IsNullOrEmpty(username)).ToArray();
         }
 
+        private User FindUser(string username)
+        {
+            return EntitiesContext.Users
+                .FirstOrDefault(u => username.Equals(u.Username, StringComparison.OrdinalIgnoreCase));
```
chenriksson: It's worked so far, but likely haven't tested against case sensitive collation. I verified that AuthenticationService is using '==', so I'll update to match.

src/NuGetGallery/Security/SecurityPolicyService.cs
agr: Maybe check using the `IsNullOrWhitespace` then?

chenriksson: Left as IsNullOrEmpty since it's used most consistently in the Gallery. Maybe we should revisit throughout at some point.

Issue #498, opened 2012-05-09
Cache-Control header should not be Private when serving packages


matthewskelton:
Currently, NuGetGallery leaves the Cache-Control header as Private for the "GetPackageApi" action. This means that GET requests to pull packages cannot be cached by downstream web caches (such as Squid/Varnish).

In order to allow web caches to cache GET requests for reading packages, we need a change in ApiController.cs - explicitly set the cacheability like this:

```
        this.Response.Cache.SetCacheability(HttpCacheability.Public);
        this.Response.Cache.SetExpires(DateTime.Now.AddDays(days));
```

Being able to cache GET read requests means that we can offload the load on the NuGet Gallery server. This is particularly important when running NuGet Gallery locally, with 80+ developers and upwards of 60 build pipelines.

I am happy to submit a pull request. We have this configuration working in our environment, and it works a treat with Squid - blazing fast performance from NuGet Gallery!

PR #4073, opened 2017-06-08
SemVer2 - Missing db index on Packages table #498

xavierdecoster:
Noticed an increase in DTU consumption on the gallery db after deploying semver2, and it's caused by a missing index on the Packages table.

The Query Processor estimates that implementing the following index could improve the query cost by 99.5235%.

Affected EF query:

```sql
(@p__linq__0 int, @p__linq__4 int,@p__linq__5 int)

SELECT	[GroupBy1].[A1] AS [C1]
FROM 
( 
	SELECT	COUNT(1) AS [A1]
    FROM	[dbo].[Packages] AS [Extent1]
    WHERE	([Extent1].[Deleted] <> 1) 
		AND (([Extent1].[SemVerLevelKey] = @p__linq__0) 
			OR (([Extent1].[SemVerLevelKey] IS NULL) AND (@p__linq__0 IS NULL))
			) 
		AND ((
			CASE 
			WHEN (@p__linq__4 = @p__linq__5) 
			THEN [Extent1].[IsLatestSemVer2] 
			ELSE [Extent1].[IsLatest] 
			END) = 1)
)  AS [GroupBy1]
```


joelverhagen:
What are the parameters in this case? What code path does this primarily affect?

xavierdecoster:
@joelverhagen https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Controllers/ODataV2FeedController.cs#L86

src/NuGetGallery/Migrations/201706080930506_AddIndexSemVerLevelKeyPackageRegistrationKey.cs
```diff
@@ -0,0 +1,18 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddIndexSemVerLevelKeyPackageRegistrationKey : DbMigration
+    {
+        public override void Up()
+        {
+            Sql("IF NOT EXISTS(SELECT * FROM sys.indexes WHERE name = 'nci_wi_Packages_SemVerLevelKey_PackageRegistrationKey' AND object_id = OBJECT_ID('Packages')) CREATE NONCLUSTERED INDEX [nci_wi_Packages_SemVerLevelKey_PackageRegistrationKey] ON [dbo].[Packages] ([SemVerLevelKey],[IsPrerelease],[Deleted]) INCLUDE ([PackageRegistrationKey],[Description],[Tags])");
```
scottbommarito: Any reason for these not to be in a single migration?

xavierdecoster: Yes, they were separate PRs, and have been evaluated separately to find out whether the change was for the better or not.

xavierdecoster:
Want to point out these indices have already been applied (see offline discussion), and monitoring has clearly shown they had a positive effect on gallery db DTU consumption.

Issue #499, opened 2012-05-10
Gallery assumes it is running on port 80 and so cannot be run behind a proxy


matthewskelton:
If you try to run NuGet Gallery on a port other than 80 (e.g. behind a caching web proxy or content switch), package installation from the NuGet client fails because in the response from NuGet Gallery, the ReadStreamUri contains the port number of the NuGet Gallery running instance.

This means that the NuGet client cannot download the package (due to firewall rules).

I think this is due to the implementation in https://github.com/NuGet/NuGetGallery/blob/master/Website/DataServices/V2Feed.svc.cs:

```
       var httpContext = new HttpContextWrapper(HttpContext.Current);
        var urlHelper = new UrlHelper(new RequestContext(httpContext, new RouteData()));

        string url = urlHelper.PackageDownload(FeedVersion, package.Id, package.Version);

        return new Uri(url, UriKind.Absolute);
```

This seems to assume that the Uri context at the NuGet Gallery is valid to return to the user, but this is not the case if running behind a proxy/content switch.

I'll submit a pull request once I have worked out the fix - we might need a config setting telling Gallery what the "public" Uri endpoint should be, and have Gallery read that, rather than having Gallery try to determine it itself.

PR #4074, opened 2017-06-08
SemVer2 - Missing db index on Packages table for partial search #499

xavierdecoster:
Targeting the other EF migration PR to avoid collisions in migration order.


chenriksson:
This will merge to dev-eng-498, then both to master? Looks like #4073 was applied to PROD, but #4074 still needs to be done.

xavierdecoster:
@chenriksson exactly. They're both applied to PROD, but on PROD I ran the index creation query through SSMS (because of EF migrations time-out). The migration checks if the index exists, so whenever we apply it, it's basically going to be a no-op :)

PR #4093, opened 2017-06-09
v2 package-versions auto-complete endpoint should exclude deleted versions

### LINKED ISSUE #4092: v2 package-versions auto-complete endpoint should exclude deleted versions
endpoint: `https://www.nuget.org/api/v2/package-versions/<id>`

xavierdecoster:
Fixes #4092


xavierdecoster:
@joelverhagen this is a production issue, so I'd suggest we get the fix in asap, and log another issue to track the need for testing/improving this particular piece of code.

xavierdecoster:
Tracking issue for test-coverage/improvement https://github.com/NuGet/NuGetGallery/issues/4094

Issue #4009, opened 2017-05-26
Page: Downloads


skofman1:
https://dist.nuget.org/index.html

PR #4098, opened 2017-06-09
[Redesign] Downloads page

joelverhagen:
/cc @microsoftsam

Address https://github.com/nuget/nugetgallery/issues/4009.

screencapture-localhost-downloads-1497025416400

### DIFF
--- a/src/NuGetGallery/Views/Pages/Downloads.cshtml
+++ b/src/NuGetGallery/Views/Pages/Downloads.cshtml
@@ -14,13 +14,13 @@
         <div class="col-md-4">
             <h2 class="ms-fontSize-xxl">Windows x86 Commandline</h2>
             <ul class="list-unstyled list-tools">
-                <li><a title="nuget.exe - recommended latest" href="https://dist.nuget.org/win-x86-commandline/latest/nuget.exe">nuget.exe - recommended latest</a></li>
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/Views/Pages/Downloads.cshtml
```diff
@@ -0,0 +1,105 @@
...
+            </a>
+        </li>
+    </ul>
+    <!-- /ko -->
+</script>
+
+@section BottomScripts {
+    <script type="text/javascript">
+        $(function () {
+            $.ajax({
```
scottbommarito: Should we consider making this call on the server side and then caching it, like we were considering for the About page?

joelverhagen: Caching is not necessary if its client side. We're hitting a CDN so downloading this small JSON file is worth it to always have up-to-date results.

Server side ended up being a lot of trouble since the code was _almost_ like the existing static content stuff but just far enough that the engineering effort would be high. We can revisit this later if we think making it server side provides much benefit.

src/NuGetGallery/Views/Pages/Downloads.cshtml
```diff
@@ -0,0 +1,105 @@
+﻿@{
+    ViewBag.Title = "Contact";
```
loic-sharma: This should be `Downloads`, right?

joelverhagen: Yup!

PR #4105, opened 2017-06-09
Update banner, transient message, and package details page

joelverhagen:
1. Banner should have the alert icon
1. Transient message should not fill the width of the screen
1. Update correct icons for display package
1. Move dependencies below versions
1. Hide dependencies by default (with expander)

screencapture-localhost-packages-newtonsoft-json-9-0-1-1497052627211

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/common.js
+++ b/src/NuGetGallery/Scripts/gallery/common.js
@@ -25,7 +25,7 @@
 
     // source: http://stackoverflow.com/questions/400212/how-do-i-copy-to-the-clipboard-in-javascript
     // enhancement with special case for IEs, otherwise the temp textarea will be visible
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -93,7 +93,27 @@
 
             document.body.removeChild(textArea);
         }
-    }
+    };
+
+    nuget.configureExpander = function (prefix, lessMessage, moreMessage) {
```
loic-sharma: Should this update the `aria-expanded` attribute?

joelverhagen: Bootstrap JS does this for us.

PR #4108, opened 2017-06-12
[Redesign] Error pages

scottbommarito:
Including [400](https://localhost/errors/400), [404](https://localhost/errors/404), [500](https://localhost/errors/500), and [Read-only Mode (503)](https://localhost/errors/readonly).

Things to note:
- We didn't have 400 before ([see here](https://www.nuget.org/packages?q=char%28)), just a page that was shown on anti-forgery token errors. This change expands that page to be used for every 400.
- I added endpoints that you can hit to show every page for those that did not have it. Previously, only 404 and 500 did.

400:
image

404:
image

500:
image

Read-only (503):
image

@microsoftsam


src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -31,6 +32,36 @@
     @Alert(htmlContent, "danger", "ErrorBadge", isAlertRole)
 }
 
+@helper ErrorPage(UrlHelper url, System.Web.Mvc.HtmlHelper html, string errorNumber, string errorName, Func<MvcHtmlString, HelperResult> errorTextMain, Func<MvcHtmlString, HelperResult> errorTextSub = null)
```
loic-sharma: Would it be better if `ErrorPage` was a shared view that other pages include with `Html.Partial(...)`? For an example, see the `_ListPackage` view.

microsoftsam:
Looks good.

src/NuGetGallery/App_Code/ViewHelpers.cshtml
scottbommarito: I have to pass HTML to the helper, which doesn't seem to be easily supported by `Html.Partial`. Since it's pretty simple as-is, I think it's better to keep it like this.

PR #4112, opened 2017-06-12
[Redesign]Package Stats/Package Version Stats Page(s)

ryuyu:
Note that this changeset requires https://github.com/NuGet/NuGetGallery/pull/4111 to function correctly as it uses a shared library.

This includes the changes for the pages:
https://www.nuget.org/stats/packages/Newtonsoft.Json.Schema 
https://www.nuget.org/stats/packages/system.diagnostics.contracts/4.0.1

Includes the following changes:
Updated to match style of redesign
Collapsible table
Loading spinner
Narrow screen compatible.
Removed Operation
Check box list no longer moves around the page.

Verified tab ordering, focus highlighting.
Adds aria-labels to non-anchor table cells for screen reader accessibility.

@microsoftsam @joelverhagen @scottbommarito @loic-sharma 

Screenshot:
image


src/NuGetGallery/Views/Statistics/_PivotTable.cshtml
```diff
@@ -1,77 +1,116 @@
 ﻿@model StatisticsPackagesViewModel
 
+<h1 class="statistics-report-title col-sm-12 page-heading">
+    Package Downloads for <a href="@Url.Package(Model.PackageId)">@Model.PackageId</a>
```
scottbommarito: Is this supposed to be the same title for both the all versions and specific versions page?

It appears like both are `Package Downloads for <id>` right now.
I would expect the specific versions page to be `Package Downloads for <id> <version>`.

src/NuGetGallery/Controllers/StatisticsController.cs
```diff
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using NuGet.Versioning;
```
joelverhagen: Sort usings

src/NuGetGallery/Controllers/StatisticsController.cs
```diff
@@ -19,14 +21,12 @@ public partial class StatisticsController
         private static readonly string[] PackageDownloadsByVersionDimensions = new[] {
             Constants.StatisticsDimensions.Version,
             Constants.StatisticsDimensions.ClientName,
-            Constants.StatisticsDimensions.ClientVersion,
-            Constants.StatisticsDimensions.Operation
+            Constants.StatisticsDimensions.ClientVersion
```
joelverhagen: super-nit: i've started putting commas at the end of the last item to make future merges a little less noisy

src/NuGetGallery/Scripts/perpackagestatsgraphs.js
```diff
@@ -1,123 +1,52 @@
...
-    if ($('#report-ClientName').length) {
-        if (Modernizr.svg) {
-            drawDownloadsByClientNameBarChart();
-        }
-    }
-    if ($('#report-Operation').length) {
-        if (Modernizr.svg) {
-            drawDownloadsByOperation();
-        }
+var graphData;
```
joelverhagen: spoke offline about this, these files should be a stats.min.js bundle that contains D3

Issue #4116, opened 2017-06-13
Address latest design changes on the package details page


joelverhagen:
Known changes:

- Expander chevron on the same line as the Dependencies and Version History headings

Other new changes to that page should be addressed as well.

PR #4123, opened 2017-06-13
Make Dependencies and Version History headings collapsable

joelverhagen:
Also, use plus/minus icons for expanding the version list.

https://github.com/NuGet/NuGetGallery/issues/4116


joelverhagen:
screencapture-localhost-packages-newtonsoft-json-1497395446281

PR #4125, opened 2017-06-13
[Redesign] Created Tool to Export Gallery's Static Assets

loic-sharma:
This will be used by the status and blog websites to consume the Gallery's assets.

### DIFF
--- a/tools/Export-StaticAssets.ps1
+++ b/tools/Export-StaticAssets.ps1
@@ -1,0 +1,21 @@
+<#
+.SYNOPSIS
+    Exports Compiled Non-Minified Static Assets
+
+.PARAMETER CssExportPath
+    The path to export the Gallery's CSS assets
+
+.PARAMETER JsExportPath
+    The path to export the Gallery's JS assets
+#>
+param(
+    [Parameter(Mandatory=$true)]
+    [string] $CssExportPath,
+
+    [Parameter(Mandatory=$true)]
+    [string] $JsExportPath)
+
+$galleryPath = Join-Path $PSScriptRoot "\..\src\NuGetGallery"
+
+Copy-Item (Join-Path $galleryPath "\Content\gallery\css\*") $CssExportPath
+Copy-Item (Join-Path $galleryPath "\Scripts\gallery\*") $JsExportPath
--- a/tools/Export-StaticAssets.ps1
+++ b/tools/Export-StaticAssets.ps1
@@ -17,5 +17,15 @@ param(
 
 $galleryPath = Join-Path $PSScriptRoot "\..\src\NuGetGallery"
 
+# Export all CSS assets
 Copy-Item (Join-Path $galleryPath "\Content\gallery\css\*") $CssExportPath
-Copy-Item (Join-Path $galleryPath "\Scripts\gallery\*") $JsExportPath
+
+# Export JS assets that do not have the "page-" prefix
+$scriptsPath = Join-Path $galleryPath "\Scripts\Gallery"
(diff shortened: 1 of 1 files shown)


tools/Export-StaticAssets.ps1
```diff
@@ -0,0 +1,31 @@
...
+    [string] $JsExportPath)
+
+$galleryPath = Join-Path $PSScriptRoot "\..\src\NuGetGallery"
+
+# Export all CSS assets
+Copy-Item (Join-Path $galleryPath "\Content\gallery\css\*") $CssExportPath
+
+# Export JS assets that do not have the "page-" prefix
+$scriptsPath = Join-Path $galleryPath "\Scripts\Gallery"
+$scripts = Get-ChildItem $scriptsPath | ? { $_.Name.StartsWith("page-") -eq $false }
```
joelverhagen: You can exclude `stats-` as well.

joelverhagen: (check latest `feature-redesign`)

tools/Export-StaticAssets.ps1
```diff
@@ -0,0 +1,31 @@
...
+    The path to export the Gallery's JS assets
+#>
+param(
+    [Parameter(Mandatory=$true)]
+    [string] $CssExportPath,
+
+    [Parameter(Mandatory=$true)]
+    [string] $JsExportPath)
+
+$galleryPath = Join-Path $PSScriptRoot "\..\src\NuGetGallery"
```
joelverhagen: How will we share images, common .cshtml, etc?

joelverhagen:
It might be worth investigating whether we can make a NuGet package with content files so that these things can be versioning and consumed easily. Not sure if it will make the whole thing easier or not.

xavierdecoster:
Why not use npm for these static assets?

tools/Export-StaticAssets.ps1
scottbommarito: I have updated the script to share images. IMO if we're sharing cshtml then we should just move the status page into the same project.

tools/Export-StaticAssets.ps1
scottbommarito: I don't see any scripts inside `Scripts\Gallery` that start with `stats-`. There are some in `Scripts\` that do, but this script only copies the JS inside `Scripts\Gallery`.

tools/Export-StaticAssets.ps1
joelverhagen: https://github.com/NuGet/NuGetGallery/tree/feature-redesign/src/NuGetGallery/Scripts/gallery

tools/Export-StaticAssets.ps1
joelverhagen: https://github.com/NuGet/NuGetGallery/tree/feature-redesign/src/NuGetGallery/Scripts/gallery

tools/Export-StaticAssets.ps1
joelverhagen: Status page should not be the same project. Gallery is too big and complex for a status page.

Same repo is a possibility but there are a couple of issues:

1. Status page is private right now.
1. Even if they are the same repo, they can still get out of sync since they are different deployments.
1. Status page will be uninteresting to almost all of gallery users IMO. It is tightly coupled with pingdom, azure storage, and the presence of a V3 API.

scottbommarito:
Decided to take another path for this. Will reopen if necessary.

Issue #4215, opened 2017-06-16
SemVer 2.0 - improve messaging for SemVer 2.0 upload and packages page


skofman1:
Exact messages will be provided by @anangaur 
Edit: Proving the details:

**Problem**
SemVer 2.0 packages will not be visible from legacy clients (i.e clients before NuGet 4.3 release). In order to make sure that publishers are aware that they are about to publish a SemVer 2 package, they should be warned appropriately so that this is a conscious decision and not an inadvertent one. 

Read more on [Identifying SemVer v2.0.0 Packages](https://github.com/NuGet/Home/wiki/SemVer2-support-for-nuget.org-%28server-side%29#identifying-semver-v200-packages)

One of the concerns is that a build metadata (e.g. 2.1.0**+buildmetadata**) also constitutes a Semver 2 version though this build metadata does not form the identity of the package. 

**Solution**
The idea here is to warn enough at the client (https://github.com/NuGet/Home/issues/5493) and on the NuGet.org server (this issue) when there is an intent or when the package is published.

Following is the list of cases and NuGet should identify to determine a SemVer 2 package and corresponding warning messages NuGet should output:
**Case 1:** Package has prerelease SemVer 2 versioning
**Warning1:** 
Title: This package has a SemVer2 package version
Message: This package will only be available to download with SemVer2 compatible NuGet clients, such as Visual Studio 2017 (version 15.3) and above or NuGet client 4.3 and above. [Read more](https://go.microsoft.com/fwlink/?linkid=852248)

**Case 2:** Package has a non-semver2 version, but is considered semver2 due to a dependency declaration
**Warning2**
Title: This package is considered a SemVer2 package as it has dependency on SemVer2 package(s)
Message: This package will only be available to download with SemVer2 compatible NuGet clients, such as Visual Studio 2017 (version 15.3) and above or NuGet client 4.3 and above. [Read more](https://go.microsoft.com/fwlink/?linkid=852248)

**Case 3:** Package has dependecy declared as SemVer 1 but the available dependency matching default dependency resolution Lowest itself is SemVer 2 (due to build metadata)
**Warning2** as described above.

NuGet Server should warn for the above scenarios during:
1. Package upload - Verify step
2. Package page

xavierdecoster:
Proposal: the message should be _in your face_ right next to the submit button on the VerifyPackage page.
Awaiting the final message, here's a suggestion.

When a package has a semver2 version itself:

[...]

PR #4239, opened 2017-06-20
Warn package authors when uploading a semver2 package through the web site

xavierdecoster:
This PR is a proposal we can start working from. Awaiting final wording from @anangaur. 

I'd suggest to have the warning as close as possible to the submit button to ensure maximum visibility (people don't always read messages at the top of the page).

When a package has a semver2 version itself:

image

When a package has a non-semver2 version, but is considered semver2 due to a dependency declaration:

image

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1223,18 +1223,24 @@ namespace NuGetGallery
                 }
             }
 
+            var dependencyGroups = packageMetadata.GetDependencyGroups();
+
             var model = new VerifyPackageRequest
             {
                 Id = packageMetadata.Id,
                 Version = packageMetadata.Version.ToFullStringSafe(),
                 OriginalVersion = packageMetadata.Version.OriginalVersion,
+                HasSemVer2Version = packageMetadata.Version.IsSemVer2,
+                HasSemVer2Dependency = dependencyGroups.Any(d => d.Packages.Any(
+                                p => (p.VersionRange.HasUpperBound && p.VersionRange.MaxVersion.IsSemVer2)
+                                    || (p.VersionRange.HasLowerBound && p.VersionRange.MinVersion.IsSemVer2))),
                 LicenseUrl = packageMetadata.LicenseUrl.ToEncodedUrlStringOrNull(),
                 Listed = true,
                 Language = packageMetadata.Language,
                 MinClientVersion = packageMetadata.MinClientVersion,
                 FrameworkReferenceGroups = packageMetadata.GetFrameworkReferenceGroups(),
                 Dependencies = new DependencySetsViewModel(
-                    packageMetadata.GetDependencyGroups().AsPackageDependencyEnumerable()),
+                    dependencyGroups.AsPackageDependencyEnumerable()),
                 DevelopmentDependency = packageMetadata.GetValueFromMetadata("developmentDependency"),
                 Edit = new EditPackageVersionRequest
                 {
--- a/src/NuGetGallery/Views/Packages/VerifyPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/VerifyPackage.cshtml
@@ -146,12 +146,12 @@
         }
 
 
-                                    @if (!String.IsNullOrEmpty(Model.LicenseUrl))
+        @if (!String.IsNullOrEmpty(Model.LicenseUrl))
         {
             <li>
                 <h4>Requires License Acceptance
                     @{
-                                            var formid2 = "Edit_RequiresLicenseAcceptance";
+                        var formid2 = "Edit_RequiresLicenseAcceptance";
                         <button type="button" class="edit-button" id="@(formid2 + "Button")"><span class="icon-edit" title="Edit Requires License Acceptance"></span></button>
                         <button type="button" class="undo-button" id="@("Undo" + formid2)" style="display: none"><span class="icon-undo"></span></button>
                     }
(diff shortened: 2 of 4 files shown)


src/NuGetGallery/Views/Packages/VerifyPackage.cshtml
```diff
@@ -201,8 +201,44 @@
...
+            <span class="icon-stack warningsign">
+                <i class="icon-circle icon-stack-base"></i>
+                <i class="icon-exclamation"></i>
+            </span>
+            @if (Model.HasSemVer2Version)
+            {
+                <text>
+                    Warning: This package has a SemVer2 package version. <br />
+                    <span style="font-weight: normal; font-style:italic;">
+                        This package will only be available for download with SemVer2-compatible NuGet clients, such as Visual Studio 2017 (version 15.3) and above or NuGet client 4.3.0 and above.<br />
```
joelverhagen: `SemVer 2.0.0` instead of `SemVer2`

xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/4239/commits/57fe6cfe939a717f59d0a2f6d7c63fc9d0a76d30

src/NuGetGallery/Views/Packages/VerifyPackage.cshtml
```diff
@@ -201,8 +201,40 @@
...
+                    Warning: This package is considered a SemVer 2.0.0 package as it has a package dependency declaring a SemVer 2.0.0 version range.<br />
+                </text>
+            }
+            <span style="font-weight: normal; font-style:italic;">
+                This package will only be available for download with SemVer 2.0.0 compatible NuGet clients, such as Visual Studio 2017 (version 15.3) and above or NuGet client 4.3.0 and above.<br />
+                Reach out to support if you have more questions.
+            </span>
+            <br />
+            <br />
+            <input id="verifyUploadSubmit" type="submit" value="Submit" title="Verify Details &amp; Submit" />
```
scottbommarito: Two things:
1 - I don't like that we have the HTML for this button in two places.
2 - I think it looks a little odd that the button is inside the alert. Additionally, we don't have any buttons inside other alerts.

I think we could resolve both of these issues by moving the button outside of the alert.

xavierdecoster: I really really want to have this warning as much in the way as possible so authors _have to_ read it..

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1223,18 +1223,24 @@ private string GetSecurePushPolicyDescriptions()
...
             }
 
+            var dependencyGroups = packageMetadata.GetDependencyGroups();
+
             var model = new VerifyPackageRequest
             {
                 Id = packageMetadata.Id,
                 Version = packageMetadata.Version.ToFullStringSafe(),
                 OriginalVersion = packageMetadata.Version.OriginalVersion,
+                HasSemVer2Version = packageMetadata.Version.IsSemVer2,
```
xavierdecoster: Please note that the message is different when the package has a semver2 version itself, or only has a semver2 package dependency.

src/NuGetGallery/Views/Packages/VerifyPackage.cshtml
xavierdecoster: Moving the button out of the warning banner. Seems _in your face_ enough indeed. As long as it is close to the actionable button of this page so authors can't just skip it "by accident" :)

image

src/NuGetGallery/Controllers/PackagesController.cs
xavierdecoster: I see the value from technical point of view to reuse the `SemVerLevelKey` class to determine `IsSemVer2`. However, I also see value in really being able to point the author to exactly what piece of metadata is causing the warning. We can only do that by distinguishing both scenarios. Would like to hear what @anangaur thinks about this one :)

xavierdecoster:
Ping @anangaur

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Oh I forgot about the method being different. Can we split the `SemVerLevelKey.ForPackage` method into two parts--`HasSemVer2Version` and `HasSemVer2Dependency` and then call those methods here?

src/NuGetGallery/Views/Packages/VerifyPackage.cshtml
scottbommarito: Yea I didn't want to move the warning away--I like it right next to the button--but I just wanted to move the button out of the yellow. Looks great!

anangaur:
I agree we should make a differentiation. And this differentiation should be conistent across sever and client. The text needs some minor tweaks though and a Fwd link and I will be able to provide the same, later.
I see the following cases:
1. The package itself is SemVer 2 - **Msg1**
2a. Package has dependencies declared in SemVer 2 - **Msg 2**
2b. Package has dependecy declared as SemVer 1 but the package itself is SemVer 2 (due to build metadata) - **Msg 2**

xavierdecoster:
@anangaur about case 2b: I think what you describe there is what we understood to be a [non-goal](https://github.com/NuGet/Home/wiki/SemVer2-support-for-nuget.org-%28server-side%29). There's no way we can detect that a dependency that would be resolved is semver2 for a given non-semver2 dependency version range. That would require analysis of all restore graphs, which is currently not taken into account to determine whether the package is semver2. So we can only detect two cases: 1 and 2a (or just 2 :))

xavierdecoster:
Thanks for the reviews! Holding off until I receive final wording from @anangaur.

anangaur:
@xavierdecoster I thought we can still figure out case 2b not for all dependencies but for 1st level. Atleast thats what I understood based on the discussions with @joelverhagen and @ryuyu 

**Case 1:** We should be able to detect PackageA is SemVer 2 as direct dependency is SemVer 2
PackageA 1.0.0
|--- PackageB 2.0.0 (SemVer 2 as the real version is 2.0.0+buildmetadata)

**Case 2:** We cannot determine Package A to be SemVer 2 as the dependency is deeper than 1 level. Here PackageC is dependent on a SemVer package and not PackageA directly.

PackageA 1.0.0
|--- PackageC 3.0.0
........| --- PackageB 2.0.0 (SemVer 2 as the real version is 2.0.0+buildmetadata)

xavierdecoster:
@anangaur @joelverhagen @ryuyu let's have the discussion here, as hallway discussions don't make it across the ocean ;)

So, we can indeed query case 1, but isn't that raising a warning on what may be a temporary situation? 

The nuspec itself does not contain any semver2 version or version range, and at any point in time, a semver2 package dependency that matches the minimum constraint may be listed or unlisted.

e.g. we may not be raising the warning at upload time when the lowest package version matching the lowerbound is v2.0.1. A week later, 2.0.0+metadata is unlisted, still matching the constraint, turning this package suddenly into a semver2 package.

The opposite may also be true: we warn about the package being semver2 when 2.0.0+metadata is the lowerbound best match, whilst it may be unlisted some time later and a semver1 package version becomes the best match for this lowerbound.

In both cases, the (lack of) warning may be experienced as inconsistent, whilst now we only take into account what is physically in the nuspec. I much prefer consistency and easy to understand warnings :)

shishirx34:
Correct me here, if I am misunderstanding something
> 1. The package itself is SemVer 2 - Msg1
> 2a. Package has dependencies declared in SemVer 2 - Msg 2
> 2b. Package has dependecy declared as SemVer 1 but the package itself is SemVer 2 (due to build  metadata) - Msg 2

Isn't 1. and 2b. the same described here?

xavierdecoster:
@shishirx34 rephrased (to how I understood it :))

> 2b. Package has dependecy declared as SemVer 1 but the **available dependency matching default dependency resolution Lowest** itself is SemVer 2 (due to build metadata) - Msg 2

anangaur:
@xavierdecoster  thanks for rephrasing. The flip side of not reporting warning for 2b is that the publisher will be happy but the package will still end up being not compatible on the older clients. Now finding out the real issue here would be non-trivial for the publisher when this issue is reported.

xavierdecoster:
See my comment here on **Case 3**: https://github.com/NuGet/NuGetGallery/issues/4215#issuecomment-311272659

> I do want to point out that none of the server APIs are considering the package to be SemVer 2.0.0 in Case 3.
> 
> These packages are not considered to be SemVer 2.0.0 and e.g. NOT HIDDEN on V2 OData endpoints, as all we check is the package's nuspec manifest.
> 
> This is one of the edge cases described in the non-goals (paragraph 2) that may lead to (gracefully handled) restore failures on non-SemVer2 compatible clients.

PR #4241, opened 2017-06-20
[Redesign] Contact Owners Page

loic-sharma:
Redesigned the contact owners page.

contact-owners

@microsoftsam

### DIFF
--- a/src/NuGetGallery/Views/Packages/ContactOwners.cshtml
+++ b/src/NuGetGallery/Views/Packages/ContactOwners.cshtml
@@ -1,43 +1,68 @@
 ﻿@model ContactOwnersViewModel
 @{
     ViewBag.Tab = "Packages";
-    var owners = Model.Owners.Flatten(@<text>@item.Username</text>);
+    Layout = "~/Views/Shared/Gallery/Layout.cshtml";
 }
 
-<h1 class="page-heading">Contact the Owners of "@Model.PackageId.Abbreviate(50)"</h1>
-@if (Model.Owners.Any())
-{
-    <p class="message">
-        By submitting this form, you agree to <em>disclose your email address</em> 
-        to the package owners listed below so they can reply to you.
-    </p>
-    using (Html.BeginForm())
-    {
-        <fieldset class="form">
-            <legend>Contact Owners</legend>
-            @Html.AntiForgeryToken()
-            <div class="form-field">
-                <label>To</label>
-                <p class="authors"><strong>@owners</strong></p>
+<section role="main" class="container main-container page-contact-owners">
+    <div class="row forgot-form">
+        <div class="col-sm-offset-2 col-sm-8">
+            <h1 class="text-center">Contact the Owners of <a href="@Url.Package(Model.PackageId)">@Model.PackageId.Abbreviate(50)</a></h1>
+            @if (Model.Owners.Any())
+            {
+                @ViewHelpers.AlertInfo(
+                    @<text>
+                        By submitting this form, you agree to disclose your email address
+                        to the package owners listed below so they can reply to you.
+                    </text>
+                );
+
+                using (Html.BeginForm())
+                {
+                    @Html.AntiForgeryToken()
+
+                    <div class="form-group">
+                        <label>To</label>
+                        <p>
+                            @foreach (var owner in Model.Owners)
+                            {
+                                <a href="@Url.User(owner)" title="View @owner.Username's profile" class="ms-font-xl">@owner.Username</a>
+                            }
+                        </p>
+                    </div>
+                    <div class="form-group">
+                        <label>From</label>
+                        <p class="ms-font-xl">@CurrentUser.EmailAddress</p>
+                    </div>
+                    <div class="form-group @Html.HasErrorFor(m => m.Message)">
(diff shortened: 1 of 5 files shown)


scottbommarito:
Why is there a cancel button? None of our other forms have a button like this.

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
```diff
@@ -1,43 +1,68 @@
...
+                        </div>
+                    </div>
+                }
+            }
+            else
+            {
+                @ViewHelpers.AlertWarning(
+                    @<text>
+                        Sorry, the owners of this package do not allow contacting them through this form.
+                        Try visiting the project home page for "@Model.PackageId" in order to contact the package owner.
```
scottbommarito: This message is a little bit...awkward.

1 - It's called "Project Site" on the display package page. It's not entirely clear that's what this message is referring to.
2 - Can't we just link the project url here directly?
3 - What happens if the package doesn't have a project url?

joelverhagen: I think this is meant to be a very generic suggestion rather than a call out to use a different part of our UI (although that could be a convenient way to do this). Even if there is no project URL baked into the .nuspec, we can suggest that the user contacts the package owners via another channel. Perhaps we can broaden the recommendation to be "project site or source repository" or something even more generic.

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
```diff
@@ -1,43 +1,68 @@
...
-            <div class="form-field">
-                @Html.LabelFor(m => m.Message)
-                @Html.TextAreaFor(m => m.Message, 10, 50, null)
-                @Html.ValidationMessageFor(m => m.Message)
-            </div>
-            <div class="form-field">
-                @Html.CheckBoxFor(m => m.CopySender)
-                @Html.LabelFor(m => m.CopySender, new { @class = "checkbox" })
+<section role="main" class="container main-container page-contact-owners">
+    <div class="row forgot-form">
```
joelverhagen: What's `forgot-form` for?

scottbommarito: We should rename this CSS class

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
```diff
@@ -1,43 +1,68 @@
...
+        <div class="col-sm-offset-2 col-sm-8">
```
scottbommarito: Move this to 10 columns `md` with offset 1 as discussed.

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
scottbommarito: Good point. I just think we should change this message to better convey that we're basically saying "please find another means of contact for this owner".

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
loic-sharma: I don't think this class is needed. This was likely a copy-paste mistake.

loic-sharma:
Removed the cancel button and improved the error message if none of the authors can be contacted.

loic-sharma:
:bell:

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
```diff
@@ -1,43 +1,74 @@
...
+                        @Html.ShowValidationMessagesFor(m => m.CopySender)
+                    </div>
+                    <div class="form-group">
+                        <input type="submit" class="btn btn-primary form-control" value="Send" />
+                    </div>
+                }
+            }
+            else
+            {
+                if (string.IsNullOrEmpty(Model.ProjectUrl))
```
joelverhagen: Nice one.

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
```diff
@@ -1,43 +1,74 @@
...
+                {
+                    @ViewHelpers.AlertWarning(
```
scottbommarito: You can combine this `if` `else`:

```
@ViewHelpers.AlertWarning(
    @<text>
        Sorry, the owners of this package do not allow contacting them through this form.
        @if (string.IsNullOrEmpty(Model.ProjectUrl))
        {
            <text>
                You can try contacting the package owners of "@Model.PackageId" through its <a href="@Model.ProjectUrl">Project Site</a>.
            </text>
        }
    </text>
);
```

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
```diff
@@ -1,43 +1,74 @@
...
+                        @<text>
+                            Sorry, the owners of this package do not allow contacting them through this form.
```
scottbommarito: I'd really like to see something that explicitly says `"You'll have to find a different way to contact them."` or something similar. Otherwise, users who encounter this don't have a clear direction of what to do next.

Issue #4252, opened 2017-06-20
Column layouts formalization affecting certain pages


jonwchu:
We formalized which pages should use a specific column layout and should be adjusted on the current designs. 

**4 Column Layout (6 on XS Screen Size): Authentication**
These pages have all their content (including title and text) centered in the page:

- Sign-in 
- Register
- Forgot Password
- Reset Password 

**10 Column Layout: Forms**
These pages have centered title but left-aligned text:

- Report Abuse
- Contact Support
- Contact Owners
- Manage Owners
- Package Listing
- Upload Package
- Edit Package
- Account Info

**12 Column Layout: Content Pages**
These pages have left-aligned title and text.

jonwchu:
Mock-ups on InVision have been updated to reflect the new column layouts.

PR #4253, opened 2017-06-20
[Redesign] Updated Forgot Password Pages' Column Layouts

loic-sharma:
This moves them to a 4-column layout with centered text to be consistent with #4252 

forgot-password

password-sent

@microsoftsam

### DIFF
--- a/src/NuGetGallery/Views/Users/ForgotPassword.cshtml
+++ b/src/NuGetGallery/Views/Users/ForgotPassword.cshtml
@@ -5,12 +5,13 @@
 }
 
 <section role="main" class="container main-container page-forgot-password">
-    <h1>Forgot Password</h1>
+    <div class="col-sm-4 col-sm-offset-4 text-center">
+        <h1>Forgot Password</h1>
 
-    <p>We are sorry to hear you forgot your NuGet.org account password. Enter your email below and we will send instructions to reset your password.</p>
+        <p>We are sorry to hear you forgot your NuGet.org account password. Enter your email below and we will send instructions to reset your password.</p>
 
-    <div class="row forgot-form">
-        <div class="col-sm-offset-3 col-sm-6">
+        <div class="row forgot-form text-left">
+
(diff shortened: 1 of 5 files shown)


src/NuGetGallery/Views/Users/ForgotPassword.cshtml
```diff
@@ -5,12 +5,13 @@
 }
 
 <section role="main" class="container main-container page-forgot-password">
-    <h1>Forgot Password</h1>
+    <div class="col-sm-4 col-sm-offset-4 text-center">
+        <h1>Forgot Password</h1>
 
-    <p>We are sorry to hear you forgot your NuGet.org account password. Enter your email below and we will send instructions to reset your password.</p>
+        <p>We are sorry to hear you forgot your NuGet.org account password. Enter your email below and we will send instructions to reset your password.</p>
```
scottbommarito: I don't like the center on this text and the forgot password text...I think it should be left-align.

microsoftsam: Yeah I agree, and for the sake of consistency with the other pages (sign in), it would make sense to be full width and left aligned.

scottbommarito: I think we're planning on moving Sign In to this format--not full width and centered `<p>` text. I think we should keep it left-aligned `<p>` text.

loic-sharma: Yes, the sign-in page will be 4 columns as well with centered text. @scottbommarito: I think that having centered text on the forgot page but left-aligned text on the password sent page would be jarring. Thoughts?

scottbommarito: I was thinking both left-aligned. I'm fine with both centered as well.

loic-sharma:
🔔

PR #4265, opened 2017-06-22
[Redesign] Toggle "slider"

scottbommarito:
100% CSS!

This is on the report abuse page (`https://localhost/packages/<id>/<version>/ReportAbuse`)

image

image

image

@microsoftsam


joelverhagen:
What browsers and devices have we tried. There's a lot of tricky CSS in there.

Issue #4281, opened 2017-06-26
When IsLatest is not set on a package user profile page doesn't load


skofman1:
For example https://www.nuget.org/profiles/kzu
The package Moq doesn't have IsLatest or IsLatestStable set.

Call stack:
System.InvalidOperationException:
   at System.Linq.Enumerable.Single (System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089)
   at NuGetGallery.PackageService.MergeLatestPackagesByOwner (NuGetGallery, Version=4.3.0.0, Culture=neutral, PublicKeyToken=null)
   at NuGetGallery.PackageService.FindPackagesByOwner (NuGetGallery, Version=4.3.0.0, Culture=neutral, PublicKeyToken=null)
   at NuGetGallery.UsersController.Profiles (NuGetGallery, Version=4.3.0.0, Culture=neutral, PublicKeyToken=null)
   at lambda_method (Anonymously Hosted DynamicMethods Assembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null)
   at System.Web.Mvc.ReflectedActionDescriptor.Execute (System.Web.Mvc, Version=5.2.3.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
   at System.Web.Mvc.ControllerActionInvoker.InvokeActionMethod (System.Web.Mvc, Version=5.2.3.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
   at System.Web.Mvc.Async.AsyncControllerActionInvoker.<BeginInvokeSynchronousActionMethod>b__39 (System.Web.Mvc, Version=5.2.3.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
   at System.Web.Mvc.Async.AsyncResultWrapper+WrappedAsyncResult`2.CallEndDelegate (System.Web.Mvc, Version=5.2.3.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
   at System.Web.Mvc.Async.AsyncControllerActionInvoker.EndInvokeActionMethod (System.Web.Mvc, Version=5.2.3.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
[...]

PR #4282, opened 2017-06-26
Fix InvalidOperationException from profile when multiple version are set to latest

chenriksson:
Fix for #4281

Note that this selects later version if multiple since order by is descending, whereas DisplayPackage history selects earlier version due to using First with order ascending.

Will file separate bug to fix IsLatest flags with reflow package, as well as trying to make selection of later latest consistent.

### DIFF
--- a/src/NuGetGallery/Services/PackageService.cs
+++ b/src/NuGetGallery/Services/PackageService.cs
@@ -301,7 +301,7 @@ namespace NuGetGallery
 
             foreach (var latestStablePackagesById in latestStablePackageVersions.ToList().GroupBy(p => p.PackageRegistration.Id))
             {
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -301,16 +301,18 @@ private void MergeLatestStablePackagesByOwner(User user, bool includeUnlisted, D
 
             foreach (var latestStablePackagesById in latestStablePackageVersions.ToList().GroupBy(p => p.PackageRegistration.Id))
             {
+                // Use First() instead of Select() to get the latest package, in case there are multiple latest due to concurrency issue
```
agr: You mean `Single` instead of `Select`, right?
(below, on line 344, too)

chenriksson: yes :-)

PR #4283, opened 2017-06-26
DB schema changes for repo URL and ReadMe

laurenshen:
-Reviewed with Christy, Fay, and Josh but opening up for more feedback


src/NuGetGallery.Core/Entities/PackageHistory.cs
```diff
@@ -65,6 +66,7 @@ public PackageHistory(Package package)
         public string IconUrl { get; set; }
         public string LicenseUrl { get; set; }
         public string ProjectUrl { get; set; }
+        public string RepositoryUrl { get; set; }
```
scottbommarito: Are we sure we don't want to have any state here about the ReadMe? I wouldn't store the whole thing but wouldn't the same enum we use in `PackageEdits` (added/changed, deleted, unchanged) make sense to track here?

src/NuGetGallery/Migrations/201706221634506_AddRepositoryURL_ReadMe.Designer.cs
```diff
@@ -0,0 +1,29 @@
+// <auto-generated />
+namespace NuGetGallery.Migrations
+{
+    using System.CodeDom.Compiler;
+    using System.Data.Entity.Migrations;
+    using System.Data.Entity.Migrations.Infrastructure;
+    using System.Resources;
+    
+    [GeneratedCode("EntityFramework.Migrations", "6.1.3-40302")]
+    public sealed partial class AddRepositoryURL_ReadMe : IMigrationMetadata
```
scottbommarito: I see 2 migrations with same name here.
- `201706221609417_AddRepositoryURL_ReadMe`
- `201706221634506_AddRepositoryURL_ReadMe`

As far as I can tell, according to the `.csproj`, `201706221609417_AddRepositoryURL_ReadMe` isn't even used. Can we remove all of the `201706221609417_AddRepositoryURL_ReadMe` files?

If it IS used, can we rename it to something else that better describes what it does or combine both into a single migration?

tests/NuGetGallery.Core.Facts/Entities/PackageFacts.cs
```diff
@@ -22,6 +22,7 @@ public void ApplyPackageEditUpdatesPackage()
                     FlattenedAuthors = "OldAuthors",
                     Copyright = "OldCopyright",
                     ReleaseNotes = "OldReleaseNotes",
+                    RepositoryUrl = "OldRepositoryUrl",
```
scottbommarito: Can we add `RepositoryUrl` to some other unit tests? I'm sure we have some tests on `Package` and `PackageHistories` that involve specifying all of the fields on the objects.

src/NuGetGallery.Core/Entities/PackageHistory.cs
laurenshen: Package History table is deprecated and Christy just filed an issue to investigate removing it altogether, but we left the Repository URL since it was package metadata

scottbommarito: IMO until we remove it entirely we should still add some ReadMe metadata to this file--should be really quick.

src/NuGetGallery/Migrations/201706221609417_AddRepositoryURL_ReadMe.cs
```diff
@@ -0,0 +1,18 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddRepositoryURL_ReadMe : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.Packages", "HasReadMe", c => c.Boolean(nullable: false));
```
joelverhagen: What does this do to existing records?

src/NuGetGallery/Migrations/201706221609417_AddRepositoryURL_ReadMe.cs
```diff
@@ -0,0 +1,18 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddRepositoryURL_ReadMe : DbMigration
```
joelverhagen: Is this a duplicate of the migration below?

src/NuGetGallery/Migrations/201706221609417_AddRepositoryURL_ReadMe.cs
laurenshen: HasReadMe column will default to false for existing records

tests/NuGetGallery.Core.Facts/Entities/PackageFacts.cs
laurenshen: PackageFacts.cs is the only existing test I see related to the fields; do you know if where there are other tests?

src/NuGetGallery/Migrations/201706262253021_AddRepositoryURL_ReadMe.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddRepositoryURL_ReadMe : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.Packages", "RepositoryUrl", c => c.String());
+            AddColumn("dbo.Packages", "HasReadMe", c => c.Boolean(nullable: false));
```
skofman1: when you apply this migration on an existing table, nullable: false will cause the migration to fail (since existing row don't have this value), or will cause all the column to be set with a default false value (which will take a looooong time). So nullable should be true.

src/NuGetGallery/Migrations/201706221609417_AddRepositoryURL_ReadMe.cs
skofman1: See my comment bellow. If you want to apply this to prod, please check how long it will take to apply, and the DTU impact.

tests/NuGetGallery.Core.Facts/Entities/PackageFacts.cs
scottbommarito: if that's true then that's fine

src/NuGetGallery.Core/Entities/PackageEdit.cs
```diff
@@ -44,6 +44,11 @@ public class PackageEdit
         public string IconUrl { get; set; }
         public string LicenseUrl { get; set; }
         public string ProjectUrl { get; set; }
+        public string RepositoryUrl { get; set; }
+        /// <summary>
+        /// changed, unchanged, deleted
+        /// </summary>
+        public string ReadMeState { get; set; }
```
scottbommarito: Can we put these in constants somewhere?

```
public const string ReadMeChanged = "changed";
public const string ReadMeUnchanged = "unchanged";
public const string ReadMeDeleted = "deleted";
```

laurenshen: Fay is going to put those in on her side, and we're changing "unchanged" to stay as null since the default for ReadMeState is null

PR #4284, opened 2017-06-26
[Redesign]Upload and Verify Package

ryuyu:
This PR merges the Upload and Verify pages, and updates them for the redesign.

Behaviour Changes:
Uploading a package will load the verify in page (previous behaviour: navigate to a new verify page).
Cancelling a verify will cancel in page.
In progress upload will be indicated and can be cancelled on page.
Icon preview was added to verify. Updates dynamically when iconUrl field is changed.
Upload/Verify sections of the page are collapsible.

Upload:
image

Awaiting Verification:
image

Continue Upload:
image

Error:
image


src/Bootstrap/less/theme/base.less
```diff
@@ -85,6 +85,20 @@ body {
...
+  h4 > span {
+    border-bottom: 1px solid #d8d8d8;
+    margin-bottom: 3px;
+  }
+}
+
+.dependency-group {
+
+  li span {
+    color: #666;
```
loic-sharma: This is the same as `@gray-light`, right?

src/Bootstrap/less/theme/page-upload.less
```diff
@@ -0,0 +1,75 @@
+.page-upload {
+    h2 {
```
loic-sharma: Joel's account page is doing this change globally in base.less. Maybe copy his changes there and remove the duplication when both changes are merged?

src/Bootstrap/less/theme/page-upload.less
```diff
@@ -0,0 +1,75 @@
...
+        .ms-Icon {
+            font-size: 0.6em;
+            position: relative;
+            top: -2px;
+        }
+
+        a {
+            color: #000;
```
loic-sharma: Use `@gray-base` here

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+        if (!form.action) {
+            return;
+        }
+
+        // count the number of file fields which have selected files
+        var totalFile = 0;
+        $('input[type=file]', form).each(function (index, el) { if (el.value) totalFile++; });
+
+        // only show progress indicator if the form actually uploads some files
+        if (totalFile > 0) {
```
loic-sharma: Is this still needed? Looks like dead code.

src/NuGetGallery/ViewModels/DependencySetsViewModel.cs
```diff
@@ -58,10 +58,13 @@ public DependencyViewModel(string id, string versionSpec)
                 {
                     VersionSpec = VersionRange.Parse(versionSpec).PrettyPrint();
                 }
+
+                PackageUrl = UrlExtensions.Package(new System.Web.Mvc.UrlHelper(System.Web.HttpContext.Current.Request.RequestContext), id);
```
loic-sharma: Can you clean this up with usings? Or are there name conflicts?

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+            $('#input-select-file').click();
+        })
+
+        $('#input-select-file').on('change', function () {
+            clearErrors();
+            var fileName = $('#input-select-file').val().split("\\").pop();
+
+            if (fileName.length > 0) {
+                $('#file-select-feedback').attr('placeholder', fileName);
+                // Whether the cancel fails or not, we want to upload the next one.
```
scottbommarito: This comment confused me at first. Basically the behavior here is "when you select a file to upload, we cancel any ongoing upload and then upload the new file". Could we change the comment so this better explains that?
```
// Cancel any ongoing upload, and then start the new upload.
// If the cancel fails, still try to upload the new one.
```

I expect that the majority of times people will try to begin a new upload, they will not have an ongoing upload. This "always try to cancel first" behavior will add an extra request to the mix. This isn't too big of a deal but it's something to note.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+            bindData(InProgressPackage);
+        }
+
+        if (_isWebkitBrowser) {
+            constructIframe(jQueryUrl);
+        }
+    }
+
+    function startUploadAsync(callback, error) {
+        // count the number of file fields which have selected files
```
scottbommarito: If we know there's only one `input[type=file]`, why do we need to count them all? Can't we just check the expected id itself? There could be some odd behavior if we ever add additional file inputs to the page, e.g. ReadMe support.

 Also I'd say `input file elements` instead of `file fields` in the comment.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+            data: new FormData($('#' + _formId)[0]),
+
+            cache: false,
+            contentType: false,
+            processData: false,
+
+            success: function (model, resultCodeString, fullResponse) {
+                bindData(model);
+                endProgressBar();
+                if (callback) {
```
scottbommarito: this can be simplified to
```
callback && callback();
```
right?

this also applies to the `if (error) { error(); }` below as well

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+            type: 'POST',
+
+            data: new FormData($('#verify-metadata-form')[0]),
+
+            cache: false,
+            contentType: false,
+            processData: false,
+
+            success: function (model, resultCodeString, fullResponse) {
+                document.location = model.location;
```
scottbommarito: Shouldn't this call `callback`? If not, then this function should not accept a `callback`.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+            contentType: false,
+            processData: false,
+
+            success: function (model, resultCodeString, fullResponse) {
+                bindData(model);
+                endProgressBar();
+                if (callback) {
+                    callback();
+                }
+
```
loic-sharma: Extra line here.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+                    error();
+                }
+            }
+
+        });
+    }
+
+    function displayErrors(errors) {
+        if (errors == null || errors.length < 1) {
+            return;
```
scottbommarito: My understanding is that this will happen when our AJAX request failed and provided a null or empty response JSON. When does that happen?

Isn't there a better way for us to handle this than no-op'ing?
1 - Display a default error message
2 - Automatically retry the request

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        var failureContainer = $("#validation-failure-container");
+        var failureListContainer = document.createElement("div");
```
scottbommarito: Why do we need to recreate this element every time? You should be able to get away with leaving the element there and using `.show()` and `.hide()` in conjunction with `ko.applyBindings`.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+        $("#validation-failure-list").remove();
+    }
+
+    function bindData(model) {
+        $("#verify-package-block").remove();
+        if (model == null) {
+            return;
+        }
+
+        var reportContainerElement = document.createElement("div");
```
scottbommarito: Why do we need to recreate this element every time? You should be able to get away with leaving the element there and using `.show()` and `.hide()` in conjunction with `ko.applyBindings`. I would prefer to avoid creating elements in the DOM as much as possible, and it's probably confusing for a user using accessibility tools to understand an element that pops into the page as `aria-expanded = true` and never has an `aria-expanded = false` state.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+        ko.applyBindings({ data: model }, reportContainerElement);
+
+        $('#verify-cancel-button').on('click', function () {
+            $('#verify-cancel-button').attr('disabled', 'disabled');
+            $('#verify-cancel-button').attr('value', 'Cancelling');
+            $('#verify-cancel-button').addClass('.loading');
+            $('#verify-submit-button').attr('disabled', 'disabled');
+            $('#input-select-file').val("");
+            $('#file-select-feedback').attr('placeholder', 'Browse to select a package file...');
+            // Whether the cancel fails or not, we want to upload the next one.
```
scottbommarito: I think this is leftover from copying code above--it doesn't make sense in this context.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+
+        window.nuget.configureExpander(
+            "verify-package-form",
+            "ChevronRight",
+            "Verify",
+            "ChevronDown",
+            "Verify");
+    }
+
+    function submitForm(form) {
```
scottbommarito: Does this function do anything now? Looks like you removed all the code that actually made it do things. Should probably delete it.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
...
+        }
+    }
+
+    function setProgressIndicator(percentComplete, fileName) {
+        $("#upload-progress-bar").width(percentComplete + "%")
+            .attr("aria-valuenow", percentComplete)
+            .text(percentComplete + "%");
+    }
+
+    // obsolete
```
scottbommarito: I would delete it rather than leave it laying around.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1191,6 +1326,7 @@ private string GetSecurePushPolicyDescriptions()
             return Redirect(urlFactory(package));
         }
 
+        [Obsolete]
```
loic-sharma: Can you disable the obsolete warnings on the unit tests that still use this method?

```
#pragma warning disable 0612
...
#pragma warning restore 0612
```

src/Bootstrap/less/theme/page-upload.less
ryuyu: If that page is doing it globally, we can just drop it here.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -201,13 +253,15 @@ public virtual ActionResult UploadPackageProgress()
             if (uploadFile == null)
             {
                 ModelState.AddModelError(String.Empty, Strings.UploadFileIsRequired);
-                return View();
+                Response.StatusCode = 400;
+                return Json(new List<string>() { Strings.UploadFileIsRequired });
```
ryuyu: done

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1264,20 +1400,30 @@ private string GetSecurePushPolicyDescriptions()
...
             Package package;
             using (Stream uploadFile = await _uploadFileService.GetUploadFileAsync(currentUser.Key))
             {
                 if (uploadFile == null)
                 {
                     TempData["Message"] = "Your attempt to verify the package submission failed, because we could not find the uploaded package file. Please try again.";
-                    return new RedirectResult(Url.UploadPackage());
+
+                    Response.StatusCode = 400;
+                    return Json(new List<string>() { "Your attempt to verify the package submission failed, because we could not find the uploaded package file. Please try again." });
```
ryuyu: No idea. I just copied the string that was used before. I would like to try to limit the scope of these changes of possible. I assume this was done this way because it was added later and the string wasn't reused anywhere right now.

src/NuGetGallery/Views/Packages/VerifyPackage.cshtml
```diff
@@ -3,15 +3,13 @@
 @{
```
ryuyu: I was leaving it since i left the controller method. We can clean it up if anybody feels strongly.

src/NuGetGallery/Views/Packages/UploadPackage.cshtml
```diff
@@ -2,48 +2,85 @@
 @{
     ViewBag.Title = "Upload Package";
     ViewBag.Tab = "Upload";
+    Layout = "~/Views/Shared/Gallery/Layout.cshtml";
 }
 
-<h1 class="page-heading">Upload Your Package</h1>
+<div class="page-upload container main-container">
+    <h1 class="page-heading col-xs-10 col-xs-offset-1 text-center">Upload Your Package</h1>
```
ryuyu: Done.

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
```diff
@@ -0,0 +1,207 @@
...
+    <!-- ko ifnot: $data.Link -->
+    <p data-bind="text: DisplayText"></p>
+    <!-- /ko-->
+    <!-- /ko -->
+    <!-- ko ifnot: DisplayText -->
+    <p><i data-bind="text:'(none specified)'"></i></p>
+    <!-- /ko -->
+</script>
+
+<script type="text/html" id="display-dependencysets">
```
ryuyu: dependencyset is one thing, so i left this as the name.

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
```diff
@@ -0,0 +1,207 @@
...
+            <li>
+                <h4>
+                    <span data-bind="text: _propkey"></span>
+                </h4>
+                <ul data-bind="foreach: _data[_propkey]" class="list-unstyled dependency-group">
+                    <li>
+                        @*class="dependency-container">*@
+                        <!-- ko if: $data -->
+                        <a class="dependency-id" data-bind="attr: { href: PackageUrl }, text: Id"></a>
+                        <span class="dependency-versionspec" data-bind="text: VersionSpec"></span>
```
ryuyu: VersionSpec is the thing on the model, so i left the name.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,316 @@
+﻿var AsyncFileUploadManager = new function () {
```
ryuyu: I left it the way it was before.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
ryuyu: I think the if block is more clear.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
ryuyu: Moved the relocatin to a callback function so verify can have other behaviours if we want later.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
ryuyu: Sure we can change it to display a default error message. Don't think we should automatically retry. On GET only requests, retrying has no adverse effect, but for POST, every request should be an explicit action.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
ryuyu: knockout doesn't allow you to call applyBindings to an element more than once. Since we get a brand new model every time, and our model is not mutating, we have to clear what we had before and reapply.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
ryuyu: Removed.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
ryuyu: Looks like it tried to load jquery in an iframe when in a webkit based browser. I'm not sure why it was doing this, but i would prefer not to drop it until it's understood that we don't need it. In any case, i don't call it any more, so it's mostly left so the work to do it doesn't have to be done again later if it turns out to be necessary.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -170,16 +170,40 @@ public virtual ActionResult UploadPackageProgress()
...
+                        packageMetadata = PackageMetadata.FromNuspecReader(
+                            package.GetNuspecReader());
+                    }
+                    catch (Exception ex)
+                    {
+                        TempData["Message"] = ex.GetUserSafeMessage();
+                        return View(model);
+                    }
+
+                    model.IsUploadInProgress = true;
```
joelverhagen: >IsUploadInProgress [](start = 26, length = 18)

The upload is already done, right?

joelverhagen: Perhaps this mean the user has not yet "confirmed"/"submitted" the upload? Name is a bit unclear.

---
In reply to: [124669087](https://github.com/NuGet/NuGetGallery/pull/4284#discussion_r124669087) [](ancestors = 124669087)

joelverhagen:
>         public virtual async Task<ActionResult> UploadPackage(HttpPostedFileBase uploadFile)

Return type should be `JsonResult` right?

---
Refers to: src/NuGetGallery/Controllers/PackagesController.cs:213 in 251d623. [](commit_id = 251d623c500ecb16ea99579bfe5a07fa53f0b8db, deletion_comment = False)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -170,16 +170,40 @@ public virtual ActionResult UploadPackageProgress()
         public async virtual Task<ActionResult> UploadPackage()
```
joelverhagen: >UploadPackage [](start = 48, length = 13)

Does this need anti forgery attribute?

joelverhagen:
>                     return new HttpStatusCodeResult(409, "Cannot upload file because an upload is already in progress.");

This should be JSON right?

---
Refers to: src/NuGetGallery/Controllers/PackagesController.cs:221 in 251d623. [](commit_id = 251d623c500ecb16ea99579bfe5a07fa53f0b8db, deletion_comment = False)

src/NuGetGallery/Controllers/PackagesController.cs
ryuyu: Yes that would be more correct. This is more a reference to there being a package submission in progress. We can rename it if you think it makes more sense.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1264,20 +1317,30 @@ private string GetSecurePushPolicyDescriptions()
...
                 }
 
                 var nugetPackage = await SafeCreatePackage(currentUser, uploadFile);
                 if (nugetPackage == null)
                 {
+
+                    Response.StatusCode = 400;
                     // Send the user back
-                    return new RedirectResult(Url.UploadPackage());
+                    return Json(new string [] { "There was an error. Please try again" });
```
joelverhagen: >There [](start = 49, length = 5)

Let's call out here that we don't know much, e.g. "There was an unexpected error. Please try uploading the package again." 

Also, these should be in Strings.resx

src/NuGetGallery/Scripts/gallery/async-file-upload.js
joelverhagen: This is really a type, not just a function.

---
In reply to: [124366026](https://github.com/NuGet/NuGetGallery/pull/4284#discussion_r124366026) [](ancestors = 124366026)

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,294 @@
...
+    var _actionUrl;
+    var _cancelUrl;
+    var _submitVerifyUrl;
+    var _isWebkitBrowser = false; // $.browser.webkit is not longer supported on jQuery
```
joelverhagen: >_isWebkitBrowser [](start = 8, length = 16)

Unused?

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,294 @@
...
+    var _iframeId = '__fileUploadFrame';
+    var _formId;
+    var _pollingInterval = 250;
```
joelverhagen: >_pollingInterval [](start = 8, length = 16)

Name this with "ms" suffix or leave comment about what unit this is (ms presumably)

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -0,0 +1,294 @@
...
+        $("#upload-progress-bar-container").show();
+        setTimeout(getProgress, 100);
+    }
+
+    function endProgressBar() {
+        $("#upload-progress-bar-container").hide();
+        _isUploadInProgress = false;
+    }
+
+    function getProgress() {
```
joelverhagen: >getProgress [](start = 13, length = 11)

Only one of these should be in flight at a time.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
joelverhagen: Let's remove if it's not called (we can always recover it from history if we later find it's necessary). If you think we may need it, let's do the investigation and use it if necessary. 

---
In reply to: [124424011](https://github.com/NuGet/NuGetGallery/pull/4284#discussion_r124424011) [](ancestors = 124424011)

src/NuGetGallery/Views/Packages/UploadPackage.cshtml
```diff
@@ -2,48 +2,99 @@
 @{
     ViewBag.Title = "Upload Package";
     ViewBag.Tab = "Upload";
+    Layout = "~/Views/Shared/Gallery/Layout.cshtml";
 }
 
-<h1 class="page-heading">Upload Your Package</h1>
-
-@ViewHelpers.UploadSequence(1)
+<div class="page-upload container main-container">
```
joelverhagen: >div [](start = 1, length = 3)

role="main"
<section>

src/NuGetGallery/Views/Packages/UploadPackage.cshtml
```diff
@@ -2,48 +2,99 @@
...
 
 @section BottomScripts
 {
     @* Right now this is the only page that uses this script. If we increase our usage of it, we should put it in our bundles *@
-    <script type="text/javascript" src="@Url.Content("~/Scripts/async-file-upload.js")"></script>
+    <script type="text/javascript" src="@Url.Content("~/Scripts/gallery/async-file-upload.js")"></script>
     <script type="text/javascript">
+        var InProgressPackage = @Html.Raw(Json.Encode(Model.InProgressUpload));
+        $("#upload-progress-bar-container").hide();
+        $("#validation-failure-container").hide();
```
joelverhagen: Use CSS "hidden" class for this instead of using jQuery

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
```diff
@@ -0,0 +1,191 @@
+﻿<script type="text/html" id="validation-errors">
+    <div data-bind="foreach:$data" class="validation-error-message-list">
```
joelverhagen: >div [](start = 5, length = 3)

Can we do a similar thing here that we did for the slider, i.e. share more between server and client side rendering?

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
```diff
@@ -0,0 +1,191 @@
...
+                    </li>
+                </ul>
+            </li>
+        </ul>
+    </div>
+</script>
+
+<script type="text/html" id="verify-metadata-template">
+    <div class="collapse in" id="verify-package-form">
+        <div data-bind="if: $data" class="col-sm-12">
```
joelverhagen: >col-sm-12 [](start = 42, length = 9)

These col-sm-12 should not be necessary. This whole form should be in a col-sm-10.

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
```diff
@@ -0,0 +1,191 @@
...
+                    <input data-bind="value: $data.DevelopmentDependency" type="hidden" name="DevelopmentDependency" class="form-control" />
+                </div>
+
+                <div class="verify-package-field form-group readonly">
+                    <div class="verify-package-field-heading">Dependencies</div>
+                    <!-- ko if: $data.Dependencies -->
+                    <div data-bind="template: {name: 'display-dependencysets', data: { DependencySets: Dependencies.DependencySets, OnlyHasAllFrameworks: Dependencies.OnlyHasAllFrameworks }}"></div>
+                    <!-- /ko -->
+                    <!-- ko ifnot: $data.Dependencies -->
+                    <p><i data-bind="text:'(none specified)'"></i></p>
```
joelverhagen: >bind [](start = 31, length = 4)

don't data-bind constants

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
```diff
@@ -0,0 +1,191 @@
...
+                    <!-- /ko -->
+                </div>
+
+                <!--BEGIN EDITABLE FIELDS-->
+                <div class="verify-package-field form-group editable">
+                    <div class="verify-package-field-heading">Title</div>
+                    <!-- ko if: $data.Edit.hasOwnProperty('Title') -->
+                    <input data-bind="value: Edit.Title" type="text" class="form-control" name="Edit.Title" id="title-field" />
+                    <!-- /ko -->
+                    <!-- ko if: $data.Edit.hasOwnProperty('VersionTitle') -->
```
joelverhagen: >VersionTitle [](start = 59, length = 12)

Can we align these names potentially? This seems weird

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
```diff
@@ -0,0 +1,191 @@
...
+                    <div class="verify-package-field-heading">Copyright</div>
+                    <input data-bind="value: Edit.Copyright" type="text" class="form-control" name="Edit.Copyright" />
+                </div>
+
+                <div class="verify-package-field form-group editable">
+                    <div class="verify-package-field-heading">Tags (space separated - e.g. 'ASP.NET Template MVC')</div>
+                    <input data-bind="value: Edit.Tags" type="text" class="form-control" name="Edit.Tags" />
+                </div>
+
+                <!-- ko if: $data.LicenseUrl -->
```
joelverhagen: >$data [](start = 28, length = 5)

Isn't `$data.` unnecessary? Can't you just use `LicenseUrl`?

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
```diff
@@ -0,0 +1,191 @@
...
+                </div>
+            </form>
+        </div>
+
+        <div class="verify-package-field form-group btn-toolbar row">
+            <div class="col-xs-6">
+                <input type="button" class="btn btn-primary btn-block" value="Submit" id="verify-submit-button" />
+            </div>
+            <div class="col-xs-6">
+                <input type="button" class="btn btn-block" value="Cancel" id="verify-cancel-button" />
```
joelverhagen: nit: btn-default

joelverhagen:
:clock1:

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
ryuyu: Done

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
```diff
@@ -24,6 +42,7 @@ public class VerifyPackageRequest
         public bool Listed { get; set; }
         public EditPackageVersionRequest Edit { get; set; }
         public NuGetVersion MinClientVersion { get; set; }
```
joelverhagen: Do we really need both of these? Ideally we just have one, likely the `string` one unless you want a `JsonConverter` that know how to properly serialize a `NuGetVersion`.

src/Bootstrap/less/theme/base.less
joelverhagen: Can we use a gray variable here?

src/NuGetGallery/Views/Packages/UploadPackage.cshtml
```diff
@@ -2,48 +2,101 @@
...
+</div>
 
 @section BottomScripts
 {
     @* Right now this is the only page that uses this script. If we increase our usage of it, we should put it in our bundles *@
-    <script type="text/javascript" src="@Url.Content("~/Scripts/async-file-upload.js")"></script>
+    <script type="text/javascript" src="@Url.Content("~/Scripts/gallery/async-file-upload.js")"></script>
     <script type="text/javascript">
+        var InProgressPackage = @Html.Raw(Json.Encode(Model.InProgressUpload));
+        //$("#upload-progress-bar-container").hide();
```
joelverhagen: >("# [](start = 11, length = 3)

Delete commented code

src/NuGetGallery/Scripts/gallery/async-file-upload.js
ryuyu: Currently, but left in place because i don't fully understand what the below related function creating the iframe was for.

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
ryuyu: Has to do with knockout behaviour when LicenseUrl is null.

src/NuGetGallery/Views/Packages/UploadPackage.cshtml
ryuyu: removed.

src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
ryuyu: I'm not 100% sure what we would share here. We can revisit this when we have a more concrete idea of what we want to do.

src/NuGetGallery/Controllers/PackagesController.cs
ryuyu: This is the GET method that returns the initially rendered page. Doesn't need antiforgery.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
ryuyu: Knockout 3 doesn't like calling ko.applyBindings multiple times on the same element, so we kill the old one and make a new one. The old way worked in Knockout 2.x.

PR #4291, opened 2017-06-28
[Redesign] Email Confirmation

scottbommarito:
Did some minor refactoring to the controller code because it expected the user to specify all of the `ConfirmationViewModel`'s fields manually instead of automatically populating them from a `User` ([this line, case in point](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Views/Users/ConfirmationRequired.cshtml#L4)).

The following pictures aren't all possible states, but they should be a good sampler of the possibilities.

image

image

image

@microsoftsam

### DIFF
--- a/src/NuGetGallery/Views/Users/Confirm.cshtml
+++ b/src/NuGetGallery/Views/Users/Confirm.cshtml
@@ -1,61 +1,78 @@
 ﻿@model ConfirmationViewModel
-
-<h1 class="page-heading">Confirm Registration</h1>
-
-@if (Model.SuccessfulConfirmation)
-{
-    <p>
-        Account confirmed! Your email address has been verified.
-    </p>
-    <p>
-        You may now upload packages and make your mark in
-        the community.
-    </p>
+@{
+    ViewBag.Title = "Confirm Your Account";
+    Layout = "~/Views/Shared/Gallery/Layout.cshtml";
 }
-else
-{
-    if (Model.WrongUsername)
-    {
-        <p class="message error">
-            You cannot confirm a different user's email address.
-        </p>
-        <p>
-            Please check that you used the link from the right confirmation email, and are logged in as the right user.
-        </p>
-        <p>
-            <a href="@Url.ConfirmationRequired()" >Resend confirmation email.</a>
-        </p>
-    }
-    else if (Model.DuplicateEmailAddress)
-    {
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Views/Users/Confirm.cshtml
```diff
@@ -1,61 +1,79 @@
...
+                                You cannot confirm a different user's email address!
+                            </p>
+                            <p>
+                                Please check that you used the link from the right confirmation email, and are logged in as the right user.
+                            </p>
+                        }
+                        else if (Model.DuplicateEmailAddress)
+                        {
+                            <p>
+                                Your email address is confirmed as belonging to another user account (likely another account you own).
```
loic-sharma: What do you think of simplifying this message to just `Your email already belongs to another account.`?

scottbommarito: Nice catch!

`Your email address has already been confirmed by another account` I think is a little better. If somebody has two accounts with the same unconfirmed email and then confirms one, your message is a little ambiguous.

src/NuGetGallery/Views/Users/Confirm.cshtml
```diff
@@ -1,61 +1,79 @@
...
-    else
-    {
-        <p class="message error">
-            We could not confirm your email address. The token was invalid or expired.
-        </p>
-        <p>
-             Please check that you used the link from the confirmation email we sent.
-        </p>
-    }
+<div class="container main-container">
```
joelverhagen: `<section>`
`role="main"`

src/NuGetGallery/Views/Users/ConfirmationRequired.cshtml
```diff
@@ -1,55 +1,59 @@
...
-                <legend>Verify Email Address</legend>
-                @Html.AntiForgeryToken()
-
-                <p>
-                    Click the button to send confirmation email.
-                </p>
-                <p>
-                    <input type="submit" value="Send Confirmation Email" title="Send Confirmation Email" style="width: auto; margin: 5px;"/>
-                    to @(Model.UnconfirmedEmailAddress) @Html.ActionLink("(change)", "Account", "Users")
+<div class="container main-container">
```
joelverhagen: same

PR #4292, opened 2017-06-28
Add X-NuGet-Warning header to response when pushing SemVer2 package

xavierdecoster:
Related to https://github.com/nuget/nugetgallery/issues/4215


xavierdecoster:
@joelverhagen added

PR #4297, opened 2017-06-28
[Redesign] Unlist/Delete Package Page

loic-sharma:
delete

@microsoftsam

### DIFF
--- a/src/Bootstrap/less/theme/page-delete-package.less
+++ b/src/Bootstrap/less/theme/page-delete-package.less
@@ -1,0 +1,9 @@
+.page-delete-package {
+  h1 {
+    margin-bottom: 0px;
+  }
+
+  #show-delete-package {
+    color: @brand-danger;
+  }
+}
--- a/src/Bootstrap/less/theme/page-delete-package.less
+++ b/src/Bootstrap/less/theme/page-delete-package.less
@@ -1,6 +1,6 @@
 .page-delete-package {
   h1 {
-    margin-bottom: 0px;
+    margin-bottom: 0;
   }
 
   #show-delete-package {
--- a/src/NuGetGallery/RequestModels/DeletePackagesRequest.cs
+++ b/src/NuGetGallery/RequestModels/DeletePackagesRequest.cs
@@ -27,8 +27,10 @@ namespace NuGetGallery
         [Display(Name = "Signature")]
         public string Signature { get; set; }
 
+        [Display(Name = "Keep the package ID and version reserved")]
         public bool SoftDelete { get; set; }
 
+        [Display(Name = "Remove the package registration when all packages are deleted")]
         public bool DeleteEmptyPackageRegistration { get; set; }
 
         public IEnumerable<ReportPackageReason> ReasonChoices { get; set; }
--- a/src/NuGetGallery/Views/Packages/Delete.cshtml
+++ b/src/NuGetGallery/Views/Packages/Delete.cshtml
@@ -22,24 +22,23 @@
 
             <div class="panel-collapse collapse in" aria-expanded="true" id="select-version">
                 <div class="form-group">
-                    <label for="VersionSelector">Select Package Version</label>
-                    <select id="VersionSelector" class="form-control">
-                        @foreach (var package in Model.PackageVersions.Where(p => p.Deleted == false))
-                        {
-                            <option value="@Url.DeletePackage(package)"
-                                @if (Model.IsCurrent(package))
-                                {
-                                    @:selected="selected"
-                                }
-                            >
-                                @package.Version
-                                @if (package.LatestVersionSemVer2)
-                                {
-                                    @:(latest)
-                                }
-                            </option>
-                        }
-                    </select>
+                    <label for="input-select-version" id="package-version-label">Select Package Version</label>
+
+                    @Html.DropDownList("version-selection",
+                        new SelectList(
+                            Model.PackageVersions
+                                 .Where(p => p.Deleted == false)
+                                 .Select(p => new
+                                 {
+                                     text = p.Version + (p.LatestVersionSemVer2 ? " (latest)" : string.Empty),
+                                     url = Url.DeletePackage(p)
+                                 }), "url", "text", Url.DeletePackage(Model)),
+                                 new
+                                 {
+                                     @class = "form-control",
+                                     title = "Selection A Version...",
+                                     id = "input-select-version"
+                                 })
                 </div>
             </div>
 
@@ -193,11 +192,11 @@
                 var referrer = document.referrer.toLowerCase();
 
                 if (referrer.endsWith("/delete") && referrer.indexOf("@Model.Id".toLowerCase()) != -1) {
-                    $("#VersionSelector").focus();
+                    $("#input-select-version").focus();
                 }
             }
 
-            $('#VersionSelector').change(function () {
+            $('#input-select-version').change(function () {
                 window.location.href = this.value;
             });
 
--- a/src/NuGetGallery/Views/Packages/Delete.cshtml
+++ b/src/NuGetGallery/Views/Packages/Delete.cshtml
@@ -2,138 +2,205 @@
 @{
     ViewBag.Title = "Delete Package";
     ViewBag.Tab = "Packages";
+    Layout = "~/Views/Shared/Gallery/Layout.cshtml";
 }
 
-<h1 class="page-heading"><em>@Model.Title.Abbreviate(40) @Model.Version</em> Listing</h1>
-@if (!User.IsAdministrator())
-{
-    <p class="message">
-        Permanently deleting packages is not supported, but you can control how they are listed.
-    </p>
-}
-else
-{
-    <h2>Manage package listing</h2>
-}
-<p>
-    Unlisting a package hides the package from search results and all NuGet commands, but packages
-    are still available for download. For example, they can still be downloaded as dependencies to
-    other packages.
-</p>
+<section class="container main-container page-delete-package">
+    <div class="row">
+        <div class="col-md-10 col-md-offset-1">
+            <h1 class="text-center">Manage Package Listing</h1>
+            <div class="text-center ms-font-xxl">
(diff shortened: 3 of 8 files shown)


src/NuGetGallery/RequestModels/DeletePackagesRequest.cs
```diff
@@ -27,8 +27,10 @@ public DeletePackagesRequest()
         [Display(Name = "Signature")]
         public string Signature { get; set; }
 
+        [Display(Name = "Keep the package id and version reserved")]
```
joelverhagen: nit: uppercase ID

src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -2,138 +2,201 @@
...
-                    @Html.ValidationMessageFor(m => m.DeletePackagesRequest.Reason, null, new {@class = "block-validation-error"})
-                </div>
+                @if (!User.IsAdministrator())
+                {
+                    @ViewHelpers.AlertInfo(
+                        @<text>
+                            Note - Permanently deleting packages is not support, but you can control how they are listed.
+                        </text>
+                    )
+                    <h3>Why can&#8217;t I delete my package?</h3>
```
joelverhagen: Can't we use a normal apostrophe?

src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -2,138 +2,201 @@
...
+                        the package.
+                    </p>
+                }
 
-                <div class="form-field">
-                    @Html.LabelFor(m => m.DeletePackagesRequest.Signature)
-                    @Html.EditorFor(m => m.DeletePackagesRequest.Signature)
-                    @Html.ValidationMessageFor(m => m.DeletePackagesRequest.Signature)
-                </div>
+                @using (Html.BeginForm("UpdateListed", "Packages", FormMethod.Post))
```
joelverhagen: nit: Post is default right?

src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -2,138 +2,201 @@
...
+            "ChevronRight",
+            "Delete Package Version",
+            "ChevronDown",
+            "Delete Package Version");
+
+    </script>
     <script>
-        $(function() {
+        $(function () {
+            if (document.referrer.endsWith("/Delete") && document.referrer.indexOf("@Model.Id") != -1) {
```
joelverhagen: can this be null?

joelverhagen: er, `document.referrer`

loic-sharma: From some quick research, it doesn't seem like it. That being said, I think I'll make this resilient to null/undefined just to be on the safe side.

src/Bootstrap/less/theme/base.less
```diff
@@ -166,9 +166,17 @@ body {
...
+  .ms-Icon {
+    font-size: 0.6em;
+    position: relative;
+    top: -2px;
+  }
+
+  a, a:hover, a:focus, a:active {
+    color: @text-color;
+    text-decoration: none;
+  }
```
ryuyu: I think merging latest feature-redesign will pull in this change.

src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -2,138 +2,201 @@
...
+            </h2>
 
-        <input type="submit" value="Save" title="Save Changes" />
-        <a class="cancel" href="@Url.Action("DisplayPackage")" title="Cancel Changes and go back to package page.">Cancel</a>
-    </fieldset>
-}
+            <div class="panel-collapse collapse in" aria-expanded="true" id="select-version">
+                <div class="form-group">
+                    <label for="VersionSelector">Select Package Version</label>
+                    <select id="VersionSelector" class="form-control">
```
ryuyu: Do we need any special tags here for accessibility?

src/NuGetGallery/Views/Packages/Delete.cshtml
ryuyu: I think so. https://msdn.microsoft.com/en-us/library/system.web.mvc.html.formextensions.beginform(v=vs.118).aspx#M:System.Web.Mvc.Html.FormExtensions.BeginForm%28System.Web.Mvc.HtmlHelper,System.String,System.String%29

src/NuGetGallery/Views/Packages/Delete.cshtml
scottbommarito: You and @ryuyu have very very different code for the selectors. I was under the impression that they were essentially the same, functionally.

[He seems to have used `@Html.DropDownList`](https://github.com/NuGet/NuGetGallery/pull/4307/files#diff-034e8bf00ce4865a1836ba7cd4fa676bR33) and you constructed the drop-down yourself. Can we be more consistent with this? I'd rather see `@Html.DropDownList` IMO.

src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -2,138 +2,201 @@
...
-                <div id="form-field-reason" class="form-field">
-                    @Html.LabelFor(m => m.DeletePackagesRequest.Reason)
-                    @Html.EnumDropDownListFor(m => m.DeletePackagesRequest.Reason, Model.DeletePackagesRequest.ReasonChoices, "<Choose a Reason>")
-                    @Html.ValidationMessageFor(m => m.DeletePackagesRequest.Reason, null, new {@class = "block-validation-error"})
-                </div>
+                @if (!User.IsAdministrator())
+                {
+                    @ViewHelpers.AlertInfo(
+                        @<text>
+                            Note - Permanently deleting packages is not support, but you can control how they are listed.
```
scottbommarito: `is not support` -> `is not supported`

src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -2,138 +2,201 @@
...
+                    )
+                    <h3>Why can&#8217;t I delete my package?</h3>
+                    <p>
+                        Our policy is to only permanently delete NuGet packages that really need it, such as
+                        packages that contain passwords, malicious/harmful code, etc. This policy is very similar
+                        to the policies employed by other package managers such as
+                        <a href="http://help.rubygems.org/kb/gemcutter/removing-a-published-rubygem">Ruby Gems</a>.
+                    </p>
+                    <p>
+                        Unlisting the package will prevent users from finding it by searching the gallery and it will prevent the dependency resolver from discovering the package.
```
scottbommarito: what is `the dependency resolver`?

```
...and it will prevent it from being discovered while resolving dependencies of other packages.
```

src/NuGetGallery/Views/Packages/Delete.cshtml
scottbommarito: [Can we align this text with our docs?](https://docs.microsoft.com/en-us/nuget/policies/deleting-packages)

We could potentially just replace it with a link to the docs and just leave in the link to make a support request.

src/NuGetGallery/Views/Packages/Delete.cshtml
scottbommarito: Yup, `FormMethod.Post` is unnecessary.

src/NuGetGallery/Views/Packages/Delete.cshtml
loic-sharma: Good catch, thanks

src/NuGetGallery/Views/Packages/Delete.cshtml
loic-sharma: I think that's just another way of saying package manager. I'll change it since it'll be a little clearer that way.

src/NuGetGallery/Views/Packages/Delete.cshtml
loic-sharma: Great suggestion, will do!

src/NuGetGallery/Views/Packages/Delete.cshtml
loic-sharma: The `for` attribute on the label should be enough. I'll switch to using Razor helpers to ensure we get everything we need though.

loic-sharma:
:bell: I did a rather large rebase and addressed all of the feedback. I'll merge this in at the end of the day if there are no more issues! :)

src/Bootstrap/less/theme/page-delete-package.less
```diff
@@ -0,0 +1,9 @@
+.page-delete-package {
+  h1 {
+    margin-bottom: 0px;
```
joelverhagen: nit: `0` instead of `0px` (it's idiomatic to leave zero unitless)

src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -2,138 +2,203 @@
...
     </div>
-}
+</section>
 
 @section BottomScripts {
+    <script type="text/javascript">
+        window.nuget.configureExpander(
+            "select-version",
+            "ChevronRight",
+            "Select Version",
```
joelverhagen: Use `configureExpanderHeading` for these three

src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -2,138 +2,203 @@
...
+            "Delete Package Version",
+            "ChevronDown",
+            "Delete Package Version");
+
+    </script>
     <script>
-        $(function() {
+        $(function () {
+            if (!!document.referrer) {
+                var referrer = document.referrer.toLowerCase();
```
joelverhagen: Let's track an issue for the "right" fix here, which means the dropdown does not navigate and instead loads the data using AJAX.

loic-sharma: Good idea, opened #4319

Issue #4185, opened 2017-06-14
[Redesign] Empty Tag Backgrounds on Package Search


loic-sharma:
[Packages with empty tags](https://preview-dev.nugettest.org/packages?q=WebGrease) seem to show an empty tag background.

<img width="769" alt="screen shot 2017-06-14 at 3 40 16 pm" src="https://user-images.githubusercontent.com/737941/27157681-c98a0e74-5117-11e7-9489-d69383f91b30.png">

Issue #4184, opened 2017-06-14
[Redesign] Profile page aside section with small window is at bottom


scottbommarito:
![image](https://user-images.githubusercontent.com/18014088/27157695-d8df6eb4-5117-11e7-8a3f-a757b0ab25c1.png)

It would look a lot better at the top.

PR #4303, opened 2017-06-29
[Redesign] Fix Several Package Listing Issues

### LINKED ISSUE #4173: [Redesign] List Packages with small window looks odd
image

I think it would look much better if:
- The margin between the icon and the id was increased.
- The margin between the gray text rows (download statistics, last updated, and latest version) was increased.
- Each portion of the gray text (download statistics, last updated, and latest version) was on its own line.
- The id and authors were always on separate lines.

Additionally, If the package id is longer than the screen size, without spaces, it gets truncated.

image

This applies to the profile page as well.

loic-sharma:
* Made Profile Picture appear before packages on mobile profile page
* Made package lists' statistics and management links flat on mobile
* Hid package icons on very narrow screens
* Fixed empty tag issue

## Wide Package Listing View

list-wide

## Narrow Package Listing View

list-narrow

## Profile Page

narrow-profile

Fixed #4173, #4185, and #4184

### DIFF
--- a/src/NuGetGallery/Views/Users/Profiles.cshtml
+++ b/src/NuGetGallery/Views/Users/Profiles.cshtml
@@ -7,7 +7,21 @@
 
 <div class="container main-container page-profile">
     <section role="main" class="row">
-        <article class="col-md-8">
+        <aside class="col-md-4 col-md-push-8 profile-details">
+            @ViewHelpers.GravatarImage(Model.EmailAddress, Model.Username, 332)
+
+            <div class="statistics">
+                <div class="statistic">
+                    <div class="value">@Model.AllPackages.Count.ToNuGetNumberString()</div>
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Views/Users/Profiles.cshtml
```diff
@@ -7,7 +7,21 @@
 
 <div class="container main-container page-profile">
     <section role="main" class="row">
-        <article class="col-md-8">
+        <aside class="col-md-4 col-md-push-8 profile-details">
```
joelverhagen: 👍 right tool for the job

ryuyu:
Are we going to fix the package title link color in a future PR?

loic-sharma:
Good call Peter, fixed the link color.

PR #4309, opened 2017-06-30
[Redesign] Fixed link color on package search and profile page

loic-sharma:
The global styling changes for h2 links (intended only for collapsible sections) broke the styling for package search and profile page by turning their links black. This change fixes that by removing the h2 around the link.

### DIFF
--- a/src/Bootstrap/less/theme/common-list-packages.less
+++ b/src/Bootstrap/less/theme/common-list-packages.less
@@ -8,8 +8,8 @@
     font-size: 15px;
 
     .package-header {
-      h2 {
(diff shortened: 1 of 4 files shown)


src/Bootstrap/less/theme/common-list-packages.less
```diff
@@ -8,8 +8,8 @@
     font-size: 15px;
 
     .package-header {
-      h2 {
-        display: inline;
+      .package-title {
+        font-size: 34px;
```
joelverhagen: `@font-size-h2`

PR #4310, opened 2017-06-30
[Redesign]Some Test Fixes

ryuyu:
This is a quick PR to fix up some of the tests we have caused to fail with redesign work so far.
 * Removes some dead tests
 * Minimal fixes to existing test to get them to pass.

@joelverhagen @scottbommarito @loic-sharma

### DIFF
--- a/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
+++ b/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
@@ -162,10 +162,10 @@ namespace NuGetGallery
                     uploadFileService: fakeUploadFileService);
                 controller.SetCurrentUser(TestUtility.FakeUser);
 
-                var result = await controller.CancelUpload() as RedirectToRouteResult;
+                var result = await controller.CancelUpload();
 
-                Assert.False(result.Permanent);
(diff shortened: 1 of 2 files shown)


tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -162,10 +162,10 @@ public class TheCancelVerifyPackageAction
                     uploadFileService: fakeUploadFileService);
                 controller.SetCurrentUser(TestUtility.FakeUser);
 
-                var result = await controller.CancelUpload() as RedirectToRouteResult;
+                var result = await controller.CancelUpload();
 
-                Assert.False(result.Permanent);
-                Assert.Equal("UploadPackage", result.RouteValues["Action"]);
+                Assert.True(result is JsonResult);
```
joelverhagen: `IsType`

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -162,10 +162,10 @@ public class TheCancelVerifyPackageAction
...
+                Assert.Equal(null, result.Data);
```
joelverhagen: `Assert.Null`

Issue #4314, opened 2017-07-03
[Redesign] Curated packages remove link jumps to top of page


joelverhagen:
![image](https://user-images.githubusercontent.com/94054/27800648-304cf4a8-5fcf-11e7-8ae6-0cf456fad89b.png)

I think this is because you made it a link with href `#` and did not prevent the default action (which for anchors is navigation).

The URL in the browser ends up looking like this:
https://localhost/curated-feeds/windows8-packages#

Issue #4191, opened 2017-06-14
[Redesign] Most Downloaded ... pages have no gutter


scottbommarito:
![image](https://user-images.githubusercontent.com/18014088/27158028-946d93e4-5119-11e7-8dee-6a271f694067.png)

ryuyu:
Probably caused by encompassing row.

Issue #4179, opened 2017-06-14
[Redesign][Page-About]No gutter around content


ryuyu:
Screenshot:
![image](https://user-images.githubusercontent.com/11051729/27157523-be412aa8-5116-11e7-87ab-c7b89b67c516.png)

Issue #4141, opened 2017-06-14
About the Gallery does not load images in IE 11


joelverhagen:
https://preview-dev.nugettest.org/policies/About

![image](https://user-images.githubusercontent.com/94054/27155919-b13d31dc-510f-11e7-9c5c-0aabfc8e8448.png)

scottbommarito:
Apparently this is because IE 11 doesn't support `Promise`s.

Given that IE (and other browsers that don't support `Promise`s) are such a small share of the market and the list of contributors is a nonessential functionality of the page, I think I will just add a check that if `Promise`s are not enabled, the script will not run.

scottbommarito:
Leaving this issue open because at some point in the future we should fix the script to work on IE 11, but the temporary fix (disabling the script when `Promise`s are not available) has been merged.

PR #4317, opened 2017-07-03
[Redesign] Minor bugfixes

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/4191
https://github.com/NuGet/NuGetGallery/issues/4179
https://github.com/NuGet/NuGetGallery/issues/4141
https://github.com/NuGet/NuGetGallery/issues/4135
https://github.com/NuGet/NuGetGallery/issues/4314


joelverhagen:
Let's keep #4141 open for now. We are officially supporting IE 11 so having a feature that just doesn't work seems right. Low priority for now since the page is still usable.

Issue #4308, opened 2017-06-30
Reflow package should recalculate IsLatest state


chenriksson:
see: https://github.com/NuGet/NuGetGallery/issues/4281

Earlier issue fixed the profile page to be resilient to multiple IsLatest versions. This issue is to allow Reflow Package to recalculate latest versions to correct the invalid state.

PR #4318, opened 2017-07-03
Update reflow to update IsLatest state

chenriksson:
See:
- https://github.com/NuGet/NuGetGallery/issues/4308
- https://github.com/NuGet/Engineering/issues/545


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -815,7 +815,7 @@ private void ValidatePackageTitle(PackageMetadata packageMetadata)
             }
         }
 
-        public async Task UpdateIsLatestAsync(PackageRegistration packageRegistration, bool commitChanges = true)
+        public virtual async Task UpdateIsLatestAsync(PackageRegistration packageRegistration, bool commitChanges = true)
```
loic-sharma: Is the `virtual` needed here?

joelverhagen: It's for the mock.

PR #4321, opened 2017-07-03
[Redesign] Use black for text color and @text-color in LESS

joelverhagen:
https://github.com/NuGet/NuGetGallery/issues/4289


src/Bootstrap/less/variables.less
```diff
@@ -29,7 +29,7 @@
 //** Background color for `<body>`.
 @body-bg:               #fff;
 //** Global text color on `<body>`.
-@text-color:            @gray-dark;
+@text-color:            #000;
```
loic-sharma: Nit: you could use `gray-base` here.

joelverhagen: I find the name kind of weird, but sure.

PR #4328, opened 2017-07-05
[Redesign] Add Dropdown Menu

loic-sharma:
# Desktop

## Closed Dropdown

dropdown-closed

## Opened Dropdown

dropdown-opened

# Mobile

## Closed Dropdown

mobile-closed

## Opened Dropdown

mobile-opened

### DIFF
--- a/src/Bootstrap/less/navbar.less
+++ b/src/Bootstrap/less/navbar.less
@@ -605,8 +605,12 @@
     @media (max-width: @grid-float-breakpoint-max) {
       // Dropdowns get custom display
       .open .dropdown-menu {
+        .row {
+          margin-right: 0;
+        }
         > .dropdown-header {
           border-color: @navbar-inverse-border;
+          color: @navbar-inverse-link-color;
         }
         .divider {
           background-color: @navbar-inverse-border;
--- a/src/NuGetGallery/UrlExtensions.cs
+++ b/src/NuGetGallery/UrlExtensions.cs
@@ -278,6 +278,11 @@ namespace NuGetGallery
             return url.Action(actionName: "Account", controllerName: "Users");
         }
 
+        public static string ManageMyApiKeys(this UrlHelper url)
+        {
+            return url.Action(actionName: "ApiKeys", controllerName: "Users");
+        }
+
         public static string ManageMyPackages(this UrlHelper url)
         {
             return url.Action(actionName: "Packages", controllerName: "Users");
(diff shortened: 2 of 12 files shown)


src/Bootstrap/less/navbar.less
```diff
@@ -605,8 +605,12 @@
     @media (max-width: @grid-float-breakpoint-max) {
       // Dropdowns get custom display
       .open .dropdown-menu {
+        .row {
+          margin-right: 0;
```
loic-sharma: This is a hack. Adding a div with the class `row` causes the menu to overflow, thereby adding a horizontal scrollbar to the bottom of the navigation bar when in mobile view port size. This removes the margin adding by the row so that the div no longer overflows.

loic-sharma:
**NOTE**: There is a known accessibility issue with the link that opens the dropdown: focusing this link will not have an outline. This is caused by line 25-27 of `dropdowns.less`, and apparently, this is necessary for mouse users.

For more information, see this [issue](https://github.com/twbs/bootstrap/issues/16103).

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -273,6 +273,21 @@ public static string DeletePackage(this UrlHelper url, IPackageVersionModel pack
                 });
         }
 
+        public static string AccountSettings(this UrlHelper url)
```
joelverhagen: Merge conflict here 😉

joelverhagen:
When using the dropdown it feels like it disappears too easily. For example, if I click any non-link space in the dropdown, e.g. the profile picture, the dropdown disappears. I think it should only disappear if someone clicks outside of the dropdown.

Related to this is a more minor issue. When you click one of the links in the dropdown, the dropdown immediately collapses. I feel lilke it should stay open until the page navigates away. This is important so if someone accidentally clicks the wrong dropdown open, they should be able to click a different one until the point in time that the page navigate completes (this would make it consistent with other anchor clicks).

loic-sharma:
:bell: Fixed merge conflicts and addressed feedback.

src/NuGetGallery/Views/Users/Profiles.cshtml
```diff
@@ -8,7 +8,7 @@
 <div class="container main-container page-profile">
     <section role="main" class="row">
         <aside class="col-md-4 col-md-push-8 profile-details">
-            @ViewHelpers.GravatarImage(Model.EmailAddress, Model.Username, 332)
+            @ViewHelpers.GravatarImage(Model.EmailAddress ?? Model.UnconfirmedEmailAddress, Model.Username, 332)
```
loic-sharma: This is a bug fix unrelated to the dropdown. Currently, viewing a user's profile whose email is unconfirmed will cause an exception to be thronw.

PR #4307, opened 2017-06-30
[Redesign]Edit Package Page

ryuyu:
Updating edit package page for new design.

Changes:
Submit button is disabled until an edit is made.
Version selector is now on page. Note that the current behaviour for version selection is a page navigation, with javascript hijacking referrer for keyboard nav.
Moved styles for metadata edit form to a common sheet.
Confirmation dialog on navigate without submitting changes changed to using browser native confirm dialog.
Update edit endpoints to return JSON.

Notes:
There is a block that creates reset functions for each input. This is work that has in progress design. It is put in place to ease adding this feature later.

Screenshot:
image

PR #4329, opened 2017-07-05
[Redesign]Upload page contrast fix

ryuyu:
Change from using placeholder attribute to value attribute.
This is ok since the input box is readonly and for feedback on the selected file.
Changing to value also picks up styles that make the contrast compliant.

@joelverhagen @scottbommarito @loic-sharma @microsoftsam


ryuyu:
This change was included in https://github.com/NuGet/NuGetGallery/pull/4307

PR #4338, opened 2017-07-06
Expose api/status with HTTP + fix functional tests

skofman1:
Fixes: https://github.com/NuGet/NuGetGallery/issues/4336

Noticed HttpToHttps functional tests don't run during deployment, so fixed it.

### DIFF
--- a/src/NuGetGallery/App_Start/OwinStartup.cs
+++ b/src/NuGetGallery/App_Start/OwinStartup.cs
@@ -106,7 +106,8 @@ namespace NuGetGallery
                 }
                 else
                 {
-                    app.UseForceSsl(config.Current.SSLPort, new[] { config.Current.ForceSslExclusion });
+                    var paths = config.Current.ForceSslExclusion.Split(';');
+                    app.UseForceSsl(config.Current.SSLPort, paths);
                 }
             }
 
--- a/src/NuGetGallery/App_Start/OwinStartup.cs
+++ b/src/NuGetGallery/App_Start/OwinStartup.cs
@@ -21,7 +21,6 @@ using NuGetGallery.Authentication;
 using NuGetGallery.Authentication.Providers;
 using NuGetGallery.Authentication.Providers.Cookie;
 using NuGetGallery.Configuration;
-using NuGetGallery.Helpers;
 using NuGetGallery.Infrastructure;
 using Owin;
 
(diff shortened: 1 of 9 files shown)


src/NuGetGallery/App_Start/OwinStartup.cs
```diff
@@ -106,7 +106,8 @@ public static void Configuration(IAppBuilder app)
                 }
                 else
                 {
-                    app.UseForceSsl(config.Current.SSLPort, new[] { config.Current.ForceSslExclusion });
+                    var paths = config.Current.ForceSslExclusion.Split(';');
```
joelverhagen: Today, this code does not add an empty string or whitespace exclusion. This is now possible with a config value of `;;;;` or something like that. We should filter out empty or whitespace before passing `paths` into `UseForceSsl`.

src/NuGetGallery/Configuration/AppConfiguration.cs
```diff
@@ -29,8 +29,8 @@ public class AppConfiguration : IAppConfiguration
         public int SSLPort { get; set; }
 
         /// <summary>
-        /// A string containing a path exluded from
-        /// forcing the HTTP to HTTPS redirection.
+        /// A string containing a path exluded from forcing the HTTP to HTTPS redirection. 
+        /// To provide multiple paths separate them with ;
         /// </summary>
         [DefaultValue("")]
         public string ForceSslExclusion { get; set; }
```
joelverhagen: Can this type be a `IEnumerable<string>`? It looks like we have some other complex types below like `MailAddress` enabled by a `TypeConverter`.

skofman1: I checked this option, and we currently don't support arrays as part of configuration, and using TypeConverter for this will introduce unnecessary complexity.

src/NuGetGallery/App_Start/OwinStartup.cs
skofman1: This is a configuration property that should only be used by people that know what they are doing (or even better, should never be changed). Given this, I don't think we should do excessive correctness checks on the input. If we start with ;;; we can also check the path has a valid format, etc. All are unnecessary.

src/NuGetGallery/Configuration/AppConfiguration.cs
joelverhagen: Any consumer of `IAppConfiguration` therefore has to perform the semicolon splitting and filtering out of useless whitespaces. The semicolon thing is essentially a workaround due to the fact that we can't express enumeration in our current config format. The runtime interaction with the config values should be as simple as possible.

(take the example of `MailAddress`, which is not just a simple string)

skofman1: Fixed

Issue #4190, opened 2017-06-14
[Redesign] Package Statistics / Package Version Statistics page looks odd with small window


scottbommarito:
![image](https://user-images.githubusercontent.com/18014088/27157988-576fa9b4-5119-11e7-8879-11b611a4a9cf.png)

- Package title is truncated.
- Margins between last updated timestamp and checkboxes should be increased.

![image](https://user-images.githubusercontent.com/18014088/27157928-0a7f747c-5119-11e7-8e84-f224cf13a4a2.png)

- Graph title is truncated (should wrap)
- "NuGet.Core-based Down..." is truncated (should wrap)
- Strange smudge at bottom of graph (the x-axis titles of the graph).

Issue #4170, opened 2017-06-14
[Redesign] Stats Label is Truncated on OS X Chrome


loic-sharma:
[This page](https://preview-dev.nugettest.org/stats/packages/BaseTestPackage?groupby=ClientName) clips the "N" in the label "NuGet.Core-based Dow.."

<img width="1079" alt="screen shot 2017-06-14 at 3 20 56 pm" src="https://user-images.githubusercontent.com/737941/27157147-1f627bf4-5115-11e7-9189-1ab6bdc5b69c.png">

ryuyu:
Addressed by #4340

PR #4340, opened 2017-07-06
[Redesign] Some Miscellaneous bug fixes

ryuyu:
Addresses the following issues:
 * [Redesign] Stats Label is Truncated on OS X Chrome #4170
 * [Redesign] Package Statistics / Package Version Statistics page looks odd with small window #4190

@joelverhagen @scottbommarito @loic-sharma

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/stats-perpackagestatsgraphs.js
+++ b/src/NuGetGallery/Scripts/gallery/stats-perpackagestatsgraphs.js
@@ -1,6 +1,6 @@
 ﻿
 var graphData;
-var axisLabelCharLimit = 20;
+var axisLabelCharLimit = 19;
 
 var packageDisplayGraphs = function (data) {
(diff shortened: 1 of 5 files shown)


src/NuGetGallery/Scripts/gallery/stats-perpackagestatsgraphs.js
```diff
@@ -1,6 +1,6 @@
 ﻿
 var graphData;
-var axisLabelCharLimit = 20;
+var axisLabelCharLimit = 19;
```
joelverhagen: Where does this number come from? Probably trial and error, right? If so, comment as such. 19 is a weird constant 😄

ryuyu: I'll add a comment that it was from trial and error.

PR #4342, opened 2017-07-06
[Redesign] Ensure package version is provided in Unlist/Delete form

### LINKED ISSUE #4323: Choosing package list view trash icon leads to failure path
## To Recreate

* Go to the Manage My Packages page. 
* Click on the Trash Icon.
* Attempt to unlist package (unselect check box and click save)

## Observered behaviour

* Oops error page

## Expected behaviour

* For my part I believed that clicking this trash icon would unlist the whole package (not just versions of the package). In fact, I think it tries to unlist just the current package.
* EIther way, it should not error

## Additional Information
* Sometimes you want to remove a whole package i.e. all versions. There is no easy way to do this, and there ought to be.

loic-sharma:
Mirror of #4341, fixes #4323 on redesign.

### DIFF
--- a/src/NuGetGallery/Views/Packages/Delete.cshtml
+++ b/src/NuGetGallery/Views/Packages/Delete.cshtml
@@ -80,6 +80,7 @@
                 @using (Html.BeginForm("UpdateListed", "Packages"))
                 {
                     @Html.AntiForgeryToken()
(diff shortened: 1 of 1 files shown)


joelverhagen:
Nit ideally this would be addressed using a merge commit, but it's not a big deal.

PR #4350, opened 2017-07-07
[Redesign] Add Admin link to dropdown

loic-sharma:
admindropdown

### DIFF
--- a/src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
+++ b/src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
@@ -86,6 +86,11 @@
                                             <span class="dropdown-email">@CurrentUser.EmailAddress.Abbreviate(35)</span>
                                         </div>
                                     </li>
+                                    @if (Request.IsAuthenticated && User.IsInRole(Constants.AdminRoleName))
+                                    {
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -86,6 +86,11 @@
                                             <span class="dropdown-email">@CurrentUser.EmailAddress.Abbreviate(35)</span>
                                         </div>
                                     </li>
+                                    @if (Request.IsAuthenticated && User.IsInRole(Constants.AdminRoleName))
```
scottbommarito: Are we not removing the admin tab?

https://github.com/NuGet/NuGetGallery/blob/feature-redesign/src/NuGetGallery/Views/Shared/Gallery/Header.cshtml#L54

joelverhagen: This could be seen two ways...

Consider the "Upload" link.

loic-sharma: Yes, the idea is to be able to access the admin link from two different places.

PR #4351, opened 2017-07-07
[Redesign] Gallery changes for Status page redesign

scottbommarito:
In order to support sharing assets with the status page, I needed to make some modifications to `Layout.cshtml`.

The general idea is that
- all links should use a url helper that can be defined by the status page to link to www.nuget.org instead of an MVC routing call.
- add a flag to `Layout.cshtml` for behavior that will never occur on the status page (see `ViewBag.ShowUserInNavBar`)

This also includes the `CSS` for the status page, which is specified using `LESS`. I have filed an issue to split this into its own file and remove it from the gallery bootstrap theme. (https://github.com/NuGet/NuGetGallery/issues/4352)


joelverhagen:
Please merge latest on `feature-redesign`.

scottbommarito:
@joelverhagen Finished merging!

src/Bootstrap/less/theme/page-status.less
```diff
@@ -0,0 +1,119 @@
+.page-status {
+    @status-general-margin-bottom: 24px;
+    
+    .status-general {
+        padding: 13px;
+        margin-bottom: @status-general-margin-bottom;
+        
+        p {
+            margin: 0px;
```
joelverhagen: nit unitless zero

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -56,48 +56,51 @@
...
-                                    <li role="presentation"><a href="@Url.UploadPackage()" role="menuitem">Upload Package</a></li>
-                                    <li class="divider"></li>
-                                    <li role="presentation"><a href="@Url.LogOff()" role="menuitem">Sign Out</a></li>
-                                </ul>
-                            </li>
-                        }
-                    </ul>
+                    @if (ViewBag.ShowUserInNavBar != false)
+                    {
+                        <ul class="nav navbar-nav navbar-right navbar-seperated" role="menu">
```
scottbommarito: no changes to the insides of this--I only added the `if (ViewBag.ShowUserInNavBar != false)`

src/Bootstrap/less/theme/page-status.less
```diff
@@ -0,0 +1,119 @@
+.page-status {
```
joelverhagen: Please file an issue to track splitting this out to its own css file (so that gallery doesn't have to ship status styles and vice verse).

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
joelverhagen: Add `w=1` query parameter to the PR URL.

src/Bootstrap/less/theme/page-status.less
scottbommarito: https://github.com/NuGet/NuGetGallery/issues/4352

PR #4365, opened 2017-07-10
[Redesign] Address PM pass redesign issues (scottbom)

scottbommarito:
1. **Most Downloaded Packages/Versions:** font is differently sized than other similar pages.
image
2. **Static Content:** moved from `src/NuGetGallery/App_Data` to the static content folders (`content/*`) so that they can be deployed
3. **Manage Packages:** add spacing between package controls and tags
image


content/DEV/Privacy-Policy.md
```diff
@@ -1,60 +1,59 @@
 # NuGet Website Privacy Statement
```
joelverhagen: Are these safe to deploy to the old site? No issues?

joelverhagen: Spoke offline. This will be deployed when we go live.

joelverhagen:
:shipit:

PR #4366, opened 2017-07-10
[Redesign]Enable drag and drop file upload.

ryuyu:
When we went from a regular file input to a masked file input, we lost the ability to drag and drop files to select a package.
This PR re-enables this feature through javascript.

@joelverhagen @jonwchu @scottbommarito @loic-sharma

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/async-file-upload.js
+++ b/src/NuGetGallery/Scripts/gallery/async-file-upload.js
@@ -6,7 +6,8 @@ var AsyncFileUploadManager = new function () {
     var _submitVerifyUrl;
     var _isWebkitBrowser = false; // $.browser.webkit is not longer supported on jQuery
     var _iframeId = '__fileUploadFrame';
-    var _formId;
+    var _uploadFormId;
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -54,7 +100,7 @@ var AsyncFileUploadManager = new function () {
             url: _actionUrl,
             type: 'POST',
 
-            data: new FormData($('#' + _formId)[0]),
+            data: _uploadFormData,
```
joelverhagen: What if the values of the form fields change after `prepareUploadFormData` is called? Will `_uploadFormData` contain stale data?

ryuyu: This is entirely possible, and yes it would contain stale data. But prepareFormData, which updates this, is called on every user interactive path that calls this.

joelverhagen: Sounds good.

PR #4394, opened 2017-07-13
[Prefix] DB changes for Prefix reservation

shishirx34:
Adding tables and relevant columns.


src/NuGetGallery.Core/Entities/ReservedPrefix.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class ReservedPrefix : IEntity
+    {
+        public ReservedPrefix() : this(null, false)
+        {
+        }
+
+        public ReservedPrefix(string pattern, bool isPublicNamespace)
+        {
+            Pattern = pattern;
+            IsPublicNamespace = isPublicNamespace;
```
joelverhagen: This means that the namespace is opened up by the owner to allow everyone to push?

src/NuGetGallery.Core/Entities/ReservedPrefix.cs
```diff
@@ -0,0 +1,35 @@
...
+            PackageRegistrations = new HashSet<PackageRegistration>();
+            ReservedPrefixOwners = new HashSet<User>();
+        }
+
+        [StringLength(CoreConstants.MaxPackageIdLength)]
+        [Required]
+        public string Pattern { get; set; }
```
joelverhagen: We talked about an exact match vs. prefix match. Does this somehow cover both?

src/NuGetGallery.Core/Entities/PackageRegistration.cs
```diff
@@ -13,15 +13,21 @@ public PackageRegistration()
...
         [Required]
         public string Id { get; set; }
 
         public int DownloadCount { get; set; }
+
+        public bool Verified { get; set; }
+
         public virtual ICollection<User> Owners { get; set; }
         public virtual ICollection<Package> Packages { get; set; }
+        public virtual ICollection<ReservedPrefix> ReservedPrefixes { get; set; }
```
skofman1: in what scenario will we need to map PackageRegistration to ReserverPrefixes?

src/NuGetGallery.Core/Entities/ReservedPrefix.cs
shishirx34: Yes

src/NuGetGallery.Core/Entities/ReservedPrefix.cs
shishirx34: Had offline talk with Joel. Changing this to "Value" and adding a new column "IsExactMatch" for cases of "jquery" vs "jquery*" patterns

src/NuGetGallery.Core/Entities/PackageRegistration.cs
shishirx34: one of the scenario, lets say we want to delete a reserved prefix and it matches an ID, we want to find which other prefixes does this ID match to determine if the "Verified" flag needs to be removed or not? I think it will be useful in the future as well to show all the prefixes an ID matches, in account page may be.

cristinamanum: For the account page scenario  will not that information come from the User information?
What is the delete flow and what are the delete scenarios? 

There are not any tests in this PR.

src/NuGetGallery.Core/Entities/ReservedPrefix.cs
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+
+namespace NuGetGallery
+{
+    public class ReservedPrefix : IEntity
+    {
+        public ReservedPrefix() : this(null, false, false)
```
joelverhagen: nit: label null and bool params

src/NuGetGallery.Core/Entities/PackageRegistration.cs
shishirx34: That would be for users to find which prefixes they own. But for a given package which all prefix it matches would come from this list. Its possible you are a co-owner on a package which matches a prefix owned by someone else so it won't show up in your account page. I am just spitballing here, but you see the point of keeping this mapping of package ids to prefixes, right? 

The delete flow would be when a prefix is deleted we want to remove the "verified" tag on a package reservation **only if** there are no other reserved prefixes matching this package ID owned by one of the owners of this package.

I am planning to add tests with the next PR for adding the admin flow of adding/removing prefixes. That way we will have complete picture. I will look into it and if possible add a few tests in here as well.

src/NuGetGallery.Core/Entities/ReservedPrefix.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        [StringLength(CoreConstants.MaxPackageIdLength)]
+        [Required]
+        public string Value { get; set; }
+
+        public bool IsPublicNamespace { get; set; }
+
+        public bool IsExactMatch { get; set; }
+
+        public virtual ICollection<PackageRegistration> PackageRegistrations { get; set; }
```
skofman1: I'm concerned about unnecessary joins and loading a lot of not needed info. For example, a prefix can have multiple package registrations. Not every time I load a prefix I need those registrations (that can be in the hundreds). If you write the code that loads ReservedPrefix carefully, this can be avoided. But you should be aware.

src/NuGetGallery.Core/Entities/PackageRegistration.cs
skofman1: How do you map a package to a prefix?
Is Microsoft.Asp.Bla is under Microsoft.*, Microsoft.Asp.*, both?

src/NuGetGallery/Migrations/201707131923337_PrefixReservation.cs
```diff
@@ -0,0 +1,68 @@
+namespace NuGetGallery.Migrations
+{
```
skofman1: header

src/NuGetGallery.Core/Entities/PackageRegistration.cs
shishirx34: Its under both. When you allocate a reserved prefix to a user, we will check all the packages owned by that user and check if that package matches this reserved prefix. If yes, we add the mapping to it.

src/NuGetGallery.Core/Entities/ReservedPrefix.cs
shishirx34: Sure, I will keep it in mind and optimize it.

src/NuGetGallery/Migrations/201707131923337_PrefixReservation.cs
```diff
@@ -0,0 +1,68 @@
...
+                        User_Key = c.Int(nullable: false),
+                    })
+                .PrimaryKey(t => new { t.ReservedPrefix_Key, t.User_Key })
+                .ForeignKey("dbo.ReservedPrefixes", t => t.ReservedPrefix_Key, cascadeDelete: true)
+                .ForeignKey("dbo.Users", t => t.User_Key, cascadeDelete: true)
+                .Index(t => t.ReservedPrefix_Key)
+                .Index(t => t.User_Key);
+            
+            AddColumn("dbo.Users", "Verified", c => c.Boolean(nullable: false));
+            AddColumn("dbo.PackageRegistrations", "Verified", c => c.Boolean(nullable: false));
```
skofman1: Do we need to have "Verified" explicitly on PackageRegistrations table, or perhaps simply finding the registration has a prefix in dbo.PackageRegistrationReservedPrefixes is enough?

shishirx34: I think having it on the PR table is better. We don't need to do additional lookups to determine if a visual indicator is necessary, this makes it easier for auxiliary job too. I would keep it on this table.

skofman1: You are introducing complexity by forcing two tables to be consistent. For example, when updating a package with a prefix you are going to require a transaction. IMO, it's better to keep it simple, and have the correct indices for a quick join.

shishirx34: Had an offline talk with Svetlana. We are in agreement that it would be beneficial to have this property on PackageRegistrations given that this would be set most cases only once plus not doing this would involve making joins on every query to the package details page(high frequency). Leaving this in.

src/NuGetGallery.Core/Entities/ReservedPrefix.cs
```diff
@@ -0,0 +1,39 @@
...
+            ReservedPrefixOwners = new HashSet<User>();
+        }
+
+        [StringLength(CoreConstants.MaxPackageIdLength)]
+        [Required]
+        public string Value { get; set; }
+
+        public bool IsPublicNamespace { get; set; }
+
+        public bool IsExactMatch { get; set; }
```
chenriksson: IsExactMatch is when the 'ReservedPrefix' isn't a prefix? :-)

Maybe the entity should be ReservedNamespace, with column IsPrefix? I would probably also consider renaming IsPublicNamespace to SharedNamespace to indicate that it was shared by the original owner.

src/NuGetGallery.Core/Entities/User.cs
```diff
@@ -36,8 +37,13 @@ public User(string username)
         public string Username { get; set; }
 
         public virtual ICollection<Role> Roles { get; set; }
+
         public bool EmailAllowed { get; set; }
 
+        public bool Verified { get; set; }
```
chenriksson: As mentioned before, I would consider removing User.Verified. I don't think it makes sense on the User. It's really a package state when the package ID matches a reserved namespace, right?

src/NuGetGallery.Core/Entities/ReservedPrefix.cs
shishirx34: Yes that is what it is. 

I am fine with those names as well.

PR #4414, opened 2017-07-18
Retrieval and Display for READMEs on Display Packages page 

JarahSi:
Please let me know if you have any concerns that I may have not addressed. We are going to contact legal over the use of Clamp.js

readme collapse 2
readme collapse 3
readme collapse


.vs/config/applicationhost.config
```diff
@@ -1,1020 +1,1040 @@
...
-                <add name="WebServiceHandlerFactory-ISAPI-4.0_64bit" path="*.asmx" verb="GET,HEAD,POST,DEBUG" modules="IsapiModule" scriptProcessor="%windir%\Microsoft.NET\Framework64\v4.0.30319\aspnet_isapi.dll" preCondition="classicMode,runtimeVersionv4.0,bitness64" responseBufferLimit="0" />
-                <add name="HttpRemotingHandlerFactory-rem-ISAPI-4.0_64bit" path="*.rem" verb="GET,HEAD,POST,DEBUG" modules="IsapiModule" scriptProcessor="%windir%\Microsoft.NET\Framework64\v4.0.30319\aspnet_isapi.dll" preCondition="classicMode,runtimeVersionv4.0,bitness64" responseBufferLimit="0" />
-                <add name="HttpRemotingHandlerFactory-soap-ISAPI-4.0_64bit" path="*.soap" verb="GET,HEAD,POST,DEBUG" modules="IsapiModule" scriptProcessor="%windir%\Microsoft.NET\Framework64\v4.0.30319\aspnet_isapi.dll" preCondition="classicMode,runtimeVersionv4.0,bitness64" responseBufferLimit="0" />
-                <add name="svc-ISAPI-4.0_64bit" path="*.svc" verb="*" modules="IsapiModule" scriptProcessor="%windir%\Microsoft.NET\Framework64\v4.0.30319\aspnet_isapi.dll" preCondition="classicMode,r
```
chenriksson: This file should be reverted.

src/NuGetGallery/App_Start/AppActivator.cs
```diff
@@ -205,6 +205,7 @@ private static void BundlingPostStart()
             BundleTable.Bundles.Add(homeScriptBundle);
 
             var displayPackageScriptBundle = new ScriptBundle("~/Scripts/gallery/page-display-package.min.js")
+                .Include("~/Scripts/gallery/clamp.js")
```
chenriksson: This should be minified, at least for Release builds

src/NuGetGallery/Constants.cs
```diff
@@ -25,11 +25,15 @@ public static class Constants
...
         public const string NuGetPackageFileExtension = ".nupkg";
+        public const string MarkdownFileExtension = ".md";
+        public const string HtmlFileExtension = ".html";
         public const string PackageFileDownloadUriTemplate = "packages/{0}/{1}/download";
         public const string PackageFileSavePathTemplate = "{0}.{1}{2}";
+        public const string ActiveReadMeFileSavePathTemplate = "active/{0}/{1}{2}";
         public const string PackageFileBackupSavePathTemplate = "{0}/{1}/{2}.{3}";
 
         public const string PackagesFolderName = "packages";
+        public const string PackageReadMeFolderName = "readmes";
```
chenriksson: Looks like you and @fayrose coordinated on these, except for the variable for readme container where @fayrose has 'ReadMeContainerName'. Be careful when merging these.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -434,6 +434,15 @@ public virtual JsonResult UploadPackageProgress()
 
             var model = new DisplayPackageViewModel(package, packageHistory);
 
+            if (package.HasReadMe.HasValue && package.HasReadMe.Value)
```
chenriksson: This could probably be shortened to 'if (package.HasReadMe ?? false)'

src/NuGetGallery/Migrations/201706221634506_AddRepositoryURL_ReadMe.cs
```diff
@@ -0,0 +1,26 @@
...
+        public override void Down()
+        {
+            DropColumn("dbo.PackageHistories", "RepositoryUrl");
+            DropColumn("dbo.PackageEdits", "ReadmeModified");
+            DropColumn("dbo.PackageEdits", "RepositoryUrl");
+            DropColumn("dbo.Packages", "HasReadMe");
+            DropColumn("dbo.Packages", "RepositoryUrl");
+        }
+    }
+}
```
chenriksson: Schema/Migrations changes should be in your PR, only in @laurenshen 's. The "20170622*" migrations files are obsolete and likely were kept during auto-merge on your last rebase... you should delete them from your branch.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1555,5 +1564,6 @@ private static bool IsDifferent(string posted, string package)
             // Ignore those pesky '\r' characters which screw up comparisons.
             return !String.Equals(posted.Replace("\r", ""), package.Replace("\r", ""), StringComparison.Ordinal);
         }
+
```
chenriksson: Would revert this empty line

src/NuGetGallery/Scripts/gallery/clamp.js
```diff
@@ -0,0 +1,261 @@
+/*!
```
chenriksson: Would put beside other scripts in NuGetGallery/Scripts, not 'gallery' subfolder. Also, we should include both full and minified versions, and use minified for Release builds.

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -13,7 +13,30 @@ $(function () {
                 copyButton.popover('destroy');
```
chenriksson: Put in /Scripts like other scripts

src/NuGetGallery/Services/CloudBlobFileStorageService.cs
```diff
@@ -54,6 +54,7 @@ public CloudBlobFileStorageService(ICloudBlobClient client, IAppConfiguration co
             return await blob.ExistsAsync();
         }
 
+
```
chenriksson: Remove this extra line

src/NuGetGallery/Services/CloudBlobFileStorageService.cs
```diff
@@ -154,7 +156,6 @@ public CloudBlobFileStorageService(ICloudBlobClient client, IAppConfiguration co
                 case Constants.UploadsFolderName:
                     creationTask = PrepareContainer(folderName, isPublic: false);
                     break;
-
```
chenriksson: Keep this line - the empty lines denote different blocks

src/NuGetGallery/Services/IFileStorageService.cs
```diff
@@ -17,6 +17,7 @@ public interface IFileStorageService
 
         Task<Stream> GetFileAsync(string folderName, string fileName);
 
+
```
chenriksson: Remove this extra line

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -128,12 +155,33 @@ private static string BuildFileName(Package package)
...
 
-            return BuildFileName(
+            return BuildPackageFileName(
                 package.PackageRegistration.Id, 
                 String.IsNullOrEmpty(package.NormalizedVersion) ?
                     NuGetVersionFormatter.Normalize(package.Version) :
                     package.NormalizedVersion);
         }
+        
+        private static string BuildReadmeFileName(Package package)
```
chenriksson: You should coordinate changes in this file with @fayrose 

Her PR has 'BuildReadmeFileName(Package package, string extension);', which supports both MD and HTML formats... so you should probably stick with that.

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -85,11 +85,38 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
 
         public async Task<Stream> DownloadPackageFileAsync(Package package)
         {
-            var fileName = BuildFileName(package);
+            var fileName = BuildPackageFileName(package);
             return (await _fileStorageService.GetFileAsync(Constants.PackagesFolderName, fileName));
         }
+        
+        public async Task<Stream> DownloadReadmeFileAsync(Package package)
```
chenriksson: What format? Would include fileExtension arg, and/or have separate APIs for MD/HTML formats.

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -85,11 +85,38 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
...
+            {
+                throw new ArgumentNullException("Package cannot be null!");
+            }
+            var fileName = BuildReadmeFileName(package);
+            return (await _fileStorageService.GetFileAsync(Constants.PackageReadMeFolderName, fileName));
+        }
+
+        private static string BuildPackageFileName(string id, string version)
+        {
+
```
chenriksson: Formatting - avoid random empty lines like this

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -108,13 +135,13 @@ private static string BuildFileName(string id, string version)
...
-                Constants.PackageFileSavePathTemplate,
+                pathTemplate,
                 id.ToLowerInvariant(),
                 version.ToLowerInvariant(),
-                Constants.NuGetPackageFileExtension);
+                extension);
         }
-
-        private static string BuildFileName(Package package)
+
```
chenriksson: Remove tab

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -199,6 +199,35 @@
...
+                        <span>Documentation</span>
+                    </a>
+                </h2>
+            </div>
+
+            <div id="readme-container" class="collapse in">
+                <div id="readme-clamped" class="collapse in">
+                    @Html.Raw(Model.ReadmeHTML)
+                </div>
+                <script src="~/Scripts/gallery/clamp.js" type="text/javascript">
```
chenriksson: Can this be moved to bottomScripts?

Were you able to avoid the initial display of the full MD on page load, before truncation?

tests/NuGetGallery.Facts/Services/FileSystemFileStorageServiceFacts.cs
```diff
@@ -290,6 +290,119 @@ public class TheGetFileMethod
             }
         }
```
chenriksson: Our CI shows unit tests failing for your branch, so have a look if they're passing. Make sure to run before any push.

- NuGetGallery.PackageFacts+PositiveScenarios.ApplyPackageEditUpdatesPackage [FAIL]

tests/NuGetGallery.Facts/Services/FileSystemFileStorageServiceFacts.cs
```diff
@@ -599,6 +712,19 @@ public class TheSaveFileMethod
                     return false;
                 }
             }
+
+            
+        }
```
chenriksson: Remove these empty lines

tests/NuGetGallery.Facts/Services/FileSystemFileStorageServiceFacts.cs
```diff
@@ -599,6 +712,19 @@ public class TheSaveFileMethod
...
+
+        static string BuildReadmeFileName(
```
chenriksson: Is this used?

tests/NuGetGallery.Facts/Services/PackageFileServiceFacts.cs
```diff
@@ -432,7 +432,7 @@ public class TheStorePackageFileInBackupLocationAsyncMethod
             }
         }
 
-        static string BuildFileName(
+        static string BuildPackageFileName(
             string id,
             string version)
         {
```
chenriksson: What about a test for the new code in PackagesController?

src/NuGetGallery/Constants.cs
JarahSi: I think we need to switch to mine because it fits the naming scheme already established with the folder name titles

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -434,6 +434,15 @@ public virtual JsonResult UploadPackageProgress()
 
             var model = new DisplayPackageViewModel(package, packageHistory);
 
+            if (package.HasReadMe.HasValue && package.HasReadMe.Value)
+            {
+                //Reads the README file and push to the view
+                using (var reader = new StreamReader(await _packageFileService.DownloadReadmeFileAsync(package), Encoding.UTF8))
+                {
+                    model.ReadmeHTML = reader.ReadToEnd();
```
loic-sharma: `await reader.ReadToEndAsync();`?

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -85,11 +85,38 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
...
-            var fileName = BuildFileName(package);
+            var fileName = BuildPackageFileName(package);
             return (await _fileStorageService.GetFileAsync(Constants.PackagesFolderName, fileName));
         }
+        
+        public async Task<Stream> DownloadReadmeFileAsync(Package package)
+        {
+            if(package == null)
+            {
+                throw new ArgumentNullException("Package cannot be null!");
```
loic-sharma: Instead of the string, do `nameof(package)` instead here.

tests/NuGetGallery.Facts/Services/FileSystemFileStorageServiceFacts.cs
```diff
@@ -290,6 +290,119 @@ public class TheGetFileMethod
             }
         }
 
+        public class TheGetReadMeMethod
```
scottbommarito: I don't think these tests are necessary.

They seem to be identical to `TheGetFileMethod` above, just with an additional test where the desired path is one for a readme. Additionally, there is no `GetReadMe` method on `FileSystemFileStorage`, so this class' name is very confusing.

Generally, unit tests one piece of the structure. Even though `PackageFileService` uses `FileSystemFileStorage`, it isn't necessary to have tests on `FileSystemFileStorage` for the exact values `PackageFileService` passes in. There are already tests in `FileSystemFileStorage` that cover what `PackageFileService` is passing in. The purpose of unit tests is that each part tests itself, and then when you put them together, everything works because everything else tested correctly.

You should have tests for `PackageFileService`'s `DownloadReadmeFileAsync`, but these tests should use a `Mock<IFileStorage>` that is setup such that it returns the expected result given that they are passed what you expect `DownloadReadmeFileAsync` passed it. These tests will verify that `DownloadReadmeFileAsync` works as expected. The other tests on `IFileStorage` will verify that the methods that `DownloadReadmeFileAsync` calls return as expected.

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -70,7 +70,7 @@ public void SetPendingMetadata(PackageEdit pendingMetadata)
         public DependencySetsViewModel Dependencies { get; set; }
         public IEnumerable<DisplayPackageViewModel> PackageVersions { get; set; }
         public string Copyright { get; set; }
-
+        public string ReadmeHTML { get; set; }
```
scottbommarito: I believe in proper camelcase, this should be
```
ReadMeHtml
```

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -434,6 +434,21 @@ public virtual JsonResult UploadPackageProgress()
 
             var model = new DisplayPackageViewModel(package, packageHistory);
 
+            if (package.HasReadMe ?? false)
+            {
+                var readmeStream = await _packageFileService.DownloadReadmeFileAsync(package, Constants.HtmlFileExtension);
```
chenriksson: It may be worth adding a simple overload for the Html version.

i.e., PackageFileService.DownloadReadMeHtmlFileAsync(Package)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -434,6 +434,21 @@ public virtual JsonResult UploadPackageProgress()
...
+                
+                if(readmeStream != null)
+                {
+                    //Reads the README file and push to the view
+                    using (var reader = new StreamReader(await _packageFileService.DownloadReadmeFileAsync(package, Constants.HtmlFileExtension), Encoding.UTF8))
+                    {
+                        model.ReadMeHtml = await reader.ReadToEndAsync();
+                    }
+                }
+
```
chenriksson: nit: remove empty line

src/NuGetGallery/Constants.cs
```diff
@@ -25,11 +25,16 @@ public static class Constants
         public const string PackageContentType = "binary/octet-stream";
         public const string OctetStreamContentType = "application/octet-stream";
         public const string NuGetPackageFileExtension = ".nupkg";
+        public const string MarkdownFileExtension = ".md";
+        public const string HtmlFileExtension = ".html";
```
chenriksson: Just discovered that we have similar constants in NuGetGallery.ContentService. Maybe you can remove from the ContentService (including JsonContentFileExtension) and reference the ones here?

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -85,11 +85,37 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
...
+            var fileName = BuildReadmeFileName(package, extension);
+            return (await _fileStorageService.GetFileAsync(Constants.PackagesReadMeFolderName, fileName));
+        }
+
+        private static string BuildPackageFileName(string id, string version)
+        {
+            return BuildFileName(Constants.PackageFileSavePathTemplate, id, version, Constants.NuGetPackageFileExtension);
+        }
+
+        public static string BuildReadmeFileName(string id, string version, string extension)
```
chenriksson: private

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -85,11 +85,37 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
...
+        {
+            return BuildFileName(Constants.ReadMeFileSavePathTemplateActive, id, version, extension);
+        }
+
+        private static string BuildReadMeMarkdownFileName(string id, string version)
```
chenriksson: Can this one be removed?

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -161,6 +208,6 @@ private static string BuildBackupFileName(string id, string version, string hash
                 version,
                 HttpServerUtility.UrlTokenEncode(hashBytes),
                 Constants.NuGetPackageFileExtension);
-        }
+        }
```
chenriksson: nit: remove extra whitespace

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -199,6 +199,35 @@
                 </div>
             </div>
 
+
```
chenriksson: nit: remove extra line

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -455,6 +455,55 @@ public class TheDisplayPackageMethod
...
+
+                // Act
+                var result = await controller.DisplayPackage("Foo", null);
+
+                // Assert
+                var model = ResultAssert.IsView<DisplayPackageViewModel>(result);
+                Assert.Equal("Foo", model.Id);
+                Assert.Equal("1.1.1", model.Version);
+                Assert.Equal("A test package!", model.Title);
+                Assert.Equal("<p>Hello World!</p>", model.ReadMeHtml);
```
chenriksson: Would remove other asserts, since this test only verifies that readme is set. Maybe add 'Assert.IsNull(model.ReadMeHtml) to the non-README test.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -455,6 +455,55 @@ public class TheDisplayPackageMethod
...
+                var controller = CreateController(
+                    packageService: packageService, indexingService: indexingService, packageFileService: fileService);
+                controller.SetCurrentUser(TestUtility.FakeUser);
+                
+                var stream = new MemoryStream();
+                var writer = new StreamWriter(stream);
+                writer.Write("<p>Hello World!</p>");
+                writer.Flush();
+                stream.Position = 0;
+                var readMeStream = stream;
```
chenriksson: Don't need both stream and readMeStream. Looks like MemoryStream has a constructor that could simplify this:

e.g., MemoryStream(Encoding.UTF8.GetBytes(text))

tests/NuGetGallery.Facts/Services/FileSystemFileStorageServiceFacts.cs
chenriksson: Looks like the new code added was PackageFileService.DownloadReadMeFileAsync. This is what you should test, in PackageFileServiceFacts.cs.

chenriksson:
@JarahSi Not sure why, but applicationHost.config is still showing as edited. Also, the old migrations files still need to be deleted. None of this should show up in the PR.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -434,13 +434,19 @@ public virtual JsonResult UploadPackageProgress()
...
             {
-                //Reads the README file and push to the view
-                using (var reader = new StreamReader(await _packageFileService.DownloadReadmeFileAsync(package), Encoding.UTF8))
+                var readmeStream = await _packageFileService.DownloadReadmeFileAsync(package, Constants.HtmlFileExtension);
+                
+                if(readmeStream != null)
                 {
-                    model.ReadMe = reader.ReadToEnd();
+                    //Reads the README file and push to the view
+                    using (var reader = new StreamReader(await _packageFileService.DownloadReadmeFileAsync(package, Constants.HtmlFileExtension), Encoding.UTF8))
```
scottbommarito: No need to call `_packageFileService.DownloadReadmeFileAsync` again--just use the value from before.

scottbommarito: Also add a `using` to the statement above.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -199,15 +199,34 @@
...
 
-            @if (Model.ReadMe != null)
-            {
-                // displays the README to the Package Description
-            <div>
-                <p>
-                    @Html.Raw(Model.ReadMe)
-                </p>
+
+            <div id="readme-collapser-container" class="">
```
scottbommarito: Don't need to give an id to this because you don't use it.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -455,6 +455,55 @@ public class TheDisplayPackageMethod
                 Assert.Equal("1.1.1", model.Version);
                 Assert.Equal("A test package!", model.Title);
             }
+
+            [Fact]
+            public async Task GivenAValidPackageWithReadMeItDisplaysReadMe()
```
scottbommarito: Nice test!

scottbommarito: We should also have the following tests:
- package has no readme, checks that readme is empty
- package has a readme but download fails, check that it handles it gracefully

skofman1:
>     %IIS_BIN%\config\schema\IIS_schema.xml.

please undo changes to this file

---
Refers to: .vs/config/applicationhost.config:7 in ae22c09. [](commit_id = ae22c09f296a55f85f4d94914194256a662a4b3c, deletion_comment = True)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -434,6 +434,22 @@ public virtual JsonResult UploadPackageProgress()
 
             var model = new DisplayPackageViewModel(package, packageHistory);
 
+            if (package.HasReadMe ?? false)
```
skofman1: HasReadMe will be null for old packages, right? I recommend hiding the fact that it's nullable in lower levels (like DB read logic) and here using a method that returns true/false

JarahSi: what do you mean by that?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -434,6 +434,22 @@ public virtual JsonResult UploadPackageProgress()
...
+            {
+                try
+                {
+                    //Reads the README file and push to the view
+                    using (var reader = new StreamReader(await _packageFileService.DownloadReadmeFileAsync(package, Constants.HtmlFileExtension), Encoding.UTF8))
+                    {
+                        model.ReadMeHtml = await reader.ReadToEndAsync();
+                    }
+                }
+                catch (Exception)
```
skofman1: catching all exceptions is a bad practice. You might catch exceptions like OutOfMemory, and hide an issue that causes the service to crash.
If you go over the code is GetBlobContentAsync you will see they throw (and catch) only some exceptions (like StorageExceptions).

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -434,6 +434,22 @@ public virtual JsonResult UploadPackageProgress()
...
+                {
+                    model.ReadMeHtml = null;
+                }
+            }
+
             if (package.IsOwner(User))
```
skofman1: In case the readme load fails you should show an error page (i.e. let the exception propagate). We shouldn't hide the error.

src/NuGetGallery/Migrations/201706221609417_AddRepositoryURL_ReadMe.cs
```diff
@@ -0,0 +1,18 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddRepositoryURL_ReadMe : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.Packages", "HasReadMe", c => c.Boolean(nullable: false));
```
skofman1: should be nullable: true
When you run this migration in production the result will be that the column will be added, and a value of "false" will be set for all packages. However, because the Packages table is huge, this will take a long time (minutes), and will cause a performance impact on the DB. Hence. make the column nullable, and in your code, assume that the column value can be null

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -85,11 +85,32 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
...
         public async Task<Stream> DownloadPackageFileAsync(Package package)
         {
-            var fileName = BuildFileName(package);
+            var fileName = BuildPackageFileName(package);
             return (await _fileStorageService.GetFileAsync(Constants.PackagesFolderName, fileName));
         }
+        
+        public async Task<Stream> DownloadReadmeFileAsync(Package package, string extension)
+        {
+            if(package == null)
```
skofman1: nit: space after if. Newline after }

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -128,12 +149,33 @@ private static string BuildFileName(Package package)
...
+            return BuildPackageFileName(
                 package.PackageRegistration.Id, 
                 String.IsNullOrEmpty(package.NormalizedVersion) ?
                     NuGetVersionFormatter.Normalize(package.Version) :
                     package.NormalizedVersion);
         }
+        
+        private static string BuildReadmeFileName(Package package, string extension)
+        {
+            if (package == null)
```
skofman1: same

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -454,6 +455,130 @@ public class TheDisplayPackageMethod
...
+                // Act
+                var result = await controller.DisplayPackage("Foo", null);
+
+                // Assert
+                var model = ResultAssert.IsView<DisplayPackageViewModel>(result);
+                Assert.Equal("<p>Hello World!</p>", model.ReadMeHtml);
+            }
+
+            [Fact]
+            public async Task GivenAPackageWthNoReadMeShowsThatReadMeIsEmpty()
```
skofman1: With

skofman1:
:clock1:

src/NuGetGallery/Controllers/PackagesController.cs
skofman1: Lets chat f2f when u r here

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -454,6 +455,127 @@ public class TheDisplayPackageMethod
...
+                var result = await controller.DisplayPackage("Foo", null);
+
+                // Assert
+                var model = ResultAssert.IsView<DisplayPackageViewModel>(result);
+                Assert.Equal("<p>Hello World!</p>", model.ReadMeHtml);
+            }
+
+            [Fact]
+            public async Task GivenAPackageWithNoReadMeShowsThatReadMeIsEmpty()
+            {
```
chenriksson: What's the difference between this test and the one below? Both look like they return null stream from Download and verify that readme isn't set.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -454,6 +455,127 @@ public class TheDisplayPackageMethod
                 Assert.Equal("Foo", model.Id);
                 Assert.Equal("1.1.1", model.Version);
                 Assert.Equal("A test package!", model.Title);
+                Assert.Null(model.ReadMeHtml);
+            }
+
+            [Fact]
+            public async Task GivenAValidPackageWithReadMeItDisplaysReadMe()
+            {
```
chenriksson: The code looks same as the test below for the missing readme. Consider using a helper method that accepts an argument for what stream should be returned by the package file service.

tests/NuGetGallery.Facts/Services/PackageFileServiceFacts.cs
```diff
@@ -432,15 +433,42 @@ public class TheStorePackageFileInBackupLocationAsyncMethod
             }
         }
 
+        public class TheRetrieveReadMeMethod
```
chenriksson: Should this be TheDownloadReadMeFileAsyncMethod?

tests/NuGetGallery.Facts/Services/PackageFileServiceFacts.cs
```diff
@@ -432,15 +433,42 @@ public class TheStorePackageFileInBackupLocationAsyncMethod
...
+        {
+            [Fact]
+            public async Task WillDownloadReadMeAsync()
+            {
```
chenriksson: Would use the 'Arrange, Act, Assert' comments

src/NuGetGallery/Constants.cs
chenriksson: Agreed we can clean up later, as this is also common to @fayrose 's PR.

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -128,12 +149,33 @@ private static string BuildFileName(Package package)
...
+            {
+                throw new ArgumentNullException(nameof(package));
+            }
+
+            if (package.PackageRegistration == null ||
+                String.IsNullOrWhiteSpace(package.PackageRegistration.Id) ||
+                (String.IsNullOrWhiteSpace(package.NormalizedVersion) && String.IsNullOrWhiteSpace(package.Version)))
```
chenriksson: nit: lowercase first letter in 'string.IsNullOrWhiteSpace'

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -199,6 +199,38 @@
                 </div>
             </div>
 
+            @if (Model.ReadMeHtml != null)
+            {
+                <div id="readme-collapser-container" class="">
```
chenriksson: Empty class="" can be removed

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -454,6 +455,70 @@ public class TheDisplayPackageMethod
...
+                //Arrange
+                var readMeStream = new MemoryStream(Encoding.UTF8.GetBytes("<p>Hello World!</p>"));
+                
+                //Act
+                var result = await CreateANewViewModelForReadMeTests((Stream)readMeStream);
```
chenriksson: Don't need cast to Stream here

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -454,6 +455,70 @@ public class TheDisplayPackageMethod
...
+
+                //Assert
+                var model = ResultAssert.IsView<DisplayPackageViewModel>(result);
+                Assert.Null(model.ReadMeHtml);
+            }
+
+            [Fact]
+            public async Task GivenAPackageWithNoReadMeShowsThatReadMeIsEmpty()
+            {
+
```
chenriksson: nit: remove empty line

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1960,6 +2025,38 @@ public class TheSetLicenseReportVisibilityMethod
                 Assert.Equal(@"~\Bar.cshtml", ((RedirectResult)result).Url);
             }
         }
+
+        private static async Task<ActionResult> CreateANewViewModelForReadMeTests(Stream stream)
```
chenriksson: Not crazy about this name... maybe rename to GetDisplayPackageResultWithReadMeStream, and rename stream arg to readMeHtmlStream?

tests/NuGetGallery.Facts/Services/PackageFileServiceFacts.cs
```diff
@@ -9,6 +9,7 @@
 using System.Web.Mvc;
 using Moq;
 using Xunit;
+using System.Text;
```
chenriksson: sort - System namespaces go first

tests/NuGetGallery.Facts/Services/PackageFileServiceFacts.cs
```diff
@@ -432,15 +433,44 @@ public class TheStorePackageFileInBackupLocationAsyncMethod
...
+                    {
+                        Id = "Foo",
+                    },
+                    Version = "01.1.01",
+                };
+                fileStorageSvc.Setup(f => f.GetFileAsync(It.IsAny<string>(), It.IsAny<string>())).Returns(Task.FromResult((Stream)stream)).Verifiable();
+
+                //Act
+                var result = await service.DownloadReadmeFileAsync(package, Constants.HtmlFileExtension);
+                var reader = new StreamReader(result);
```
chenriksson: usings around reader and stream

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -13,7 +13,30 @@ $(function () {
...
             }, 1000);
         });
-    }
+    }    
+    $clamp(document.getElementById("readme-clamped"),
```
fayrose: Replace `document.getElementById("readme-clamped")` with equivalent jQuery for cleanliness

`$("#readme-clamped")`

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -85,11 +85,32 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
...
+        {
+            if (package == null)
+            {
+                throw new ArgumentNullException("Package cannot be null!");
+            }
+            var fileName = BuildReadmeFileName(package, extension);
+            return (await _fileStorageService.GetFileAsync(Constants.PackagesReadMeFolderName, fileName));
+        }
+
+        private static string BuildPackageFileName(string id, string version)
```
fayrose: Can combine BuildPackageFileName and BuildReadMeFileName into one function taking in id, version and extension.

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -85,11 +85,32 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
...
+        {
+            return BuildFileName(Constants.PackageFileSavePathTemplate, id, version, Constants.NuGetPackageFileExtension);
+        }
+
+        private static string BuildReadmeFileName(string id, string version, string extension)
```
fayrose: Change to BuildReadMeFileName in accordance with convention

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -128,12 +149,33 @@ private static string BuildFileName(Package package)
...
 
-            return BuildFileName(
+            return BuildPackageFileName(
                 package.PackageRegistration.Id, 
                 String.IsNullOrEmpty(package.NormalizedVersion) ?
                     NuGetVersionFormatter.Normalize(package.Version) :
                     package.NormalizedVersion);
         }
+        
+        private static string BuildReadmeFileName(Package package, string extension)
```
fayrose: Change to BuildReadMeFileName; consolidate with BuildPackageFileName

src/NuGetGallery/Scripts/gallery/page-display-package.js
JarahSi: That did not work

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -434,6 +434,25 @@ public virtual JsonResult UploadPackageProgress()
...
+                    {
+                        // Reads the README file and push to the view
+                        using (var reader = new StreamReader(readmeStream, Encoding.UTF8))
+                        {
+                            model.ReadMeHtml = await reader.ReadToEndAsync();
+                        }
+                    }
+                    else if (User.IsInRole(Constants.AdminRoleName) || package.IsOwner(User))
+                    {
+                        TempData["Message"] = ("The ReadMe for this package could not be found");
```
scottbommarito: nit: no need for parenthesis

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -85,11 +85,27 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
...
         }
+        
+        public async Task<Stream> DownloadReadmeFileAsync(Package package, string extension)
+        {
+            if (package == null)
+            {
+                throw new ArgumentNullException("Package cannot be null!");
+            }
+            var fileName = BuildReadmeFileName(package, extension);
+            return (await _fileStorageService.GetFileAsync(Constants.PackagesReadMeFolderName, fileName));
```
scottbommarito: two things

1 - can get rid of parenthesis
2 - can return the `Task` directly instead of needing to make this method `async`
```
public Task<Stream> DownloadReadmeFileAsync(Package package, string extension)
{
    if (package == null)
    {
        throw new ArgumentNullException("Package cannot be null!");
    }
    
    var fileName = BuildReadmeFileName(package, extension);
    return _fileStorageService.GetFileAsync(Constants.PackagesReadMeFolderName, fileName);
}
```

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
scottbommarito: It was a little bit clearer to understand before he moved it into a helper method (which was the right move).

Basically, this test is "if we say there's a readme, but we don't get a readme (missing or the download fails), what happens?". The test below is "if we say there's no readme, what do we do?". The test above is "if we say there's a readme and we successfully get the readme, what happens?".

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -454,6 +455,69 @@ public class TheDisplayPackageMethod
...
+                //Assert
+                var model = ResultAssert.IsView<DisplayPackageViewModel>(result);
+                Assert.Null(model.ReadMeHtml);
+            }
+
+            [Fact]
+            public async Task GivenAPackageWithNoReadMeShowsThatReadMeIsEmpty()
+            {
+                // Arrange
+                var packageService = new Mock<IPackageService>();
```
scottbommarito: Could we use the same helper method as the other ones but add a `bool hasReadme` to the parameters?

tests/NuGetGallery.Facts/Services/PackageFileServiceFacts.cs
```diff
@@ -432,15 +433,48 @@ public class TheStorePackageFileInBackupLocationAsyncMethod
             }
         }
 
+        public class TheDownloadReadMeFileAsyncMethod
+        {
+            [Fact]
+            public async Task WillDownloadReadMeAsync()
```
scottbommarito: No test for failed download?

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
chenriksson: Maybe rename to 'GivenPackageWithHasReadMeSetAndNoBlob_ReturnsModelWithoutReadme'?

chenriksson:
Agree with @scottbommarito 's remaining comments, then I'm :shipit:

PR #4420, opened 2017-07-19
Adding support for EU cookie compliance

chenriksson:
There will be 3 PRs altogether: Gallery, Shims, Deployment

When shim is deployed, banner will display if client IP is from region which requires it. Message indicates that user gives consent if they continue to navigate, and does not include any buttons. Provided JS detects and sets consent cookie if user clicks on any non-external links, which automatically closes the banner.

I've verified the banner locally, by spoofing the consent status. I've verified AI logs in DEV, but am not able to test the banner there yet since domain isn't enabled. Will see if there's a way to enable this.

Note:
- Onus is on us to monitor logs and report if service isn't working.
- API includes 'fire and forget' background refresh; with initial refresh done when service is created (see DefaultDependenciesModule.RegisterCookieComplianceService)

### DIFF
--- a/src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
+++ b/src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+using System.Threading.Tasks;
+using System.Web;
+using NuGetGallery.Diagnostics;
+
+namespace NuGetGallery.Cookies
+{
+    /// <summary>
+    /// Base cookie compliance service with access to some Gallery resources.
+    /// </summary>
+    public abstract class CookieComplianceServiceBase : ICookieComplianceService
+    {
+        private string _siteName;
+        private IDiagnosticsSource _diagnostics;
+
+        protected string SiteName
+        {
+            get
+            {
+                return _siteName ?? throw new InvalidOperationException("Cookie compliance service has not been initialized");
+            }
+        }
+
+        protected IDiagnosticsSource Diagnostics
+        {
+            get
+            {
+                return _diagnostics ?? throw new InvalidOperationException("Cookie compliance service has not been initialized");
+            }
+        }
+
+        protected string Locale
+        {
+            get
+            {
+                return CultureInfo.CurrentCulture.Name;
+            }
+        }
+
+        public virtual Task InitializeAsync(string siteName, IDiagnosticsService diagnostics)
+        {
+            _siteName = siteName;
+            _diagnostics = diagnostics.GetSource("CookieComplianceService");
+            return Task.Delay(0);
+        }
+        
+        public abstract bool CanWriteNonEssentialCookies(HttpRequestBase request);
+
+        public abstract bool NeedsConsentForNonEssentialCookies(HttpRequestBase request);
+
+        public abstract string GetConsentMarkup();
+
+        public abstract string[] GetConsentScripts();
+
+        public abstract string[] GetConsentStylesheets();
+    }
+}
--- a/src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
+++ b/src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
@@ -17,7 +17,7 @@ namespace NuGetGallery.Cookies
         private string _siteName;
         private IDiagnosticsSource _diagnostics;
 
-        protected string SiteName
+        protected internal string SiteName
         {
             get
             {
@@ -25,7 +25,7 @@ namespace NuGetGallery.Cookies
             }
         }
 
-        protected IDiagnosticsSource Diagnostics
+        protected internal IDiagnosticsSource Diagnostics
         {
             get
             {
@@ -33,7 +33,7 @@ namespace NuGetGallery.Cookies
             }
         }
 
-        protected string Locale
+        protected internal string Locale
         {
             get
             {
--- a/src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
+++ b/src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
@@ -14,6 +14,8 @@ namespace NuGetGallery.Cookies
     /// </summary>
     public abstract class CookieComplianceServiceBase : ICookieComplianceService
     {
+        private const string DiagnosticsSourceName = "CookieComplianceService";
+
         private string _siteName;
         private IDiagnosticsSource _diagnostics;
 
@@ -21,7 +23,7 @@ namespace NuGetGallery.Cookies
         {
             get
             {
-                return _siteName ?? throw new InvalidOperationException("Cookie compliance service has not been initialized");
+                return _siteName ?? throw new InvalidOperationException(CoreStrings.CookieComplianceServiceNotInitialized);
             }
         }
 
@@ -29,7 +31,7 @@ namespace NuGetGallery.Cookies
         {
             get
             {
-                return _diagnostics ?? throw new InvalidOperationException("Cookie compliance service has not been initialized");
+                return _diagnostics ?? throw new InvalidOperationException(CoreStrings.CookieComplianceServiceNotInitialized);
             }
         }
 
@@ -44,7 +46,7 @@ namespace NuGetGallery.Cookies
         public virtual Task InitializeAsync(string siteName, IDiagnosticsService diagnostics)
         {
             _siteName = siteName;
-            _diagnostics = diagnostics.GetSource("CookieComplianceService");
+            _diagnostics = diagnostics.GetSource(DiagnosticsSourceName);
             return Task.Delay(0);
         }
         
--- a/src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
+++ b/src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
@@ -40,6 +40,8 @@ namespace NuGetGallery.Cookies
 
         public abstract bool NeedsConsentForNonEssentialCookies(HttpRequestBase request);
 
+        public abstract CookieConsentMessage GetConsentMessage(HttpRequestBase request, string locale = null);
+
         public abstract string GetConsentMarkup(HttpRequestBase request, string locale = null);
 
         public abstract IEnumerable<string> GetConsentScripts(HttpRequestBase request, string locale = null);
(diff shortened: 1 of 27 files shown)


src/NuGetGallery.Core/Diagnostics/IDiagnosticsSource.cs
```diff
@@ -0,0 +1,18 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace NuGetGallery.Diagnostics
+{
+    public interface IDiagnosticsSource
+    {
+        void TraceEvent(TraceEventType type, int id, string message,
```
joelverhagen: We should have a method on here to track exceptions. This would call into AI's `TrackException` method.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -486,27 +489,46 @@ private static IAuditingService CombineServices(IEnumerable<IAuditingService> se
...
+            var service = GetAddInServices<ICookieComplianceService>(builder).FirstOrDefault() ?? new NullCookieComplianceService();
+            
+            builder.RegisterInstance(service)
+                .AsSelf()
+                .As<ICookieComplianceService>()
+                .SingleInstance();
+            
+            // Initialize the service on App_Start to avoid any performance degradation during initial requests.
+            var siteName = configuration.GetSiteRoot(true);
+            HostingEnvironment.QueueBackgroundWorkItem(async cancellationToken => await service.InitializeAsync(siteName, diagnostics));
```
joelverhagen: How can we be sure this initialization is completed before the first request?

src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                return CultureInfo.CurrentCulture.Name;
+            }
+        }
+
+        public virtual Task InitializeAsync(string siteName, IDiagnosticsService diagnostics)
+        {
+            _siteName = siteName;
+            _diagnostics = diagnostics.GetSource(DiagnosticsSourceName);
+            return Task.Delay(0);
```
joelverhagen: nit: `Task.Yield` is a bit more explicit than a "0 sleep".

src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+        
+        public abstract bool CanWriteNonEssentialCookies(HttpRequestBase request);
+
+        public abstract bool NeedsConsentForNonEssentialCookies(HttpRequestBase request);
+
+        public abstract string GetConsentMarkup();
+
+        public abstract string[] GetConsentScripts();
```
joelverhagen: Nit, could we make this interface expose `IEnumerable<string>` or `IReadOnlyList<string>` (if indexing is required)?

src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
```diff
@@ -0,0 +1,63 @@
...
+using NuGetGallery.Diagnostics;
+
+namespace NuGetGallery.Cookies
+{
+    /// <summary>
+    /// Base cookie compliance service with access to some Gallery resources.
+    /// </summary>
+    public abstract class CookieComplianceServiceBase : ICookieComplianceService
+    {
+        private const string DiagnosticsSourceName = "CookieComplianceService";
```
joelverhagen: Instead of a constant can we use one of the properties on `this.GetType()` so that the source name matches that actual type name?

src/NuGetGallery/Views/Shared/Layout.cshtml
```diff
@@ -26,8 +26,14 @@
         @RenderSection("TopScripts", required: false)
         @ViewHelpers.ReleaseMeta()
         @ViewHelpers.InstrumentationScript()
+
+        @ViewHelpers.CookieComplianceScripts()
```
joelverhagen: Could we move this to the `feature-redesign` branch? This integration will be different there.

My understanding is the deadline for this work is after the expected roll-out date for the redesign.

joelverhagen: Must these scripts be in the `<head>`? If you put them at the end of the `<body>` then the loading of these scripts will not block the page render. Note that this is the case with almost all scripts in `feature-redesign`.

src/NuGetGallery/Controllers/NuGetContext.cs
```diff
@@ -14,11 +15,15 @@ public class NuGetContext
...
         {
             Config = DependencyResolver.Current.GetService<IGalleryConfigurationService>();
+            Cookies = DependencyResolver.Current.GetService<ICookieComplianceService>();
 
             _currentUser = new Lazy<User>(() => ctrl.OwinContext.GetCurrentUser());
         }
 
         public IGalleryConfigurationService Config { get; internal set; }
+
+        public ICookieComplianceService Cookies { get; }
```
skofman1: >Cookies [](start = 40, length = 7)

Cookies is a confusing name... the full name  CookieComplianceService will be more readable

src/NuGetGallery/Views/NuGetViewBase.cs
```diff
@@ -25,6 +26,11 @@ public User CurrentUser
         {
             get { return NuGetContext.CurrentUser; }
         }
+        
+        public ICookieComplianceService Cookies
```
skofman1: Same.. please use full name CookieComplianceService

src/NuGetGallery.Core/Diagnostics/IDiagnosticsSource.cs
```diff
@@ -0,0 +1,18 @@
...
+            [CallerMemberName] string member = null, [CallerFilePath] string file = null, [CallerLineNumber] int line = 0);
+
+        void PerfEvent(string name, TimeSpan time, IEnumerable<KeyValuePair<string, object>> payload);
```
skofman1: Where is this code used and how?

src/NuGetGallery.Core/Properties/AssemblyInfo.cs
```diff
@@ -8,6 +8,8 @@
 using System.Runtime.InteropServices;
 
 [assembly: AssemblyTitle("NuGetGallery.Core")]
+[assembly: InternalsVisibleTo("NuGetGallery.Core.Facts")]
```
skofman1: nit: can we move this line under [assembly: AssemblyCulture("")]  ?
The internals visible is not part of the assembly properties like title and complany

src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
chenriksson: Yield doesn't return Task. Really, I want Task.CompletedTask, but Gallery.Core doesn't use .NET 4.6 yet. Should it?

joelverhagen: `await Task.Yield()`. Even better than that `Task.FromResult(0)` (which can be a static).

I think it would be breaking to switch to 4.6. If I recall, we have some jobs that depend on NuGetGallery.Core and are not yet 4.6.

src/NuGetGallery/Views/Shared/Layout.cshtml
chenriksson: Per discussion - will go ahead and roll out before redesign. Once deployment/shims are merged, redesign will need to FI before redeployment. We'll also need to register various Gallery domains with the service so the consent banner actually displays... getting "Domain not registered" in AI right now.

chenriksson: It makes sense for now since Analytics and Insights are currently in head. Is redesign optimized to have all these as bottom scripts? Once we FI to redesign we can probably refactor to use a bottom script and display banner on page load.

src/NuGetGallery.Core/Diagnostics/IDiagnosticsSource.cs
chenriksson: This class was moved as-is from Gallery->Gallery.Core to extend diagnostics to Shims. Shims implements a logger interface provided by the external service which allows us to hook into their tracing.

Per our discussion, I'll add an Exception API here so that exceptions can be written to AppInsights. I can log an issue in case we want to clean up Gallery diagnostics/telemetry/etc in the future.

src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
scottbommarito: `await Task.FromResult(0)` is what I see everywhere

src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
scottbommarito: `nameof(CookieComplianceServiceBase)`?

chenriksson: nameof won't reflect the child type. Going with GetType().Name.

src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -128,30 +129,34 @@
...
     if (!string.IsNullOrEmpty(iKey))
     {
-        <!-- Telemetry -->
-        <script type="text/javascript">
-            var appInsights = window.appInsights || function (config) {
-                function s(config) {
-                    t[config] = function () {
-                        var i = arguments;
-                        t.queue.push(function () { t[config].apply(t, i) })
+        var cookies = DependencyResolver.Current.GetService<ICookieComplianceService>();
```
shishirx34: would `cookieService` be more appropriate name?

src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
joelverhagen: Yeah, `nameof` is compile-time. `GetType()` is run-time.

src/NuGetGallery/Views/Shared/Layout.cshtml
```diff
@@ -28,6 +28,16 @@
         @ViewHelpers.InstrumentationScript()
     </head>
     <body class="s-noclickonce">
+        @if (CookieConsentMessage != null)
+        {
+            <div id="service-alert">
```
joelverhagen: spoke offline, this should have `role="alert"` (default implementation has it.

Issue #4426, opened 2017-07-19
[Redesign] Avatars are not HiDPI


NickCraver:
A simple fix here, but a carry over from the old design. With so much else looking crisp at HiDPI, it seems like a shame not to fix the avatars!

User avatars from gravatar are `?s=32` in a 32px box, which doesn't look good on a HiDPI display. If we changed these URLs to `?s=64` instead, you get a HiDPI image and a crisp look. Example difference:

Current (?s=32):
<img width="318" alt="screen shot 2017-07-19 at 13 39 47" src="https://user-images.githubusercontent.com/454813/28381152-cf6c7ec2-6c87-11e7-85b9-1d0059aadf14.png">

HiDPI (?s=64):
<img width="292" alt="screen shot 2017-07-19 at 13 40 03" src="https://user-images.githubusercontent.com/454813/28381166-d61e4bf6-6c87-11e7-9e3c-6d46cd1bb6d1.png">

/cc @joelverhagen

joelverhagen:
Resolved. Not yet deployed.

PR #4429, opened 2017-07-19
[Redesign] Clean up Gravatar code and load images twice the element size (for high DPI)

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/4426

Before:
image

After:
image

### DIFF
--- a/src/NuGetGallery/App_Code/ViewHelpers.cshtml
+++ b/src/NuGetGallery/App_Code/ViewHelpers.cshtml
@@ -161,32 +161,6 @@
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -161,32 +161,6 @@
     }
 }
 
-@helper OwnersGravatar(IEnumerable<User> owners, int size, UrlHelper url, bool showName = true)
```
joelverhagen: Dead code.

PR #4434, opened 2017-07-20
[Redesign] Make package ID more prominent than package title (but still keep title visible)

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/4413.

Top of package details page:
image

Search result:
image

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -1,7 +1,7 @@
 @model DisplayPackageViewModel
 
 @{
-    ViewBag.Title = Model.Title + " " + Model.Version;
+    ViewBag.Title = Model.Id + " " + Model.Version;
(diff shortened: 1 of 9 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -15,7 +15,7 @@
     <meta name="twitter:card" content="summary">
     <meta name="twitter:site" content="@("@nuget")">
 
-    <meta property="og:title" content="@Model.Title" />
+    <meta property="og:title" content="@Model.Id" />
```
loic-sharma: Should this have the version too?

joelverhagen: Yes, I think it should.

joelverhagen: Fixed.

Issue #4441, opened 2017-07-20
[Redesign] text inputs that are expected to be filled with email address should have type='email' instead of text


agr:
Example, the [Register form](https://preview.nuget.org/users/account/SignUp). Email input has type `text`, should be `email`.
This would aid with mobile browsing.

joelverhagen:
Fixed and deployed to https://preview.nuget.org/.

Issue #4440, opened 2017-07-20
[Redesign] Account page form data floats to top


shishirx34:
The Account page in new redesign shows the data floating to the top against the title. It should be centered.

![image](https://user-images.githubusercontent.com/1646506/28434033-d3f9e6d0-6d42-11e7-9be9-115277a2252e.png)

![image](https://user-images.githubusercontent.com/1646506/28434050-dc8151e4-6d42-11e7-8c16-3fd1ea4d217f.png)

joelverhagen:
Fixed and deployed to https://preview.nuget.org/.

Issue #4432, opened 2017-07-19
[Redesign] Small feedbacks from Twitter and offline


joelverhagen:
From https://twitter.com/shanselman/status/887795754021957633, offline with @unniravindranathan and @jonwchu, etc.

- [x] stats page
  - [x] "Client Version" -> NuGet Client Version"
  - [x] reverse order of client versions
- [x] footer
  - [x] headings to h3
  - [x] first row 1em
  - [x] second row is .85em
- [x] packages list
  - [x] package title height 0.9
  - [x] bold downloads: weight 500
  - [x] 5px at the top of package-details, not 10
- [x] jumbotron
  - [x] too much top padding: 5px
  - [x] check mobile too
    - [x] note there are two settings for padding-top (based on screen width)
- [x] inconsistent pagination padding
  - [x] https://twitter.com/chad_ramos/status/887804942903382017
  - [x] clear the float?

NickCraver:
Just a small tip: you can format with "- [ ] Item" for checkboxes in issues, ideal use case! Like this:

- [x] Item 1
- [ ] Item 2
  - [x] Subitem A
  - [ ] Subitem B

joelverhagen:
Fixed and deployed to https://preview.nuget.org/.

Issue #4419, opened 2017-07-19
[Redesign] Packages page is not mobile friendly.


ChrisMaddock:
The new NuGet.org looks great - big improvement!

The site seems to be mostly mobile friendly, with the exception of the package index page (/packages) which looks like this. I assume this is just an oversight.

ChrisMaddock:
![](http://i.imgur.com/yHBTswB.jpg)

I've tried this in both Chrome and Safari on iOS 10.

joelverhagen:
Resolved and deployed to https://preview.nuget.org/.

ChrisMaddock:
Great - thanks!

PR #4442, opened 2017-07-20
[Redesign] Address various points of feedback from the community

joelverhagen:
https://github.com/NuGet/NuGetGallery/issues/4432
https://github.com/NuGet/NuGetGallery/issues/4419
https://github.com/NuGet/NuGetGallery/issues/4440
https://github.com/NuGet/NuGetGallery/issues/4441
https://github.com/NuGet/NuGetGallery/issues/4443

Observe the bugs and each commit separately.

### DIFF
--- a/src/Bootstrap/less/theme/common-list-packages.less
+++ b/src/Bootstrap/less/theme/common-list-packages.less
@@ -12,6 +12,8 @@
         font-size: @font-size-h2;
         font-weight: 300;
         line-height: 0.9;
+        word-break: break-all;
+        word-break: break-word;
       }
 
       .package-by {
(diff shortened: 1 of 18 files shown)


joelverhagen:
🔔

src/Bootstrap/less/theme/common-list-packages.less
```diff
@@ -12,6 +12,8 @@
         font-size: @font-size-h2;
         font-weight: 300;
         line-height: 0.9;
+        word-break: break-all;
```
scottbommarito: why do we have the same property twice, set to different values?

joelverhagen: `break-all` is supported on more browsers than `break-word`.

joelverhagen: `break-word` will override `break-all` if the browser understands it.

PR #4089, opened 2017-06-09
Update Semver2 package details message (add Rider)

AndreyAkinshin:

PR #4448, opened 2017-07-21
Update Semver2 package details message (add Paket)

forki:


gep13:
While in principle I have no objection with this PR (I think it's addition would be a welcome hat-tip to Paket) I do think that listing Paket before the NuGet CLI is not helping with it's possibility for getting merged.

abdullin:
@gep13 are there any guidelines defining order in which open source NuGet clients have to be featured?

gep13:
>@abdullin said...
are there any guidelines defining order in which open source NuGet clients have to be featured?

I can say with certainty, even though I haven't checked, that there are no such guidelines.  However, given that the documentation in question is hosted on `nuget.org`, I would regard it as common decency to list the NuGet CLI as the first in the list of possible CLI tools to consume packages from the website.  Although, this in just my opinion you understand...

cdrnet:
The one which is currently, both with and without this PR, listed first is Visual Studio, not NuGet CLI. Both VisualStudio and Paket can be considered as third party clients (with different governance, ownership than NuGet). What makes them so fundamentally different?

NickCraver:
> Both Visual Studio and Paket can be considered as third party clients

This is not correct. The NuGet clients in Visual Studio and the CLI itself are both maintained by the NuGet team and supported (this term has significant meaning). The order of any lists like this is priority users are expected to try them in, the same as from highest to lowest volume. If a user is on NuGet.org, they are far more likely to have the NuGet CLI than they are to have the Paket CLI, that's just the install base reality by a huge margin.

One can argue the order doesn't matter, in which case they shouldn't mind changing it when shown it does. Or they can argue it does matter, which is an issue against the PR as-is as @gep13 says. But one can't argue it both ways, they have to pick one.

I personally have no objection to Paket being included, and am assuming the best intentions here with naivety in ordering. IMO, fixing it would show a willingness to work with each other.

gep13:
>@NickCraver said...
I personally have no objection to Paket being included, and am assuming the best intentions here with naivety in ordering. IMO, fixing it would show a willingness to work with each other.

Couldn't agree more.

From my own point of view (and again that is all that it is) I took the current wording which is:

`This package is only available for download with Visual Studio 2017 (version 15.3) and above or with NuGet client 4.3.0 and above.`

To mean:

`This package is only available for download <from these IDE's> or with <these CLI tools>`

As a result, this PR (https://github.com/NuGet/NuGetGallery/pull/4089) which extends the list of IDE's that can be used made sense, preserving the order as suggested by @NickCraver.

Should these two PR's be merged, I would suggest that the end result should be something like:

`This package is only available for download with Visual Studio 2017 (version 15.3) and above, or JetBrains Rider, or with NuGet client 4.3.0 and above or Paket 5.0 and above.`

abdullin:
@gep13 To make is easier to accept the final PR it could even be reworded as:

`This package is only available for download with Visual Studio 2017 (version 15.3) and above, or with NuGet client 4.3.0 and above, or JetBrains Rider or Paket 5.0 and above.`

This properly highlights VS & NuGet as the primary tools, while still indicating compatibility for the other tools used by the community.

gep13:
@abdullin I can't argue with that.

forki:
No order intended. Just copied what the rider guys did in #4089. Order now changed

drewgillies:
Thanks a bunch for submitting this PR, @forki --I'm spinning up an internal conversation here so we can figure out next steps.

jcjiang:
Took a look through and I have no objections.

drewgillies:
@forki --okay, let's get this one merged. There are a couple things that need to happen before we proceed. I checked our readme and they look documented enough to me, but I'd like your feedback--the readme is written for you, not those of us on the NuGet team. If these weren't clear enough I'd like to update our readme. We need 2 things to happen for our internal tracking and deployment to have all of the boxes checked:
- We need you to file (or identify an existing) issue to connect this to. Please file the issue and connect this PR to it using the Connect Issue button.
- This change needs to be based on our `dev` branch rather than our `master` branch. So, please rebase this on `dev` and retarget the PR using the Edit button (it may be automatic after the rebase and push).
Thanks again! We appreciate your contribution and want to get it in the product. Let me know if any of this needs clarifying, and again, please let me know if our readme could be better.

joelverhagen:
_(copied from a similar PR, to improve visibility: https://github.com/NuGet/NuGetGallery/pull/4089#issuecomment-816893151)_

Hey folks, this SemVer 2.0.0 warning banner was removed a while back. From our data we found that over 95% of clients support SemVer 2.0.0 parsing and the SemVer 2.0.0 V2 and V3 protocols. This meant the warning banner was more noisy that helpful for the majority of users.

Context: https://github.com/NuGet/NuGetGallery/issues/7836
Client context: https://github.com/NuGet/Home/issues/5201
NuGet.org change: https://github.com/NuGet/NuGetGallery/pull/7837

Looks like Paket is mentioned in the docs:
https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#semantic-versioning-200

Do you feel that a further change is necessary at this point?

Apologies on the extensive delay.

drewgillies:
@joelverhagen ignore my confused emoji above--mouse slip when doing a +1, and apparently there's no way to un-emoji.

joelverhagen:
Hey all, I'm going to close this PR per https://github.com/NuGet/NuGetGallery/issues/4448#issuecomment-816894162. If you feel that further changes are still required, please feel free to open a new issue! Thanks!

Issue #4424, opened 2017-07-19
[Redesign] Relative dates for "a year ago" aren't as useful as previous format


NickCraver:
The version history date format should revert back, it's far less usable now. As an example, here's Dapper:
<img width="808" alt="screen shot 2017-07-19 at 13 23 38" src="https://user-images.githubusercontent.com/454813/28380527-a47a93ea-6c85-11e7-8774-eb6455f15db1.png">

Note the old date format:

<img width="601" alt="screen shot 2017-07-19 at 13 23 22" src="https://user-images.githubusercontent.com/454813/28380538-ab1ca60c-6c85-11e7-93dd-e784d8504ca3.png">

I propose that "a year ago" should never be a string that's shown, it should revert to the date at that point. Likely anything over a short duration of relative time should show the full date. Otherwise we get lists of "a year ago" all over the place and everyone has to hover to get the data they want.

We do something similar on Stack Overflow, we never present over "2 days ago", a date is shown after that. The English-only algorithm for relative dates we use is:

```c#
private static string ToRelativeTimePast(DateTime dt, DateTime utcNow, bool includeTime = true)
{
    var ts = utcNow - dt;
    var delta = ts.TotalSeconds;

    if (delta < 1) return "just now";
    if (delta < 60) return ts.Seconds == 1 ? "1 sec ago" : ts.Seconds.ToString() + " secs ago";
    if (delta < 3600 /*60 mins * 60 sec*/) return ts.Minutes == 1 ? "1 min ago" : ts.Minutes.ToString() + " mins ago";
    if (delta < 86400 /*24 hrs * 60 mins * 60 sec*/) return ts.Hours == 1 ? "1 hour ago" : ts.Hours.ToString() + " hours ago";

    var days = ts.Days;
    if (days == 1) return "yesterday";
    if (days <= 2) return days.ToString() + " days ago";
    if (utcNow.Year == dt.Year) return dt.ToString(includeTime ? "MMM %d 'at' %H:mmm" : "MMM %d");
    return dt.ToString(includeTime ? @"MMM %d \'yy 'at' %H:mmm" : @"MMM %d \'yy");
}
```

joelverhagen:
I talked to @jonwchu about this and we agree to lean on moment.js's thresholds. When the units turn to years, we will start showing the date as "July 19, 2017".

joelverhagen:
Resolved but not yet deployed.

Issue #4280, opened 2017-06-24
File upload date is UTC


LarryEllis:
The date that renders on the package details is based on UTC.   Should this be converted in the UI to local time, or is there a way to do so?

I have not been able to find a timezone configuration, if one is needed.

skofman1:
@joelverhagen , does this repro on the redesigned page?

joelverhagen:
Potentially related: https://github.com/NuGet/NuGetGallery/issues/4084

joelverhagen:
Are you looking at https://www.nuget.org for this? If so, all of the timestamp we store internally should be UTC. It's a view or display concern of what timezone it should be converted to for human readability. In our website redesign (currently in progress), these timestamps are converted to browser local time using moment.js. 

Could you take a look at our preview site and let me know if it looks better for you?
https://preview.nuget.org/packages/Newtonsoft.Json/

Also, could you provide a screenshot or specific example of this?

joelverhagen:
This is resolved in the redesign. 

The server sends down timestamps in UTC. Then the client (browser) converts these UTC timestamps to relative times (11 hours ago) with respect to the browser's local time.

ctolkien:
![image](https://user-images.githubusercontent.com/515955/28394404-3696d98c-6d30-11e7-9745-72791ee43a61.png)

With the current preview build - this is not correct. I just pushed a new package live and it's out by 10 hours (which is our UTC offset).

joelverhagen:
I'll take another look -- thanks for the heads up!

joelverhagen:
Resolved but not yet deployed.

PR #4457, opened 2017-07-22
[Redesign] Assume dates are UTC (even though their DateTimeKind is Unspecified)

joelverhagen:
Also, never show relative years -- show full dates at that point

Fix https://github.com/NuGet/NuGetGallery/issues/4280
Fix https://github.com/NuGet/NuGetGallery/issues/4424

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/common.js
+++ b/src/NuGetGallery/Scripts/gallery/common.js
@@ -58,7 +58,7 @@
         unparsedValue = ('' + unparsedValue).replace(/,/g, '');
         var parsedValue = parseInt(unparsedValue);
         return parsedValue;
-    }
+    };
 
     // source: http://stackoverflow.com/questions/400212/how-do-i-copy-to-the-clipboard-in-javascript
     // enhancement with special case for IEs, otherwise the temp textarea will be visible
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -200,7 +200,7 @@
 
         var allButLast = items.slice(0, -1).join(', ');
         var last = items.slice(-1)[0];
-        return [allButLast, last].join(items.length < 2 ? '' : (items.length == 2 ? ' and ' : ', and '));
+        return [allButLast, last].join(items.length < 2 ? '' : items.length === 2 ? ' and ' : ', and ');
```
skofman1: === ?

joelverhagen: `===` matches type in addition to what `==` does. Equal value and equal type. It's a best practice.

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -126,13 +126,13 @@ private static string GetClientVersion()
 
         private static string GetAccountCreationDate(User user)
         {
-            return user.CreatedUtc != null ? user.CreatedUtc.Value.ToString("o") : "N/A";
+            return user.CreatedUtc != null ? user.CreatedUtc.Value.ToString("O") : "N/A";
```
loic-sharma: You can use `user.CreatedUtc?.Value.ToString("O")  ?? "N/A"` to be consistent with line 135.

joelverhagen: Good idea.

joelverhagen: Fixed.

PR #4460, opened 2017-07-24
Merge latest dev in feature-redesign

joelverhagen:


src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -22,6 +22,21 @@
     </div>
 }
 
+@if (CookieConsentMessage != null)
+{
+    <div class="container-fluid banner banner-notice text-center">
```
chenriksson: This is the new layout for redesign? Is role=alert getting applied?

joelverhagen: > This is the new layout for redesign?

Yes

>  Is role=alert getting applied?

Not yet. Good catch!!

joelverhagen: Fixed.

PR #4461, opened 2017-07-24
Fix password autocomplete

skofman1:
Fixes this: https://github.com/NuGet/Engineering/issues/622

### DIFF
--- a/src/NuGetGallery/Views/Users/Account.cshtml
+++ b/src/NuGetGallery/Views/Users/Account.cshtml
@@ -319,12 +319,12 @@
                              @Html.ValidationSummaryFor("ChangePassword")
                              <div class="form-field">
                                  @Html.LabelFor(m => m.ChangePassword.OldPassword)
-                                 @Html.PasswordFor(m => m.ChangePassword.OldPassword)
+                                 @Html.PasswordFor(m => m.ChangePassword.OldPassword, new { autocomplete = "off" })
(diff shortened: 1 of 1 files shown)


skofman1:
@blowdart , looks good?

joelverhagen:
What does the experience in the browser looks like when this attribute is not applied (i.e. current state in PROD)?

skofman1:
Chrome for example lets you choose which password to apply in the field

skofman1:
Created a bug Joel: https://github.com/NuGet/Engineering/issues/623

joelverhagen:
@skofman1 thanks!!

src/NuGetGallery/Views/Users/Account.cshtml
```diff
@@ -319,12 +319,12 @@
                              @Html.ValidationSummaryFor("ChangePassword")
                              <div class="form-field">
                                  @Html.LabelFor(m => m.ChangePassword.OldPassword)
-                                 @Html.PasswordFor(m => m.ChangePassword.OldPassword)
+                                 @Html.PasswordFor(m => m.ChangePassword.OldPassword, new { autocomplete = "off" })
```
xavierdecoster: You sure this works? AFAIK, browsers are disregarding `autocomplete=off` on password fields (IE11, FF, Chrome).

Issue #4450, opened 2017-07-21
[Redesign] Stats page breaks when the thousands seperator is not a comma


joelverhagen:
Steps:

1. Open Chrome.
1. Change the Chrome language to German (`de-DE`): https://support.google.com/chrome/answer/173424?co=GENIE.Platform%3DDesktop&hl=en
1. Go to https://preview.nuget.org/stats

![image](https://user-images.githubusercontent.com/94054/28476341-9379e728-6e04-11e7-8da6-030f22ff4b11.png)

joelverhagen:
If a number is sent to client via JSON, it should be a number, not a string. When displaying the number to the user, we should use `.toLocaleString()` on the client side and `.ToNuGetNumberString()` on the server side.

PR #4468, opened 2017-07-26
[Redesign]Make svg render culture invariant.

ryuyu:
Addresses issue https://github.com/NuGet/NuGetGallery/issues/4450
The problem was that when the svg was being generated, the floating point numbers were being rendered with commas as decimal separators, which SVG render didn't understand.

Fix was to make this part of rendering culture invariant.

@joelverhagen

### DIFF
--- a/src/NuGetGallery/Views/Statistics/Index.cshtml
+++ b/src/NuGetGallery/Views/Statistics/Index.cshtml
@@ -1,4 +1,5 @@
 ﻿@model StatisticsPackagesViewModel
+@using System.Globalization
 @{
     ViewBag.Title = "Statistics";
     ViewBag.Tab = "Statistics";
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Statistics/Index.cshtml
```diff
@@ -112,11 +113,11 @@
                                 var fraction = chartFullWidthInPercent / showVersions.Count();
                                 foreach (var item in showVersions)
                                 {
-                                    <rect class="graph-bar" id="graph-bar-@(item.Version)" x="@(widthOffsetPercent + index * fraction)%" y="@(chartFullHeightInPercent-item.Downloads/max*chartFullHeightInPercent)%" width="@(fraction * (1.0 - gapPercent))%" height="@(item.Downloads/max*chartFullHeightInPercent)%">
+                                    <rect class="graph-bar" id="graph-bar-@(item.Version)" x="@((widthOffsetPercent + index * fraction).ToString(CultureInfo.InvariantCulture))%" y="@((chartFullHeightInPercent-item.Downloads/max*chartFullHeightInPercent).ToString(CultureInfo.InvariantCulture))%" width="@((fraction * (1.0 - gapPercent)).ToString(CultureInfo.InvariantCulture))%" height="@((item.Downloads/max*chartFullHeightInPercent).ToString(CultureInfo.InvariantCulture))%">
```
joelverhagen: Can we an `@function` for this like `@Format(...)`? This is a view concern and is pretty verbose right now.

Issue #4458, opened 2017-07-22
[Redesign] Focus search element immediately when landing on the front page


maartenba:
It would be nice for the search box to be focused immediately when visiting the new NuGet.org website, much like on NuGet.org vCurrent. This would mean people are able to just start typing, hit enter, find results. No need to move the mouse and click the search box.

joelverhagen:
We need to check about the accessibility guidance of this (/cc @microsoftsam), but I like the idea.

NickCraver:
If not on the homepage, it really should still happen on `/packages` IMO.

microsoftsam:
I think a better way to solve this is to provide a mechanism to bypass blocks, since you have content that's repeated on multiple pages (each page has a search box, each page has the menu, and each page has the content section).

This is the "hidden menu" concept we've previously talked about. A menu when focused provides the user the ability to "Jump to Search", or "Jump to Content", or "Change language", etc. 

So I think focusing the search box should only be done on pages where that is the likely first intractable element for a user.

MAS Standard in question here: WCAG 2.0 2.4.1 Bypass Blocks

maartenba:
Any chance a "focus the element when user starts typing" could be added within those rules?

microsoftsam:
^^ That's what github does. You'd want to make sure you'd on no activeElement in your handler, but that could work.

joelverhagen:
Fixed and deployed to https://preview.nuget.org/.

PR #4469, opened 2017-07-26
[Redesign]Autofocus search

ryuyu:
If no element has focus (or body has focus), typing will jump to search box.
As suggested in https://github.com/NuGet/NuGetGallery/issues/4458

@joelverhagen @microsoftsam

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/common.js
+++ b/src/NuGetGallery/Scripts/gallery/common.js
@@ -312,9 +312,20 @@ $(function () {
 
     $(document).on('keydown', function (e) {
         var code = (e.keyCode || e.which);
+        var isValidInputCharacter =
+            ((code >= 48 && code <= 57)           // numbers 0-9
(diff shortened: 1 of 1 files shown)


microsoftsam:
Nice!

src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -309,4 +309,12 @@ $(function () {
     $(document).on('click', '.dropdown-menu', function (e) {
         e.stopPropagation();
     });
+
+    $(document).on('keydown', function (e) {
+        var code = (e.keyCode || e.which);
+
```
joelverhagen: Please comment about the `code != 9` part.

joelverhagen: Code comment -- that's Tab.

joelverhagen: Should we make this more generic? Perhaps the key should be alpha + numbers + symbols?

PR #4489, opened 2017-07-31
[Redesign] Address three points of feedback from the survey

joelverhagen:
- Remove transition CSS from .collapsing class since the transition time is 0 in the JS
  - This was causing weirdness in Edge.
- Automatically expand the version list if the selected version is in the collapsed section
- Fix ctrl + click on data-track links

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -28,6 +28,15 @@
     }
 }
 
+@helper VersionListDivider(int rowCount, bool versionsExpanded)
+{
+    if (rowCount == Constants.VisibleVersions + 1)
+    {
+        @:</tbody><tbody class="no-border collapse @(versionsExpanded ? "in" : string.Empty)" id="hidden-versions">
+    }
+}
+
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -301,26 +302,31 @@
...
-                        @{var rowCount = 0;}
+                        @{
+                            var rowCount = 0;
+                            var versionsExpanded = Model.PackageVersions.Skip(VisibleVersions).Any(v => v.IsCurrent(Model));
+                        }
                         @foreach (var packageVersion in Model.PackageVersions)
                         {
                             if (!packageVersion.Deleted && (packageVersion.Listed || (Model.IsOwner(User) || User.IsAdministrator())))
                             {
                                 rowCount++;
```
scottbommarito: This and the `if` statement below are identical between both branches of the outer `if` statement. Could we use a helper function of some kind to minimize the code duplication?

joelverhagen: Fixed.

PR #4500, opened 2017-08-02
Ensure SSL v3 is disabled and that TLS v1.2 is enabled

loic-sharma:


loic-sharma:
Tested on dev with release ` 4.3.0-loshar-tls12-12456`, this passed all functional tests.

Issue #4492, opened 2017-08-02
[Redesign] Make links in the description and release notes fields clickable


joelverhagen:

anangaur:
https://github.com/NuGet/NuGetGallery/issues/4023

Issue #4491, opened 2017-08-02
[Redesign] Revisit font sizes on the redesign site


joelverhagen:
There was a lot of feedback on the font sizes on the preview site. Major points of feedback:

1. [x] ID size on the search results - too big
1. [x] Footer text - too small

anangaur:
Smaller font size for the Package Ids and separator line removed:
![image](https://user-images.githubusercontent.com/14800916/29131487-e2a2623a-7ce1-11e7-9b41-f1a16434dc2e.png)

PR #4503, opened 2017-08-03
[Redesign] Add "include prerelease" checkbox and polish search results

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/4491
Fix https://github.com/NuGet/NuGetGallery/issues/4492

screencapture-localhost-packages-1501777306383 1

### DIFF
--- a/src/NuGetGallery/OData/SearchService/SearchAdaptor.cs
+++ b/src/NuGetGallery/OData/SearchService/SearchAdaptor.cs
@@ -25,14 +25,14 @@ namespace NuGetGallery.OData
         /// </summary>
         internal const int MaxPageSize = 100;
 
-        public static SearchFilter GetSearchFilter(string q, int page, string sortOrder, string context, string semVerLevel)
+        public static SearchFilter GetSearchFilter(string q, int page, bool includePrerelease, string sortOrder, string context, string semVerLevel)
         {
             var searchFilter = new SearchFilter(context)
             {
                 SearchTerm = q,
                 Skip = (page - 1) * Constants.DefaultPackageListPageSize, // pages are 1-based. 
                 Take = Constants.DefaultPackageListPageSize,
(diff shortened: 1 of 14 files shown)


src/NuGetGallery/OData/SearchService/SearchAdaptor.cs
```diff
@@ -25,14 +25,14 @@ public static class SearchAdaptor
         /// </summary>
         internal const int MaxPageSize = 100;
 
-        public static SearchFilter GetSearchFilter(string q, int page, string sortOrder, string context, string semVerLevel)
+        public static SearchFilter GetSearchFilter(string q, int page, string sortOrder, string context, string semVerLevel, bool includePrerelease)
```
scottbommarito: nit: In most methods, `includePrerelease` comes before `semVerLevel` in the parameters.

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -51,4 +60,31 @@
...
     </div>
-</section>
\ No newline at end of file
+</section>
+
+@section bottomScripts {
+    <script type="text/javascript">
+        $(function () {
+            $("#include-prerelease").on('change', function () {
+                var parameters = {};
```
scottbommarito: Is there any way we could simplify this script?

It seems a lot for something that should be innately handled by the form.

joelverhagen: The idea is to create a minimal URL with the knowledge of the server side defaults. This cannot be achieved with a form.

src/NuGetGallery/OData/SearchService/SearchAdaptor.cs
joelverhagen: Sure.

PR #3050, opened 2016-05-28
Implement auto linking URLs on Package Details - Fix #67

robertmuehsig:
This PR should implement the (pretty old) feature idea from this issue https://github.com/NuGet/NuGetGallery/issues/67

I used a RegEx that seems to identify URLs fine and it wraps them in a a-tag. As far as I tested, there shouldn't be any XSS attacks possible, because the first step is to HtmlEncode the content, which is the same way as before.

Some screenshots from my tests (with modified packages - the real authors didn't inject the xss stuff ;))

image

image

image

image

The RegEx seems to be "clever" in the sense that only urls that start with http will be catched - so JSON.NET is just fine. 
If you need to test some other combinations: https://regex101.com/r/kI9xT9/2

PR #4505, opened 2017-08-03
[Redesign] Make links clickable in description and release notes.

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/4492

## Bootstrap:
image

## jQuery
image

### DIFF
--- a/src/NuGetGallery/Helpers/HtmlExtensions.cs
+++ b/src/NuGetGallery/Helpers/HtmlExtensions.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Text.RegularExpressions;
 using System.Web;
 using System.Web.Mvc;
 using System.Web.Mvc.Html;
@@ -52,10 +53,28 @@ namespace NuGetGallery.Helpers
 
         public static IHtmlString PreFormattedText(this HtmlHelper self, string text)
         {
-            return self.Raw(HttpUtility
-                .HtmlEncode(text)
-                .Replace("\n", "<br />")
-                .Replace("  ", "&nbsp; "));
+            // Encode HTML entities. Important! Security!
+            var encodedText = HttpUtility.HtmlEncode(text);
+
+            // Turn HTTP and HTTPS URLs into links.
+            // Source: https://stackoverflow.com/a/4750468
+            encodedText = RegexEx.TryReplaceWithTimeout(
+                encodedText,
+                @"((http|https):\/\/[\w\-_]+(\.[\w\-_]+)+([\w\-\.,@?^=%&amp;:/~\+#]*[\w\-\@?^=%&amp;/~\+#])?)",
+                match => $"<a href=\"{match.Value}\">{match.Value}</a>",
+                RegexOptions.IgnoreCase);
+
+            // Replace new lines with the <br /> tag.
+            encodedText = encodedText.Replace("\n", "<br />");
+
+            // Replace more than one space in a row with a space then &nbsp;.
+            encodedText = RegexEx.TryReplaceWithTimeout(
+                encodedText,
+                "  +",
+                match => " " + string.Join(string.Empty, Enumerable.Repeat("&nbsp;", match.Value.Length - 1)),
+                RegexOptions.None);
+
+            return self.Raw(encodedText);
         }
 
         public static IHtmlString ValidationSummaryFor(this HtmlHelper html, string key)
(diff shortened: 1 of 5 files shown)


src/NuGetGallery/Helpers/HtmlExtensions.cs
```diff
@@ -52,10 +53,44 @@ public static IHtmlString BreakWord(this HtmlHelper self, string text)
...
+            // Encode HTML entities. Important! Security!
+            text = HttpUtility.HtmlEncode(text);
+
+            // Turn HTTP and HTTPS URLs into links.
+            try
+            {
+                // Source: https://stackoverflow.com/a/4750468
+                text = Regex.Replace(
+                    text,
+                    @"https?://[\w\-_]+(\.[\w\-_]+)+([\w\-\.,@?^=%&amp;:/~\+#]*[\w\-\@?^=%&amp;/~\+#])?",
```
chenriksson: Should it be const, and should we use the same regex for project/icon URLs? (see Constants.UrlValidationRegEx)

Maybe there should be helper for the try/catch of the RegEx with const timeout?

src/NuGetGallery/Helpers/HtmlExtensions.cs
```diff
@@ -52,10 +53,44 @@ public static IHtmlString BreakWord(this HtmlHelper self, string text)
 
         public static IHtmlString PreFormattedText(this HtmlHelper self, string text)
         {
-            return self.Raw(HttpUtility
-                .HtmlEncode(text)
-                .Replace("\n", "<br />")
-                .Replace("  ", "&nbsp; "));
+            // Encode HTML entities. Important! Security!
+            text = HttpUtility.HtmlEncode(text);
```
chenriksson: nit: maybe store in separate local from original, unencoded text; for debugging and to ensure this line isn't accidentally omitted.

src/NuGetGallery/Helpers/HtmlExtensions.cs
```diff
@@ -52,10 +53,44 @@ public static IHtmlString BreakWord(this HtmlHelper self, string text)
...
+            // Replace new lines with the <br /> tag.
+            text = text.Replace("\n", "<br />");
+
+            // Replace more than one space in a row with a space then &nbsp;
+            try
+            {
+                text = Regex.Replace(
+                    text,
+                    "  +",
+                    match => " " + string.Join(string.Empty, Enumerable.Repeat("&nbsp;", match.Value.Length - 1)),
```
chenriksson: Doesn't the original logic achieve the same without a regex? Why is regex better?
.Replace("  ", "&nbsp; "));

src/NuGetGallery/Helpers/HtmlExtensions.cs
joelverhagen: Sounds good.

src/NuGetGallery/Helpers/HtmlExtensions.cs
joelverhagen: Dots instead of spaces.
Old: `.....` -> `&nbsp;.&nbsp;..`
New: `.....` -> `.&nbsp;&nbsp;&nbsp;&nbsp;`

src/NuGetGallery/Helpers/HtmlExtensions.cs
joelverhagen: I'll take a look at this. Thanks.

src/NuGetGallery/Helpers/HtmlExtensions.cs
chenriksson: Ah, thought old might account for the appended space for the next match. i.e.:
`.....` -> `&nbsp;&nbsp;&nbsp;&nbsp;.`

src/NuGetGallery/Helpers/HtmlExtensions.cs
joelverhagen: ` Constants.UrlValidationRegEx` is much more permissive. This is existing behavior for edit so I am hesitant to change that.

The regexes have similar characteristics but are different in key ways:

1. `UrlValidationRegEx` is anchored to the end of the string (via `$`), this is not appropriate for this case.
1. This new regex is lazy with its last group meaning that trailing dots will not be matched (for example). This makes sense only when the URL is in a longer string with sentence/paragraphs etc (vs. a field that _just_ has the URL).

joelverhagen: > Maybe there should be helper for the try/catch of the RegEx with const timeout?
Done.

robertmuehsig:
He @joelverhagen I just discovered this feature and I really like it. I tried more or less the same with a PR last year, which was closed due to XSS security considerations: https://github.com/NuGet/NuGetGallery/pull/3050
Just to make sure it's "safer" this time 👍

PR #4512, opened 2017-08-05
Add page to view ownership requests 

scottbommarito:
(Quick weekend project I did.)

Currently, there is no way for the pending owner of a package ownership request to view and accept the request outside of the email that is sent. As a result, if the email fails to send, the pending owner will never know that they were ever requested and--even if they do know they were requested--cannot accept the request!

This PR adds a page for users to manage their package ownership requests. It shows both incoming and outgoing requests. From this page, users can accept and reject requests for them to become owners of a package and can cancel requests they had made.

image

The link to this page is in the dropdown menu at the top of the page.

image

Will tidy up and add some tests later.


anangaur:
This is really cool. A big +1 from my side :) A couple of things:
1. I think this option should move being a separate option in the profile drop down to something under "Manage packages"
2. I would call the incoming/outgoing titles to:

- Pending ownership requests (incoming)
- Pending ownership requests (outgoing)
3. These should only show up when there is one or more pending requests.

Can you file an issue for the same?

src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace NuGetGallery
+{
+    public class OwnerRequestsListItemViewModel
+    {
+        public PackageOwnerRequest Request { get; private set; }
+
+        public Package Package { get; private set; }
+
+        public OwnerRequestsListItemViewModel(PackageOwnerRequest request, IPackageService packageService)
+            : this(request, packageService.FindPackageByIdAndVersion(request.PackageRegistration.Id, null, SemVerLevelKey.SemVer2, true))
```
loic-sharma: This is a lot of code to stash away in the `this` call. Could you instead move the `FindPackageByIdAndVersion` inside of the code block itself and used named parameters for the `null` and `true`?

loic-sharma: If the second constructor isn't used, simply merge into one constructor.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1078,7 +1081,19 @@ public virtual ActionResult Edit(string id, string version)
 
         [Authorize]
         [RequiresAccountConfirmation("accept ownership of a package")]
-        public virtual async Task<ActionResult> ConfirmOwner(string id, string username, string token)
+        public virtual Task<ActionResult> ConfirmOwner(string id, string username, string token)
+        {
+            return HandleOwnershipRequest(id, username, token, true);
```
loic-sharma: For `HandleOwnershipRequest`, can you use a named parameter for boolean so it's clear what it's for? Same thing with line 1093.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1107,13 +1122,56 @@ public virtual ActionResult Edit(string id, string version)
...
+
+                await _packageService.AddPackageOwnerAsync(package, user);
+
+                SendAddPackageOwnerNotification(package, user, result.Item1, result.Item2);
+
+                return View(new PackageOwnerConfirmationModel(id, username, ConfirmOwnershipResult.Success));
+            }
+            else
+            {
+                await _packageService.RemovePackageOwnerAsync(package, user);
```
loic-sharma: Should we have a notification there, similar to `SendAddPackageOwnerNotification` but for the denied case?

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -125,6 +125,7 @@
                                     <li class="divider"></li>
                                     <li role="presentation"><a href="@Url.ManageMyPackages()" role="menuitem">Manage Packages</a></li>
                                     <li role="presentation"><a href="@Url.UploadPackage()" role="menuitem">Upload Package</a></li>
+                                    <li role="presentation"><a href="@Url.OwnerRequests()" role="menuitem">Package Owner Requests</a></li>
```
loic-sharma: What do you think about `Package Ownership Requests` instead? Is that too verbose?

scottbommarito: @anangaur wants to merge the page with "Manage Packages", so it won't appear in this list anymore

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I'm not sure. Currently when you remove a pending owner via the `Manage Package Owners` page it doesn't send a mail. If we add a notification here, we should also make sure to add the notification there.

scottbommarito:
@anangaur I made the UI changes requested!

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1078,7 +1081,19 @@ public virtual ActionResult Edit(string id, string version)
 
         [Authorize]
         [RequiresAccountConfirmation("accept ownership of a package")]
-        public virtual async Task<ActionResult> ConfirmOwner(string id, string username, string token)
+        public virtual Task<ActionResult> ConfirmOwner(string id, string username, string token)
```
skofman1: anti forgery token

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -458,6 +458,11 @@ where VersionRange.Parse(d.VersionSpec).Satisfies(packageVersion)
                 new PackageRegistrationAuditRecord(package, AuditedPackageRegistrationAction.RemoveOwner, user.Username));
         }
 
+        public PackageOwnerRequest GetPackageOwnerRequestAsync(PackageRegistration package, User requestingUser, User pendingUser)
```
skofman1: this method is not async but the name contains Async

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -916,10 +921,20 @@ private static Package FindPackage(IEnumerable<Package> packages, Func<Package,
             return packages.First(pv => pv.Version.Equals(version.OriginalVersion, StringComparison.OrdinalIgnoreCase));
         }
 
-        private PackageOwnerRequest FindExistingPackageOwnerRequest(PackageRegistration package, User pendingOwner)
+        private PackageOwnerRequest FindExistingPackageOwnerRequestByPending(PackageRegistration package, User pendingOwner)
```
skofman1: this name is confusing. Consider: FindPendingOwnerRequestsByPackageAndOwner

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1107,13 +1122,56 @@ public virtual ActionResult Edit(string id, string version)
...
+            {
+                await _packageService.RemovePackageOwnerAsync(package, user);
+
+                return View("ConfirmOwner", new PackageOwnerConfirmationModel(id, username, ConfirmOwnershipResult.Rejected));
+            }
+        }
+
+        [Authorize]
+        [RequiresAccountConfirmation("cancel ownership of a package")]
+        public virtual async Task<ActionResult> CancelOwner(string id, string requestingUsername, string pendingUsername)
```
skofman1: This name is confusing.. how about CancelPendingOwnershipRequest

src/NuGetGallery/Services/IPackageService.cs
```diff
@@ -79,6 +79,8 @@ public interface IPackageService
 
         Task RemovePackageOwnerAsync(PackageRegistration package, User user);
 
+        PackageOwnerRequest GetPackageOwnerRequestAsync(PackageRegistration package, User requestingUser, User pendingUser);
```
skofman1: this method is not Async..

src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
```
skofman1: file header

skofman1:
> using System;

file header

---
Refers to: src/NuGetGallery/ViewModels/OwnerRequestsListViewModel.cs:1 in ad5e9ea. [](commit_id = ad5e9eab351cb24e607868d18e83e9decc3af727, deletion_comment = False)

src/NuGetGallery/ViewModels/OwnerRequestsViewModel.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Collections.Generic;
+
+namespace NuGetGallery
+{
+    public class OwnerRequestsViewModel
+    {
+        public OwnerRequestsListViewModel Incoming { get; private set; }
```
skofman1: file header

src/NuGetGallery/ViewModels/OwnerRequestsViewModel.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        public OwnerRequestsListViewModel Outgoing { get; private set; }
+
+        public OwnerRequestsViewModel(IEnumerable<PackageOwnerRequest> incoming, IEnumerable<PackageOwnerRequest> outgoing, User currentUser, IPackageService packageService)
+            : this(new OwnerRequestsListViewModel(incoming, "Incoming", currentUser, packageService), new OwnerRequestsListViewModel(outgoing, "Outgoing", currentUser, packageService))
```
skofman1: >Incoming [](start = 61, length = 8)

const

skofman1: >Outgoing [](start = 144, length = 8)

const

src/NuGetGallery.Core/Entities/PackageOwnerRequest.cs
```diff
@@ -9,9 +9,10 @@ public class PackageOwnerRequest
         : IEntity
     {
         public int PackageRegistrationKey { get; set; }
+        public virtual PackageRegistration PackageRegistration { get; set; }
         public int NewOwnerKey { get; set; }
-        public User NewOwner { get; set; }
-        public User RequestingOwner { get; set; }
+        public virtual User NewOwner { get; set; }
```
skofman1: why is this virtual?

skofman1:
>         public int Key { get; set; }

UTs?

---
Refers to: src/NuGetGallery.Core/Entities/PackageOwnerRequest.cs:19 in ad5e9ea. [](commit_id = ad5e9eab351cb24e607868d18e83e9decc3af727, deletion_comment = False)

src/NuGetGallery/Services/PackageService.cs
loic-sharma: Is this method actually needed?

src/NuGetGallery/Services/PackageService.cs
loic-sharma: Or how about `FindPendingOwnerRequestsByPendingOwner`?

loic-sharma: Or `FindOwnershipRequestsbyPendingOwner`?

src/NuGetGallery/Views/Packages/ConfirmOwner.cshtml
```diff
@@ -28,6 +36,20 @@
...
+        {
+            <p class="text-center">
+                You have rejected a request to become an owner of the
+                '<strong><a href="@Url.Package(@Model.PackageId)" title="The @Model.PackageId package">@Model.PackageId</a></strong>' package!
+            </p>
+        }
+        else if (Model.Result == ConfirmOwnershipResult.Cancelled)
+        {
+            <p class="text-center">
+                You have cancelled your request for <strong>@Model.Username</strong> become an owner of the
```
loic-sharma: `You have cancelled your request for <strong>@Model.Username</strong> become an owner` -> `  You have cancelled your request for <strong>@Model.Username</strong> to become an owner`

loic-sharma: (I added a to)

src/NuGetGallery/Views/Users/_OwnerRequestsList.cshtml
```diff
@@ -0,0 +1,82 @@
...
+                                <td class="col-sm-2 text-right align-middle">
+                                    <div class="btn-group">
+                                        @{
+                                            if (CurrentUser.Key == requestItem.Request.NewOwnerKey)
+                                            {
+                                                <a href="@Url.ConfirmOwner(packageId, CurrentUser.Username, requestItem.Request.ConfirmationCode)" class="btn" title="Confirm Ownership" aria-label="Confirm Ownership of @packageId">
+                                                    <i class="ms-Icon ms-Icon--Accept" aria-hidden="true"></i>
+                                                </a>
+
+                                                <a href="@Url.RejectOwner(packageId, CurrentUser.Username, requestItem.Request.ConfirmationCode)" class="btn" title="Reject Ownership" aria-lable="Reject Ownership Request of @packageId">
```
loic-sharma: `aria-lable` -> `aria-label`. Same with line 68.

src/NuGetGallery.Core/Entities/PackageOwnerRequest.cs
scottbommarito: Required by EntityFramework for fetching the data from the database.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: done--also changed the other route names as well to match

src/NuGetGallery/Services/PackageService.cs
scottbommarito: `GetPendingOwnershipRequest` because it's not necessarily just "by pending".

src/NuGetGallery/Services/IPackageService.cs
scottbommarito: fixed

src/NuGetGallery/Services/PackageService.cs
scottbommarito: it is used by `CancelPendingOwnershipRequest`

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -29,6 +30,7 @@ public partial class UsersController
             ICuratedFeedService feedsQuery,
             IUserService userService,
             IPackageService packageService,
+            IEntityRepository<PackageOwnerRequest> packageOwnerRequestRepository,
```
joelverhagen: This is lifting data access layer (DAL) and presentation concerns. There should be a service that exposes the necessary operations on the `PackageOwnerRequest`. I see that `JsonApiController` already has this. It should be fixed too.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -154,10 +157,18 @@ public virtual ActionResult Packages()
             var user = GetCurrentUser();
             var packages = _packageService.FindPackagesByOwner(user, includeUnlisted: true)
                 .Select(p => new ListPackageItemViewModel(p)).OrderBy(p => p.Id).ToList();
+            
+            var requests = _packageOwnerRequestRepository.GetAll();
+            
+            var incoming = requests.Where(r => r.NewOwnerKey == user.Key);
+            var outgoing = requests.Where(r => r.RequestingOwnerKey == user.Key);
+
+            var ownerRequests = new OwnerRequestsViewModel(incoming, outgoing, user, _packageService);
```
joelverhagen: We should materialize the collection before passing it to the view model.

src/NuGetGallery/ViewModels/OwnerRequestsListViewModel.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
```
joelverhagen: license header

src/NuGetGallery/ViewModels/OwnerRequestsListViewModel.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace NuGetGallery
+{
+    public class OwnerRequestsListViewModel
+    {
+        public IEnumerable<OwnerRequestsListItemViewModel> RequestItems { get; private set; }
+
+        public string Name { get; private set; }
+
+        public User CurrentUser { get; private set; }
```
joelverhagen: Can be `{ get; }`

src/NuGetGallery/ViewModels/OwnerRequestsListViewModel.cs
```diff
@@ -0,0 +1,23 @@
...
+        
+        public OwnerRequestsListViewModel(IEnumerable<PackageOwnerRequest> requests, string name, User currentUser, IPackageService packageService)
+        {
+            RequestItems = requests.Select(r => new OwnerRequestsListItemViewModel(r, packageService)).ToArray();
```
joelverhagen: nit: `ToList()` requires fewer allocations.

src/NuGetGallery/ViewModels/OwnerRequestsViewModel.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Collections.Generic;
+
+namespace NuGetGallery
+{
+    public class OwnerRequestsViewModel
+    {
+        public OwnerRequestsListViewModel Incoming { get; private set; }
+
+        public OwnerRequestsListViewModel Outgoing { get; private set; }
```
joelverhagen: `{ get; }`

src/NuGetGallery/Views/Packages/ConfirmOwner.cshtml
```diff
@@ -12,11 +12,19 @@
...
                 || Model.Result == ConfirmOwnershipResult.AlreadyOwner)
             {
-                @:Email Confirmed
+                @:Ownership Confirmed
+            }
+            else if (Model.Result == ConfirmOwnershipResult.Rejected)
+            {
+                @:Ownership Rejected
+            }
+            else if (Model.Result == ConfirmOwnershipResult.Cancelled)
```
joelverhagen: Can this be a `switch`?

src/NuGetGallery/Views/Users/_OwnerRequestsList.cshtml
```diff
@@ -0,0 +1,82 @@
...
+        </h2>
+        <div class="panel-collapse collapse in" id="requests-@Model.Name" aria-expanded="true">
+            <p>
+                You have <b>@Model.RequestItems.Count().ToNuGetNumberString() @Model.Name.ToLowerInvariant() @requestsString</b>
+            </p>
+            <div class="list-packages" role="list">
+                <table class="table">
+                    <thead>
+                        <tr class="manage-package-headings">
+                            <th class="col-sm-1"></th>
```
joelverhagen: This table should align to changes already made on the `_UserPackagesList` such as hiding the icon on `xs`.

src/NuGetGallery/ViewModels/OwnerRequestsListViewModel.cs
loic-sharma: Does it? [`ToArray()`](https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/ToCollection.cs#L11) uses [`EnumerableHelpers.ToArray`](https://github.com/dotnet/corefx/blob/f17f1e847aeab830de77f8a46656339d7b0f1b43/src/Common/src/System/Collections/Generic/EnumerableHelpers.cs#L66), which does "the same growth logic as in List<T>" (see comment [here](https://github.com/dotnet/corefx/blob/f17f1e847aeab830de77f8a46656339d7b0f1b43/src/Common/src/System/Collections/Generic/EnumerableHelpers.cs#L135))

joelverhagen: `List` supports capacity greater than length. `Array` does not. Therefore `ToArray` requires another a copy at the end of the allocation process so the returned array has the exact correct length. Take a look at the implementation of `ToArray` vs. `ToList`. The implementation in .NET Framework is easier to read through (ILSpy > System.Core > `Enumerable`).

src/NuGetGallery/Services/PackageService.cs
scottbommarito: oh i see...yeaaaaa

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1102,7 +1102,7 @@ public virtual Task<ActionResult> RejectOwner(string id, string username, string
 
             if (!String.Equals(username, User.Identity.Name, StringComparison.OrdinalIgnoreCase))
             {
-                return View(new PackageOwnerConfirmationModel(id, username, ConfirmOwnershipResult.NotYourRequest));
+                return View("ConfirmOwner", new PackageOwnerConfirmationModel(id, username, ConfirmOwnershipResult.NotYourRequest));
```
loic-sharma: Should the view be called `"ConfirmPendingOwner"` to stay consistent with the controller's actions?

src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
```diff
@@ -10,16 +10,11 @@ public class OwnerRequestsListItemViewModel
         public PackageOwnerRequest Request { get; private set; }
 
         public Package Package { get; private set; }
-
+
```
loic-sharma: Remove whitespace here.

src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
```diff
@@ -10,16 +10,11 @@ public class OwnerRequestsListItemViewModel
...
         public OwnerRequestsListItemViewModel(PackageOwnerRequest request, IPackageService packageService)
-            : this(request, packageService.FindPackageByIdAndVersion(request.PackageRegistration.Id, null, SemVerLevelKey.SemVer2, true))
-        {
-        }
-
-        public OwnerRequestsListItemViewModel(PackageOwnerRequest request, Package package)
         {
             Request = request;
-            Package = package;
+            Package = packageService.FindPackageByIdAndVersion(request.PackageRegistration.Id, null, SemVerLevelKey.SemVer2, true);
```
loic-sharma: Nit: named parameters for `null` and `true`

src/NuGetGallery/ViewModels/OwnerRequestsViewModel.cs
scottbommarito: done

src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
scottbommarito: done

src/NuGetGallery/Services/PackageService.cs
scottbommarito: done

src/NuGetGallery/Views/Packages/ConfirmOwner.cshtml
scottbommarito: done

src/NuGetGallery/Views/Users/_OwnerRequestsList.cshtml
scottbommarito: done

src/NuGetGallery/ViewModels/OwnerRequestsListViewModel.cs
scottbommarito: done

src/NuGetGallery/ViewModels/OwnerRequestsListViewModel.cs
scottbommarito: done

src/NuGetGallery/ViewModels/OwnerRequestsListViewModel.cs
scottbommarito: I would prefer to leave it as `ToArray`, because `Array`s are immutable, but `List`s are not.

src/NuGetGallery/ViewModels/OwnerRequestsViewModel.cs
scottbommarito: done

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1103,7 +1106,19 @@ public virtual ActionResult Edit(string id, string version)
 
         [Authorize]
```
skofman1: Where is the antiforgery token attribute?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1103,7 +1106,19 @@ public virtual ActionResult Edit(string id, string version)
...
         [RequiresAccountConfirmation("accept ownership of a package")]
```
skofman1: also, since it's a POST it requires an HttpPost attribute

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1103,7 +1106,19 @@ public virtual ActionResult Edit(string id, string version)
...
-        public virtual async Task<ActionResult> ConfirmOwner(string id, string username, string token)
+        public virtual Task<ActionResult> ConfirmPendingOwnershipRequest(string id, string username, string token)
+        {
+            return HandleOwnershipRequest(id, username, token, accept: true);
+        }
+
+        [Authorize]
+        [RequiresAccountConfirmation("reject ownership of a package")]
+        public virtual Task<ActionResult> RejectPendingOwnershipRequest(string id, string username, string token)
```
skofman1: same

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1124,21 +1139,64 @@ public virtual ActionResult Edit(string id, string version)
...
+            else
+            {
+                await _packageService.RemovePackageOwnerAsync(package, user);
+
+                return View("ConfirmOwner", new PackageOwnerConfirmationModel(id, username, ConfirmOwnershipResult.Rejected));
+            }
+        }
+
+        [Authorize]
+        [RequiresAccountConfirmation("cancel ownership of a package")]
```
skofman1: please fix the text in the message to match the action: "cancel pending ownership request"

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1124,21 +1139,64 @@ public virtual ActionResult Edit(string id, string version)
...
+        public virtual async Task<ActionResult> CancelPendingOwnershipRequest(string id, string requestingUsername, string pendingUsername)
```
skofman1: same regarding attributes; HttpPost +AntiForgery

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1124,21 +1139,64 @@ public virtual ActionResult Edit(string id, string version)
...
+        {
+            if (!String.Equals(requestingUsername, User.Identity.Name, StringComparison.OrdinalIgnoreCase))
```
skofman1: string

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: It's expected to be called from outside the website. It can't have the attribute.

src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
scottbommarito: done

src/NuGetGallery/Controllers/PackagesController.cs

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
scottbommarito: done (no longer applicable)

src/NuGetGallery/Services/PackageService.cs
scottbommarito: merged this method with the GetPending method

src/NuGetGallery/ViewModels/OwnerRequestsViewModel.cs
scottbommarito: replaced with `nameof(Incoming)` and `nameof(Outgoing)`

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: done

src/NuGetGallery/Controllers/PackagesController.cs

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: These are `GET` requests, but they do need `HttpGet`.

skofman1: A request that changes the state of the service can't be a GET request

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1124,21 +1139,64 @@ public virtual ActionResult Edit(string id, string version)
...
+            }
+
+            var request = _packageService.GetPackageOwnershipRequest(package, requestingUser, pendingUser);
+            if (request == null)
+            {
+                return HttpNotFound();
+            }
 
-            SendAddPackageOwnerNotification(package, user, result.Item1, result.Item2);
+            await _packageService.RemovePackageOwnerAsync(package, pendingUser);
```
chenriksson: There's 2 queries for the owner request since RemovePackageOwnerAsync also handles confirmed owners. Instead, you could expose RemovePackageOwnerRequestAsync on the service that just tries to delete pending request if it exists.

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: Cancel and Reject are almost identical, except one is called by the pending and the other by the requesting owner. Why not just combine them and require that it be the pending or requesting?

chenriksson: Note it also resembles JsonApiController.RemovePackageOwner. Maybe that could be combined as well?

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -917,10 +922,20 @@ private static Package FindPackage(IEnumerable<Package> packages, Func<Package,
...
-        private PackageOwnerRequest FindExistingPackageOwnerRequest(PackageRegistration package, User pendingOwner)
+        private PackageOwnerRequest GetPendingOwnershipRequest(PackageRegistration package, User pendingOwner)
+        {
+            return (from request in _packageOwnerRequestRepository.GetAll()
+                    where request.PackageRegistrationKey == package.Key && 
+                        request.NewOwnerKey == pendingOwner.Key
+                    select request).FirstOrDefault();
+        }
+
+        private PackageOwnerRequest GetPendingOwnershipRequest(PackageRegistration package, User requestingOwner, User pendingOwner)
```
chenriksson: It's a little strange that only the requesting owner can cancel the request here, but all owners can cancel the request via the manage owners view.

src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace NuGetGallery
+{
+    public class OwnerRequestsListItemViewModel
+    {
+        public PackageOwnerRequest Request { get; }
+
+        public Package Package { get; }
+        
+        public OwnerRequestsListItemViewModel(PackageOwnerRequest request, IPackageService packageService)
```
chenriksson: View models are just state objects, right? So, should the caller query and just provide the package here? Same with other classes below.

src/NuGetGallery/ViewModels/OwnerRequestsViewModel.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System.Collections.Generic;
+
+namespace NuGetGallery
+{
+    public class OwnerRequestsViewModel
+    {
+        public OwnerRequestsListViewModel Incoming { get; }
+
+        public OwnerRequestsListViewModel Outgoing { get; }
```
chenriksson: What does Incoming and Outgoing mean? Are these the current user's pending ownership requests and requesting ownership requests? If so, maybe PendingRequests and OriginatingRequests?

src/NuGetGallery/Views/Packages/ConfirmOwner.cshtml
```diff
@@ -28,6 +36,20 @@
...
+                '<strong><a href="@Url.Package(@Model.PackageId)" title="The @Model.PackageId package">@Model.PackageId</a></strong>' package!
+            </p>
+        }
+        else if (Model.Result == ConfirmOwnershipResult.Cancelled)
+        {
+            <p class="text-center">
+                You have cancelled your request for <strong>@Model.Username</strong> to become an owner of the
+                '<strong><a href="@Url.Package(@Model.PackageId)" title="The @Model.PackageId package">@Model.PackageId</a></strong>' package!
+            </p>
+        }
```
chenriksson: Are these notifications useful? Would it be more useful to notify the requester that the pending owner rejected?

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -638,6 +638,45 @@ public static string RemovePackageOwner(this UrlHelper url)
                 hostName: GetConfiguredSiteHostName(url));
         }
 
+        public static string ConfirmPendingOwnershipRequest(this UrlHelper url, string packageId, string username, string confirmationCode)
+        {
+            return url.Action(
```
loic-sharma: >Action [](start = 23, length = 6)

You should probably follow Xavier's pattern of specifying the `protocol` and `hostName`

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: fixed it by adding `IPackageOwnerRequestService`

src/NuGetGallery/Views/Packages/ConfirmOwner.cshtml
scottbommarito: yes it can!

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: There's about 3 different actions with different names that map to this view. I'm not sure what it should be called.

src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
scottbommarito: done

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: not true, we have this today

src/NuGetGallery/Controllers/PackagesController.cs

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: good catch!

scottbommarito: I will use `IPackageOwnerRequestService`'s delete method

src/NuGetGallery/ViewModels/OwnerRequestsViewModel.cs
scottbommarito: Incoming = requests for YOU to be an owner of somebody else's package

Outgoing = requests you created for somebody else to be an owner of one of your packages

This is the header text that's used to display the requests on the page, so I'd prefer to leave it with this title. The subtext on the page explains it better. I'll add a comment.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1124,21 +1139,64 @@ public virtual ActionResult Edit(string id, string version)
...
+
+                SendAddPackageOwnerNotification(package, user, result.Item1, result.Item2);
+
+                return View("ConfirmOwner", new PackageOwnerConfirmationModel(id, username, ConfirmOwnershipResult.Success));
+            }
+            else
+            {
+                await _packageService.RemovePackageOwnerAsync(package, user);
+
+                return View("ConfirmOwner", new PackageOwnerConfirmationModel(id, username, ConfirmOwnershipResult.Rejected));
```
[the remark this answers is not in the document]
scottbommarito: I'm going to add a rejection notification as well as a cancel notification.

src/NuGetGallery/Views/Packages/ConfirmOwner.cshtml
scottbommarito: These are messages displayed on the page when you cancel the request. I have added a notification which is sent by `IMessageService`.

src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
scottbommarito: It's only the constructor that uses the `IPackageService`--it doesn't save the service.

src/NuGetGallery/Services/PackageService.cs
scottbommarito: Well this view only shows requests you've sent. I could change it to show all requests for any packages you own, but that seems a bit excessive for me.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I think they are different enough and simple enough that there isn't a huge benefit from combining them.

PR #4523, opened 2017-08-09
optimize svgs

SimonCropp:
saves approx 60%


SimonCropp:
feel free to pick or choose any of these

SimonCropp:
given how many of these are loaded on the home page it is worth the effort

joelverhagen:
Could you explain how this compression works? Is it meant to be lossless?

SimonCropp:
@joelverhagen i did a two pass approach

 1. go through by hand an remove things that are generally redundant. eg the title tag, the link attribute, id attributes (since i assume u are not styling by id or manipulating svg via code), comments etc. generally these should be settings on whatever tool u use to produce the svg assets, from some of the svg source it seems to be  http://www.bohemiancoding.com/sketch.
 2. run them all through a svg compression lib (eg https://compressor.io/compress). the first step will often result in many redundant grouping (multiple nested `<g>` with no attributes) and this step will collapse them

and yes it is meant to be lossless. TBH this was mostly meant to illustrate that there is a non trivial amount of payload u have which is redundant. in the long run you prob want to work this type of optimisation into your process. this can be part of your manual review process (if there is not much churn in graphics assets ), or done in whatever tool u use to produce/edit svgs, or even done on deploymentor your http pipeline through svg compression libs (a quick google gives this for example https://www.nuget.org/packages/Top.ImageCompressor.Client).

i also not running the pngs through a compressor also give a saving of ~60%

joelverhagen:
I think there are some issues with the compression:

image

As a side note, have you tried SVGO? It seems to have a lot of plug-ins and to be widely used.

SimonCropp:
@joelverhagen noted. i will take a look

PR #4525, opened 2017-08-09
Merge feature-redesign into dev

joelverhagen:
I will not be squashing commits. There will be a standard merge commit, due to the size of the merge.


skofman1:
Too big to review, will just trust you on this :) 
:shipit:

Issue #4527, opened 2017-08-09
Long package IDs make manage package controls wrap line


joelverhagen:
![image](https://user-images.githubusercontent.com/94054/29101050-9b28b300-7c64-11e7-9a26-4415a8016027.png)

PR #4532, opened 2017-08-09
Improve manage package page handling of long IDs

joelverhagen:
1. Allow long IDs to wrap on the manage packages package
1. Don't let the package controls to wrap unless screen size is xs
1. Hide images on xs
1. Allow browser to control column widths

Fix https://github.com/NuGet/NuGetGallery/issues/4527

Normal display:
screencapture-localhost-account-packages-1502292959515

Long IDs simulated by displaying `ID + ID`:
screencapture-localhost-account-packages-1502292578159

### DIFF
--- a/src/Bootstrap/dist/css/bootstrap-theme.css
+++ b/src/Bootstrap/dist/css/bootstrap-theme.css
@@ -368,10 +368,20 @@ img.package-icon {
   margin-bottom: 0;
 }
 .user-package-list .manage-package-listing .package-icon {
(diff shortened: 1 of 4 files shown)


scottbommarito:
Code looks good, can we see a picture or two at `xs`?

joelverhagen:
## `sm`
screencapture-localhost-account-packages-1502300835071

## `xs`

screencapture-localhost-account-packages-1502300879620

PR #4543, opened 2017-08-10
[Prefix] Add Reserved namespace service and tests

shishirx34:
- This is the first part for the Admin page changes.  
- This PR adds the service to interface with the database and related tests. The admin page and controller changes will be part of subsequent PRs. 
- Please note I have not added auditing as part of this PR since it would involve modifying shims, it will be a separate PR as well.


src/NuGetGallery/Services/IPackageService.cs
```diff
@@ -84,5 +84,7 @@ public interface IPackageService
         void EnsureValid(PackageArchiveReader packageArchiveReader);
 
         Task IncrementDownloadCountAsync(string id, string version, bool commitChanges = true);
+
+        Task UpdatePackageVerifiedStatusAsync(IList<PackageRegistration> package, bool isVerified);
```
xavierdecoster: Does this need to be `IList`? Could it be `IReadOnlyCollection` instead?

src/NuGetGallery/Services/IReservedNamespaceService.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        Task AddOwnerToReservedNamespaceAsync(ReservedNamespace prefix, User user);
+
+        Task DeleteOwnerFromReservedNamespaceAsync(ReservedNamespace prefix, User user);
+
+        ReservedNamespace FindReservedNamespaceForPrefix(string prefix);
+
+        IList<ReservedNamespace> FindAllReservedNamespacesForPrefix(string prefix, bool getExactMatches);
+
+        IList<ReservedNamespace> FindReservedNamespacesForPrefixList(IList<string> prefixList);
```
xavierdecoster: Same here, perhaps `IReadOnlyCollection`, both for input as output?

src/NuGetGallery/Services/IReservedNamespaceService.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    public interface IReservedNamespaceService
```
xavierdecoster: For function results, I generally prefer to return `IReadOnlyCollection` to tell consumers they have no business in modifying the resultset :)

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,242 @@
...
+            IUserService userService,
+            IPackageService packageService,
+            IAuditingService auditing)
+            : this()
+        {
+            EntitiesContext = entitiesContext;
+            ReservedNamespaceRepository = reservedNamespaceRepository;
+            UserService = userService;
+            PackageService = packageService;
+            // TODO: Add auditing
```
xavierdecoster: If there's a tracking issue, no need for the comment.

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,242 @@
...
+        {
+            if (newNamespace == null)
+            {
+                throw new ArgumentNullException(nameof(newNamespace));
+            }
+
+            var matchingReservedNamespaces = FindAllReservedNamespacesForPrefix(newNamespace.Value, !newNamespace.IsPrefix);
+            if (matchingReservedNamespaces.Count() > 0)
+            {
+                throw new InvalidOperationException($"The specified namespace is already reserved or is a more liberal namespace.");
```
xavierdecoster: Should we move the messages into resources?

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,242 @@
...
+                // Delete verified flags on corresponding packages for this prefix if it is the only prefix matching the 
+                // package registration.
+                if (namespaceToDelete.IsSharedNamespace == false)
+                {
+                    var packageRegistrationsToMarkUnVerified = namespaceToDelete
+                        .PackageRegistrations
+                        .Where(pr => pr.ReservedNamespaces.Count() == 1)
+                        .ToList();
+
+                    if (packageRegistrationsToMarkUnVerified.Count() > 0)
```
xavierdecoster: `if(packageRegistrationsToMarkUnVerified.Any())` may perform better

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,242 @@
...
+                if (namespaceToDelete == null)
+                {
+                    throw new InvalidOperationException($"Namespace '{existingNamespace.Value}' not found.");
+                }
+
+                // Delete verified flags on corresponding packages for this prefix if it is the only prefix matching the 
+                // package registration.
+                if (namespaceToDelete.IsSharedNamespace == false)
+                {
+                    var packageRegistrationsToMarkUnVerified = namespaceToDelete
```
xavierdecoster: Weird casing of `UnVerified`, isn't it `Unverified`, or even `ToMarkNonVerified` or `ToMarkNotVerified`?

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,242 @@
...
+            return ReservedNamespaceRepository.GetAll()
+                .Where(prefixMatch)
+                .ToList();
+        }
+
+        public IList<ReservedNamespace> FindReservedNamespacesForPrefixList(IList<string> prefixList)
+        {
+            return (from dbPrefix in ReservedNamespaceRepository.GetAll()
+                    join queryPrefix in prefixList
+                    on dbPrefix.Value.ToLower() equals queryPrefix.ToLower()
```
xavierdecoster: Is the `.ToLower()` really needed?

tests/NuGetGallery.Facts/Services/ReservedNamespaceServiceFacts.cs
```diff
@@ -0,0 +1,502 @@
...
+                    It.Is<ReservedNamespace>(
+                        rn => rn.Value == newNamespace.Value
+                            && rn.IsPrefix == newNamespace.IsPrefix
+                            && rn.IsSharedNamespace == newNamespace.IsSharedNamespace)));
+
+            service.MockReservedNamespaceRepository.Verify(x => x.CommitChangesAsync());
+        }
+
+        [Fact]
+        public async Task NullNamespaceReservationThrowsException()
```
xavierdecoster: Looks like a duplicate test of https://github.com/NuGet/NuGetGallery/pull/4543/files#diff-7861183f77ed349c8796810257ff2d06R35

xavierdecoster: Perhaps I missed it, but didn't see any test coverage for invalid namespaces (e.g. empty string, whitespaces only, unsupported characters, ...)

src/NuGetGallery/Services/IPackageService.cs
skofman1: Please change "verified" to "reserved". The word verified is used for security..

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -954,5 +954,22 @@ private void NotifyIndexingService()
                 }
             }
         }
+
+        public virtual async Task UpdatePackageVerifiedStatusAsync(IList<PackageRegistration> packageRegistrationList, bool isVerified)
```
skofman1: why does this need to be virtual?

src/NuGetGallery/Services/ReservedNamespaceService.cs
skofman1: you should also add the list of namespaces you found to the message

---
In reply to: [132390783](https://github.com/NuGet/NuGetGallery/pull/4543#discussion_r132390783) [](ancestors = 132390783)

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        public async Task AddReservedNamespaceAsync(ReservedNamespace newNamespace)
+        {
+            if (newNamespace == null)
+            {
+                throw new ArgumentNullException(nameof(newNamespace));
+            }
+
+            var matchingReservedNamespaces = FindAllReservedNamespacesForPrefix(newNamespace.Value, !newNamespace.IsPrefix);
```
skofman1: use named parameter

tests/NuGetGallery.Facts/Services/ReservedNamespaceServiceFacts.cs
```diff
@@ -0,0 +1,502 @@
...
+using NuGetGallery.TestUtils;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace NuGetGallery.Services
+{
+    public class ReservedNamespaceServiceFacts
```
skofman1: other UT classes in our code have a separate internal class per UTs of a method. I recommend you do the same here

tests/NuGetGallery.Facts/TestUtils/TestServiceUtility.cs
```diff
@@ -0,0 +1,172 @@
...
+        public TestableUserServiceWithDBFaking(FakeEntitiesContext context = null)
+        {
+            FakeEntitiesContext = context ?? new FakeEntitiesContext();
+            Config = (MockConfig = new Mock<IAppConfiguration>()).Object;
+            UserRepository = new EntityRepository<User>(FakeEntitiesContext);
+            Auditing = new TestAuditingService();
+        }
+    }
+
+    public class TestableReservedNamespaceService : ReservedNamespaceService
```
skofman1: why do you need to mock using inheritance? As far as I remember this pattern doesn't appear anywhere in the code base

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,242 @@
...
+                throw new ArgumentNullException(nameof(existingNamespace));
+            }
+
+            EntitiesConfiguration.SuspendExecutionStrategy = true;
+            using (var transaction = EntitiesContext.GetDatabase().BeginTransaction())
+            {
+                var namespaceToDelete = FindReservedNamespaceForPrefix(existingNamespace.Value);
+                if (namespaceToDelete == null)
+                {
+                    throw new InvalidOperationException($"Namespace '{existingNamespace.Value}' not found.");
```
loic-sharma: In this case, `EntitiesConfiguration.SuspendExecutionStrategy` is not set back to `false`. I think this would be a better pattern:

```
try
{
  EntitiesConfiguration.SuspendExecutionStrategy = true;
}
finally
{
  EntitiesConfiguration.SuspendExecutionStrategy = false;
}
```

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,242 @@
...
+        {
+            return (from request in ReservedNamespaceRepository.GetAll()
+                    where request.Value.Equals(prefix, StringComparison.OrdinalIgnoreCase)
+                    select request).FirstOrDefault();
+        }
+
+        public IList<ReservedNamespace> FindAllReservedNamespacesForPrefix(string prefix, bool getExactMatches)
+        {
+            Expression<Func<ReservedNamespace, bool>> prefixMatch =
+                dbPrefix => getExactMatches
```
loic-sharma: Consider doing the `getExactMatches` check when assigning `prefixMatch` instead of when evaluating `prefixMatch`.

src/NuGetGallery/Services/ReservedNamespaceService.cs
xavierdecoster: I thought our DB collation was case-insensitive too

src/NuGetGallery/Services/ReservedNamespaceService.cs
shishirx34: Added one, https://github.com/NuGet/NuGetGallery/issues/4549, removed TODO

src/NuGetGallery/Services/PackageService.cs
shishirx34: To override in testing MOQ needs it to be virtual. This is true for other functions above too. See `UpdateIsLatestAsync`

tests/NuGetGallery.Facts/Services/ReservedNamespaceServiceFacts.cs
shishirx34: You are right, it is a duplicate.

shishirx34: Added more tests.

tests/NuGetGallery.Facts/TestUtils/TestServiceUtility.cs
shishirx34: See TestableUserService. We do have a precedence for such format.

src/NuGetGallery/Services/IPackageService.cs
shishirx34: I think we agreed on "verified" after much deliberation because it came across the intent that we want to state. "Verified" is an identification property more than a security, plus we have client code based off of this property. I would rather keep it as "verified".

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        public async Task DeleteReservedNamespaceAsync(ReservedNamespace existingNamespace)
+        {
+            if (existingNamespace == null)
+            {
+                throw new ArgumentNullException(nameof(existingNamespace));
+            }
+
+            EntitiesConfiguration.SuspendExecutionStrategy = true;
```
shishirx34: This is better idea than doing try..finally.

src/NuGetGallery/Services/ReservedNamespaceService.cs
shishirx34: covered with `using` IDisposable

shishirx34:
@xavierdecoster, @loic-sharma, @scottbommarito - addressed feedback, please review.

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,262 @@
...
+
+            if (value.Length > CoreConstants.MaxPackageIdLength)
+            {
+                throw new ArgumentException(string.Format(
+                    CultureInfo.CurrentCulture,
+                    Strings.ReservedNamespace_NamespaceExceedsLength,
+                    CoreConstants.MaxPackageIdLength));
+            }
+
+            if (!NamespaceRegex.IsMatch(value))
```
agr: Probably, using an `IsMatch` overload that accepts the [timeout](https://msdn.microsoft.com/en-us/library/hh160210(v=vs.110).aspx) would be safer.

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,262 @@
...
+using System.Linq;
+using System.Linq.Expressions;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    public class ReservedNamespaceService : IReservedNamespaceService
+    {
+        private static readonly Regex NamespaceRegex = new Regex(@"^\w+([_.-|\w+])*$", RegexOptions.IgnoreCase | RegexOptions.ExplicitCapture);
```
agr: Add some tests to validate the regex correctness.

src/NuGetGallery/Services/ReservedNamespaceService.cs
agr: Dont' insist though. The length check above would likely cover the potential issues.

shishirx34: while I agree its a good practice, I think this regex is not as complex to warrant a timeout. I will keep it consistent with the ID matching and leave it as is.

src/NuGetGallery/Services/IReservedNamespaceService.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// Remove the specified user as an owner from the reserved namespace.
+        /// Also, all the package registrations owned by this user which match the 
+        /// specified namespace only will be marked as unverifed.
+        /// </summary>
+        /// <param name="prefix">The reserved namespace to modify</param>
+        /// <param name="username">The user to remove the ownership for the namespace</param>
+        /// <returns>Awaitable Task</returns>
+        Task DeleteOwnerFromReservedNamespaceAsync(string prefix, string username);
+
+        ReservedNamespace FindReservedNamespaceForPrefix(string prefix);
```
agr: Documentation here and below?..

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,263 @@
...
+using System.Linq.Expressions;
+using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    public class ReservedNamespaceService : IReservedNamespaceService
+    {
+        private static readonly Regex NamespaceRegex = new Regex(@"^\w+([_.-]\w+)*[.]?$", RegexOptions.IgnoreCase | RegexOptions.ExplicitCapture);
```
agr: There is no alpha characters in the regex, so `RegexOptions.IgnoreCase` option is not needed.

agr: You might also want to add the `RegexOptions.Compiled` here.

src/NuGetGallery.Core/Entities/SuspendDbExecutionStrategy.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Define the execution strategy for the EntitiesConfiguration for connection resiliency and retries
+    /// </summary>
+    public class SuspendDbExecutionStrategy : IDisposable
```
scottbommarito: Can you refactor the old code that also suspended the execution strategy to use this class as well? (I believe it's all in `IPackageDeleteService`.) Should be super simple.

src/NuGetGallery/Services/ReservedNamespaceService.cs
scottbommarito: @blowdart has said in the past to *always* use timeouts with regexp.

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,263 @@
...
+                        CultureInfo.CurrentCulture, Strings.ReservedNamespace_NamespaceNotFound, existingNamespace));
+
+                // Delete verified flags on corresponding packages for this prefix if it is the only prefix matching the 
+                // package registration.
+                if (!namespaceToDelete.IsSharedNamespace)
+                {
+                    var packageRegistrationsToMarkUnverified = namespaceToDelete
+                        .PackageRegistrations
+                        .Where(pr => pr.ReservedNamespaces.Count() == 1)
+                        .ToList();
```
agr: There is no need to call `ToList()` here. The subsequent `Any()` call will work with any `IEnumerable`.

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -0,0 +1,263 @@
...
+        public IReadOnlyCollection<ReservedNamespace> GetReservedNamespacesForId(string id)
+        {
+            return (from request in ReservedNamespaceRepository.GetAll()
+                    where id.StartsWith(request.Value)
+                    select request).ToList();
+        }
+
+        public static void ValidateNamespace(string value)
+        {
+            // Same restrictions as that of NuGetGallery.Core.Packaging.PackageIdValidator except for the regex change, a namespace could end in a '.'
```
scottbommarito: What if we just called the method in `PackageIdValidator`, BUT if it ends with a ".", remove the "." from the string we pass to the method? It should reduce the duplicated code.

tests/NuGetGallery.Facts/TestUtils/TestServiceUtility.cs
```diff
@@ -0,0 +1,172 @@
...
+using NuGetGallery.Framework;
+using System.Collections.Generic;
+using System.Data.Entity;
+using System.Linq;
+
+namespace NuGetGallery.TestUtils
+{
+    public static class PackageServiceUtility
+    {
+        private static readonly string _packageHashForTests = "NzMzMS1QNENLNEczSDQ1SA==";
```
agr: `const`?

tests/NuGetGallery.Facts/Services/ReservedNamespaceServiceFacts.cs
```diff
@@ -0,0 +1,569 @@
...
+    {
+        public class AllocateNamespace
```
scottbommarito: These nested classes should be named after the methods they test.

E.g. `TheAddReservedNamespaceAsyncMethod` is the convention in the codebase.

tests/NuGetGallery.Facts/Services/ReservedNamespaceServiceFacts.cs
```diff
@@ -0,0 +1,569 @@
...
+            [InlineData("  ")]
+            public async Task InvalidNamespaceThrowsException(string value)
+            {
+                var service = new TestableReservedNamespaceService();
+
+                await Assert.ThrowsAsync<ArgumentException>(async () => await service.DeleteReservedNamespaceAsync(value));
+            }
+
+            [Fact]
+            public async Task NonExistingNamespaceThrowsException()
```
scottbommarito: nit: `NonExisting` isn't grammatically correct, it should either be `Nonexistent` or `Missing`.

(IMO `Missing` is best)

src/NuGetGallery.Core/Entities/SuspendDbExecutionStrategy.cs
shishirx34: sure, let me check

src/NuGetGallery/Services/ReservedNamespaceService.cs
shishirx34: I don't think it would be a good idea, particularly with regexs, it sounds a lot hacky. I would rather duplicate.

tests/NuGetGallery.Facts/Services/ReservedNamespaceServiceFacts.cs
shishirx34: `Nonexistent` seems appropriate, `Missing` means, it was there and now its not.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -954,5 +954,21 @@ private void NotifyIndexingService()
...
         }
+
+        public virtual async Task UpdatePackageVerifiedStatusAsync(IReadOnlyCollection<PackageRegistration> packageRegistrationList, bool isVerified)
+        {
+            var allPackageRegistrations = _packageRegistrationRepository.GetAll();
+            var packageRegistrationsToUpdate = allPackageRegistrations
+                .Where(pr => packageRegistrationList.Any(prl => prl.Id == pr.Id))
+                .ToList();
+
+            packageRegistrationsToUpdate
```
loic-sharma: Move this into the `if (packageRegistrationsToUpdate.Count > 0)`?

PR #4490, opened 2017-07-31
Add AntiForgery tokens to some actions that were missing them

scottbommarito:
Mostly admin routes.

These include
- LuceneController
- SupportRequestController
- SecurityPolicyController
- CuratedPackagesController
- UsersController

https://github.com/NuGet/Engineering/issues/638

PR #4545, opened 2017-08-10
Merge dev into master

### LINKED ISSUE #4513: Average Download stats calculated incorrectly.

joelverhagen:
This has the UI redesign in addition to:

098e77736965d9309672cf8f52fde9fd3a485c35 - Fix #4513: avg download count calculated incorrectly. (#4520)
8c3cb749062917890a8c87dbc7d39a2c644cc9ac - Redirect to HTTPS when using packages.nuget.org (#4516)
75c022ef06e32a4cbd20c834cbb76e2007bd314d - DEV and INT, point to the new index.json locations (#4511)
ed3c98ef2f8f87c7adaf179a5b6dffe74836b709 - Ensure SSL v3 is disabled and that TLS v1.2 is enabled
94769fd9f7da4132ab14012d7a97a209eb232ede - Add AntiForgery tokens to some actions that were missing them (#4490)
3a1e4f41449044bee4b93e601ea7e234714fb1b6 - Cookie service init fix (#4488)


joelverhagen:
Merged with f44a86fb1507465b60baf70928b175a19213b394.

PR #4551, opened 2017-08-11
Gracefully handle unconfirmed email addresses in password reset

### LINKED ISSUE #4548: Resetting Password on Account with Unconfirmed Email Causes 500
The [`ForgotPassword`](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Controllers/UsersController.cs#L177) route will result in a 500 if the account's email is unconfirmed. This is due to the `ForgotPassword` route not catching the `InvalidOperationException` that AuthenticationService's [`GeneratePasswordResetToken`](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Authentication/AuthenticationService.cs#L391) throws if the user's email address is unconfirmed.

loic-sharma:
Fixes #4548

### DIFF
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -194,6 +194,8 @@ namespace NuGetGallery
                 }
                 catch (Exception e)
                 {
+                    e.Log();
+
                     ModelState.AddModelError("Email", e.GetUserSafeMessage());
                 }
             }
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -182,13 +182,20 @@ namespace NuGetGallery
 
             if (ModelState.IsValid)
             {
-                var user = await _authService.GeneratePasswordResetToken(model.Email, Constants.PasswordResetTokenExpirationHours * 60);
-                if (user != null)
+                try
                 {
-                    return SendPasswordResetEmail(user, forgotPassword: true);
-                }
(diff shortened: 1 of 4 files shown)


scottbommarito:
Can we write a test that tests that `UserController` gracefully handles this?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -182,13 +182,20 @@ public virtual ActionResult ForgotPassword()
...
+                    var user = await _authService.GeneratePasswordResetToken(model.Email, Constants.PasswordResetTokenExpirationHours * 60);
+                    if (user != null)
+                    {
+                        return SendPasswordResetEmail(user, forgotPassword: true);
+                    }
 
-                ModelState.AddModelError("Email", Strings.CouldNotFindAnyoneWithThatEmail);
+                    ModelState.AddModelError("Email", Strings.CouldNotFindAnyoneWithThatEmail);
+                }
+                catch (Exception e)
```
skofman1: you shouldn't catch all exceptions.. catch UserSafeException explicitly..

loic-sharma: I followed the pattern used in [PackagesController](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Controllers/PackagesController.cs#L193). Catching only `UserSafeException` means that all other exceptions will bubble through to users as 500s. If we're worried about exceptions, I think we should catch them here and then log them.

loic-sharma:
Added test and logged exceptions.

PR #4555, opened 2017-08-14
Introduce a type to represent the different password reset results (instead of null and exceptions)

joelverhagen:
I'm only submitting this PR because the work was already done on my devbox. It should have no functional different from @loic-sharma's https://github.com/NuGet/NuGetGallery/pull/4551 aside from a clarified message to the user. I poorly communicated my intent to fix issue https://github.com/NuGet/NuGetGallery/issues/4548.

There are three potential outcomes from this reset flow:
1. Success (the user exists and is confirmed)
1. User not found
1. User not confirmed (the user exists but has not confirmed his email)

I have added a type to clearly represent these three states. In general we should not be using exceptions as non-exceptional signals. Additionally, null is ambiguous signal in many cases so it should be avoided especially in a tri(/N)-state outcomes like this.

### DIFF
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -182,21 +182,19 @@ namespace NuGetGallery
 
             if (ModelState.IsValid)
             {
-                try
+                var result = await _authService.GeneratePasswordResetToken(model.Email, Constants.PasswordResetTokenExpirationHours * 60);
+                switch (result.Type)
                 {
-                    var user = await _authService.GeneratePasswordResetToken(model.Email, Constants.PasswordResetTokenExpirationHours * 60);
-                    if (user != null)
-                    {
-                        return SendPasswordResetEmail(user, forgotPassword: true);
-                    }
-
-                    ModelState.AddModelError("Email", Strings.CouldNotFindAnyoneWithThatEmail);
-                }
-                catch (Exception e)
-                {
-                    e.Log();
-
-                    ModelState.AddModelError("Email", e.GetUserSafeMessage());
+                    case PasswordResetResultType.UserNotConfirmed:
+                        ModelState.AddModelError("Email", Strings.UserIsNotYetConfirmed);
+                        break;
+                    case PasswordResetResultType.UserNotFound:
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -182,21 +182,19 @@ public virtual ActionResult ForgotPassword()
...
+                    case PasswordResetResultType.UserNotConfirmed:
+                        ModelState.AddModelError("Email", Strings.UserIsNotYetConfirmed);
+                        break;
+                    case PasswordResetResultType.UserNotFound:
+                        ModelState.AddModelError("Email", Strings.CouldNotFindAnyoneWithThatUsernameOrEmail);
+                        break;
+                    case PasswordResetResultType.Success:
+                        return SendPasswordResetEmail(result.User, forgotPassword: true);
+                    default:
+                        throw new NotImplementedException($"The passwword reset result type '{result.Type}' is not supported.");
```
skofman1: password

skofman1: We never want the user to see this error, so there should be a UT to make sure this doesn't happened.

joelverhagen: It will result in a 500 and the message will not be visible -- which is correct. Will add UT to protect.

joelverhagen: Added unit test covering this case.

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -301,6 +301,21 @@ public class TheForgotPasswordAction : TestContainer
                 Assert.IsNotType(typeof(RedirectResult), result);
                 Assert.Contains(Strings.UserIsNotYetConfirmed, result.ViewData.ModelState["Email"].Errors.Select(e => e.ErrorMessage));
             }
+
+            [Fact]
+            public async Task ThrowsNotImplementedExceptionWhenResultTypeIsUnknown()
```
skofman1: The UT should protect from the following scenario: someone added a new value to  but forgot to update the switch

joelverhagen: Spoke offline. Fixed.

Issue #4558, opened 2017-08-14
Redirect www.nuget.org/nuget.exe endpoint to dist


scottbommarito:
Currently this points to a very old version of `nuget.exe` (2.8.6) hosted in a blob storage that is not used otherwise. We should either
- deprecate the endpoint if it has no use
- redirect it to `dist`

scottbommarito:
The endpoint still has a lot of usage (13k in the last day). We should immediately redirect it to version 2.8.6 on dist and then consider migrating it to the latest version.

anangaur:
Linking https://github.com/NuGet/NuGetGallery/issues/4559

PR #4563, opened 2017-08-15
Redirect /nuget.exe endpoint to dist.nuget.org, nuget.exe 2.8.6

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/4558.

I will not merge this until @anangaur has posted about this on the announcements repo.

### DIFF
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -31,6 +31,8 @@ namespace NuGetGallery
     public partial class ApiController
         : AppController
     {
+        private const string NuGetExeUrl = "https://dist.nuget.org/win-x86-commandline/v2.8.6/nuget.exe";
+
         public IApiScopeEvaluator ApiScopeEvaluator { get; set; }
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -31,6 +31,8 @@ namespace NuGetGallery
     public partial class ApiController
         : AppController
     {
+        private const string NuGetExeUrl = "https://dist.nuget.org/win-x86-commandline/v2.8.6/nuget.exe";
```
shishirx34: why not use the latest url?

joelverhagen: This is an endpoint that hasn't changed in 4 years. It is being hit pretty frequently, roughly 12,000 times a day. I would rather not change to NuGet.exe 3.x/4.x since it would likely break some people.

shishirx34: Its fine I guess, but just wondering aren't our clients backwards compatible though? In most likely cases using the latest shouldn't break people using this route, push/install should just work.

PR #4576, opened 2017-08-18
Add autofocus to the searchbar search element

### LINKED ISSUE #4575: Please put focus on the search input control when nuget.org loads
I frequently show up to nuget.org with a package name on my clipboard to paste. *PLEASE* set the focus to the `search` **\<input>** element on the homepage so that when the page loads we can just paste (Ctrl-v) and hit Enter to get our package search results.

guardrex:
Fixes #4575

If ur interested. This is an unsolicited but :pray: hopeful PR. This is an insanely easy problem to fix; however, I do understand that if this partial appears on other pages in other contexts that it might pose a problem. In that case, I suggest a one-off partial for the homepage. If this element doesn't appear elsewhere, I hope you'll take this PR (and save me from *death by a thousand clicks*).


jeff-pang:
+1

PR #4595, opened 2017-08-25
README support (rebase, cleanup and polish)

chenriksson:
PR base is 'dev', but I'll be merging to 'feature-readme' on completion until we have OSS approval.

I'm also doing this to preserve authors for 93233d4  (@fayrose) and a994621 (@JarahSi). On merge, I plan to squash PR commits, except these. Note that these 2 commits were PR'd, but I'm including them for completeness due to merge conflicts and other changes.

Scope: This PR is solely for cleanup and polishing of the feature as-is. I expect another PR to cover changes for security (mini-markdown/encoding) or feature modifications (i.e., registration/version scoping, repoURL).


src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Takes in a Stream representing a readme file in markdown, converts it to HTML and 
+        /// returns a Stream representing the HTML version of the readme.
+        /// </summary>
+        /// <param name="readMeMarkdownStream">Stream containing a readMe in markdown</param>
+        /// <returns>A stream with the HTML version of the readMe</returns>
+        public static async Task<Stream> GetReadMeHtmlStream(Stream readMeMarkdownStream)
+        {
+            var markdown = await readMeMarkdownStream.ReadToEndAsync();
+            var html = CommonMarkConverter.Convert(markdown);
+            var sanitizedHtml = Sanitizer.Value.Sanitize(html).Trim();
```
chenriksson: This will likely change in next PR. Will likely have whitelist of approved tags and encode others, and not sure if we'll use sanitizer. Will be reviewing with security.

tests/NuGetGallery.Facts/Helpers/ReadMeHelperFacts.cs
```diff
@@ -0,0 +1,344 @@
...
+            }
+        }
+
+        /// <summary>
+        /// HTML sanitzation test scenarios, taken from the Microsoft Anti-XSS library (GetSafeHtmlFragment).
+        /// Differences between Anti-XSS and HtmlSanitizer (ignoring whitespace) are specified inline. Note that
+        /// input is treated as markdown, and some differences may be due to the markdown to Html conversion.
+        /// </summary>
+        /// <see cref="https://wpl.codeplex.com/SourceControl/latest#release/v4.3/Microsoft.Security.Application.HtmlSanitization.Tests/SanitizerTests.cs" />
+        public static TheoryData<string, string, string> AntiXssTestScenarios = new TheoryData<string, string, string> {
```
chenriksson: This will likely change, depending on whether we encode only or encode and sanitize.

src/NuGetGallery.Core/Entities/Package.cs
```diff
@@ -125,9 +125,25 @@ public Package()
         public string RepositoryUrl { get; set; }
 
         /// <summary>
-        /// Signifies whether or not ReadMe exists (optimization for pulling from blob storage)
+        /// Nullable flag stored in the database. Callers should use the HasReadMe property instead.
         /// </summary>
-        public bool? HasReadMe { get; set; }
+        [Column("HasReadMe")]
+        public bool? HasReadMeInternal { get; set; }
```
cristinamanum: Can this property be private or internal?

src/NuGetGallery.Core/Entities/PackageEdit.cs
```diff
@@ -45,10 +69,6 @@ public class PackageEdit
         public string LicenseUrl { get; set; }
         public string ProjectUrl { get; set; }
         public string RepositoryUrl { get; set; }
-        /// <summary>
-        /// changed, null (means unchanged), deleted
-        /// </summary>
-        public string ReadMeState { get; set; }
```
cristinamanum: Would it make sense as the ReadMeState to be an enum?

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+
+namespace NuGetGallery.Helpers
+{
+    internal static class ReadMeHelper
+    {
+        internal const string TypeUrl = "Url";
+        internal const string TypeFile = "File";
+        internal const string TypeWritten = "Written";
+        internal const int MaxFileSize = 40000;
+        private const int UrlTimeout = 10000;
```
cristinamanum: Are these milliseconds? Maybe a comment or embedded in the name information about this?

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+        private const string UriHostRequirement = "raw.githubusercontent.com";
+
+        private static Lazy<HtmlSanitizer> Sanitizer = new Lazy<HtmlSanitizer>(() => new HtmlSanitizer());
```
cristinamanum: nitpick: the default constructor will be used in Lazy. Should not need to explicitly calling it.

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            var readMeUri = new Uri(readMeUrl);
+            if (readMeUri.Host != UriHostRequirement)
+            {
+                throw new ArgumentException("Url must link to a raw markdown file hosted on Github. [" + UriHostRequirement + "]");
+            }
+
+            var webRequest = WebRequest.Create(readMeUrl);
+            webRequest.Timeout = UrlTimeout;
+
```
cristinamanum: Will we fail gracefully on timeout error?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -996,32 +995,45 @@ public virtual ActionResult Delete(string id, string version)
...
+                return Json(404, new [] { string.Format(Strings.PackageWithIdAndVersionNotFound, id, version) });
             }
 
             if (!package.IsOwner(User))
             {
-                return Json(403, new string[] { Strings.Unauthorized });
+                return Json(403, new [] { Strings.Unauthorized });
             }
 
+            // create model from package
```
loic-sharma: >create [](start = 15, length = 6)

Nit: please capitalize comments and add punctuation to remain consistent

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+            else
+            {
+                throw new InvalidOperationException("Form data contains an invalid ReadMeType.");
+            }
+
+            readMeStream = readMeStream.AsSeekableStream();
+            if (readMeStream.Length >= MaxFileSize)
+            {
+                throw new ArgumentException("ReadMe file exceeds size limitations. (" + MaxFileSize + ")");
```
loic-sharma: >throw [](start = 16, length = 5)

Use string interpolation here

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                readMeStream = new MemoryStream(Encoding.UTF8.GetBytes(readMeRequest.ReadMeWritten));
+            }
+            else if (readMeType.Equals(TypeFile, StringComparison.InvariantCultureIgnoreCase))
+            {
+                readMeStream = readMeRequest.ReadMeFile.InputStream;
+            }
+            else
+            {
+                throw new InvalidOperationException("Form data contains an invalid ReadMeType.");
```
loic-sharma: >throw [](start = 16, length = 5)

Adding the readMeType may be useful down the road for debugging

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// then a repository URL inputted via the website, then a repository URL entered through the nuspec.
+        /// </summary>
+        /// <param name="readMeRequest">The current package's form data submitted through the verify page</param>
+        /// <param name="packageMetadata">The package metadata from the nuspec file</param>
+        /// <returns>A stream with the encoded ReadMe file</returns>
+        public static async Task<Stream> GetReadMeMarkdownStream(ReadMeRequest readMeRequest)
+        {
+            Stream readMeStream;
+            var readMeType = readMeRequest.ReadMeType;
+            if (readMeType.Equals(TypeUrl, StringComparison.InvariantCultureIgnoreCase))
```
loic-sharma: >if [](start = 12, length = 2)

Is there a reason why a `HasReadMe` uses a switch and `GetReadMeMarkdownStream` uses if statements? If not, could we make them consistent?

src/NuGetGallery/RequestModels/ReadMeRequest.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Web.Mvc;
+
+namespace NuGetGallery
+{
+    public class ReadMeRequest
+    {
+        public virtual HttpPostedFileBase ReadMeFile { get; set; }
+
+        [AllowHtml]
+        public virtual string ReadMeWritten { get; set; }
```
loic-sharma: >ReadMeWritten [](start = 30, length = 13)

Please add some comments explaining the fields and what they represent

src/NuGetGallery/Services/IPackageFileService.cs
```diff
@@ -38,5 +45,10 @@ public interface IPackageFileService
         ///     Downloads the package from the file storage and reads it into a Stream asynchronously.
         /// </summary>
         Task<Stream> DownloadPackageFileAsync(Package packge);
+
+        /// <summary>
+        ///     Downloads the README from the file storage and reads it into a Stream asynchronously.
+        /// </summary>
+        Task<Stream> DownloadReadmeFileAsync(Package package, bool pending = false);
```
loic-sharma: >DownloadReadmeFileAsync [](start = 21, length = 23)

Nit: we should be consistent on `Readme` vs `ReadMe`. Personally I prefer `Readme`

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -83,19 +100,34 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
...
+            var fileName = BuildFileName(package, Constants.PackageFileSavePathTemplate, Constants.NuGetPackageFileExtension);
+            return _fileStorageService.GetFileAsync(Constants.PackagesFolderName, fileName);
         }
 
-        private static string BuildFileName(string id, string version)
+        public Task<Stream> DownloadReadmeFileAsync(Package package, bool pending = false)
+        {
+            if (package == null)
+            {
+                throw new ArgumentNullException("Package cannot be null!");
```
loic-sharma: >ArgumentNullException [](start = 26, length = 21)

Use `nameof(package)` instead of custom string.

src/NuGetGallery.Core/Entities/Package.cs
loic-sharma: If any changes are done here, do the same thing to `ReadMeStateInternal` on `PackageEdit.cs` 

---
In reply to: [135591600](https://github.com/NuGet/NuGetGallery/pull/4595#discussion_r135591600) [](ancestors = 135591600)

src/NuGetGallery.Core/Entities/PackageEdit.cs
```diff
@@ -58,5 +78,40 @@ public void Apply(string hashAlgorithm, string hash, long packageFileSize)
         {
             Package.ApplyEdit(this, hashAlgorithm, hash, packageFileSize);
         }
+
+        private PackageEditReadMeState GetReadMeStateString()
+        {
+            var state = ReadMeStateInternal ?? string.Empty;
+            
+            if (state.Equals("changed", StringComparison.InvariantCultureIgnoreCase))
```
loic-sharma: >InvariantCultureIgnoreCase [](start = 57, length = 26)

`InvariantCultureIgnoreCase` seems unnecessary considering the state should only be set by `SetReadMeStateString`. Also, could we use constants for `changed`/`deleted` strings?

tests/NuGetGallery.Facts/Services/PackageFileServiceFacts.cs
```diff
@@ -154,7 +155,7 @@ public void WillThrowIfPackageIsMissingVersion()
                 var service = CreateService(fileStorageSvc: fileStorageSvc);
                 var packageRegistraion = new PackageRegistration { Id = "theId" };
                 var package = new Package { PackageRegistration = packageRegistraion, NormalizedVersion = null, Version = "01.01.01" };
-                fileStorageSvc.Setup(x => x.CreateDownloadFileActionResultAsync(new Uri("http://fake"), It.IsAny<string>(), BuildFileName("theId", "1.1.1")))
+                fileStorageSvc.Setup(x => x.CreateDownloadFileActionResultAsync(new Uri("http://fake"), It.IsAny<string>(), BuildFileName("theId", "1.1.1",Constants.NuGetPackageFileExtension, Constants.PackageFileSavePathTemplate)))
```
loic-sharma: >, [](start = 154, length = 1)

Add space here.

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+            if (readMeUri.Host != UriHostRequirement)
+            {
+                throw new ArgumentException("Url must link to a raw markdown file hosted on Github. [" + UriHostRequirement + "]");
+            }
+
+            var webRequest = WebRequest.Create(readMeUrl);
+            webRequest.Timeout = UrlTimeout;
+
+            var response = await webRequest.GetResponseAsync();
+            return response.GetResponseStream().AsSeekableStream();
```
agr: If I remember correctly, `.AsSeekableStream()` in case the original stream is not seekable will read all data into memory, which may consume dozens of megabytes of memory. We probably need an alternative for which we'd be able to specify max bytes to read.

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Converts a ReadMe's url to a file stream.
+        /// </summary>
+        /// <param name="readMeUrl">A link to the raw ReadMe markdown file</param>
+        /// <returns>A stream to allow the file to be read</returns>
+        private static async Task<Stream> GetReadMeStreamFromUrl(string readMeUrl)
+        {
+            var readMeUri = new Uri(readMeUrl);
+            if (readMeUri.Host != UriHostRequirement)
+            {
+                throw new ArgumentException("Url must link to a raw markdown file hosted on Github. [" + UriHostRequirement + "]");
```
agr: Method is not Github specific, so I'd remove mentioning it from the exception message.

src/NuGetGallery/Web.config
```diff
@@ -24,16 +24,17 @@
     <!-- DEPLOYMENT SPECIFIC -->
     <!-- ******************* -->
     <!-- These should change on every deployment (to rotate credentials, etc.) -->
-    <add key="Gallery.StorageType" value="FileSystem" />
+    <add key="Gallery.StorageType" value="AzureStorage" />
```
agr: Revert the whole file.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1893,18 +1932,83 @@ public static IEnumerable<object[]> WillApplyEdits_Data
...
+                yield return new object[] { new EditPackageVersionRequest() { ReleaseNotes = "notes123" } };
+                yield return new object[] { new EditPackageVersionRequest() { Summary = "summary new" } };
+                yield return new object[] { new EditPackageVersionRequest() { Tags = "tag1new tag2new" } };
+                yield return new object[] { new EditPackageVersionRequest() { VersionTitle = "title" } };
+                yield return new object[] { new EditPackageVersionRequest() { ReadMeState = PackageEditReadMeState.Unchanged } };
+            }
+        }
+
+        [Theory]
+        [MemberData("WillApplyReadMe_Data")]
```
agr: nit: `[MemberData(nameof(WillApplyReadMe_Data)]`?

src/NuGetGallery/Scripts/gallery/page-edit-readme.js
```diff
@@ -0,0 +1,213 @@
...
+        }
+        else if (readMeFile) {
+            readMeType = "File";
+        }
+    }
+
+    // prepare form data with antiforgery token
+    var formData = new FormData();
+    var token = $('[name=__RequestVerificationToken]').val();
+    formData.append("__RequestVerificationToken", token);
```
joelverhagen: >append [](start = 13, length = 6)

We have a helper for this now -- check out manage owners page.

src/NuGetGallery/Scripts/gallery/page-edit-readme.js
```diff
@@ -0,0 +1,213 @@
...
+
+function clearReadMeError() {
+    if (!$("#readme-errors").hasClass("hidden")) {
+        $("#readme-errors").addClass("hidden");
+        $("#readme-error-content").text("");
+    }
+    $("#preview-readme-button").removeAttr("disabled");
+}
+
+function changeReadMeFormTab() {
```
joelverhagen: >changeReadMeFormTab [](start = 9, length = 19)

[https://getbootstrap.com/docs/3.3/javascript/#tabs](https://getbootstrap.com/docs/3.3/javascript/#tabs)

src/NuGetGallery/Scripts/gallery/page-edit-readme.js
```diff
@@ -0,0 +1,213 @@
...
+    });
+
+    $('#readme-select-feedback').on('click', function () {
+        $('#readme-select-file').click();
+    });
+
+    $('#readme-select-file').on('change', function () {
+        clearReadMeError();
+        displayReadMeEditMarkdown();
+        var fileName = $('#readme-select-file').val().split("\\").pop();
```
joelverhagen: >"\\" [](start = 60, length = 4)

This assumes Windows, right? What is the behavior on Mac/Linux?

src/NuGetGallery/Scripts/gallery/page-edit-readme.js
```diff
@@ -0,0 +1,213 @@
...
+        if (fileName.length > 0 && validateReadMeFileName(fileName)) {
+            $('#readme-select-feedback').attr('placeholder', fileName);
+            clearReadMeError();
+        } else if (fileName.length > 0) {
+            $('#readme-select-feedback').attr('placeholder', 'Browse to select a package file...');
+            displayReadMeError("Please enter a markdown file with one of the following extensions: '.md', '.mkdn', '.mkd', '.mdown', '.markdown', '.txt' or '.text'.");
```
joelverhagen: >'. [](start = 99, length = 2)

This list should be shared with the validation list below.

src/NuGetGallery/Scripts/gallery/page-edit-readme.js
```diff
@@ -0,0 +1,213 @@
...
+function previewReadMeAsync(callback, error) {
+    // collect and validate readme data
+    var readMeType = $("input[name='ReadMe.ReadMeType']:checked").val();
+    var readMeUrl = $("#ReadMeUrlInput").val();
+    var readMeWritten = $("#readme-written").val();
+    var readMeFileInput = $("#readme-select-file");
+    var readMeFileName = readMeFileInput && readMeFileInput[0] ? readMeFileInput.val().split("\\").pop() : null;
+    var readMeFile = readMeFileName && validateReadMeFileName(readMeFileName) ? readMeFileInput[0].files[0] : null;
+
+    if (readMeType === undefined) {
```
joelverhagen: >readMeType [](start = 8, length = 10)

When would this be undefined?

src/NuGetGallery/Scripts/gallery/page-edit.js
```diff
@@ -173,7 +181,27 @@ var EditViewManager = new function () {
         });
 
         $("#verify-collapser-container").removeClass("hidden");
+        $("#submit-collapser-container").removeClass("hidden");
+
+        window.nuget.configureExpander(
```
joelverhagen: >configureExpander [](start = 21, length = 17)

configureExpanderHeading

src/NuGetGallery/Scripts/gallery/page-edit.js
```diff
@@ -173,7 +181,27 @@ var EditViewManager = new function () {
...
+            "verify-package-form",
+            "ChevronRight",
+            "Verify",
+            "ChevronDown",
+            "Verify");
+        window.nuget.configureExpander(
```
joelverhagen: >configureExpander [](start = 21, length = 17)

configureExpanderHeading

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -83,19 +100,34 @@ public Task StorePackageFileInBackupLocationAsync(Package package, Stream packag
...
+            }
+
+            var formatter = pending ?
```
joelverhagen: >formatter [](start = 16, length = 9)

nit: format (formatter is a weird name for a string)

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -218,6 +218,33 @@
...
+                    <h2>
+                        <a href="#" role="button" data-toggle="collapse" data-target="#readme-container"
+                           aria-expanded="true" aria-controls="readme-container" id="show-readme-container">
+                            <i class="ms-Icon ms-Icon--ChevronDown" aria-hidden="true"></i>
+                            <span>Documentation</span>
+                        </a>
+                    </h2>
+                </div>
+                <div id="readme-container" class="collapse in">
+                    <div id="readme-clamped" class="collapse in" style="max-height: 228.57px">
```
joelverhagen: >228.57px [](start = 84, length = 8)

Why?

joelverhagen: max-height I mean. Also, styles should not be inline.

---
In reply to: [135848095](https://github.com/NuGet/NuGetGallery/pull/4595#discussion_r135848095) [](ancestors = 135848095)

src/NuGetGallery/Views/Packages/Edit.cshtml
```diff
@@ -61,34 +61,27 @@
...
-
-            @Html.Partial("_EditMetadata")
+            @Html.Partial("_EditForm")
         </div>
     </div>
 </section>
 
 @section BottomScripts {
     @Scripts.Render("~/Scripts/jquery-ui-1.10.3.js")
+    @Scripts.Render("~/Scripts/gallery/bootstrap.js")
```
joelverhagen: >bootstrap [](start = 39, length = 9)

bootstrap.js should already be included

jquery-ui should be added to a bundle

src/NuGetGallery/Views/Packages/Edit.cshtml
```diff
@@ -61,34 +61,27 @@
...
     <script>
         $(function () {
         var viewModel = @Html.Raw(Json.Encode(Model.Edit));
         var realModel = {
             "Id": '@Model.PackageId',
             "Version": '@Model.Version',
-            "Edit": viewModel
+            "Edit": viewModel,
+            "ReadMe": {
+                "ReadMeType": "Written",
```
joelverhagen: >Written [](start = 31, length = 7)

Why is this constant written here? the rest of these constants are here since they are server side mastered.

src/NuGetGallery/Views/Packages/UploadPackage.cshtml
```diff
@@ -57,31 +57,25 @@
...
 @section BottomScripts
 {
     @* Right now this is the only page that uses this script. If we increase our usage of it, we should put it in our bundles *@
-    <script type="text/javascript" src="@Url.Content("~/Scripts/gallery/async-file-upload.js")"></script>
+    @Scripts.Render("~/Scripts/gallery/page-edit-readme.js")
+    @Scripts.Render("~/Scripts/gallery/async-file-upload.js")
     <script type="text/javascript">
         var InProgressPackage = @Html.Raw(Json.Encode(Model.InProgressUpload));
-        window.nuget.configureExpanderHeading("upload-package-form");
+        window.nuget.configureExpander(
```
joelverhagen: >configureExpander [](start = 21, length = 17)

Why not configureExpanderHeading?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1076,6 +1091,68 @@ public virtual ActionResult Edit(string id, string version)
...
+        {
+            var oldReadMe = package.HasReadMe ?
+                NormalizeNewLines(await (await _packageFileService.DownloadReadmeFileAsync(package)).ReadToEndAsync()) :
+                null;
+
+            var newReadMe = ReadMeHelper.HasReadMe(edit?.ReadMe) ?
+                NormalizeNewLines(await (await ReadMeHelper.GetReadMeMarkdownStream(edit.ReadMe)).ReadToEndAsync()) :
+                null;
+
+            var hasReadMe = !string.IsNullOrWhiteSpace(newReadMe);
```
joelverhagen: >IsNullOrWhiteSpace [](start = 36, length = 18)

When would we have a null or whitespace readme here? can't this be filtered out sooner?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1076,6 +1091,68 @@ public virtual ActionResult Edit(string id, string version)
...
+            if (hasReadMe && !newReadMe.Equals(oldReadMe))
+            {
+                await _packageFileService.SaveReadMeFileAsync(package, new MemoryStream(Encoding.UTF8.GetBytes(newReadMe)));
```
joelverhagen: >SaveReadMeFileAsync [](start = 42, length = 19)

Is there a maximum size for a Readme? This could be a lot in memory right?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -435,9 +432,11 @@ public virtual JsonResult UploadPackageProgress()
                 if (pendingMetadata != null)
                 {
                     model.SetPendingMetadata(pendingMetadata);
+                    pendingReadMe = pendingMetadata.ReadMeState == PackageEditReadMeState.Changed;
                 }
             }
 
+            model.ReadMeHtml = await GetPackageReadMeHtml(package, pendingReadMe);
```
joelverhagen: >GetPackageReadMeHtml [](start = 37, length = 20)

Wait, so we are downloading and rendering the MD for every page view? I am concerned about perf impact.

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            var webRequest = WebRequest.Create(readMeUrl);
```
joelverhagen: >WebRequest [](start = 29, length = 10)

Why WebRequest? HttpClient is the newer API.

src/NuGetGallery/Helpers/ReadMeHelper.cs
joelverhagen: Yes. We we should deal with the raw, unbuffered network stream and only read MaxFileSize bytes.

---
In reply to: [135660095](https://github.com/NuGet/NuGetGallery/pull/4595#discussion_r135660095) [](ancestors = 135660095)

src/NuGetGallery.Core/StreamExtensions.cs
```diff
@@ -20,5 +21,13 @@ public static Stream AsSeekableStream(this Stream stream)
             memoryStream.Position = 0;
             return memoryStream;
         }
+
+        public static async Task<string> ReadToEndAsync(this Stream stream)
+        {
+            using (var reader = new StreamReader(stream, System.Text.Encoding.UTF8))
```
joelverhagen: >UTF8 [](start = 78, length = 4)

Why specify encoding here? StreamReader can try to detect encoding

src/NuGetGallery.Core/Entities/Package.cs
chenriksson: It looks like EF 6+ allows [mapping of non-public properties](https://stackoverflow.com/questions/7619955/mapping-private-property-entity-framework-code-first), but it looks ugly and may require reflection. Would prefer to keep it simple.

src/NuGetGallery.Core/Entities/PackageEdit.cs
chenriksson: SQL Azure doesn't support enums. This change is exposing the property to the Gallery as an enum, but mapping it to a string DB column.

src/NuGetGallery/Helpers/ReadMeHelper.cs
chenriksson: I'm removing HtmlSanitizer altogether. Per security feedback, will encode and not sanitize.

src/NuGetGallery/Helpers/ReadMeHelper.cs
chenriksson: No, but it should be case insensitive.

src/NuGetGallery/Services/IPackageFileService.cs
chenriksson: Explorers mostly used 'ReadMe', and corresponding migration has already been applied. Due to this I'll conform to 'ReadMe'.

src/NuGetGallery.Core/Entities/PackageEdit.cs
chenriksson: Agree it should be unnecessary, but inclined to leave in as a safe guard esp since I've kept the mapped ReadMeStateInternal column as public. Added constants.

src/NuGetGallery/Helpers/ReadMeHelper.cs
chenriksson: It is, due to the UriHostRequirement restriction. We could debate whether to keep this restriction.

tests/NuGetGallery.Facts/Helpers/ReadMeHelperFacts.cs
chenriksson: Going to encode only (all markdown text), per security recommendation.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
chenriksson: Ah, right. I think this was a workaround we added for the demo. For some reason clamp wasn't behaving after the last rebase. I'll investigate the right fix.

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: Yes, this was the recommendation from @blowdart . The issue is that persisting the converted HTML could mean we are persisting insecure data, should there be a vulnerability in our conversion (or the OSS we use). Yes, this does mean a perf hit.

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: see [ReadMeHelper.MaxFileSize](https://github.com/NuGet/NuGetGallery/pull/4595/files#diff-4db82a8bd6284e61c6b57ff713f51ebcR19). Throws exception if file size is exceeded.

Let me know if you recommend different limit - we could start smaller. Yes, will be a lot in memory especially considering the additional conversion to html.

src/NuGetGallery/Views/Packages/Edit.cshtml
chenriksson: jquery-ui doesn't look necessary here

src/NuGetGallery.Core/Entities/PackageEdit.cs
dtivel: EF supports enums natively.  Why not let EF serialize/deserialize the enum?

src/NuGetGallery/Helpers/ReadMeHelper.cs
dtivel: The problem is that units were lost when the idea was translated into code and a unitless data type of `int` was chosen.  There is already a built-in primitive for time intervals.

Change the data type to `TimeSpan` and initialize the field with an appropriate static method.  For example,

    private static readonly TimeSpan UrlTimeout = TimeSpan.FromSeconds(10);

Then, any call site can obtain the value in the units it wants:

    webRequest.Timeout = (int)UrlTimeout.TotalMilliseconds;

This eliminates the need for a comment explaining the intended unit of measurement.

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// </summary>
+        /// <param name="formData">A ReadMeRequest with the ReadMe data from the form.</param>
+        /// <returns>Whether there is a ReadMe to upload.</returns>
+        public static bool HasReadMe(ReadMeRequest formData)
+        {
+            switch (formData?.ReadMeType)
+            {
+                case TypeUrl:
+                    var readMeUrl = formData.ReadMeUrl;
+                    return !string.IsNullOrWhiteSpace(readMeUrl) && Uri.IsWellFormedUriString(readMeUrl, UriKind.Absolute);
```
dtivel: How about also checking that `readMeUrl` uses the HTTPS scheme (`Uri.UriSchemeHttps`)?

src/NuGetGallery/Helpers/ReadMeHelper.cs
dtivel: Without the restriction and any further restrictions, `readMeUrl` could be an HTTP endpoint on an internal (Azure) network.  It could also be a file:///-reference; the readme feature could be used to display arbitrary local file content from the web server.   Neither of these are desirable for nuget.org-hosted instance of NuGetGallery.

Use of `HttpClient` vs `WebRequest` probably eliminates the non-HTTP/-HTTPS use cases, but I haven't examined at it in depth.

src/NuGetGallery/Scripts/gallery/page-edit-readme.js
chenriksson: This matches [async-file-upload](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Scripts/gallery/async-file-upload.js#L64), but does seem to assume Windows.

Looks like both should change to:
.split(/(\\|\/)/g).pop()

Assuming we can't depend on [.name](https://www.w3schools.com/jsref/prop_fileupload_name.asp) property, since it seems to lack IE support.

src/NuGetGallery/Scripts/gallery/page-edit-readme.js
chenriksson: Originally, was undefined on edit - but code will change with latest refactoring.

src/NuGetGallery.Core/Entities/PackageEdit.cs
chenriksson: EF does, but I didn't think SQL Azure did; I'll try to verify this. However, note that the migration has already been deployed so any change here would require a new migration.

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: ReadMeHelper.HasReadMe doesn't guarantee that the content is non-null / non-empty. See @dtivel 's comments. This is why this was done here, but I'll see if there's anything I can do to clean it up.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1076,6 +1091,68 @@ public virtual ActionResult Edit(string id, string version)
...
+            else
+            {
+                edit.ReadMeState = PackageEditReadMeState.Unchanged;
+            }
+
+            return hasReadMe;
+        }
+
+        /// <summary>
+        /// Get the HTML that results from markdown conversion, encoding and sanitization.
```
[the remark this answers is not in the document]
chenriksson: Can't tell what 'this' links to. Updated comment to remove sanitization reference.

chenriksson:
Pushed some updates. More cleanup and fixes are coming, including:
- More PR feedback, tests
- File upload fix
- Clamp fix
- Telemetry

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+            return readMeStream;
+        }
+
+        /// <summary>
+        /// Gets the readme.md source data from a url.
+        /// </summary>
+        /// <param name="readMeRequest">Request containing readme.md source data.</param>
+        /// <returns>Stream containing the readme.md data.</returns>
+        private static async Task<Stream> GetReadMeStreamFromUrl(string readMeUrl)
```
agr: Argument name in the documentation does not match the name in the method declaration.

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,140 @@
...
+                readMeStream = await GetReadMeStreamFromUrl(readMeRequest.SourceUrl);
+            }
+            else if (TypeWritten.Equals(readMeType, StringComparison.InvariantCultureIgnoreCase))
+            {
+                readMeStream = new MemoryStream(Encoding.UTF8.GetBytes(readMeRequest.SourceText));
+            }
+            else if (TypeFile.Equals(readMeType, StringComparison.InvariantCultureIgnoreCase))
+            {
+                var uploadFileName = readMeRequest.SourceFile.FileName;
+                if (!Path.GetExtension(uploadFileName).Equals(Constants.MarkdownFileExtension, StringComparison.OrdinalIgnoreCase))
```
agr: Inconsistent `StringComparison` usage: 4 lines above `StringComparison.InvariantCultureIgnoreCase` is used. I suggest using `StringComparison.OrdinalIgnoreCase` everywhere.

src/NuGetGallery/Scripts/gallery/page-edit-readme.js
chenriksson: Simplified to only accept .md uploads, validate server-side same as with nupkg

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
chenriksson: Replaced clampJS with our custom expanders

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1455,6 +1534,27 @@ private string GetSecurePushPolicyDescriptions()
...
+            }
+
+            try
+            {
+                var readMeHtml = await (await ReadMeHelper.GetReadMeHtmlStream(formData)).ReadToEndAsync();
+                return Json(new [] { readMeHtml });
+            }
+            catch (Exception ex)
+            {
+                return Json(400, new string[] { "Failed to convert markdown to Html: " + ex.Message });
```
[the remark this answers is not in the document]
chenriksson: No longer handler conversion exception

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,159 @@
...
+                throw new ArgumentException(Strings.ReadMeUrlHostInvalid, nameof(readMeMdUrl));
+            }
+
+            using (var client = new HttpClient() { Timeout = UrlTimeout, MaxResponseContentBufferSize = MaxMdLengthBytes })
+            {
+                using (var readMeMdStream = await client.GetStreamAsync(readMeMdUrl))
+                {
+                    using (var reader = new StreamReader(readMeMdStream))
+                    {
+                        return await reader.ReadToEndAsync();
```
agr: Missing the text length check.

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// <param name="readMeMdUrl">ReadMe.md URL, which must be a raw github file.</param>
+        /// <returns>Markdown content.</returns>
+        private static async Task<string> GetReadMeMdFromUrlAsync(string readMeMdUrl)
+        {
+            if (!Uri.IsWellFormedUriString(readMeMdUrl, UriKind.Absolute) || new Uri(readMeMdUrl).Host != UrlHostRequirement)
+            {
+                throw new ArgumentException(Strings.ReadMeUrlHostInvalid, nameof(readMeMdUrl));
+            }
+
+            using (var client = new HttpClient() { Timeout = UrlTimeout, MaxResponseContentBufferSize = MaxMdLengthBytes })
```
agr: My previous comment disappeared, but setting the `MaxResponseContentBufferSize` does not limit the amount of bytes read from the response.

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// </summary>
+        /// <param name="readMeRequest">ReadMe request from Verify or Edit package page.</param>
+        /// <returns>Markdown content.</returns>
+        public static async Task<string> GetReadMeMdAsync(ReadMeRequest readMeRequest)
+        {
+            var readMeType = readMeRequest?.SourceType;
+
+            if (TypeWritten.Equals(readMeType, StringComparison.InvariantCultureIgnoreCase))
+            {
+                if (readMeRequest.SourceText.Length > MaxMdLengthBytes)
```
agr: This checks the text length in characters, not in bytes (same in the `GetReadMeMdFromPostedFileAsync`). Likely not a big deal, but then the code that reads it from the network should be consistent and use the same logic.

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -0,0 +1,159 @@
...
+            using (var client = new HttpClient() { Timeout = UrlTimeout, MaxResponseContentBufferSize = MaxMdLengthBytes })
+            {
+                using (var readMeMdStream = await client.GetStreamAsync(readMeMdUrl))
+                {
+                    using (var reader = new StreamReader(readMeMdStream))
+                    {
+                        return await reader.ReadToEndAsync();
+                    }
+                }
+            };
```
agr: Extra semicolon?

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -144,16 +145,19 @@ public static string GetReadMeHtml(string readMeMd)
...
             }
 
-            using (var client = new HttpClient() { Timeout = UrlTimeout, MaxResponseContentBufferSize = MaxMdLengthBytes })
+            using (var client = new HttpClient() { Timeout = UrlTimeout })
             {
-                using (var readMeMdStream = await client.GetStreamAsync(readMeMdUrl))
+                using (var response = await client.GetAsync(readMeMdUrl))
                 {
-                    using (var reader = new StreamReader(readMeMdStream))
+                    if (response.Content.Headers.ContentLength > MaxMdLengthBytes)
```
agr: `Content-Length` header is not required and may contain invalid data, so we can't rely on it to figure out response length. We should probably just try reading first `MaxMdLengthBytes + 1` bytes from the stream. If we end up reading more than `MaxMdLengthBytes` throw exception, otherwise, try converting bytes to string.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1076,6 +1092,84 @@ public virtual ActionResult Edit(string id, string version)
...
+                edit.ReadMeState = PackageEditReadMeState.Unchanged;
+            }
+
+            return hasReadMe;
+        }
+
+        /// <summary>
+        /// Get the package README.md from storage.
+        /// </summary>
+        private async Task<string> GetPackageReadMeMdAsync(Package package, bool isPending = false)
```
joelverhagen: >GetPackageReadMeMdAsync [](start = 35, length = 23)

There is a lot of logic in this controller. The class is HUGE. Could we introduce a ReadMeService?

src/NuGetGallery/Helpers/ReadMeHelper.cs
joelverhagen: Yeah, agreed. An 8000 byte buffer is a very reasonable size. In fact, I bet the buffer used by ReadAsStringAsync is a similar size already.

---
In reply to: [138472860](https://github.com/NuGet/NuGetGallery/pull/4595#discussion_r138472860) [](ancestors = 138472860)

src/NuGetGallery/Helpers/ReadMeHelper.cs
```diff
@@ -147,15 +147,24 @@ public static string GetReadMeHtml(string readMeMd)
...
 
-                    return await response.Content.ReadAsStringAsync();
+                        if (offset == buffer.Length)
+                        {
+                            throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture,
+                                Strings.ReadMeMaxLengthExceeded, MaxMdLengthBytes));
+                        }
+                    }
+                    
+                    return Encoding.UTF8.GetString(buffer).Trim('\0');
```
agr: `Encoding.UTF8.GetString(buffer, 0, bytesRead)` shouldn't require trimming.

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: @joelverhagen - refactored ReadMeHelper into a ReadMeService.

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -0,0 +1,261 @@
...
+        /// <returns>Markdown content.</returns>
+        private static async Task<string> GetReadMeMdFromPostedFileAsync(HttpPostedFileBase readMeMdPostedFile)
+        {
+            if (!Path.GetExtension(readMeMdPostedFile.FileName).Equals(Constants.MarkdownFileExtension, StringComparison.InvariantCulture))
+            {
+                throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture,
+                    Strings.ReadMePostedFileExtensionInvalid, Constants.MarkdownFileExtension));
+            }
+
+            if (readMeMdPostedFile.ContentLength > MaxMdLengthBytes)
```
joelverhagen: >readMeMdPostedFile [](start = 16, length = 18)

Can we use the same buffering technique as below?

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -0,0 +1,261 @@
...
+                        offset += bytesRead;
+
+                        if (offset == buffer.Length)
+                        {
+                            throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture,
+                                Strings.ReadMeMaxLengthExceeded, MaxMdLengthBytes));
+                        }
+                    }
+                    
+                    return Encoding.UTF8.GetString(buffer).Trim('\0');
```
joelverhagen: >UTF8 [](start = 36, length = 4)

Assuming UTF-8 here is probably correct in most cases but this seems to be ignoring any `charset` parameter to `Content-Type`. I would be curious to know what encoding `HttpClient.GetStringAsync` uses when decoding the network stream. I would hope it inspects the response headers.

joelverhagen: Yes, I looked at the implementation of `HttpClient` and it first tries to use `charset`, then tries looking for a BOM, then defaults to UTF-8. 

---
In reply to: [139208022](https://github.com/NuGet/NuGetGallery/pull/4595#discussion_r139208022) [](ancestors = 139208022)

src/NuGetGallery/Services/PackageFileService.cs
```diff
@@ -13,6 +14,9 @@ namespace NuGetGallery
 {
     public class PackageFileService : IPackageFileService
     {
+        private const string ReadMeFilePathTemplateActive = "active/{0}/{1}{2}";
```
joelverhagen: >ReadMeFilePathTemplateActive [](start = 29, length = 28)

nit comment on what each string param is

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -0,0 +1,261 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        ///  Normalize new lines, used for readme.md change detection.
+        /// </summary>
+        private string NormalizeNewLines(string content)
+        {
+            return content?.Replace("\r\n", "\n").Replace("\n", Environment.NewLine).Trim();
```
joelverhagen: Nit, this can be done with one pass using a regex

src/NuGetGallery/Services/ReadMeService.cs
joelverhagen: Why are we explicitly trimmed \0?

chenriksson: Encoding.GetString doesn't trim, but leaves trailing '\0' for any unused buffer space.

chenriksson:
Also added HasReadMe to AuditedPackage, which is saved to storage auditing. Doesn't look like shim change is necessary, as it doesn't copy all of the package properties.

PR #4599, opened 2017-08-29
Move pager to center of the page

khellang:
I just think this

image

looks much better than this

image

Also, I couldn't find the pager at first, cause I expected it to be in the center (which is also the Bootstrap default.)

### DIFF
--- a/src/Bootstrap/dist/css/bootstrap-theme.css
+++ b/src/Bootstrap/dist/css/bootstrap-theme.css
@@ -374,7 +374,6 @@ img.package-icon {
   }
 }
 .pager {
-  float: right;
   margin-top: 75px;
   margin-bottom: 0;
 }
--- a/src/Bootstrap/less/theme/common-list-packages.less
+++ b/src/Bootstrap/less/theme/common-list-packages.less
@@ -66,5 +66,4 @@
 .pager {
   margin-top: 75px;
   margin-bottom: 0px;
-  float: right;
 }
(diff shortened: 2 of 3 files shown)


joelverhagen:
Thanks for the contribution. Since this is purely UI, I have sent this over to our designer for approval. Concerning the implementation, the .less needs to be compiled into .css. This can be done by executing `grunt` (no parameters) in the `src/Bootstrap` directory.

khellang:
Are you actually checking in generated artifacts?

joelverhagen:
> Are you actually checking in generated artifacts?

Yes. We chose to a) not compile the .less at runtime and b) not require users that `git clone` the repository to install npm/grunt.

joelverhagen:
Also, designer signed off so once the compiled .css is checked in, I can merge this into dev!

khellang:
> not compile the .less at runtime

Sure, fair enough. I was thinking about compile time. Before you deploy. To avoid having me compile it manually and the noise it makes in source control.

> not require users that `git clone` the repository to install npm/grunt.

But by having me compile the CSS and check it in, isn't this exactly what you're doing? 🤔

joelverhagen:
> But by having me compile the CSS and check it in, isn't this exactly what you're doing? 🤔 

Yeah, that's a fair point, but we felt that most users that `git clone` the repo just want to run it or perhaps want to make C#/Razor changes. CSS changes historically have been comparatively rare.

khellang:
Done :+1:

khellang:
BTW, I don't think it's too much to ask devs to install NPM and do a restore/build (it could even be part of the existing build.cmd/ps1) to run a front-end web project in 2017 😄

joelverhagen:
Rebased on `dev` and merged with f62358a7e012db3ad3d08a16775ebb032563890d.

Issue #4571, opened 2017-08-17
Package statistics graph contains wrong values


justdmitry:
Graph with package download counts ([like this](https://www.nuget.org/stats/packages/Newtonsoft.Json?groupby=Version)) is displaying wrong values.

Version with download count "3 649 326" shows bar with height of 3, with " 700 877" downloads - bar with height of 700.

It seems that graph parses values until first space.

Here is screenshot:
![statistics-bug](https://user-images.githubusercontent.com/3540597/29405767-92deccb8-8347-11e7-825b-3a45bb37d2c6.png)

joelverhagen:
@justdmitry, thanks for reporting this. Could you send us the "Accept-Language" header your browser is sending to NuGet.org? This can be found in Chrome developer tools, Network tab under Request Headers.

agr:
I'd guess it's `ru-RU`.

justdmitry:
Yes, it's ru-RU: `Accept-Language: ru-RU,ru;q=0.8,en-US;q=0.5,en;q=0.3`

PR #4602, opened 2017-08-30
Update Number display to clientside formatting.

ryuyu:
Addresses https://github.com/NuGet/NuGetGallery/issues/4571

@joelverhagen @jonwchu

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/stats-dimensions.js
+++ b/src/NuGetGallery/Scripts/gallery/stats-dimensions.js
@@ -124,15 +124,10 @@ var setupHiddenRows = function (data) {
                     content.attr("href", item.Uri);
                     content.text(item.Data);
                 } else {
-                    if (item.IsNumeric) {
-                        content = $(document.createElement("span"));
-                        content.attr("aria-label", parseInt(item.Data).toLocaleString());
-                        content.text(parseInt(item.Data).toLocaleString());
-                    } else {
-                        content = $(document.createElement("span"));
-                        content.attr("aria-label", item.Data);
-                        content.text(item.Data);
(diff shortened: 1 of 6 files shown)


tests/NuGetGallery.Facts/Controllers/StatisticsControllerFacts.cs
```diff
@@ -518,7 +518,7 @@ public class StatisticsControllerFacts
             }
 
             Assert.Equal(603, sum);
-            Assert.Equal("603", actualReport.Total);
+            Assert.Equal("603", actualReport.Total.ToNuGetNumberString());
```
joelverhagen: Just `Assert.Equal(int, int)`

src/NuGetGallery/Scripts/gallery/stats-dimensions.js
```diff
@@ -122,12 +122,19 @@ var setupHiddenRows = function (data) {
...
-                    content.attr("aria-label", item.Data);
+                    if (item.IsNumeric) {
+                        content = $(document.createElement("span"));
+                        content.attr("aria-label", parseInt(item.Data).toLocaleString());
+                        content.text(parseInt(item.Data).toLocaleString());
+                    } else {
+                        content = $(document.createElement("span"));
+                        content.attr("aria-label", item.Data);
+                        content.text(item.Data);
+                    }
```
shishirx34: nit: can be simplified.
```
var textValue = item.IsNumeric ? parseInt(item.Data).toLocaleString() : item.Data;
content = $(document.createElement("span");
content.attr("aria-label", textValue);
content.text(textValue);
```

PR #4611, opened 2017-09-04
Package details page warning for SemVer is inconsistent with other places

### LINKED ISSUE #4552: Package details page warning for SemVer is inconsistent with other places.
Expected:
image
This is what we see while uploading a SemVer package

Actual:
image

xavierdecoster:
Fixes #4552

### DIFF
--- a/src/NuGetGallery/ViewModels/PackageViewModel.cs
+++ b/src/NuGetGallery/ViewModels/PackageViewModel.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using NuGet.Versioning;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/ViewModels/PackageViewModel.cs
```diff
@@ -24,7 +25,13 @@ public PackageViewModel(Package package)
             Version = String.IsNullOrEmpty(package.NormalizedVersion) ?
                 NuGetVersionFormatter.Normalize(package.Version) :
                 package.NormalizedVersion;
-            
+
+            HasSemVer2Version = NuGetVersion.Parse(_fullVersion).IsSemVer2;
+            HasSemVer2Dependency = package.Dependencies.ToList()
```
loic-sharma: Just to double-check - does `Dependencies` include dependencies of dependencies?

xavierdecoster: Those are the top-level dependencies only.

loic-sharma: Oh. Does that mean that a package is not considered a semver 2 package if one of its dependencies has a semver 2 dependency?

xavierdecoster: Yep! See the non-goals in the SemVer2 spec for nuget.org ;) https://github.com/NuGet/Home/wiki/SemVer2-support-for-nuget.org-%28server-side%29#non-goals

loic-sharma: I see, thanks for the info! :)

PR #4612, opened 2017-09-05
Use SiteRoot config setting for generated URLs

xavierdecoster:
When having a proxy in front of the gallery deployment, the gallery currently generates hyperlinks (e.g. in emails, UI, ...) that use the deployment host name instead of the proxy's.

This fix leverages the `SiteRoot` config setting instead of the resolved host name of the deployment's environment.

### DIFF
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -151,14 +151,14 @@ namespace NuGetGallery
                 try
                 {
                     var package = PackageService.FindPackageByIdAndVersion(
-                        id, 
-                        version, 
-                        SemVerLevelKey.SemVer2, 
+                        id,
+                        version,
+                        SemVerLevelKey.SemVer2,
                         allowPrerelease: false);
 
                     if (package == null)
                     {
-                       return new HttpStatusCodeWithBodyResult(HttpStatusCode.NotFound, String.Format(CultureInfo.CurrentCulture, Strings.PackageWithIdAndVersionNotFound, id, version));
+                        return new HttpStatusCodeWithBodyResult(HttpStatusCode.NotFound, String.Format(CultureInfo.CurrentCulture, Strings.PackageWithIdAndVersionNotFound, id, version));
                     }
                     version = package.NormalizedVersion;
 
@@ -174,7 +174,7 @@ namespace NuGetGallery
                 {
                     QuietLog.LogHandledException(e);
 
-			        // Database was unavailable and we don't have a version, return a 503
+                    // Database was unavailable and we don't have a version, return a 503
                     return new HttpStatusCodeWithBodyResult(HttpStatusCode.ServiceUnavailable, Strings.DatabaseUnavailable_TrySpecificVersion);
                 }
             }
@@ -261,7 +261,7 @@ namespace NuGetGallery
             {
                 await AuthenticationService.RemoveCredential(user, credential);
             }
-            
+
             TelemetryService.TrackVerifyPackageKeyEvent(id, version, user, User.Identity, result?.StatusCode ?? 200);
 
             return (ActionResult)result ?? new EmptyResult();
@@ -285,7 +285,7 @@ namespace NuGetGallery
             {
                 return new HttpStatusCodeWithBodyResult(HttpStatusCode.Forbidden, Strings.ApiKeyNotAuthorized);
             }
-            
+
             if (CredentialTypes.IsPackageVerificationApiKey(credential.Type))
             {
                 // Secure path: verify that verification key matches package scope.
@@ -370,7 +370,7 @@ namespace NuGetGallery
                             {
                                 message = ex.Message;
                             }
-                            
+
                             return new HttpStatusCodeWithBodyResult(HttpStatusCode.BadRequest, message);
                         }
 
@@ -399,7 +399,7 @@ namespace NuGetGallery
                         {
                             // Check if API key allows pushing a new package id
                             if (!ApiKeyScopeAllows(
-                                subject: nuspec.GetId(), 
+                                subject: nuspec.GetId(),
                                 requestedActions: NuGetScopes.PackagePush))
                             {
                                 // User cannot push a new package ID as the API key scope does not allow it
@@ -414,8 +414,8 @@ namespace NuGetGallery
                                 // Audit that a non-owner tried to push the package
                                 await AuditingService.SaveAuditRecordAsync(
                                     new FailedAuthenticatedOperationAuditRecord(
-                                        user.Username, 
-                                        AuditedAuthenticatedOperationAction.PackagePushAttemptByNonOwner, 
+                                        user.Username,
+                                        AuditedAuthenticatedOperationAction.PackagePushAttemptByNonOwner,
                                         attemptedPackage: new AuditedPackageIdentifier(
                                             nuspec.GetId(), nuspec.GetVersion().ToNormalizedStringSafe())));
 
@@ -427,7 +427,7 @@ namespace NuGetGallery
 
                             // Check if API key allows pushing the current package id
                             if (!ApiKeyScopeAllows(
-                                packageRegistration.Id, 
+                                packageRegistration.Id,
                                 NuGetScopes.PackagePushVersion, NuGetScopes.PackagePush))
                             {
                                 // User cannot push a package as the API key scope does not allow it
@@ -460,7 +460,7 @@ namespace NuGetGallery
                         };
 
                         var package = await PackageService.CreatePackageAsync(
-                            packageToPush, 
+                            packageToPush,
                             packageStreamMetadata,
                             user,
                             commitChanges: false);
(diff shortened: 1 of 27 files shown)


src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -495,16 +495,16 @@ public virtual Task<ActionResult> CreatePackagePost()
...
                             new PackageAuditRecord(package, AuditedPackageAction.Create, PackageCreatedVia.Api));
 
                         // Notify user of push
                         MessageService.SendPackageAddedNotice(package,
-                            Url.Action("DisplayPackage", "Packages", routeValues: new { id = package.PackageRegistration.Id, version = package.NormalizedVersion }, protocol: Request.Url.Scheme),
-                            Url.Action("ReportMyPackage", "Packages", routeValues: new { id = package.PackageRegistration.Id, version = package.NormalizedVersion }, protocol: Request.Url.Scheme),
-                            Url.Action("Account", "Users", routeValues: null, protocol: Request.Url.Scheme));
+                            Url.Package(package.PackageRegistration.Id, package.NormalizedVersion, Request.Url.Scheme),
+                            Url.ReportPackage(package.PackageRegistration.Id, package.NormalizedVersion, Request.Url.Scheme),
+                            Url.AccountSettings(Request.Url.Scheme));
```
xavierdecoster: Wondering if we should just default to `https` everywhere...

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -19,47 +23,94 @@ public static string Current(this UrlHelper url)
...
-            if (path.StartsWith("~/", StringComparison.OrdinalIgnoreCase))
+            return url.RequestContext.HttpContext.Request.IsSecureConnection ? "https" : "http";
+        }
+
+        private static string GetConfiguredSiteRoot(UrlHelper url)
+        {
+            if (_configuration == null)
             {
-                path = VirtualPathUtility.ToAbsolute(path, url.RequestContext.HttpContext.Request.ApplicationPath);
+                SetConfigurationService(new ConfigurationService(new SecretReaderFactory(new DiagnosticsService())));
```
skofman1: >SecretReaderFactory [](start = 69, length = 19)

1. This will cause reading all configured secrets from keyvault. Is there a way to load ConfigurationService using DI?
2. Consider making this code thread safe

skofman1:
> using System.Collections.Specialized;

are all changes covered by UTs? Should we add functional tests to cover those changes and additional issues we found (for example OData API pointing to traffic manager)

---
Refers to: tests/NuGetGallery.Facts/TestUtils/TestUtility.cs:5 in 8635236. [](commit_id = 8635236e23cb767fb2bcb8029c0634012d2558f3, deletion_comment = False)

src/NuGetGallery/Controllers/ApiController.cs
xavierdecoster: @joelverhagen @skofman1 thoughts on this?

joelverhagen: I think using the request scheme is right. When `IAppConfiguration.RequireSSL` is enabled, we redirect to HTTPS so this should already be reflecting configuration. Consider the case where gallery is running HTTP, not HTTPS, and the site has no valid certificate.

xavierdecoster: Check! There are code paths currently that are hard-coded to use `http` (e.g. in `JsonApiController`, the `SendPackageOwnerRequest` e-mails contain `http` hyperlinks, no matter what value the `IAppConfiguration.RequireSSL` setting has).

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -19,47 +23,94 @@ public static string Current(this UrlHelper url)
...
+        private static string GetConfiguredSiteHostName(UrlHelper url)
+        {
+            var siteRoot = GetConfiguredSiteRoot(url);
+            return new Uri(siteRoot).Host;
         }
 
         public static string Home(this UrlHelper url)
         {
-            return url.RouteUrl(RouteName.Home);
+            return url.RouteUrl(
```
loic-sharma: >RouteUrl [](start = 23, length = 8)

Could we create a `RouteUrl` helper that automatically provides the `protocol` and `hostName` using `GetProtocol` and `GetconfiguredSiteHostName`?

src/NuGetGallery/UrlExtensions.cs
xavierdecoster: Updated. This commit will set the configuration service once, when initializing the DI container: https://github.com/NuGet/NuGetGallery/pull/4612/commits/99c0e918ba4e8914b368f510805340d7c58f40ca

All of the code in this class are extension methods to `UrlHelper`.

src/NuGetGallery/Controllers/ApiController.cs
xavierdecoster: @joelverhagen Actually, you gave me an idea. I think it's nicer if the links we generate as part of e-mail body are HTTPS if the server is configured to require SSL. [This commit](https://github.com/NuGet/NuGetGallery/pull/4612/commits/a3dae2b600f8224b51cd47a5c873aba0598411ba) takes that config setting into consideration. Thoughts?

joelverhagen: Looks great.

xavierdecoster:
@skofman1 we used to have web UI tests checking for broken links if I'm not mistaken. There's also the `UrlExtensionFacts` test class, but other than that, there's not much unit tests covering this at present.

src/NuGetGallery/UrlExtensions.cs
xavierdecoster: That's what these extension methods are: helper methods that automatically provide `protocol` and `hostName`. AFAIK, [it's not possible to override the default resolution/behavior](https://stackoverflow.com/questions/8492798/is-it-possible-to-override-urlhelper-generateurl-across-asp-net-mvc-framework/8492905) of protocol and host name provided by `UrlHelper`.

Btw, these extension methods were there already. I just updated their behavior, and ensured we use them in those places that were relying on `Url.Action` or `Url.RouteUrl`.

skofman1:
@xavierdecoster , we need to have test coverage to make sure not only that the links are not broken, but also that they point to the correct URL (and not traffic manager). Ideas?

xavierdecoster:
@skofman1 open for any ideas, need to give this some thought. Note that even Azure API Management policies may affect this, so unit tests don't seem like the most solid approach to me...

In fact, all traffic manager URLs are currently dynamically replaced with the proxy URL using an API Management policy (content rewrite rule).

skofman1:
@xavierdecoster , so perhaps functional tests are a better place for those tests. We need to make sure that the API management config is correct, and the rule wasn't dropped.

xavierdecoster:
@joelverhagen added unit test to detect usages of default UrlHelper routing in .cshtml files https://github.com/NuGet/NuGetGallery/pull/4612/commits/3d7ab393146e4e265b760145747cedeab42ebfe2

image

tests/NuGetGallery.Facts/Views/UrlHelperFacts.cs
```diff
@@ -0,0 +1,97 @@
...
+            // in all razor files of the NuGetGallery web project.
+            AssertNoViolationsOfUnsupportedTermInRazorFiles(unsupportedTerm);
+        }
+
+        private void AssertNoViolationsOfUnsupportedTermInRazorFiles(string unsupportedTerm)
+        {
+            var violations = new ConcurrentBag<Tuple<int, string, string>>();
+
+            Parallel.ForEach(
+                GetRazorFiles(),
```
joelverhagen: We should assert that there is at least one Razor file to catch working directory issues

tests/NuGetGallery.Facts/Views/UrlHelperFacts.cs
```diff
@@ -0,0 +1,97 @@
...
+                file =>
+                {
+                    var lineNumber = 0;
+                    foreach (var line in File.ReadLines(file))
+                    {
+                        lineNumber++;
+
+                        if (line.Contains(unsupportedTerm))
+                        {
+                            violations.Add(new Tuple<int, string, string>(lineNumber, file, line.TrimStart(' ').TrimEnd(' ')));
```
joelverhagen: nit: Tuple.Create to imply types

tests/NuGetGallery.Facts/Views/UrlHelperFacts.cs
xavierdecoster: Fixed https://github.com/NuGet/NuGetGallery/pull/4612/commits/6fb7d1d198af259765143bfa339d3be9f0d3cea6

tests/NuGetGallery.Facts/Views/UrlHelperFacts.cs
xavierdecoster: Fixed https://github.com/NuGet/NuGetGallery/pull/4612/commits/b52a807b062a2c6f5e46a9caeba680083ed52ab3

PR #4613, opened 2017-09-05
Delete snapshots

cristinamanum:
Force delete snapshots on delete blob action.

### DIFF
--- a/src/NuGetGallery/Services/CloudBlobWrapper.cs
+++ b/src/NuGetGallery/Services/CloudBlobWrapper.cs
@@ -47,9 +47,14 @@ namespace NuGetGallery
         public Task DeleteIfExistsAsync()
         {
             return Task.Factory.FromAsync<bool>(
-                (cb, state) => _blob.BeginDeleteIfExists(cb, state), 
-                ar => _blob.EndDeleteIfExists(ar),
-                state: null);
+                    (cb, state) => _blob.BeginDeleteIfExists(deleteSnapshotsOption: DeleteSnapshotsOption.IncludeSnapshots,
+                    accessCondition: null,
+                    options: null,
+                    operationContext: null,
+                    callback: cb,
+                    state: state),
+                    ar => _blob.EndDeleteIfExists(ar),
+                    state: null);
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Services/CloudBlobWrapper.cs
```diff
@@ -47,9 +47,14 @@ public string ETag
...
-                (cb, state) => _blob.BeginDeleteIfExists(cb, state), 
-                ar => _blob.EndDeleteIfExists(ar),
-                state: null);
+                    (cb, state) => _blob.BeginDeleteIfExists(deleteSnapshotsOption: DeleteSnapshotsOption.IncludeSnapshots,
+                    accessCondition: null,
+                    options: null,
+                    operationContext: null,
+                    callback: cb,
+                    state: state),
+                    ar => _blob.EndDeleteIfExists(ar),
```
loic-sharma: Can replace `ar => _blob.EndDeleteIfExists(ar)` with just `_blob.EndDeleteIfExists`

cristinamanum: We can. If not strong opinion about it I would let the code structure like it was before.

src/NuGetGallery/Services/CloudBlobWrapper.cs
```diff
@@ -47,9 +47,14 @@ public string ETag
...
             return Task.Factory.FromAsync<bool>(
-                (cb, state) => _blob.BeginDeleteIfExists(cb, state), 
-                ar => _blob.EndDeleteIfExists(ar),
-                state: null);
+                    (cb, state) => _blob.BeginDeleteIfExists(deleteSnapshotsOption: DeleteSnapshotsOption.IncludeSnapshots,
+                    accessCondition: null,
+                    options: null,
+                    operationContext: null,
+                    callback: cb,
+                    state: state),
```
agr: Can we have the 5 lines above indented more? They are arguments to the `BeginDeleteIfExists` and not to the `FromAsync`.

Issue #782, opened 2013-02-08
Search:The search bar can provide a gesture to add search filters


bhuvak:
We have enabled search using "id", "author" "description" "title" and "tags".
But it is not very discoverable to the end user.
It would be good to provide a drop down arrow with the search bar to add search filters. Take a look at the "test explorer" windows's search ( Visual studi -->Test --> Window -->Test explorer" ) for example. Attaching a screen shot for the same.
![SearchFilter](https://f.cloud.github.com/assets/2625465/141330/ea41eada-7244-11e2-8cbc-907658d32118.PNG)

robertmuehsig:
This is a pretty old issue, but why not give it a try in the fresh look. 
I'm not sure if this is the solution for this issue, it is more like in #1210 suggested - a small hint how the search syntax works on NuGet.org and everywhere else:

![image](https://user-images.githubusercontent.com/756703/29477921-bbfca268-846a-11e7-9187-42ffaed4c402.png)

Simple idea: If you enter a search term, but you get 0 package results the small hint appears and points to: https://docs.microsoft.com/nuget/consume-packages/finding-and-choosing-packages#search-syntax

This is more or less like the github search experience if you enter nothing and just press search:

![image](https://user-images.githubusercontent.com/756703/29478014-21afd882-846b-11e7-93c4-0e146e880c9b.png)

It is not super fancy, but should be accessible from all kinds of clients (mobile / tablet / screenreader etc.)

I also played with the top-message, but currently I can't inject HTML (the HTML will be html encoded)  to generate the link and maybe this isn't the right place at all.

![image](https://user-images.githubusercontent.com/756703/29478086-82e99d90-846b-11e7-8f35-1684946759b6.png)


What do you think?

robertmuehsig:
Any comments @skofman1 @joelverhagen ?

skofman1:
@robertmuehsig , I like this idea! @anangaur , thoughts?

joelverhagen:
Great idea. This is just linking to existing docs, so it seems perfect!

anangaur:
Yepp.. Lets get going.. 👍

PR #4616, opened 2017-09-05
Adding a search hint if query results in no items

robertmuehsig:
Based on the idea posted here: https://github.com/NuGet/NuGetGallery/issues/782 

Maybe the issue then already solved that way - nothing to fancy. It should look like this:

image

The link points to this location:
https://docs.microsoft.com/nuget/consume-packages/finding-and-choosing-packages#search-syntax

### DIFF
--- a/src/NuGetGallery/Views/Shared/ListPackages.cshtml
+++ b/src/NuGetGallery/Views/Shared/ListPackages.cshtml
@@ -48,6 +48,20 @@
         </div>
     </div>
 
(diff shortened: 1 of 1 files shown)


skofman1:
@anangaur

AnandG-it:
LGTM 👍

anangaur:
LGTM 👍

robertmuehsig:
That was fast - thanks for merging!

PR #4622, opened 2017-09-06
Fix incorrect test setup where Package did not have a Version

xavierdecoster:


tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -399,8 +399,8 @@ public class TheDisplayPackageMethod
                             Id = "Foo",
                             Owners = new List<User>()
                         },
-                        Version = "2.0.0a",
-                        NormalizedVersion = "2.0.0a",
+                        Version = "2.0.0",
```
skofman1: why change this?

tests/NuGetGallery.Facts/Controllers/CuratedFeedsControllerFacts.cs
```diff
@@ -189,14 +189,15 @@ public class TheListPackagesAction
                     Id = "RedPill",
                     Key = 2,
                     DownloadCount = 0,
-                    Packages = new []
+                    Packages = new[]
```
loic-sharma: Why remove the space between the `new` and the `[]`? `[]` is essentially a type...

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
xavierdecoster: It's an invalid version :)

tests/NuGetGallery.Facts/Controllers/CuratedFeedsControllerFacts.cs
xavierdecoster: auto-formatting of doc in VS

PR #4633, opened 2017-09-07
Add AddPackageLicenseReport2 which supports a larger Value field

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/743.

Our license report job is failing because the value generated here...
https://github.com/NuGet/NuGetGallery/blob/9eb4bc49f7f34339b0fc4a6183f9265509961cdd/src/NuGetGallery/Infrastructure/AddPackageLicenseReport.sql#L135
... is more than 128 characters.
https://github.com/NuGet/NuGetGallery/blob/9eb4bc49f7f34339b0fc4a6183f9265509961cdd/src/NuGetGallery/Infrastructure/AddPackageLicenseReport.sql#L55

The fix is to add a new stored procedure with a fix. I chose to use a new one so rollback of job or DB can be done freely and independently.

There will be another PR to update the UpdateLicenseReports job to use the new sproc.


joelverhagen:
Note that it was possible to do an `ALTER` on the existing procedure. I chose to not do this since a) a bug in my new migration would require both a code rollback and a DB rollback (instead of just a code rollback) and b) I wanted to set a safer pattern for future, similar changes.

src/NuGetGallery/Migrations/SqlResourceMigration.cs
```diff
@@ -9,16 +10,38 @@ namespace NuGetGallery.Migrations
...
+        {
+            if (_sqlFileDown == null)
+            {
+                return;
+            }
+
+            Stream stream = typeof(ExecuteELMAHSql).Assembly.GetManifestResourceStream(sqlFile);
             using (var streamReader = new StreamReader(stream))
             {
                 var statements = streamReader.ReadToEnd().Split(Go, StringSplitOptions.RemoveEmptyEntries);
```
agr: This looks super dangerous. Substring "GO" may appear anywhere.

agr: I would also prefer to see [`File.ReadAllText`](https://msdn.microsoft.com/en-us/library/ms143368(v=vs.110).aspx) used here, but it makes no functional difference.

joelverhagen: This is existing code.

True, it's not ideal, but any issue would be quickly caught by a manual test or by execution against DEV, INT. Keep in mind this isn't dynamic input being split but rather static files checked in to the repo. 

Perhaps splitting by `\r\nGO;\r\n` would be sufficient?

agr: Code tends to change and we have a (very subtle) requirement that the T-SQL code should not contain "GO" anywhere except on the batch boundaries. I agree though that in 99% of cases of unexpected "GO" the result would be an invalid SQL query, which would be noticed.

New lines might improve the situation, but would not prevent it completely: SQL allows multi-line strings, so one may have `GO` on a separate line inside string literal. Semicolon is not needed though.

src/NuGetGallery/Infrastructure/AddPackageLicenseReport2.Up.sql
```diff
@@ -0,0 +1,133 @@
...
+
+CREATE PROCEDURE [dbo].[AddPackageLicenseReport2]
+(
+     @sequence INT,
+     @packageId NVARCHAR(128),
+     @version NVARCHAR(64),
+     @reportUrl NVARCHAR(256) = NULL,
+     @licenseNames LicenseNamesList READONLY,
+     @comment NVARCHAR(256) = NULL,
+	 @whatIf BIT = 0
```
agr: indentation

src/NuGetGallery/Infrastructure/AddPackageLicenseReport2.Up.sql
```diff
@@ -0,0 +1,133 @@
...
+        [Key] INT NOT NULL PRIMARY KEY
+    )
+
+    -- Check input
+
+    IF @sequence  IS NULL RETURN
+    IF @packageId IS NULL RETURN
+    IF @version   IS NULL RETURN
+
+	DECLARE @returnTable TABLE
```
agr: more indentation :)

src/NuGetGallery/Infrastructure/AddPackageLicenseReport2.Up.sql
```diff
@@ -0,0 +1,133 @@
...
+	(
+		[Action] nvarchar(128),
+		[Table] nvarchar(128),
+		[Value] nvarchar(2000)
```
agr: `nvarchar(max)`? What are we saving by specifying length here?

joelverhagen: I had that for a while, but was unsure of the support for `nvarchar(max)` for table variables. After some reading, I think it's fine.

src/NuGetGallery/Migrations/SqlResourceMigration.cs
```diff
@@ -9,16 +10,38 @@ namespace NuGetGallery.Migrations
...
+        }
+
+        private void ExecuteSqlFile(string sqlFile)
+        {
+            if (_sqlFileDown == null)
+            {
+                return;
+            }
+
+            Stream stream = typeof(ExecuteELMAHSql).Assembly.GetManifestResourceStream(sqlFile);
```
chenriksson: use var?

joelverhagen: Fixed.

src/NuGetGallery/Infrastructure/AddPackageLicenseReport2.Up.sql
joelverhagen: Fixed.

src/NuGetGallery/Infrastructure/AddPackageLicenseReport2.Up.sql
joelverhagen: Fixed.

src/NuGetGallery/Infrastructure/AddPackageLicenseReport2.Up.sql
joelverhagen: Fixed.

src/NuGetGallery/Migrations/SqlResourceMigration.cs
joelverhagen: Improved delimited marginally and left comment about the caveats.

PR #4437, opened 2017-07-20
Add support for Paket guidance in display package.

isaacabraham:

PR #4638, opened 2017-09-09
Add support for Paket CLI.

isaacabraham:
Revised PR that replaces #4437 to add support for Paket as a third-party NuGet client.

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -23,6 +23,15 @@
             CommandPrefix = "> ",
             InstallPackageCommand = string.Format("dotnet add package {0} --version {1}", Model.Id, Model.Version)
         },
+
+        new ThirdPartyPackageManagerViewModel()
+        {
+            Id = "paket-cli",
+            Name = "Paket CLI",
+            CommandPrefix = "> ",
+            InstallPackageCommand = string.Format("paket add {0} --version {1}", Model.Id, Model.Version),
+            ContactUrl = "https://fsprojects.github.io/Paket/"
+        },
     };
 }
 @section SocialMeta {


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -23,6 +23,15 @@
...
             InstallPackageCommand = string.Format("dotnet add package {0} --version {1}", Model.Id, Model.Version)
         },
+
+        new ThirdPartyPackageManagerViewModel()
+        {
+            Id = "paket-cli",
+            Name = "Paket CLI",
+            CommandPrefix = "> ",
+            InstallPackageCommand = string.Format("paket add {0} --version {1}", Model.Id, Model.Version),
+            ContactUrl = "https://fsprojects.github.io/Paket/"
```
loic-sharma: Is this the best support URL for Paket? I can't seem to find anything myself on Paket's website - would a link to their GitHub issues be better?

forki: What are you looking for exactly? The docs are very large maybe we can highlight stuff more when you tell us what you are looking for.

anangaur: IMO, something like https://www.nuget.org/policies/Contact would be better but I would not block on this link. If in future there is a better page that summarizes all ways to reach out to the maintainers in case there is a problem/query - please update the link here.

loic-sharma:
Thanks! This pull request been merged and will be deployed as part of our next deployment. You can track this in issue https://github.com/NuGet/NuGetGallery/issues/4668.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
forki: understood. Here it is: https://github.com/NuGet/NuGetGallery/pull/4671

Issue #4379, opened 2017-07-12
[Prefix] Restrict uploads for prefix matching Ids


shishirx34:
Modify the upload flow to check for prefixes when matching Ids.

PR #4639, opened 2017-09-10
[Prefix] Restrict uploads for prefix matching IDs

shishirx34:
This PR addresses #4379. Now on a package push/upload we have additional validation for a new package id. The package ID matching a reserved namespace will allowed to be uploaded only if it that user is the owner of the namespace, or the namespace is shared with public to upload. Also added tests.

Screenshots:

Upload flow:
image

Command line push:
image


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -113,7 +113,7 @@ public void EnsureValid(PackageArchiveReader packageArchiveReader)
         /// <exception cref="InvalidPackageException">
         /// This exception will be thrown when a package metadata property violates a data validation constraint.
         /// </exception>
-        public async Task<Package> CreatePackageAsync(PackageArchiveReader nugetPackage, PackageStreamMetadata packageStreamMetadata, User user, bool commitChanges = true)
+        public async Task<Package> CreatePackageAsync(PackageArchiveReader nugetPackage, PackageStreamMetadata packageStreamMetadata, User user, bool commitChanges = true, bool isVerified = false)
```
joelverhagen: Nit, don't add default param. We own all the code paths calling this method so we should explicitly state the `isVerified` parameter. Makes it a lot easier to understand the code.

src/NuGetGallery/Services/IReservedNamespaceService.cs
```diff
@@ -45,6 +45,15 @@ public interface IReservedNamespaceService
         Task DeleteOwnerFromReservedNamespaceAsync(string prefix, string username);
 
         /// <summary>
+        /// Add the specified package registration to the reserved namespace
+        /// </summary>
+        /// <param name="prefix">The reserved namespace to modify</param>
+        /// <param name="pr">The package registration to be added</param>
+        /// <param name="commitChanges">Commit changes to the database, defaults to true</param>
+        /// <returns>Awaitable Task</returns>
+        Task AddPackageRegistrationToNamespaceAsync(string prefix, PackageRegistration pr, bool commitChanges = true);
```
joelverhagen: The other method `AddReservedNamespaceAsync` is failing locally with:

```
LINQ to Entities does not recognize the method 'Boolean StartsWith(System.String, System.StringComparison)' method, and this method cannot be translated into a store expression.
```

From this code in `ReservedNamespaceService.cs`:
```csharp
prefixMatch = dbPrefix => dbPrefix.Value.StartsWith(prefix, StringComparison.OrdinalIgnoreCase);
```

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -394,17 +398,32 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+                                subject: nuspec.GetId(),
                                 requestedActions: NuGetScopes.PackagePush))
                             {
                                 // User cannot push a new package ID as the API key scope does not allow it
                                 return new HttpStatusCodeWithBodyResult(HttpStatusCode.Unauthorized, Strings.ApiKeyNotAuthorized);
                             }
+
+                            // For new package id verify that the user owns any matching namespace 
+                            // or the user is allowed to push to a matching namespace
+                            var anyoneCanPushToThisNamespace = matchingNamespaces.Count == 0
```
joelverhagen: Is it possible to have the following data?

Prefix | Anyone can push
------ | -------------------
`A.`   | false
`A.B.` | true
`A.B.C.` | false

It looks like it is possible if they are added in the order stated above (since `A.B.` is not a prefix of `A.` ✔️  and `A.B.C.` is not a prefix of `A.B.` or `A.` ✔️). Because if this data is allowed and the guy who is pushing does not own any of the reserved namespaces himself, then pushing `A.B.C.D` would be allowed when it really shouldn't.

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -196,6 +196,32 @@ public class ReservedNamespaceService : IReservedNamespaceService
             }
         }
 
+        // Add the Package registration entry to the reserved namespace, the provided package registration
+        // should be an entry in the database or an entity from memory to be referenced.
+        public async Task AddPackageRegistrationToNamespaceAsync(string prefix, PackageRegistration pr, bool commitChanges = true)
```
joelverhagen: In this whole service I only see `IsPrefix` used once -- how is it supposed to be used?

src/NuGetGallery/Services/IReservedNamespaceService.cs
joelverhagen: You can remove the `StringComparison.OrdinalIgnoreCase` parts because our SQL Server collation is case-insensitive.

src/NuGetGallery/Controllers/ApiController.cs
joelverhagen: In general the long prefix should be evaluated first right?

joelverhagen: In general the longer prefix should be evaluated first right?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -307,21 +310,43 @@ public virtual JsonResult UploadPackageProgress()
...
+
+                    return Json(400, new string[] {
                         string.Format(
                             CultureInfo.CurrentCulture,
                             Strings.UploadPackage_MinClientVersionOutOfRange,
                             nuspec.GetMinClientVersion()) });
                 }
 
-                var packageRegistration = _packageService.FindPackageRegistrationById(nuspec.GetId());
+                var id = nuspec.GetId();
```
joelverhagen: There seems to be new, duplicated logic between `PackagesController` and `ApiController`. Can this be shared?

joelverhagen: If the logic was shared somehow, then you would not need to duplicate as much testing in these two controllers.

src/NuGetGallery/Strings.Designer.cs
```diff
@@ -1138,6 +1138,15 @@ public class Strings {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to This package ID has been reserved. Please request access to push to this reserved namespace from the owner of this reserved prefix or re-upload the package with a different ID.
+        /// </summary>
+        public static string UploadPackage_IdNamespaceConflict {
+            get {
```
skofman1: Was this message provided by Daniel? Please sync with him on wording

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -394,17 +398,32 @@ public virtual Task<ActionResult> CreatePackagePost()
                         }
 
                         // Ensure that the user can push packages for this partialId.
-                        var packageRegistration = PackageService.FindPackageRegistrationById(nuspec.GetId());
+                        var id = nuspec.GetId();
+                        var packageRegistration = PackageService.FindPackageRegistrationById(id);
+                        var matchingNamespaces = ReservedNamespaceService.GetReservedNamespacesForId(id);
```
skofman1: There is a lot of code/logic duplication between ApiController changes and PackageContoller. I recommend moving as much logic as possible to ReservedNamespaceService. It can expose a single method: IsPushAllowed(User user, string id)

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -460,10 +479,14 @@ public virtual Task<ActionResult> CreatePackagePost()
                         };
 
                         var package = await PackageService.CreatePackageAsync(
-                            packageToPush, 
+                            packageToPush,
```
skofman1: Here you add "knowledge" about reserved namespaces to PackageService. Consider keeping all this "knowledge" in a single class: 
ReservedNamespaceService will update the registration table. PackageService will not change. This way all the logic remains in ReservedNamespaceService .

joelverhagen: I spoke to @shishirx34 offline a bit about this -- there are certainly multiple valid ways of looking at this but the way we landed on at the time was that `PackageService` is the owner of both `Packages` and `PackageRegistrations` table.

src/NuGetGallery/Strings.Designer.cs
shishirx34: Yes plan to run this by Daniel.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: I do not think there was any specific requirement about such complications. I think we should keep it simple and if there is a public namespace that matches for any ID, we allow to push. In such cases, `A.B.C.D` will be a public namespace.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: Yes, like @joelverhagen stated, we let the service be the delegate to take care of its own entities and that is how we split the services vertically, I think we have kept this process true through out our codebase and I think we should keep it consistent. 

Also, I think we aren't necessarily adding knowledge of reserved namespaces, its the property of a `PackageRegistration`, this package registration property could very well be set in the future by other scenarios and not necessarily by the reserved namespaces, as such it should be a responsibility of the `PackageService`

src/NuGetGallery/Services/ReservedNamespaceService.cs
shishirx34: Yes, it is supposed to be used to identify the namespaces
1. The namespace could be absolute or a prefix, if its a prefix we shouldn't allow a more liberal namespace to be reserved. 
2. Also, use the prefix property to mark the existing package registrations as verified, however using the `startsWith` function makes the use of `IsPrefix` redundant in this case. 
3. Another case would be to check for returning the matching namespaces to allow/disallow the Id during upload, I missed adding this earlier. Adding it now, good catch. Will add tests for the method `GetReservedNamespacesForId`

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: Good idea. Will do.

src/NuGetGallery/Controllers/PackagesController.cs
shishirx34: As per @skofman1's suggestion I moved the following code to the `ReservedNamespaceService`. However, I think we will still need the tests, since these controller tests ensure that the correct error/flow is persisted from Package upload/push view.

src/NuGetGallery/Services/ReservedNamespaceService.cs
shishirx34: Added relevant tests.

shishirx34:
@skofman1, @joelverhagen - addressed comments.

src/NuGetGallery/Controllers/ApiController.cs
skofman1: According to the DB schema defined, in order to mark a registration as reserved, there should be changes to two table: PackageRegistrations, PackRegPrefixMap (as defined in the design). Updating only one of the tables, will break contract and get the DB into an invalid state. Hence, updating each of the tables in a different service class, and forcing the Controller to be aware of this logic, makes the code hard to modify, and it might be easily broken by future devs.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -394,17 +399,28 @@ public virtual Task<ActionResult> CreatePackagePost()
                         }
 
                         // Ensure that the user can push packages for this partialId.
-                        var packageRegistration = PackageService.FindPackageRegistrationById(nuspec.GetId());
+                        var id = nuspec.GetId();
+                        var packageRegistration = PackageService.FindPackageRegistrationById(id);
+                        var isPushAllowed = ReservedNamespaceService
```
skofman1: >isPushAllowed [](start = 28, length = 13)

The return values of IsPushAllowed are relevant only in the case packageResgistration == null. 
Moving this code inside the if statement will improve perf in case a package version is pushed.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -25,6 +25,7 @@
 using NuGetGallery.Packaging;
 using NuGetGallery.Security;
 using PackageIdValidator = NuGetGallery.Packaging.PackageIdValidator;
+using System.Collections.Generic;
```
cristinamanum: At the beginning of using directives.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -394,17 +399,28 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+                        var packageRegistration = PackageService.FindPackageRegistrationById(id);
+                        var isPushAllowed = ReservedNamespaceService
+                            .IsPushAllowed(id, user, out IReadOnlyCollection<ReservedNamespace> userOwnedNamespaces);
+
                         if (packageRegistration == null)
                         {
                             // Check if API key allows pushing a new package id
                             if (!ApiKeyScopeAllows(
-                                subject: nuspec.GetId(), 
+                                subject: nuspec.GetId(),
```
cristinamanum: nuspec.GetId already calculated above in var id = ....

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -460,11 +476,15 @@ public virtual Task<ActionResult> CreatePackagePost()
                         };
 
                         var package = await PackageService.CreatePackageAsync(
-                            packageToPush, 
+                            packageToPush,
                             packageStreamMetadata,
                             user,
+                            isVerified: userOwnedNamespaces.Any(),
```
cristinamanum: Is it enough to have a task existing only? Even if that task ends in faulted state?

src/NuGetGallery/Controllers/ApiController.cs
joelverhagen: If this is an unnecessary complication, the "add namespace" code should block this sort of data (`A.B.C.` in my example) from being persisted since it is meaningless or unsupported data.

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -196,6 +196,32 @@ public class ReservedNamespaceService : IReservedNamespaceService
...
+            if (pr == null)
+            {
+                throw new ArgumentNullException(nameof(pr));
+            }
+
+            var namespaceToModify = FindReservedNamespaceForPrefix(prefix)
+                ?? throw new InvalidOperationException(string.Format(
+                    CultureInfo.CurrentCulture, Strings.ReservedNamespace_NamespaceNotFound, prefix));
+
+            namespaceToModify.PackageRegistrations.Add(pr);
```
cristinamanum: How is the local namespaceToModify used?

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -196,6 +196,32 @@ public class ReservedNamespaceService : IReservedNamespaceService
...
+
+            if (commitChanges)
```
cristinamanum: On commit will the changes saved to database? What are the cases when the method will be invoked with commitChnages = false. Maybe it would be good to add few comments at the method level?

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: Not my change, but I see this is littered through out the code. I will update only the `id` to keep the changes minimal.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: The task shouldn't enter the faulted state. If the code reaches here the push should be allowed. But you are right, to make it foolproof, I added additional checks to make sure that it doesn't break.

src/NuGetGallery/Services/ReservedNamespaceService.cs
shishirx34: added comments.

src/NuGetGallery/Services/ReservedNamespaceService.cs
shishirx34: added comments.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: Sounds reasonable. Would it be ok with you if I add that as part of admin page changes, since that is a more related code. I would keep the unrelated changes here to minimal. I have made a note for this.

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -196,6 +196,40 @@ public class ReservedNamespaceService : IReservedNamespaceService
...
+        /// This method fetches the reserved namespace matching the prefix and adds the 
+        /// package registration entry to the reserved namespace, the provided package registration
+        /// should be an entry in the database or an entity from memory to be committed.
+        /// </summary>
+        /// <param name="prefix">The prefix value of the reserved namespace to modify</param>
+        /// <param name="pr">The package registration entity entry to be added.</param>
+        /// <param name="commitChanges">Flag to commit the modifications to the database, if set to false
+        /// the caller of this method should take care of saving changes for entities context.</param>
+        /// <returns>Awaitable task</returns>
+        public async Task AddPackageRegistrationToNamespaceAsync(string prefix, PackageRegistration pr, bool commitChanges = true)
```
loic-sharma: >pr [](start = 100, length = 2)

Please  rename `pr` to something more readable, like `registration` or `packageRegistration`

shishirx34:
Addressed the feedback. @skofman1 @joelverhagen take look. 

It seems there are currently broken tests in `dev`, holding off on resolving merge conflicts until the tests are fixed.

src/NuGetGallery/Controllers/ApiController.cs
joelverhagen: Yup, that's fine! Thanks. I just want to avoid some future day where we have existing that conflicts with newly defined requirements.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            _reservedNamespaceService.IsPushAllowed(id, user, out IReadOnlyCollection<ReservedNamespace> userOwnedNamespaces);
+
+            var shouldMarkIdVerified = userOwnedNamespaces != null && userOwnedNamespaces.Any();
+            var package = await _packageService.CreatePackageAsync(
+                nugetPackage,
+                packageStreamMetadata,
+                user,
+                isVerified: shouldMarkIdVerified,
+                commitChanges: false);
```
dtivel: Why is this hard-coded to `false` and not use the method argument `commitChanges`?

tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
```diff
@@ -486,7 +539,79 @@ public class TheCreatePackageAction
...
+                controller.MockPackageService.Verify(x => x.CreatePackageAsync(It.IsAny<PackageArchiveReader>(), It.IsAny<PackageStreamMetadata>(), It.IsAny<User>(), It.IsAny<bool>(), false));
+                controller.MockEntitiesContext.VerifyCommitted();
+            }
+
+            [Fact]
+            public async Task WillCreatePackageIfIdMatchesAnOwnedNamespace()
+            {
+                // Arrange
+                var user1 = new User { Key = 1, Username = "random1" };
+                var user2 = new User { Key = 2, Username = "random2" };
```
dtivel: This user isn't used anywhere.

tests/NuGetGallery.Facts/Services/ReservedNamespaceServiceFacts.cs
```diff
@@ -497,6 +564,149 @@ public class TheDeleteOwnerFromReservedNamespaceAsyncMethod
...
+                var service = new TestableReservedNamespaceService(reservedNamespaces: testNamespaces, users: testUsers);
+
+                var isPushAllowed = service.IsPushAllowed(id, lastUser, out IReadOnlyCollection<ReservedNamespace> matchingNamespaces);
+                Assert.Empty(matchingNamespaces);
+                Assert.True(isPushAllowed);
+            }
+
+            [Theory]
+            [InlineData("Microsoft.Aspnet")]
+            [InlineData("Microsoft.Aspnet.Newpackage")]
```
dtivel: Somewhere do you have tests for case-sensitivity?

src/NuGetGallery/Services/PackageUploadService.cs
shishirx34: You are right, I missed it. I will create a separate PR to fix it, this was merged before.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            _packageService = packageService;
+            _reservedNamespaceService = reservedNamespaceService;
+        }
+
+        public async Task<Package> GeneratePackageAsync(string id, PackageArchiveReader nugetPackage, PackageStreamMetadata packageStreamMetadata, User user, bool commitChanges)
+        {
+            _reservedNamespaceService.IsPushAllowed(id, user, out IReadOnlyCollection<ReservedNamespace> userOwnedNamespaces);
+
+            var shouldMarkIdVerified = userOwnedNamespaces != null && userOwnedNamespaces.Any();
```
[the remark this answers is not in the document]
shishirx34: The check of push allowed is an extra unnecessary check thats why I ignored it. Don't mind adding it for completion.

tests/NuGetGallery.Facts/Services/ReservedNamespaceServiceFacts.cs
shishirx34: Its kind of tough to unit test case-sensitivity with regards to entityframework. Our DB collation is set to be case-insensitive. And testing this in memory with mocks seems useless since it will be overridden in the test framework.

PR #4664, opened 2017-09-12
Add package status column and migration

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/4644.


src/NuGetGallery/Migrations/201709111714021_AddPackageStatusKey.cs
```diff
@@ -0,0 +1,61 @@
...
+    SET PackageStatusKey = 1
+    WHERE Deleted = 1 AND PackageStatusKey = 0
+
+    SELECT @UpdateCount = @@ROWCOUNT
+
+    WAITFOR DELAY @Delay
+END
+");
+
+            Sql("IF NOT EXISTS(SELECT * FROM sys.indexes WHERE name = 'nci_wi_Packages_PackageStatusKeyListed' AND object_id = OBJECT_ID('Packages')) CREATE INDEX [nci_wi_Packages_PackageStatusKeyListed] ON [dbo].[Packages] ([PackageStatusKey] ASC, [Listed] ASC) INCLUDE ([Description], [FlattenedDependencies], [IsPrerelease], [PackageRegistrationKey], [Tags], [Version]) WITH (ONLINE = ON)");
```
loic-sharma: >Sql [](start = 12, length = 3)

Would adding the index first and then updating `PackageStatusKey` reduce the migration time?

joelverhagen: We purposefully do not have an index with both `Deleted` and `PackageStatusKey` so not noticeably.

The query...
```sql
UPDATE TOP (@PerIteration) dbo.Packages
SET PackageStatusKey = 1
WHERE Deleted = 1 AND PackageStatusKey = 0
```

... will utilize the `nci_wi_Packages_Deleted` index, which is enough for this purpose.

src/NuGetGallery.Core/Entities/Package.cs
```diff
@@ -209,6 +209,11 @@ public Package()
 
         public bool Deleted { get; set; }
 
+        /// <summary>
+        /// The package status key, referring to the <see cref="PackageStatus"/> enum.
+        /// </summary>
+        public PackageStatus PackageStatusKey { get; set; }
```
skofman1: Reading an enum directly from DB might introduce a forward-compatibility issue in the future. Consider the following scenario: we introduce a new status key value (3). We deploy code that writes this value, and it appears in the DB. Than we recognize an issue and rollback. The old code can't parse packagestatus=3, and crashes.
To overcome this, you should read an int from DB, and transform it later to enum. Also, the enum should have "Unsupported" value, and application should handle it correctly.

dtivel: After reading [this blog post](https://blogs.msdn.microsoft.com/efdesign/2011/06/29/enumeration-support-in-entity-framework/), it sounds like Entity Framework enum handling follows the same rules as C# enum handling; in particular, EF will not fail when handling a value that does not map to a defined enum member.

Excerpt thereof:

"This means that the declaration of Enums in EDM will have the same elements that they have in C#, including the optionally specified underlying type, case sensitive member names, optional values, etc., and that the same rules are used to choose values not specified, e.g. the default underlying type is also Int32 and the implicit value of members is the previous member plus one, starting with zero for the first member."

It's worth testing that our current version of EF still matches the original design intent, but at least this scenario doesn't sound like it will be a concern.

tests/NuGetGallery.Core.Facts/PackageStatusFacts.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Xunit;
+
+namespace NuGetGallery
+{
+    public class PackageStatusFacts
```
dtivel: Nice

src/NuGetGallery.Core/Entities/Package.cs
joelverhagen: Unexpected enum values are handled without exception by Entity Framework. For our application, there are two sides to this concern I think:

1. Query side -- what will our queries do with unexpected enum values? Well, most queries "opt in" to a specific `PackageStatus` instead of "opting out" a set of package statuses. Excluding a set of package statuses is the part that is liable to have issues with unexpected enum values. For example, we have `WHERE PackageStatusKey = 0` (`package.PackageStatusKey == PackageStatus.Available`) is most places where this column is considered. The cases where `!=` is used unexpected enum values must be acceptable.
1. Entity mapping -- what does EF do with an int value that doesn't have a corresponding enum value? Well, it just casts the int to the enum. This is perfectly acceptable in C#. In general, any in-memory operations on enums must consider enum values that do not have names, e.g. with `default` switch clause. This is regardless of EF involvement.

skofman1: It's great that EF supports enums in this way! Now all that's left, like Joel explained, is to make sure our code is handling forward compatibility correctly.

PR #4665, opened 2017-09-12
Use package status instead of deleted bit

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/4645.

Depends on https://github.com/NuGet/NuGetGallery/pull/4664. I will change the base branch once that one goes in.

### DIFF
--- a/src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedPackage.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedPackage.cs
@@ -44,6 +44,7 @@ namespace NuGetGallery.Auditing.AuditedEntities
         public int? UserKey { get; private set; }
         public bool Deleted { get; private set; }
         public bool HasReadMe { get; private set; }
+        public int PackageStatusKey { get; private set; }
 
         public static AuditedPackage CreateFrom(Package package)
         {
@@ -83,8 +84,11 @@ namespace NuGetGallery.Auditing.AuditedEntities
                 Key = package.Key,
                 MinClientVersion = package.MinClientVersion,
                 UserKey = package.UserKey,
+#pragma warning disable CS0612 // Type or member is obsolete
                 Deleted = package.Deleted,
-                HasReadMe = package.HasReadMe
+#pragma warning restore CS0612 // Type or member is obsolete
+                HasReadMe = package.HasReadMe,
+                PackageStatusKey = (int)package.PackageStatusKey,
             };
         }
     }
--- a/src/NuGetGallery.Core/SemVerLevelKey.cs
+++ b/src/NuGetGallery.Core/SemVerLevelKey.cs
@@ -8,16 +8,47 @@ using NuGet.Versioning;
 
 namespace NuGetGallery
 {
-    // Hard-coded for now, but we can easily expand to use an additional Sql table to join with 
-    // when supporting additional semVerLevel's is needed.
-
     /// <summary>
     /// Helper class to use to determine the SemVer level of a package version.
     /// </summary>
     public static class SemVerLevelKey
     {
-        public static readonly string SemVerLevel2 = "2.0.0";
+        public const string SemVerLevel2 = "2.0.0";
         private static readonly NuGetVersion _semVer2Version = NuGetVersion.Parse(SemVerLevel2);
+
+        /// <summary>
+        /// Since <see cref="Unknown"/> is not a constant, we must explicitly construct the expression to force
+        /// that value to be considered a constant. If <see cref="Unknown"/> is not considered a constant, the
+        /// expression that is converted to SQL by Entity Framework is less efficient.
+        /// </summary>
+        private static readonly Lazy<Expression<Func<Package, bool>>> _isUnknown = new Lazy<Expression<Func<Package, bool>>>(() =>
+        {
+            var parameter = Expression.Parameter(typeof(Package), "p");
+            var property = Expression.Property(parameter, nameof(Package.SemVerLevelKey));
+            return Expression.Lambda<Func<Package, bool>>(
+                Expression.Equal(
+                    property,
+                    Expression.Constant(Unknown)),
+                parameter);
+        });
+
+        private static readonly Lazy<Expression<Func<Package, bool>>> _isSemVer2 = new Lazy<Expression<Func<Package, bool>>>(() =>
+        {
+            /// Since <see cref="Unknown"/> is not a constant, we must explicitly construct the expression to force
+            /// that value to be considered a constant. If <see cref="Unknown"/> is not considered a constant, the
+            /// expression that is converted to SQL by Entity Framework is less efficient.
+            var parameter = Expression.Parameter(typeof(Package), "p");
+            var property = Expression.Property(parameter, nameof(Package.SemVerLevelKey));
+            return Expression.Lambda<Func<Package, bool>>(
+                Expression.Or(
+                    Expression.Equal(
+                        property,
+                        Expression.Constant(Unknown)),
+                    Expression.Equal(
+                        property,
+                        Expression.Constant(SemVer2, typeof(int?)))),
+                parameter);
+        });
 
         /// <summary>
         /// This could either indicate being SemVer1-compliant, or non-SemVer-compliant at all (e.g. System.Versioning pattern).
@@ -28,7 +59,7 @@ namespace NuGetGallery
         /// Indicates being SemVer2-compliant, but not SemVer1-compliant.
         /// This key corresponds to semVerLevel=2.0.0
         /// </summary>
-        public static readonly int SemVer2 = 2;
+        public const int SemVer2 = 2;
 
         /// <summary>
         /// Identifies the SemVer-level of a package based on original version string and dependency versions.
@@ -106,17 +137,27 @@ namespace NuGetGallery
         /// </summary>
         /// <param name="semVerLevel">The SemVer-level string indicating the SemVer-level to comply with.</param>
         /// <returns><c>True</c> if compliant; otherwise <c>false</c>.</returns>
-        public static Expression<Func<Package, bool>> IsPackageCompliantWithSemVerLevel(string semVerLevel)
+        public static Expression<Func<Package, bool>> IsPackageCompliantWithSemVerLevelPredicate(string semVerLevel)
         {
             // Note: we must return an expression that Linq to Entities is able to translate to SQL
             var parsedSemVerLevelKey = ForSemVerLevel(semVerLevel);
 
             if (parsedSemVerLevelKey == SemVer2)
             {
-                return p => p.SemVerLevelKey == Unknown || p.SemVerLevelKey == parsedSemVerLevelKey;
+                return IsSemVer2Predicate();
             }
 
-            return p => p.SemVerLevelKey == Unknown;
+            return IsUnknownPredicate();
+        }
+
+        public static Expression<Func<Package, bool>> IsSemVer2Predicate()
+        {
+            return _isSemVer2.Value;
+        }
+
+        public static Expression<Func<Package, bool>> IsUnknownPredicate()
+        {
+            return _isUnknown.Value;
         }
     }
 }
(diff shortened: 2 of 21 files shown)


src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -87,10 +88,10 @@ public IHttpActionResult GetCount(ODataQueryOptions<V1FeedPackage> options)
...
             var packages = _packagesRepository.GetAll()
                 .Include(p => p.PackageRegistration)
-                .Where(p => p.PackageRegistration.Id.Equals(id, StringComparison.OrdinalIgnoreCase) 
-                        && !p.IsPrerelease 
-                        && !p.Deleted 
-                        && p.SemVerLevelKey == SemVerLevelKey.Unknown);
+                .Where(p => p.PackageRegistration.Id.Equals(id, StringComparison.OrdinalIgnoreCase) &&
+                            !p.IsPrerelease &&
+                            p.PackageStatusKey == PackageStatus.Available)
+                .Where(SemVerLevelKey.IsUnknown());
```
loic-sharma: >IsUnknown [](start = 38, length = 9)

Nit: is there a better name to would indicate that `IsUnknown` returns *an expression* that checks if the package is unknown?

src/NuGetGallery/Services/PackageNamingConflictValidator.cs
```diff
@@ -56,7 +56,7 @@ public bool IdConflictsWithExistingPackageTitle(string registrationId)
             registrationId = registrationId.ToLowerInvariant();
 
             return _packageRepository.GetAll()
-                .Any(p => !p.Deleted && p.Title.ToLower() == registrationId);
+                .Any(p => p.PackageStatusKey != PackageStatus.Deleted && p.Title.ToLower() == registrationId);
```
loic-sharma: >Deleted [](start = 62, length = 7)

Consider adding a comment here that you MUST check more than just `Available` packages as this small nuance is crucial  :)

joelverhagen: Sure

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -470,7 +470,7 @@ where VersionRange.Parse(d.VersionSpec).Satisfies(packageVersion)
                 return;
             }
 
-            if (package.Deleted)
+            if (package.PackageStatusKey == PackageStatus.Deleted)
```
loic-sharma: >package.PackageStatusKey == PackageStatus.Deleted [](start = 16, length = 49)

Should this be `package.PackageStatusKey != PackageStatus.Active`? Or do we want users to list/unlist packages that vail validation?

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -836,11 +836,11 @@ private void ValidatePackageTitle(PackageMetadata packageMetadata)
             // If the last listed package was just unlisted, then we won't find another one
             var latestPackage = FindPackage(
                 packageRegistration.Packages,
-                p => !p.Deleted && p.Listed && p.SemVerLevelKey == SemVerLevelKey.Unknown);
+                p => p.PackageStatusKey == PackageStatus.Available && p.Listed && p.SemVerLevelKey == null);
```
loic-sharma: >p.SemVerLevelKey == null [](start = 82, length = 24)

Shouldn't all these p.SemVerLevelKey be using `SemverLevelKey.IsUnknown`? If so, don't foget lines 843, 854, and 881

src/NuGetGallery/Services/PackageService.cs
joelverhagen: List/unlist on "Validating" packages is supported, per @karann-msft's design.

src/NuGetGallery/Services/PackageService.cs
joelverhagen: I matched existing `SemVerLevelKey` precedent. We cannot use `IsUnknown` inside of a `Func<Package, bool>`.

joelverhagen: I'll see if it's worth refactoring.

src/NuGetGallery/Controllers/ODataV1FeedController.cs
joelverhagen: Renamed to `IsUnknownPredicate`.

src/NuGetGallery/Services/PackageService.cs
joelverhagen: Found a good solution.

src/NuGetGallery/Services/PackageNamingConflictValidator.cs
```diff
@@ -55,8 +55,9 @@ public bool IdConflictsWithExistingPackageTitle(string registrationId)
 
             registrationId = registrationId.ToLowerInvariant();
 
+            // We allow any package status aside from soft-deleted to be considered for ID/title conflicts.
             return _packageRepository.GetAll()
-                .Any(p => !p.Deleted && p.Title.ToLower() == registrationId);
+                .Any(p => p.PackageStatusKey != PackageStatus.Deleted && p.Title.ToLower() == registrationId);
```
skofman1: As discussed in the previous PR, we shouldn't assume the behavior here will be correct for future status keys, hence we should use an explicit subset

skofman1: We can leave it like this for now, and flip the condition if we ever introduce a status that's shouldn't block upload

src/NuGetGallery.Core/SemVerLevelKey.cs
```diff
@@ -8,18 +8,50 @@
 
 namespace NuGetGallery
 {
-    // Hard-coded for now, but we can easily expand to use an additional Sql table to join with 
-    // when supporting additional semVerLevel's is needed.
-
     /// <summary>
-    /// Helper class to use to determine the SemVer level of a package version.
+    /// Helper class to use to determine the SemVer level of a package version. Hard-coded for now, but we can easily
+    /// expand to use an additional Sql table to join with when supporting additional semVerLevel's is needed.
```
shishirx34: nit: ...additional semverLevel as needed.

joelverhagen: I think this comment is unnecessary actually. It's unlikely we will add a table for SemVer level.

src/NuGetGallery/Areas/Admin/Controllers/DeleteController.cs
```diff
@@ -84,7 +84,7 @@ private DeleteSearchResult CreateDeleteSearchResult(Package package)
                     : NuGetVersion.Parse(package.Version).ToNormalizedString(),
                 DownloadCount = package.DownloadCount,
                 Listed = package.Listed,
-                Deleted = package.Deleted
+                Deleted = package.PackageStatusKey == PackageStatus.Deleted,
```
shishirx34: So this code would be going in only after the migration from 'Deleted' => 'Package.StatusKey' is done?

src/NuGetGallery.Core/SemVerLevelKey.cs
joelverhagen: Comment revised.

src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -87,10 +88,10 @@ public IHttpActionResult GetCount(ODataQueryOptions<V1FeedPackage> options)
...
             var packages = _packagesRepository.GetAll()
                 .Include(p => p.PackageRegistration)
-                .Where(p => p.PackageRegistration.Id.Equals(id, StringComparison.OrdinalIgnoreCase) 
-                        && !p.IsPrerelease 
-                        && !p.Deleted 
-                        && p.SemVerLevelKey == SemVerLevelKey.Unknown);
+                .Where(p => p.PackageRegistration.Id.Equals(id, StringComparison.OrdinalIgnoreCase) &&
+                            !p.IsPrerelease &&
+                            p.PackageStatusKey == PackageStatus.Available)
+                .Where(SemVerLevelKey.IsUnknownPredicate());
```
shishirx34: nit: the operators on subsequent lines rather than at the end, just for consistency.

src/NuGetGallery/Controllers/ODataV2FeedController.cs
```diff
@@ -188,9 +188,9 @@ public class ODataV2FeedController
         {
             var packages = _packagesRepository.GetAll()
                 .Include(p => p.PackageRegistration)
-                .Where(p => p.PackageRegistration.Id.Equals(id, StringComparison.OrdinalIgnoreCase)
-                            && !p.Deleted)
-                .Where(SemVerLevelKey.IsPackageCompliantWithSemVerLevel(semVerLevel));
+                .Where(p => p.PackageStatusKey == PackageStatus.Available &&
+                            p.PackageRegistration.Id.Equals(id, StringComparison.OrdinalIgnoreCase))
```
shishirx34: nit: conditional operator on next line, maybe?

src/NuGetGallery/Queries/AutoCompleteDatabasePackageVersionsQuery.cs
```diff
@@ -14,7 +14,7 @@ public class AutoCompleteDatabasePackageVersionsQuery
         private const string _sqlFormat = @"SELECT p.[Version]
 FROM Packages p (NOLOCK)
 	JOIN PackageRegistrations pr (NOLOCK) on pr.[Key] = p.PackageRegistrationKey
-WHERE p.[Deleted] <> 1 AND p.[Listed] = 1 AND {0} AND pr.ID = {{0}}
+WHERE p.[PackageStatusKey] = 0 AND p.[Listed] = 1 AND {0} AND pr.ID = {{0}}
```
shishirx34: may be use the `PackageStatus.Available` instead of the hard coded value here?

src/NuGetGallery/Areas/Admin/Controllers/DeleteController.cs
joelverhagen: Yes.

src/NuGetGallery/Controllers/ODataV1FeedController.cs
joelverhagen: I've seen both in similar frequency.

Regex | Hit count
--- | ---
` {4,}&&` | 91
` {4,}\|\|` | 30
`&&\s*$` | 99
`\|\|\s*$` | 21

src/NuGetGallery/Queries/AutoCompleteDatabasePackageVersionsQuery.cs
joelverhagen: Good idea!

src/NuGetGallery/Controllers/ODataV1FeedController.cs
joelverhagen: Spoke offline. I like the logical operator at the beginning of the line better.

chenriksson:
The migration should correct existing data, to set PackageStatus.Deleted in cases of Package.Deleted. However, there's a window between when the migration ran and when the deployment finishes where packages could be deleted old-style. We should probably verify/correct data just after deployment as well, right?

joelverhagen:
> The migration should correct existing data, to set PackageStatus.Deleted in cases of Package.Deleted. However, there's a window between when the migration ran and when the deployment finishes where packages could be deleted old-style. We should probably verify/correct data just after deployment as well, right?

That's right. I plan that as part of the deployment steps. The alternative is to say "don't delete things" during that window. I prefer the former since it does not disrupt the support requests flow.

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedPackage.cs
```diff
@@ -86,6 +87,7 @@ public static AuditedPackage CreateFrom(Package package)
 #pragma warning disable CS0612 // Type or member is obsolete
                 Deleted = package.Deleted,
 #pragma warning restore CS0612 // Type or member is obsolete
+                HasReadMe = package.HasReadMe,
```
shishirx34: Why are these changes showing up here? Shouldn't this be in DEV already?

joelverhagen: It is. I added a comma at the end of a line.

PR #4670, opened 2017-09-13
Making branding and footer urls configurable.

ryuyu:
This changeset makes the following updates:
Adds configuration values for Privacy Policy, Terms of Use, About, and Branding

@joelverhagen @karann-msft


src/NuGetGallery/Views/Shared/Gallery/Footer.cshtml
```diff
@@ -32,23 +32,45 @@
...
-                    </div>
-                    <div class="col-md-11 footer-release-info">
+                    <div class="col-md-12 footer-release-info">
                         <p>
-                            &copy; @DateTime.UtcNow.Year .NET Foundation -
-                            <a href="@Url.About()">About</a> -
-                            <a href="@Url.Terms()">Terms of Use</a> -
-                            <a href="@Url.Privacy()">Privacy Policy</a>
+                            @if (!String.IsNullOrEmpty(this.Config.Current.ExternalBrandingMessage))
+                            {
```
skofman1: The default experience for ppl who run Gallery locally should be that .NET foundation logo is shown, term of use, provacy policy, etc. unchanged

src/NuGetGallery/Web.config
```diff
@@ -128,6 +128,11 @@
     <add key="Gallery.ConfirmEmailAddresses" value="true" />
     <add key="Gallery.CollectPerfLogs" value="false" />
     <add key="Gallery.IsODataFilterEnabled" value="false" />
+    <add key="Gallery.ExternalAboutUrl" value="" />
+    <add key="Gallery.ExternalPrivacyPolicyUrl" value="" />
```
skofman1: We don't actually need all this config to brand the gallery for Microsoft. I recommend not adding config that's not required, but your call

joelverhagen: I agree. Couldn't we just have a `Gallery.UseMicrosoftBranding` bool to toggle between the two experiences? Or do we really think this will be an extension point?

ryuyu: I wanted to avoid having the Microsoft branding bleed into the OSS codebase. This way allows us to brand for Microsoft purely through configuration and leave no mention in the codebase itself. Also, all configuration is optional and anybody not specifying will get the default old experience (making this change now).

scottbommarito: I agree with @ryuyu, I can't imagine why anybody besides us would want to use Microsoft branding. Allowing these things to be configurable to any value makes more sense from an open source perspective.

joelverhagen: Sounds fine!

src/NuGetGallery/Views/Shared/Gallery/Footer.cshtml
ryuyu: Done!

ryuyu:
That test should still pass. The default without CustomBranding flag set is now exactly what it was before.

scottbommarito:
The test will pass when it's set to the default values, but we won't set it to the default values.

ryuyu:
ah yes. Good point. Test should be removed then. I'll do that.

scottbommarito:
I think you can keep some of the parts of the test, such as the copyright, privacy policy, and terms checks. You just need to rename it and remove the references to DotNetFoundation.

ryuyu:
Test won't be doing the same thing at all any more. I would prefer to disable it altogether as the values it is checking are configurable now. Copyright will be configured as will terms of use url and privacy policy url.

src/NuGetGallery/Web.config
```diff
@@ -128,6 +128,12 @@
     <add key="Gallery.ConfirmEmailAddresses" value="true" />
     <add key="Gallery.CollectPerfLogs" value="false" />
     <add key="Gallery.IsODataFilterEnabled" value="false" />
+    <add key="Gallery.UseCustomBranding" value="false" />
```
skofman1: UseCustomBranding config is not needed. You can simply check if the other values are provided (like you already do)

src/NuGetGallery/Views/Shared/Gallery/Footer.cshtml
```diff
@@ -32,24 +32,68 @@
...
+                                }
+                                    else
+                                    {
+                                        @:@(this.Config.Current.ExternalBrandingMessage) -
+                                }
+                                }
+                                @if (!String.IsNullOrEmpty(this.Config.Current.ExternalAboutUrl))
+                                {
+                                    @:<a href="@this.Config.Current.ExternalAboutUrl">About</a> -
+                            }
```
loic-sharma: The closing braces aren't well aligned here.

ryuyu: I think I fixed this. Razor editor did some weird autoformatting

skofman1:
Looks great! please also add functional tests to make sure nuget.org is configured correctly.

Issue #4646, opened 2017-09-11
Upload workflow change (in gallery code)


joelverhagen:
- Enqueue to SB
- Set package status to Available initially and Validating after some period of stabilization in PROD
- Abstraction for `git clone` code vs. NuGet.org code

joelverhagen:
This is done. It will be validated at the end of the feature.

PR #4677, opened 2017-09-14
Enqueue to validation queue when a package is created

joelverhagen:
Depends on PR https://github.com/NuGet/NuGetGallery/pull/4665.

Progress on https://github.com/NuGet/NuGetGallery/issues/4646.

### DIFF
--- a/src/NuGetGallery.Core/PackageStatus.cs
+++ b/src/NuGetGallery.Core/PackageStatus.cs
@@ -15,5 +15,19 @@ namespace NuGetGallery
         /// version are still reserved.
         /// </summary>
         Deleted = 1,
+
+        /// <summary>
+        /// The package is in the process of being validated. The validation is being done asynchronously. When the
+        /// validation is complete it will move to the <see cref="Available"/>, <see cref="Deleted"/>, or
+        /// <see cref="FailedValidation"/> state. When packages are in this state, they are not available but the
+        /// package ID and version are still reserved.
+        /// </summary>
+        Validating = 2,
+
+        /// <summary>
+        /// The package has completed validation and the result has come back as a failure. When packages are in this
+        /// state, they are not available but the package ID and version are still reserved.
+        /// </summary>
+        FailedValidation = 3,
     }
 }
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -17,6 +17,8 @@ using Autofac;
 using Autofac.Core;
 using Elmah;
 using Microsoft.WindowsAzure.ServiceRuntime;
+using NuGet.Services.ServiceBus;
+using NuGet.Services.Validation;
 using NuGetGallery.Areas.Admin;
 using NuGetGallery.Areas.Admin.Models;
 using NuGetGallery.Auditing;
@@ -208,6 +210,11 @@ namespace NuGetGallery
                 .As<IPackageUploadService>()
                 .InstancePerLifetimeScope();
 
+            builder.RegisterType<ValidationService>()
+                .AsSelf()
+                .As<IValidationService>()
+                .InstancePerLifetimeScope();
+
             builder.RegisterType<ReadMeService>()
                 .AsSelf()
                 .As<IReadMeService>()
@@ -287,6 +294,8 @@ namespace NuGetGallery
                     break;
             }
 
+            RegisterAsynchronousValidation(builder, configuration);
+
             RegisterAuditingServices(builder, defaultAuditingService);
 
             RegisterCookieComplianceService(builder, configuration, diagnosticsService);
@@ -322,6 +331,38 @@ namespace NuGetGallery
             }
 
             ConfigureAutocomplete(builder, configuration);
+        }
+
+        private void RegisterAsynchronousValidation(ContainerBuilder builder, ConfigurationService configuration)
+        {
+            builder
+                .RegisterType<ServiceBusMessageSerializer>()
+                .As<IServiceBusMessageSerializer>();
+
+            builder
+                .RegisterType<PackageValidationEnqueuer>()
+                .As<IPackageValidationEnqueuer>();
+
+            if (configuration.Current.AsynchronousPackageValidationEnabled)
+            {
+                builder
+                    .RegisterType<AsynchronousPackageValidationInitiator>()
+                    .As<IPackageValidationInitiator>();
+
+                builder
+                    .Register(c => new TopicClientWrapper(
+                        configuration.ServiceBus.Validation_ConnectionString,
+                        configuration.ServiceBus.Validation_TopicName))
+                    .As<ITopicClient>()
+                    .SingleInstance()
+                    .OnRelease(x => x.Close());
+            }
+            else
+            {
+                builder
+                    .RegisterType<ImmediatePackageValidator>()
+                    .As<IPackageValidationInitiator>();
+            }
         }
 
         private static void ConfigureSearch(ContainerBuilder builder, IGalleryConfigurationService configuration)
--- a/src/NuGetGallery/Configuration/AppConfiguration.cs
+++ b/src/NuGetGallery/Configuration/AppConfiguration.cs
@@ -63,6 +63,8 @@ namespace NuGetGallery.Configuration
         /// </summary>
         public bool AzureStorageReadAccessGeoRedundant { get; set; }
 
+        public bool AsynchronousPackageValidationEnabled { get; set; }
+
         /// <summary>
         /// Gets the URI to the search service
         /// </summary>
--- a/src/NuGetGallery/Configuration/ConfigurationService.cs
+++ b/src/NuGetGallery/Configuration/ConfigurationService.cs
@@ -21,22 +21,19 @@ namespace NuGetGallery.Configuration
     {
         protected const string SettingPrefix = "Gallery.";
         protected const string FeaturePrefix = "Feature.";
+        protected const string ServiceBusPrefix = "AzureServiceBus.";
         private bool _notInCloud;
         private readonly Lazy<string> _httpSiteRootThunk;
         private readonly Lazy<string> _httpsSiteRootThunk;
-        private ISecretReaderFactory _secretReaderFactory;
-        private Lazy<ISecretInjector> _secretInjector;
-        private Lazy<IAppConfiguration> _lazyAppConfiguration;
-        private Lazy<FeatureConfiguration> _lazyFeatureConfiguration;
+        private readonly ISecretReaderFactory _secretReaderFactory;
+        private readonly Lazy<ISecretInjector> _secretInjector;
+        private readonly Lazy<IAppConfiguration> _lazyAppConfiguration;
+        private readonly Lazy<FeatureConfiguration> _lazyFeatureConfiguration;
+        private readonly Lazy<IServiceBusConfiguration> _lazyServiceBusConfiguration;
 
         public ConfigurationService(ISecretReaderFactory secretReaderFactory)
         {
-            if (secretReaderFactory == null)
-            {
-                throw new ArgumentNullException(nameof(secretReaderFactory));
-            }
-
-            _secretReaderFactory = secretReaderFactory;
+            _secretReaderFactory = secretReaderFactory ?? throw new ArgumentNullException(nameof(secretReaderFactory));
             _secretInjector = new Lazy<ISecretInjector>(InitSecretInjector, isThreadSafe: false);
 
             _httpSiteRootThunk = new Lazy<string>(GetHttpSiteRoot);
@@ -44,6 +41,7 @@ namespace NuGetGallery.Configuration
 
             _lazyAppConfiguration = new Lazy<IAppConfiguration>(() => ResolveSettings().Result);
             _lazyFeatureConfiguration = new Lazy<FeatureConfiguration>(() => ResolveFeatures().Result);
+            _lazyServiceBusConfiguration  = new Lazy<IServiceBusConfiguration>(() => ResolveServiceBus().Result);
         }
 
         public static IEnumerable<PropertyDescriptor> GetConfigProperties<T>(T instance)
@@ -64,6 +62,8 @@ namespace NuGetGallery.Configuration
         public IAppConfiguration Current => _lazyAppConfiguration.Value;
 
         public FeatureConfiguration Features => _lazyFeatureConfiguration.Value;
+
+        public IServiceBusConfiguration ServiceBus => _lazyServiceBusConfiguration.Value;
 
         /// <summary>
         /// Gets the site root using the specified protocol
@@ -165,6 +165,11 @@ namespace NuGetGallery.Configuration
         private async Task<IAppConfiguration> ResolveSettings()
         {
             return await ResolveConfigObject(new AppConfiguration(), SettingPrefix);
+        }
+
+        private async Task<IServiceBusConfiguration> ResolveServiceBus()
+        {
+            return await ResolveConfigObject(new ServiceBusConfiguration(), ServiceBusPrefix);
         }
 
         protected virtual string GetCloudSetting(string settingName)
--- a/src/NuGetGallery/Configuration/IServiceBusConfiguration.cs
+++ b/src/NuGetGallery/Configuration/IServiceBusConfiguration.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Configuration
+{
+    /// <summary>
+    /// Configuration values related to Service Bus integration.
+    /// </summary>
+    public interface IServiceBusConfiguration
+    {
+        /// <summary>
+        /// The connection string to use when connecting to the validation topic. This connection string should not
+        /// contain the topic name as the name is explicitly specified by <see cref="Validation_TopicName"/>. This
+        /// connection string only needs to have the "Send" privilege. This topic is used for requesting asynchronous
+        /// validation of packages.
+        /// </summary>
+        string Validation_ConnectionString { get; set; }
+
+        /// <summary>
+        /// The name of the Azure Service Bus topic to send validation messages to. This topic name is used at the same
+        /// time as the <see cref="Validation_ConnectionString"/>.
+        /// </summary>
+        string Validation_TopicName { get; set; }
+    }
+}
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -482,8 +482,7 @@ namespace NuGetGallery
                             id,
                             packageToPush,
                             packageStreamMetadata,
-                            user,
-                            commitChanges: false);
+                            user);
 
                         await AutoCuratePackage.ExecuteAsync(package, packageToPush, commitChanges: false);
 
--- a/src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
+++ b/src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
@@ -1,0 +1,62 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using NuGet.Services.Validation;
+using NuGetGallery.Configuration;
+using NuGetGallery.Diagnostics;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Initiates asynchronous validation on a package by enqueuing a message containing the package identity and a new
+    /// <see cref="Guid"/>. The <see cref="Guid"/> represents a unique validation request.
+    /// </summary>
+    public class AsynchronousPackageValidationInitiator : IPackageValidationInitiator
+    {
+        private readonly IPackageValidationEnqueuer _enqueuer;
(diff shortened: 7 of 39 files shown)


src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
```diff
@@ -0,0 +1,30 @@
...
+        private readonly IPackageValidationEnqueuer _enqueuer;
+
+        public AsynchronousPackageValidationInitiator(IPackageValidationEnqueuer enqueuer)
+        {
+            _enqueuer = enqueuer;
+        }
+
+        public async Task StartValidationAsync(Package package)
+        {
+            package.PackageStatusKey = PackageStatus.Available;
```
joelverhagen: For now, mark packages as `Available` immediately. Only during the final phase of rolling this feature out will we set the package status to `Validating` here.

src/NuGetGallery/packages.config
```diff
@@ -82,10 +82,13 @@
   <package id="NuGet.Packaging" version="4.3.0-preview1-2524" targetFramework="net46" />
   <package id="NuGet.Packaging.Core" version="4.3.0-preview1-2524" targetFramework="net46" />
   <package id="NuGet.Protocol" version="4.3.0-preview1-2524" targetFramework="net46" />
+  <package id="NuGet.Services.Abstractions" version="2.2.3-dev-13985" targetFramework="net46" />
   <package id="NuGet.Services.KeyVault" version="1.0.0.0" targetFramework="net46" />
-  <package id="NuGet.Services.Owin" version="2.2.3" targetFramework="net46" />
   <package id="NuGet.Services.Logging" version="2.2.3.0" targetFramework="net46" />
+  <package id="NuGet.Services.Owin" version="2.2.3" targetFramework="net46" />
   <package id="NuGet.Services.Platform.Client" version="3.0.29-r-master" targetFramework="net46" />
+  <package id="NuGet.Services.ServiceBus" version="2.2.3-dev-13985" targetFramework="net46" />
```
joelverhagen: Before merging, I will publish a stable version of these new NuGet.Service.* dependencies.

src/NuGetGallery/Configuration/AppConfiguration.cs
```diff
@@ -66,6 +66,14 @@ public class AppConfiguration : IAppConfiguration
         /// </summary>
         public bool AzureStorageReadAccessGeoRedundant { get; set; }
 
+        public bool AsynchronousPackageValidationEnabled { get; set; }
+
+        [DisplayName("AzureServiceBus.Validation.ConnectionString")]
+        public string AzureServiceBus_Validation_ConnectionString { get; set; }
```
skofman1: Can we introduce a class to contain all service bus config?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -314,6 +318,38 @@ protected override void Load(ContainerBuilder builder)
             ConfigureAutocomplete(builder, configuration);
         }
 
+        private void RegisterAsynchronousValidation(ContainerBuilder builder, ConfigurationService configuration)
+        {
+            builder
+                .RegisterType<ServiceBusMessageSerializer>()
```
skofman1: Can we avoid making the code aware that we are using service bus? Won't IValidationMessageSerializer do the job?

src/NuGetGallery.Core/PackageStatus.cs
```diff
@@ -15,5 +15,19 @@ public enum PackageStatus
         /// version are still reserved.
         /// </summary>
         Deleted = 1,
+
+        /// <summary>
+        /// The package is in the process of being validated. The validation is being done asynchronously. When the
+        /// validation is complete it will either move to the <see cref="Available"/> or the <see cref="Deleted"/>
+        /// state. When packages are in this state, they are not available but the package ID and version are still
```
shishirx34: Can't the state move to `FailedValidation` state as well in here?

src/NuGetGallery/Configuration/AppConfiguration.cs
joelverhagen: I would rather not. This is a case very similar to our Azure Storage configuration, which is also in `IAppConfiguration`. Until we have formal design for re-organization of configuration I prefer to align with what we already have for Azure Storage. Otherwise, we may end up having multiple approaches done here.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
joelverhagen: DI container code needs to be aware, unless we want to implement modules in the dependent libraries, which is worse since that would mean Autofac dependency goes everywhere.

Concerning the name, the serialization of messages has close affinity with the queue implementation. For example, Service Bus SDK uses data contract serialization for data and has a property bag for metadata. Other queue implement X may require entirely different serialization methods.

Note that taking a data model `PackageValidationMessageData` and tossing it on the queue is abstracted away at the `IPackageValidationEnqueuer` level. The implementation of `IPackageValidationInitiator` does not know anything about Service Bus.

There is a middle ground where we try to make a good guess on how to serialize messages by, say turning each data object into a `string` of data and `Dictionary<string, string>` for metadata, but this is making assumptions we have no reason to make. Even if we do this, we would still need an implementation to turn these two things (`string` and `Dictionary<string, string>`) to a Service Bus `BrokeredMessage`. This would still be visible here in the DI set up and would very likely have the same name.

src/NuGetGallery.Core/PackageStatus.cs
joelverhagen: Yes, nice catch.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -314,6 +318,38 @@ protected override void Load(ContainerBuilder builder)
...
+                .As<IServiceBusMessageSerializer>();
+
+            builder
+                .RegisterType<PackageValidationEnqueuer>()
```
chenriksson: Are the serializer and enqueuer part of the NuGet.Services.* dependencies? Assuming these are in separate PR / repo?

joelverhagen: They are in the ServerCommon repo:
https://github.com/NuGet/ServerCommon/tree/dev/src

src/NuGetGallery/Configuration/AppConfiguration.cs
joelverhagen: I'll see if I can come up with a good place to put these configuration values that can easily be adapted by other groupings.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
joelverhagen: If the shape of the interfaces seems incorrect, please let me know on this PR. I will apply the changes in all of the necessary places.

src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
chenriksson: Available is the default, so this shouldn't be necessary - right? Makes me wonder if we should have a different, disabled default that requires us to explicitly set this to Available.

joelverhagen: `Available` is default, yes. I think this should remain the default since it matches the DB default. I put this line here more-so to be defensive and to call out the location where we will soon have `... = PackageStatus.Validating`.

src/NuGetGallery/Configuration/AppConfiguration.cs
joelverhagen: I've added `IServiceBusConfiguration`.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -138,6 +111,8 @@ public void EnsureValid(PackageArchiveReader packageArchiveReader)
             packageRegistration.Packages.Add(package);
             await UpdateIsLatestAsync(packageRegistration, commitChanges: false);
 
+            await _validationInitiator.StartValidationAsync(package);
```
cristinamanum: Will it be possible for this await to throw an exception? If yes what is the status of the package on the failure? Will it be actionable for the user?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
scottbommarito: Can we move the registration of `ServiceBusMessageSerializer` and `PackageValidationEnqueuer` inside the if statement (`AsynchronousPackageValidationEnabled`)? Are either of these needed if `AsynchronousPackageValidationEnabled` is disabled?

src/NuGetGallery/Web.config
```diff
@@ -543,6 +547,30 @@
   </runtime>
   <system.serviceModel>
     <serviceHostingEnvironment aspNetCompatibilityEnabled="true" multipleSiteBindingsEnabled="true" />
+    <extensions>
+      <!-- In this extension section we are introducing all known service bus extensions. User can remove the ones they don't need. -->
```
scottbommarito: Can you explain exactly what these are for? Is this implying that we need to add these extensions to use service bus?

tests/NuGetGallery.Facts/Services/AsynchronousPackageValidationInitiatorFacts.cs
```diff
@@ -0,0 +1,96 @@
...
+                    {
+                        Id = "NuGet.Versioning"
+                    },
+                    Version = "4.3.0",
+                    PackageStatusKey = (PackageStatus)(-1),
+                };
+            }
+        }
+
+        public abstract class Base
```
scottbommarito: I was confused at first where these fields were coming from.

I suggest
1 - Changing `Base` to something more descriptive (e.g. `MethodFactsBase`).
2 - Moving it to the top of the file so it the first thing people read when try to read this file.

Same for `ImmediatePackageValidatorFacts.Base` as well.

joelverhagen: I think it's more confusing to put test infrastructure above the tests. Facts/asserts/etc are the most important part of a test class and should therefore be most visible.

src/NuGetGallery/Web.config
```diff
@@ -38,6 +38,10 @@
     
     <add key="Gallery.AzureStorageReadAccessGeoRedundant" value="false" />
     <!-- If the storage account has to fall-back to a secondary (when Read Access Geo Redundant is enabled in Azure storage), change Gallery.AzureStorageReadAccessGeoRedundant to true -->
+
+    <add key="Gallery.AsynchronousPackageValidationEnabled" value="false" />
+    <add key="AzureServiceBus.Validation.ConnectionString" value="" />
+    <add key="AzureServiceBus.Validation.TopicName" value="" />
```
chenriksson: Do we need the async flag, or can we assume that if the validation connection string is provided?

chenriksson: If ConnectionString is missing, I'm assuming package pushes will fail and we'll see exceptions in AI.

If TopicName is missing, would consumers just not receive the messages? Should we verify that this config is set if async is enabled?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
joelverhagen: I considered that but I thought it would be easier to put as much as possible into the DI container and let the run time figure out what is necessary. That's easier than trying to think through the minimal set of types required for activation. Only in cases where we choose between implementation A or B should we consider not adding something to the container.

src/NuGetGallery/Services/PackageService.cs
joelverhagen: If we fail to write to Service Bus an exception will be thrown. The package has not been written to DB or packages container yet so there is no concern of inconsistency.

src/NuGetGallery/Web.config
joelverhagen: These are added to the web.config by the Service Bus package. I would rather leave the defaults as-is. From what I can tell from online documentation, they basically enable features that come along with `Microsoft.ServiceBus`.

src/NuGetGallery/Web.config
joelverhagen: I'll leave some comments about this. Essentially, if `Gallery.AsynchronousPackageValidationEnabled` is `true` both the `ConnectionString` and `TopicName` must be set.

src/NuGetGallery/Web.config
scottbommarito: Does this comment come automatically with the service bus package?

Because I'd prefer to see something like
```
<!-- Service Bus extensions (added automatically by the Azure.ServiceBus package) -->
```

joelverhagen: > Does this comment come automatically with the service bus package?

Yes.

> Because I'd prefer to see something like

Sure.

src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
```diff
@@ -0,0 +1,30 @@
...
+
+            var data = new PackageValidationMessageData(
+                package.PackageRegistration.Id,
+                package.Version,
+                Guid.NewGuid());
+            await _enqueuer.StartValidationAsync(data);
```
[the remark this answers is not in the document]
joelverhagen: Yes, great idea!

src/NuGetGallery/Configuration/ConfigurationService.cs
```diff
@@ -21,29 +21,27 @@ public class ConfigurationService : PoliteCaptcha.IConfigurationSource, IGallery
...
         private readonly Lazy<string> _httpsSiteRootThunk;
-        private ISecretReaderFactory _secretReaderFactory;
-        private Lazy<ISecretInjector> _secretInjector;
-        private Lazy<IAppConfiguration> _lazyAppConfiguration;
-        private Lazy<FeatureConfiguration> _lazyFeatureConfiguration;
+        private readonly ISecretReaderFactory _secretReaderFactory;
+        private readonly Lazy<ISecretInjector> _secretInjector;
+        private readonly Lazy<IAppConfiguration> _lazyAppConfiguration;
+        private readonly Lazy<FeatureConfiguration> _lazyFeatureConfiguration;
+        private readonly Lazy<IServiceBusConfiguration> _lazyServiceBusConfiguration;
```
skofman1: >IServiceBusConfiguration [](start = 30, length = 24)

Like :)

src/NuGetGallery/Configuration/IServiceBusConfiguration.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Configuration
+{
+    public interface IServiceBusConfiguration
+    {
+        string Validation_ConnectionString { get; set; }
```
skofman1: Why do we need the validation prefix?

joelverhagen: When we have other Service Bus integrations, this will be useful to disambiguate: for example, when we disconnect V3 from V2 Feed.

src/NuGetGallery/Services/IPackageValidationInitiator.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    public interface IPackageValidationInitiator
```
agr: Documentation?..

joelverhagen: Done.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
skofman1: Spoke offline.

src/NuGetGallery/Configuration/IServiceBusConfiguration.cs
skofman1: If the plan is to have a separate interface and implementation pair for each service bus connection, it seems redundant. You can simply have a class called ValidationServiceBusConfiguration.
On the other hand, if we want all service bus config classes to have a shared interface, we can achieve that as well using DI.

joelverhagen: I think we should group all Service Bus configuration keys in the `IServiceBusConfiguration` interface. I don't see much value in a one or two property configuration interface.

joelverhagen:
I have made a change where we will not enqueue if the gallery is in read-only mode. This saves us from enqueuing a non-actionable Service Bus message on nuget.exe push against a read-only gallery, since the DB commit will fail.

src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
```diff
@@ -15,18 +16,26 @@ namespace NuGetGallery
...
             _enqueuer = enqueuer;
+            _appConfiguration = appConfiguration;
             _diagnosticsSource = diagnosticsService.SafeGetSource(nameof(AsynchronousPackageValidationInitiator));
         }
 
         public async Task StartValidationAsync(Package package)
         {
+            if (_appConfiguration.ReadOnlyMode)
+            {
+                throw new ReadOnlyModeException("Enqueuing unavailable: the gallery is currently in read only mode, with limited service. Please try again later.");
```
skofman1: Can we move this message to Strings?

joelverhagen: Sure, but note that this string is not user-visible. It will only appear in our exception logs. I think it makes sense for it to be in `Strings` since other enqueuers might use it.

joelverhagen: Fixed.

joelverhagen:
This merge is pending a stable version of `NuGet.Services.Validations`.

joelverhagen:
Everyone, I have made some significant changes since I first sent out this PR. 

- Removed `commitChanges` parameter from some places where it was always `false`. This was necessary so that `PackageUploadService` could handle the Service Bus enqueueing for validation instead of `PackageService` (PackageService.CreatePackageAsync(..., commitChanges: true) was preventing this move).
- Introduced `IValidationService` for all the validation-y goodness. This will have admin and revalidation logic.

tests/NuGetGallery.Facts/Services/ValidationServiceFacts.cs
```diff
@@ -0,0 +1,68 @@
...
+                _initiator
+                    .Setup(x => x.StartValidationAsync(It.IsAny<Package>()))
+                    .ReturnsAsync(packageStatus);
+
+                // Act
+                await _target.StartValidationAsync(_package);
+
+                // Assert
+                _packageService.Verify(
+                    x => x.UpdatePackageStatusAsync(_package, packageStatus, false),
```
loic-sharma: >false [](start = 77, length = 5)

Please add a named parameter for the `false` :)

joelverhagen: Doesn't work in `Expression`s

src/NuGetGallery/Services/IValidationService.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// The service for interacting with the asynchronous validation pipeline.
+    /// </summary>
+    public interface IValidationService
+    {
+        /// <summary>
+        /// Starts the asynchronous validation for the provided package and puts the package in the correct state. The
+        /// commit to the database is the responsibility of the caller.
+        /// </summary>
+        /// <param name="package">The package to start validation for.</param>
+        Task StartValidationForNewPackageAsync(Package package);
```
cristinamanum: Why the name StartValiadtionForNewPackage and not only StartValidationForPackage?

joelverhagen: I've changed the method name to `StartValidationAsync`.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -482,8 +482,7 @@ public virtual Task<ActionResult> CreatePackagePost()
                             id,
                             packageToPush,
                             packageStreamMetadata,
-                            user,
-                            commitChanges: false);
+                            user);
```
skofman1: Although commitChanges is always false, I think it adds to the readability of the code. Consider adding it back.

tests/NuGetGallery.Facts/Services/ValidationServiceFacts.cs
loic-sharma: Huh, I didn't know that. Good to know.

src/NuGetGallery/Controllers/ApiController.cs
joelverhagen: Dead code should not be maintained! We should get into the mindset of having limited locations where we commit changes.

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -228,10 +228,8 @@ public class ReservedNamespaceService : IReservedNamespaceService
         /// </summary>
```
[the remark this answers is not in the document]
joelverhagen: Done.

joelverhagen:
Merged with an unsigned version of ServerCommon packages. This matches existing behavior with `NuGet.Services.KeyVault`. This feature is currently disabled in DEV, INT, and PROD.

PR #4661, opened 2017-09-11
Fix and consolidate routing in NuGetGallery unit test configurations.

xavierdecoster:
This PR contains various fixes in NuGetGallery test infrastructure, which surface now as the gallery is checking for the `Gallery.SiteRoot` config setting. In other words: test routing and config setup must reflect a real possible scenario. 

In addition, `FormatException` were thrown due to invalid, dummy values for e-mail fields, as well as `NullReferenceException` due to service mocks that did not have all dependencies injected.

Tests are now more consistent:
* `Gallery.SiteRoot` = "http://localhost/"
* `IAppConfiguration.RequireSSL` = "true"
* all requests are HTTPS, assertions assume HTTPS (unless otherwise specified)
* we target fewer service mocks, more real implementations

PR #4680, opened 2017-09-14
Properly generate relative/absolute URLs.

xavierdecoster:
Follow-up PR for https://github.com/NuGet/NuGetGallery/pull/4661.

The change to rely on the `Gallery.SiteRoot` config setting had the side-effect of no longer returning any relative URLs. This broke some functional tests that relied on the assumption that these URLs were relative. As the `Gallery.SiteRoot` setting is the same for both staging and production slots in any environment, this became a hard to solve problem. Properly using relative URLs (and absolute ones where applicable, e.g. in outbound e-mails) fixes this and works around the test infrastructure complexity.

### DIFF
--- a/src/NuGetGallery/UrlExtensions.cs
+++ b/src/NuGetGallery/UrlExtensions.cs
@@ -52,6 +52,9 @@ namespace NuGetGallery
 
         private static string GetConfiguredSiteHostName()
         {
+            // It doesn't matter which value we pass on here for the useHttps parameter.
+            // We're just interested in the host, which is the same for both, 
+            // as it all results from the same 'NuGetGallery.SiteRoot' URL value.
             var siteRoot = GetSiteRoot(useHttps: true);
             return new Uri(siteRoot).Host;
         }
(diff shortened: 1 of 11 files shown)


src/NuGetGallery/UrlExtensions.cs
```diff
@@ -787,5 +733,47 @@ internal static string EnsureTrailingSlash(string url)
...
+
+        private static string GetRouteLink(
+            UrlHelper url,
+            string routeName,
+            bool relativeUrl,
+            RouteValueDictionary routeValues = null)
+        {
+            var protocol = GetProtocol(url);
+            var hostName = GetConfiguredSiteHostName();
+            var siteRoot = $"{protocol}://{hostName}";
```
loic-sharma: >siteRoot [](start = 16, length = 8)

This can be moved to within the `if (relativeUrl)` check.

loic-sharma: Same thing with line 747

---
In reply to: [139019197](https://github.com/NuGet/NuGetGallery/pull/4680#discussion_r139019197) [](ancestors = 139019197)

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -34,104 +34,93 @@ public static string Absolute(this UrlHelper url, string path)
...
-        private static string GetConfiguredSiteHostName(UrlHelper url)
+        internal static string GetSiteRoot(bool useHttps)
+        {
+            return _configuration.GetSiteRoot(useHttps);
+        }
+
+        private static string GetConfiguredSiteHostName()
         {
-            var siteRoot = _configuration.GetSiteRoot(useHttps: url.RequestContext.HttpContext.Request.IsSecureConnection);
+            var siteRoot = GetSiteRoot(useHttps: true);
```
loic-sharma: >true [](start = 49, length = 4)

`useHttps: true`? Shouldn't this do similar checking to `GetProtocol`? If this intentional, could you leave a comment explaining why?

PR #4698, opened 2017-09-18
Improve Third-Party Installation Tabs' UI

loic-sharma:
Made the height of first-party vs third-party tabs consistent.

First-party tab (there is 37 more pixels of space below the tab):

one

Third-party tab (no changes here):

two

### DIFF
--- a/src/Bootstrap/less/theme/page-display-package.less
+++ b/src/Bootstrap/less/theme/page-display-package.less
@@ -81,6 +81,10 @@
       text-decoration: underline;
     }
 
+    .tab-firstparty {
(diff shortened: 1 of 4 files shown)


src/Bootstrap/less/theme/page-display-package.less
```diff
@@ -78,45 +78,49 @@
       border: 0;
       color: #fff;
       font-weight: 600;
-      text-decoration: underline;
+      text-decoratinolo: underline;
```
loic-sharma: This is a mistake; will fix.

Issue #4623, opened 2017-09-06
404 redirect gives IIS 404 page


khellang:
A 404 redirect gives IIS 404 page:

![image](https://user-images.githubusercontent.com/582487/30132522-e72c0678-9350-11e7-8157-709ce2a855e6.png)

Try going to https://www.nuget.org/packages/Newtonsoft.Json/NotFound

PR #4699, opened 2017-09-18
Bug fix: custom error pages should be shown on 400, 404, 500

skofman1:
Additional change: cleanup old unused error pages

Fixes: https://github.com/NuGet/NuGetGallery/issues/4623


src/NuGetGallery/Web.config
```diff
@@ -294,9 +294,8 @@
       <!-- Remove Default HTTP Handler -->
       <remove path="*" verb="GET,HEAD,POST" />
     </httpHandlers>
-    <!-- Note: The Error.html file is in the ~/Public folder, but the URL Rewrite rules below take ~/Error.html and rewrite it appropriately. -->
-    <customErrors mode="RemoteOnly" defaultRedirect="~/Error.html">
-      <error statusCode="400" redirect="~/Errors/400" />
+    <customErrors mode="RemoteOnly" defaultRedirect="~/Errors/500" redirectMode="ResponseRedirect">
+      <error statusCode="400" redirect="~/Errors/400?error=1" />
       <error statusCode="404" redirect="~/Errors/404" />
```
joelverhagen: Should re pl

joelverhagen: These could theoretically lead of 404 and 500 as well depending on the logic of the action, right?

src/NuGetGallery/Web.config
```diff
@@ -294,9 +294,8 @@
       <!-- Remove Default HTTP Handler -->
       <remove path="*" verb="GET,HEAD,POST" />
     </httpHandlers>
-    <!-- Note: The Error.html file is in the ~/Public folder, but the URL Rewrite rules below take ~/Error.html and rewrite it appropriately. -->
-    <customErrors mode="RemoteOnly" defaultRedirect="~/Error.html">
-      <error statusCode="400" redirect="~/Errors/400" />
+    <customErrors mode="RemoteOnly" defaultRedirect="~/Errors/500" redirectMode="ResponseRedirect">
+      <error statusCode="400" redirect="~/Errors/400?error=1" />
```
joelverhagen: Spoke offline. Would `?` or `?error` work just as well?

loic-sharma: Could we have a comment explaining what this hack does? :)

skofman1: Apparently just ? works 🌞

src/NuGetGallery/Web.config
skofman1: I don't see a problem with keeping the query params for 404/500. There was a specific issue with 400 with <script></script> in the query string that causes continuous failures. This won't happen for 404/500.

Issue #4676, opened 2017-09-14
Package description breaks word in Firefox and Edge


BrainCrumbz:
Hello there

We just edited a package with the new gallery UI, in Firefox, and noticed that package description breaks word midway through.  
We tried with Edge: issue **is present** there. We tried with Chrome: **no issue**.

Some screenshots:

**Firefox**
![immagine](https://user-images.githubusercontent.com/3185573/30441103-b8738486-9978-11e7-9234-104dcd751199.png)

**Edge**
![immagine](https://user-images.githubusercontent.com/3185573/30441129-c9b76550-9978-11e7-8870-407137d4589c.png)

Inspecting element in Firefox, it shows it does not like how word-break CSS is set:

![immagine](https://user-images.githubusercontent.com/3185573/30441307-583be864-9979-11e7-884a-2d1800c9a80c.png)

while that's fine for Chrome:

![immagine](https://user-images.githubusercontent.com/3185573/30441360-7a628326-9979-11e7-9fe9-a2d5be87d8e5.png)

The package URL (of course we might change description in the future): https://www.nuget.org/packages/NSpec.VsAdapter/1.0.0.

* Chrome v60.0.3112.113 (64bit)
* Firefox v55.0.3 (32 bit)
* Edge v40.15063.0.0
* OS Windows 10 64bit

HTH

BrainCrumbz:
Side note: on Chrome words are not broken, but you might consider justifying those text blocks. Sorry if this has already been discussed elsewhere.

PR #4704, opened 2017-09-19
Fix improper word breaking in long text fields

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/4676.

`break-all` breaks in the middle of words. Instead use the different `break-word` forms (for maximum browser compatibility).

More details here:
https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/

Tested:
1. All browsers (Edge, IE, Firefox, Chrome)
1. Search results page and display page page
1. Long ID
1. Long Title
1. Long description
1. Long version
1. Small windows

### DIFF
--- a/src/Bootstrap/dist/css/bootstrap-theme.css
+++ b/src/Bootstrap/dist/css/bootstrap-theme.css
@@ -254,8 +254,11 @@ img.package-icon {
   margin-left: 20px;
 }
 .break-word {
-  word-break: break-all;
+  word-break: normal;
   word-break: break-word;
+  word-wrap: break-word;
+
+  overflow-wrap: break-word;
 }
 #edit-metadata-form-container .loading:after {
   display: inline-block;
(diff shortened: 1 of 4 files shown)


agr:
`word-break: break-word;` is non-standard and should not be left by itself without standard-compliant fallback. Should we drop it, too?

joelverhagen:
Browsers not supporting `word-break: break-word` fall back to the default with is `normal`. At least that's behavior I saw in IE and Edge.

agr:
Browsers should fallback to whatever was in effect before the `word-break: break-word;` (not to `normal` specifically), i.e. they should just ignore lines they don't understand (if I am not mistaken). If we want to be sure that it falls back to `normal`, we should specify it explicitly and not rely on it being set properly on parent element.

joelverhagen:
Ah, good point. We don't use `word-break` anywhere else in the document so it's the same net effect but we should be defensive. Will fix.

joelverhagen:
> Browsers should fallback to whatever was in effect before the `word-break: break-word;` (not to `normal` specifically), i.e. they should just ignore lines they don't understand (if I am not mistaken). If we want to be sure that it falls back to `normal`, we should specify it explicitly and not rely on it being set properly on parent element.

Fixed.

Issue #4688, opened 2017-09-15
Reserved package ID prefix:  grammatical error in tooltip


dtivel:
Pluralize "owner" in tooltip.  It should read:  "...one of the owners..."

![image](https://user-images.githubusercontent.com/12734758/30485365-03751530-99e2-11e7-9990-e2951d882202.png)

shishirx34:
fixing this with next merge.

PR #4708, opened 2017-09-20
[Prefix] Admin page for namespace reservation

shishirx34:
Also addresses #4688 

Admin page:

image


skofman1:
Will this UI work for reserving exact match prefixes. For example: reserve JQuery , but not JQuery*

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -74,19 +74,16 @@ public class ReservedNamespaceService : IReservedNamespaceService
...
-                // package registration.
-                if (!namespaceToDelete.IsSharedNamespace)
-                {
-                    var packageRegistrationsToMarkUnverified = namespaceToDelete
-                        .PackageRegistrations
-                        .Where(pr => pr.ReservedNamespaces.Count() == 1)
-                        .ToList();
+                // Delete verified flags on corresponding packages for this prefix if 
+                // it is the only prefix matching the package registration.
+                var packageRegistrationsToMarkUnverified = namespaceToDelete
```
skofman1: please add a UT covering this case

skofman1: please add a UT covering this case

src/NuGetGallery/Areas/Admin/Views/ReservedNamespace/Index.cshtml
```diff
@@ -0,0 +1,237 @@
...
+
+@ViewHelpers.AjaxAntiForgeryToken(Html)
+
+<section>
+    <article id="stage">
+        <div class="message" style="display: none" data-bind="text: message, visible: message"></div>
+        <h2>Reserve Namespace</h2>
+        <form>
+            <div class="form-horizontal">
+                <input type="text" placeholder="Search for a prefix" autocomplete="on" autofocus data-bind="value: prefixSearchQuery" />
```
skofman1: consider adding a tooltip/help text with the allowed query format

src/NuGetGallery/Areas/Admin/ViewModels/ReservedNamespaceResultModel.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public ReservedNamespace prefix;
+
+        public bool isExisting;
+
+        public string[] registrations;
+
+        public string[] owners;
+
+        public ReservedNamespaceResultModel(): this(null, false) { }
```
skofman1: you can leave the default ctor empty. The result will be the same as this(null, false)
Also, this will allow you to simplify the parametrized ctor by removing the if and validating the input (
if (reservedNamespace == null) throw

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                Prefixes = resultModel
+            };
+
+            return Json(results, JsonRequestBehavior.AllowGet);
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<JsonResult> AddPrefix(ReservedNamespace prefix)
```
skofman1: >prefix [](start = 66, length = 6)

"prefix" should actually be "namespace"?
In this case, we should rename the param and method name

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                return Json(new { success = false, message = ex.Message });
+            }
+        }
+
+        private static string[] GetPrefixesFromQuery(string query)
+        {
+            return query.Split(',', '\r', '\n',';')
+                .Select(prefix => prefix.Trim())
+                .Where(prefix => !string.IsNullOrEmpty(prefix)).ToArray();
```
chenriksson: Instead of where, can use StringSplitOptions.RemoveEmptyEntries

src/NuGetGallery/Areas/Admin/ViewModels/ReservedNamespaceResultModel.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+
+namespace NuGetGallery.Areas.Admin.ViewModels
+{
+
+    public class ReservedNamespaceResultModel
```
chenriksson: nit: we could probably mark our model classes as sealed

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
```diff
@@ -0,0 +1,118 @@
...
+
+using NuGetGallery.Areas.Admin.ViewModels;
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Web.Mvc;
+
+namespace NuGetGallery.Areas.Admin.Controllers
+{
+    public class ReservedNamespaceController : AdminControllerBase
```
chenriksson: Should we have tests for the admin controller?

src/NuGetGallery/Areas/Admin/Views/ReservedNamespace/Index.cshtml
```diff
@@ -0,0 +1,237 @@
...
+            }
+        </div>
+    </article>
+
+</section>
+
+@section BottomScripts {
+    <script src="@Url.Content("~/Scripts/knockout-2.2.1.js")"></script>
+    <script src="@Url.Content("~/Scripts/gallery/common.js")"></script>
+    <script>
```
chenriksson: nit: could use Scripts.Render(...), and should probably put type="text/javascript" (even though we don't consistently do it)

chenriksson: Other admin pages use ko 2.2.1, but should we be using 3.4.2 like other site pages? Should we just include bundle "site.min.js"?

src/NuGetGallery/Strings.resx
```diff
@@ -585,4 +585,19 @@ For more information, please contact '{2}'.</value>
   <data name="ReadMeUrlHostInvalid" xml:space="preserve">
     <value>ReadMe URL must be a raw markdown file hosted on GitHub.</value>
   </data>
+  <data name="ReservedNamespace_OwnerAdded" xml:space="preserve">
+    <value>The user '{0}' is now an owner of the prefix '{1}'.</value>
+  </data>
+  <data name="ReservedNamespace_OwnerRemoved" xml:space="preserve">
+    <value>The user '{0}' removed from owners list of prefix '{1}'.</value>
```
loic-sharma: >removed [](start = 26, length = 7)

Should this be "The user '{0}' *was* removed from..."?

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        public IReservedNamespaceService ReservedNamespaceService { get; private set; }
```
loic-sharma: >IReservedNamespaceService [](start = 15, length = 25)

Does this need to be public?

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
loic-sharma: >ToArray [](start = 64, length = 7)

Maybe do `ToList` instead of `ToArray`? The `GetPrefixesFromQuery` helper is only used once and its result is always converted to a `List`.

src/NuGetGallery/Areas/Admin/ViewModels/ReservedNamespaceResultModel.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public ReservedNamespace prefix;
```
loic-sharma: >prefix [](start = 33, length = 6)

Consider changing all public properties to public properties with only getters.

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
```diff
@@ -94,5 +94,16 @@
                 Manage User Security Policies
             </p>
         </li>
+        <li>
+            <h2>
+                <a class="action-menu-link" href="@Url.Action(actionName: "Index", controllerName: "ReservedNamespace")">
```
loic-sharma: Does these links not need Xavier's URL fixes? If they do, you should create a helper in `UrlExtensions` for all of these `Url.Action`s.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -958,9 +958,10 @@ private void NotifyIndexingService()
 
         public virtual async Task UpdatePackageVerifiedStatusAsync(IReadOnlyCollection<PackageRegistration> packageRegistrationList, bool isVerified)
         {
+            var list = packageRegistrationList.Select(prl => prl.Id);
             var allPackageRegistrations = _packageRegistrationRepository.GetAll();
             var packageRegistrationsToUpdate = allPackageRegistrations
-                .Where(pr => packageRegistrationList.Any(prl => prl.Id == pr.Id))
+                .Where(pr => list.Contains(pr.Id))
```
loic-sharma: >Contains [](start = 34, length = 8)

`list` is a LINQ `IEnumerable` - each call to `Contains` will cause `list` to be reevaluated. You should turn the list into an `Array` or a `HashSet` first, and then call `Contains`.

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
```diff
@@ -0,0 +1,118 @@
...
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<JsonResult> AddPrefix(ReservedNamespace prefix)
+        {
+            try
+            {
+                await ReservedNamespaceService.AddReservedNamespaceAsync(prefix);
+                return Json(new { success = true, message = string.Format(Strings.ReservedNamespace_PrefixAdded, prefix.Value) });
+            }
+            catch (Exception ex)
```
joelverhagen: Shouldn't catch `Exception` -- catch more specific

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
loic-sharma: If you change this, also update line 103 of ReservedNamespace's Index.cshtml too

---
In reply to: [140036318](https://github.com/NuGet/NuGetGallery/pull/4708#discussion_r140036318) [](ancestors = 140036318)

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                return Json(new { success = false, message = ex.Message });
```
loic-sharma: Should we log all these exceptions?

shishirx34: These exceptions are not actionable for us as such. These are user action exceptions which provide the way for user(admin) to act on it. Also, I have another task for logging and auditing for reserved namespace, will be taking care of all logging in that task.

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
shishirx34: `RemoveEmptyEntries` doesn't remove the whitespace strings, since the trim happens after split. I will leave it as is.

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
shishirx34: I don't think so, I kept the actions consistent with other admin pages, which haven't been modified.

src/NuGetGallery/Areas/Admin/Views/ReservedNamespace/Index.cshtml
shishirx34: I would rather not make unrelated changes in here. keeping the admin pages consistent, so that when/if we change we could make the changes in batch.

src/NuGetGallery/Services/PackageService.cs
shishirx34: good catch. will use HashSet

src/NuGetGallery/Services/ReservedNamespaceService.cs
shishirx34: added

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
loic-sharma: @xavierdecoster can you confirm this is intended? :)

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
```diff
@@ -0,0 +1,118 @@
...
+        [HttpGet]
+        public ActionResult Index()
+        {
+            return View();
+        }
+
+        [HttpGet]
+        public virtual JsonResult SearchPrefix(string query)
+        {
+            var prefixQueries = GetPrefixesFromQuery(query);
```
cristinamanum: Is any check for null or empty is necessary?

joelverhagen:
Spoke offline about this: if a reserved namespace is shared, we are saying that all namespaces under that must also be considered shared. If that is the case, the "Add" flow should force all namespaces under a shared namespace to be shared so that the data matches today's implementation.

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
```diff
@@ -0,0 +1,118 @@
...
+            catch (Exception ex)
+            {
+                return Json(new { success = false, message = ex.Message });
+            }
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<JsonResult> RemovePrefix(ReservedNamespace prefix)
+        {
```
cristinamanum: Do we need to audit the delete action?

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                return Json(new { success = false, message = ex.Message });
+            }
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<JsonResult> RemoveOwner(ReservedNamespace prefix, string owner)
+        {
+            try
```
cristinamanum: Same as above. Do we want to audit this action?

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
shishirx34: Yes, auditing will be part of a separate PR since it might need changes for shims.

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
shishirx34: yes, part of a separate task.

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
shishirx34: yes, added one.

src/NuGetGallery/Areas/Admin/Controllers/ReservedNamespaceController.cs
shishirx34: added

shishirx34:
@joelverhagen - addressed the edge case.

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
xavierdecoster: If it works on DEV (which has APIM configured), then we're good :)

PR #4711, opened 2017-09-20
Hide readme form on validation errors

chenriksson:
-Update bindData to call bindReadMeData when model is null
-Update bindReadMeData to hide form when model is null
-Remove unmatched ending form tag

### DIFF
--- a/src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
+++ b/src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
@@ -175,7 +175,7 @@
                     </div>
                 </div>
                 <!-- /ko -->
-            </form>
+            @*</form>*@
         </div>
 
         <!-- ko if: $data.IsSemVer2 -->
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Views/Packages/_EditMetadata.cshtml
```diff
@@ -175,7 +175,7 @@
                     </div>
                 </div>
                 <!-- /ko -->
-            </form>
+            @*</form>*@
```
joelverhagen: What is this about?

chenriksson: When edit support was added, the form was extracted into an _EditForm partial but only the begin tag was removed.

see:
https://github.com/NuGet/NuGetGallery/commit/0434052dba95a1526df5f961bc09ebb9fc5b24f9#diff-5058b8e804fd7b6e9e60a29de7f6fff6L46

ryuyu: So should this be removed altogether?

chenriksson: Yes - see link above. It should have been removed when the form tags were extracted into _EditForm.

Issue #4706, opened 2017-09-19
Package upload doesn't validate edits


chenriksson:
Repro:
1. Go to https://int.nugettest.org
2. Upload package
3. Delete authors or other required metadata
4. Click Submit

Package will upload and create package edit without showing any validation error

Note, a related issue is that the upload/edit form fields disappear on model validation errors... repro below. Choosing not to fix since form is so long that user won't see the errors; and because upload uses a general error handler for submit/cancel/failure.

Repro 1:
1. Go to https://int.nugettest.org
2. Edit existing package
3. Delete authors or other required metadata
4. Click Submit

Validation error displays, but metadata form fields disappear.

PR #4712, opened 2017-09-20
Add model validation for package upload

chenriksson:
Adding model validation to VerifyPackage, same as done for Edit

Issue is #4706. Note that Edit/Upload currently would require refresh on validation errors because handleErrors does a bindData(null). Isn't ideal, but leaving as-is because:
- Upload uses single error handler for cancel, submit, upload failure
- Upload/Edit form is long, so it's not apparent that validation failed unless user scrolls up


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1293,6 +1293,12 @@ private string GetSecurePushPolicyDescriptions()
         {
             var currentUser = GetCurrentUser();
 
+            if (!ModelState.IsValid)
```
loic-sharma: Why do you check the `ModelState` after fetching the `currentUser`?

PR #4719, opened 2017-09-21
[Prefix] Fix the bug with absolute namespace reservation for an owner

shishirx34:
There was a bug which would mark all package registrations under an owner which started with the reserved prefix even if the prefix is an absolute namespace.

This fixes it, plus added test.

### DIFF
--- a/src/NuGetGallery/Services/ReservedNamespaceService.cs
+++ b/src/NuGetGallery/Services/ReservedNamespaceService.cs
@@ -142,6 +142,9 @@ namespace NuGetGallery
 
                 // Mark all packages owned by this user that start with the given namespace as verified.
                 var allPackageRegistrationsForUser = PackageService.FindPackageRegistrationsByOwner(userToAdd);
+
+                // We need 'AsQueryable' here because FindPackageRegistrationsByOwner returns an IEnumerable
+                // and to evaluate the predicate server side, the casting is essential.
                 var packageRegistrationsMatchingNamespace = allPackageRegistrationsForUser
                     .AsQueryable()
                     .Where(predicate)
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -130,10 +130,21 @@ public class ReservedNamespaceService : IReservedNamespaceService
                     throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, Strings.ReservedNamespace_UserAlreadyOwner, username));
                 }
 
+                Expression<Func<PackageRegistration, bool>> predicate;
+                if (namespaceToModify.IsPrefix)
```
cristinamanum: an we use ? and  :  ?

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -130,10 +130,21 @@ public class ReservedNamespaceService : IReservedNamespaceService
...
+                else
+                {
+                    predicate = registration => registration.Id.Equals(namespaceToModify.Value);
+                }
+
                 // Mark all packages owned by this user that start with the given namespace as verified.
                 var allPackageRegistrationsForUser = PackageService.FindPackageRegistrationsByOwner(userToAdd);
                 var packageRegistrationsMatchingNamespace = allPackageRegistrationsForUser
-                    .Where(pr => pr.Id.StartsWith(namespaceToModify.Value, StringComparison.OrdinalIgnoreCase))
+                    .AsQueryable()
```
joelverhagen: Spoke offline -- we should leave a comment that the `predicate` is still being evaluated in SQL Server and that `FindPackageRegistrationsByOwner` should return an `IQueryable`.

src/NuGetGallery/Services/ReservedNamespaceService.cs
shishirx34: The ternary operators cannot evaluate lambda expressions without explicit casting. This is much readable.

PR #4721, opened 2017-09-21
Move logic needed by validation orchestrator out of PackageService into CorePackageService 

joelverhagen:
Also, add the `UpdatePackageStatusAsync` method for package status transitions.

I'm open to naming suggestions, but I thought this makes sense since it's in the `NuGetGallery.Core` assembly.


src/NuGetGallery.Core/Services/CorePackageService.cs
```diff
@@ -0,0 +1,199 @@
...
+    public class CorePackageService : ICorePackageService
+    {
+        protected readonly IEntityRepository<Package> _packageRepository;
+        
+        public CorePackageService(IEntityRepository<Package> packageRepository)
+        {
+            _packageRepository = packageRepository;
+        }
+
+        public virtual async Task UpdatePackageStatusAsync(Package package, PackageStatus packageStatus, bool commitChanges = true)
```
skofman1: newPackageStatus?

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -60,6 +60,172 @@ public class PackageServiceFacts
             return packageService.Object;
         }
 
+        public class TheUpdatePackageStatusMethod
```
chenriksson: nit: Should core tests be split out into a different class in Gallery.Core.Facts?

src/NuGetGallery.Core/Services/CorePackageService.cs
```diff
@@ -0,0 +1,199 @@
...
+            if (package.PackageStatusKey != packageStatus)
+            {
+                package.Deleted = packageStatus == PackageStatus.Deleted;
+                package.PackageStatusKey = packageStatus;
+
+                /// If the package is being made available, consider this change as an "edit" so that the package
+                /// appears for cursors on the <see cref="Package.LastEdited"/> field.
+                if (packageStatus == PackageStatus.Available)
+                {
+                    package.LastEdited = DateTime.UtcNow;
```
chenriksson: UpdateIsLatest sets LastModified.... does that not make the package appear for cursors as well?

src/NuGetGallery.Core/Services/CorePackageService.cs
```diff
@@ -0,0 +1,199 @@
...
+                    package.IsLatestSemVer2 ||
+                    package.IsLatestStableSemVer2)
+                {
+                    await UpdateIsLatestAsync(package.PackageRegistration, commitChanges: false);
+                }
+            }
+
+            if (commitChanges)
+            {
+                // Even if the package status did not change, commit the changes in the entity context to pick ip any
```
skofman1: up

chenriksson: nit: pick *up*

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
joelverhagen: I considered this but saw no value since the resulting test coverage is the same. Since you pointed it out I will fix it since I guess it's more "correct".

src/NuGetGallery.Core/Services/CorePackageService.cs
joelverhagen: I presume you mean `LastUpdated` instead of `LastModified`? 

`LastUpdated` is for any change to the package record.
`LastEdited` is for changes to the package that need V3 to re-evaluate the package.

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
chenriksson: It's more correct... but probably only worth fixing if there's a chance that Core tests would be run independently of Gallery. Not a huge deal.

src/NuGetGallery.Core/Services/CorePackageService.cs
chenriksson: Got it - thanks!

src/NuGetGallery.Core/Services/CorePackageService.cs
joelverhagen: Fixed.

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
joelverhagen: Fixed.

src/NuGetGallery.Core/Services/CorePackageService.cs
joelverhagen: Fixed.

src/NuGetGallery.Core/Services/CorePackageService.cs
joelverhagen: Fixed.

PR #4722, opened 2017-09-21
Add ownership requests to manage packages page

scottbommarito:
Currently, there is no way for the pending owner of a package ownership request to view and accept the request outside of the email that is sent. As a result, if the email fails to send, the pending owner will never know that they were ever requested and--even if they do know they were requested--cannot accept the request!

This PR adds management of package ownership requests to the manage packages page. It shows both incoming and outgoing requests. Users can accept and reject requests for them to become owners of a package and can cancel requests they had made.

image

This is a continuation of https://github.com/NuGet/NuGetGallery/pull/4512, which had to be closed and rebased due to strange issues with GitHub.


src/NuGetGallery/Migrations/201709202249402_AddPackageOwnershipRequestsPage.cs
```diff
@@ -0,0 +1,20 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddPackageOwnershipRequestsPage : DbMigration
+    {
+        public override void Up()
+        {
+            CreateIndex("dbo.PackageOwnerRequests", "PackageRegistrationKey");
```
scottbommarito: This appears to be required because I added `public virtual PackageRegistration PackageRegistration { get; set; }` to `PackageOwnerRequest`, which did not have a way to get the `PackageRegistration` object before.

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -612,31 +612,87 @@ public static string RemovePackageOwner(this UrlHelper url, bool relativeUrl = t
...
 
-        public static string ConfirmationUrl(
+        public static string ConfirmPendingOwnershipRequest(
+            this UrlHelper url,
+            string packageId,
+            string username,
+            string confirmationCode,
+            bool relativeUrl = true)
+        {
+            var rvd = new RouteValueDictionary
```
loic-sharma: >rvd [](start = 16, length = 3)

I'm not a huge fan of acronyms, could we use `routeValues` to be consistent with other snippets in this file?

scottbommarito: I used it because I saw `rvd` elsewhere in the file, but it turns out the place I saw it is the only other place it is...lol

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1137,21 +1157,72 @@ public virtual ActionResult Delete(string id, string version)
...
+                return View("ConfirmOwner", new PackageOwnerConfirmationModel(id, username, ConfirmOwnershipResult.AlreadyOwner));
+            }
+
+            var request = _packageOwnerRequestService.GetPackageOwnershipRequest(package, user, token);
+            if (request == null)
+            {
+                return View("ConfirmOwner", new PackageOwnerConfirmationModel(id, username, ConfirmOwnershipResult.Failure));
+            }
+
+            var requestingUser = request.RequestingOwner;
```
loic-sharma: >requestingUser [](start = 16, length = 14)

This is only used in the `else` on line 1183, consider moving the variable there?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -154,10 +156,16 @@ public virtual ActionResult Packages()
             var user = GetCurrentUser();
             var packages = _packageService.FindPackagesByOwner(user, includeUnlisted: true)
                 .Select(p => new ListPackageItemViewModel(p)).OrderBy(p => p.Id).ToList();
+
```
loic-sharma: 159 has whitespace here.

src/NuGetGallery/Views/Users/_OwnerRequestsList.cshtml
```diff
@@ -0,0 +1,78 @@
...
+        </h2>
+        <div class="panel-collapse collapse in" id="requests-@Model.Name" aria-expanded="true">
+            <p>
+                You have <b>@Model.RequestItems.Count().ToNuGetNumberString() @Model.Name.ToLowerInvariant() @requestsString</b>
+            </p>
+            <div class="list-packages" role="list">
+                <table class="table">
+                    <thead>
+                        <tr class="manage-package-headings">
+                            <th class="col-sm-1"></th>
```
joelverhagen: Why specify column sizes here? Browser can manage column sizes on its own -- which is preferable for overflow cases.

src/NuGetGallery/Views/Users/_OwnerRequestsList.cshtml
```diff
@@ -0,0 +1,78 @@
...
+                <i class="ms-Icon ms-Icon--ChevronDown" aria-hidden="true"></i>
+                <span>Pending Ownership Requests (@Model.Name)</span>
+            </a>
+        </h2>
+        <div class="panel-collapse collapse in" id="requests-@Model.Name" aria-expanded="true">
+            <p>
+                You have <b>@Model.RequestItems.Count().ToNuGetNumberString() @Model.Name.ToLowerInvariant() @requestsString</b>
+            </p>
+            <div class="list-packages" role="list">
+                <table class="table">
```
joelverhagen: This should be styled a lot like `_UserPackagesList.cshtml` right?

src/NuGetGallery/Views/Users/_OwnerRequestsList.cshtml
scottbommarito: I thought I got rid of them, my bad.

joelverhagen:
Spoke offline about this.

If we are doing work in this area to move the package ownership request experience to the web UI, we should at the same time:

1. Remove the token from the URL
1. Change the accept/reject endpoint to POST
1. Make the email link to either link to the manage packages page
   - Alternatively, link to a GET version of the accept/reject endpoints which require an additional click.

src/NuGetGallery/Views/Users/_OwnerRequestsList.cshtml
scottbommarito: yup

scottbommarito:
@joelverhagen I agree that doing those things is the ideal fix, but since we'd like to limit time working on this and this PR is in a good state anyway, I will open up an issue with your feedback for the backlog.

PR #4740, opened 2017-09-25
Add support for X-NuGet-Protocol-Version header

skofman1:
Functional tests will be sent in a different PR


src/NuGetGallery/Security/RequireMinClientVersionForPushPolicy.cs
```diff
@@ -76,8 +86,16 @@ public override SecurityPolicyResult Evaluate(UserSecurityPolicyEvaluationContex
...
 
-            var clientVersion = GetClientVersion(context);
-            if (clientVersion == null || clientVersion < minClientVersion)
+            // Do we have X-NuGet-Protocol-Version header?
+            var protocolVersion = GetProtocolVersion(context);
+
+            if (protocolVersion == null)
+            {
+                // Do we have X-NuGet-Client-Version header?
+                protocolVersion = GetClientVersion(context);
```
joelverhagen: Let's ignore (or block -- which is the same effect if we are requiring this header) client version ≥ first client version that uses `X-NuGet-Protocol-Version`. Given that we don't know this version yet, let's set it to a sufficiently high version and have a work item to lower it once client has shipped with the fix. That way the code is in place and if we forget/de-prioritize the lowering it will still work in the future.

skofman1: We will need to actively deprecate the header only if we see clients continue to use it in the long run. Since this is currently not a problem, and deprecation is expected to happen organically, created this issue to track: https://github.com/NuGet/NuGetGallery/issues/4742

PR #4747, opened 2017-09-26
Implement "revalidate" gesture for asynchronous validation

joelverhagen:
Depends on https://github.com/NuGet/NuGetGallery/pull/4677. Address https://github.com/NuGet/Engineering/issues/765.

### DIFF
--- a/src/NuGetGallery/UrlExtensions.cs
+++ b/src/NuGetGallery/UrlExtensions.cs
@@ -486,7 +486,24 @@ namespace NuGetGallery
         {
             return GetActionLink(
                 url,
-                "Reflow",
+                nameof(PackagesController.Reflow),
+                "Packages",
+                relativeUrl,
+                routeValues: new RouteValueDictionary
+                {
+                    { "id", package.Id },
+                    { "version", package.Version }
+                });
+        }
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/UrlExtensions.cs
```diff
@@ -479,16 +479,27 @@ public static string UploadPackageProgress(this UrlHelper url, bool relativeUrl
                 });
         }
 
-        public static string ReflowPackage(
-            this UrlHelper url,
-            IPackageVersionModel package,
-            bool relativeUrl = true)
```
loic-sharma: >relativeUrl [](start = 17, length = 11)

Out of curiosity, why do you remove the `relativeUrl` parameter?

joelverhagen: It is unused.

loic-sharma: Personally I would rather remain consistent with the other public method signatures than to remove an unused parameter.

joelverhagen: Okay, I'll add it back.

PR #4759, opened 2017-09-27
Split some of PackageFileService and FileStorageService into NuGetGallery.Core

joelverhagen:
Orchestrator needs to upload to the packages container when a package becomes available. Therefore, I had to split `IPackageFileService` into two, introducing a new `ICorePackageFileService`. Ideally I would have moved the whole service down to `NuGetGallery.Core` but there were some web service specific methods on `IPackageFileService` that don't make sense for orchestrator (e.g. `CreateDownloadPackageActionResultAsync` which requires an MVC dependency).

`IFileStorageService` has some constants which relate to all sorts of container names and content types not specific to the `packages` container. Instead of splitting these methods in half, I chose to just push the entire methods and their associated constants down to `NuGetGallery.Core` instead of having to do `override` in `NuGetGallery`. I think this is the right move since it is simpler and more likely than not we'll have to push more into `NuGetGallery.Core` as time goes on.

By extension, `IFileStorageService` also had to be split up because it also has an MVC-dependent method, thus introducing `ICoreFileStorageService`.

Finally, I split the test classes up so that implementations now in `NuGetGallery.Core` are tested by `NuGetGallery.Core.Facts`.

In general I like this pattern since we have less and less logic in our web app and more and more logic in a project which is agnostic of entry point and is abstracted away from view concerns (like .cshtml or HTTP redirects).


joelverhagen:
Note that this PR introduces no new behavior. It is simply a refactoring.

src/NuGetGallery/Services/IPackageFileService.cs
```diff
@@ -25,21 +26,11 @@ public interface IPackageFileService
         Task DeletePackageFileAsync(string id, string version);
```
loic-sharma: >DeletePackageFileAsync [](start = 13, length = 22)

It seems that more of these methods could be moved down to `ICoreFileStorageService`, like this `DeletePacakgeFileAsync`. Are you restricting the move to just methods needed by the Orchestrator?

joelverhagen: Correct. `IPackageFileService` does a lot more than what orchestrator foreseeably needs. You'll notice that I moved a lot more down in the case of `IFileStorageService` since those methods are more generic and could reasonably be used by other orchestrator needs (i.e. auditing).

src/NuGetGallery/Strings.Designer.cs
```diff
@@ -1035,7 +1026,7 @@ public class Strings {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to The ID prefix of this package has been reserved for one of the owners of this package by NuGet.org..
+        ///   Looks up a localized string similar to The ID prefix of this package has been reserved for one of the owner of this package by NuGet.org..
```
skofman1: can u undo this?

shishirx34: Yes, please, don't override my bug fix. 😌

joelverhagen: Fixed. Somehow the XML doc and the actual resource were out of sync.

PR #4763, opened 2017-09-27
Log user safe exceptions

chenriksson:
Fix for: https://github.com/NuGet/Engineering/issues/809

### DIFF
--- a/src/NuGetGallery/Services/TelemetryService.cs
+++ b/src/NuGetGallery/Services/TelemetryService.cs
@@ -5,11 +5,14 @@ using System;
 using System.Collections.Generic;
 using System.Security.Principal;
 using System.Web;
+using NuGetGallery.Diagnostics;
 
 namespace NuGetGallery
 {
     public class TelemetryService : ITelemetryService
     {
+        private IDiagnosticsSource _trace;
+
         // Event types
         public const string ODataQueryFilterEvent = "ODataQueryFilter";
         public const string PackagePushEvent = "PackagePush";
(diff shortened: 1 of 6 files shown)


skofman1:
Are those logged as traces?

skofman1:
In order to be able to monitor AI exceptions, we need AI exceptions to contain only exceptions that indicate a service issue. In this specific case, the exception is caused by a user mistake, and should be logged as a trace.

chenriksson:
@skofman1 Will update PR to log as traces. Also need to investigate why user isn't getting the descriptive message (i.e., Id conflicts with Title of another package).

chenriksson:
@skofman1 @joelverhagen Updated PR to send these exceptions to trace instead

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -36,6 +39,31 @@ public class TelemetryService : ITelemetryService
...
+        {
+            if (diagnosticsService == null)
+            {
+                throw new ArgumentNullException(nameof(diagnosticsService));
+            }
+
+            _trace = diagnosticsService.GetSource("TelemetryService");
+        }
+
+        // Used by ODataQueryVerifier. Should consider refactoring to make this non-static.
```
xavierdecoster: Should we log an issue to track this as an actionable clean-up task?

chenriksson: Good idea... opened #4795

PR #4764, opened 2017-09-28
Move package commit logic to IPackageUploadService

joelverhagen:
Right now, `ApiController` and `PackagesController` both have the logic where we reliably write to `packages` container and to the database whilst still handling failures. This logic is duplicated.

I have moved this logic to `IPackageUploadService.CommitPackageAsync`. Right now, this is just a refactoring but the goal in a subsequent PR is to write (and delete in failure cases) packages to the `validating` container if the package has `PackageStatus.Validating`. Otherwise, existing behavior is maintained.

Progress on https://github.com/NuGet/NuGetGallery/issues/4646.

### DIFF
--- a/tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
+++ b/tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
@@ -7,6 +7,7 @@ using System.Collections.Specialized;
 using System.Data;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
 using System.Web;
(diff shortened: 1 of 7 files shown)


tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
```diff
@@ -23,6 +23,7 @@
 using NuGetGallery.Security;
 using Xunit;
 using NuGetGallery.Configuration;
+using System.Linq;
```
chenriksson: nit: sort

joelverhagen: Fixed.

PR #4766, opened 2017-09-28
Write to the validation container if the package starts in the Validating status

joelverhagen:
Depends on https://github.com/NuGet/NuGetGallery/pull/4764. Progress on https://github.com/NuGet/NuGetGallery/issues/4646.

This writes the new .nupkg to the `validation` container if the package is starting in the `Validating` status instead of the `Available` status. No product code actually sets the `Validating` status, but eventually the `AsynchronousPackageValidationInitiator` will do it .

https://github.com/NuGet/NuGetGallery/blob/d6b55e7f15f2d7963f382d4e9719ce6ede5eed1f/src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs#L59

### DIFF
--- a/tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
+++ b/tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
@@ -173,9 +173,52 @@ namespace NuGetGallery
 
         public class TheCommitPackageMethod : FactsBase
         {
-            [Fact]
-            public async Task SavesPackageToStorageAndDatabase()
+            public static IEnumerable<object[]> SupportedPackageStatuses => new[]
             {
+                new object[] { PackageStatus.Available },
+                new object[] { PackageStatus.Validating },
+            };
+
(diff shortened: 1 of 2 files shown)


tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
```diff
@@ -191,6 +234,25 @@ public class TheCommitPackageMethod : FactsBase
             }
 
             [Fact]
+            public async Task SavesPackageToStorageAndDatabaseWhenValidating()
```
loic-sharma: >SavesPackageToStorageAndDatabaseWhenValidating [](start = 30, length = 46)

Maybe this is overkill, but should we add a check that `SavePackageFileAsync` is never called when using `PackageStatus.Validating`?

joelverhagen: Thought about that, but originally decided against it since this kind of explodes as the number of states goes up. But I guess we can worry about that when we get there. I'll add this.

PR #4768, opened 2017-09-28
Add text to API Keys page

skofman1:
image


src/NuGetGallery/Views/Users/ApiKeys.cshtml
```diff
@@ -52,6 +52,8 @@
                 <p>
                     <strong>Always keep your API keys a secret!</strong> If one of your keys is accidentally revealed,
                     you can always generate a new one at any time. You can also remove existing API keys if necessary.
+                    <br />
+                    <strong>Note:</strong> We recommend using NuGet.exe v4.1.0 or above to push packages.
```
scottbommarito: I think for emphasis, the note should be on its own line (a new `<p>` tag).

skofman1: done

Issue #4683, opened 2017-09-14
Admin delete packages page should include whether the package is soft-deleted or not


scottbommarito:
![image](https://user-images.githubusercontent.com/18014088/30456429-64758ade-9958-11e7-89f4-690ac695e51b.png)

If you saw this, you would have no idea that 0.0.1 and 0.0.2 are already soft-deleted, and 0.0.3 is available.

This should likely be integrated with the new `PackageStatusKey` field. (https://github.com/NuGet/NuGetGallery/issues/4645)

joelverhagen:
It should also include the Owner(s).

joelverhagen:
> This should likely be integrated with the new `PackageStatusKey` field. 

This could be done with or without the `PackageStatusKey`. The `Deleted` flag would work just fine too.

Issue #4648, opened 2017-09-11
Update the delete UI wrt Validating and FailedValidation states


joelverhagen:
Deleting a Validating or FailedValidation package deletes from the "validating" nupkg container instead of "packages" container.

Search box on admin UI should search all states (including Validating and FailingValidation).

joelverhagen:
This will be validated as part of asynchronous validation E2E testing.

Issue #4230, opened 2017-06-19
Delete Packages search UI does not show any message for empty search results


mishra14:
stemming from https://github.com/NuGet/NuGetGallery/issues/4229

If the admin delete search does not have any matches, there is no message displayed indicating it and as a user, it seems like I am just pressing a dead button.

I dont think this affects external customers.

Issue #3856, opened 2017-05-02
Delete package UI should de-dupe package ID


joelverhagen:
When you enter the same ID twice, its versions appear 2 times each.

To repro, uncheck "Keep package id and version reserved" and check "Remove the package registration when all packages are deleted".

![image](https://cloud.githubusercontent.com/assets/94054/25636515/72f801fe-2f36-11e7-8824-020e8becd8ba.png)

After attempting to delete them all:
![image](https://cloud.githubusercontent.com/assets/94054/25636540/89d5c2ee-2f36-11e7-8ca1-137a7c0932e9.png)

Exception:
```
System.NullReferenceException
Object reference not set to an instance of an object.

System.NullReferenceException: Object reference not set to an instance of an object.
   at NuGetGallery.Auditing.PackageAuditRecord..ctor(Package package, AuditedPackageAction action, String reason)
   at NuGetGallery.PackageDeleteService.CreateAuditRecord(Package package, PackageRegistration packageRegistration, AuditedPackageAction action, String reason)
   at NuGetGallery.PackageDeleteService.<HardDeletePackagesAsync>d__10.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGetGallery.PackagesController.<Delete>d__34.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Web.Mvc.Async.TaskAsyncActionDescriptor.EndExecute(IAsyncResult asyncResult)
   at System.Web.Mvc.Async.AsyncControllerActionInvoker.<>c__DisplayClass37.<BeginInvokeAsynchronousActionMethod>b__36(IAsyncResult asyncResult)
   at System.Web.Mvc.Async.AsyncControllerActionInvoker.EndInvokeActionMethod(IAsyncResult asyncResult)
   at System.Web.Mvc.Async.AsyncControllerActionInvoker.AsyncInvocationWithFilters.<InvokeActionMethodFilterAsynchronouslyRecursive>b__3d()
[...]

PR #4774, opened 2017-09-28
Backup package in validation package container during delete flow (and other little fixes)

joelverhagen:
... and some little polishes that are low hanging fruit.

Address https://github.com/NuGet/NuGetGallery/issues/3856, https://github.com/NuGet/NuGetGallery/issues/4230, https://github.com/NuGet/NuGetGallery/issues/4648, and https://github.com/NuGet/NuGetGallery/issues/4683.

Many results:
screencapture-localhost-admin-delete-1506629861853

No results:
screencapture-localhost-admin-delete-1506629851995

_...this time without secrets..._
image

### DIFF
--- a/src/NuGetGallery/Areas/Admin/Controllers/DeleteController.cs
+++ b/src/NuGetGallery/Areas/Admin/Controllers/DeleteController.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Web.Mvc;
+using System.Web.Routing;
 using NuGet.Versioning;
 using NuGetGallery.Areas.Admin.ViewModels;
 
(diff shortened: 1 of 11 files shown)


joelverhagen:
🔔

src/NuGetGallery/Areas/Admin/Controllers/DeleteController.cs
```diff
@@ -83,8 +106,28 @@ private DeleteSearchResult CreateDeleteSearchResult(Package package)
...
+                PackageStatus = package.PackageStatusKey.ToString(),
+                Owners = package
+                    .PackageRegistration
+                    .Owners
+                    .Select(u => u.Username)
+                    .OrderBy(u => u)
+                    .Select(username => new UserViewModel
+                    {
+                        Username = username,
+                        ProfileUrl = Url.GetActionLink(
```
loic-sharma: >GetActionLink [](start = 41, length = 13)

Why aren't you using `Url.User` here? Also, what's the `area` route value? It seems the `UsersController.Profiles` only accepts a `username` and `page` parameter.

joelverhagen: By default, `area` is the current area, which is `Admin` in this case, which results in a 404 (since the profile page is not in the `Admin` area).

loic-sharma: Oh, I see. Would it be better to explicitly set the area for `Url.User`? I feel that `Url.GetActionLink` is an implementation detail that would be best kept private.

PR #4775, opened 2017-09-28
File storage read URI generation

agr:
Need a URI to an uploaded package for the validation orchestrator to know how to obtain the file to validate.
This change implements SAS-enabled URI generation for the cloud storage, the local file storage leaves it not implemented.

### DIFF
--- a/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
@@ -110,6 +110,19 @@ namespace NuGetGallery
             await blob.SetPropertiesAsync();
         }
 
+        public async Task<Uri> GetFileReadUriAsync(string folderName, string fileName, DateTimeOffset? endOfAccess)
+        {
+            folderName = folderName ?? throw new ArgumentNullException(nameof(folderName));
+            fileName = fileName ?? throw new ArgumentNullException(nameof(fileName));
+            if (endOfAccess.HasValue && endOfAccess < DateTimeOffset.UtcNow)
+            {
+                throw new ArgumentOutOfRangeException(nameof(endOfAccess), $"{nameof(endOfAccess)} is in the past");
+            }
+            ICloudBlobContainer container = await GetContainer(folderName);
+            var blob = container.GetBlobReference(fileName);
+            return new Uri(blob.Uri, blob.GetSharedReadSignature(endOfAccess));
+        }
+
         protected async Task<ICloudBlobContainer> GetContainer(string folderName)
         {
             ICloudBlobContainer container;
@@ -226,19 +239,6 @@ namespace NuGetGallery
                 });
 
             return container;
-        }
-
-        public async Task<Uri> GetFileReadUriAsync(string folderName, string fileName, DateTimeOffset? endOfAccess)
-        {
-            folderName = folderName ?? throw new ArgumentNullException(nameof(folderName));
-            fileName = fileName ?? throw new ArgumentNullException(nameof(fileName));
-            if (endOfAccess.HasValue && endOfAccess < DateTimeOffset.UtcNow)
-            {
-                throw new ArgumentOutOfRangeException(nameof(endOfAccess), $"{nameof(endOfAccess)} is in the past");
-            }
-            ICloudBlobContainer container = await GetContainer(folderName);
-            var blob = container.GetBlobReference(fileName);
-            return blob.GetSharedReadUri(endOfAccess);
         }
 
         private struct StorageResult
--- a/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
@@ -232,6 +232,10 @@ namespace NuGetGallery
         {
             folderName = folderName ?? throw new ArgumentNullException(nameof(folderName));
             fileName = fileName ?? throw new ArgumentNullException(nameof(fileName));
+            if (endOfAccess.HasValue && endOfAccess < DateTimeOffset.UtcNow)
+            {
+                throw new ArgumentOutOfRangeException(nameof(endOfAccess), $"{nameof(endOfAccess)} is in the past");
+            }
             ICloudBlobContainer container = await GetContainer(folderName);
             var blob = container.GetBlobReference(fileName);
             return blob.GetSharedReadUri(endOfAccess);
--- a/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
@@ -25,14 +25,14 @@ namespace NuGetGallery
 
         public async Task DeleteFileAsync(string folderName, string fileName)
         {
-            ICloudBlobContainer container = await GetContainer(folderName);
+            ICloudBlobContainer container = await GetContainerAsync(folderName);
             var blob = container.GetBlobReference(fileName);
             await blob.DeleteIfExistsAsync();
         }
 
         public async Task<bool> FileExistsAsync(string folderName, string fileName)
         {
-            ICloudBlobContainer container = await GetContainer(folderName);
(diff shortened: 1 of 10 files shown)


src/NuGetGallery.Core/Services/CorePackageFileService.cs
```diff
@@ -80,6 +80,18 @@ public Task DeleteValidationPackageFileAsync(string id, string version)
             return _fileStorageService.DeleteFileAsync(CoreConstants.ValidationFolderName, fileName);
         }
 
+        public Task<Uri> GetValidationPackageReadUriAsync(Package package, DateTimeOffset? endOfAccess)
```
loic-sharma: Do we need any sanity checks on `endOfAccess` to make sure it isn't too far in the future?

src/NuGetGallery/Services/CloudBlobWrapper.cs
```diff
@@ -124,6 +124,19 @@ public Task FetchAttributesAsync()
...
 
+        public Task<Uri> GetSharedReadUriAsync(DateTimeOffset? endOfAccess)
+        {
+            var accessPolicy = new SharedAccessBlobPolicy
+            {
+                SharedAccessExpiryTime = endOfAccess,
+                Permissions = SharedAccessBlobPermissions.Read
+            };
+
+            var token = this._blob.GetSharedAccessSignature(accessPolicy);
```
joelverhagen: Can we make this a slimmer wrapper and just return the token here?

src/NuGetGallery/Services/CloudBlobWrapper.cs
```diff
@@ -124,6 +124,19 @@ public Task FetchAttributesAsync()
                 state: null);
         }
 
+        public Task<Uri> GetSharedReadUriAsync(DateTimeOffset? endOfAccess)
```
joelverhagen: Why `async`?

src/NuGetGallery.Core/Services/CorePackageFileService.cs
joelverhagen: Unit tests.

src/NuGetGallery/Services/FileSystemFileStorageService.cs
```diff
@@ -173,6 +173,18 @@ public Task<bool> IsAvailableAsync()
             return Task.FromResult(Directory.Exists(_configuration.FileStorageDirectory));
         }
 
+        public Task<Uri> GetFileReadUriAsync(string folderName, string fileName, DateTimeOffset? endOfAccess)
+        {
+            // technically, we would be able to generate the file:/// url here, but we don't need it right now
```
joelverhagen: 👍

src/NuGetGallery/Services/CloudBlobWrapper.cs
agr: Initially, I expected the `GetSharedAccessSignature` to be `async` (as most Azure API are)...

src/NuGetGallery.Core/Services/CorePackageFileService.cs
agr: It seems like an unnecessary limitation. Might be worth checking that it's not in the past though.

src/NuGetGallery/Services/CloudBlobWrapper.cs
agr: The `token` here is not actually a token, but a complete query part of the URI (a string that starts with "?" and contains a bunch of key-value pairs in it) and it's blob specific, so it make little sense in existing on its own (I am not sure how could anyone would use it other than append it to the blob URI) and on the other hand it would be easy to misuse it (append to a wrong blob URI for example).

joelverhagen: I mean, storage SDK itself has the method so... 😄

joelverhagen: Can we just move the token + URI concatenation to the `CloudBlobCoreFileStorageService`? Then the wrapper is thin and the concatenation can be unit tested.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -228,6 +228,19 @@ private static string GetContentType(string folderName)
...
 
+        public async Task<Uri> GetFileReadUriAsync(string folderName, string fileName, DateTimeOffset? endOfAccess)
+        {
+            folderName = folderName ?? throw new ArgumentNullException(nameof(folderName));
+            fileName = fileName ?? throw new ArgumentNullException(nameof(fileName));
+            if (endOfAccess.HasValue && endOfAccess < DateTimeOffset.UtcNow)
+            {
+                throw new ArgumentOutOfRangeException(nameof(endOfAccess), $"{nameof(endOfAccess)} is in the past");
+            }
+            ICloudBlobContainer container = await GetContainer(folderName);
```
cristinamanum: If await on GetContainer, it would be good to have it called GetContainerAsync

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -228,6 +228,19 @@ private static string GetContentType(string folderName)
             return container;
         }
 
+        public async Task<Uri> GetFileReadUriAsync(string folderName, string fileName, DateTimeOffset? endOfAccess)
+        {
```
cristinamanum: Why does endOfAccess need to be nullable?

src/NuGetGallery.Core/Services/ICoreFileStorageService.cs
```diff
@@ -23,6 +24,15 @@ public interface ICoreFileStorageService
         /// <returns>A <see cref="IFileReference"/> representing the file reference</returns>
         Task<IFileReference> GetFileReferenceAsync(string folderName, string fileName, string ifNoneMatch = null);
 
+        /// <summary>
+        /// Generates the storage file URI (which is optionally time limited)
+        /// </summary>
+        /// <param name="folderName">The folder containing the file.</param>
+        /// <param name="fileName">The file within the <paramref name="folderName"/>.</param>
+        /// <param name="endOfAccess">Optional end of access timestamp.</param>
```
cristinamanum: This is not optional, Maybe it has a specific behavior for null.

src/NuGetGallery.Core/Services/ISimpleCloudBlob.cs
```diff
@@ -26,5 +26,16 @@ public interface ISimpleCloudBlob
         Task UploadFromStreamAsync(Stream packageFile, bool overwrite);
 
         Task FetchAttributesAsync();
+
+        /// <summary>
+        /// Generates the URI that can be used to read the contents of the blob
+        /// using the produced URI only (without storage account credentials).
+        /// </summary>
+        /// <param name="endOfAccess">
+        /// Optional "end of access" timestamp. After the specified time,
```
cristinamanum: Not optional

src/NuGetGallery/Services/CloudBlobWrapper.cs
```diff
@@ -124,6 +124,19 @@ public Task FetchAttributesAsync()
                 state: null);
         }
 
+        public Uri GetSharedReadUri(DateTimeOffset? endOfAccess)
```
cristinamanum: In case that we need another permission in the future we would need to recreate same method with one change for Permission. Would we like to add Permissions as argument and change the method name?

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
agr: This is a generic file access layer, it should be able to produce non-time limited links. Package service is more limited and hence the time limit is required there.

src/NuGetGallery/Services/CloudBlobWrapper.cs
agr: I do not foresee other permissions to be needed. If service requires the permissions to create or update blobs it probably needs full set of credentials to storage account with write access.

Issue #4753, opened 2017-09-26
Content Service Caching isn't being used


ryuyu:
The contentService that is used by the PagesController (https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Controllers/PagesController.cs) doesn't appear to being persisted.

Every time the PagesController is called, a new contentService instance is given to it. This means that the old instance with the cached content is actually not used. This leads to us always retrieving the content from blob storage and never hitting the cache.

I believe this is due to the fact that we don't register the content service as a single instance with the DI container (not sure i'm using the right terminology here). I think adding a reference to the content service indicating that it should be single instance in https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs should fix this issue (it did for me locally).

PR #4776, opened 2017-09-28
Fix caching of static files.

ryuyu:
Addresses https://github.com/NuGet/NuGetGallery/issues/4753
Adds a new property so we can tell if a storageDependent is supposed to be a SingleInstance or not.

@joelverhagen @skofman1

### DIFF
--- a/src/NuGetGallery/App_Start/StorageDependent.cs
+++ b/src/NuGetGallery/App_Start/StorageDependent.cs
@@ -17,12 +17,14 @@ namespace NuGetGallery
             string bindingKey,
             string azureStorageConnectionString,
             Type implementationType,
-            Type interfaceType)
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/App_Start/StorageDependent.cs
```diff
@@ -54,14 +57,20 @@ public StorageDependent SetBindingKey(string bindingKey)
         /// </summary>
         public Type InterfaceType { get; }
 
+        /// <summary>
+        /// Indicates if the implementation type should be declared as a singleton in DI container.
```
loic-sharma: Super nit: `DI` -> `dependency injection`. Acronyms are evil and should be eradicated :)

PR #4791, opened 2017-10-02
Update links to docs on register page to use new configuration

ryuyu:
Updates the Terms of Use and Privacy Policy links on register page to use the links provided by the customization.

@joelverhagen @skofman1 @karann-msft

### DIFF
--- a/src/NuGetGallery/UrlExtensions.cs
+++ b/src/NuGetGallery/UrlExtensions.cs
@@ -756,16 +756,31 @@ namespace NuGetGallery
 
         public static string Terms(this UrlHelper url, bool relativeUrl = true)
         {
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/UrlExtensions.cs
```diff
@@ -756,16 +756,31 @@ public static string ContactOwners(this UrlHelper url, string id, bool relativeU
 
         public static string Terms(this UrlHelper url, bool relativeUrl = true)
         {
+            if (!String.IsNullOrEmpty(_configuration.Current.ExternalTermsOfUseUrl))
+            {
+                return _configuration.Current.ExternalTermsOfUseUrl;
```
joelverhagen: Can we observe the `relativeUrl` parameter here? Or can we throw if it does not match the configured URL? (e.g. if it is true?)

Issue #4786, opened 2017-10-02
Gallery-Broken link when navigating from admin page to account page


skofman1:
Repro:
1. Login
2. Go to Admin page
3. Click on user name (to go to account page)

Expected: Account page is opened
Actual: 404

The link to account page is: https://www.nuget.org/Admin/Users/Account

joelverhagen:
This is due to an implicit `Admin` area associated with those links.

PR #4792, opened 2017-10-02
Specify the empty string area in UrlExtensions unless otherwise specified

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/4786.

/cc @xavierdecoster, could you approve this one? You've been doing a lot of work in this area.

/cc @loic-sharma, we talked about this before. I dug into the code via ILSpy and read online. Looks like always specifying the empty area is the right move.

### DIFF
--- a/src/NuGetGallery/UrlExtensions.cs
+++ b/src/NuGetGallery/UrlExtensions.cs
@@ -7,11 +7,13 @@ using System.Globalization;
 using System.Web;
 using System.Web.Mvc;
 using System.Web.Routing;
+using NuGetGallery.Areas.Admin;
 
 namespace NuGetGallery
 {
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/UrlExtensions.cs
```diff
@@ -776,10 +769,7 @@ public static string Admin(this UrlHelper url, bool relativeUrl = true)
                 "Index",
                 "Home",
                 relativeUrl,
-                routeValues: new RouteValueDictionary
-                {
-                    { "area", "Admin" }
-                });
+                area: "Admin");
```
loic-sharma: >Admin [](start = 23, length = 5)

Could we make this  string a const somewhere?

PR #4796, opened 2017-10-03
Latest Version on Manage Packages does not show SemVer 2.0.0 packages

### LINKED ISSUE #4782: Latest Version on Manage Packages does not show SemVer 2.0.0 packages
Latest Version on Manage Packages does not show SemVer 2.0.0 packages in some scenarios.

In the below example all packages have SemVer 2.0.0 stable versions. Only SleetLib shows the correct version, Sleet for example should be 2.0.0.

image

xavierdecoster:
fixes #4782

### DIFF
--- a/src/NuGetGallery/Views/Users/_UserPackagesList.cshtml
+++ b/src/NuGetGallery/Views/Users/_UserPackagesList.cshtml
@@ -49,7 +49,7 @@
                             }
                             </td>
                             <td class="align-middle text-nowrap">@package.TotalDownloadCount.ToNuGetNumberString()</td>
-                            <td class="align-middle text-nowrap">@package.Version</td>
+                            <td class="align-middle text-nowrap">@package.FullVersion.Abbreviate(25)</td>
                             <td class="text-right align-middle package-controls">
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Users/_UserPackagesList.cshtml
```diff
@@ -49,7 +49,7 @@
                             }
                             </td>
                             <td class="align-middle text-nowrap">@package.TotalDownloadCount.ToNuGetNumberString()</td>
-                            <td class="align-middle text-nowrap">@package.Version</td>
+                            <td class="align-middle text-nowrap">@package.FullVersion.Abbreviate(25)</td>
```
joelverhagen: Do we need to abbreviate? Can we instead add wrapping? (we do this on package details page)

xavierdecoster: See https://github.com/NuGet/NuGetGallery/issues/4782#issuecomment-333562540. Metadata may become very long, and the column is really not that wide.

PR #4798, opened 2017-10-03
4701: Return 404, not 200, when package doesn't exist.

rafcioki:


src/NuGetGallery/Controllers/StatisticsController.cs
```diff
@@ -178,7 +178,14 @@ public StatisticsController(IStatisticsService statisticsService, IAggregateStat
...
 
-            return Json(await GetPackageDownloadsByVersionReport(id, groupby), JsonRequestBehavior.AllowGet);
+            var packageStatisticsReport = await GetPackageDownloadsByVersionReport(id, groupby);
+
+            if (packageStatisticsReport == null)
+            {
+                return new HttpStatusCodeResult(HttpStatusCode.NotFound);
+            }
+
+            return Json(packageStatisticsReport, JsonRequestBehavior.AllowGet);
```
joelverhagen: There is an overload `Json(int httpStatusCode, object data, JsonRequestBehavior behavior)` which allows you to both return `Content-Type: application/json` and set the status code to 200 or 404. Could you use that instead?

At that point you can change the method to return `Task<JsonResult>` instead of the more generic `Task<ActionResult>`.

src/NuGetGallery/Controllers/StatisticsController.cs
```diff
@@ -178,7 +178,14 @@ public StatisticsController(IStatisticsService statisticsService, IAggregateStat
                 return new HttpStatusCodeResult(HttpStatusCode.NotFound);
             }
 
-            return Json(await GetPackageDownloadsByVersionReport(id, groupby), JsonRequestBehavior.AllowGet);
+            var packageStatisticsReport = await GetPackageDownloadsByVersionReport(id, groupby);
+
+            if (packageStatisticsReport == null)
```
joelverhagen: Could we have unit tests to cover this change?
https://github.com/NuGet/NuGetGallery/blob/master/tests/NuGetGallery.Facts/Controllers/StatisticsControllerFacts.cs

rafcioki:
Thanks for your comments @joelverhagen. Pushed another commit to cover the stuff you've mentioned, please take a look when you'll have a moment.

joelverhagen:
Merged with efb23fcd310080e99b02f115ec9ee82c822bca8d.

PR #4818, opened 2017-10-09
Update local reference urls to redirect to external if specified.

ryuyu:
@joelverhagen @skofman1 @karann-msft

### DIFF
--- a/src/NuGetGallery/Controllers/PagesController.cs
+++ b/src/NuGetGallery/Controllers/PagesController.cs
@@ -50,7 +50,7 @@ namespace NuGetGallery
         [HttpGet]
         public virtual ActionResult About()
         {
-            if (!String.IsNullOrEmpty(_galleryConfigurationService.Current.ExternalAboutUrl))
+            if (!string.IsNullOrEmpty(_galleryConfigurationService.Current.ExternalAboutUrl))
             {
                 return Redirect(_galleryConfigurationService.Current.ExternalAboutUrl);
             }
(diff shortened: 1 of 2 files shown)


joelverhagen:
Can we get unit tests on this?

tests/NuGetGallery.Facts/Controllers/PagesControllerFacts.cs
```diff
@@ -84,5 +85,63 @@ public class TheContactAction : TestContainer
...
+            }
+
+            [Fact]
+            public void ExternalTermsOfUseUrlRedirects()
+            {
+                var customUrl = "http://TermsOfUse";
+                var configuration = GetConfigurationService();
+                
+                var controller = GetController<PagesController>();
+                var result = controller.Terms().Result;
```
joelverhagen: use `await` by making the test `async Task`.

tests/NuGetGallery.Facts/Controllers/PagesControllerFacts.cs
```diff
@@ -84,5 +85,63 @@ public class TheContactAction : TestContainer
...
+                Assert.IsType<ViewResult>(result);
+
+                configuration.Current.ExternalTermsOfUseUrl = customUrl;
+                controller = GetController<PagesController>();
+                result = controller.Terms().Result;
+
+                Assert.IsType<RedirectResult>(result);
```
joelverhagen: This should be a different "Fact" (i.e. test)

joelverhagen: In other words, one act+assert per test

src/NuGetGallery/Controllers/PagesController.cs
```diff
@@ -46,6 +50,11 @@ public virtual ActionResult Page(string pageName)
         [HttpGet]
         public virtual ActionResult About()
         {
+            if (!String.IsNullOrEmpty(_galleryConfigurationService.Current.ExternalAboutUrl))
```
skofman1: string instead of String everywhere

scottbommarito:
@ryuyu @skofman1 @karann-msft 
What's the status of this?

skofman1:
@ryuyu , is this still relevant? Which issue does this resolve?

Issue #4823, opened 2017-10-10
Limit the upload size on the packages.


shishirx34:
I am not sure what the upload size restrictions on nupkgs are, but we have a package of size 655MB, which is huge. May be limit the size of packages to 500MB?

The nupkg in question: 
```
https://api.nuget.org/packages/visualcpptools.community.vs2017layout.14.11.25547.nupkg
```
This has caused an index delay due to timeout issues in uploading huge files to blob storage in China.

shishirx34:
It seems that we do have the IIS restriction on large size request via `maxAllowedContentLength` set to `250MB`. However, this doesn't seem to prevent the CLI from uploading the large size file. 

From what I gather online, with IIS 7 and above there are two values that you need to set for restricting request size: `maxAllowedContentLength` and `maxRequestLength`. Currently the `maxRequestLength` is set to a value of ~2000GB, but the smaller of the two values should take a precedence. For some reason this doesn't seem to be taking effect from CLI, you can still upload large size files from CLI. The web UI upload flow doesn't let anyone upload large size file. There is possibly another [bug](https://github.com/NuGet/NuGetGallery/issues/4833) with handling large file uploads correctly.

We need to set `maxRequestLength` value to `256000` (250MB => 256000KB) to restrict the uploads from CLI with large file size and possibly try to find why CLI upload flow doesn't respect the `maxAllowedContentLength` property.

PR #4826, opened 2017-10-11
Restrict upload size for packages

shishirx34:
Fixes https://github.com/NuGet/NuGetGallery/issues/4823. This should also prevent the index lag caused by humongous file sizes, like the issue that happened today.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -236,6 +236,7 @@ namespace NuGetGallery
             var currentUser = GetCurrentUser();
             if (uploadFile != null && uploadFile.ContentLength > Constants.MaxUploadFileSizeInBytes)
             {
+                ModelState.AddModelError(String.Empty, string.Format(Strings.UploadFileSizeExceedsMaxLimit, Constants.MaxUploadFileSizeInMB));
                 return (Json(400, new[] { string.Format(Strings.UploadFileSizeExceedsMaxLimit, Constants.MaxUploadFileSizeInMB) }));
             }
 
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -234,7 +234,7 @@ namespace NuGetGallery
         public virtual async Task<JsonResult> UploadPackage(HttpPostedFileBase uploadFile)
         {
             var currentUser = GetCurrentUser();
-            if (uploadFile.ContentLength > Constants.MaxUploadFileSizeInBytes)
+            if (uploadFile != null && uploadFile.ContentLength > Constants.MaxUploadFileSizeInBytes)
             {
                 return (Json(400, new[] { string.Format(Strings.UploadFileSizeExceedsMaxLimit, Constants.MaxUploadFileSizeInMB) }));
             }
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Strings.resx
```diff
@@ -600,4 +600,7 @@ For more information, please contact '{2}'.</value>
   <data name="CannotEnqueueDueToReadOnly" xml:space="preserve">
     <value>Enqueuing unavailable: the gallery is currently in read only mode, with limited service. Please try again later.</value>
   </data>
+  <data name="UploadFileSizeExceedsMaxLimit" xml:space="preserve">
+    <value>The uploaded file is too big. Maximum allowed size is {0}MB for the upload file.</value>
```
shishirx34: Feel free to suggest better wording here.

scottbommarito:
I would make the max file size configurable, so that people hosting the gallery can set it as they want.

Otherwise, PR looks great
________________________________

joelverhagen:
How many packages do we have today that violate this limit?

I thought we had this validation already. Did we lose it at some point?

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1905,6 +1905,27 @@ public class TheUploadFileActionForPostRequests
...
+                        GetConfigurationService(),
+                        uploadFileService: fakeUploadFileService);
+                    controller.SetCurrentUser(TestUtility.FakeUser);
+
+                    var postedFileBase = new Mock<HttpPostedFileBase>();
+                    postedFileBase.Setup(x => x.ContentLength).Returns(251 * 1024 * 1024 /*251MB*/);
+                    var result = await controller.UploadPackage(postedFileBase.Object) as JsonResult;
+
+                    Assert.NotNull(result);
+                    Assert.False(controller.ModelState.IsValid);
```
joelverhagen: Assert message here to make sure the right validation is failing.

joelverhagen:
How did we come up with the value 250 MB?

shishirx34:
Svetlana pointed out that there is an IIS setting that should restrict big file uploads. I am testing this out.
The size value was determined by the documentation [here](https://docs.microsoft.com/en-us/nuget/policies/nuget-faq).

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
shishirx34: sure.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -234,6 +234,11 @@ public virtual JsonResult UploadPackageProgress()
         public virtual async Task<JsonResult> UploadPackage(HttpPostedFileBase uploadFile)
         {
             var currentUser = GetCurrentUser();
+            if (uploadFile != null && uploadFile.ContentLength > Constants.MaxUploadFileSizeInBytes)
```
chenriksson: Can we move the 'if (uploadFile == null)' check above this, so we don't need another null check?

src/NuGetGallery/Strings.resx
chenriksson: How about just "The uploaded file exceeded the package size limit of 250MB."

shishirx34:
Setting this via ASP.NETs property instead of code enforcer.

PR #4829, opened 2017-10-11
Add notification message to edit/verify package view

xavierdecoster:


src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -88,6 +88,31 @@
...
+        )
+}
+
+@helper AlertPackageVerifyRecommendation()
+{
+    @AlertWarning(
+    @<text>
+        Please verify the package metadata below.<br/>
+        <em>
+            To make changes, we recommend you cancel the upload, make the edits in the nuspec, and upload a new package.
```
joelverhagen: I don't think we should prescribe "edits to the nuspec". Yes, that is one way to fix the issue but just as likely they can fix their project file and rerun `nuget.exe pack`. Maybe we can be more generic and say "edit the package metadata" and link to some docs about what the possibilities are:

1. update csproj then repack
1. update input .nuspec then repack

joelverhagen: @rohit21agrawal, what do you think the wording should be here? The point I'm trying to get at is that we should guide users to leverage our tools to make the package have the correct metadata. Manual edits to the output nuspec is pretty much the last option we want to suggest.

xavierdecoster:
@joelverhagen followed [the spec](https://github.com/NuGet/Home/wiki/Package-Immutability#package-edit-page-for-published-packages) provided by @karann-msft last night, but I do agree with you that guiding towards manual nuspec edits may not be the desired recommendation. Preferably, the link explains the actual steps, whilst the notification message is a tad bit more generic.

src/NuGetGallery/App_Code/ViewHelpers.cshtml
scottbommarito: Something else to consider is that if we mention our client, we may also want to mention third-party clients as well.

xavierdecoster:
Updated wording based on https://github.com/NuGet/Engineering/issues/833#issuecomment-335983192. Any other changes needed to this?

xavierdecoster:
🔔

xavierdecoster:
@joelverhagen ok with the updates?

PR #4830, opened 2017-10-11
X-NuGet-Protocol-Version for Functional tests

skofman1:


tests/NuGetGallery.FunctionalTests.Core/Constants.cs
```diff
@@ -9,6 +9,7 @@ public static class Constants
         public const string NuGetHeaderApiKey = "X-NuGet-ApiKey";
         public const string NuGetHeaderClientVersion = "X-NuGet-Client-Version";
```
joelverhagen: Should this be used at all?

skofman1: X-NuGet-Client-Version is used by SecurePush tests to verify we still support this header. If we stop support 4.1 nuget.exe will break.

joelverhagen: Good point!

Issue #4510, opened 2017-08-04
Define criteria for listing clients on NuGet.org’s package details page


unniravindranathan:
For reference, see this pull request from Paket as an additional tab: https://github.com/NuGet/NuGetGallery/pull/4437

falfaddaghi:
why should there be any criteria aside from being an actual working client?  as stated in the last pr maven lists multiple clients. I don't understand the reluctance here at all.  can you maybe explain your concerns?  if nuget change a protocol and that break Paket which in turn makes the command displayed not correct how does that affect nuget.org ?

unniravindranathan:
@falfaddaghi I will admit that I don't have a well-thought out response yet, so please consider this as me just thinking out loud - curious what you think?

- Once we agree that a particular experience is core to the .NET developer experience (security, platform support, etc.) - should we mandate that the client support that? How do we reason with the urgency around supporting a particular experience - wait for everyone to be ready, etc.?

- Our data currently shows that NuGet.org is accessed by 10s of different kinds of clients, if not more (hard to know for sure, since we don't have a contract for the client to share a well know name and version :( ). What is the objective criteria we can use here to triage requests?

- We will almost certainly need to do additional more UX work to give the user more information so they can make an informed choice about who supports the client, what scenarios work, what don't, and so on.

Rickasaurus:
Trying to maintain so much control over the .NET developer experience is what has driven so many .NET developers away in the first place. The problem is that there is no one size fits all experience. We need to be able to make and use tools that fit our particular needs. For some that is out of the box one stop shopping, for others it's customizing to reduce the burden of certain kinds of common developer tasks (this is what Paket does really well). 

We are a small team with many clients and so hundreds of .NET projects. Keeping them up to date and in sync with Nuget was a maintenance nightmare. However, the standard Nuget client might be just fine for a team focused on one single or a few products.

Security is important and I think all clients should be made to be secure by enforcing a secure protocol. However, I think mandating platform support goes too far. If it works for the people who opt in to use it what's the problem? There's nothing stopping you from using the standard Nuget client in some projects and Paket in others where the dependency burden has become overwhelming.

ReedCopsey:
@unniravindranathan I'll try to reply my thoughts in line:

> Once we agree that a particular experience is core to the .NET developer experience (security, platform support, etc.) - should we mandate that the client support that? How do we reason with the urgency around supporting a particular experience - wait for everyone to be ready, etc.?

[...]

PR #4834, opened 2017-10-12
Add support for Protobuild CLI

hach-que:
This adds support for showing instructions on using [Protobuild](https://protobuild.org/) to add or install a NuGet package.

In addition, it detects packages that were created with Protobuild and thus have additional metadata included in them to make cross-platform projects work better, and recommends the use of the client in these scenarios.

image

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -6,7 +6,30 @@
 
     var absolutePackageUrl = Url.Absolute(Url.Package(Model.Id));
 
-    var packageManagers = new PackageManagerViewModel[]
+    var protobuildTerm = "add";
+    var protobuildPreferred = false;
+    var versionSuffix = "@" + Model.Version;
+    var isProtobuildEnhanced = false;
+    if (Model.Tags.Any(x => x.StartsWith("platforms=")))
+    {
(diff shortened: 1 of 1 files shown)


anangaur:
In order to evaluate whether a CLI can be listed on the package details page, we look at a [pre-defined set of criteria](https://github.com/NuGet/NuGetGallery/issues/4510#issuecomment-324422213). In the case of this PR, we evaluated that the criteria of “Significant adoption in .NET community” is not something we can currently establish and hence we will not be able to accept this PR.

NuGet.org uses the “User Agent” string supplied by the CLI to determine usage. In case you are not already setting the “User Agent” in NuGet.org requests, you should consider setting this to something that will help us uniquely identify usage of the client.

If you would like to discuss further about this, please create an issue on [NuGet/NuGetGallery](https://github.com/nuget/nugetgallery/issues) repo or reach out to us at [support@nuget.org](mailto:support@nuget.org).

anangaur:
Closing the PR for now (See above comment for details).
If you would like to discuss further about this, please create an issue on [NuGet/NuGetGallery](https://github.com/nuget/nugetgallery/issues) repo or reach out to us at [support@nuget.org](mailto:support@nuget.org).

hach-que:
We'll add `User-Agent` (since I'm having to update the binary to support the `X-NuGet-Protocol-Version: 4.1.0` header). However, it should be noted that to date there's over 300,000 downloads of packages which support the Protobuild-enhanced package protocol.

PR #4844, opened 2017-10-13
Added credential key to AuditActor

skofman1:
Moved some of the AuditActor creation code from NuGetGallery.Core to NuGetGallery to allow usage of NuGetClaims.
Let me know if this looks good, and I will apply changes to Shims.

### DIFF
--- a/src/NuGetGallery.Core/Auditing/AuditActor.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditActor.cs
@@ -8,7 +8,6 @@ using System.Net;
 using System.Net.NetworkInformation;
 using System.Net.Sockets;
 using System.Threading.Tasks;
-using System.Web;
 
 namespace NuGetGallery.Auditing
 {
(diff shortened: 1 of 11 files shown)


src/NuGetGallery.Core/Auditing/AuditActor.cs
```diff
@@ -97,6 +51,7 @@ public static Task<AuditActor> GetCurrentMachineActorAsync()
                 ipAddress,
                 $@"{Environment.UserDomainName}\{Environment.UserName}",
                 "MachineUser",
+                string.Empty,
```
scottbommarito: Should we add a parameter to this method to allow passing in this value instead of defaulting to `string.Empty` always?

tests/NuGetGallery.Core.Facts/Auditing/AuditActorTests.cs
```diff
@@ -55,33 +53,35 @@ public void Constructor_WithOnBehalfOf_AcceptsNullValues()
...
                 timeStampUtc: DateTime.MinValue);
 
-            Assert.Equal("", actor.MachineName);
-            Assert.Equal("", actor.MachineIP);
-            Assert.Equal("", actor.UserName);
-            Assert.Equal("", actor.AuthenticationType);
+            Assert.Equal(string.Empty, actor.MachineName);
+            Assert.Equal(string.Empty, actor.MachineIP);
+            Assert.Equal(string.Empty, actor.UserName);
+            Assert.Equal(string.Empty, actor.AuthenticationType);
```
scottbommarito: Add a check to `credentialKey` here.

```
Assert.Equal(string.Empty, actor.CredentialKey);
```

Same with `Constructor_WithOnBehalfOf_AcceptsEmptyStringValues`

src/NuGetGallery.Core/Auditing/AuditActor.cs
skofman1: In this specific scenario there is no "CredentialKey", since the actor is a MachineUser

src/NuGetGallery/Helpers/AuditingHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+                var claimsIdentity = context.User.Identity as ClaimsIdentity;
+                credentialKey = claimsIdentity?.GetClaimOrDefault(NuGetClaims.CredentialKey);
+            }
+
+            return Task.FromResult(new AuditActor(
+                null,
+                clientIpAddress,
+                user,
+                authType,
+                credentialKey,
```
cristinamanum: I did not see used anywhere(except tests) the new CredentialKey Get. Should it be?

chenriksson:
@skofman1 For Shims, do we need to change callerIdentity to use combination of credentialKey and usernameOrEmail?

tests/NuGetGallery.Core.Facts/Auditing/AuditActorTests.cs
skofman1: done

src/NuGetGallery/Helpers/AuditingHelper.cs
chenriksson: It's required for json serialization. Audit records will be serialized and provided to the auditing services. The cloud auditing service writes the whole record to blob storage.

skofman1:
@chenriksson , I'm adding the Shims changes now. The current plan is to add another CallerIdentity.

PR #4847, opened 2017-10-13
Add a hidden skip to content link.

ryuyu:
Page view on load:
image

Page View on First tab
image

This element is only reachable by tab navigation.

@joelverhagen @skofman1 @xavierdecoster


src/NuGetGallery/Views/Shared/Gallery/Layout.cshtml
```diff
@@ -7,7 +7,7 @@
     @RenderSection("SocialMeta", required: false)
     @RenderSection("Meta", required: false)
 
-    <title>
+    <title>
```
scottbommarito: nit: get rid of extra whitespace

Issue #4812, opened 2017-10-09
Clicking on "Show more" for Documentation has incorrect chevron


anangaur:
Feature discussion issue: https://github.com/NuGet/NuGetGallery/issues/2280

Instead of ChevronDown it's ChevronRight

![image](https://user-images.githubusercontent.com/14800916/31326318-d700a19e-ac7a-11e7-8d41-fe978f35e016.png)

PR #4849, opened 2017-10-13
Do not toggle doc chevron on clamp (#4812)

chenriksson:
The documentation section (expander heading) contains a nested expander which toggles between the full and clamped versions. Clicking 'show [more|less]' for the inner expander was triggering the parent's hide/show collapse handlers and causing the Documentation chevron to toggle incorrectly. This change only toggles icons when the current target is the one that triggered the event.


chenriksson:
Closing - no longer reproduces with clamp fix (PR #4858).

PR #4853, opened 2017-10-17
Documentation parser fixes

### LINKED ISSUE #4783: < and > from imported markdown files display as &lt; and &gt;
The gallery does not display angle brackets in code blocks correctly in the package's documentation section comes from importing a markdown README file.

# Repro
1. Edit a package. E.g. https://www.nuget.org/packages/Yarn.MSBuild/1.0.1/Edit
1. Add this URL to the "Import Readme" => "URL" box: https://raw.githubusercontent.com/natemcmaster/Yarn.MSBuild/v1.0.1/README.md. This file contains this section
````md
**Package Manager Console in Visual Studio**
```
PM> Install-Package Yarn.MSBuild
```
````

# Expected result

image

# Actual result
It displays brackets as `&gt;` and `&lt;`
image

### PR DESCRIPTION
**Pull request by chenriksson**
Fixes #4783 and #4816, plus restricts markdown links to http/s

### DIFF
--- a/src/NuGetGallery/Services/ReadMeService.cs
+++ b/src/NuGetGallery/Services/ReadMeService.cs
@@ -17,6 +17,9 @@ namespace NuGetGallery
 {
     internal class ReadMeService : IReadMeService
     {
+        private static readonly Regex BlockQuotePattern = new Regex("^ {0,3}&gt;", RegexOptions.Multiline);
+        private static readonly Regex LinkPattern = new Regex("<a href=([\"\']).*?\\1");
+
         internal const string TypeUrl = "Url";
         internal const string TypeFile = "File";
         internal const string TypeWritten = "Written";
@@ -24,6 +27,7 @@ namespace NuGetGallery
         internal const int MaxMdLengthBytes = 8000;
         private const int ReadMeClampedLineCount = 10;
         private const string UrlHostRequirement = "raw.githubusercontent.com";
+
         private static readonly TimeSpan UrlTimeout = TimeSpan.FromSeconds(10);
 
         private IPackageFileService _packageFileService;
@@ -142,30 +146,6 @@ namespace NuGetGallery
 
             return hasReadMe;
         }
-        
-        /// <summary>
-        /// HTML encode content except for markdown blockquotes.
-        /// </summary>
-        /// <param name="readMeMd">ReadMe.md content.</param>
-        /// <returns>Encoded ReadMe.md content.</returns>
-        private static string HtmlEncodeExceptBlockquotes(string readMeMd)
-        {
-            var encoded = HttpUtility.HtmlEncode(readMeMd);
-            var encodedLines = encoded.Replace("\r\n", "\n").Split('\n');
-
-            var blockQuotePattern = new Regex("^ {0,3}&gt;");
-            for (int i = 0; i < encodedLines.Length; i++)
-            {
-                var match = blockQuotePattern.Match(encodedLines[i]);
-                if (match.Success)
-                {
-                    // For now, only decode the blockquote marker and not any inline html.
-                    encodedLines[i] = "> " + encodedLines[i].Substring(match.Length);
-                }
-            }
-
-            return string.Join(Environment.NewLine, encodedLines);
-        }
 
         /// <summary>
         /// Get converted HTML for readme.md string content.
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/NuGetGallery/Services/ReadMeService.cs [hunk 1]
```diff
@@ -141,6 +142,30 @@ public bool HasReadMeSource(ReadMeRequest readMeRequest)
...
+        /// HTML encode content except for markdown blockquotes.
+        /// </summary>
+        /// <param name="readMeMd">ReadMe.md content.</param>
+        /// <returns>Encoded ReadMe.md content.</returns>
+        private static string HtmlEncodeExceptBlockquotes(string readMeMd)
+        {
+            var encoded = HttpUtility.HtmlEncode(readMeMd);
+            var encodedLines = encoded.Replace("\r\n", "\n").Split('\n');
+
+            var blockQuotePattern = new Regex("^ {0,3}&gt;");
```
loic-sharma: Would it be better if `blockQuotePattern` was a static readonly property?

loic-sharma: Also, this `HtmlEncodeExceptBlockquotes` method seems rather expensive. Would it be worthwhile to check if `readMeMd` has the `&gt;` substring, and if it doesn't, simply return `encoded`.

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -149,10 +174,65 @@ public bool HasReadMeSource(ReadMeRequest readMeRequest)
...
+                    }
+                }
+            }
+
+            // CommonMark.Net does not support link attributes, so manually inject nofollow.
+            using (var htmlWriter = new StringWriter())
+            {
+                CommonMarkConverter.ProcessStage3(document, htmlWriter, settings);
+
+                var regex = new Regex("<a href=([\"\']).*?\\1");
```
loic-sharma: Consider making this regex also a static readonly.

loic-sharma: Also, is `CommonMarkConverter` guaranteed to always generate links with the format `<a href=`? If not, this regex may not match all links.

src/NuGetGallery/Services/ReadMeService.cs
chenriksson: Will play with this to see if I can optimize it.

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -149,10 +174,65 @@ public bool HasReadMeSource(ReadMeRequest readMeRequest)
...
+                                break;
+                        }
+                    }
+
+                    var inline = node.Inline;
+                    if (inline != null)
+                    {
+                        // Block javascript in links.
+                        if ((inline.Tag == InlineTag.Link) &&
+                            (inline.TargetUrl.IndexOf("javascript", StringComparison.InvariantCultureIgnoreCase) >= 0))
```
chenriksson: Will change to only allow target urls that use http/https.

src/NuGetGallery/Services/ReadMeService.cs
chenriksson: Yes, the current version always has href attribute first, then possible title attribute. There's an open issue to add support for other link attributes, but it won't be added until the commonmark spec includes it.

chenriksson:
@blowdart Can you have a look?

blowdart:
LGTM

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -7,22 +7,27 @@
...
 using CommonMark;
-using System.Text.RegularExpressions;
+using CommonMark.Syntax;
 
 namespace NuGetGallery
 {
     internal class ReadMeService : IReadMeService
     {
+        private static readonly Regex EncodedBlockQuotePattern = new Regex("^ {0,3}&gt;", RegexOptions.Multiline);
+        private static readonly Regex CommonMarkLinkPattern = new Regex("<a href=([\"\']).*?\\1");
```
agr: I feel uneasy about using regexes to insert the `rel=nofollow`. Are you sure CommonMark always generates links that start with `<a href=`?

chenriksson: This regex was added in PR #4841. I discussed with @ryuyu and we agreed the regex was probably better than forking CommonMark.Net.

For CommonMark.NET link formatting see [HtmlFormatterSlim](https://github.com/Knagis/CommonMark.NET/blob/e77bf5a9c6f6a432fa0fb845ccdb64d9b2427bbb/CommonMark/Formatters/HtmlFormatterSlim.cs#L560).

CommonMark.NET had a request to [support link attributes](https://github.com/Knagis/CommonMark.NET/issues/101), but it won't be added unless attribute extensions are added to the underlying commonmark spec.

chenriksson: I should add timeout for these regexes, as pointed out in other PRs.

src/NuGetGallery/Services/ReadMeService.cs
chenriksson: Changed to single MultiLine regex applied to entire markdown

PR #4854, opened 2017-10-17
[Prefix] Manage package owners with respect to Prefix reservation

### LINKED ISSUE #4380: [Prefix] Handle ownership changes with regards to prefix reservation

shishirx34:
Add support for package ownership management with regards to reserved prefixes. Fixes #4380 
1. Adding an owner owning a matching reserved namespace to a package, adds verified visual indicator to the package.
2. Added restrictions on owner removal. Now the owner cannot remove other owners owning the matching reserved namespace unless the owner themselves own the namespace as well.

I have created a new service for `PackageOwnershipManagement`, a bit of refactor for consolidating code and auditing. 

TODO:
 - [ ] Add the tests for new service and relevant refactor code tests.

Tests are to follow in the same PR, please review the code meanwhile.


src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    public interface IPackageOwnershipManagementService
```
chenriksson: Can we combine this with IPackageOwnerRequestService?

Also, the name is long. Some entity-based services follow the naming pattern "I{Entity}Service". Maybe it could be "IPackageRegistrationOwnerService", or "IPackageOwnerService"?

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -55,24 +58,94 @@ public virtual ActionResult GetPackageOwners(string id, string version)
...
+                .PackageRegistration
+                .Owners
+                .Where(o => o.Username == currentUserName)
+                .FirstOrDefault();
+
+            IEnumerable<User> removableOwnersList = null;
+            IEnumerable<User> nonRemovableOwnersList = null;
+            var packageRegistrationOwners = package.PackageRegistration.Owners;
+            var canRemoveSelf = false;
+            if (currentOwner != null)
```
chenriksson: This check is a bit redundant with the !package.IsOwner check above.

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -55,24 +58,94 @@ public virtual ActionResult GetPackageOwners(string id, string version)
...
+                    .Any(owner => owner.Username == currentOwner.Username);
+
+                var packageAndReservedNamespaceOwners = packageRegistrationOwners.Intersect(allMatchingNamespaceOwners);
+
+                // An owner can remove themselves if
+                // 1. There are more than one owners on the package 
+                // 2. And either the owner does not own any matching namespace
+                // 3. Or if the owner owns a namespace there are other owners who also own the namespace
+                canRemoveSelf = (!currentOwnerOwnsNamespace && packageRegistrationOwners.Count() > 1)
+                    || (currentOwnerOwnsNamespace && packageAndReservedNamespaceOwners.Count() > 1);
```
chenriksson: Why is this relative to self? Don't the same general rules apply regardless of whether the current user or a different user is the last package or namespace owner?

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -55,24 +58,94 @@ public virtual ActionResult GetPackageOwners(string id, string version)
...
+
+                // Build a list of removable and non-removable users based on the ownership of the matching reserved namespace
+                if (!currentOwnerOwnsNamespace)
+                {
+                    // Current owner does not own the matching namespace, so it cannot remove any
+                    // package owner who owns the matching namespace.
+                    nonRemovableOwnersList = packageRegistrationOwners
+                        .Where(packageOwner => allMatchingNamespaceOwners.Any(rno => rno == packageOwner));
```
chenriksson: All namespaces are weighted equally, or should we care about the closest match / most specific?

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -55,24 +58,94 @@ public virtual ActionResult GetPackageOwners(string id, string version)
...
+
+            if (nonRemovableOwnersList != null)
+            {
+                var nonRemovableResultViewModel = from u in nonRemovableOwnersList
+                                                  select new PackageOwnersResultViewModel(
+                                                      u.Username,
+                                                      u.EmailAddress,
+                                                      isCurrentUser: u.Username == currentUserName,
+                                                      isPending: false,
+                                                      canBeRemoved: u.Username == currentUserName ? canRemoveSelf : false);
```
chenriksson: These changes look like they can be simplified.

Currently we block removing the last owner. Can we just add a flag to indicate whether owners own namespaces, and block removing the last namespace owner as well?

chenriksson: Should complex code like this go into the package owner service?

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
```diff
@@ -175,16 +175,17 @@
...
     .error(failHandler);
 
-    function Owner(name, profileUrl, imageUrl, pending, current) {
+    function Owner(name, profileUrl, imageUrl, pending, current, canBeRemoved) {
         this.name = ko.observable(name);
         this.profileUrl = ko.observable(profileUrl);
         this.imageUrl = ko.observable(imageUrl);
         this.pending = ko.observable(pending);
         this.current = current;
+        this.canBeRemoved = ko.observable(canBeRemoved);
```
chenriksson: Per comments in JsonApiController, should we send isNamespaceOwner flag instead of canBeRemoved and just block removing the last isNamespaceOwner owner?

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
```diff
@@ -0,0 +1,182 @@
...
+                            _reservedNamespaceService.AddPackageRegistrationToNamespace(mn.Value, packageRegistration));
+
+                    await _entitiesContext.SaveChangesAsync();
+                }
+
+                await _packageService.AddPackageOwnerAsync(packageRegistration, user);
+
+                await RemovePendingOwnershipRequestAsync(packageRegistration, user);
+
+                transaction.Commit();
```
chenriksson: Does this need an explicit transaction and suspended execution strategy, or could you defer SaveChangesAsync until the end? Note, you're going to lose retry capabilities by suspending the strategy.

Same elsewhere.

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            using (var strategy = new SuspendDbExecutionStrategy())
+            using (var transaction = _entitiesContext.GetDatabase().BeginTransaction())
+            {
+                var userOwnedMatchingNamespacesForId = user
```
skofman1: there is no handling here for the prefix/not prefix case, like in ReservedNamespaceService.GetReservedNamespacesForId

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <param name="user">The user to add as an owner to the package.</param>
+        Task AddPackageOwnerAsync(PackageRegistration packageRegistration, User user);
+
+        /// <summary>
+        /// Add the pending ownership request.
+        /// </summary>
+        /// <param name="packageRegistration">The package registration that has pending ownership request.</param>
+        /// <param name="requestingOwner">The user to requesting to add the pending owner.</param>
+        /// <param name="newOwner">The user to be added for from pending ownership.</param>
+        Task<PackageOwnerRequest> AddPendingOwnershipRequestAsync(PackageRegistration packageRegistration, User requestingOwner, User newOwner);
```
cristinamanum: Would not SendOwnershipRequestAsync a better name? Pending seems to be more a property of the resulted request.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
+        /// <param name="packageRegistration">The package registration that has pending ownership request.</param>
+        /// <param name="user">The user to be removed from pending ownership.</param>
+        Task RemovePendingOwnershipRequestAsync(PackageRegistration packageRegistration, User user);
+
+        /// <summary>
+        /// Remove the pending ownership request.
+        /// </summary>
+        /// <param name="request">The package owner request to be removed.</param>
+        Task RemovePendingOwnershipRequestAsync(PackageOwnerRequest request);
```
cristinamanum: Why is this method needed (why the above one is not enough)?

src/NuGetGallery/Controllers/JsonApiController.cs
shishirx34: all are equal.

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -55,24 +58,94 @@ public virtual ActionResult GetPackageOwners(string id, string version)
...
+                var currentOwnerOwnsNamespace = allMatchingNamespaceOwners
+                    .Any(owner => owner.Username == currentOwner.Username);
+
+                var packageAndReservedNamespaceOwners = packageRegistrationOwners.Intersect(allMatchingNamespaceOwners);
+
+                // An owner can remove themselves if
+                // 1. There are more than one owners on the package 
+                // 2. And either the owner does not own any matching namespace
+                // 3. Or if the owner owns a namespace there are other owners who also own the namespace
+                canRemoveSelf = (!currentOwnerOwnsNamespace && packageRegistrationOwners.Count() > 1)
```
scottbommarito: What about shared namespaces? Shouldn't a user be able to remove the owner of a package's reserved namespace if the namespace is shared?

src/NuGetGallery/Controllers/JsonApiController.cs
shishirx34: changing this to determine removal logic on client side.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
scottbommarito: This method is identical to `IPackageOwnershipRequestService.AddPackageOwnershipRequest`. Please use that method instead.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
scottbommarito: This method is identical to `IPackageOwnershipRequestService.DeletePackageOwnershipRequest`. Please use that method instead.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <param name="packageRegistration">The package registration that is intended to get ownership.</param>
+        /// <param name="user">The user to add as an owner to the package.</param>
+        Task RemovePackageOwnerAsync(PackageRegistration packageRegistration, User user);
+
+        /// <summary>
+        /// Remove the pending ownership request.
+        /// </summary>
+        /// <param name="packageRegistration">The package registration that has pending ownership request.</param>
+        /// <param name="user">The user to be removed from pending ownership.</param>
+        Task RemovePendingOwnershipRequestAsync(PackageRegistration packageRegistration, User user);
```
scottbommarito: You should add this to `IPackageOwnershipRequestService` or use the methods in `IPackageOwnershipRequestService` instead.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
scottbommarito: I don't think we need a new service at all. The methods in this class appear to either be

- Things that `IPackageService` already does, but without handling reserved namespaces.
or
- Things that `IPackageOwnerRequestService` already does.

The `IPackageService` methods should be updated to handle reserved namespaces and the `IPackageOwnerRequestService` methods should be used instead.

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
shishirx34: That is the beauty of `StartsWith` below, it works for both cases, prefix/not prefix.

src/NuGetGallery/Controllers/JsonApiController.cs
shishirx34: no, same conditions apply for shared namespaces too.

src/NuGetGallery/Controllers/JsonApiController.cs
shishirx34: I think the result set construction should be the part of the controller. Also, I have simplified this code by moving logic to client side for removal.

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
shishirx34: Yes, addressed.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
scottbommarito: Spoke with @shishirx34 offline, this is ok because `IPackageOwnershipRequestService` is meant to be a low level service and `IPackageOwnershipManagementService` is meant to be a high level service.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
scottbommarito: Spoke with @shishirx34 offline, this is ok because `IPackageOwnershipRequestService` is meant to be a low level service and `IPackageOwnershipManagementService` is meant to be a high level service.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
scottbommarito: Spoke with @shishirx34 offline, this is ok because `IPackageOwnershipRequestService` is meant to be a low level service and `IPackageOwnershipManagementService` is meant to be a high level service.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
scottbommarito: Spoke with @shishirx34 offline, this is ok because `IPackageOwnershipRequestService` is meant to be a low level service and `IPackageOwnershipManagementService` is meant to be a high level service.

shishirx34: Like @scottbommarito said, this service is a high level service and not an "I{Entity}Service", we are moving to scenario based service, same as `PackageUploadService` created in my previous PR. This is a high level service which would use entity services like a microservice and move the business logic from controllers to here. I will look into renaming the service into something shorter.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
shishirx34: `SendOwnershipRequestAsync` sounds like it would actually send the request(like email) which is not the case, it simply adds the request as a pending approval with regards to `PackageOwnerRequestService` and hence the name.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
shishirx34: I added this as an optimization, however, I did not use it. I will remove this to keep it simple.

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
shishirx34: Some of the methods do `commit` their changes, which I don't want to happen, and that would entail refactoring these methods to not commit if we were to defer `SaveChangesAsync`. I agree this is somewhat inconsistent, but I would rather not do unnecessary changes in this PR. Also, given that this route is not a frequent execution I think its fine to have the transaction here.

src/NuGetGallery/Services/IPackageOwnershipManagementService.cs
shishirx34: Renamed this method to `AddPackageOwnershipRequestAsync` instead which is more appropriate.

shishirx34:
Addressed the feedback. @scottbommarito - removed the `PackageOwnerRequestService` completely from the controllers. 

Adding tests next.

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
skofman1: What about the case when an owner that owns namespace ABC (not a prefix, but exact match), is added to a package called ABCD? According to this logic ABCD will be marked as verified, but it shouldn't be.

shishirx34: Ah, yes you are right.

shishirx34: Fixed and added test.

shishirx34:
@skofman1 @scottbommarito @chenriksson - signoffs please?

tests/NuGetGallery.Facts/Services/PackageOwnershipManagementServiceFacts.cs
```diff
@@ -0,0 +1,417 @@
...
+            [Fact]
+            public async Task NullUserThrowsException()
+            {
+                var service = CreateService();
+                var testPackageRegistrations = ReservedNamespaceServiceTestData.GetRegistrations();
+                await Assert.ThrowsAsync<ArgumentNullException>(async () => await service.AddPackageOwnerAsync(packageRegistration: testPackageRegistrations.First(), user: null));
+            }
+
+            [Fact]
+            public async Task NewOwnerIsAddedSuccessfullyToTheRegistration()
```
scottbommarito: This test adds an owner who had an existing package ownership request. We should have another test for adding an owner without an existing package ownership request, which is supported by the code as far as I can tell.

tests/NuGetGallery.Facts/Services/PackageOwnershipManagementServiceFacts.cs
```diff
@@ -0,0 +1,417 @@
...
+            {
+                var package = new PackageRegistration { Key = 2, Id = "Microsoft.Aspnet.Package1" };
+                var pendingOwner = new User { Key = 100, Username = "aspnet" };
+                var packageService = new Mock<IPackageService>();
+                var packageOwnerRequestService = new Mock<IPackageOwnerRequestService>();
+
+                var service = CreateService(packageService: packageService, packageOwnerRequestService: packageOwnerRequestService);
+                await service.AddPackageOwnerAsync(package, pendingOwner);
+
+                packageService.Verify(x => x.AddPackageOwnerAsync(It.Is<PackageRegistration>(pr => pr == package), It.Is<User>(u => u == pendingOwner)));
```
scottbommarito: nit: I'm pretty sure you can simplify this with
```
x.AddPackageOwnerAsync(package, pendingOwner)
```

tests/NuGetGallery.Facts/Services/PackageOwnershipManagementServiceFacts.cs
```diff
@@ -0,0 +1,417 @@
...
+                var existingNamespace2 = new ReservedNamespace("microsoft.aspnet.", isSharedNamespace: false, isPrefix: true);
+                pendingOwner.ReservedNamespaces.Add(existingNamespace);
+                pendingOwner.ReservedNamespaces.Add(existingNamespace2);
+
+                var reservedNamespaceService = new Mock<IReservedNamespaceService>();
+
+                var service = CreateService(reservedNamespaceService: reservedNamespaceService);
+                await service.AddPackageOwnerAsync(package, pendingOwner);
+
+                reservedNamespaceService.Verify(x => x.AddPackageRegistrationToNamespace(It.IsAny<string>(), It.IsAny<PackageRegistration>()), Times.Between(2, 2, Range.Inclusive));
```
scottbommarito: `Times.Between(...)` -> `Times.Exactly(2)`

I think it would be better to check exactly the parameters you expect here, e.g.
```
x.AddPackageRegistrationToNamespace("microsoft.", package)
```
and
```
x.AddPackageRegistrationToNamespace("microsoft.aspnet.", package)
```

src/Bootstrap/less/theme/page-manage-owners.less
```diff
@@ -28,8 +28,21 @@
...
+        position: relative;
+        top: 3px;
+      }
+    }
+
     .remove-owner a:hover, .remove-owner a:active {
       color: darken(@brand-danger, 10%);
     }
+
+    .remove-owner a:hover, .remove-owner a:active {
```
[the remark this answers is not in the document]
shishirx34: corrected to `.remove-owner-disabled`

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -55,24 +55,43 @@ public virtual ActionResult GetPackageOwners(string id, string version)
...
+            var allMatchingNamespaceOwners = package
+                .PackageRegistration
+                .ReservedNamespaces
+                .SelectMany(rn => rn.Owners)
+                .Distinct();
+
+            var packageAndReservedNamespaceOwners = packageRegistrationOwners.Intersect(allMatchingNamespaceOwners);
+            var packageOwnersOnly = packageRegistrationOwners.Except(packageAndReservedNamespaceOwners);
+
+            var owners = from u in packageAndReservedNamespaceOwners
```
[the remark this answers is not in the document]
shishirx34: Its the same. I don't care enough, will change it.

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
```diff
@@ -0,0 +1,187 @@
...
+                transaction.Commit();
+            }
+
+            await _auditingService.SaveAuditRecordAsync(
+                new PackageRegistrationAuditRecord(packageRegistration, AuditedPackageRegistrationAction.AddOwner, user.Username));
+        }
+
+        public async Task<PackageOwnerRequest> AddPackageOwnershipRequestAsync(PackageRegistration packageRegistration, User requestingOwner, User newOwner)
+        {
+            if (packageRegistration == null)
```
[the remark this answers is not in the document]
shishirx34: It doesnt.

src/NuGetGallery/Views/Packages/ManagePackageOwners.cshtml
```diff
@@ -50,12 +50,30 @@
...
                             <span>Remove</span>
                         </a>
                     </div>
+                    <!-- /ko -->
+                    <!-- ko ifnot: pending() 
+                        || $parent.hasMoreThanOneOwner() 
+                        && (!isNamespaceOwner() || ($parent.isCurrentNamespaceOwner() && $parent.hasMoreThanOneNamespaceOwners())) -->
+                    <div class="col-md-3 remove-owner-disabled">
+                        <a class="icon-link"
+                           title="Cannot remove the only owner or reserved namespace restrictions apply.">
```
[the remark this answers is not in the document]
shishirx34: I am using the title as the tooltip, we aren't going to show Alert here, because it is not going to be actionable or invokable at all thats why I want to merge it into the single message. I would leave it as is unless the meaning is unclear.

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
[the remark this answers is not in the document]
scottbommarito: I think you should move these checks into `PackageOwnerRequestService` then, it'll throw an exception otherwise.

Issue #4770, opened 2017-09-28
Markdown documentation do not show formatting propoerly


anangaur:
Feature discussion issue: https://github.com/NuGet/NuGetGallery/issues/2280

Add the following markdown to the documentation for a package:

```
**Bold text**

*Italics*

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum
```

**Note** the line breaks.

Expected: 
1. The markdown formatting and line breaks should be rendered properly in the preview (without clicking on **Show more**)
2. If all the text is shown in preview mode (without clicking on **Show more**), then "Show more** option should **not** be shown to users.

Actual:
1. The markdown formatting shows properly but line breaks are removed
2. Even when all text is shown in preview mode, "Show more** option is shown.

PR #4858, opened 2017-10-17
Documentation clamp fixes

### LINKED ISSUE #4812: Clicking on "Show more" for Documentation has incorrect chevron

chenriksson:
Fixes #4812 and #4770 

This revives the client-side clamping from @JarahSi. I've disabled the native (-webkit-line-clamp) support due to formatting issues possibly caused by conflicts with the redesign css. Instead, it does javascript clamping which adjusts the max height.

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -281,21 +281,19 @@
                     </h2>
                 </div>
                 <div id="readme-container" class="collapse in">
-                    <div id="readme-clamped" class="readme collapse in">
-                        @Html.Raw(Model.ReadMeHtmlClamped)
+                    <div id="readme-less" class="collapse in">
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -281,21 +281,19 @@
                     </h2>
                 </div>
                 <div id="readme-container" class="collapse in">
-                    <div id="readme-clamped" class="readme collapse in">
-                        @Html.Raw(Model.ReadMeHtmlClamped)
+                    <div id="readme-less" class="collapse in">
+                        @Html.Raw(Model.ReadMeHtml)
```
scottbommarito: I would leave a comment here stating that this id gets "clamped", because otherwise it's a little funny that you have the same content twice in a row.

PR #4860, opened 2017-10-18
Log product information on push+verify events

skofman1:
Extracts client information from user agent and logs on push/verify id events.
https://github.com/NuGet/Engineering/issues/797

### DIFF
--- a/src/NuGetGallery/Helpers/HttpContextBaseExtensions.cs
+++ b/src/NuGetGallery/Helpers/HttpContextBaseExtensions.cs
@@ -59,7 +59,7 @@ namespace NuGetGallery
         /// User-Agent: Mozilla/<version> (<system-information>) <platform> (<platform-details>) <extensions>
         /// Thus, extracting the part before the first '(', should give us product and version tokens in MOST cases.
         /// </summary>
-        public static string GetProductInformation(this HttpContextBase httpContext)
+        public static string GetClientInformation(this HttpContextBase httpContext)
         {
             string userAgent = httpContext.Request.Headers[Constants.UserAgentHeaderName];
             string result = string.Empty;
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Helpers/HttpContextBaseExtensions.cs
```diff
@@ -51,6 +51,37 @@ public static string GetConfirmationReturnUrl(this HttpContextBase httpContext)
...
+                else
+                {
+                    result = userAgent;
+                }
+
+                result = result.Trim();
+            }
+
+            return result;
+        }
```
cristinamanum: Why we not return the user agent value as it is?

loic-sharma: This heuristic may lose valuable information and it'd be easy to recreate this value in Application Insight's Analytics. I would favor ad-hoc parsing as well.

skofman1: This is not new code.. it was copied from ClientInformationTelemetryEnricher, and moved here to allow sharing.
User-Agent header contains not only the client name and version (which is the info we actually need), but also platform information, which we don't care about. The reason to split and not log the entire value is to reduce size, since this data is logged for every request. The approach has proved itself over the course of several months, and you can see this data in AI.

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -31,6 +31,7 @@ public class TelemetryService : ITelemetryService
         public const string AccountCreationDate = "AccountCreationDate";
         public const string ClientVersion = "ClientVersion";
         public const string ProtocolVersion = "ProtocolVersion";
+        public const string ProductInformation = "ProductInformation";
```
chenriksson: Should this key be "ClientInfo" like before? "ProductInformation" seems confusing when taken outside the context of a user-agent string.

skofman1: fixed

skofman1:
🔔

Issue #4653, opened 2017-09-11
Admin for Validation tables


joelverhagen:
This is a read-only UI. Revalidation gesture is tracked elsewhere.

This is likely a new admin UI page.

- Needs search by package ID + version, for drill down
- Shows all validation sets and validation records

PR #4861, opened 2017-10-18
Add UI for validation tables (just orchestrator tables)

joelverhagen:
https://github.com/NuGet/NuGetGallery/issues/4653

## Admin page (notice last link)

screencapture-localhost-admin-1508288815230

## Initial page

screencapture-localhost-admin-validation-search-1508288756724

## Search by ID

screencapture-localhost-admin-validation-search-1508288660270

## Search by ID + version

screencapture-localhost-admin-validation-search-1508288671683

## Search by validation set tracking ID

screencapture-localhost-admin-validation-search-1508288685885

## Search by validation ID

screencapture-localhost-admin-validation-search-1508288704997

### DIFF
--- a/src/NuGetGallery.Core/Entities/IEntityRepository.cs
+++ b/src/NuGetGallery.Core/Entities/IEntityRepository.cs
@@ -7,12 +7,11 @@ using System.Threading.Tasks;
 namespace NuGetGallery
 {
     public interface IEntityRepository<T>
-        where T : class, IEntity, new()
+        where T : class, new()
     {
         Task CommitChangesAsync();
         void DeleteOnCommit(T entity);
-        T GetEntity(int key);
         IQueryable<T> GetAll();
-        int InsertOnCommit(T entity);
+        void InsertOnCommit(T entity);
     }
 }
--- a/src/NuGetGallery/Areas/Admin/Controllers/HomeController.cs
+++ b/src/NuGetGallery/Areas/Admin/Controllers/HomeController.cs
@@ -1,24 +1,32 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Web.Mvc;
+using NuGetGallery.Areas.Admin.ViewModels;
+using NuGetGallery.Configuration;
 
 namespace NuGetGallery.Areas.Admin.Controllers
 {
     public partial class HomeController : AdminControllerBase
     {
         private readonly IContentService _content;
+        private readonly IGalleryConfigurationService _config;
 
-        public HomeController(IContentService content)
+        public HomeController(IContentService content, IGalleryConfigurationService config)
         {
-            _content = content;
+            _content = content ?? throw new ArgumentNullException(nameof(content));
+            _config = config ?? throw new ArgumentNullException(nameof(config));
         }
 
         [HttpGet]
         public virtual ActionResult Index()
         {
-            return View();
+            var viewModel = new HomeViewModel(
+                showValidation: _config.Current.AsynchronousPackageValidationEnabled);
+
+            return View(viewModel);
         }
 
         [HttpGet]
--- a/src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
+++ b/src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
@@ -1,0 +1,63 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Web.Mvc;
+using NuGet.Versioning;
+using NuGetGallery.Areas.Admin.Services;
+using NuGetGallery.Areas.Admin.ViewModels;
+
+namespace NuGetGallery.Areas.Admin.Controllers
+{
+    public class ValidationController : AdminControllerBase
+    {
+        private readonly ValidationAdminService _validationAdminService;
+
+        public ValidationController(ValidationAdminService validationAdminService)
+        {
+            _validationAdminService = validationAdminService ?? throw new ArgumentNullException(nameof(validationAdminService));
+        }
+
+        [HttpGet]
+        public virtual ActionResult Index()
+        {
+            return View(nameof(Index), new ValidationPageViewModel());
+        }
+
+        [HttpGet]
+        public virtual ActionResult Search(string q)
+        {
+            var packageValidationSets = _validationAdminService.Search(q ?? string.Empty);
+
+            // Sort by lexigraphically package ID then put newer stuff on top.
+            var groups = packageValidationSets
+                .OrderBy(x => x.PackageId)
+                .ThenByDescending(x => NuGetVersion.Parse(x.PackageNormalizedVersion))
+                .ThenByDescending(x => x.PackageKey)
+                .ThenByDescending(x => x.Created)
+                .ThenByDescending(x => x.Key)
+                .GroupBy(x => x.PackageKey);
+
+            var validatedPackages = new List<ValidatedPackageViewModel>();
+            foreach (var group in groups)
+            {
+                foreach (var set in group)
+                {
+                    // Put completed validations first then put new validations on top.
+                    set.PackageValidations = set.PackageValidations
+                        .OrderByDescending(x => x.ValidationStatus)
+                        .ThenByDescending(x => x.ValidationStatusTimestamp)
+                        .ToList();
+                }
+
+                var deletedStatus = _validationAdminService.GetPackageDeletedStatus(group.Key);
+                var validatedPackage = new ValidatedPackageViewModel(group.ToList(), deletedStatus);
+                validatedPackages.Add(validatedPackage);
+            }
+
+            return View(nameof(Index), new ValidationPageViewModel(q, validatedPackages));
+        }
+    }
+}
--- a/src/NuGetGallery/Areas/Admin/Models/DeletedStatus.cs
+++ b/src/NuGetGallery/Areas/Admin/Models/DeletedStatus.cs
@@ -1,12 +1,12 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Web.Mvc;
-
-namespace NuGetGallery.Areas.Admin.Controllers
+namespace NuGetGallery.Areas.Admin.Models
 {
-    [Authorize(Roles="Admins")]
-    public class AdminControllerBase : AppController
+    public enum PackageDeletedStatus
     {
+        SoftDeleted,
+        NotDeleted,
+        Unknown,
     }
 }
--- a/src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
+++ b/src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
@@ -1,0 +1,199 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Data.Entity;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using NuGet.Services.Validation;
+using NuGet.Versioning;
+using NuGetGallery.Areas.Admin.Models;
+
+namespace NuGetGallery.Areas.Admin.Services
+{
+    public class ValidationAdminService
+    {
+        private readonly IEntityRepository<PackageValidationSet> _validationSets;
+        private readonly IEntityRepository<PackageValidation> _validations;
+        private readonly IEntityRepository<Package> _packages;
+
+        public ValidationAdminService(
+            IEntityRepository<PackageValidationSet> validationSets,
+            IEntityRepository<PackageValidation> validations,
+            IEntityRepository<Package> packages)
+        {
+            _validationSets = validationSets ?? throw new ArgumentNullException(nameof(validationSets));
+            _validations = validations ?? throw new ArgumentNullException(nameof(validations));
+            _packages = packages ?? throw new ArgumentNullException(nameof(packages));
+        }
+
+        /// <summary>
+        /// Fetch a list of package validation sets matching the provided query. The query is a line seperated list of
+        /// identifiers. These identifiers can refer to packages, sets, or validations. An empty list is returned if no
+        /// sets are matched.
+        /// </summary>
+        public IReadOnlyList<PackageValidationSet> Search(string query)
+        {
+            var lines = ParseQueryToLines(query);
+
+            // Search for matching validation sets using various methods of parsing the lines.
+            var validationSets = new Dictionary<long, PackageValidationSet>();
+            foreach (var line in lines)
+            {
+                SearchByValidationSetTrackingId(validationSets, line);
+                SearchByValidationId(validationSets, line);
+                SearchByValidationSetKey(validationSets, line);
+                SearchByPackageIdAndVersion(validationSets, line);
+                SearchByPackageId(validationSets, line);
+            }
+
+            return validationSets
+                .Values
+                .ToList();
+        }
+
+        /// <summary>
+        /// Determines if deleted status of the provided package key. This method is unable to differentiate between
+        /// a hard deleted package and a package that never existed in the first place. Therefore,
+        /// <see cref="PackageDeletedStatus.Unknown"/> is returned if the package key is not found.
+        /// </summary>
+        public PackageDeletedStatus GetPackageDeletedStatus(int packageKey)
+        {
+            var package = _packages
+                .GetAll()
+                .Where(x => x.Key == packageKey)
+                .FirstOrDefault();
+
+            if (package == null)
+            {
+                return PackageDeletedStatus.Unknown;
+            }
+            else if (package.PackageStatusKey == PackageStatus.Deleted)
+            {
+                return PackageDeletedStatus.SoftDeleted;
+            }
+
+            return PackageDeletedStatus.NotDeleted;
+        }
+
+        internal static IReadOnlyList<string> ParseQueryToLines(string query)
+        {
+            // Collapse redundant spaces.
+            var normalizedQuery = Regex.Replace(
+                query,
+                @"[^\S\r\n]+",
+                " ",
+                RegexOptions.None,
+                TimeSpan.FromSeconds(10));
+
+            // Split lines and trim.
+            var lines = new List<string>();
+            var uniqueLines = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            using (var stringReader = new StringReader(normalizedQuery))
+            {
+                string line;
+                while ((line = stringReader.ReadLine()) != null)
+                {
+                    var trimmedLine = line.Trim();
+                    if (trimmedLine.Length > 0 && uniqueLines.Add(trimmedLine))
+                    {
+                        lines.Add(trimmedLine);
+                    }
+                }
+            }
+
+            return lines;
+        }
+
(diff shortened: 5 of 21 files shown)


src/NuGetGallery.Core/Entities/IEntityRepository.cs
```diff
@@ -7,12 +7,11 @@
 namespace NuGetGallery
 {
     public interface IEntityRepository<T>
-        where T : class, IEntity, new()
+        where T : class, new()
```
joelverhagen: Simplified this interface to enable use on the `ValidationEntitiesContext`.

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -510,11 +512,19 @@ public static string UploadPackageProgress(this UrlHelper url, bool relativeUrl
...
+                    { "version", version }
                 });
         }
 
+        public static string RevalidatePackage(
+            this UrlHelper url,
+            IPackageVersionModel package,
+            bool relativeUrl = true)
+        {
+            return url.RevalidatePackage(package.Id, package.Version);
```
loic-sharma: >RevalidatePackage [](start = 23, length = 17)

Pass the `relativeUrl` down.

src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        [HttpGet]
+        public virtual ActionResult Index()
+        {
+            return View(nameof(Index), new ValidationPageViewModel());
+        }
+
+        [HttpGet]
+        public virtual async Task<ActionResult> Search(string vq)
```
loic-sharma: >vq [](start = 62, length = 2)

Nit: `vq` is a strange name, could we use something like `query` instead?

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
```diff
@@ -0,0 +1,176 @@
...
+            foreach (var line in lines)
+            {
+                await SearchByValidationSetTrackingIdAsync(validationSets, line);
+                await SearchByValidationIdAsync(validationSets, line);
+                await SearchByValidationSetKeyAsync(validationSets, line);
+                await SearchByPackageIdAndVersionAsync(validationSets, line);
+                await SearchByPackageIdAsync(validationSets, line);
+            }
+
+            var outputValidationSets = validationSets
```
loic-sharma: >outputValidationSets [](start = 16, length = 20)

Nit: return this directly instead of creating a variable?

src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
joelverhagen: `v`alidation `q`uery. I'll give the parameter a different name but I wanted to keep the query string short.

src/NuGetGallery/UrlExtensions.cs
joelverhagen: Nice catch

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            // Split lines and trim whitespace.
+            var lines = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            using (var stringReader = new StringReader(query))
+            {
+                string line;
+                while ((line = stringReader.ReadLine()) != null)
+                {
+                    var normalizedLine = Regex.Replace(line.Trim(), @"\s+", " ");
+                    lines.Add(line);
```
loic-sharma: >line [](start = 30, length = 4)

Add the `normalizedLine`, not the `line`. Also, `skip `normalizedLine`s that are empty.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
```diff
@@ -0,0 +1,176 @@
...
+        public async Task<IReadOnlyList<PackageValidationSet>> SearchAsync(string query)
+        {
+            // Split lines and trim whitespace.
+            var lines = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            using (var stringReader = new StringReader(query))
+            {
+                string line;
+                while ((line = stringReader.ReadLine()) != null)
+                {
+                    var normalizedLine = Regex.Replace(line.Trim(), @"\s+", " ");
```
loic-sharma: >Replace [](start = 47, length = 7)

Could you run the `Regex.Replace` once on the entire query instead?

src/NuGetGallery/Areas/Admin/Views/Validation/Index.cshtml
```diff
@@ -0,0 +1,146 @@
...
+            <li>Package ID and version: <code>NuGet.Versioning 4.3.0</code></li>
+            <li>Validation set tracking ID: <code>e6db6737-bd77-44cd-af8b-74826e054d36</code></li>
+            <li>Validation ID: <code>218d075b-99b6-4151-ba12-121872ffe7fe</code></li>
+        </ul>
+    }
+
+    @{ bool isFirst = true; }
+    @foreach (var package in Model.Packages)
+    {
+        if (!isFirst)
```
loic-sharma: >isFirst [](start = 13, length = 7)

Should this be `if (isFirst)`?

src/NuGetGallery/Areas/Admin/Views/Validation/Index.cshtml
```diff
@@ -0,0 +1,146 @@
...
+                                    <th>Key</th>
+                                    <th>Type</th>
+                                    <th>Started</th>
+                                    <th>Status</th>
+                                    <th>Status Timestamp</th>
+                                </tr>
+                            </thead>
+                            <tbody>
+                                @foreach (var validation in set.PackageValidations)
+                        {
```
loic-sharma: >{ [](start = 24, length = 1)

Curly brace is off here.

src/NuGetGallery/Areas/Admin/Controllers/HomeController.cs
```diff
@@ -1,24 +1,32 @@
...
+            _config = config ?? throw new ArgumentNullException(nameof(config));
         }
 
         [HttpGet]
         public virtual ActionResult Index()
         {
-            return View();
+            var viewModel = new HomeViewModel(
+                showValidation: _config.Current.AsynchronousPackageValidationEnabled);
+
```
cristinamanum: Will this ever be false?

src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
loic-sharma: I would recommend `q` instead of `vq` as an abbreviation for the query string :)

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
```diff
@@ -0,0 +1,176 @@
...
+            IEntityRepository<PackageValidationSet> validationSets,
+            IEntityRepository<PackageValidation> validations,
+            IEntityRepository<Package> packages)
+        {
+            _validationSets = validationSets ?? throw new ArgumentNullException(nameof(validationSets));
+            _validations = validations ?? throw new ArgumentNullException(nameof(validations));
+            _packages = packages ?? throw new ArgumentNullException(nameof(packages));
+        }
+
+        public async Task<IReadOnlyList<PackageValidationSet>> SearchAsync(string query)
```
cristinamanum: It would be great to have comments at the method level.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
```diff
@@ -0,0 +1,176 @@
...
+                    .FirstOrDefaultAsync();
+
+                if (validation != null)
+                {
+                    validationSets[validation.PackageValidationSet.Key] = validation.PackageValidationSet;
+                }
+            }
+        }
+
+        private async Task SearchByValidationSetKeyAsync(Dictionary<long, PackageValidationSet> validationSets, string line)
```
cristinamanum: line argument is more like a key. Can it be renamed to key?

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            long integer;
+            if (long.TryParse(line, out integer))
+            {
+                var validationSet = await _validationSets
+                    .GetAll()
+                    .Include(x => x.PackageValidations)
+                    .Where(x => x.Key == integer)
+                    .FirstOrDefaultAsync();
+
```
cristinamanum: What is the state if the collection has more than 1 element? Should this case be asserted? (same for the other similar methods)

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
```diff
@@ -0,0 +1,176 @@
...
+                return DeletedStatus.SoftDeleted;
+            }
+
+            return DeletedStatus.NotDeleted;
+        }
+
+        private async Task SearchByValidationSetTrackingIdAsync(Dictionary<long, PackageValidationSet> validationSets, string line)
+        {
+            Guid guid;
+            if (Guid.TryParse(line, out guid))
```
chenriksson: nit: the SearchByX methods are very similar; I wonder if you could refactor for both code reuse and slight optimization.

For instance, if the caller does TryParse it would know whether to call specific SearchByX methods. Also, then you may be able to combine the core _validationSets SearchBy query into a single method that takes a filter.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            // Search for matching validation sets using various methods of parsing the lines.
+            var validationSets = new Dictionary<long, PackageValidationSet>();
+            foreach (var line in lines)
+            {
+                await SearchByValidationSetTrackingIdAsync(validationSets, line);
+                await SearchByValidationIdAsync(validationSets, line);
+                await SearchByValidationSetKeyAsync(validationSets, line);
+                await SearchByPackageIdAndVersionAsync(validationSets, line);
+                await SearchByPackageIdAsync(validationSets, line);
```
cristinamanum: Does the order of invocation for these methods matter?  Could be cases when they will override dictionary data for one call to the other?

src/NuGetGallery/Web.config
```diff
@@ -172,6 +172,7 @@
   <connectionStrings>
     <add name="Gallery.SqlServer" connectionString="Data Source=(localdb)\mssqllocaldb; Initial Catalog=NuGetGallery; Integrated Security=True; MultipleActiveResultSets=True" providerName="System.Data.SqlClient" />
     <add name="Gallery.SupportRequestSqlServer" connectionString="Data Source=(localdb)\mssqllocaldb; Initial Catalog=SupportRequest; Integrated Security=True; MultipleActiveResultSets=True" providerName="System.Data.SqlClient" />
+    <add name="Validation.SqlServer" connectionString="Data Source=(localdb)\mssqllocaldb; Initial Catalog=Validation; Integrated Security=True; MultipleActiveResultSets=True" providerName="System.Data.SqlClient" />
```
skofman1: what would be the F5 experience of this? Will someone running the gallery on a clean PC have issue? We don't run migrations for Validation DB as part of the setup-devenvironment.ps1.

src/NuGetGallery/Areas/Admin/Models/DeletedStatus.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Areas.Admin.Models
+{
+    public enum DeletedStatus
```
skofman1: PackageDeletedStatus for clearity

src/NuGetGallery/Areas/Admin/Services/ValidationEntityRepository.cs
```diff
@@ -0,0 +1,40 @@
...
+using NuGet.Services.Validation;
+
+namespace NuGetGallery.Areas.Admin.Services
+{
+    public class ValidationEntityRepository<T> : IEntityRepository<T>
+        where T : class, new()
+    {
+        private readonly ValidationEntitiesContext _entities;
+
+        public ValidationEntityRepository(ValidationEntitiesContext entities)
```
chenriksson: Could EntityRepository<T> be made more generic so that it could be reused for Validation? Maybe introduce an IEntitiesContextBase interface for the APIs shared across all EntityContexts:

- SaveChangesAsync
- Set
- DeleteOnCommit
- SetCommandTimeout
- GetDatabase

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
skofman1: can u add a timeout to the regex. Barry will like it :)

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
skofman1: nit: u can define the guid inline

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
```diff
@@ -0,0 +1,176 @@
...
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using NuGet.Services.Validation;
+using NuGet.Versioning;
+using NuGetGallery.Areas.Admin.Models;
+
+namespace NuGetGallery.Areas.Admin.Services
+{
+    public class ValidationAdminService
```
skofman1: didnt see UTs for this code

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
```diff
@@ -105,5 +106,19 @@
                 Manage Namespace Reservation
             </p>
         </li>
+        @if (Model.ShowValidation)
```
skofman1: I like this approach. We show items in this menu that are not relevant for certain configs, and we shouldn't.

src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
joelverhagen: I was thinking this would collide with filling in the search box at the top of the page, but this behavior only exists on the package search page. Will fix.

src/NuGetGallery/Areas/Admin/Controllers/HomeController.cs
joelverhagen: The default `git clone` case is this (and NuGet.org, today).

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Sure

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Line is not the key in all cases. It is just something that will be parsed for a DB query. I would rather keep the parameter name consistent.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: This is not possible since `Key` is the primary key of the `PackageValidationSets` table.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Order does not matter since the dictionary contains the whole package validation set.

src/NuGetGallery/Web.config
joelverhagen: F5 will not see this validation UI since asynchronous validation is disabled. The setup script should not create this database since it's not the default scenario (IMO).

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Good catch.

src/NuGetGallery/Areas/Admin/Services/ValidationEntityRepository.cs
joelverhagen: I don't think an new base interface would help here. I will try to have a shared base class (implementation, not interface).

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Spoke extensively offline about this. I think they are not necessary in this case since the code is pretty simple but I would rather set the pattern here as there is no rush.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: I chose not to perform these optimizations until we experience some painful performance. As the code is today, I think it's more readable.

src/NuGetGallery/Areas/Admin/Services/ValidationEntityRepository.cs
joelverhagen: These cannot share a base class since there is not common interface between `IEntitiesContext` and `ValidationEntitiesContext`.

src/NuGetGallery/UrlExtensions.cs

src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
joelverhagen: Fixed.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Fixed.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Fixed.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Fixed.

src/NuGetGallery/Areas/Admin/Views/Validation/Index.cshtml
joelverhagen: Removed entirely.

src/NuGetGallery/Areas/Admin/Views/Validation/Index.cshtml
joelverhagen: Fixed.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Fixed.

src/NuGetGallery/Areas/Admin/Models/DeletedStatus.cs
joelverhagen: Fixed.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Fixed.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Fixed.

src/NuGetGallery/Areas/Admin/Services/ValidationAdminService.cs
joelverhagen: Fixed.

PR #4877, opened 2017-10-20
Fix: no need to show both verify and edit warnings on edit package page.

xavierdecoster:
Moved the warning out of the form to avoid showing both on regular edit page.

### DIFF
--- a/src/NuGetGallery/Views/Packages/_EditForm.cshtml
+++ b/src/NuGetGallery/Views/Packages/_EditForm.cshtml
@@ -11,8 +11,7 @@
         </h2>
     </div>
 
-    <div id="verify-package-container" class="hidden">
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Views/Packages/_EditForm.cshtml
```diff
@@ -11,8 +11,7 @@
         </h2>
     </div>
 
-    <div id="verify-package-container" class="hidden">
-        @ViewHelpers.AlertPackageVerifyRecommendation()
+    <div id="verify-package-container">
```
loic-sharma: What's the purpose of this div now? Can we get rid of it?

xavierdecoster: Nope, still needed, it's the place holder for the content that gets injected through JavaScript.

PR #4882, opened 2017-10-23
Fix NullReferenceException on GET Users/Packages

xavierdecoster:
The EF queries for retrieving `PackageOwnerRequest` did not include the `PackageRegistration` relationship, which the view models now seem to rely on.

cc @scottbommarito

### DIFF
--- a/src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
+++ b/src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
@@ -1,23 +1,18 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/ViewModels/OwnerRequestsListItemViewModel.cs
```diff
@@ -1,23 +1,18 @@
...
     public class OwnerRequestsListItemViewModel
     {
-        public PackageOwnerRequest Request { get; }
-
-        public Package Package { get; }
-        
         public OwnerRequestsListItemViewModel(PackageOwnerRequest request, IPackageService packageService)
         {
             Request = request;
             Package = packageService.FindPackageByIdAndVersion(request.PackageRegistration.Id, version: null, semVerLevelKey: SemVerLevelKey.SemVer2, allowPrerelease: true);
```
xavierdecoster: This line was throwing a `NullReferenceException`.

Issue #4807, opened 2017-10-05
Hard delete of package doesn't delete package owner requests


chenriksson:
Repo:
- Create package
- Add new owner to the package
- Hard delete the package through the Admin panel
- Go to Account->Manage Packages

You'll see 500 server error due to NullReferenceException in OwnerRequestsListItemViewModel. The cause is that the PackageOwnerRequest still exists in the DB, but the PackageRegistration was deleted.

chenriksson:
It looks like hard deletes also don't delete from the PackageRegistrationOwners table, but this doesn't block any UI or cause exceptions.

scottbommarito:
Hard deletes should delete from the `PackageRegistrationOwners` table according to this

https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Services/PackageDeleteService.cs#L25

scottbommarito:
I checked locally and it does in fact delete from the `PackageRegistrationOwners` table--I think if it didn't, the commit would fail.

chenriksson:
Closing - should be fixed with #4898

PR #4898, opened 2017-10-24
Hard delete of package should delete package owner requests

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/4807


loic-sharma:
Why don't we have a one-to-many relationship between a package registration and its package owner requests?

scottbommarito:
@loic-sharma We should. This is just a quick fix for that problem (and also the problem that `PackageRegistrationOwners` doesn't have the relationship either).

PR #4901, opened 2017-10-25
[Prefix] List reserved namespaces in the manage packages page

shishirx34:
The view looks like this:

image


src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
```diff
@@ -0,0 +1,63 @@
+﻿@model ReservedNamespaceViewModel
+@{
+    if (!Model.ReservedNamespaces.Any())
+    {
+        return;
+    }
+}
+
+<div class="row user-package-list">
+    <div class="col-md-12">
```
joelverhagen: `col-xs-12`

src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
```diff
@@ -0,0 +1,63 @@
...
+        <h2>
+            <a href="#" role="button" data-toggle="collapse" data-target="#reservednamespaces"
+               aria-expanded="true" aria-controls="reservednamespaces" id="show-reservednamespaces">
+                <i class="ms-Icon ms-Icon--ChevronDown" aria-hidden="true"></i>
+                <span>My Reserved Package Ids</span>
```
joelverhagen: `IDs` not `Ids`

src/NuGetGallery/ViewModels/ReservedNamespaceViewModel.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace NuGetGallery
+{
+    public class ReservedNamespaceViewModel
+    {
+        public IEnumerable<ReservedNamespaceListItemViewModel> ReservedNamespaces { get; }
+
+        public ReservedNamespaceViewModel(ICollection<ReservedNamespace> reservedNamespacesList)
+        {
+            ReservedNamespaces = reservedNamespacesList
+                .ToList()
```
loic-sharma: The `ToList` should be after the `Select`, right?

src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
```diff
@@ -0,0 +1,63 @@
...
+                                <td class="align-middle">
+                                    <a target="_blank" href="@Url.Search(reservedNamespace.Value)">@reservedNamespace.GetPattern()</a>
+                                </td>
+                                <td class="align-middle">
+                                    @foreach (var owner in reservedNamespace.Owners)
+                                    {
+                                        <a target="_blank" href="@Url.User(owner.Username)">@owner.Username</a>
+                                    }
+                                </td>
+                                @if(reservedNamespace.IsPublic)
```
loic-sharma: Add a space between `@if` and `(`

src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
shishirx34: looks like `col-md-12` behaves correctly across different dimensions. Will leave as is.

src/NuGetGallery/ViewModels/ReservedNamespaceViewModel.cs
shishirx34: Don't need it actually, leftover from previous change. removed.

src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
shishirx34: formatted the document.

src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
scottbommarito: IMO I'd prefer
```
My Reserved Packages
```
or even
```
My Reserved Package Namespaces
```

I don't think we explicitly say `IDs` prominently in the UI anywhere else.

src/NuGetGallery/ViewModels/ReservedNamespaceViewModel.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery
+{
+    public class ReservedNamespaceViewModel
```
scottbommarito: To follow the pattern of all of our other list view models this should be `ReservedNamespaceListViewModel`

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -20,6 +20,10 @@
     }
 
     @{
+        @Html.Partial("_ReservedNamespacesList", Model.ReservedNamespaces)
```
scottbommarito: nit: No need to wrap this in `@{ ... }`

src/NuGetGallery/ViewModels/ReservedNamespaceListItemViewModel.cs
```diff
@@ -0,0 +1,38 @@
...
+        public bool IsPrefix { get; }
+
+        public IEnumerable<User> Owners { get; }
+
+        public ReservedNamespaceListItemViewModel(ReservedNamespace reservedNamespace)
+        {
+            Value = reservedNamespace.Value;
+            IsPublic = reservedNamespace.IsSharedNamespace;
+            IsPrefix = reservedNamespace.IsPrefix;
+            Owners = reservedNamespace.Owners.Select(owner => owner);
```
scottbommarito: Why do you need this `Select`?

src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
```diff
@@ -0,0 +1,63 @@
+﻿@model ReservedNamespaceViewModel
+@{
+    if (!Model.ReservedNamespaces.Any())
+    {
+        return;
+    }
+}
+
+<div class="row user-package-list">
```
scottbommarito: Given that this is the second clone of `_UserPackagesList.cshtml` that we've implemented, can we create some sort of ViewHelpers function that can create a list like this? Every time one page changes, we have to change every other page.

src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
```diff
@@ -0,0 +1,63 @@
...
+                            <th>Owners</th>
+                            <th>Upload Restrictions</th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        @foreach (var reservedNamespace in @Model.ReservedNamespaces)
+                        {
+                            <tr class="manage-package-listing">
+                                <td class="align-middle hidden-xs">
+                                    <img class="reserved-indicator-icon img-responsive"
```
scottbommarito: nit: Do we have this same chunk of code everywhere we embed the reserved package icon? Is there some way we could put it in a shared place?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -161,11 +161,13 @@ public virtual ActionResult Packages()
             var outgoing = _packageOwnerRequestService.GetPackageOwnershipRequests(requestingOwner: user);
 
             var ownerRequests = new OwnerRequestsViewModel(incoming, outgoing, user, _packageService);
+            var reservedPrefixes = new ReservedNamespaceViewModel(user.ReservedNamespaces);
```
scottbommarito: Are we concerned we're adding too many lists to `ManagePackages`? It has the user's packages, their owner requests, and now their reserved namespaces. It's ok now, but it just seems like we're dumping tons of content there.

shishirx34: There isn't a better place for this, I don't think this warrants another page either. Also, given that this  would show up only if you have a reserved namespace and same goes with owner requests I think its fine.

src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
shishirx34: This is as per @diverdan92 's spec. I think IDs clearly articulates the message here.

src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
shishirx34: I don't think we should refactor the view elements much, this isn't exactly the same piece of code, the fallback urls are different, its easier to change the view elements that are aggregated together.

src/NuGetGallery/ViewModels/ReservedNamespaceListItemViewModel.cs
shishirx34: don't need it, reminiscent of some view model changes I was making. removed it.

src/NuGetGallery/Views/Users/_ReservedNamespacesList.cshtml
shishirx34: Perhaps we can, but I wouldn't want to make it for this PR given that its a much bigger change for refactoring and affecting other components for thorough testing.

Issue #4788, opened 2017-10-02
Get rid of obsolete notifications on package details view


xavierdecoster:
To reduce the number of potential notification messages shown on the package details page, two notification messages will be removed.

* pending edits message
* v4.1.0 min client requirement for push message

Those are the 2 top notifications on the below screenshot:

![screenshot](https://user-images.githubusercontent.com/880728/31988627-0c03fd76-b970-11e7-9f1f-33ba8cc9ad3d.png)

xavierdecoster:
@anangaur there's also a "Last-Edit-Failed" notification still. I guess that should also be removed?

> Your last edit to this package was not successfully applied.  You can edit the package and submit the edit again to retry the edit.

anangaur:
I am assuming when the edit fails users would would know that immediately? If not, then I am assuming this shows when the job to edit fails and in that case we should show the notification unless we send a mail to the user for this case.

xavierdecoster:
Didn't find any email notification sending code in the `HandlePackageEdits` job, so my guess is that this notification in the UI is currently the only indicator of edit failure.

As the job is to be retired anyway, my vote is to keep the notification in the gallery around until all pending edits have been processed once we disabled edits. A subsequent deployment can then retire the gallery UI notification.

anangaur:
@xavierdecoster Does this notification show while the edits are in progress (through Jobs)? Or only when its processed and failed for some reason? I thought its the later.

xavierdecoster:
The `Pending-Edits` notification was shown while the edits were in progress. This one is removed by this PR.
The `Edit-Failed` notification was shown after failure, and is still there (not removed as part of this PR). Reason: when deploying this, notifications may still be pending and fail (rare situation). The `Edit-Failed` notification can then be removed in subsequent deployment (once Edit feature has been retired completely).

PR #4902, opened 2017-10-25
Get rid of obsolete notifications on package details view

xavierdecoster:
Fixes https://github.com/NuGet/NuGetGallery/issues/4788 (see issue for details)

### DIFF
--- a/src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
@@ -11,7 +11,6 @@ namespace NuGetGallery.Auditing
         List,
         Unlist,
         Edit,
-        UndoEdit,
         Verify
     }
 }
--- a/src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+
 namespace NuGetGallery.Auditing
 {
     public enum AuditedPackageAction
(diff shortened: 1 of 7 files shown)


joelverhagen:
I see your comment to @anangaur, but it seems weird to leave the "edit failed" message if we are removing the "pending edit". I think we should remove both at the same time, but your call!

src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
```diff
@@ -11,7 +11,6 @@ public enum AuditedPackageAction
         List,
         Unlist,
         Edit,
-        UndoEdit,
```
joelverhagen: Consider marking this as `[Obselete]` instead of removing it.

xavierdecoster: That's an option too. Is this used elsewhere? And if it is, shouldn't we break upon upgrading the consumed `NuGetGallery.Core` dependency to bring it under our attention (not sure an Obsolete warning message is intrusive enough :)).

xavierdecoster:
@joelverhagen the thing is that edits may still be pending at the time of deployment, so we may need to remove this particular Edit-Failed notification in a second deployment?

src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
joelverhagen: I am more thinking that since it's a serialized enum, it will allow us to deserialize the audit later.

joelverhagen:
> @joelverhagen the thing is that edits may still be pending at the time of deployment, so we may need to remove this particular Edit-Failed notification in a second deployment?

Seems fine. I think failed edits are a very rare occurrence.

src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
xavierdecoster: Gotcha. Marked as obsolete in https://github.com/NuGet/NuGetGallery/pull/4902/commits/26074f43622329c02b044c28a899594171dfba8d

Issue #4870, opened 2017-10-19
Add evaluation of API key Organization scopes


chenriksson:
Includes the following work:
- Add claim for Organization scope
- Evaluate Organization scope from:
  - ApiController.CreatePackageInternal
  - ApiController.VerifyPackageKeyAsync
- Propagate Organization scope to temp keys at ApiController.CreatePackageVerificationKey
- Add auditing for Organization scope

chenriksson:
Closing, as we will verify organizations work with bug bash once we onboard organization.

Issue #4868, opened 2017-10-19
Organizations schema changes


chenriksson:
Updates to Gallery DB schema to support Organizations

chenriksson:
Closing... will be verified with next deployment when we manually test single organization account.

PR #4904, opened 2017-10-25
Schema changes for Organization membership

chenriksson:
Core schema changes required for membership in organizations. Tracked by #4868 

The following were excluded, but can be added in a future migration once the design is finalized:
- API key organization scopes; will be added as part of authentication work (#4870)
- Tracking organization associated with push

/cc @anangaur


src/NuGetGallery.Core/Entities/Membership.cs
```diff
@@ -0,0 +1,60 @@
...
+        public bool IsAdmin { get; set; }
+
+        /// <summary>
+        /// Whether the <see cref="Member"/> is a collaborator for the <see cref="Organization"/>.
+        /// 
+        /// Collaborators have the following capabilities:
+        /// - Pushing new package versions
+        /// - Unlisting packages
+        /// </summary>
+        public bool IsCollaborator
```
joelverhagen: We should be careful with this, because I don't think it is usable in LINQ to SQL.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -118,6 +127,25 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
...
                            .MapRightKey("RoleKey"));
+            
+            modelBuilder.Entity<Organization>()
+                .HasKey(o => o.Key)
+                .HasRequired<User>(o => o.Account)
+                .WithOptional(u => u.Organization)
+                .WillCascadeOnDelete(false); // Disabled to prevent multiple cascade paths.
+
+            modelBuilder.Entity<Membership>()
+                .HasKey(m => new { m.OrganizationKey, m.MemberKey })
```
joelverhagen: Why the composite key? EF will already be creating an index for `OrganizationKey`.

src/NuGetGallery.Core/Entities/User.cs
```diff
@@ -24,6 +35,38 @@ public User(string username)
...
 
+        /// <summary>
+        /// <see cref="Organization"/> represented by this account, if any.
+        /// </summary>
+        public Organization Organization { get; set; }
+        
+        /// <summary>
+        /// Whether this is an <see cref="Organization"/> account.
+        /// </summary>
+        public bool IsOrganization
```
joelverhagen: AFAIK this is unusable in LINQ to SQL.

src/NuGetGallery.Core/Entities/Membership.cs
joelverhagen: In general I prefer methods here since it's a stronger signal to the caller that this is not a real SQL property and should only be used after a query is completed (i.e. in memory).

src/NuGetGallery.Core/Entities/EntitiesContext.cs
chenriksson: EF requires a PK on the Membership table, and throws an exception if I don't have it.

This is essentially a many-to-many table where the 2 foreign keys become the composite PK, same as with the PackageRegistrationOwners table. The only difference is that this includes additional state (IsAdmin), so I need to treat it as two one-to-many relations.

joelverhagen: Oh, I see `MemberKey` is a `Key` on the `Users` table. I misunderstood. Perfect 👌

src/NuGetGallery.Core/Entities/Membership.cs
chenriksson: Removed Membership.IsCollaborator and User.IsOrganization properties. If callers need something like that, we can add back helpers in later PRs.

src/NuGetGallery.Core/Entities/Membership.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// Note that there is no database contraint preventing memberships of Organizations into other
+        /// Organizations. For now this is restricted by the Gallery, but could be considered in the
+        /// future if we want to support Organization teams.
+        /// </summary>
+        public virtual User Member { get; set; }
+
+        /// <summary>
+        /// Whether the <see cref="Member"/> is an administrator for the <see cref="Organization"/>.
+        /// 
+        /// Administrators have the following capabilities:
```
scottbommarito: Should say explicitly that members that are not admins are collaborators.

It might not be best to list out all of the capabilities here. This comment is unlikely to change if we change the actual capabilities, so I'm concerned this list will become outdated and misleading.

If you do leave the list, you should specify what collaborators can do.

src/NuGetGallery.Core/Entities/Organization.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Linq;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// The organization associated with a <see cref="NuGetGallery.User" /> account.
+    /// 
+    /// The Users table contains both User and Organization accounts. The Organizations table exists both
+    /// as a constraint for Membership as well as a place for possible Organization-only settings. If User
+    /// and Organization settings diverge, we can consider creating a separate UserSettings table as well.
```
scottbommarito: nit: there are already fields in the `Users` table that don't apply to `Organizations`.

Ideally we should already create `UserSettings` but the value of doing so is just too low to prioritize it.

tests/NuGetGallery.Facts/TestUtils/FakeEntitiesContext.cs
```diff
@@ -98,6 +98,18 @@ public IDbSet<User> Users
...
 
+        public IDbSet<Organization> Organizations
+        {
+            get
+            {
+                return Set<Organization>();
+            }
+            set
+            {
+                throw new NotSupportedException();
```
scottbommarito: Why do these all have `set`? Is `set` ever used? Can we remove `set` and get rid of these `NotSupportedException`s?

src/NuGetGallery.Core/Entities/Organization.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// <summary>
+        /// Organization memberships.
+        /// </summary>
+        public virtual ICollection<Membership> Memberships { get; set; }
+
+        /// <summary>
+        /// Account (User) name.
+        /// </summary>
+        public string Name
```
joelverhagen: Don't think this works in LINQ to SQL. Not sure tho.

joelverhagen:
In general I think we should avoid convenience properties on EF entities since it is unclear from the interface if they are usable in LINQ to SQL. Maybe we can add analogous methods as they are needed by our business logic. 

Also, what SQL does this migration generate?

chenriksson:
@scottbommarito I don't think we need a feature branch for this feature. Most of these changes should be transparent to the user until we enable onboarding. Let me know if there are concerns with this.

tests/NuGetGallery.Facts/TestUtils/FakeEntitiesContext.cs
chenriksson: The setters exist on the IEntitiesContext properties. The test context doesn't support it, so throw ensures it isn't used.

scottbommarito: Didn't realize these were tests, woops

src/NuGetGallery.Core/Entities/Organization.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// <summary>
+        /// Organization account (User) foreign key.
+        /// </summary>
+        public int AccountKey { get; set; }
+
+        /// <summary>
+        /// Organization account (User).
+        /// </summary>
+        public User Account { get; set; }
```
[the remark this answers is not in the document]
chenriksson: Removed

Issue #4787, opened 2017-10-02
UploadPackage.cshtml's ValidationSummary not showing VerifyPackage errors


chenriksson:
Regression from redesign.

#4763 logs exceptions as traces, so we can see what happened.

This issue is to fix the validation summary so the user sees validation errors again. I think the summary is working for upload, but not verify.

scottbommarito:
@chenriksson does this issue suggest that ANY issue with the verify package form doesn't show up?

Is there a quick repro for this that would help me fix this quickly?

chenriksson:
Repro:
1. Upload package with Id {X} and Title {Y} as User {A}
2. Upload package with Id {Y} as User {B}
  - During the Verify step, make sure the package Title is also set to {Y}
3. Submit the upload

Expected:
User should see validation error "The ID of your package, '{Y}', is similar to the title of an existing package, which can cause confusion with our users. Please modify the ID of your package and try uploading again." (see issue #809)

Actual:
User just sees error "An unexpected error occurred. Contact support for assistance."

scottbommarito:
Upon investigation, VerifyPackage errors are shown--the scenario you've given me is a specific error case that we are not catching and returning a message for.

chenriksson:
Deployed on 2017.11.30 - please verify in PROD!

skofman1:
@scottbommarito , please verify and close

scottbommarito:
Verified in prod.

PR #4905, opened 2017-10-25
Fix readme telemetry

chenriksson:
Fixes bug from #827 where TrackPackageReadMeChangeEvent AI was logged under package push event name.

In addition, wanted to add tests to ensure that each API uses the correct event name. This required some refactoring to replace the static Telemetry class with an interface/wrapper for TelemetryClient.


src/NuGetGallery/OData/QueryAllowed/ODataQueryVerifier.cs
```diff
@@ -124,12 +123,11 @@ public static ODataQueryFilter V1Search
             }
             catch (Exception ex)
             {
-                var telemetryProperties = new Dictionary<string, string>();
-                telemetryProperties.Add(TelemetryService.CallContext, $"{telemetryContext}:{nameof(AreODataOptionsAllowed)}");
-                telemetryProperties.Add(TelemetryService.IsEnabled, $"{isFeatureEnabled}");
-
                 // Log and do not throw
```
scottbommarito: nit: this comment is not very useful

src/NuGetGallery/Services/ITelemetryService.cs
```diff
@@ -18,6 +19,14 @@ public interface ITelemetryService
 
         void TrackVerifyPackageKeyEvent(string packageId, string packageVersion, User user, IIdentity identity, int statusCode);
 
+        /// <summary>
+        /// Create a trace for an exception (Application Insights->traces)
```
scottbommarito: Why are we creating traces for any exceptions? Why not just track them all as exceptions?

scottbommarito: nit: this `ITelemetryService` isn't inherently for Application Insights--there could be another implementation that logs to somewhere else. We probably shouldn't explicitly call out Application Insights here.

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -5,20 +5,24 @@
...
+            public const string PackageReadMeChanged = "PackageReadMeChanged";
+        }
 
-        // Event types
-        public const string ODataQueryFilterEvent = "ODataQueryFilter";
-        public const string PackagePushEvent = "PackagePush";
-        public const string CreatePackageVerificationKeyEvent = "CreatePackageVerificationKeyEvent";
-        public const string VerifyPackageKeyEvent = "VerifyPackageKeyEvent";
-        public const string PackageReadMeChangeEvent = "PackageReadMeChanged";
+        private IDiagnosticsSource _trace;
```
scottbommarito: Can we rename this to `_diagnosticsSource`?

tests/NuGetGallery.Facts/Framework/Fakes.cs
```diff
@@ -80,11 +80,11 @@ public Fakes()
             {
                 Id = "FakePackage",
                 Owners = new List<User> {Owner},
-                Packages = new List<Package>
-                {
-                    new Package {Version = "1.0"},
-                    new Package {Version = "2.0"}
-                }
+            };
+            Package.Packages = new List<Package>
```
scottbommarito: nit: why didn't you leave this nested?

tests/NuGetGallery.Facts/Services/TelemetryServiceFacts.cs
```diff
@@ -20,6 +25,167 @@ public void ThrowsIfDiagnosticsServiceIsNull()
...
+
+                // Assert
+                service.TelemetryClient.Verify(c => c.TrackEvent(eventName,
+                    It.IsAny<IDictionary<string, string>>(),
+                    It.IsAny<IDictionary<string, double>>()),
+                    Times.Once);
+            }
+            
+            [Fact]
+            public void TrackPackageReadMeChangeEventThrowsIfPackageIsNull()
```
scottbommarito: Can we put tests for the same event in inner classes?

```
public class ThePackageReadMeChangeEvent
{
    [Fact]
    public void TrackPackageReadMeChangeEventThrowsIfPackageIsNull()
    {
        ...
    }
}
```

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -39,7 +39,12 @@ public class DefaultDependenciesModule : Module
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Maintainability", "CA1502:CyclomaticComplexity", Justification = "This code is more maintainable in the same function.")]
         protected override void Load(ContainerBuilder builder)
         {
-            var diagnosticsService = new DiagnosticsService();
+            var telemetryClient = TelemetryClientWrapper.Instance;
+            builder.RegisterInstance(telemetryClient)
+                .As<ITelemetryClient>()
```
joelverhagen: I'm not sure this guy should be a singleton.

From what I can tell online it looks like `TelemetryClient` inherit data from the request context:
https://stackoverflow.com/questions/34709278/using-static-instance-of-telemetryclient-with-application-insights#comment57259442_34709678

src/NuGetGallery/Services/ITelemetryService.cs
chenriksson: I added trace for user safe exceptions in #809 to help with support requests, due to bug #4787.

tests/NuGetGallery.Facts/Framework/Fakes.cs
chenriksson: I needed to populate the reverse relationship (Package.PackageRegistration), but that can't be done within the PackageRegistration initialization.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
skofman1: This was a Singleton before (see Telemetry.cs), so no change in behavior here.

---
In reply to: [147026697](https://github.com/NuGet/NuGetGallery/pull/4905#discussion_r147026697) [](ancestors = 147026697)

chenriksson: Ah, thanks! Will update.

src/NuGetGallery/Services/TelemetryClientWrapper.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// <summary>
+    /// Wrapper for the Application Insights TelemetryClient class.
+    /// </summary>
+    public class TelemetryClientWrapper : ITelemetryClient
+    {
+        internal static TelemetryClientWrapper Instance = new TelemetryClientWrapper();
+
+        private TelemetryClientWrapper()
+        {
+            UnderlyingClient = new TelemetryClient();
```
skofman1: Consider making this thread safe. Probably nothing will happen if TelemetryClient gets initialized a bunch of times, but still....

tests/NuGetGallery.Facts/Services/TelemetryServiceFacts.cs
chenriksson: All of the TrackXEvent APIs are thin wrappers around TrackEvent and share some of the same tests (e.g., TrackEventNames, TrackEventNamesIncludesAllEvents). This is why I chose to keep them together.

src/NuGetGallery/Services/TelemetryClientWrapper.cs
chenriksson: Should be safe... TelemetryClient is initialized in ctor, so each wrapper would have a single client instance. The wrapper is also implemented as a singleton with private ctor and static instance. I did update it to use Lazy for deferred initialization, which should be thread safe.

chenriksson:
@joelverhagen Removed all non-EF (schema) helpers from the entities.

Totally onboard with the separation of EF entities from helpers, which could help prevent accidental schema changes as well.  I believe EF schema generation originally used partials for this, but with code first maybe it makes sense for us to create extensions for entities instead?

We already have this problem with existing entities... maybe I should file an issue to refactor, so that this new pattern is adopted?

joelverhagen:
> We already have this problem with existing entities... maybe I should file an issue to refactor, so that this new pattern is adopted?

Seems reasonable. One extreme would be mapping DAL entities to business logic types and baking all query patterns we need down into the DAL. Then we can have all of the business logic/convenience properties and methods that we want -- on the BL types. This is extreme and a big deviation from what we have today though. Perhaps a middle ground is writing all of these convenience methods/properties as methods of in our `*Service` classes.

chenriksson:
@joelverhagen Oops, my last comment was intended for PR #4904. I've got too many tabs open :-)

Issue #4886, opened 2017-10-24
[Organizations] ExtensionMethods.IsOwner should check for the user's membership in an organization that owns the package


scottbommarito:
This will allow organization members to perform actions on packages owned by their organizations.

We should also add a parameter for whether or not the member must be an admin or not. This parameter should default to false.

scottbommarito:
https://github.com/NuGet/NuGetGallery/pull/4923

PR #4916, opened 2017-10-26
Modify ExtensionMethods.IsOwner to support Organizations

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/4886

`ExtensionMethods.IsOwner` will now return true if the user is an admin of an organization that owns the package.

Added a parameter (`allowCollaborator`) to return true if the user is a collaborator of an organization that owns the package.

Removed the old `IsOwnerOrAdmin` method and replaced it with `IsOwner`'s `allowAdmin` parameter.

Also did the same for `ListPackageItemViewModel.IsOwner`.

### DIFF
--- a/src/NuGetGallery/ExtensionMethods.cs
+++ b/src/NuGetGallery/ExtensionMethods.cs
@@ -206,8 +206,8 @@ namespace NuGetGallery
                     .Where(u => u.Organization != null)
                     .Any(u =>
                     {
-                        var members = allowCollaborator ? u.Organization.Members : u.Organization.Administrators;
-                        return members.Any(m => m.Username == user.Identity.Name);
+                        var members = allowCollaborator ? u.Organization.Memberships : u.Organization.Memberships.Where(m => m.IsAdmin);
+                        return members.Any(m => m.Member.Username == user.Identity.Name);
                     });
         }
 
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/ExtensionMethods.cs
```diff
@@ -177,40 +177,62 @@ public static bool AnySafe<T>(this IEnumerable<T> items, Func<T, bool> predicate
...
 
-        public static bool IsOwner(this Package package, User user)
+        public static bool IsOwner(this Package package, User user, bool allowAdmin = false, bool allowCollaborator = false)
         {
-            return package.PackageRegistration.IsOwner(user);
+            return package.PackageRegistration.IsOwner(user, allowAdmin, allowCollaborator);
         }
 
-        public static bool IsOwnerOrAdmin(this PackageRegistration package, IPrincipal user)
+        public static bool IsOwner(this PackageRegistration package, IPrincipal user, bool allowAdmin = false, bool allowCollaborator = false)
```
joelverhagen: Please make `allowAdmin` default `true`. In general admins should be able to do anything owners do unless otherwise opted out (like uploading a new version).

Make `allowCollaborator` a different method.

src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs
```diff
@@ -49,18 +49,22 @@ public bool UseVersion
...
             {
                 return false;
             }
-            return Owners.Any(u => u.Username == user.Identity.Name);
-        }
-
-        public bool IsOwnerOrAdmin(IPrincipal user)
-        {
-            return IsOwner(user) || user.IsAdministrator();
+            return
```
joelverhagen: This should use one of the methods above.

scottbommarito:
After offline discussion with @ryuyu and @joelverhagen, I'll be creating a more explicit permissions system instead of refactoring this function, which is basically "does user have owner permissions" right now.

PR #4918, opened 2017-10-27
Pakage URL generation method and checks whether package file exists

agr:
Validation.Orchestrator needs a way to get public package URL, too (for the revalidation case when we don't have a package in validation container, also needed for the non-blocking validation case).
Also, methods that check for package file existence would allow to figure out where package file is located on Orchestrator side without trying to download either file.

### DIFF
--- a/src/NuGetGallery.Core/Services/CorePackageFileService.cs
+++ b/src/NuGetGallery.Core/Services/CorePackageFileService.cs
@@ -34,6 +34,18 @@ namespace NuGetGallery
             return _fileStorageService.GetFileAsync(CoreConstants.PackagesFolderName, fileName);
         }
 
+        public Task<Uri> GetPackageUrlAsync(Package package)
+        {
+            var fileName = BuildFileName(package, CoreConstants.PackageFileSavePathTemplate, CoreConstants.NuGetPackageFileExtension);
+            return _fileStorageService.GetFileReadUriAsync(CoreConstants.PackagesFolderName, fileName, endOfAccess: null);
+        }
+
+        public Task<bool> DoesPackageFileExistAsync(Package package)
+        {
+            var fileName = BuildFileName(package, CoreConstants.PackageFileSavePathTemplate, CoreConstants.NuGetPackageFileExtension);
+            return _fileStorageService.FileExistsAsync(CoreConstants.PackagesFolderName, fileName);
+        }
+
         public Task SaveValidationPackageFileAsync(Package package, Stream packageFile)
         {
             if (packageFile == null)
@@ -108,6 +120,12 @@ namespace NuGetGallery
                 CoreConstants.NuGetPackageFileExtension);
 
             return _fileStorageService.GetFileReadUriAsync(CoreConstants.ValidationFolderName, fileName, endOfAccess);
+        }
+
+        public Task<bool> DoesValidationPackageFileExistAsync(Package package)
+        {
+            var fileName = BuildFileName(package, CoreConstants.PackageFileSavePathTemplate, CoreConstants.NuGetPackageFileExtension);
+            return _fileStorageService.FileExistsAsync(CoreConstants.ValidationFolderName, fileName);
         }
 
         protected static string BuildFileName(Package package, string format, string extension)
--- a/src/NuGetGallery.Core/Services/CorePackageFileService.cs
+++ b/src/NuGetGallery.Core/Services/CorePackageFileService.cs
@@ -34,7 +34,7 @@ namespace NuGetGallery
             return _fileStorageService.GetFileAsync(CoreConstants.PackagesFolderName, fileName);
         }
 
-        public Task<Uri> GetPackageUrlAsync(Package package)
+        public Task<Uri> GetPackageReadUriAsync(Package package)
         {
             var fileName = BuildFileName(package, CoreConstants.PackageFileSavePathTemplate, CoreConstants.NuGetPackageFileExtension);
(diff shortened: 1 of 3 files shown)


src/NuGetGallery.Core/Services/CorePackageFileService.cs
```diff
@@ -34,6 +34,18 @@ public Task<Stream> DownloadPackageFileAsync(Package package)
             return _fileStorageService.GetFileAsync(CoreConstants.PackagesFolderName, fileName);
         }
 
+        public Task<Uri> GetPackageUrlAsync(Package package)
```
joelverhagen: This name should mimic the `GetValidationPackageReadUriAsync` method, e.g. `GetPackageReadUriAsync`. The fact that one is private and requires SAS and another does not is an implementation detail that does not need to be differentiated on the interface.

joelverhagen: Well, that's my opinion at least. Thoughts?

agr: You are probably right. My initial idea was to make a method that generates public URL, but that's impossible in general case: it is not required to store package in publicly available storage (i.e. file storage for personal setups), so name is too generic. I'll rename and update the documentataion

src/NuGetGallery.Core/Services/ICorePackageFileService.cs
```diff
@@ -20,6 +20,24 @@ public interface ICorePackageFileService
...
         /// <summary>
+        /// Generates the URL for the specified package in the public container for available packages.
+        /// </summary>
+        /// <param name="package">The package metadata.</param>
+        /// <returns>Package download URL</returns>
+        /// <remarks>
+        /// The returned URL is only intended to be used by the internal tooling and not for the user:
+        /// it might not make any sense to external users as it can be, for example, a file:/// URL.
+        /// </remarks>
+        Task<Uri> GetPackageReadUriAsync(Package package);
```
chenriksson: 'Read' is a bit misleading in the API name; what's important is that it's the public packages container.

I like how you used 'ValidationPackage' in other APIs, which is obvious from the name. Could we use a consistent pattern like this:
- Get**Package**UriAsync or Get**PublicPackage**UriAsync => 'packages' container
- Get**ValidationPackage**UriAsync => 'validation' container
- Get**BackupPackage**UriAsync => 'package-backups' container

src/NuGetGallery.Core/Services/ICorePackageFileService.cs
```diff
@@ -20,6 +20,24 @@ public interface ICorePackageFileService
...
+
+        /// <summary>
+        /// Checks whether package file exists in the public container for available packages
+        /// </summary>
+        /// <param name="">The package metadata</param>
+        /// <returns>True if file exists, false otherwise</returns>
+        Task<bool> DoesPackageFileExistAsync(Package package);
```
chenriksson: nit: 'File' could be removed from these API names to shorten, and since storage is an abstraction (could be file or blob). Maybe this would be a hassle though, since existing APIs use this pattern.

tests/NuGetGallery.Core.Facts/Services/CorePackageFileServiceFacts.cs
```diff
@@ -412,6 +412,89 @@ public class TheGetValidationPackageReadUriAsyncMethod : FactsBase
             }
         }
 
+        public class TheGetPackageReadUriMethod : FactsBase
+        {
+            [Fact]
+            public async Task WillThrowIfPackageIsNull()
+            {
+                var ex = await Assert.ThrowsAsync<ArgumentNullException>(() => _service.GetPackageReadUriAsync(null));
+                Assert.Equal(ex.ParamName, "package");
```
chenriksson: nit: expected should come before actual in Assert

tests/NuGetGallery.Core.Facts/Services/CorePackageFileServiceFacts.cs
```diff
@@ -412,6 +412,89 @@ public class TheGetValidationPackageReadUriAsyncMethod : FactsBase
...
+                await _service.GetPackageReadUriAsync(_package);
+
+                string filename = BuildFileName(_package.PackageRegistration.Id, _package.NormalizedVersion, CoreConstants.NuGetPackageFileExtension, CoreConstants.PackageFileSavePathTemplate);
+
+                _fileStorageService.Verify(
+                    x => x.GetFileReadUriAsync(PackagesFolderName, filename, null),
+                    Times.Once);
+                _fileStorageService.Verify(
+                    x => x.GetFileReadUriAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<DateTimeOffset?>()),
+                    Times.Once);
```
chenriksson: Are we calling GetFileReadUriAsync twice? Why two verifies?

src/NuGetGallery.Core/Services/ICorePackageFileService.cs
agr: The naming follows seemingly established pattern for methods that already exist: `DownloadPackageFileAsync`, `SavePackageFileAsync`. No "public" in either of them.

tests/NuGetGallery.Core.Facts/Services/CorePackageFileServiceFacts.cs
agr: First verifies that it was called with expected arguments exactly once, second verifies that it was called with any arguments exactly once, i.e. that it wasn't called with anything else but the expected arguments.

PR #4919, opened 2017-10-27
Add support for building signed&unsigned nugetgallery.core packages

skofman1:


build.ps1
```diff
@@ -7,6 +7,7 @@ param (
     [switch]$CleanCache,
     [string]$SimpleVersion = '1.0.0',
     [string]$SemanticVersion = '1.0.0-zlocal',
+    [string]$PackageSuffix,
```
joelverhagen: Should this default to `''`? Does this work locally without parameters?

skofman1: Works without providing a parameter

PR #4923, opened 2017-10-27
Use PackagePermissionsService to determine actions a user can perform on a package

scottbommarito:
This replaces `ExtensionMethods.IsOwner`.

I went through all previous calls of `IsOwner` and replaced them with a `HasPermissions` call with the right `Permission`. I also added checks to `_UserPackagesList.cshtml` to only show actions that the user can perform.


src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,173 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Principal;
+
+namespace NuGetGallery
+{
+    public static class PackagePermissionsService
+    {
+        public enum PermissionLevel
```
joelverhagen: I don't think should not be nested. It makes the call sites too verbose.

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,173 @@
...
+        internal static IEnumerable<Permission> GetPermissions(PermissionLevel permissionLevel)
+        {
+            switch (permissionLevel)
+            {
+                case PermissionLevel.Owner:
+                case PermissionLevel.OrganizationAdmin:
+
+                    yield return Permission.ReportMyPackage;
+
+                    goto case PermissionLevel.SiteAdmin;
```
joelverhagen: nit: can we write this without `goto`?

scottbommarito: yes, but it will be significantly more verbose

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Linq;
+using System.Security.Principal;
+
+namespace NuGetGallery
+{
+    public static class PackagePermissionsService
+    {
+        public enum PermissionLevel
+        {
+            None,
```
joelverhagen: Perhaps this could be call `Anonymous` -- this enum almost role

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: done

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,173 @@
...
+                case PermissionLevel.SiteAdmin:
+
+                    yield return Permission.ManagePackageOwners;
+
+                    goto case PermissionLevel.OrganizationCollaborator;
+
+                case PermissionLevel.OrganizationCollaborator:
+
+                    yield return Permission.DisplayMyPackage;
+                    yield return Permission.Upload;
```
joelverhagen: How is the notion of `UploadNewVersion` captured? Is that out of scope for this particular PR?

tests/NuGetGallery.Facts/Services/PackagePermissionsServiceFacts.cs
```diff
@@ -0,0 +1,159 @@
...
+                            PackagePermissionsService.Permission.Edit,
+                            PackagePermissionsService.Permission.Delete
+                        }
+                    },
+                    {
+                        PackagePermissionsService.PermissionLevel.SiteAdmin,
+                        new []
+                        {
+                            PackagePermissionsService.Permission.DisplayMyPackage,
+                            PackagePermissionsService.Permission.Upload,
```
joelverhagen: This is new for `SiteAdmin` right?

tests/NuGetGallery.Facts/Services/PackagePermissionsServiceFacts.cs
```diff
@@ -0,0 +1,159 @@
...
+                {
+                    { PackagePermissionsService.PermissionLevel.None, new PackagePermissionsService.Permission[0] },
+                    {
+                        PackagePermissionsService.PermissionLevel.OrganizationCollaborator,
+                        new []
+                        {
+                            PackagePermissionsService.Permission.DisplayMyPackage,
+                            PackagePermissionsService.Permission.Upload,
+                            PackagePermissionsService.Permission.Edit,
+                            PackagePermissionsService.Permission.Delete
```
joelverhagen: super nit: trailing `,` so that adding new permissions only effects on line (easier to read diff)

tests/NuGetGallery.Facts/Services/PackagePermissionsServiceFacts.cs
scottbommarito: Yes, good catch.

I thought it makes sense, but I'm fine removing the permission

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: I think this should be renamed to `UploadNewVersion`. These permissions only affect existing packages.

joelverhagen: Aren't `Upload` and `UploadNewVersion` different permissions tho?

tests/NuGetGallery.Facts/Services/PackagePermissionsServiceFacts.cs
scottbommarito: done

src/NuGetGallery/Services/PackagePermissionsService.cs
joelverhagen: verbose is better IMO for something so critical as permissions.

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: Yes, but `UploadNewId` won't come into play until the user can choose the owner of the new package. These permissions are "I have this existing package, what can I do to it."

E.g. out of scope for this PR

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: IMO `None` makes more sense because they have "no" permissions with the package

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: Which is why I made the tests very explicit. If someone wants to change the permissions they will need to change the tests in the exact same way.

joelverhagen: Implementation should be as clear or clearer than tests. People trying to understand our code should not have to dig through unit tests for comprehension. Alternative is to have docs concerning this matrix of permissions.

skofman1: please no goto in code

---
In reply to: [147482420](https://github.com/NuGet/NuGetGallery/pull/4923#discussion_r147482420) [](ancestors = 147482420)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -392,7 +392,7 @@ public virtual JsonResult UploadPackageProgress()
             if (package == null
                 || ((package.PackageStatusKey == PackageStatus.Validating
                      || package.PackageStatusKey == PackageStatus.FailedValidation)
-                    && !package.IsOwnerOrAdmin(User)))
+                    && !PackagePermissionsService.HasPermission(package, User, Permission.DisplayMyPackage)))
```
skofman1: DisplayMyPackage isn't the best name here. Consider: DisplayHiddenPackage/DisplayPrivatePackage

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1136,7 +1136,7 @@ public virtual Task<ActionResult> RejectPendingOwnershipRequest(string id, strin
             }
 
             var user = GetCurrentUser();
-            if (package.IsOwner(user))
+            if (package.Owners.Any(owner => owner.Key == user.Key))
```
skofman1: >  if (package.Owners.Any(owner => owner.Key == user.Key)) [](start = 10, length = 57)

This should be handled by the PackagePermissionsService . We will need to add Org logic here soon, so it doesn't make sense to have to logic here

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,173 @@
...
+    public enum PermissionLevel
+    {
+        None,
+        Owner,
+        SiteAdmin,
+        OrganizationAdmin,
+        OrganizationCollaborator
+    }
+
+    public enum Permission
```
skofman1: This is not Permission, this is Action

src/NuGetGallery/Services/PackagePermissionsService.cs
ryuyu: can we just drop the goto and let it fall through? it looks like that is bascially what is happening here anyway?

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: changed

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: correct, didn't notice this

scottbommarito: upon looking over this case, this case shouldn't be handled by the `PackagePermissionsService`, because this is for adding a user as a co-owner

I didn't review the remainder of the code and did find some cases that used `Owners.Any` directly that should be using `PackagePermissionsService` however

ryuyu:
We should probably also use this to hide the Manage Package Owners button in the Manage Packages page for packages that users are collaborators on.

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: Thought about it more and I agree that `Anonymous` is better

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: C# does not allow falling through, which is why this exists as is

regardless, I rewrote it to be explicit, as @joelverhagen requested

src/NuGetGallery/Controllers/PackagesController.cs
skofman1: `upon looking over this case, this case shouldn't be handled by the PackagePermissionsService, because this is for adding a user as a co-owner` - we are handling co-ownership for packages, why shouldnt this be handled by PackagePermissionsService?

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,215 @@
...
+        Edit,
+        Delete,
+        ManagePackageOwners,
+        ReportMyPackage,
+    }
+
+    public static class PackagePermissionsService
+    {
+        private static readonly IDictionary<PermissionLevel, IEnumerable<Action>> _allowedActions = 
+            new Dictionary<PermissionLevel, IEnumerable<Action>>
```
joelverhagen: 🥇 this is beautiful

tests/NuGetGallery.Facts/Services/PackagePermissionsServiceFacts.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace NuGetGallery.Services
+{
+    public class PackagePermissionsServiceFacts
+    {
+        public class TheGetPermissionLevelMethod
+        {
+            private int _key = 0;
+
+            [Flags]
+            public enum ReturnsExpectedPermissionLevels_State
```
joelverhagen: 👍

scottbommarito:
@ryuyu it does that

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: yea i mean it didn't use `HasPermission`, but it uses `GetPermissionLevels` which used to be private

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: using `DisplayPrivatePackage`

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,215 @@
...
+        Owner,
+        SiteAdmin,
+        OrganizationAdmin,
+        OrganizationCollaborator
+    }
+
+    public enum PackageAction
+    {
+        DisplayPrivatePackage,
+        UploadNewVersion,
```
skofman1: UploadNewVersion is different then uploading a new package id?

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,215 @@
...
+        Edit,
+        Delete,
```
skofman1: >Delete [](start = 8, length = 6)

Delete is actually Unlist?

scottbommarito: Yes...the UI and API endpoints are named Delete. I can add comments

skofman1: IMO it's better to rename to Unlist, cause an SiteAdmin can actually do a Delete.

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,215 @@
...
+    {
+        Anonymous,
+        Owner,
+        SiteAdmin,
+        OrganizationAdmin,
+        OrganizationCollaborator
+    }
+
+    public enum PackageAction
+    {
```
chenriksson: PermissionLevel and PackageAction should be extracted into separate files.

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: yes, these are all actions on an existing package or id

`UploadNewId` is a right on an organization or a reserved namespace, not an existing package.

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: renamed

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            if (principal == null)
+            {
+                return new[] { PermissionLevel.Anonymous };
+            }
+
+            return GetPermissionLevels(
+                owners, 
+                principal.IsAdministrator(), 
+                u => UserMatchesPrincipal(u, principal));
```
chenriksson: nit: why not just pass principal to the core GetPrincipalLevels. One less arg, and no need for allocating the lambda.

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,215 @@
...
+
+    public static class PackagePermissionsService
+    {
+        private static readonly IDictionary<PermissionLevel, IEnumerable<PackageAction>> _allowedActions = 
+            new Dictionary<PermissionLevel, IEnumerable<PackageAction>>
+            {
+                {
+                    PermissionLevel.Anonymous,
+                    new PackageAction[0]
+                },
```
chenriksson: nit: why include anonymous? HasPermissions could just check, and return false immediately w/o linq.

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,215 @@
...
+    {
+        DisplayPrivatePackage,
+        UploadNewVersion,
+        Edit,
+        Delete,
+        ManagePackageOwners,
+        ReportMyPackage,
+    }
+
+    public static class PackagePermissionsService
```
chenriksson: nit: Is a static class really a service? Wondering if this class should be shortened to just PackagePermissions?

Also, 'Permissions' is redundant. Maybe 'PackagePermissionsService.HasPermissions(...)' could change to 'PackagePermissions.ActionAllowed(...)' or something.

src/NuGetGallery/Services/PackagePermissionsService.cs
```diff
@@ -0,0 +1,215 @@
...
+            var permissionLevels = GetPermissionLevels(owners, user);
+            return HasPermission(permissionLevels, action);
+        }
+
+        private static bool HasPermission(IEnumerable<PermissionLevel> permissionLevels, PackageAction action)
+        {
+            return permissionLevels.Any(permissionLevel => _allowedActions[permissionLevel].Contains(action));
+        }
+
+        public static IEnumerable<PermissionLevel> GetPermissionLevels(Package package, User user)
```
chenriksson: The GetPermissionLevels APIs seem like they should be private. The only use outside this class is a check if the user is the owner. Maybe you instead expose a single IsOwner API?

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: done

src/NuGetGallery/Services/PackagePermissionsService.cs
skofman1: I don't see an "UploadNewId" Action here

scottbommarito: It will come in a future PR

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: Not sure what you mean exactly...are you suggesting `UserMatchesPrincipal` and `UserMatchesUser` should return a `Func<User, bool>`?

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: not possible because I wanted to separate the allowed actions from the permission levels

scottbommarito: correction: if we ever decide `Anonymous` users can do actions, such as `ReportAbuse`, they would need state here

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: AFAIK it fits all the definitions of being a service, despite being a static class.

Will change to `ActionAllowed`

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: I like them being public, because I think there's a pretty high likelihood we will use them to check other states besides `IsOwner` in the future.

src/NuGetGallery/Services/PackagePermissionsService.cs
chenriksson: Just saying if anonymous has no actions, no need for the extra allocations... in case this is on the fast path. For instance:

```
private static bool IsActionAllowed(IEnumerable<PermissionLevel> permissionLevels, PackageAction action)
{
    return permissionLevels != null && permissionLevels.Any(permissionLevel => _allowedActions[permissionLevel].Contains(action));
}
public static IEnumerable<PermissionLevel> GetPermissionLevels(IEnumerable<User> owners, IPrincipal principal)
{
    if (principal == null)
    {
        return null;
    }
    ...
```

src/NuGetGallery/Services/PackagePermissionsService.cs
chenriksson: Just a suggestion, but I think it opens up complexity to the callers that this service was intended for.

src/NuGetGallery/Services/PackagePermissionsService.cs
scottbommarito: Ah, well I think I prefer it like this because
1 - it's easier to extend and add actions to anonymous if we decide we need to
2 - it matches the structure of the rest of the permission levels

Issue #4888, opened 2017-10-24
[Organizations] Only admins can accept package ownership requests on behalf of an organization


scottbommarito:
This should be an easy change to `PackagesController.HandleOwnershipRequest` which uses the `ExtensionMethods.IsOwner` with `IsOrganizationAdmin` set to true.

PR #4924, opened 2017-10-28
[Organizations] Allow admins to accept package ownership requests on behalf of an organization

scottbommarito:
This involved doing some more refactoring than I had expected on the `PackagePermissionsService` for the main purpose of
1 - distinguishing permissions between package permissions, user permissions, and organization permissions
2 - creating an action that is REJECTED if a user is a part of a permission level (see `PackagePermissionRestrictedActions.AcceptOwnership`

https://github.com/NuGet/NuGetGallery/issues/4888


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1123,8 +1123,10 @@ public virtual Task<ActionResult> RejectPendingOwnershipRequest(string id, strin
             {
                 return HttpNotFound();
             }
-
-            if (!string.Equals(username, User.Identity.Name, StringComparison.OrdinalIgnoreCase))
+            
+            var user = _userService.FindByUsername(username);
+            var permissionLevels = PermissionsService.GetPermissionLevels(user, GetCurrentUser());
+            if (!(permissionLevels.Contains(PermissionLevel.Owner) || permissionLevels.Contains(PermissionLevel.OrganizationAdmin)))
```
loic-sharma: >if (!(permissionLevels.Contains(PermissionLevel.Owner) || permissionLevels.Contains(PermissionLevel.OrganizationAdmin))) [](start = 12, length = 120)

Can you explain this line? `username` is the pending owner, correct? Does that mean that a pending owner should have `PermissionLevel.Owner` or `PermissionLevel.OrganizationAdmin` permissions?

scottbommarito: I should rewrite this to be more clear, but username is the pending user but we aren't comparing it against the pending user, we're comparing it against the current user, who may not be the pending user.

src/NuGetGallery/Services/UserPermissionRestrictedActions.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    public static class UserPermissionRestrictedActions
+    {
+        /// <summary>
+        /// If a user is requested to be an owner of a package, this user can accept the request on behalf of the other user.
+        /// </summary>
+        public static IPermissionRestrictedAction AcceptPackageOwnershipOnBehalfOf =
+            new PermissionRestrictedActionIncludeLevel(new PermissionLevel[]
+            {
+                PermissionLevel.Owner,
```
skofman1: Owner doesn't make sense here, because it's a package related word. Perhaps "Self"?

src/NuGetGallery/Services/PackagePermissionRestrictedActions.cs
```diff
@@ -0,0 +1,90 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// <see cref="IPermissionRestrictedAction"/>s that operate on a <see cref="Package"/> or <see cref="PackageRegistration"/>.
+    /// </summary>
+    public static class PackagePermissionRestrictedActions
```
chenriksson: You switched from actions by permissionLevel (PackagePermissionService._allowedActions) to permissionLevels by action. Why this change?

src/NuGetGallery/Services/PermissionLevel.cs
```diff
@@ -6,27 +6,27 @@ namespace NuGetGallery
     public enum PermissionLevel
```
chenriksson: Why not use Flags enum for PermissionLevel? Wouldn't this simplify the APIs?

```
public class PackagesController {
    public void CreatePackageInteral(...) {
        ...
        if (new package) {
            if (!Permissions.Matches(Level.Self | Level.OrganizationAdmin)) { ... }
        } else {
            if (!Permissions.Matches(Level.Self | Level.OrganizationAdmin | Level.OrganizationCollaborator)) { ... }
        }
```

src/NuGetGallery/Services/PackagePermissionRestrictedActions.cs
scottbommarito: Because I changed the logic of how actions determine if they are allowed.

Part of the reason for this was that `AcceptOwnershipOfPackage` is NOT allowed if the user is an `Owner`, but allowed for all other roles. This made the structure much more complex and no longer possible to store in a single dictionary.

src/NuGetGallery/Services/PermissionLevel.cs
scottbommarito: Callers don't use `PermissionLevel`, it's completely handled by the `PermissionsService`.

I like having `IEnumerable<PermissionLevel>` better because it doesn't hide the state that `PermissionLevel` is used here as a set.

src/NuGetGallery/Services/PackagePermissionRestrictedActions.cs
```diff
@@ -0,0 +1,90 @@
...
+                PermissionLevel.OrganizationAdmin,
+                PermissionLevel.SiteAdmin,
+            });
+
+        /// <summary>
+        /// The user can report the package as the package's owner.
+        /// This is usually used for requesting deletion of packages.
+        /// </summary>
+        public static IPermissionRestrictedAction ReportMyPackage =
+            new PermissionRestrictedActionIncludeLevel(new PermissionLevel[]
```
joelverhagen: `PermissionInclusionList`

src/NuGetGallery/Services/PackagePermissionRestrictedActions.cs
```diff
@@ -0,0 +1,90 @@
...
+{
+    /// <summary>
+    /// <see cref="IPermissionRestrictedAction"/>s that operate on a <see cref="Package"/> or <see cref="PackageRegistration"/>.
+    /// </summary>
+    public static class PackagePermissionRestrictedActions
+    {
+        /// <summary>
+        /// The user can accept ownership of the package.
+        /// </summary>
+        public static IPermissionRestrictedAction AcceptOwnership =
```
joelverhagen: This should not be consider a permission. We should just as easily redirect them to the "success" message. We've greatly complicated the code for a non-permission case.

src/NuGetGallery/Services/PackagePermissionRestrictedActions.cs
joelverhagen: In fact, if we can get ride of the "Exclude" type this can just be a `HashSet<PermissionLevel>`

src/NuGetGallery/Services/PermissionLevel.cs
chenriksson: It doesn't need to be exposed entirely to the caller. The point I'm getting at is that I think you can get away with just an int allocation for each action; as opposed to a dictionary of levels or arrays of ints for each.
```
GalleryActions {
    int PushPackage = Level.Owner | Level.OrganizatinAdmin;
    int PushPackageVersion = Level.Owner | Level.OrganizatinAdmin | Level.OrganizationCollaborator;
    ...
}
public void CreatePackageInternal(...) {
    if (new package) {
        if !(Permisssions.Matches(GalleryActions.PushPackage) { ... }
    } else {
        if !(Permisssions.Matches(GalleryActions.PushPackageVersion) { ... }
    }
}
```

src/NuGetGallery/Services/UserPermissionRestrictedActions.cs
scottbommarito: Renaming this from `UserActions` to `AccountActions` so `Owner` makes sense in this context.

src/NuGetGallery/Services/PermissionLevel.cs
scottbommarito: am using `int`s now

src/NuGetGallery/Services/PackagePermissionRestrictedActions.cs
scottbommarito: using @chenriksson's suggestion to use `[Flags]` on the enum so there's no need to store the array

src/NuGetGallery/Services/PackagePermissionRestrictedActions.cs
scottbommarito: removed

src/NuGetGallery/Services/PermissionsService.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            return IsActionAllowed(package.PackageRegistration, principal, actionPermissionLevel);
+        }
+
+        public static bool IsActionAllowed(PackageRegistration packageRegistration, IPrincipal principal, PermissionLevel actionPermissionLevel)
+        {
+            return IsActionAllowed(packageRegistration.Owners, principal, actionPermissionLevel);
+        }
+
+        public static bool IsActionAllowed(IEnumerable<User> owners, IPrincipal principal, PermissionLevel actionPermissionLevel)
```
skofman1: this is confusing. It's not clear that the owners are package owners and who is the logged in user. Please add comments and rename params to more specific names, like "package owners".

Also, there are a lot of public methods in this class. Are all used externally? Can some be made private?

scottbommarito: they are all used externally

they aren't explicitly package owners--they could be reserved namespace owners, package owners, or any group of owners of some entity. I will rename the `user` and `principal` parameters so that it is more clear that they are intended to be the current user, however.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -293,6 +293,11 @@ public ActionResult HealthProbe()
             await AuditingService.SaveAuditRecordAsync(
                 new PackageAuditRecord(package, AuditedPackageAction.Verify));
 
+            if (!PermissionsService.IsActionAllowed(package, user, PackageActions.UploadNewVersion))
```
chenriksson: PermissionsService should not be necessary in ApiController.

The HasAnyScopeThatAllows calls below ensure that the API key scopes match, which includes action, subject (packages), and owners.

src/NuGetGallery/Services/AccountActions.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Actions that a <see cref="User"/> can perform on another <see cref="User"/>.
+    /// </summary>
+    public static class AccountActions
```
chenriksson: Separate classes for AccountActions/PackageActions is odd, esp since there's only 1 account action. Why not just combine them all into a NuGetGalleryActions?

joelverhagen: Maybe we can group `AccountActions` and `PackageActions` into one static class with nested classes a la `Actions.Account` and `Actions.Package`. It might be convenient to have these two in the same file. Just a thought.

src/NuGetGallery/Services/PackageActions.cs
```diff
@@ -0,0 +1,63 @@
...
+    public static class PackageActions
+    {
+        /// <summary>
+        /// The user can see private information about the package.
+        /// </summary>
+        public static PermissionLevel DisplayPrivatePackage = 
+            PermissionLevel.Owner | 
+            PermissionLevel.OrganizationAdmin | 
+            PermissionLevel.SiteAdmin | 
+            PermissionLevel.OrganizationCollaborator;
```
chenriksson: This looks really nice!

src/NuGetGallery/Services/PermissionsService.cs
```diff
@@ -0,0 +1,159 @@
...
+
+            if (matchingMembers.Any())
+            {
+                permissionLevel |= PermissionLevel.OrganizationCollaborator;
+            }
+
+            return permissionLevel;
+        }
+
+        private static bool UserMatchesPrincipal(User user, IPrincipal principal)
```
chenriksson: Consider extracting small extensions of User/Principle into extension classes, for reuse across services. For instance:
```
class UserExtensions {
    bool MatchesUser(this User first, User second);
}
class PrincipalExtensions {
    bool MatchesUser(this IPrincipal principal, User user);
}
```

src/NuGetGallery/Services/PermissionsService.cs
```diff
@@ -0,0 +1,159 @@
...
+        internal static PermissionLevel GetPermissionLevel(IEnumerable<User> owners, User currentUser)
+        {
+            if (currentUser == null)
+            {
+                return PermissionLevel.Anonymous;
+            }
+
+            return GetPermissionLevel(
+                owners, 
+                currentUser.IsInRole(Constants.AdminRoleName),
```
chenriksson: Consider extracting into extension method, similar to the Principal extension...
```
class UserExtensions {
    bool IsAdministrator(this User user);
}
```

chenriksson:
Remove PermissionsService from ApiController (redundant/unnecessary), consider other feedback. Then :shipit:

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: git merge failed me, will fix

src/NuGetGallery/Services/AccountActions.cs
scottbommarito: there will be more coming very soon

scottbommarito: not sure...i will reconsider this later when the classes are fully fleshed out

i want to make sure there's not too many user actions that this just becomes unwieldy and i have to move it back anyway

PR #4929, opened 2017-10-30
Fix permission-related renaming issue

scottbommarito:
Also turned on MvcBuildViews compilation by default, so Visual Studio fails the build when this happens in the future. (It was already enabled in our build scripts.)

### DIFF
--- a/src/NuGetGallery/NuGetGallery.csproj
+++ b/src/NuGetGallery/NuGetGallery.csproj
@@ -30,7 +30,7 @@
     <CodeContractsAssemblyMode>0</CodeContractsAssemblyMode>
     <MvcProjectUpgradeChecked>true</MvcProjectUpgradeChecked>
     <UseGlobalApplicationHostFile />
-    <MvcBuildViews Condition=" '$(MvcBuildViews)' == '' ">true</MvcBuildViews>
+    <MvcBuildViews Condition=" '$(MvcBuildViews)' == '' ">false</MvcBuildViews>
     <NuGetPackageImportStamp>
     </NuGetPackageImportStamp>
     <Use64BitIISExpress />
--- a/src/NuGetGallery/NuGetGallery.csproj
+++ b/src/NuGetGallery/NuGetGallery.csproj
@@ -30,7 +30,7 @@
     <CodeContractsAssemblyMode>0</CodeContractsAssemblyMode>
     <MvcProjectUpgradeChecked>true</MvcProjectUpgradeChecked>
     <UseGlobalApplicationHostFile />
-    <MvcBuildViews Condition=" '$(MvcBuildViews)' == '' ">false</MvcBuildViews>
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -30,7 +30,7 @@
     <CodeContractsAssemblyMode>0</CodeContractsAssemblyMode>
     <MvcProjectUpgradeChecked>true</MvcProjectUpgradeChecked>
     <UseGlobalApplicationHostFile />
-    <MvcBuildViews Condition=" '$(MvcBuildViews)' == '' ">false</MvcBuildViews>
```
joelverhagen: I am not a big fan of this being on by default. To slows builds waaaaaaay down in VS 🐌. For me it's like a 10x factor. IMO this should only be done from the `build.ps1`.

xavierdecoster: It may be useful to have this locally too e.g. for `Release` build configuration? It allows to quickly check view compilation locally without having to wait for a full CI build.

joelverhagen: That's reasonable. BTW we are already doing this locally in build.ps1:
https://github.com/NuGet/NuGetGallery/blob/5e44c0786caf249bc50571f2e0dcc2bbbde30a9d/build.ps1#L84

scottbommarito: The reason I decided to turn this on was because the lack of visibility on `cshtml` compilation failures has caused me to push broken builds twice because of tooling issues. For example, this was caused by a rename that renamed everything except this `cshtml` file. I don't use the `build.ps1` locally--I rely on the fact that if the VS build succeeds, the code compiles.

I haven't seen nearly that much degradation in performance with this setting on, although it is noticeable.

I'm totally fine with setting this to `Release`-only, but I'm not sure how to do that.

scottbommarito: I'm going to take this change out of the PR for now just because I'd like to get this fix in ASAP

joelverhagen: `'$(MvcBuildViews)' == '' And '$(Configuration)' != 'Release'`

joelverhagen: Or something like that

scottbommarito: I'll file an issue!

chenriksson: That fix assumes you build Release before merge to detect cshtml compilation errors before CI. Maybe a better test is to just run `git clean` and `build.ps1` to test the same way CI would... nothing we can enforce, but a good practice for the team?

joelverhagen: I always run `build.ps1` locally to catch this sort of error and unit test regressions. Seems like a best practice IMHO.

PR #4935, opened 2017-10-31
Add owner scope to API keys

chenriksson:
Remaining work:
- ApiController scope evaluation
- Auditing
- APIKeys.cshtml UI


src/NuGetGallery.Core/Entities/Scope.cs
```diff
@@ -30,6 +38,14 @@ public Scope()
...
         }
 
+        public Scope(string owner, string subject, string allowedAction)
+        {
+            Owner = owner;
+            Subject = subject;
+            AllowedAction = allowedAction;
+        }
+
+        // Deprecated: Should be removed once ApiKeys.cshtml is updated to support owner scope.
```
scottbommarito: Wouldn't this just be as simple as changing a constructor call to pass the current username as owner?

You can add the ability to choose what user later. I would just prefer not having deprecated methods if it's very easy to fix.

chenriksson: Yes, but then I'd need to update all callers. I'm minimizing changes in this PR (schema and copying org scope for temp keys). I'll remove this ctor when I start requiring the owner, which happens when I enable the UI.

src/NuGetGallery.Core/Entities/Scope.cs
```diff
@@ -16,6 +15,15 @@ public class Scope
         [JsonIgnore]
         public int CredentialKey { get; set; }
 
+        /// <summary>
+        /// Package owner (user or organization) scoping.
+        /// </summary>
+        [JsonProperty("o")]
+        public string Owner { get; set; }
```
joelverhagen: Why isn't this a user key?

joelverhagen: The "identity" of a user is the `Key` property.

chenriksson: Good question... there was debate about whether keys can be scoped to multiple owners (yourself, and your organization(s)). We decided that we'd start with a single owner, but my impression was that we might change that.

I typed as string so that it could change with the same schema. Alternatively, I could change to FK and we move to many-many relation if we change our minds on supporting multiple owners.

chenriksson: @joelverhagen Switched owner scope to FK. This does make more sense - if we support multiple owners per API key, we can just add more scopes.

PR #4936, opened 2017-10-31
Admin - deleteaccount

cristinamanum:
Add the delete admin account workflow. Tests are in progress and will be added in the same pull request.

The pull request contains:
1. The admin UI changes.
- A setting in the admin page to search and delete an account 
- Delete option on the Profile view for the admin to Delete the user
- The UI page for the Admin to submit the delete action.
- The UI page for the admin to see the status of his initiated delete action.

2. Two sql changes

- One column added to the User table  keeping the IsDeleted status
- One table "AccountDeletes" that keeps the information for deletion time stamp, signature of the admin, the admin and deleted user  keys. 

3. DeleteAccount service and the necessary APIs to execute the delete/clean up actions.

Observation:
With this change it is enabled a scenario that was disabled previously: Remove the last owner of a package. This change is necessary in order to clean up the account.


scottbommarito:
Pictures for each UI page please

src/NuGetGallery.Core/Entities/AccountDelete.cs
```diff
@@ -0,0 +1,36 @@
...
+        public int Key { get; set; }
+
+        /// <summary>
+        /// The date when the account was deleted.
+        /// </summary>
+        [Required]
+        public DateTime DeletedOn { get; set; }
+
+        /// <summary>
+        /// The User that had the account deleted.
```
shishirx34: may be: `The user that requested the account to be delete.`

src/NuGetGallery.Core/Entities/AccountDelete.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// </summary>
+        public User DeletedAccount { get; set; }
```
shishirx34: `DeleteRequestAccount` or `RequestAccount` may be appropriate, given that the class is named `AccountDelete`

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -280,6 +280,16 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
...
                     .WithOptional();
+
+            modelBuilder.Entity<AccountDelete>()
+           .HasKey(a => a.Key)
+           .HasRequired(a => a.DeletedAccount);
+
+            modelBuilder.Entity<AccountDelete>()
+            .HasRequired(a => a.DeletedBy)
+            .WithMany()
+            .WillCascadeOnDelete(false);
```
shishirx34: Indentation

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -275,6 +275,11 @@ public static void RegisterUIRoutes(RouteCollection routes)
                 new { controller = "Users", action = "ChangeEmailSubscription" });
 
             routes.MapRoute(
+                RouteName.AdminDeleteAccount,
+                "admin/deleteuseraccount",
```
shishirx34: nit: `admin/deleteuser` seems sufficient for route name. Don't feel too strongly, up to you.

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
```diff
@@ -0,0 +1,48 @@
...
+            };
+            return View(model);
+        }
+
+        [HttpGet]
+        public virtual ActionResult Search(string query)
+        {
+            var results = new List<DeleteAccountSearchResult>();
+            var result = new DeleteAccountSearchResult();
+            if (query != null)
```
shishirx34: `if (!string.IsNullOrWhitespace(query))`

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
```diff
@@ -0,0 +1,48 @@
...
+                if (user != null)
+                {
+                    result.AccountName = user.Username;
+                }
+            }
+            if(result.AccountName!=null)
+            {
+                results.Add(result);
+            }
+            return Json( results, JsonRequestBehavior.AllowGet);
```
shishirx34: please format for new lines after `}`

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteUserAccount.cshtml
```diff
@@ -0,0 +1,83 @@
+﻿@using NuGetGallery
+@model NuGetGallery.Areas.Admin.ViewModels.DeleteUserAccountViewModel
+@{
+    ViewBag.Title = "Delete Account" + Model.AccountName;
```
shishirx34: nit: `"Delete Account " + Model.AccountName;` **spacing**

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteUserAccount.cshtml
```diff
@@ -0,0 +1,83 @@
...
+    ViewBag.MdPageColumns = Constants.ColumnsFormMd;
+    Layout = "~/Views/Shared/Gallery/Layout.cshtml";
+}
+
```
shishirx34: extra new line

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteUserAccount.cshtml
```diff
@@ -0,0 +1,83 @@
...
+        <div class="form-group">
+            <h1 class="text-center">Delete Account</h1>
+            <div class="text-center ms-font-xxl">
+                <a href="@Url.User(@Model.User)">@Model.AccountName</a>
+            </div>
+
+            @ViewHelpers.AlertDanger(@<p>
+                <b class="keywords">Once the account is deleted, it cannot be undone!</b> <br />
+                Deleting the account will: <br />
+                1. Revoke associated API key/s. <br />
```
shishirx34: `key(s)`

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/Index.cshtml
```diff
@@ -0,0 +1,67 @@
...
+
+@{
+    ViewBag.Title = "Delete account";
+}
+
+<section>
+    <article id="stage">
+        <h2>Delete Account</h2>
+        <form>
+            <textarea placeholder="Search for account to delete. Only one account at a time." autocomplete="off" autofocus style="width: 100%;" rows="5" data-bind="value: searchQuery"></textarea><br />
```
shishirx34: may be change this to `input type='text'` since you want to allow only single user account search possible?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -90,6 +95,42 @@ public virtual ActionResult Account()
...
+        public virtual ActionResult DeleteUserAccount(string accountName)
+        {
+            var user = _userService.FindByUsername(accountName);
+            if(user == null)
+            {
+                return HttpNotFound("User not found.");
+            }
+
+            var listPackageItems = _packageService
+                 .FindPackagesByOwner(user, true)
```
shishirx34: named parameter for `true`

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -90,6 +95,42 @@ public virtual ActionResult Account()
...
+        [Authorize(Roles = "Admins")]
+        [HttpPost]
+        [RequiresAccountConfirmation("delete account")]
+        [ValidateAntiForgeryToken]
+        public virtual async Task<ActionResult> DeleteUserAccount(DeleteUserAccountViewModel model)
+        {
+            var user = _userService.FindByUsername(model.AccountName);
+            var admin = GetCurrentUser();
+            var status = await _deleteAccountService.DeleteGalleryUserAccountAsync(user, admin, model.Signature, model.Unlist);
+            return  View("DeleteUserAccountStatus", new DeleteUserAccountStatusViewModel() {AccountName = model.AccountName , OperationStatus = status .Item2});
```
shishirx34: nit: extra spacing after `return`

shishirx34: and after `status`? Weird this didn't give you compile time error?

shishirx34: can you format this line for spacings.

src/NuGetGallery.Core/Entities/AccountDelete.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <summary>
+        /// The User(admin) that executed the delete action.
+        /// </summary>
+        public User DeletedBy { get; set; }
```
joelverhagen: Do we need the key of the `DeletedBy` and `DeletedAccount` navigation properties? If this is working for you locally, perhaps EF is adding them automagically? If so, I think it's better to be explicit.

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public DeleteAccountController(IUserService userService)
+        {
+            _userService = userService;
+        }
+
+        [HttpGet]
+        public virtual ActionResult Index()
+        {
+            var model = new DeleteAccountRequest
```
joelverhagen: nit `DeletedAccountRequest();`.

src/NuGetGallery/Migrations/201710302343008_AddAccountDelete.cs
```diff
@@ -0,0 +1,39 @@
...
+                        DeletedAccount_Key = c.Int(nullable: false),
+                        DeletedBy_Key = c.Int(nullable: false),
+                    })
+                .PrimaryKey(t => t.Key)
+                .ForeignKey("dbo.Users", t => t.DeletedAccount_Key, cascadeDelete: true)
+                .ForeignKey("dbo.Users", t => t.DeletedBy_Key)
+                .Index(t => t.DeletedAccount_Key)
+                .Index(t => t.DeletedBy_Key);
+            
+            AddColumn("dbo.Users", "IsDeleted", c => c.Boolean(nullable: false));
```
joelverhagen: This needs a default value, right? I think if it's not nullable and records already exist the migration will fail.

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -275,6 +275,11 @@ public static void RegisterUIRoutes(RouteCollection routes)
...
+                new { controller = "Users", action = "DeleteUserAccount" });
```
joelverhagen: nit nameof for the `action`

src/NuGetGallery/Areas/Admin/ViewModels/DeleteAccountSearchResult.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGetGallery.Areas.Admin.ViewModels
+{
+    public class DeleteAccountSearchResult
+    {
```
joelverhagen: Can we make these immutable with ctor and `{ get; }` (read only props)? It makes the code easier to understand.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -335,7 +376,7 @@ public virtual ActionResult ResetPassword(bool forgot)
         public virtual ActionResult Profiles(string username, int page = 1)
         {
             var user = _userService.FindByUsername(username);
-            if (user == null)
+            if (user == null || user.IsDeleted)
```
joelverhagen: Why? The user can't even log in right?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// 1. The user will be removed as owner from his owned packages.
+        /// 2. Any of the packages that become orphaned as this result will be unlisted if the unlistOrphanPackages is set to true.
+        /// 3. Any owned namespaces will be released.
+        /// 4. The user credentials will be cleaned.
+        /// 5. The user data will be cleaned.
+        /// </summary>
+        /// <param name="userName">The NuGet user name</param>
+        /// <param name="unlistOrphanPackages">True if the orphan packages will be unlisted.</param>
+        /// <returns>A list with information regarding the status of each clean-up operation.</returns>
+        public async Task<Tuple<bool, List<string>>> DeleteGalleryUserAccountAsync(User useToBeDeleted, User admin, string signature, bool unlistOrphanPackages)
```
joelverhagen: What if the user is currently logged in. Can we forceably log them out?

joelverhagen: Please no tuple. Can we have a `class` that is descriptive of the output?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,213 @@
...
+            if (admin == null)
+            {
+                throw new ArgumentNullException(nameof(admin));
+            }
+
+            var ownedPackages = _packageService.FindPackagesByOwner(useToBeDeleted, includeUnlisted: true).ToList();
+            List<string> executionDetails = new List<string>();
+            bool result = true;
+
+            result = (await RemoveOwnership(useToBeDeleted, admin, unlistOrphanPackages, ownedPackages, executionDetails)) ?
```
joelverhagen: These need to be done during the same transaction. So we should have only one `CommitChangesAsync`.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,213 @@
...
+                            (await RemoveReservedNamespaces(useToBeDeleted, executionDetails) ?
+                                (await RemoveSecurityPolicies(useToBeDeleted, executionDetails) ?
```
joelverhagen: I think this code layout is actually more confusing that a series of `if`s with early returns.

src/NuGetGallery/Views/Users/Profiles.cshtml
```diff
@@ -7,9 +7,8 @@
 
 <section role="main" class="container main-container page-profile">
     <section role="main" class="row">
-        <aside class="col-md-3 col-md-push-9 profile-details">
+        <aside class="col-md-3 col-md-push-9  profile-details">
```
joelverhagen: nit: we don't need this extra space

joelverhagen:
What is the SQL that this migration generates?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,213 @@
...
+            _auditingService = auditingService ?? throw new ArgumentNullException(nameof(auditingService));
+            _userService = userService ?? throw new ArgumentNullException(nameof(userService));
+            _securityPolicyService = securityPolicyService ?? throw new ArgumentNullException(nameof(securityPolicyService));
+            _authService = authService ?? throw new ArgumentNullException(nameof(authService));
+        }
+
+        /// <summary>
+        /// Will clean-up the data related with an user account.
+        /// The result will be:
+        /// 1. The user will be removed as owner from his owned packages.
```
loic-sharma: Please use gender neutral terms: `his` -> `their`.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,213 @@
...
+                    Signature = signature
+                };
+                _accountDeleteRepository.InsertOnCommit(accountDelete);
+                await _accountDeleteRepository.CommitChangesAsync();
+                executionDetails.Add($"{nameof(InsertDeleteAccount)}: Succeeded");
+                return true;
+            }
+            catch (Exception e)
+            {
+                executionDetails.Add($"{nameof(InsertDeleteAccount)}: {e.ToString()}");
```
loic-sharma: `{e.ToString()}` should be just `{e}`

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -90,6 +95,42 @@ public virtual ActionResult Account()
         }
 
         [HttpGet]
+        [Authorize(Roles = "Admins")]
+        public virtual ActionResult DeleteUserAccount(string accountName)
+        {
+            var user = _userService.FindByUsername(accountName);
+            if(user == null)
```
scottbommarito: We should do something here if the user is already deleted.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -90,6 +95,42 @@ public virtual ActionResult Account()
...
+            return View("DeleteUserAccount", model);
+        }
+
+        [Authorize(Roles = "Admins")]
+        [HttpPost]
+        [RequiresAccountConfirmation("delete account")]
+        [ValidateAntiForgeryToken]
+        public virtual async Task<ActionResult> DeleteUserAccount(DeleteUserAccountViewModel model)
+        {
+            var user = _userService.FindByUsername(model.AccountName);
```
scottbommarito: What if `user` is null?

What if `user` is already deleted?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// 3. Any owned namespaces will be released.
+        /// 4. The user credentials will be cleaned.
+        /// 5. The user data will be cleaned.
+        /// </summary>
+        /// <param name="userName">The NuGet user name</param>
+        /// <param name="unlistOrphanPackages">True if the orphan packages will be unlisted.</param>
+        /// <returns>A list with information regarding the status of each clean-up operation.</returns>
+        public async Task<Tuple<bool, List<string>>> DeleteGalleryUserAccountAsync(User useToBeDeleted, User admin, string signature, bool unlistOrphanPackages)
+        {
+            if (useToBeDeleted == null)
```
scottbommarito: Should throw if `userToBeDeleted` is already deleted.

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: `useToBeDeleted` -> `userToBeDeleted`

src/NuGetGallery.Core/Entities/AccountDelete.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace NuGetGallery
+{
+    public class AccountDelete
```
chenriksson: Per our discussion, would consider inheritance here... though may depend on how/if we support org deletion.

```
public class DeletedAccount : User
{
    public DateTime DeletedOn { get; set; }
    public User DeletedBy { get; set; }
}
```

If you soft delete only to keep the username reserved, you could also consider filtering out deleted accounts entirely:

```
public class EntitiesContext {
    protected override void OnModelCreating(DbModelBuilder modelBuilder)
    {
        modelBuilder.Ignore<DeletedAccount>();
```

Or, if you Gallery still needs to access deleted accounts you could check with:

```
if (user is DeletedAccount)
```

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -90,6 +95,42 @@ public virtual ActionResult Account()
...
+            var listPackageItems = _packageService
+                 .FindPackagesByOwner(user, true)
+                 .Select(p => new ListPackageItemViewModel(p))
+                 .ToList();
+            var model = new DeleteUserAccountViewModel
+            {
+                Packages = listPackageItems,
+                User = user,
+                AccountName = user.Username,
+                HasOrphanPackages = listPackageItems.Where(p => p.Owners.Count <= 1).Any()
```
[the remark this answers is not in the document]
chenriksson: Is the package still orphaned if the only remaining co-owners are orphaned organization(s)?

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace NuGetGallery.Areas.Admin.Controllers
+{
+    public partial class DeleteAccountController : AdminControllerBase
+    {
+        private readonly IUserService _userService;
+
+        public DeleteAccountController(IUserService userService)
+        {
+            _userService = userService;
```
skofman1: null check

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        [HttpGet]
+        public virtual ActionResult Search(string query)
+        {
+            var results = new List<DeleteAccountSearchResult>();
+            var result = new DeleteAccountSearchResult();
+            if (query != null)
+            {
+                var user = _userService.FindByUsername(query);
```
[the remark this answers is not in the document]
skofman1: Organizations is out of scope for this work. 

---
In reply to: [148116556](https://github.com/NuGet/NuGetGallery/pull/4936#discussion_r148116556) [](ancestors = 148116556)

src/NuGetGallery/Areas/Admin/ViewModels/DeleteUserAccountViewModel.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public string AccountName { get; set; }
+
+        [Required(ErrorMessage = "Please sign using your name.")]
+        [StringLength(1000)]
+        [Display(Name = "Signature")]
+        public string Signature { get; set; }
+
+        [Display(Name = "Unlist the packages without an user.")]
+        public bool Unlist { get; set; }
```
skofman1: ShouldUnlist

skofman1:
>         [ValidateAntiForgeryToken]

 [ValidateAntiForgeryToken]

---
Refers to: src/NuGetGallery/Controllers/UsersController.cs:65 in 0a0f23e. [](commit_id = 0a0f23ec0936f3fd3ffaf1b8927ac2c2487fc6bc, deletion_comment = False)

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -90,6 +95,42 @@ public virtual ActionResult Account()
         }
 
         [HttpGet]
+        [Authorize(Roles = "Admins")]
+        public virtual ActionResult DeleteUserAccount(string accountName)
```
skofman1: [ValidateAntiForgeryToken]

src/NuGetGallery/Controllers/UsersController.cs
skofman1: can the owners count of a package be <1 before the delete?

src/NuGetGallery/Controllers/UsersController.cs
joelverhagen: Oh right, I got this confused with a logged in page.

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
joelverhagen: Er, that was a typo. What I meant is you don't need the property initializer `{ ... }` syntax. Just parens.

src/NuGetGallery.Core/Entities/AccountDelete.cs
cristinamanum: Updated.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
cristinamanum: Updated.

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
cristinamanum: updated

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteUserAccount.cshtml
cristinamanum: updated

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteUserAccount.cshtml
cristinamanum: updated

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteUserAccount.cshtml
cristinamanum: updated

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/Index.cshtml
cristinamanum: I tried it, it works but it looks worse. I let it as it is.

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: updated

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: updated

src/NuGetGallery.Core/Entities/AccountDelete.cs
cristinamanum: EF is creating the right keys. I added the keys explicitly anyway.

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
cristinamanum: updated

src/NuGetGallery/Migrations/201710302343008_AddAccountDelete.cs
cristinamanum: The migration does not fall. It inserts false.

src/NuGetGallery/App_Start/Routes.cs
cristinamanum: This is the pattern used throughout the entire file.

src/NuGetGallery/Areas/Admin/ViewModels/DeleteAccountSearchResult.cs
cristinamanum: Updated

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: The user will not be  forceably logged out?

joelverhagen: I'm not sure! We should verify this, i.e. what happens when you delete the account of a logged in user. What is the user's experience? Can they create more API keys is they are already on the API keys page? Etc.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: updated

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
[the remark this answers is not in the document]
cristinamanum: Updated to not return deleted users.

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/Index.cshtml
```diff
@@ -0,0 +1,67 @@
...
+                        })
+                        .error(function(jqXhr, textStatus, errorThrown) {
+                            alert("Error: " + errorThrown);
+                        })
+                        .always(function () {
+                            $self.doneSearching(true);
+                        });
+                };
+                this.generateProfileUrl = function (u) {
+                    return '/profiles/' + u.AccountName;
```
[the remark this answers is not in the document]
cristinamanum: We use a similar pattern for packages delete scenario.(Areas\Admin\Views\Delete\Index.cshtml)

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: Updated to handle these cases. Good catch.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,213 @@
...
+using System.Threading.Tasks;
+using NuGetGallery.Auditing;
+using NuGetGallery.Authentication;
+using NuGetGallery.Security;
+
+namespace NuGetGallery
+{
+    public class DeleteAccountService : IDeleteAccountService
+    {
+        private readonly IEntityRepository<Package> _packageRepository;
```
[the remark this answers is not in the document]
cristinamanum: Removed the _packagesRepository. _entitiesContext it is used now to delete data as a single transaction.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Updated to handle this case.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,213 @@
...
+                                            await InsertDeleteAccount(useToBeDeleted, admin, signature, executionDetails)
+                                         : false)
+                                     : false)
+                                 : false)
+                             : false)
+                         : false;
+            return new Tuple<bool, List<string>>(result, executionDetails);
+        }
+
+        private async Task<bool> InsertDeleteAccount(User user, User admin, string signature, List<string> executionDetails)
```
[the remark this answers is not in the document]
cristinamanum: All of the above were updated to perform the changes as a transaction.

src/NuGetGallery.Core/Entities/AccountDelete.cs
cristinamanum: I will keep the current design. The Deleted Account is not an user, it is an association of the admin - deleted account .

src/NuGetGallery.Core/Entities/User.cs
```diff
@@ -137,5 +139,18 @@ public bool IsInRole(string roleName)
...
+            EmailAddress = null;
+            UnconfirmedEmailAddress = null;
+            EmailAllowed = false;
+            EmailConfirmationToken = null;
+            PasswordResetToken = null;
+            NotifyPackagePushed = false;
+            LastFailedLoginUtc = null;
+            FailedLoginCount = 0;
+            IsDeleted = true;
+        }
```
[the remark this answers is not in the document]
cristinamanum: Updated.

src/NuGetGallery/Areas/Admin/ViewModels/DeleteUserAccountViewModel.cs
```diff
@@ -0,0 +1,33 @@
...
+        public DeleteUserAccountViewModel()
+        {
+            Unlist = true;
+        }
+
+        public List<ListPackageItemViewModel> Packages { get; set; }
+
+        public User User { get; set; }
+
+        public string AccountName { get; set; }
```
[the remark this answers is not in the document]
cristinamanum: AccountName is used to pass back the account info from the view.

src/NuGetGallery/Areas/Admin/ViewModels/DeleteUserAccountViewModel.cs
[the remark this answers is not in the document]
cristinamanum: Updated.

src/NuGetGallery/Controllers/UsersController.cs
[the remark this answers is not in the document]
cristinamanum: Currently the package will not be orphaned in the above case.

src/NuGetGallery/Services/DeleteAccountService.cs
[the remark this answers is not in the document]
cristinamanum: They will be unverified. No - the co-owners of the packages will need to take ownership of the namespace.

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
cristinamanum: Updated.

src/NuGetGallery/Areas/Admin/ViewModels/DeleteUserAccountViewModel.cs
cristinamanum: Updated

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: Short answer:Yes
Long answer: from data point of view yes. From the application point of view, the scenarios was blocked for the last couple of years. With this change it will open again. This is related with the observation in the PR comments.

src/NuGetGallery/Controllers/UsersController.cs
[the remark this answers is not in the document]
cristinamanum: Updated

src/NuGetGallery/App_Start/Routes.cs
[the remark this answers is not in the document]
cristinamanum: Updated.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Updated. This code was refactored but the suggestion was implemented.

src/NuGetGallery/Views/Users/Profiles.cshtml
cristinamanum: updated

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: The code was refactored to have the data updates as a single transaction.

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/Index.cshtml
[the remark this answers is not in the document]
scottbommarito: We should make sure this works with APIM.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -362,11 +362,7 @@ where VersionRange.Parse(d.VersionSpec).Satisfies(packageVersion)
 
         public async Task RemovePackageOwnerAsync(PackageRegistration package, User user)
         {
-            if (package.Owners.Count == 1 && user == package.Owners.Single())
-            {
-                throw new InvalidOperationException("You can't remove the only owner from a package.");
-            }
-
+            //it is allowed to remove the last owner of a package
```
scottbommarito: This check needs to be moved to somewhere such that it gets checked by `JsonApiController.RemovePackageOwner` when users try to remove package owners.

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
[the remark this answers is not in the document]
chenriksson: If organizations are out of scope, DeleteAccountController.Search should filter out organization accounts.

src/NuGetGallery/Areas/Admin/Controllers/DeleteAccountController.cs
```diff
@@ -13,36 +14,30 @@ public partial class DeleteAccountController : AdminControllerBase
...
         public virtual ActionResult Search(string query)
         {
             var results = new List<DeleteAccountSearchResult>();
-            var result = new DeleteAccountSearchResult();
-            if (query != null)
+            if (!string.IsNullOrWhiteSpace(query))
             {
                 var user = _userService.FindByUsername(query);
-                if (user != null)
+                if (user !=  null && user.Username != null && !user.IsDeleted)
```
chenriksson: Why would user.Username be null?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -96,38 +96,50 @@ public virtual ActionResult Account()
...
+        [RequiresAccountConfirmation("Delete account")]
         [ValidateAntiForgeryToken]
-        public virtual async Task<ActionResult> DeleteUserAccount(DeleteUserAccountViewModel model)
+        public virtual async Task<ActionResult> Delete(DeleteUserAccountViewModel model)
         {
             var user = _userService.FindByUsername(model.AccountName);
-            var admin = GetCurrentUser();
-            var status = await _deleteAccountService.DeleteGalleryUserAccountAsync(user, admin, model.Signature, model.Unlist);
-            return  View("DeleteUserAccountStatus", new DeleteUserAccountStatusViewModel() {AccountName = model.AccountName , OperationStatus = status .Item2});
+            if (user == null || user.IsDeleted)
```
chenriksson: Per Organizations comment, would also ensure user isn't organization here.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -7,62 +7,57 @@
 using NuGetGallery.Auditing;
 using NuGetGallery.Authentication;
 using NuGetGallery.Security;
+using NuGetGallery.Areas.Admin.ViewModels;
```
chenriksson: nit: sort

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -184,7 +184,7 @@ public virtual ActionResult GetAddPackageOwnerConfirmation(string id, string use
 
                 if (request == null)
                 {
-                    await _packageOwnershipManagementService.RemovePackageOwnerAsync(model.Package, model.CurrentUser, model.User);
+                    await _packageOwnershipManagementService.RemovePackageOwnerAsync(model.Package, model.CurrentUser, model.User, commitAsTransaction:true);
```
skofman1: nit: space before true

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -90,6 +95,54 @@ public virtual ActionResult Account()
...
+                Packages = listPackageItems,
+                User = user,
+                AccountName = user.Username,
+                HasOrphanPackages = listPackageItems.Any(p => p.Owners.Count <= 1)
+            };
+            return View("DeleteUserAccount", model);
+        }
+
+        [Authorize(Roles = "Admins")]
+        [HttpPost]
```
skofman1: >HttpPost [](start = 9, length = 8)

Should the HTTP verb here be DELETE?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,182 @@
...
+            if (admin == null)
+            {
+                throw new ArgumentNullException(nameof(admin));
+            }
+            if(useToBeDeleted.IsDeleted)
+            {
+                return new DeleteUserAccountStatus()
+                {
+                    Success = false,
+                    Description = $"The account:{useToBeDeleted.Username} was already deleted. No action was performed.",
```
skofman1: if this is a user facing string it should be in Strings

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,182 @@
...
+                    }
+                }
+                else
+                {
+                    await DeleteGalleryUserAccountImplAsync(useToBeDeleted, admin, signature, unlistOrphanPackages);
+                }
+                return new DeleteUserAccountStatus()
+                {
+                    Success = true,
+                    Description = $"The account:{useToBeDeleted.Username} was deleted succesfully.",
```
skofman1: same

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// 4. The user credentials will be cleaned.
+        /// 5. The user data will be cleaned.
+        /// </summary>
+        /// <param name="useToBeDeleted">The user to be deleted.</param>
+        /// <param name="admin">The admin that will perform the delete action.</param>
+        /// <param name="signature">The admin signature.</param>
+        /// <param name="unlistOrphanPackages">If the orphaned packages will unlisted.</param>
+        /// <param name="commitAsTransaction">If the data will be persisted as a transaction.</param>
+        /// <returns></returns>
+        public async Task<DeleteUserAccountStatus> DeleteGalleryUserAccountAsync(User useToBeDeleted, User admin, string signature, bool unlistOrphanPackages, bool commitAsTransaction)
```
skofman1: The only usage I found of DeleteGalleryUserAccountAsync is in UsersController and commitAsTransaction is true. Why is this param needed?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,182 @@
...
+                    await DeleteGalleryUserAccountImplAsync(useToBeDeleted, admin, signature, unlistOrphanPackages);
+                }
+                return new DeleteUserAccountStatus()
+                {
+                    Success = true,
+                    Description = $"The account:{useToBeDeleted.Username} was deleted succesfully.",
+                    AccountName = useToBeDeleted.Username
+                };
+            }
+            catch(Exception e)
```
skofman1: The exception caught is presented as is to the user and is not logged anywhere. We should log the exception and present a general error message to the user.

src/NuGetGallery/Services/PackageService.cs
skofman1: nit: comments should start with upper case and a space after // 
This is something style cop used to enforce, and most comments in the code base look like this :)

cristinamanum: Updated

cristinamanum: Updated

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Updated

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Updated

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: There are 2 patterns currently used to commit data to the underling EF storage.
1. SaveAsync
2. Grouping multiple SaveAsync in a transaction.

Some of the services implement APIs that commit changes to db. If any of this api is wrapped in a transaction it cannot be reused by other service call to commit data as a single transaction. The usercontrollers need to commit the delete data in transaction, but other service may need to use the same api call and do the commit in their own transaction and they invoke this call with commitAsTransaction = false.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: I agree that this is a good practice in most of the cases. In the current case this is an Admin action; having the exception made intermediately available to the admin would the most beneficial in my opinion. Saving the exception data I do not think that will have any benefit as the action is executed by an human(admin) and not a tool. The admin needs to take the action as he gets the error information.  In addition saving the exception will need to save the user account information that will need to be cleaned after as it may contain sensitive data.

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: Makes sense - updated.

src/NuGetGallery/Services/DeleteAccountService.cs
skofman1: Can't we log the exception without logging sensitive information? The error here can he anything from DB timeout to out of memory..

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/Index.cshtml
```diff
@@ -0,0 +1,67 @@
...
+                <li data-bind="foreach: searchResults">
+                    <a href="#" data-bind="text: AccountName, attr: { href: $parent.generateProfileUrl($data) }"></a>
+                </li>
+            </ul>
+        </form>
+    </article>
+</section>
+
+@section BottomScripts {
+    <script src="@Url.Content("~/Scripts/knockout-2.2.1.js")"></script>
```
chenriksson: nit, consider `@Scripts.Render("~/Scripts/gallery/site.min.js")` which includes ko v3.4.2.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -90,6 +95,54 @@ public virtual ActionResult Account()
...
+            var listPackageItems = _packageService
+                 .FindPackagesByOwner(user, includeUnlisted:true)
+                 .Select(p => new ListPackageItemViewModel(p))
+                 .ToList();
+            var model = new DeleteUserAccountViewModel
+            {
+                Packages = listPackageItems,
+                User = user,
+                AccountName = user.Username,
+                HasOrphanPackages = listPackageItems.Any(p => p.Owners.Count <= 1)
```
chenriksson: Do/should we show the orphaned packages w/ download counts on the Admin view, for review by the admin?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -90,6 +95,54 @@ public virtual ActionResult Account()
...
+            };
+            return View("DeleteUserAccount", model);
+        }
+
+        [Authorize(Roles = "Admins")]
+        [HttpDelete]
```
chenriksson: nit: is verb attribute usually first?

src/NuGetGallery/RequestModels/DeleteAccountRequest.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    public class DeleteAccountRequest
+    {
+        public string AccountName {get; set; }
```
chenriksson: Do you need a request model for a single string value?

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: The user would not be able to create apis keys, upload packages etc. As discussed offline it would be a nice feature to have. To track it, I added issue: https://github.com/NuGet/NuGetGallery/issues/4966

src/NuGetGallery/Controllers/UsersController.cs

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Updated

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,181 @@
...
+                    Success = false,
+                    Description = string.Format(Strings.AccountDelete_AccountAlreadyDeleted, userToBeDeleted.Username),
+                    AccountName = userToBeDeleted.Username
+                };
+            }
+            try
+            {
+                if (commitAsTransaction)
+                {
+                    using (var strategy = new SuspendDbExecutionStrategy())
```
chenriksson: nit, would wrap SuspendDbExecutionStrategy in curly braces so someone doesn't mistakenly change its scope

src/NuGetGallery/Views/Users/_UserPackagesListForDeletedAccount.cshtml
```diff
@@ -0,0 +1,59 @@
...
+                    {
+                        <tr class="manage-package-listing">
+                            <td class="align-middle ">
+                                @if (package.Owners.Count() > 1)
+                                {
+                                    <i  class="ms-Icon ms-Icon--CompletedSolid green-check" title="After account delete the package will still at least one owner."></i>
+                                }
+                                else 
+                                {
+                                    <i  class="ms-Icon ms-Icon--StatusErrorFull red-check" title="After account delete the package will not have any owner."></i>
```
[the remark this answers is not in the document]
cristinamanum: updated

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: Updated

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            var accountDelete = new AccountDelete
+            {
+                DeletedOn = DateTime.UtcNow,
+                DeletedAccountKey = user.Key,
+                DeletedByKey = admin.Key,
+                Signature = signature
+            };
+            _accountDeleteRepository.InsertOnCommit(accountDelete);
+            await _accountDeleteRepository.CommitChangesAsync();
```
chenriksson: Does CommitChangesAsync done inside the BeginTransaction block not commit until the transaction is committed?

Other places in code pass argument for commitChanges, is that a better alternative? Note that SuspendDbExecutionStrategy means you lose retry capabilities... which maybe isn't a big deal for admin actions.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Updated

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        private async Task RemoveUserCredentials(User user)
+        {
+            var copyOfUserCred = user.Credentials.ToArray();
+            foreach (var uc in copyOfUserCred)
+            {
+                await _authService.RemoveCredential(user, uc);
```
chenriksson: RemoveUserCredentials, RemoveSecurityPolicies, RemoveReservedNamespaces, etc. will all trigger auditing, right?

If DeleteAccount also needs to clean up auditing records, I wonder if these should be done w/o auditing and a single audit record w/o PII should be created for the account being deleted. Thoughts?

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
```diff
@@ -116,41 +116,18 @@ public IEnumerable<PackageOwnerRequest> GetPackageOwnershipRequests(PackageRegis
...
-                using (var strategy = new SuspendDbExecutionStrategy())
-                using (var transaction = _entitiesContext.GetDatabase().BeginTransaction())
+                if (commitAsTransaction)
                 {
-                    // 1. Remove this package registration from the namespaces owned by this user if he is the only package owner in the set of matching namespaces
-                    // 2. Remove the IsVerified flag from package registration if all the matching namespaces are owned by this user alone (no other package owner owns a matching namespace for this PR)
-                    var allMatchingNamespacesForRegistration = packageRegistration.ReservedNamespaces;
-                    if (allMatchingNamespacesForRegistration.Any())
+                    using (var strategy = new SuspendDbExecutionStrategy())
+                    using (var transaction = _entitiesContext.GetDatabase().BeginTransaction())
```
chenriksson: We should be careful about using SuspendDbExecutionStrategy since it prevents retry (AFAIK). Is it necessary here, and if so we should add comment as to why.

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -176,51 +176,62 @@ public class ReservedNamespaceService : IReservedNamespaceService
             {
                 throw new ArgumentException(Strings.ReservedNamespace_InvalidUsername);
             }
-
-            using (var strategy = new SuspendDbExecutionStrategy())
-            using (var transaction = EntitiesContext.GetDatabase().BeginTransaction())
+            if (commitAsTransaction)
```
scottbommarito: You could probably make a helper function and remove this snippet of code from everywhere that has it.

```
public static async Task CommitAsTransaction(EntitiesContext entitiesContext, Func<Task> taskFactory, bool commitAsTransaction)
{
    if (commitAsTransaction)
    {
        using (var strategy = new SuspendDbExecutionStrategy())
        using (var transaction = EntitiesContext.GetDatabase().BeginTransaction())
        {
            await TaskFactory();
        }
    }
    else
    {
        await TaskFactory();
    }
}
```

```
public async Task DeleteOwnerFromReservedNamespaceAsync(string prefix, string username, bool commitAsTransaction = true)
{
    await CommitAsTransaction(EntitiesContext, () => DeleteOwnerFromReservedNamespaceImplAsync(prefix, username), commitAsTransaction);
}
```

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
cristinamanum: This is out of scope for this PR as it is existent code.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: The audit records clean up will be a different task.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: CommitChangeAsync is actually SaveAsync so it will not commit until transaction.commit.

src/NuGetGallery/RequestModels/DeleteAccountRequest.cs
cristinamanum: I think is clean this way.

src/NuGetGallery/Controllers/UsersController.cs

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: If needed i propose to do it in a different PR. We do not have any input if we need or not.

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/Index.cshtml
cristinamanum: If needed i propose to be done in a different PR.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: This is the pattern used in other apis.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Updated.

Issue #4549, opened 2017-08-11
[Prefix] Add auditing for Reserved namespaces


shishirx34:
Application changes and shim changes. Test as well. 

- [x] Gallery changes
- [x] Shim changes
- [x] Validate locally

skofman1:
We should also audit failed package pushes due to namespace conflict.

PR #4940, opened 2017-11-01
[Prefix] Add auditing for reserved namespaces

shishirx34:
I will send out the Shim changes once I get a build out of this merge, since it requires the changes from `NuGetGallery.Core` for the new audit record types and actions. Addresses #4549 first part.


src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
```diff
@@ -15,6 +15,7 @@ public enum AuditedPackageAction
         Edit,
         [Obsolete("Undo package edit functionality is being retired.")]
         UndoEdit,
-        Verify
+        Verify,
+        UploadFailedNamespaceConflict
```
chenriksson: We don't audit upload failures due to id/title conflict, so I don't think we should with namespace either. The only failures we currently audit are authentication related (package push attempt by non-owner, login failed).

src/NuGetGallery.Core/Auditing/AuditedPackageRegistrationAction.cs
```diff
@@ -6,6 +6,8 @@ namespace NuGetGallery.Auditing
     public enum AuditedPackageRegistrationAction
     {
         AddOwner,
-        RemoveOwner
+        RemoveOwner,
+        MarkVerified,
+        MarkUnverified
```
chenriksson: I don't think you need separate actions for MarkVerified/MarkUnverified. AuditedPackageRegistration.IsVerified will be logged when the action that caused verification is audited (i.e., package owner added, or push of new package).

src/NuGetGallery.Core/Auditing/AuditedReservedNamespaceAction.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Auditing
+{
+    public enum AuditedReservednamespaceAction
+    {
+        Allocate,
+        Deallocate,
```
chenriksson: Would rename to [Reserve/Unreserve]Namespace or [Add/Remove]NamespaceReservation, which seems more specific.

src/NuGetGallery.Core/Auditing/AuditedReservedNamespaceAction.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Auditing
+{
+    public enum AuditedReservednamespaceAction
```
chenriksson: Casing - should be AuditedReservedNamespaceAction

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedPackageRegistration.cs
```diff
@@ -8,14 +8,16 @@ public class AuditedPackageRegistration
...
 
         public static AuditedPackageRegistration CreateFrom(PackageRegistration packageRegistration)
         {
             return new AuditedPackageRegistration
             {
                 Id = packageRegistration.Id,
                 DownloadCount = packageRegistration.DownloadCount,
-                Key = packageRegistration.Key
+                Key = packageRegistration.Key,
+                IsVerified = packageRegistration.IsVerified
```
joelverhagen: nit trailing comma so that future diffs only change one line

src/NuGetGallery.Core/Auditing/PackageAuditRecord.cs
```diff
@@ -30,6 +30,9 @@ public class PackageAuditRecord : AuditRecord<AuditedPackageAction>
             Reason = reason;
         }
 
+        public PackageAuditRecord(string id, string version, AuditedPackageAction action, string reason)
+            : this(id, version, "", null, null, action, reason) { }
```
joelverhagen: name ambiguous parameters

src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
shishirx34: This is as per @skofman1's comment in #4549

src/NuGetGallery.Core/Auditing/AuditedPackageRegistrationAction.cs
shishirx34: True, will remove this.

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedReservedNamespace.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace NuGetGallery.Auditing.AuditedEntities
+{
+    public class AuditedReservedNamespace
+    {
+        public string Value { get; }
+        public bool IsPrefix { get; }
+        public bool IsSharedNamespace { get; }
+
+        public AuditedReservedNamespace(ReservedNamespace reservedNamespace)
```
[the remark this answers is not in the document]
shishirx34: I didn't see consistency here, `AuditedUserSecurityPolicy` doesn't use `CreateFrom` uses the ctor to create audit record.

I don't mind making this to use `CreateFrom` instead.

src/NuGetGallery.Core/Auditing/AuditedPackageRegistrationAction.cs
shishirx34: On second thoughts, I think we should have these actions for `AffectedRegistrations` in the reserved namespace audit, the registration records are updated for marking verified/unverified actions, which cannot be said of using `(Add/Remove)Owner` see [this](https://github.com/NuGet/NuGetGallery/pull/4940/files/6c422da7e4887d087f9f0f06aa7e5cdfc47512ac#diff-05d1e3e185bcf9424c2eb99ae86d0920R48)

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedPackageRegistration.cs
shishirx34: Putting a trailing comma on the last property of an object seems weird to me. Ticks my OCD :)

src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
shishirx34: changed this to telemetry value rather than auditing.

shishirx34:
@chenriksson - addressed feedback.

src/NuGetGallery.Core/Auditing/AuditedPackageRegistrationAction.cs
chenriksson: Good point. I was thinking the list of packages that became verified could be audited as part of the register/unregister namespace. This is probably better since the audit path will be more consistent for when packages become un-/verified.

If we want to see when a package became verified, regardless of how, do we just look under registration auditing (auditing/{package-id})?

src/NuGetGallery.Core/Auditing/ReservedNamespaceAuditRecord.cs
```diff
@@ -0,0 +1,72 @@
...
+        private static AuditedPackageRegistrationAction? GetPackageRegistrationAction(AuditedReservedNamespaceAction action)
+        {
+            switch (action)
+            {
+                case AuditedReservedNamespaceAction.AddOwner:
+                    return AuditedPackageRegistrationAction.MarkVerified;
+                case AuditedReservedNamespaceAction.RemoveOwner:
+                    return AuditedPackageRegistrationAction.MarkUnverified;
+                default:
+                    return null;
```
chenriksson: This default case shouldn't happen... I would have it throw instead of complicating the code to work with nullables.

src/NuGetGallery.Core/Auditing/ReservedNamespaceAuditRecord.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            Value = reservedNamespace.Value;
+            AffectedReservedNamespace = AuditedReservedNamespace.CreateFrom(reservedNamespace);
+            Action = action;
+
+            if (!string.IsNullOrWhiteSpace(username))
+            {
+                AffectedOwner = username;
+
+                var registrationAction = GetPackageRegistrationAction(action);
```
chenriksson: No need to call GetPackageRegistrationAction if registrations is null/empty

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -216,6 +203,30 @@ private static string GetApiKeyCreationDate(User user, IIdentity identity)
             return apiKey?.Created.ToString("O") ?? "N/A";
         }
 
+        private void TrackPackageForEvent(string eventValue, string packageId, string packageVersion, User user, IIdentity identity)
```
chenriksson: If you add AuthenticationMethod, would also match TrackPackagePushEvent. Maybe worth doing?

tests/NuGetGallery.Core.Facts/Auditing/ReservedNamespaceAuditRecordTests.cs
```diff
@@ -0,0 +1,53 @@
...
+            Assert.Equal(prefix.Value, record.Value);
+            Assert.NotNull(record.AffectedReservedNamespace);
+            Assert.NotNull(record.AffectedRegistrations);
+            Assert.NotNull(record.AffectedOwner);
+            Assert.Equal(prefix.Value, record.AffectedReservedNamespace.Value);
+            Assert.Equal(prefix.IsSharedNamespace, record.AffectedReservedNamespace.IsSharedNamespace);
+            Assert.Equal(prefix.IsPrefix, record.AffectedReservedNamespace.IsPrefix);
+            Assert.Equal(AuditedReservedNamespaceAction.AddOwner, record.Action);
+            Assert.Equal(registrationsList.Count, record.AffectedRegistrations.Length);
+            Assert.Equal(owner.Username, record.AffectedOwner);
```
chenriksson: nit: Arrange/Act/Assert comments would be useful in a long test like this

tests/NuGetGallery.Facts/Services/ReservedNamespaceServiceFacts.cs
```diff
@@ -2,12 +2,14 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Moq;
-using NuGetGallery.TestUtils;
+using Xunit;
```
chenriksson: nit: sort

src/NuGetGallery.Core/Auditing/ReservedNamespaceAuditRecord.cs
shishirx34: Will throw and add tests.

PR #4945, opened 2017-11-02
Organization scope evaluation

chenriksson:
Changes include:
- ApiKeyAuthHandler updated to verify authenticated user matches the owner scope
- ScopeEvaluator split into scope extensions which can evaluate individual scope parts
- ApiController.CreatePackageVerificationKey creates temp key using same owner scope as original key
- ApiController no longer calls PackagePermissionService, which was redundant with scope evaluation
- ExtensionMethods refactoring, extracting APIs as I created more specific extensions classes


src/NuGetGallery.Core/Entities/Organization.cs
```diff
@@ -15,12 +15,13 @@ namespace NuGetGallery
     /// <see href="https://weblogs.asp.net/manavi/inheritance-mapping-strategies-with-entity-framework-code-first-ctp5-part-2-table-per-type-tpt" />
     public class Organization : User
     {
-        public Organization() : base()
+        public Organization() : this(null)
```
joelverhagen: nit: name null param

src/NuGetGallery/Extensions/ScopeExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                throw new ArgumentNullException(nameof(scope));
+            }
+
+            if (string.IsNullOrEmpty(subject))
+            {
+                throw new ArgumentNullException(nameof(subject));
+            }
+
+            return new Regex("^" + Regex.Escape(scope.Subject).Replace(@"\*", ".*") + "$",
```
joelverhagen: We need a timeout on this regex right?

src/NuGetGallery/Authentication/Providers/ApiKey/ApiKeyAuthenticationHandler.cs
```diff
@@ -81,21 +81,30 @@ internal void WriteStatus(string message, int statusCode)
                 var authUser = await Auth.Authenticate(apiKey);
                 if (authUser != null)
                 {
+                    var credential = authUser.CredentialUsed;
+                    var user = authUser.User;
+
+                    // Ensure that the user matches the owner scope
+                    if (!user.MatchesOwnerScope(credential))
+                    {
+                        WriteStatus(Strings.ApiKeyNotAuthorized, 403);
```
cristinamanum: Instead of 403 can we use the HttpStatusCode values?

src/NuGetGallery/Infrastructure/Authentication/CredentialBuilder.cs
```diff
@@ -26,14 +27,28 @@ public Credential CreateApiKey(TimeSpan? expiration)
...
 
-            credential.Scopes.Add(new Scope(subject: id, allowedAction: NuGetScopes.PackageVerify));
+            var ownerKeys = originalApiKey.Scopes
+                .Select(s => s.OwnerKey)
+                .Distinct().ToArray();
+
+            if (ownerKeys.Length == 0)
+            {
+                // Legacy API key with no owner scope.
+                credential.Scopes = new[] { new Scope(null, subject: id, allowedAction: NuGetScopes.PackageVerify) };
```
joelverhagen: nit name nulls

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -291,23 +293,18 @@ public ActionResult HealthProbe()
             await AuditingService.SaveAuditRecordAsync(
                 new PackageAuditRecord(package, AuditedPackageAction.Verify));
 
-            if (!PackagePermissionsService.IsActionAllowed(package, user, PackageAction.UploadNewVersion))
-            {
-                return new HttpStatusCodeWithBodyResult(HttpStatusCode.Forbidden, Strings.ApiKeyNotAuthorized);
-            }
-
```
cristinamanum: Why this check is not needed?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -742,5 +704,43 @@ protected internal virtual Stream ReadPackageFromRequest()
 
             return new HttpStatusCodeResult(HttpStatusCode.NotFound);
         }
+
+        private bool HasScopeThatAllows(PackageRegistration package, params string[] requestedActions)
```
joelverhagen: The function name should state whether it's `Any` or `All` of the requested actions

src/NuGetGallery.Core/Entities/Organization.cs
chenriksson: I kept the same pattern as the User base class... but agree it's icky. I'll switch back to base(), but then will need to instantiate Members in this ctor as well.

src/NuGetGallery/Authentication/Providers/ApiKey/ApiKeyAuthenticationHandler.cs
chenriksson: Our controller APIs use HttpStatusCode since MVC's ActionResult classes support it. This uses OwinResponse, which does not. Agree it would improve readability, but not sure it's worth doing when I'd need to cast back to int.

src/NuGetGallery/Controllers/ApiController.cs
chenriksson: The PackagePermissionService calls were originally calls to check `package.IsOwner(user)`. This change resulted in not only checking owner, but also checking the action which was already being verified with the `ApiKeyScopeAllows` call below as well as the `ApiScopeRequired` attributes.

Now that API key scopes have an owner scope, the original `package.IsOwner` is evaluated inside the `ApiKeyScopeAllows` / `HasScopeThatAllows` call.

src/NuGetGallery/Extensions/ScopeExtensions.cs
chenriksson: This is pre-existing regex, and like others it didn't have timeout. I'll add it since I'm touching the code.

src/NuGetGallery/Controllers/ApiController.cs
chenriksson: Related note, I am going to update HasAnyScopeThatAllows to also check if user is still owner of package when there is an owner scope.  This will handle case where user/org was removed as package owner but we didn't revoke the api key scope for the package.

src/NuGetGallery/Authentication/Providers/ApiKey/ApiKeyAuthenticationHandler.cs
scottbommarito:
```
(int)HttpStatusCode.Forbidden
```

should be perfectly fine

Issue #3417, opened 2016-12-13
[Statistics] Introduce Community Statistics Reports


joelverhagen:
Thanks @emgarten, for bringing this up.

Look here: https://www.nuget.org/stats

![image](https://cloud.githubusercontent.com/assets/94054/21124070/86809376-c091-11e6-9c5c-909d2f234bdb.png)

It's not interesting that `System.*` packages are being downloaded a lot. They show up in templates so it does not help new cool packages catch on to crowd them out with BCL-ish packages.

We should filter out `System.*` packages by Microsoft or perhaps the closure of the `Microsoft.NETCore.App`.

xavierdecoster:
This is a known issue, and a non-trivial fix to get it right, but thanks for bringing it up again!
Anyone can own a `System.*` or `Microsoft.*` package today. The thing is, the stats database only has the W3C logs as input, which does not contain ownership data, and today there is no link to correlate between stats data and gallery data when these stats report blobs are generated. 

We do have the concept of `PackageSets` in the stats db which is a manually maintained list of package sets we treat specially (e.g. we could leverage this to filter certain packages in this report view).

skofman1:
Since a full solution is complex, lets update  `PackageSets` in the stats db with the packages above, like @xavierdecoster suggested.

joelverhagen:
This point has been brought up on Twitter (https://twitter.com/CodeScottLe/status/887796340029247488).

As a first "easy" step, @jonwchu and I think we should try excluding anything owned the [microsoft](https://preview.nuget.org/profiles/microsoft) or [aspnet](https://preview.nuget.org/profiles/aspnet). 

If this remaining set is too small, we should remove the `NETStandard.Library` closure.

joelverhagen:
Additional feedback from the community. @anangaur posted on https://github.com/NuGet/NuGetGallery/issues/4836:

> https://twitter.com/chrismckee/status/918275305835585536
> 
> On https://www.nuget.org/stats/packages, NuGet.org should show only the directly referenced packages that users add to their projects and not System.*, Microsoft.* that automatically get pulled in.

anangaur:
IMO the right fix is to be able to distinguish between "Install" action vs. download which may be update/restore/install. And use the install count as the metric to list top packages. (Or even in search ranking). 

However, any quick fix as suggested above should be good too, in the interim.

skofman1:
We can start with removing system.*  and runtime.native.*, and perhaps a few other prefixes.

[...]

PR #4955, opened 2017-11-03
[Statistics] Introduce Community Statistics Reports

loic-sharma:
# Introduce Community Statistics Reports

This change introduces the concept of "community" statistics report that will be displayed by default. These reports filter out packages owned by Microsoft and .NET Framework. Users can choose to "Show all packages" to view reports that include Microsoft's and .NET Framework's packages. With this change, packages like `System.Globalization` will no longer clutter our default statistic reports! 😄 

Note: The `JsonStatisticsService` had some serious concurrency issue and did not cache reports that it loaded from blob storage. This pull request fixes that by loading reports once an hour.

This change depends on https://github.com/NuGet/NuGet.Jobs/pull/239.
Fixes https://github.com/NuGet/NuGetGallery/issues/3417.

# Screenshots

## Statistics Homepage

one

## Detailed Report

Clicking on the "Show more..." links at the bottom of the homepage will take you to the following reports:

### Most Downloaded Packages

most-downloaded-packages

### Most Downloaded Package Versions

most-downloaded-package-versions

### DIFF
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -666,9 +666,9 @@ namespace NuGetGallery
         [ActionName("StatisticsDownloadsApi")]
         public virtual async Task<ActionResult> GetStatsDownloads(int? count)
         {
-            var result = await StatisticsService.LoadDownloadPackageVersions();
+            await StatisticsService.Refresh();
 
-            if (result.Loaded)
+            if (StatisticsService.DownloadPackageVersionsResult.Loaded)
             {
                 int i = 0;
 
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -668,12 +668,12 @@ namespace NuGetGallery
         {
             await StatisticsService.Refresh();
 
-            if (StatisticsService.DownloadPackageVersionsResult.Loaded)
+            if (StatisticsService.PackageVersionDownloadsResult.IsLoaded)
             {
                 int i = 0;
 
                 JArray content = new JArray();
-                foreach (StatisticsPackagesItemViewModel row in StatisticsService.DownloadPackageVersionsAll)
+                foreach (StatisticsPackagesItemViewModel row in StatisticsService.PackageVersionDownloads)
                 {
                     JObject item = new JObject();
 
--- a/src/NuGetGallery/Controllers/StatisticsController.cs
+++ b/src/NuGetGallery/Controllers/StatisticsController.cs
@@ -79,16 +79,16 @@ namespace NuGetGallery
 
             var model = new StatisticsPackagesViewModel
             {
-                IsDownloadPackageAvailable = _statisticsService.DownloadCommunityPackagesResult.Loaded,
-                DownloadPackagesSummary = _statisticsService.DownloadCommunityPackagesSummary,
+                IsDownloadPackageAvailable = _statisticsService.CommunityPackageDownloadsResult.IsLoaded,
+                DownloadPackagesSummary = _statisticsService.CommunityPackageDownloadsSummary,
 
-                IsDownloadPackageVersionsAvailable = _statisticsService.DownloadCommunityPackageVersionsResult.Loaded,
-                DownloadPackageVersionsSummary = _statisticsService.DownloadCommunityPackageVersionsSummary,
+                IsDownloadPackageVersionsAvailable = _statisticsService.CommunityPackageVersionDownloadsResult.IsLoaded,
+                DownloadPackageVersionsSummary = _statisticsService.CommunityPackageVersionDownloadsSummary,
 
-                IsNuGetClientVersionAvailable = _statisticsService.NuGetClientVersionResult.Loaded,
+                IsNuGetClientVersionAvailable = _statisticsService.NuGetClientVersionResult.IsLoaded,
                 NuGetClientVersion = _statisticsService.NuGetClientVersion,
 
-                IsLast6WeeksAvailable = _statisticsService.Last6WeeksResult.Loaded,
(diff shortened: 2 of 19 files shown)


anangaur:
I dont think we should have the filter on the https://www.nuget.org/stats page. The filter should only be there on the details pages:
https://www.nuget.org/stats/packageversions
https://www.nuget.org/stats/packages

anangaur:
Something like this:

image

loic-sharma:
Added a filter to both index and detailed pages. I will update the screenshots on Monday.

src/NuGetGallery/Services/JsonStatisticsService.cs
```diff
@@ -109,39 +173,58 @@ public IEnumerable<StatisticsWeeklyUsageItem> Last6Weeks
             }
         }
 
-        public async Task<StatisticsReportResult> LoadDownloadPackageVersions()
+        private Task<StatisticsReportResult> LoadDownloadPackageVersions()
+        {
+            var statisticsPackagesItemViewModels = (List<StatisticsPackagesItemViewModel>)DownloadPackageVersionsAll;
```
loic-sharma: TODO: Delete this.

src/NuGetGallery/Services/StatisticsReportName.cs
```diff
@@ -18,6 +18,21 @@ public enum StatisticsReportName
...
+        /// <summary>
+        /// Most frequently downloaded community package, specific to actual version.
+        /// This excludes Microsoft and ASP.NET packages.
+        /// </summary>
+        RecentCommunityPopularityDetail,
+        /// <summary>
+        /// Breakout by version for a community package (drill down from RecentPopularity).
+        /// This excludes Microsoft and ASP.NET packages.
+        /// </summary>
+        RecentCommunityPopularityDetail_,
```
loic-sharma: TODO: This is not needed, remove.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -704,9 +704,9 @@ protected internal virtual Stream ReadPackageFromRequest()
         [ActionName("StatisticsDownloadsApi")]
         public virtual async Task<ActionResult> GetStatsDownloads(int? count)
         {
-            var result = await StatisticsService.LoadDownloadPackageVersions();
+            await StatisticsService.Refresh();
```
joelverhagen: Why this change?

src/NuGetGallery/Views/Statistics/Index.cshtml
```diff
@@ -16,16 +16,21 @@
...
 <section role="main" class="container main-container page-statistics-overview">
     <h1>Statistics</h1>
+
     <span class="page-subheading ms-fontSize-l">
         @Html.Partial("_LastUpdated", Model)
     </span>
-
+    <p>
+        Show all packages:
+        <input type="checkbox" data-bind="checked: showAllPackageDownloads" />
```
joelverhagen: The include prerelease checkbox on [search](https://www.nuget.org/packages) and the [per-package stats page](https://www.nuget.org/stats/packages/Newtonsoft.Json?groupby=Version&groupby=ClientName) both are linkable. Meaning clicking on them updates the URL in the address bar so people can link to the results. You should do this either with a `GET` form or `replaceState` in JavaScript.

joelverhagen:
Do users really care about the old report? Is this driven by a specific customer ask? It is unclear from the linked issue if we should provide a way to toggle between them.

anangaur:
The filter (check box) should be enabled by default to filter out the Microsoft packages as that is what users want. It serves 2 purposes:
1. Soft landing for users who are used to seeing the Microsoft packages in the list.
2. Messaging: We need to tell what is it that we show on the page :)

src/NuGetGallery/Views/Statistics/Index.cshtml
loic-sharma: Good suggestion. However, due to feedback this will no longer be linkable.

src/NuGetGallery/Controllers/ApiController.cs
loic-sharma: See the description:

> The JsonStatisticsService had some serious concurrency issue and did not cache reports that it loaded from blob storage. This pull request fixes that by loading reports once every four hours.

loic-sharma:
Addressed feedback.

src/NuGetGallery/Controllers/ApiController.cs
loic-sharma: To elaborate a little: all the `Load...` reports were merged into the `Refresh` method.

src/NuGetGallery/Controllers/StatisticsController.cs
```diff
@@ -73,27 +75,23 @@ public StatisticsController(IStatisticsService statisticsService, IAggregateStat
...
+            await _statisticsService.Refresh();
 
             var model = new StatisticsPackagesViewModel
             {
-                IsDownloadPackageAvailable = availablity[0].Loaded,
-                DownloadPackagesSummary = _statisticsService.DownloadPackagesSummary,
-                IsDownloadPackageDetailAvailable = availablity[1].Loaded,
-                DownloadPackageVersionsSummary = _statisticsService.DownloadPackageVersionsSummary,
-                IsNuGetClientVersionAvailable = availablity[2].Loaded,
+                IsDownloadPackageAvailable = _statisticsService.DownloadCommunityPackagesResult.Loaded,
```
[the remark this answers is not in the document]
xavierdecoster: Same for the `Loaded` property which currently sounds like an event: `IsLoaded`?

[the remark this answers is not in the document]
loic-sharma: Will do! :)

Issue #4960, opened 2017-11-06
delete page is outdated


karann-msft:
When a package owner clicks on "Delete package", the page displays information which is not aligned to NuGet.org policy. Moreover, it is difficult to maintain this page. Instead, it should point to our docs.

joelverhagen:
@karann-msft, can you verify this on INT?

karann-msft:
There is a bug. @ryuyu why am I seeing the highlighted text for owner support request? I remember working on this with you where this should show up only for report abuse and when the reason selected is copyright/TM infringement.

![image](https://user-images.githubusercontent.com/16904420/32674265-6681ff52-c607-11e7-908d-6269348d8a29.png)

PR #4961, opened 2017-11-06
Updating package delete page

karann-msft:
Fixing the package delete page to remove outdated and potentially misleading content and adding a pointer to docs which is significantly easier to maintain. Fixing issue https://github.com/NuGet/NuGetGallery/issues/4960

### DIFF
--- a/src/NuGetGallery/Views/Packages/Delete.cshtml
+++ b/src/NuGetGallery/Views/Packages/Delete.cshtml
@@ -52,26 +52,9 @@
                 {
                     @ViewHelpers.AlertInfo(
                 @<text>
-                    Note - Permanently deleting packages is not supported, but you can control how they are listed.
+                    You can control how your packages are listed using the checkbox below. Permanent deletion is not supported. <a href="https://docs.microsoft.com/en-us/nuget/policies/deleting-packages">Read more</a>
                 </text>
                     )
-                    <h3>Why can't I delete my package?</h3>
-                    <p>
-                        We only permanently delete NuGet packages in exceptional situations, such as
-                        packages that contain passwords, malicious/harmful code, etc. For more information,
-                        please refer to our <a href="https://docs.microsoft.com/en-us/nuget/policies/deleting-packages">documentation</a>.
-                    </p>
-                    <p>
-                        Unlisting the package will prevent users from finding it by searching the gallery and it will prevent the package manager from discovering the package.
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -52,26 +52,9 @@
                 {
                     @ViewHelpers.AlertInfo(
                 @<text>
-                    Note - Permanently deleting packages is not supported, but you can control how they are listed.
+                    You can control how your packages are listed using the checkbox below. Permanent deletion is not supported. <a href="https://docs.microsoft.com/en-us/nuget/policies/deleting-packages">Read more</a>
```
scottbommarito: We should probably consider

1 - Having a short summary on WHY we don't support permanent deletion (a sentence or two instead of the current several paragraphs)
2 - Having a link to the support request page

I'm concerned that this description here is not verbose enough and it could frustrate users.

karann-msft: We shouldn't provide any reasons on the Why on a static page. This is the same as why we don't explain why you shouldn't edit package metadata. We have docs exactly for that.

What do you recommend should be the text leading up to the Link to support requests? I don't want to get into explaining hard vs soft delete.

karann-msft: Alright, added a link to the support requests. How does that look?

scottbommarito: I think something similar to the following would be totally fine and would still encourage users to go to the docs page:

```
We don't allow supporting permanently deleting packages because doing so would break any users that depend on the availability of the package. See this link to learn more. If you believe your case is exceptional and that your package should be deleted, see this link.
```

I think a short description like this is small enough to be maintainable. Also, we should consider that people running the gallery locally might not want to be redirected to docs.

scottbommarito: also support request link is definitely better

karann-msft: We don't explain the reason for not editing metadata as package signing. I don't see this as being any different. We are explaining the reason for not being able to permanently delete package and that reason is our policies don't allow for it. As for why we have that policy, they can read the docs.

PR #4964, opened 2017-11-07
[Organizations] Improvements to Manage Package Owners page for organizations

scottbommarito:
- Adds "(your organization)" to organizations accounts that you are a member of in the Current Owners list
- Fixes the "who I can remove" logic with respect to organization users
- Improves messaging when users try to add a package owner that already exists or remove a package owner that doesn't exist
- Adds organization users to `Fakes`
- **Vastly improved testing for `JsonApiController`** (will add more for `GetPackageOwners` later)

### DIFF
--- a/src/NuGetGallery/Controllers/JsonApiController.cs
+++ b/src/NuGetGallery/Controllers/JsonApiController.cs
@@ -104,7 +104,7 @@ namespace NuGetGallery
         public virtual ActionResult GetAddPackageOwnerConfirmation(string id, string username)
         {
             ManagePackageOwnerModel model;
-            if (TryGetManagePackageOwnerModel(id, username, add: true, model: out model))
+            if (TryGetManagePackageOwnerModel(id, username, isAddOwner: true, model: out model))
             {
                 return Json(new
                 {
@@ -125,7 +125,7 @@ namespace NuGetGallery
         public async Task<JsonResult> AddPackageOwner(string id, string username, string message)
         {
             ManagePackageOwnerModel model;
-            if (TryGetManagePackageOwnerModel(id, username, add: true, model: out model))
+            if (TryGetManagePackageOwnerModel(id, username, isAddOwner: true, model: out model))
             {
                 var encodedMessage = HttpUtility.HtmlEncode(message);
 
@@ -172,7 +172,7 @@ namespace NuGetGallery
         public async Task<JsonResult> RemovePackageOwner(string id, string username)
         {
             ManagePackageOwnerModel model;
-            if (TryGetManagePackageOwnerModel(id, username, add: false, model: out model))
+            if (TryGetManagePackageOwnerModel(id, username, isAddOwner: false, model: out model))
             {
                 var request = _packageOwnershipManagementService.GetPackageOwnershipRequests(package: model.Package, newOwner: model.User).FirstOrDefault();
 
@@ -290,7 +290,7 @@ namespace NuGetGallery
                 SecurePushSubscription.MinProtocolVersion, SecurePushSubscription.PushKeysExpirationInDays);
         }
 
-        private bool TryGetManagePackageOwnerModel(string id, string username, bool add, out ManagePackageOwnerModel model)
+        private bool TryGetManagePackageOwnerModel(string id, string username, bool isAddOwner, out ManagePackageOwnerModel model)
         {
             if (string.IsNullOrEmpty(id))
             {
@@ -330,13 +330,14 @@ namespace NuGetGallery
                 package.Owners.Any(o => o.MatchesUser(user)) ||
                 _packageOwnershipManagementService.GetPackageOwnershipRequests(package: package, newOwner: user).Any();
 
-            if (add && isOwner)
+            if (isAddOwner && isOwner)
             {
                 model = new ManagePackageOwnerModel(
                     string.Format(CultureInfo.CurrentCulture, Strings.AddOwner_AlreadyOwner, username));
                 return false;
             }
-            if (!add && !isOwner)
+
+            if (!isAddOwner && !isOwner)
             {
                 model = new ManagePackageOwnerModel(
                     string.Format(CultureInfo.CurrentCulture, Strings.RemoveOwner_NotOwner, username));
--- a/src/NuGetGallery/Controllers/JsonApiController.cs
+++ b/src/NuGetGallery/Controllers/JsonApiController.cs
@@ -55,7 +55,7 @@ namespace NuGetGallery
                 return new HttpUnauthorizedResult();
             }
 
-            var currentUserName = HttpContext.User.Identity.Name;
+            var currentUser = GetCurrentUser();
             var packageRegistrationOwners = package.PackageRegistration.Owners;
             var allMatchingNamespaceOwners = package
                 .PackageRegistration
@@ -69,18 +69,18 @@ namespace NuGetGallery
             var owners =
                 packageAndReservedNamespaceOwners
                 .Select(u => new PackageOwnersResultViewModel(
-                    u.Username,
-                    u.EmailAddress,
-                    isCurrentUser: u.Username == currentUserName,
-                    isPending: false,
+                    u, 
+                    currentUser,
+                    Url,
+                    isPending: false, 
                     isNamespaceOwner: true));
 
             var packageOwnersOnlyResultViewModel =
                 packageOwnersOnly
                 .Select(u => new PackageOwnersResultViewModel(
-                    u.Username,
-                    u.EmailAddress,
-                    isCurrentUser: u.Username == currentUserName,
+                    u,
+                    currentUser,
+                    Url,
                     isPending: false,
                     isNamespaceOwner: false));
 
@@ -89,21 +89,13 @@ namespace NuGetGallery
             var pending =
                 _packageOwnershipManagementService.GetPackageOwnershipRequests(package: package.PackageRegistration)
                 .Select(r => new PackageOwnersResultViewModel(
-                    r.NewOwner.Username,
-                    r.NewOwner.EmailAddress,
-                    isCurrentUser: false,
+                    r.NewOwner,
+                    currentUser,
+                    Url,
                     isPending: true,
                     isNamespaceOwner: false));
 
-            var result = owners.Union(pending).Select(o => new
-            {
-                name = o.Name,
-                profileUrl = Url.User(o.Name, relativeUrl: false),
-                imageUrl = GravatarHelper.Url(o.EmailAddress, size: Constants.GravatarImageSize),
-                current = o.Current,
-                pending = o.Pending,
-                isNamespaceOwner = o.IsNamespaceOwner
-            });
+            var result = owners.Union(pending);
 
             return Json(result, JsonRequestBehavior.AllowGet);
         }
@@ -161,10 +153,12 @@ namespace NuGetGallery
                 return Json(new
                 {
                     success = true,
-                    name = model.User.Username,
-                    profileUrl = Url.User(model.User.Username, relativeUrl: false),
-                    imageUrl = GravatarHelper.Url(model.User.EmailAddress, size: Constants.GravatarImageSize),
-                    pending = true
+                    model = new PackageOwnersResultViewModel(
+                        model.User,
+                        model.CurrentUser,
+                        Url,
+                        isPending: true,
+                        isNamespaceOwner: false)
                 });
             }
             else
--- a/src/NuGetGallery/Controllers/JsonApiController.cs
+++ b/src/NuGetGallery/Controllers/JsonApiController.cs
@@ -327,8 +327,9 @@ namespace NuGetGallery
             }
 
             var isOwner = 
-                package.Owners.Any(u => u.MatchesUser(user)) ||
+                package.Owners.Any(o => o.MatchesUser(user)) ||
                 _packageOwnershipManagementService.GetPackageOwnershipRequests(package: package, newOwner: user).Any();
+
             if (add && isOwner)
             {
                 model = new ManagePackageOwnerModel(
--- a/src/NuGetGallery/Controllers/JsonApiController.cs
+++ b/src/NuGetGallery/Controllers/JsonApiController.cs
@@ -104,7 +104,7 @@ namespace NuGetGallery
         public virtual ActionResult GetAddPackageOwnerConfirmation(string id, string username)
         {
             ManagePackageOwnerModel model;
-            if (TryGetManagePackageOwnerModel(id, username, out model))
+            if (TryGetManagePackageOwnerModel(id, username, add: true, model: out model))
             {
                 return Json(new
                 {
@@ -125,7 +125,7 @@ namespace NuGetGallery
         public async Task<JsonResult> AddPackageOwner(string id, string username, string message)
         {
             ManagePackageOwnerModel model;
-            if (TryGetManagePackageOwnerModel(id, username, out model))
+            if (TryGetManagePackageOwnerModel(id, username, add: true, model: out model))
             {
                 var encodedMessage = HttpUtility.HtmlEncode(message);
 
@@ -172,7 +172,7 @@ namespace NuGetGallery
         public async Task<JsonResult> RemovePackageOwner(string id, string username)
         {
             ManagePackageOwnerModel model;
-            if (TryGetManagePackageOwnerModel(id, username, out model))
+            if (TryGetManagePackageOwnerModel(id, username, add: false, model: out model))
             {
                 var request = _packageOwnershipManagementService.GetPackageOwnershipRequests(package: model.Package, newOwner: model.User).FirstOrDefault();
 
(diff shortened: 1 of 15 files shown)


src/NuGetGallery/ViewModels/PackageOwnersResultViewModel.cs
```diff
@@ -9,17 +12,26 @@ public class PackageOwnersResultViewModel
...
         {
-            Name = username;
-            EmailAddress = emailAddress;
-            Current = isCurrentUser;
+            Name = user.Username;
+            EmailAddress = user.EmailAddress;
+            ProfileUrl = url.User(user, relativeUrl: false);
+            ImageUrl = GravatarHelper.Url(user.EmailAddress, size: Constants.GravatarImageSize);
+            CurrentUser = PermissionsService.IsActionAllowed(user, currentUser, AccountActions.AcceptPackageOwnershipOnBehalfOf);
+            CurrentOrganization = PermissionsService.IsActionAllowed(user, currentUser, AccountActions.DisplayPrivateOrganization);
```
chenriksson: Property names don't seem to match the assignment. Should these be renamed?

CurrentUser => CanAcceptPackageOwnership
CurrentOrganization => CanDisplayPrivateOrganization

chenriksson: Or alternatively, why do these do a permissions check as opposed to just checking that user matches the current user, or matches an organization admin?

```
CurrentUser = user.Key == currentUser.Key

CurrentOrganizationAdmin = !CurrentUser
    && (user is Organization)
    && ((Organization)user).Members.Any(m => m.MemberKey == currentUser.Key && m.IsAdmin)
```

I think we need to be careful about how many times PermissionsService.IsActionAllowed is called... can't it result in many queries with each invocation? Why not just do the minimal check required?

tests/NuGetGallery.Facts/Framework/Fakes.cs
```diff
@@ -50,19 +55,19 @@ public Fakes()
...
-            };
+            CreateOrganizationUsers(ref key, credentialBuilder, "", out var organization, out var organizationAdmin, out var organizationCollaborator);
+            Organization = organization;
+            OrganizationAdmin = organizationAdmin;
+            OrganizationCollaborator = organizationCollaborator;
+
+            CreateOrganizationUsers(ref key, credentialBuilder, "Owner", out var organizationOwner, out var organizationAdminOwner, out var organizationCollaboratorOwner);
+            OrganizationOwner = organizationOwner;
+            OrganizationAdminOwner = organizationAdminOwner;
+            OrganizationCollaboratorOwner = organizationCollaboratorOwner;
```
chenriksson: nit: Organization doesn't own package, and OrganizationOwner does? Was confused by 'OrganizationAdminOwner'.... maybe OrganizationOwnerAdmin/OrganizationOwnerCollaborator is more clear?

tests/NuGetGallery.Facts/Framework/Fakes.cs
```diff
@@ -192,5 +213,76 @@ public static Mock<OwinMiddleware> CreateOwinMiddleware()
...
+        private IEnumerable<Func<PackageRegistration>> _packageRegistrations;
+        public IEnumerable<PackageRegistration> PackageRegistrations => _packageRegistrations.Select(f => f());
+
+        private IEnumerable<Func<T>> Get<T>()
+        {
+            return typeof(Fakes)
+                .GetProperties()
+                .Where(p => typeof(T).IsAssignableFrom(p.PropertyType) && p.GetMethod != null)
+                .Select<PropertyInfo, Func<T>>(p => 
+                    () => (T)p.GetMethod.Invoke(this, new object[] { }));
```
chenriksson: nit: is reflection worth is here, as opposed to just a list?

tests/NuGetGallery.Facts/Controllers/JsonApiControllerFacts.cs
```diff
@@ -17,564 +19,1057 @@ namespace NuGetGallery.Controllers
...
-            [Theory]
-            [InlineData("")]
-            [InlineData(null)]
-            public void ThrowsArgumentNullIfUsernameMissing(string username)
-            {
-                // Arrange
-                var controller = GetController<JsonApiController>();
+                [Theory]
+                [MemberData(nameof(ThrowsArgumentNullIfMissing_Data))]
+                public void ThrowsArgumentNullIfPackageIdMissing(InvokePackageOwnerModificationRequest request, string id)
```
chenriksson: In case of failure, would you know which method didn't throw? Wondering if the abstraction is worth it here.

tests/NuGetGallery.Facts/Framework/Fakes.cs
scottbommarito: I agree, `OrganizationOwnerAdmin` and `OrganizationOwnerCollaborator` is better

tests/NuGetGallery.Facts/Framework/Fakes.cs
scottbommarito: it's easier to maintain. this will prevent us from having to plumb through entities when we add additional ones--for example, `Organization` had been forgotten to be added to `ConfigureEntitiesContext`

src/NuGetGallery/ViewModels/PackageOwnersResultViewModel.cs
scottbommarito: I'm using `CurrentUser` and `CurrentOrganization` because they are used as "am I this user?" and "am I a member of this Organization?". I think potentially the names for the `Actions` used here could be better, e.g. `AccountActions.ManagePackageOwnershipOnBehalfOf` and `AccountActions.ManageOrganizationPackageOwnershipOnBehalfOf`.

`PermissionsService` shouldn't be any less performant than the method you proposed.

tests/NuGetGallery.Facts/Controllers/JsonApiControllerFacts.cs
scottbommarito: Abstraction is worth it because without it, I would need to duplicate this method x3.

I'm pretty sure you can figure out which method is which--as long as the test name isn't too long.

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -328,6 +322,23 @@ private bool TryGetManagePackageOwnerModel(string id, string username, out Manag
...
+            {
+                model = new ManagePackageOwnerModel(
+                    string.Format(CultureInfo.CurrentCulture, Strings.AddOwner_AlreadyOwner, username));
+                return false;
+            }
+            if (!add && !isOwner)
+            {
+                model = new ManagePackageOwnerModel(
+                    string.Format(CultureInfo.CurrentCulture, Strings.RemoveOwner_NotOwner, username));
+                return false;
```
shishirx34: I fail to see how we can end up in this piece of code? How will we try to remove a user who is not an owner?

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
```diff
@@ -134,7 +140,15 @@
...
                 success: function (data) {
                     if (data.success) {
-                        if (item.current) {
+                        var numCurrent = 0;
+
+                        ko.utils.arrayForEach(viewModel.owners(), function (owner) {
+                            if (owner !== item && owner.currentUser) {
+                                numCurrent++;
+                            }
+                        });
```
shishirx34: cant you simply use `item.currentUser` here instead of looking up of the deleted owner was current user?

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
```diff
@@ -170,8 +184,8 @@
                 approvedOwner++;
             }
 
-            if (owner.current === true) {
-                currentOwnerOwnsNamespace = owner.isNamespaceOwner();
+            if (owner.currentUser) {
+                currentOwnerOwnsNamespace = currentOwnerOwnsNamespace || owner.isNamespaceOwner();
```
shishirx34: could you elaborate on why this needs to be? can there be multiple instances of `currentUser` with organization?

shishirx34:
Please resolve conflicts.

src/NuGetGallery/Controllers/JsonApiController.cs
loic-sharma: The server should be resilient to client-side bugs or attacks. This is a good validation to add.

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
loic-sharma: If I understand this correctly, this is to handle the case where a user removed their organization as an owner (but the user itself was NOT an owner of the package)? This is unclear, please add a comment.

scottbommarito: @loic-sharma is correct.

`owner.currentUser` is true if `owner` is a user that the current user has the rights to access this page from. Essentially, the current user can act on behalf of ANY of these users.

`owner.currentOrganization` is true if `owner` is a user that the current user is a member of the organization of. This includes organizations that the user DOES NOT have the rights to access this page from

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
scottbommarito: as explained in a comment above, the current user can act on behalf of each `owner` for which `owner.currentUser` is true

in other words, if any of these `owner`s own the namespace, the current user owns the namespace

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
scottbommarito: I need a better naming for one or both of these fields. Any ideas?

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -328,6 +322,23 @@ private bool TryGetManagePackageOwnerModel(string id, string username, out Manag
...
+                package.Owners.Any(o => o.MatchesUser(user)) ||
+                _packageOwnershipManagementService.GetPackageOwnershipRequests(package: package, newOwner: user).Any();
+
+            if (add && isOwner)
+            {
+                model = new ManagePackageOwnerModel(
+                    string.Format(CultureInfo.CurrentCulture, Strings.AddOwner_AlreadyOwner, username));
+                return false;
+            }
+            if (!add && !isOwner)
```
skofman1: what is add?

src/NuGetGallery/Controllers/JsonApiController.cs
shishirx34: I don't disagree, I was wondering if this is getting called at all.

scottbommarito: > How will we try to remove a user who is not an owner?

Someone could modify the request to attempt to delete a user that isn't actually an owner. The UI doesn't allow it, but this guarantees that it can't happen.

src/NuGetGallery/Controllers/JsonApiController.cs
scottbommarito: This is a shared method between adding and removing package owners. If `add` is true, we are attempting to add a new owner, and if `add` is false, we are trying to remove an existing owner.

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
loic-sharma: How about `currentUser` -> `grantsCurrentUserAccess`? Can you explain the `currentOrganization` bit some more? I don't understand this bit:

> This includes organizations that the user DOES NOT have the rights to access this page from

From my understanding, `currentOrganization` is `true` if the user is a member of that organization, HOWEVER, the user may not have manage ownership privileges for that organization? If so, how about `currentOrganization` -> `currentUserIsMember`? I would still add thorough comments to both fields.

src/NuGetGallery/Controllers/JsonApiController.cs
loic-sharma: >if [](start = 12, length = 2)

Super nit: please change `if` to `else if`, or, add a new line in between the two `if` blocks

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
scottbommarito: renamed them to `grantsCurrentUserAccess` and `isCurrentUserMemberOfOrganization`

src/NuGetGallery/Controllers/JsonApiController.cs
scottbommarito: done

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
```diff
@@ -31,7 +31,49 @@
...
+                if (owner.isNamespaceOwner() === true) {
+                    namespaceOwnerCount++;
+                }
+            });
+
+            return approvedOwner >= 2
+                && (!owner.isNamespaceOwner()
+                    || (currentOwnerOwnsNamespace
+                        && namespaceOwnerCount >= 2));
+        },
```
chenriksson: I think the following block could be simplified to minimize checks, etc:

```
ko.utils.array.ForEach(this.owners(), function (owner) {
  if (owner.pending() === true) { continue; }
  
  confirmedOwnerCount++;
  
  if (owner.isNamespaceOwner() === true) {
      currentOwnerOwnsNamespace = currentOwnerOwnsNamespace || owner.grantsCurrentUserAccess;
      namespaceOwnerCount++;
      
      // at least 2 confirmed owners that own the namespace
      if (namespaceOwnerCount == 2) { return true; }
  }
}

// at least 2 confirmed owners and current owner doesn't own namespace
return confirmedOwnerCount >= 2 && !currentOwnerOwnsNamespace;
```

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
```diff
@@ -118,8 +160,15 @@
         },
 
         removeOwner: function (item) {
-            if (item.current) {
-                if (!confirm("Are you sure you want to remove yourself from the owners?")) {
+            var isOnlyUserGrantingAccessToCurrentUser = viewModel.IsOnlyUserGrantingAccessToCurrentUser();
+            var isOnlyUserGrantingAccessToCurrentUserMessage = isOnlyUserGrantingAccessToCurrentUser ? " You will no longer be able to manage the package if you do!" : "";
```
chenriksson: Do we have a way of using Strings.resx for Javascript messages? Or maybe we could at least define at top of script... like `page-api-keys.js` does?

chenriksson: Would also not use '!' for user messages.

src/NuGetGallery/Strings.resx
```diff
@@ -655,4 +655,10 @@ For more information, please contact '{2}'.</value>
   <data name="ApiKeyScopesNotAllowed" xml:space="preserve">
     <value>The current user does not have permission to create an ApiKey with the specified owner or scopes.</value>
   </data>
+  <data name="AddOwner_AlreadyOwner" xml:space="preserve">
+    <value>The user '{0}' is already an owner or pending owner of the package!</value>
+  </data>
+  <data name="RemoveOwner_NotOwner" xml:space="preserve">
+    <value>The user '{0}' is not an owner or pending owner of the package!</value>
```
chenriksson: I don't think we should use '!' for user messages, as it may come off as strong language.

chenriksson:
@scottbommarito Mostly LGTM, but please address my remaining comments then :shipit:

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
```diff
@@ -31,7 +31,49 @@
         message: ko.observable(''),
 
         IsAllowedToRemove: function (owner) {
-            return true;
+            if (isUserAnAdmin.toLocaleLowerCase() === "True".toLocaleLowerCase()
```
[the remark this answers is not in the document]
scottbommarito: haha not sure, this is existing

https://github.com/NuGet/NuGetGallery/pull/4964/files#diff-a6d4fb7c8ca9fa1c604897a46a3b7c24L155

I just moved it.

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
scottbommarito: We do. You declare the variable in the `cshtml` and then use it in the JS. Will do.

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
scottbommarito: unfortunately, due to how `ko.utils.array.ForEach` works, I don't think your optimization works--the early `return`s and `continue`s don't exit from `IsAllowedToRemove`, they exist from the nested function inside of `ko.utils.array.ForEach`

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
scottbommarito: had to implement `nuget.formatString` so that we can use C# format strings inside JS!

src/NuGetGallery/Strings.resx
scottbommarito: yup

Issue #4958, opened 2017-11-06
Auditing for API keys owner (org) scope


chenriksson:

chenriksson:
Note that ApiKeys UI from 11/30 deployment now sets owner scope for new API keys. To verify auditing, check blob audit at /{user}/{guid}-addcredential.audit.v1.json

Issue #4872, opened 2017-10-19
Update API keys UI to support Organization scopes


chenriksson:
Work includes:
- new organization drop down
- Remove push reg scope for non-admins
- Edit shouldn't allow change to organization scope

Hide if user is not member of any organizations. This should essentially hide the feature for now, since we don't yet support organization creation/migration.

chenriksson:
Majority of work was completed in S127, but there's still cleanup remaining to integrate with the PermissionsService once that PR is complete.

Moving to S128 to finish this work.

PR #4963, opened 2017-11-07
[Organizations] Allow uploading and editing packages as organization from UI

scottbommarito:
With this PR, users can upload packages from the UI and specify the owner to be an organization they are a member of.

It also makes the required modifications to the edit UI to allow it to work with the updated partial.

Full API key support will come in a future PR--this PR simply updates the existing `ApiController` code path to work with the changes necessary for `PackagesController`.

PR #4965, opened 2017-11-07
ApiKey owner (org) scope UI work

chenriksson:
Issue #4872.

Work remaining for API key owner (org) scopes:
- Auditing (#4958)
- Remove deprecated Scope.cs ctor (will do in separate PR, to minimize this PR)
- Integrate with PermissionsService

### DIFF
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -473,6 +473,9 @@ namespace NuGetGallery.Authentication
                 string providerName = credential.Type.Split('.')[1];
                 Authenticators.TryGetValue(providerName, out auther);
             }
+            
+            var scopeOwner = credential.Scopes.GetOwnerScope()
+                ?? credential.User;
 
             var credentialViewModel = new CredentialViewModel
             {
@@ -488,7 +491,7 @@ namespace NuGetGallery.Authentication
                 Description = credential.Description, 
                 Value = kind == CredentialKind.Token && credential.Description == null ? credential.Value : null,
                 Scopes = credential.Scopes.Select(s => new ScopeViewModel(
-                        s.Owner?.Username,
+                        scopeOwner.Username,
                         s.Subject,
                         NuGetScopes.Describe(s.AllowedAction)))
                     .ToList(),
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -473,9 +473,6 @@ namespace NuGetGallery.Authentication
                 string providerName = credential.Type.Split('.')[1];
                 Authenticators.TryGetValue(providerName, out auther);
             }
-            
-            var scopeOwner = credential.Scopes.GetOwnerScope()
-                ?? credential.User;
 
             var credentialViewModel = new CredentialViewModel
             {
@@ -491,7 +488,7 @@ namespace NuGetGallery.Authentication
                 Description = credential.Description, 
                 Value = kind == CredentialKind.Token && credential.Description == null ? credential.Value : null,
                 Scopes = credential.Scopes.Select(s => new ScopeViewModel(
-                        scopeOwner.Username,
+                        s.Owner?.Username ?? credential.User.Username,
                         s.Subject,
                         NuGetScopes.Describe(s.AllowedAction)))
                     .ToList(),
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -487,7 +487,11 @@ namespace NuGetGallery.Authentication
                 // Set the description as the value for legacy API keys
                 Description = credential.Description, 
                 Value = kind == CredentialKind.Token && credential.Description == null ? credential.Value : null,
-                Scopes = credential.Scopes.Select(s => new ScopeViewModel(s.Subject, NuGetScopes.Describe(s.AllowedAction))).ToList(),
+                Scopes = credential.Scopes.Select(s => new ScopeViewModel(
+                        s.Owner?.Username,
+                        s.Subject,
+                        NuGetScopes.Describe(s.AllowedAction)))
+                    .ToList(),
                 ExpirationDuration = credential.ExpirationTicks != null ? new TimeSpan?(new TimeSpan(credential.ExpirationTicks.Value)) : null
             };
 
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -226,8 +226,11 @@ namespace NuGetGallery
 
             // Get package owners (user's self or organizations)
             var owners = user.Organizations
-                .Select(o => CreateApiKeyOwnerViewModel(o.Organization, canPushNew: o.IsAdmin))
-                .ToList();
+                .Select(o => CreateApiKeyOwnerViewModel(
+                    o.Organization,
+                    // todo: move logic for canPushNew to PermissionsService
+                    canPushNew: o.IsAdmin)
+                    ).ToList();
             owners.Insert(0, CreateApiKeyOwnerViewModel(user, canPushNew: true));
 
             var model = new ApiKeyListViewModel
@@ -658,7 +661,7 @@ namespace NuGetGallery
                 Response.StatusCode = (int)HttpStatusCode.NotFound;
                 return Json(Strings.CredentialNotFound);
             }
-
+           
             var newCredential = await GenerateApiKeyInternal(
                 cred.Description,
                 BuildScopes(cred.Scopes),
@@ -694,14 +697,25 @@ namespace NuGetGallery
                 Response.StatusCode = (int)HttpStatusCode.BadRequest;
                 return Json(Strings.ApiKeyDescriptionRequired);
             }
+            if (string.IsNullOrWhiteSpace(owner))
+            {
+                Response.StatusCode = (int)HttpStatusCode.BadRequest;
+                return Json(Strings.ApiKeyOwnerRequired);
+            }
 
             // Get the owner scope
-            User scopeOwner = string.IsNullOrEmpty(owner)
-                ? GetCurrentUser()
-                : _userService.FindByUsername(owner);
+            User scopeOwner = _userService.FindByUsername(owner);
             if (scopeOwner == null)
             {
-                Response.StatusCode = (int)HttpStatusCode.NotFound;
+                Response.StatusCode = (int)HttpStatusCode.BadRequest;
+                return Json(Strings.UserNotFound);
+            }
+
+            // todo: move validation logic to PermissionsService
+            var resolvedScopes = BuildScopes(scopeOwner, scopes, subjects);
+            if (!VerifyScopes(scopeOwner, resolvedScopes))
+            {
+                Response.StatusCode = (int)HttpStatusCode.BadRequest;
                 return Json(Strings.UserNotFound);
             }
 
@@ -716,8 +730,9 @@ namespace NuGetGallery
                     expiration = TimeSpan.FromDays(Math.Min(expirationInDays.Value, _config.ExpirationInDaysForApiKeyV1));
                 }
             }
-            
-            var newCredential = await GenerateApiKeyInternal(description, BuildScopes(scopeOwner, scopes, subjects), expiration);
+
+            var newCredential = await GenerateApiKeyInternal(description, resolvedScopes, expiration);
+
             var credentialViewModel = _authService.DescribeCredential(newCredential);
             credentialViewModel.Value = newCredential.Value;
 
@@ -748,7 +763,7 @@ namespace NuGetGallery
                 return Json(Strings.CredentialNotFound);
             }
 
-            var scopeOwner = cred.Scopes.Select(x => x.Owner).Distinct().SingleOrDefault();
+            var scopeOwner = cred.Scopes.GetOwnerScope();
             var scopes = cred.Scopes.Select(x => x.AllowedAction).Distinct().ToArray();
             var newScopes = BuildScopes(scopeOwner, scopes, subjects);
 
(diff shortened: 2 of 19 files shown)


chenriksson:
- Create key has same UI if user is not member of any organization:

image

- Existing v2 key w/o owner scope:

image

- Existing v2 key with owner scope:

image

- Creation of new v2 key, will always have owner scope:

image

- Other Notes:
  - Push new action disabled for organization collaborators
  - Packages displayed depends on the selected owner

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -558,14 +568,24 @@ public virtual ActionResult PasswordChanged()
         [Authorize]
         [HttpPost]
         [ValidateAntiForgeryToken]
-        public virtual async Task<JsonResult> GenerateApiKey(string description, string[] scopes = null, string[] subjects = null, int? expirationInDays = null)
+        public virtual async Task<JsonResult> GenerateApiKey(string description, string owner, string[] scopes = null, string[] subjects = null, int? expirationInDays = null)
```
scottbommarito: I think we should change `scopes` to be `actions` here and elsewhere. It's incredibly confusing that a `Scope` is an object that contains an owner, a set of actions, and a list of subjects AND a "scope" is an action that the API key allows on the package. Now we're also referring to the owner as a "scope" too.

src/NuGetGallery/Scripts/gallery/page-api-keys.js
```diff
@@ -106,13 +101,33 @@
...
             this.PendingDescription = ko.observable();
+
+            // Package owner selection
+            this.PackageOwner = ko.observable();
+            this.PackageOwnerName = ko.pureComputed(function () {
+                return self.PackageOwner() && self.PackageOwner().Owner;
+            }, this);
+
+            // Do not default to push new scope if disabled (organization collaborator)
+            this.PushNewEnabled = ko.pureComputed(function () {
```
scottbommarito: You will probably want to do this for all three possible actions (push new, push version, and unlist) because the `PermissionsService` allows for it.

src/NuGetGallery/Views/Users/ApiKeys.cshtml
```diff
@@ -174,6 +174,10 @@
                         <!-- /ko -->
                     </li>
                 </ul>
+                <!-- ko if: Owner -->
+                <b>Package owner:</b> <span data-bind="text: Owner"></span>
```
scottbommarito: any thoughts on showing the current user if there is no owner scope?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -558,14 +568,24 @@ public virtual ActionResult PasswordChanged()
...
         {
             if (string.IsNullOrWhiteSpace(description))
             {
                 Response.StatusCode = (int)HttpStatusCode.BadRequest;
                 return Json(Strings.ApiKeyDescriptionRequired);
             }
 
+            // Get the owner scope
+            User scopeOwner = string.IsNullOrEmpty(owner)
```
scottbommarito: We need to check that the current user actually has the permissions to create an API key with this user as their owner scope. From what I can see here, someone could create an API key with any user as the owner in the scope.

On a similar note, we should check that the user has permissions to create an API key with these scopes (actions) on the owner (e.g. organization collaboration shouldn't have push new).

src/NuGetGallery/Views/Users/ApiKeys.cshtml
```diff
@@ -293,6 +297,23 @@
                 </div>
             }
         </div>
+
+        @if (Model.PackageOwners.Count > 1)
```
scottbommarito: If `Model.PackageOwners.Count == 1` you may want to pass the single option as a hidden item on the form.

It's not needed because you have the `GetCurrentUser` call in `GenerateApiKey`, but in case some code changes make it possible for the only package owner to be NOT the current user, this is more consistent.

src/NuGetGallery/ViewModels/ScopeViewModel.cs
```diff
@@ -5,11 +5,13 @@ namespace NuGetGallery
...
     public class ScopeViewModel
     {
+        public string Owner { get; set; }
         public string Subject { get; set; }
         public string AllowedAction { get; set; }
 
-        public ScopeViewModel(string subject, string allowedAction)
+        public ScopeViewModel(string owner, string subject, string allowedAction)
         {
+            Owner = owner ?? "*";
```
scottbommarito: `"*"` -> `NuGetPackagePattern.AllInclusivePattern`

src/NuGetGallery/Controllers/UsersController.cs
chenriksson: I agree. The term "scope" is used interchangeably with scope action throughout the code.

I'm ok with fixing, but prefer to do it in a separate PR as I don't want to inflate this one.

src/NuGetGallery/Scripts/gallery/page-api-keys.js
chenriksson: Push version and unlist are always available, even to organization collaborators. I only need to determine when to enable/disable push new.

src/NuGetGallery/Views/Users/ApiKeys.cshtml
chenriksson: We decided that no owner scope means that it can match any, not just the current user. That's why I chose to show "*"... but am open to a different suggestion.

src/NuGetGallery/ViewModels/ScopeViewModel.cs
chenriksson: It's not a package pattern, but will add AllInclusivePattern const to the view model.

src/NuGetGallery/Controllers/UsersController.cs
joelverhagen: Yes, client side validation is a "hint" to the user. Server side validation is the real enforcement.

src/NuGetGallery/Scripts/gallery/page-api-keys.js
```diff
@@ -106,13 +101,33 @@
...
+                return self.PackageOwner() && self.PackageOwner().CanPushNew;
+            }, this);
+            this.PushNewEnabled.subscribe(function (newValue) {
+                var defaultPushId = (newValue)
```
joelverhagen: nit: why parens around `newValue`?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -558,14 +568,24 @@ public virtual ActionResult PasswordChanged()
...
+                ? GetCurrentUser()
+                : _userService.FindByUsername(owner);
+            if (scopeOwner == null)
```
joelverhagen: Just making sure -- this code is saying all future API keys must be scoped to a single owner?

joelverhagen:
We need a final sign-off from all feature owners and @anangaur that:

1. What do API keys without an owner scope do when pushing a new ID and the user is part of at least one org. _This is the ambiguous case..._
1. What should the API key UI show for the scopes/owners/actions for these old owner-less API keys?

I keep forgetting our conclusion on these discussions 😊.

src/NuGetGallery/Scripts/gallery/page-api-keys.js
scottbommarito: always available NOW

in the future they may not be, so we use `PermissionsService` to centralize this logic

src/NuGetGallery/Views/Users/ApiKeys.cshtml
scottbommarito: Oh my bad, I will need to update my PR to understand this logic

src/NuGetGallery/Views/Users/ApiKeys.cshtml
scottbommarito: https://github.com/NuGet/NuGetGallery/pull/4963/files#diff-5058b8e804fd7b6e9e60a29de7f6fff6R60 is how I did it for upload/edit

scottbommarito:
@joelverhagen 
question 1 is handled in https://github.com/NuGet/NuGetGallery/pull/4963 (we should ask it there)
not sure about question 2

anangaur:
> What do API keys without an owner scope do when pushing a new ID and the user is part of at least one org. This is the ambiguous case...
What should the API key UI show for the scopes/owners/actions for these old owner-less API keys?

Both cases are for existing API keys and they should get associated to the account owner's account, UI or otherwise. The spec has this discussed here: https://github.com/NuGet/Home/wiki/Organizations-on-NuGet.org#what-happens-to-authors-existing-api-keys

src/NuGetGallery/Scripts/gallery/page-api-keys.js
```diff
@@ -139,7 +154,7 @@
                     return id;
                 }, self);
             }
-
+
```
scottbommarito: nit: unnecessary whitespace

src/NuGetGallery/Views/Users/ApiKeys.cshtml
scottbommarito: according to @anangaur, we decided that no owner scope means only match current user

personally, I prefer this because it forces users to use organizations correctly

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -558,14 +568,24 @@ public virtual ActionResult PasswordChanged()
...
+            {
+                Response.StatusCode = (int)HttpStatusCode.NotFound;
```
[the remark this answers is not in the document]
chenriksson: Updated

src/NuGetGallery/Scripts/gallery/page-api-keys.js
chenriksson: Updated

src/NuGetGallery/ViewModels/ScopeViewModel.cs
chenriksson: Removed per @anangaur feedback that legacy keys match only current user.

src/NuGetGallery/Controllers/UsersController.cs
chenriksson: Yes, API keys will be scoped to a single owner.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -105,23 +105,33 @@ public virtual ActionResult ApiKeys()
...
                 .ToList();
 
-            // Get package IDs
-            var packageIds = _packageService
-                .FindPackageRegistrationsByOwner(user)
-                .Select(p => p.Id)
-                .OrderBy(i => i)
+            // Get package owners (user's self or organizations)
+            var owners = user.Organizations
+                .Select(o => CreateApiKeyOwnerViewModel(o.Organization, canPushNew: o.IsAdmin))
```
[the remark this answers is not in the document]
chenriksson: Will integrate w/ PermissionsService in a follow-up PR.

src/NuGetGallery/Controllers/UsersController.cs
chenriksson: Adding server side validation, but will integrate w/ PermissionsService in follow-up PR since it's under flux right now.

src/NuGetGallery/Views/Users/ApiKeys.cshtml
chenriksson: Modified to still populate PackageOwner observable when single owner (self), but will hide the drop-down.

src/NuGetGallery/Scripts/gallery/page-api-keys.js
chenriksson: reverted

chenriksson:
@scottbommarito @joelverhagen Addressed feedback. I will integrate it with PermissionsService in a separate PR, to avoid taking dependency for now on #4963.

There's still a test I'd like to add for server-side validation, but I'm currently blocked there by a VS bug with test discovery.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -524,7 +527,11 @@ public virtual CredentialViewModel DescribeCredential(Credential credential)
                 // Set the description as the value for legacy API keys
                 Description = credential.Description, 
                 Value = kind == CredentialKind.Token && credential.Description == null ? credential.Value : null,
-                Scopes = credential.Scopes.Select(s => new ScopeViewModel(s.Subject, NuGetScopes.Describe(s.AllowedAction))).ToList(),
+                Scopes = credential.Scopes.Select(s => new ScopeViewModel(
+                        scopeOwner.Username,
```
scottbommarito: You shouldn't cache the value and use it for each scope in the list when you're enumerating through each scope anyway.

```
s.Owner?.Username ?? credential.User
```

src/NuGetGallery/Extensions/ScopeExtensions.cs
```diff
@@ -70,6 +71,19 @@ public static bool HasOwnerScope(this Scope scope)
 
             return scope.OwnerKey.HasValue;
         }
+
+        public static User GetOwnerScope(this IEnumerable<Scope> scopes)
```
scottbommarito: I would suggest removing this function and using `scopes.SingleOrDefault(s => s.Owner)` instead in all the places that you use it for the following reasons.

- it can be reduced into the one-liner above
- it will need to be removed when we allow multiple owners per scope
- its existence makes it very easy to not treat scopes appropriately (e.g. by assuming that the first owner in the scope applies to ALL scopes)

src/NuGetGallery/Strings.resx
```diff
@@ -618,4 +618,7 @@ For more information, please contact '{2}'.</value>
   <data name="AccountDelete_Success" xml:space="preserve">
     <value>The account:{0} was deleted succesfully.</value>
   </data>
+  <data name="ApiKeyOwnerRequired" xml:space="preserve">
+    <value>Can't generate an API key without a package owner.</value>
```
scottbommarito: `You must specify a package owner to generate an API key.` might be better

src/NuGetGallery/Extensions/ScopeExtensions.cs
chenriksson: IMHO, if we eventually allow multiple scope owners, having this in a central location would actually help identify callers that need updating.

src/NuGetGallery/Strings.resx
chenriksson: Was consistent with ApiKeyDescriptionRequired, but I have no strong preference. Updated.

src/NuGetGallery/ViewModels/ApiKeyViewModel.cs
```diff
@@ -20,6 +20,13 @@ public ApiKeyViewModel(CredentialViewModel cred)
                 throw new ArgumentNullException(nameof(cred));
             }
 
+            // Currently ApiKeys.cshtml has single Owner per ApiKey restriction.
+            var owner = cred
```
joelverhagen: Can you use `GetOwnerScope`  extension method here?

chenriksson: No... code looks the same, but this is querying on view models not entities.

PR #4968, opened 2017-11-07
Fixing ApiScopeRequiredAttribute regression

chenriksson:
Regression from PR #4945 
https://github.com/NuGet/NuGetGallery/pull/4945/files#diff-1463a262f19818464c6a4983b635e94aL18

When I changed type from `List<string>` to `string[]`, the property was getting overwritten by AutofacFilterProvider property injection. Fix is to make the property private.

### DIFF
--- a/src/NuGetGallery/Filters/ApiScopeRequiredAttribute.cs
+++ b/src/NuGetGallery/Filters/ApiScopeRequiredAttribute.cs
@@ -13,7 +13,7 @@ namespace NuGetGallery.Filters
     public sealed class ApiScopeRequiredAttribute 
         : AuthorizeAttribute
     {
-        public string[] ScopeActions { get; set; }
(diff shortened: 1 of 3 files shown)


skofman1:
Should this be against dev and not master?

chenriksson:
@skofman1 Thanks - fixed!

chenriksson:
@joelverhagen I'm trying, but having a hard time figuring out how to enable the autofac filter provider for the new test. Possible issues:
- The test needs Autofac's dependency resolver, but other tests are using the default
- Our version of Autofac seems to depend on HttpContext.Current

Trying to find a way around this, but I suggest I timebox this. If I'm blocking deployment too long, how about I submit the fix to dev, then open an issue to investigate adding this test?

/CC @skofman1

PR #4976, opened 2017-11-09
Hotfix for legacy api keys

chenriksson:
This was a regression from #4945


tests/NuGetGallery.Facts/Extensions/PrincipalExtensionsFacts.cs
```diff
@@ -117,6 +117,20 @@ public void WhenNoScopeClaims_ReturnsNull()
 
                 Assert.Equal(null, identity.GetScopesFromClaim());
             }
+
+            [Theory]
+            [InlineData("")]
```
joelverhagen: null? whitespace? `"null"`? How far do we want to go here?

chenriksson: Added whitespace scenario. Claim will throw if value is null, but I have test above that tests with no scope claim.

PR #4977, opened 2017-11-09
Account delete user flow

cristinamanum:
pack1
user2
user1
owners

The user flow for account delete.


src/NuGetGallery/Views/Packages/ContactOwners.cshtml
```diff
@@ -59,20 +59,31 @@
...
+                                @:You'll have to find some other way to contact them.
+                            }
+                        </text>
+                         );
+                }
+                else
+                {
+                    @ViewHelpers.AlertWarning(
+                        @<text>
+                            Sorry, this package has no owners and is not being actively maintained. You can try to find a contact for "@Model.PackageId" through its <a href="@Model.ProjectUrl">Project Site</a>
```
scottbommarito: check that the package has a `ProjectUrl` before you ask them to go to it like above

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -626,17 +626,24 @@
 
             <h2>Owners</h2>
             <ul class="list-unstyled owner-list">
-                @foreach (var owner in Model.Owners)
+                @if (Model.Owners.Count() == 0)
```
scottbommarito: Use `.Any()` instead of `.Count() == 0`

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -626,17 +626,24 @@
...
-                    <li>
-                        @if (!String.IsNullOrEmpty(owner.EmailAddress))
-                        {
-                            <a href="@Url.User(owner)" title="@owner.Username">@ViewHelpers.GravatarImage(owner.EmailAddress, owner.Username, Constants.GravatarElementSize)</a>
-                        }
-                        <a href="@Url.User(owner)" title="@owner.Username">
-                            @owner.Username
-                        </a>
-                    </li>
+                        @ViewHelpers.AlertWarning(@<text>@Strings.PackageHasNoOwnersMessage</text>)
```
scottbommarito: We don't use strings for other user-facing strings in the `.cshtml`, no need to put it in a string.

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
```diff
@@ -0,0 +1,93 @@
...
+}
+
+<section role="main" class="container main-container page-delete-account">
+    <div class="form-group">
+        <div class="form-group package-controls">
+            <h1 class="text-center">Delete Account</h1>
+            <div class="text-center ms-font-xxl">
+                <a href="@Url.User(@Model.User)">@Model.AccountName</a>
+            </div>
+            @if ((TempData["RequestFailedMessage"] != null))
```
scottbommarito: You should be able to just use `TempData["Message"]` and get rid of this.

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
```diff
@@ -0,0 +1,93 @@
...
+            {
+                var message = (string)TempData["RequestFailedMessage"];
+                @ViewHelpers.AlertDanger(@<p><b class="keywords">@message</b> <br /></p>)
+            }
+            else
+            {
+                if ((bool)ViewData["HasPendingRequest"])
```
scottbommarito: Put this in the model

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -95,6 +99,59 @@ public virtual ActionResult Account()
...
+        [Authorize]
+        public virtual ActionResult DeleteRequest()
+        {
+            var user = GetCurrentUser();
+
+            if (user == null || user.IsDeleted)
+            {
+                return HttpNotFound("User not found.");
+            }
+            bool pendingRequest = _supportRequestService.GetIssues(assignedTo: null, reason: null, issueStatusId: null, galleryUsername: user.Username)
```
scottbommarito: If someone cancels a request to delete their account, won't this still find the existing request?

It seems a bit awkward to query the support request service for whether or not a user requested deletion. Additionally, when we move off our support request system, this will need to be changed. I would rather store this in the database somehow.

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
```diff
@@ -0,0 +1,93 @@
...
+                            @Html.Partial("_UserPackagesListForDeletedAccount", new ManagePackagesListViewModel(@Model.Packages, "Packages"))
+                        }
+                    </div>
+                    <div class="form-group danger-zone">
+                        <text>
+                            @using (Html.BeginForm("SendAccountDeleteNotificationAsync", "Users", FormMethod.Post, new { id = "delete-form" }))
+                            {
+                                @Html.AntiForgeryToken()
+                                <p>
+                                    <strong>By proceeding, I understand that I am relinquishing ownership of the package/s listed above, and any existent package ID reservation/s will be dissociated with this account.</strong>
```
scottbommarito: Can we use less verbose terms than `relinquish` and `dissociate`? `losing` should be perfectly understandable

scottbommarito: We should show all the reserved namespaces associated with the account in the same way we show packages. This will make it clearer for users which reserved namespaces will be deleted and which ones have multiple users.

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -1828,6 +1829,111 @@ public void DeleteHappyAccount()
...
                 Assert.Equal<int>(1, model.Packages.Count());
             }
+
+            [Theory]
+            [InlineData(false)]
+            [InlineData(true)]
+            public void DeleteAccountRequestView (bool withPendingIssues)
+            {
+                // Arrange
+                string userName = "DeletedUser";
```
scottbommarito: this test (and the rest of the account deletion tests) should use `Fakes`

```
var fakes = Get<Fakes>();
var userToDelete = fakes.Owner;
var packageToDelete = fakes.Package;
```

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -1828,6 +1829,111 @@ public void DeleteHappyAccount()
                 Assert.Equal(userName, model.AccountName);
                 Assert.Equal<int>(1, model.Packages.Count());
             }
+
+            [Theory]
+            [InlineData(false)]
+            [InlineData(true)]
+            public void DeleteAccountRequestView (bool withPendingIssues)
```
scottbommarito: This is not a test for the `TheDeleteAccountAction`, so it should have its own method.

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
```diff
@@ -0,0 +1,93 @@
...
+                var message = (string)TempData["RequestFailedMessage"];
+                @ViewHelpers.AlertDanger(@<p><b class="keywords">@message</b> <br /></p>)
+            }
+            else
+            {
+                if ((bool)ViewData["HasPendingRequest"])
+                {
+                    @ViewHelpers.AlertWarning(@<text>
+                        <b class="keywords">Request submitted.</b> <br />
+                        It may take up to 5 business days to process Account deletion requests.<a href="https://github.com/NuGet/Home/wiki/NuGet-Account-Deletion-Workflow">Read more.</a><br />
```
loic-sharma: Add a space between the period and the link here.

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
loic-sharma: Also, no need for the passive voice here. "am relinquishing" -> "will relinquish" (or "will lose").

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -280,6 +280,11 @@ public static void RegisterUIRoutes(RouteCollection routes)
                 new { controller = "Users", action = "Delete" });
 
             routes.MapRoute(
+                RouteName.UserDeleteAccount,
+               "account/delete",
```
loic-sharma: The spacing looks off here.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -775,6 +775,7 @@ public virtual ActionResult ContactOwners(string id)
                 CopySender = true,
             };
 
+            ViewData["HasOwners"] = package.PackageRegistration.Owners.Count > 0;
```
loic-sharma: `.Owners.Count > 0` -> `.Owners.Any()`

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -95,6 +99,59 @@ public virtual ActionResult Account()
...
+                                    .Any(issue => string.Equals(issue.IssueTitle, Strings.AccountDelete_SupportRequestTitle, StringComparison.CurrentCultureIgnoreCase) && 
+                                         issue.IssueStatus.Key != 3 &&
```
loic-sharma: What does `issue.IssueStatus.Key != 3` mean? A constant for `3` could help explain this code.

src/NuGetGallery/Strings.resx
```diff
@@ -618,4 +618,13 @@ For more information, please contact '{2}'.</value>
...
     <value>The account:{0} was deleted succesfully.</value>
   </data>
+  <data name="AccountDelete_CreateSupportRequestFails" xml:space="preserve">
+    <value>The request failed to be submitted. Please try again or contact nuget support.</value>
+  </data>
+  <data name="AccountDelete_SupportRequestTitle" xml:space="preserve">
+    <value>DeleteAccountRequest</value>
+  </data>
+  <data name="PackageHasNoOwnersMessage" xml:space="preserve">
+    <value>This package has no owners and is not being actively maintained.</value>
```
loic-sharma: No need for the word "being" here.

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
loic-sharma: Super nit: the capitalization on "Account" seems weird - could we leave it lower cased?

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -1828,6 +1829,111 @@ public void DeleteHappyAccount()
...
+                    Description = "TestPackage",
+                    Key = 1,
+                    Version = "1.0.0",
+                    PackageRegistration = packageRegistration
+                };
+                packageRegistration.Packages.Add(userPackage);
+
+                List<Package> userPackages = new List<Package>() { userPackage };
+                List<Issue> issues = new List<Issue>();
+                if(withPendingIssues)
```
loic-sharma: A space is missing between the `if` and `(`

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -647,6 +647,32 @@ public void SendPackageAddedNotice(Package package, string packageUrl, string pa
             }
         }
 
+        public void SendAccountDeleteNotice(MailAddress mailAddress, string account)
+        {
+            string body = @"We received a request to delete your account {0}. If you did not initiate this request please contact the {1} team.
```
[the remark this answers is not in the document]
cristinamanum: Updated

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: For the first question - no they will not find it because it was not created.
For the second question, the result of the user action to submit the request to have the package deleted is that a new support request is created. 
As discussed offline updated to use the issuekey.

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
```diff
@@ -59,20 +59,31 @@
             }
             else
             {
-                @ViewHelpers.AlertWarning(
-                    @<text>
-                        Sorry, the owners of this package do not allow contacting them through this form.
+                if ((bool)ViewData["HasOwners"])
```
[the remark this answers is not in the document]
cristinamanum: Updated

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
cristinamanum: updated

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
cristinamanum: Updated

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -95,6 +99,59 @@ public virtual ActionResult Account()
...
+                HasOrphanPackages = listPackageItems.Any(p => p.Owners.Count <= 1),
+            };
+            
+            return View("DeleteAccount", model);
+        }
+
+        [HttpPost]
+        [Authorize]
+        [ValidateAntiForgeryToken]
+        public virtual async Task<ActionResult> SendAccountDeleteNotificationAsync()
```
skofman1: RequestAccountDeletionAsync?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -95,6 +99,59 @@ public virtual ActionResult Account()
...
+        {
+            var user = GetCurrentUser();
+            bool createSupportRequestStatus = await _supportRequestService.AddNewSupportRequestAsync(Strings.AccountDelete_SupportRequestTitle,
```
skofman1: There are no checks here:
if (user == null || user.IsDeleted)

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
cristinamanum: I think that is good to target having them in resource file. It will make easier to localize the site. Do you see any disadvantages?

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
cristinamanum: Do you mean to change the naming from "RequestFailedMessage" to "Message"?

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
cristinamanum: Updated

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
cristinamanum: This is as per spec. Following up with the spec owner.

cristinamanum: Implementation is as spec-ed. In think the info and warnings that are showed are enough for the user.

src/NuGetGallery/App_Start/Routes.cs
cristinamanum: Updated

src/NuGetGallery/Controllers/PackagesController.cs
cristinamanum: Updated

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: The code was updated. It represents the status for a specific issue. ( 3 == resolved)

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
cristinamanum: Updated

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
cristinamanum: updated

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: Updated

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: Updated

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
cristinamanum: What are the benefits of getting the user through fakes versus creating it directly?

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: I don't think we use `Async` in controller action names

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
loic-sharma: Fair enough. You should still add a space between "requests." and the "Read more." link though :)

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: The async suffix was added to follow the naming convention for the async methods. The observation that @scottbommarito made is correct. The name was changed to RequestAccountDeletion to be in sync with the current pattern.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
scottbommarito: The disadvantages are
1 - if we actually were to localize the site, it would require a huge effort anyway because this convention isn't used anywhere else
2 - it's harder to work with (you have to work with the `.resx` file instead of the rest of the code and you have to recompile if you want to change the string unlike the rest of razor) with little gain

If we reused this string somewhere else I'd agree that it was probably worth, but given that we don't, I don't think it's worth it.

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
scottbommarito: `TempData["Message"]` already writes a message to the top of the page. There's no reason to introduce `TempData["RequestFailedMessage"]` when `TempData["Message"]` does the same thing.

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
scottbommarito: `Fakes` is used in some of our tests to make it easier to get users. Unless your test requires that you create special users (for example, those that own multiple packages or are in multiple organizations), if your test is in a file that uses `Fakes`, your test should use `Fakes`.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -95,6 +99,64 @@ public virtual ActionResult Account()
...
+        public virtual async Task<ActionResult> RequestAccountDeletion()
+        {
+            var user = GetCurrentUser();
+
+            if (user == null || user.IsDeleted)
+            {
+                return HttpNotFound("User not found.");
+            }
+
+            bool createSupportRequestStatus = await _supportRequestService.AddNewSupportRequestAsync(Strings.AccountDelete_SupportRequestTitle,
```
scottbommarito: nit: I think `bool isSupportRequestCreated` is a better name for this

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -647,6 +647,36 @@ public void SendPackageAddedNotice(Package package, string packageUrl, string pa
             }
         }
 
+        public void SendAccountDeleteNotice(MailAddress mailAddress, string account)
+        {
+            string body = @"We received a request to delete your account {0}. If you did not initiate this request please contact the {1} team imediatelly.
```
scottbommarito: `We received a request` -> `We have received a request`

`your account {0}` -> `your account, '{0}'`

`imediatelly` -> `immediately`

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -647,6 +647,36 @@ public void SendPackageAddedNotice(Package package, string packageUrl, string pa
...
+{2}When your account will be deleted, we will:
```
scottbommarito: We haven't needed to use `Environment.NewLine`s in previous emails. Have you tested that the email sent looks correctly and doesn't have extra newlines?

src/NuGetGallery/Strings.resx
```diff
@@ -618,4 +618,13 @@ For more information, please contact '{2}'.</value>
   <data name="AccountDelete_Success" xml:space="preserve">
     <value>The account:{0} was deleted succesfully.</value>
   </data>
+  <data name="AccountDelete_CreateSupportRequestFails" xml:space="preserve">
+    <value>The request failed to be submitted. Please try again or contact nuget support.</value>
```
scottbommarito: Two things
1 - `nuget` -> `NuGet`
2 - This should use [`IAppConfiguration.Brand`](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Configuration/IAppConfiguration.cs#L126) so that if somebody deploys the gallery themselves and changes the brand, it doesn't say `NuGet support` anyway

An easy alternative may be just remove the use of `nuget` and just show `contact support`.

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
```diff
@@ -59,20 +59,40 @@
...
+                                @:You'll have to find some other way to contact them.
+                            }
+                        </text>
+                         );
+                }
+                else
+                {
+                    @ViewHelpers.AlertWarning(
+                        @<text>
+                            Sorry, this package has no owners and is not being actively maintained.
```
scottbommarito: I would say it's not being actively maintained ON the `NuGet Gallery`. Someone might still be working on it on `GitHub` or some other site, but is not maintaining this package on `NuGet`.

`is not being actively maintained on @Config.Brand.`

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
```diff
@@ -59,20 +59,40 @@
...
+
+                            @if (!string.IsNullOrEmpty(Model.ProjectUrl))
+                            {
+                                @:You can try contacting the package owners of "@Model.PackageId" through its <a href="@Model.ProjectUrl">Project Site</a>
```
scottbommarito: We just said it doesn't have any `package owners`, so we should say
```
You can try contacting someone maintaining "@Model.PackageId" through its <a href="@Model.ProjectUrl">Project Site</a>
```

Also in the `else`
```
You'll have to find someone maintaining the package through other means.
```

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
```diff
@@ -0,0 +1,93 @@
...
+                var message = (string)TempData["RequestFailedMessage"];
+                @ViewHelpers.AlertDanger(@<p><b class="keywords">@message</b> <br /></p>)
+            }
+            else
+            {
+                if (Model.HasPendingRequests)
+                {
+                    @ViewHelpers.AlertWarning(@<text>
+                        <b class="keywords">Request submitted.</b> <br />
+                        It may take up to 5 business days to process Account deletion requests. <a href="https://go.microsoft.com/fwlink/?linkid=862770">Read more.</a><br />
```
scottbommarito: `Account` -> `account`

(I don't think account should be capitalized here.)

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -1829,6 +1830,119 @@ public void DeleteHappyAccount()
...
+        public class TheDeleteAccountRequestAction : TestContainer
+        {
+            [Theory]
+            [InlineData(false)]
+            [InlineData(true)]
+            public void DeleteAccountRequestView(bool withPendingIssues)
+            {
+                // Arrange
+                string userName = "DeletedUser";
+                string emailAddress = $"{userName}@coldmail.com";
```
scottbommarito: `coldmail.com`...that's brilliant hahaha

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -1829,6 +1830,119 @@ public void DeleteHappyAccount()
...
+
+                List<Package> userPackages = new List<Package>() { userPackage };
+                List<Issue> issues = new List<Issue>();
+                if ( withPendingIssues )
+                {
+                    issues.Add(new Issue()
+                    {
+                        IssueTitle = Strings.AccountDelete_SupportRequestTitle,
+                        OwnerEmail = emailAddress,
+                        IssueStatus = new IssueStatus() { Key = 1, Name = "OneIssue" }
```
scottbommarito: Use the `IssueKeys` constant instead of `1`.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
cristinamanum: Removed.

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
cristinamanum: I prefer to let it specific for the scenario that it targets.

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
```diff
@@ -58,6 +58,11 @@ public IReadOnlyCollection<Issue> GetIssues(int? assignedTo = null, string reaso
             return queryable.ToList();
         }
 
+        public IReadOnlyCollection<Issue> GetOpenIssues( Func<Issue,bool> filter )
```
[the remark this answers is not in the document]
cristinamanum: The issues key is private and I preferred to not change it as I think that was a reason for making it private. Why is weird to have a predicate that enables filtering?

src/NuGetGallery/Controllers/UsersController.cs

src/NuGetGallery/Services/MessageService.cs
cristinamanum: Updated.

src/NuGetGallery/Services/MessageService.cs
cristinamanum: I think that the email looks good(attached).
email

src/NuGetGallery/Strings.resx
cristinamanum: Updated to not use NuGet in the notification.

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
cristinamanum: Updated

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
cristinamanum: The first is as spec-ed. The second, the scenario is about finding the owners of the package. I think that the message should refer to the owners.

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
cristinamanum: Agree - updated.

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
scottbommarito: `IssueKeys` should be public...there's no reason to have "magic values" around the code when we can just make the enum public and get rid of them.

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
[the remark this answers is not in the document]
scottbommarito: As I said in another comment, `IssueKeys` should be public. There's no reason to have "magic values" around the code when we can just make the enum public and get rid of them. I think it was probably private before because it wasn't needed outside of that class, but now it is so it should be changed.

It's weird to have a predicate because
1 - syntactically, predicates aren't as elegant or powerful as LINQ
2 - it requires every caller to write a filtering function when the filtering is likely done on the same fields (the user the request is for and the title, for example)

src/NuGetGallery/Services/MessageService.cs
scottbommarito: I agree it looks good.

However, it might not look consistent with the rest of our emails, because they don't have newlines between every line.

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
scottbommarito: It's a minor grammatical change. I think it's ok to change despite how it was spec'ed

My confusion is just that the statement above says "this package has no owners" (and not "this package has no owners on NuGet.org"), but then we are talking about it having owners again. It's not too big of a deal but I think changing the phrasing as detailed would improve the clarity of the statement.

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
cristinamanum: I understand your point. If we would like to refactor the code around support it would be in a different PR.

src/NuGetGallery/Views/Packages/ContactOwners.cshtml
cristinamanum: The code updated to be : Sorry, this package has no owners and is not being actively maintained on NuGet.org

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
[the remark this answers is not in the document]
cristinamanum: Updated to make IssuesKeys public and use them on this code path.

Issue #4717, opened 2017-09-21
We should not show any details about the deleted package on the details page


anangaur:
Deleted packages are rarest of rare cases on NuGet.org and should be never done. However, showing the details of the deleted package and ways to install the package is unnecessary and confusing. We should not show all these details.

E.g:
https://www.nuget.org/packages/Microsoft.Azure.Management.Monitor/0.17.0-preview
![image](https://user-images.githubusercontent.com/14800916/30709156-1c20a1a2-9eb6-11e7-8f88-728972e140c0.png)

scottbommarito:
I will note that only SOFT-deleted packages show this information. HARD-deleted packages are completely invisible.

anangaur:
I think none of them should show for a normal user because the package doesn't exist. I am okay with the page being shown.

PR #4978, opened 2017-11-09
Hide package metadata when the package is deleted

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/4717.

I recommend ignoring whitespace in the diff:
https://github.com/NuGet/NuGetGallery/pull/4978/files?w=1

This also add a default warning banner for local development (like DEV and INT).

## Before

screencapture-localhost-packages-nuget-server-2-14-0-1510258411423

## After

screencapture-localhost-packages-nuget-server-2-14-0-1510258828190

### DIFF
--- a/src/NuGetGallery/Web.config
+++ b/src/NuGetGallery/Web.config
@@ -56,7 +56,7 @@
     <add key="Gallery.Environment" value="Development" />
     
     <!-- banner warning on the top - leave blank to remove the banner from the site. -->
-    <add key="Gallery.WarningBanner" value="" />
+    <add key="Gallery.WarningBanner" value="This is the local development environment." />
 
     <add key="Gallery.FacebookAppId" value="" />
     <!-- Set this if you have a Application Insights instrumentation key to use for your gallery deployment -->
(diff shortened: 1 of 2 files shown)


joelverhagen:
I disagree. These metadata values are essentially dead. It would add complexity to the code and provides no administrative/owner benefit.

scottbommarito:
They're dead, but I think they could help admins and owners understand what was deleted in case a package was deleted accidentally or a user has forgotten what package was deleted.

I suppose it's not too big of a deal, however.

anangaur:
The metadata info is irrelevant once the package is deleted. The only information to persist would be to show the owner of the package that the package was deleted.

src/NuGetGallery/Web.config
```diff
@@ -56,7 +56,7 @@
     <add key="Gallery.Environment" value="Development" />
     
     <!-- banner warning on the top - leave blank to remove the banner from the site. -->
-    <add key="Gallery.WarningBanner" value="" />
+    <add key="Gallery.WarningBanner" value="This is the local development environment." />
```
loic-sharma: Is this change intended to be checked-in? Personally I would prefer to not see this while developing.

joelverhagen: I think it's a good practice to have it. When using the delete UI or upload, it should be very clear which environment you are working on.

If you feel strongly I can remove it.

skofman1: Just make sure our prod config overrides this value explicitly 😆

PR #5001, opened 2017-11-14
Improve performance of the PermissionsService

scottbommarito:
Previously, the `PermissionsService` would
1 - Get a list of all permissions that the user has
2 - Check that list with the permissions required by the action

This PR now allows the `PermissionsService` to short-circuit and stop checking permissions when it finds that the user has a permission that's required.

E.g. if I am an owner of the package and this action is allowed if I have owner permissions, I don't need to check if I am a member of any organizations that own the package.

### DIFF
--- a/src/NuGetGallery/Services/PermissionsService.cs
+++ b/src/NuGetGallery/Services/PermissionsService.cs
@@ -78,7 +78,7 @@ namespace NuGetGallery
         {
             if (currentUser == null)
             {
-                return PermissionLevelsMatch(PermissionLevel.Anonymous, actionPermissionLevel);
+                return PermissionLevelsIntersect(PermissionLevel.Anonymous, actionPermissionLevel);
             }
 
             return HasPermission(
@@ -92,7 +92,7 @@ namespace NuGetGallery
         {
             if (currentPrincipal == null)
             {
-                return PermissionLevelsMatch(PermissionLevel.Anonymous, actionPermissionLevel);
+                return PermissionLevelsIntersect(PermissionLevel.Anonymous, actionPermissionLevel);
             }
 
             return HasPermission(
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Services/PermissionsService.cs
```diff
@@ -72,58 +71,55 @@ public static bool IsActionAllowed(User account, User currentUser, PermissionLev
...
-            if (entityOwners == null)
+            if ((entityOwners == null || !entityOwners.Any()) &&
+                PermissionLevelsMatch(PermissionLevel.Anonymous, actionPermissionLevel))
             {
-                return permissionLevel;
+                return true;
             }
 
-            if (isUserAdmin)
+            if (isUserAdmin &&
```
skofman1: you can get a minor perf improvement if you switch between the admin check and the owner check.
Admin performing an action is not as common as user performing an action, so you can avoid the check in  most cases.

src/NuGetGallery/Services/PermissionsService.cs
```diff
@@ -133,17 +129,24 @@ private static PermissionLevel GetPermissionLevel(IEnumerable<User> entityOwners
...
             {
-                permissionLevel |= PermissionLevel.OrganizationCollaborator;
+                return true;
             }
 
-            return permissionLevel;
+            return PermissionLevelsMatch(PermissionLevel.Anonymous, actionPermissionLevel);
+        }
+
+        private static bool PermissionLevelsMatch(PermissionLevel first, PermissionLevel second)
```
joelverhagen: nit: This is more like `PermissionLevelsIntersect` right? Match to me sounds like equality.

tests/NuGetGallery.Facts/Services/PermissionsServiceFacts.cs
```diff
@@ -103,10 +103,17 @@ private void CreateOrganizationOwnerAndAddUserAsMember(List<User> owners, User u
 
             private void AssertPermissionLevels(IEnumerable<User> owners, User user, PermissionLevel expectedLevel)
             {
-                Assert.Equal(expectedLevel, PermissionsService.GetPermissionLevel(owners, user));
+                for (int i = 0; i < Enum.GetValues(typeof(PermissionLevel)).Cast<int>().Max() * 2; i++)
```
joelverhagen: What's this math for?

joelverhagen:
Was there a perceived performance impact? How was this measured?

tests/NuGetGallery.Facts/Services/PermissionsServiceFacts.cs
chenriksson: Looks like the max value for permissions level (i.e., IsOwner | IsSiteAdmin | IsOrganizationAdmin | IsOrganizationCollaborator), and is also used for test data above. Maybe create a readonly class variable for it?

chenriksson: nit: My only concern with `ReturnsExpectedPermissionLevels` is whether it would be obvious from a test failure what the issue is. In the case of failure, will debugging be necessary to determine the issue? Looks comprehensive though.

scottbommarito:
@joelverhagen performance was never measured, but it seems very reasonable that in the majority of cases, skipping the organizations check when we know the user is an owner should be more performant.

tests/NuGetGallery.Facts/Services/PermissionsServiceFacts.cs
scottbommarito: yup, it's for iterating through all possible combinations of permissions

scottbommarito: added readonly variable

PR #5002, opened 2017-11-14
[Organizations] Support for UI upload for organizations

scottbommarito:
The changes to allow organization members to upload packages on behalf of their organization from the UI.

**only one possible owner**
image

**multiple possible owners**
image

(mostly ported from https://github.com/NuGet/NuGetGallery/pull/4963)

### DIFF
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -468,6 +468,7 @@ namespace NuGetGallery
                             id,
                             packageToPush,
                             packageStreamMetadata,
+                            user,
                             user);
 
                         await AutoCuratePackage.ExecuteAsync(package, packageToPush, commitChanges: false);
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -468,8 +468,8 @@ namespace NuGetGallery
                             id,
                             packageToPush,
                             packageStreamMetadata,
-                            user,
-                            user);
+                            owner: user,
+                            currentUser: user);
 
                         await AutoCuratePackage.ExecuteAsync(package, packageToPush, commitChanges: false);
 
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1591,7 +1591,8 @@ namespace NuGetGallery
         [ValidateAntiForgeryToken]
         public virtual async Task<JsonResult> CancelUpload()
         {
-            await _uploadFileService.DeleteUploadFileAsync(GetCurrentUser().Key);
+            var currentUser = GetCurrentUser();
+            await _uploadFileService.DeleteUploadFileAsync(currentUser.Key);
 
             return Json(null);
         }
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -162,7 +162,7 @@ namespace NuGetGallery
 
                     var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageMetadata.Id);
 
-                    if (!_reservedNamespaceService.IsPushAllowedOnBehalfOfOwner(packageMetadata.Id, currentUser, out var userOwnerMatchingNamespaces) ||
+                    if (existingPackageRegistration == null && !_reservedNamespaceService.IsPushAllowedOnBehalfOfOwner(packageMetadata.Id, currentUser, out var userOwnerMatchingNamespaces) ||
                         (existingPackageRegistration != null && !PermissionsService.IsActionAllowed(existingPackageRegistration, currentUser, PackageActions.UploadNewVersion)))
                     {
                         // This user no longer has the rights to upload to this package. Cancel this upload.
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -273,29 +273,27 @@ namespace NuGetGallery
                 }
 
                 var id = nuspec.GetId();
-                var packageRegistration = _packageService.FindPackageRegistrationById(id);
+                var existingPackageRegistration = _packageService.FindPackageRegistrationById(id);
+
                 // For a new package id verify if the user is allowed to use it.
-                if (packageRegistration == null)
+                if (existingPackageRegistration == null && !_reservedNamespaceService.IsPushAllowedOnBehalfOfOwners(id, currentUser, out var matchingNamespaces))
                 {
-                    if (!_reservedNamespaceService.IsPushAllowedOnBehalfOfOwner(id, currentUser, out var matchingNamespaces))
-                    {
-                        ModelState.AddModelError(
-                            string.Empty, string.Format(CultureInfo.CurrentCulture, Strings.UploadPackage_IdNamespaceConflict));
+                    ModelState.AddModelError(
+                        string.Empty, string.Format(CultureInfo.CurrentCulture, Strings.UploadPackage_IdNamespaceConflict));
 
-                        var version = nuspec.GetVersion().ToNormalizedString();
-                        _telemetryService.TrackPackagePushNamespaceConflictEvent(id, version, currentUser, User.Identity);
+                    var version = nuspec.GetVersion().ToNormalizedString();
+                    _telemetryService.TrackPackagePushNamespaceConflictEvent(id, version, currentUser, User.Identity);
 
-                        return Json(409, new string[] { string.Format(CultureInfo.CurrentCulture, Strings.UploadPackage_IdNamespaceConflict) });
-                    }
+                    return Json(409, new string[] { string.Format(CultureInfo.CurrentCulture, Strings.UploadPackage_IdNamespaceConflict) });
                 }
 
                 // For existing package ID verify that the current user has the rights to upload new versions
-                if (packageRegistration != null && !PermissionsService.IsActionAllowed(packageRegistration, currentUser, PackageActions.UploadNewVersion))
+                if (existingPackageRegistration != null && !PermissionsService.IsActionAllowedOnBehalfOfOwners(existingPackageRegistration, currentUser, AccountActions.UploadNewVersionOnBehalfOf))
                 {
                     ModelState.AddModelError(
-                        string.Empty, string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, packageRegistration.Id));
+                        string.Empty, string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, existingPackageRegistration.Id));
 
-                    return Json(409, new[] { string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, packageRegistration.Id) });
+                    return Json(409, new[] { string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, existingPackageRegistration.Id) });
                 }
 
                 var nuspecVersion = nuspec.GetVersion();
@@ -1422,13 +1420,26 @@ namespace NuGetGallery
                         return Json(400, new[] { message });
                     }
                 }
-                else if (!PermissionsService.IsActionAllowed(owner, currentUser, AccountActions.UploadNewIdOnBehalfOf))
+                else
                 {
-                    // The user is not allowed to upload a new ID on behalf of the owner specified in the form
-                    var message = string.Format(CultureInfo.CurrentCulture,
-                        Strings.UploadPackage_NewIdOnBehalfOfUserNotAllowed,
-                        currentUser.Username, owner.Username);
-                    return Json(400, new[] { message });
+                    if (!PermissionsService.IsActionAllowed(owner, currentUser, AccountActions.UploadNewIdOnBehalfOf))
+                    {
+                        // The user is not allowed to upload a new ID on behalf of the owner specified in the form
+                        var message = string.Format(CultureInfo.CurrentCulture,
+                            Strings.UploadPackage_NewIdOnBehalfOfUserNotAllowed,
+                            currentUser.Username, owner.Username);
+                        return Json(400, new[] { message });
+                    }
+
+                    if (!_reservedNamespaceService.IsPushAllowed(packageMetadata.Id, owner, out var userOwnerdMatchingNamespaces))
+                    {
+                        // The owner specified in the form is not allowed to push to a reserved namespace matching the new ID
+                        var version = packageMetadata.Version.ToNormalizedString();
+                        _telemetryService.TrackPackagePushNamespaceConflictEvent(packageMetadata.Id, version, currentUser, User.Identity);
+
+                        var message = string.Format(CultureInfo.CurrentCulture, Strings.UploadPackage_IdNamespaceConflict);
+                        return Json(409, new string[] { message });
+                    }
                 }
 
                 // update relevant database tables
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1402,47 +1402,10 @@ namespace NuGetGallery
                     return Json(400, new[] { message });
                 }
 
-                var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageMetadata.Id);
-                if (existingPackageRegistration != null)
+                var permissionsCheckResult = CheckPermissionsForVerifyPackage(packageMetadata, owner, currentUser);
+                if (permissionsCheckResult != null)
                 {
-                    if (!PermissionsService.IsActionAllowed(owner, currentUser, AccountActions.UploadNewVersionOnBehalfOf))
-                    {
-                        // The user is not allowed to upload a new version on behalf of the owner specified in the form
-                        var message = string.Format(CultureInfo.CurrentCulture,
-                            Strings.UploadPackage_NewVersionOnBehalfOfUserNotAllowed,
-                            currentUser.Username, owner.Username);
-                        return Json(400, new[] { message });
-                    }
-
-                    if (!PermissionsService.IsActionAllowed(existingPackageRegistration, owner, PackageActions.UploadNewVersion))
-                    {
-                        // The owner specified in the form is not allowed to upload a new version of the package
-                        var message = string.Format(CultureInfo.CurrentCulture,
-                            Strings.VerifyPackage_OwnerInvalid,
-                            owner.Username, existingPackageRegistration.Id);
-                        return Json(400, new[] { message });
-                    }
-                }
-                else
-                {
-                    if (!PermissionsService.IsActionAllowed(owner, currentUser, AccountActions.UploadNewIdOnBehalfOf))
-                    {
-                        // The user is not allowed to upload a new ID on behalf of the owner specified in the form
-                        var message = string.Format(CultureInfo.CurrentCulture,
-                            Strings.UploadPackage_NewIdOnBehalfOfUserNotAllowed,
-                            currentUser.Username, owner.Username);
-                        return Json(400, new[] { message });
-                    }
-
-                    if (!_reservedNamespaceService.IsPushAllowed(packageMetadata.Id, owner, out var userOwnerdMatchingNamespaces))
-                    {
-                        // The owner specified in the form is not allowed to push to a reserved namespace matching the new ID
-                        var version = packageMetadata.Version.ToNormalizedString();
-                        _telemetryService.TrackPackagePushNamespaceConflictEvent(packageMetadata.Id, version, currentUser, User.Identity);
-
-                        var message = string.Format(CultureInfo.CurrentCulture, Strings.UploadPackage_IdNamespaceConflict);
-                        return Json(409, new string[] { message });
-                    }
+                    return permissionsCheckResult;
                 }
 
                 // update relevant database tables
@@ -1544,6 +1507,57 @@ namespace NuGetGallery
             {
                 location = Url.Package(package.PackageRegistration.Id, package.NormalizedVersion)
             });
+        }
+
+        private JsonResult CheckPermissionsForVerifyPackage(PackageMetadata packageMetadata, User owner, User currentUser)
+        {
+            var packageId = packageMetadata.Id;
+            var packageVersion = packageMetadata.Version;
+
+            var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageId);
+            if (existingPackageRegistration != null)
+            {
+                if (!PermissionsService.IsActionAllowed(owner, currentUser, AccountActions.UploadNewVersionOnBehalfOf))
+                {
+                    // The user is not allowed to upload a new version on behalf of the owner specified in the form
+                    var message = string.Format(CultureInfo.CurrentCulture,
+                        Strings.UploadPackage_NewVersionOnBehalfOfUserNotAllowed,
+                        currentUser.Username, owner.Username);
+                    return Json(400, new[] { message });
+                }
+
+                if (!PermissionsService.IsActionAllowed(existingPackageRegistration, owner, PackageActions.UploadNewVersion))
+                {
+                    // The owner specified in the form is not allowed to upload a new version of the package
+                    var message = string.Format(CultureInfo.CurrentCulture,
+                        Strings.VerifyPackage_OwnerInvalid,
+                        owner.Username, existingPackageRegistration.Id);
+                    return Json(400, new[] { message });
+                }
+            }
+            else
+            {
+                if (!PermissionsService.IsActionAllowed(owner, currentUser, AccountActions.UploadNewIdOnBehalfOf))
+                {
+                    // The user is not allowed to upload a new ID on behalf of the owner specified in the form
+                    var message = string.Format(CultureInfo.CurrentCulture,
+                        Strings.UploadPackage_NewIdOnBehalfOfUserNotAllowed,
+                        currentUser.Username, owner.Username);
+                    return Json(400, new[] { message });
+                }
+
+                if (!_reservedNamespaceService.IsPushAllowed(packageId, owner, out var userOwnerdMatchingNamespaces))
+                {
+                    // The owner specified in the form is not allowed to push to a reserved namespace matching the new ID
+                    var version = packageVersion.ToNormalizedString();
+                    _telemetryService.TrackPackagePushNamespaceConflictEvent(packageId, version, currentUser, User.Identity);
+
+                    var message = string.Format(CultureInfo.CurrentCulture, Strings.UploadPackage_IdNamespaceConflict);
+                    return Json(409, new string[] { message });
+                }
+            }
+
+            return null;
         }
 
         private async Task<PackageArchiveReader> SafeCreatePackage(User currentUser, Stream uploadFile)
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -368,8 +368,6 @@ namespace NuGetGallery
                 {
                     _telemetryService.TraceException(ex);
 
-                    TempData["Message"] = ex.GetUserSafeMessage();
-
                     return Json(400, new[] { ex.GetUserSafeMessage() });
                 }
             }
@@ -1413,7 +1411,6 @@ namespace NuGetGallery
                 if (owner == null)
                 {
                     var message = string.Format(CultureInfo.CurrentCulture, Strings.VerifyPackage_UserNonExistent, formData.Edit.Owner);
-                    TempData["Message"] = message;
                     return Json(400, new[] { message });
                 }
 
@@ -1426,7 +1423,6 @@ namespace NuGetGallery
                         var message = string.Format(CultureInfo.CurrentCulture,
                             Strings.UploadPackage_NewVersionOnBehalfOfUserNotAllowed,
                             currentUser.Username, owner.Username);
-                        TempData["Message"] = message;
                         return Json(400, new[] { message });
                     }
 
@@ -1436,7 +1432,6 @@ namespace NuGetGallery
                         var message = string.Format(CultureInfo.CurrentCulture,
                             Strings.VerifyPackage_OwnerInvalid,
                             owner.Username, existingPackageRegistration.Id);
-                        TempData["Message"] = message;
                         return Json(400, new[] { message });
                     }
                 }
@@ -1446,7 +1441,6 @@ namespace NuGetGallery
                     var message = string.Format(CultureInfo.CurrentCulture,
                         Strings.UploadPackage_NewIdOnBehalfOfUserNotAllowed,
                         currentUser.Username, owner.Username);
-                    TempData["Message"] = message;
                     return Json(400, new[] { message });
                 }
 
@@ -1465,8 +1459,6 @@ namespace NuGetGallery
                 catch (InvalidPackageException ex)
                 {
                     _telemetryService.TraceException(ex);
-
-                    TempData["Message"] = ex.Message;
 
                     return Json(400, new[] { ex.GetUserSafeMessage() });
                 }
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -159,8 +159,9 @@ namespace NuGetGallery
                     }
 
                     model.IsUploadInProgress = true;
-                    
-                    var verifyRequest = new VerifyPackageRequest(packageMetadata, GetPossibleOwnersForUpload(packageMetadata.Id));
+
+                    var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageMetadata.Id);
+                    var verifyRequest = new VerifyPackageRequest(packageMetadata, GetPossibleOwnersForUpload(packageMetadata.Id, existingPackageRegistration));
 
                     model.InProgressUpload = verifyRequest;
                 }
@@ -196,6 +197,8 @@ namespace NuGetGallery
                 ModelState.AddModelError(String.Empty, Strings.UploadFileMustBeNuGetPackage);
                 return Json(400, new[] { Strings.UploadFileMustBeNuGetPackage });
             }
+
+            PackageRegistration existingPackageRegistration;
 
             using (var uploadStream = uploadFile.InputStream)
             {
@@ -273,7 +276,7 @@ namespace NuGetGallery
                 }
 
                 var id = nuspec.GetId();
-                var existingPackageRegistration = _packageService.FindPackageRegistrationById(id);
+                existingPackageRegistration = _packageService.FindPackageRegistrationById(id);
 
                 // For a new package id verify if the user is allowed to use it.
                 if (existingPackageRegistration == null && !_reservedNamespaceService.IsPushAllowedOnBehalfOfOwners(id, currentUser, out var matchingNamespaces))
@@ -360,7 +363,7 @@ namespace NuGetGallery
                 }
             }
 
-            var model = new VerifyPackageRequest(packageMetadata, GetPossibleOwnersForUpload(packageMetadata.Id));
+            var model = new VerifyPackageRequest(packageMetadata, GetPossibleOwnersForUpload(packageMetadata.Id, existingPackageRegistration));
 
             return Json(model);
         }
@@ -1665,11 +1668,11 @@ namespace NuGetGallery
         /// Determines the possible owners for a package that is being uploaded by the current user.
         /// Assumes the current user has permissions to upload the package.
         /// </summary>
+        /// <param name="packageId">The package ID being uploaded to.</param>
         /// <param name="existingPackageRegistration">The package registration being uploaded to.</param>
-        internal IEnumerable<User> GetPossibleOwnersForUpload(string packageId)
+        internal IEnumerable<User> GetPossibleOwnersForUpload(string packageId, PackageRegistration existingPackageRegistration)
         {
             IEnumerable<User> possibleOwners;
-            var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageId);
             var currentUser = GetCurrentUser();
 
             if (existingPackageRegistration != null)
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1495,13 +1495,20 @@ namespace NuGetGallery
 
             var currentUser = GetCurrentUser();
 
+            // Check that the owner specified in the form is valid
+            var owner = _userService.FindByUsername(formData.Owner);
+
+            if (owner == null)
+            {
+                var message = string.Format(CultureInfo.CurrentCulture, Strings.VerifyPackage_UserNonExistent, formData.Owner);
+                return Json(400, new[] { message });
+            }
+
             Package package;
             using (Stream uploadFile = await _uploadFileService.GetUploadFileAsync(currentUser.Key))
             {
                 if (uploadFile == null)
                 {
-                    TempData["Message"] = Strings.VerifyPackage_UploadNotFound;
-
                     return Json(400, new[] { Strings.VerifyPackage_UploadNotFound });
                 }
 
@@ -1525,8 +1532,6 @@ namespace NuGetGallery
                         && String.Equals(packageMetadata.Version.ToFullStringSafe(), formData.Version, StringComparison.OrdinalIgnoreCase)
                         && String.Equals(packageMetadata.Version.OriginalVersion, formData.OriginalVersion, StringComparison.OrdinalIgnoreCase)))
                     {
-                        TempData["Message"] = Strings.VerifyPackage_PackageFileModified;
-
                         return Json(400, new[] { Strings.VerifyPackage_PackageFileModified });
                     }
                 }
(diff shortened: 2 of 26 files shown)


src/NuGetGallery/Strings.Designer.cs
```diff
@@ -19,7 +19,7 @@ namespace NuGetGallery {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "15.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "4.0.0.0")]
```
skofman1: do we need this change?

scottbommarito: honestly I see it change with every PR that touches it...I'd leave it alone and let it do its thing

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -160,7 +160,17 @@ public virtual JsonResult UploadPackageProgress()
 
                     model.IsUploadInProgress = true;
 
-                    var verifyRequest = new VerifyPackageRequest(packageMetadata);
+                    var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageMetadata.Id);
+
+                    if (existingPackageRegistration == null && !_reservedNamespaceService.IsPushAllowedOnBehalfOfOwner(packageMetadata.Id, currentUser, out var userOwnerMatchingNamespaces) ||
```
skofman1: please open a bug and send a separate PR

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -277,10 +287,7 @@ public virtual JsonResult UploadPackageProgress()
                 // For a new package id verify if the user is allowed to use it.
                 if (packageRegistration == null)
                 {
-                    var isPushAllowed = _reservedNamespaceService
-                        .IsPushAllowed(id, currentUser, out IReadOnlyCollection<ReservedNamespace> matchingNamespaces);
-
-                    if (!isPushAllowed)
+                    if (!_reservedNamespaceService.IsPushAllowedOnBehalfOfOwner(id, currentUser, out var matchingNamespaces))
```
skofman1: please don't refactor unrelated code

scottbommarito: this is a different method call--`IsPushAllowed` != `IsPushAllowedOnBehalfOfOwner`

src/NuGetGallery/Services/ReservedNamespaceActions.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace NuGetGallery
+{
+    public static class ReservedNamespaceActions
+    {
+        /// <summary>
+        /// The user can push to this reserved namespace.
+        /// </summary>
+        public static PermissionLevel PushToReservedNamespace =
+            PermissionLevel.Owner |
+            PermissionLevel.SiteAdmin;
```
skofman1: Should a site admin be here?

src/NuGetGallery/Services/AccountActions.cs
```diff
@@ -9,10 +9,34 @@ namespace NuGetGallery
...
+        /// The user can upload new package IDs on behalf of the account using the UI.
+        /// </summary>
+        public static PermissionLevel UploadNewIdOnBehalfOf =
+            PermissionLevel.Owner |
+            PermissionLevel.OrganizationAdmin;
+
+        /// <summary>
+        /// The user can upload new versions of an existing package on behalf of the account using the UI.
+        /// </summary>
+        public static PermissionLevel UploadNewVersionOnBehalfOf =
```
skofman1: Isn't this duplicate of PackageActions.UploadNewVerison?

src/NuGetGallery/Services/AccountActions.cs
```diff
@@ -9,10 +9,34 @@ namespace NuGetGallery
...
+            PermissionLevel.Owner |
+            PermissionLevel.OrganizationAdmin |
+            PermissionLevel.OrganizationCollaborator;
+
+        /// <summary>
+        /// The user can push to reserved namespaces on behalf of another owner.
+        /// </summary>
+        public static PermissionLevel PushToReservedNamespaceOnBehalfOf =
```
skofman1: 1. SiteAdmin shouldn't be able to push to a reserved namespace on someone's behalf
2. why do we need a special permission for reserved namespace push? If you are part of an organization that owns the namespace, you can push. So the permission for the namespace is derived from the org permission

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1618,6 +1680,44 @@ protected internal virtual PackageArchiveReader CreatePackage(Stream stream)
...
 
+        /// <summary>
+        /// Determines the possible owners for a package that is being uploaded by the current user.
+        /// Assumes the current user has permissions to upload the package.
+        /// </summary>
+        /// <param name="id">The ID of the package being uploaded</param>
+        internal IEnumerable<User> GetPossibleOwnersForUpload(string id)
+        {
+            IEnumerable<User> possibleOwners;
+            var existingPackageRegistration = _packageService.FindPackageRegistrationById(id);
```
skofman1: reduce the number of DB calls in the code. They are expensive. And call to FindPackageRegistrationById is already done by UploadPackage()

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1618,6 +1680,44 @@ protected internal virtual PackageArchiveReader CreatePackage(Stream stream)
...
+            var currentUser = GetCurrentUser();
+
+            if (existingPackageRegistration != null)
+            {
+                possibleOwners = existingPackageRegistration.Owners
+                    .Where(u => PermissionsService.IsActionAllowed(u, currentUser, AccountActions.UploadNewVersionOnBehalfOf));
+                if (!possibleOwners.Any())
+                {
+                    // If the user has the right to upload to the package but they are not able to upload as any of the existing owners (e.g. site admin), allow the user to upload as themselves.
```
skofman1: site admins shouldn't be able to upload new versions of existing packages that they don't own

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -69,13 +69,13 @@ public void EnsureValid(PackageArchiveReader packageArchiveReader)
...
         /// <param name="packageStreamMetadata">The <see cref="PackageStreamMetadata"/> instance providing metadata about the package stream.</param>
-        /// <param name="user">The <see cref="User"/> creating the package.</param>
+        /// <param name="owner">The <see cref="User"/> creating the package.</param>
         /// <param name="commitChanges"><c>True</c> to commit the changes to the data store and notify the indexing service; otherwise <c>false</c>.</param>
         /// <returns>Returns the created <see cref="Package"/> entity.</returns>
         /// <exception cref="InvalidPackageException">
         /// This exception will be thrown when a package metadata property violates a data validation constraint.
         /// </exception>
-        public async Task<Package> CreatePackageAsync(PackageArchiveReader nugetPackage, PackageStreamMetadata packageStreamMetadata, User user, bool isVerified)
+        public async Task<Package> CreatePackageAsync(PackageArchiveReader nugetPackage, PackageStreamMetadata packageStreamMetadata, User owner, User currentUser, bool isVerified)
```
skofman1: >User owner, User currentUser [](start = 134, length = 28)

I wonder if we could improve this.. we are duplicating owner and currentUser in multiple areas of the code. Can we add to the User class a property of "OnBehalfOf"? Other ideas?

src/NuGetGallery/Services/ReservedNamespaceActions.cs
scottbommarito: I don't see any problem with a site admin having the ability to upload to a reserved namespace. I've set up a trend of allowing site admins to do anything to any package (but not on behalf of other users). I think allowing this could be potentially useful to admins in the future in case some scenario came up that required us to reserved a package on behalf of some user.

Note: we do need to determine if pushing to a reserved namespace as a site admin marks the package verified.

src/NuGetGallery/Strings.Designer.cs
skofman1: ok

src/NuGetGallery/Controllers/PackagesController.cs
skofman1: ok

src/NuGetGallery/Services/AccountActions.cs
scottbommarito: Almost!

`PackageActions.UploadNewVersion` is:
given this package, can I upload a new version to it?

`AccountActions.UploadNewVersionOnBehalfOf` is:
given this user, can I upload a new version on behalf of them?

The UI upload flow is essentially:

Does current user have `PackageActions.UploadNewVersion` for this package?
If so, continue.
If not, then reject the upload.

Then, the package must be uploaded on behalf of an owner for which the current user has `AccountActions.UploadNewVersionOnBehalfOf` for.

src/NuGetGallery/Services/ReservedNamespaceActions.cs
skofman1: this is a functionality that doesn't exist right now. It requires a discussion. Please open an issue to track this suggestion and remove from this PR.

src/NuGetGallery/Services/AccountActions.cs
scottbommarito: 1. Yep, woops.

2. We need both permissions.
`ReservedNamespaceActions.PushToReservedNamespace` means THIS USER can push a new package registration to the namespace. We want organization admins to be able to push to a reserved namespace ON BEHALF OF an organization--they shouldn't be able to push to it themselves.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: this functionality was already added in a previous PR.

I can remove it, but it would require some test changes and we will want to have the discussion anyway.

src/NuGetGallery/Services/PackageService.cs
scottbommarito: Personally I think this is fine. To be honest I'm slightly concerned that having an `OnBehalfOf` property could potentially make it harder to understand the distinction between the user performing the action and the user who owns the package. Fortunately I don't think this "owner" and "currentUser" pattern needs to be added to anywhere else in the code.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1032,7 +1039,10 @@ public virtual ActionResult Delete(string id, string version)
             // Create edit model from the latest pending edit.
             var pendingMetadata = _editPackageService.GetPendingMetadata(package);
 
-            model.Edit = new EditPackageVersionRequest(package, pendingMetadata);
+            // No point in allowing users to specify the owner who edited the package: just show the first.
+            var possibleOwners = GetPossibleOwnersForUpload(package.PackageRegistration.Id).Take(1);
```
chenriksson: GetPossibleOwnersForUpload is only necessary for registration owner, right? Should edit just use the current user?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1618,6 +1680,44 @@ protected internal virtual PackageArchiveReader CreatePackage(Stream stream)
             }
         }
 
+        /// <summary>
+        /// Determines the possible owners for a package that is being uploaded by the current user.
+        /// Assumes the current user has permissions to upload the package.
+        /// </summary>
+        /// <param name="id">The ID of the package being uploaded</param>
+        internal IEnumerable<User> GetPossibleOwnersForUpload(string id)
```
chenriksson: GetPossibleOwnersForUpload should only be necessary for new registration, right?

Do you need to call it anywhere other than UploadPackage(HttpPostedFileBase) when it's a new registration? If not, then this method could be simplified... don't need the block where existing registration exists.

src/NuGetGallery/RequestModels/EditPackageVersionRequest.cs
```diff
@@ -23,7 +25,7 @@ public EditPackageVersionRequest()
         {
         }
 
-        public EditPackageVersionRequest(PackageMetadata packageMetadata)
+        public EditPackageVersionRequest(PackageMetadata packageMetadata, IEnumerable<User> possibleOwners)
```
chenriksson: Is possibleOwners needed on EditPackageVersionRequest, if we're just defaulting to current or first user?

src/NuGetGallery/Services/AccountActions.cs
skofman1: > Does current user have PackageActions.UploadNewVersion for this package? 

The current user gets his permission to upload the package because he is a member of an org that has this permission. Meaning, his permissions are completely derived. Also, notice that the value of `PackageActions.UploadNewVersion` is identical to `AccountActions.UploadNewVersionOnBehalfOf`

src/NuGetGallery/Services/AccountActions.cs
skofman1: Same as above, admins shouldn't be able to push on someone's behalf.

src/NuGetGallery/Controllers/PackagesController.cs
skofman1: Please remove. This is a new functionality that wasn't discussed.

src/NuGetGallery/Strings.resx
```diff
@@ -618,4 +618,19 @@ For more information, please contact '{2}'.</value>
   <data name="AccountDelete_Success" xml:space="preserve">
     <value>The account:{0} was deleted succesfully.</value>
   </data>
+  <data name="UploadPackage_NewIdOnBehalfOfUserNotAllowed" xml:space="preserve">
+    <value>User '{0}' does not have the rights to upload a package with a new ID on behalf of user '{1}'!</value>
```
chenriksson: nit: Would not use '!' at the end of user messages.

src/NuGetGallery/Strings.Designer.cs
chenriksson: I suspect this is due to switching back and forth between d15rel and other versions. We should probably stick with the d15rel version.

See https://github.com/dotnet/project-system/issues/2614

src/NuGetGallery/Services/ReservedNamespaceActions.cs
scottbommarito: https://github.com/NuGet/NuGetGallery/issues/5019

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: You're probably right--I'll need to look at this and see if returning current user has any impact on the flow.

src/NuGetGallery/Services/AccountActions.cs
scottbommarito: https://github.com/NuGet/NuGetGallery/issues/5019

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: https://github.com/NuGet/NuGetGallery/issues/5019

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1397,13 +1407,57 @@ private string GetSecurePushPolicyDescriptions()
...
+                    {
+                        // The user is not allowed to upload a new version on behalf of the owner specified in the form
+                        var message = string.Format(CultureInfo.CurrentCulture,
+                            Strings.UploadPackage_NewVersionOnBehalfOfUserNotAllowed,
+                            currentUser.Username, owner.Username);
+                        TempData["Message"] = message;
+                        return Json(400, new[] { message });
+                    }
+
+                    if (!PermissionsService.IsActionAllowed(existingPackageRegistration, owner, PackageActions.UploadNewVersion))
```
scottbommarito: This check shouldn't succeed if `owner` is not a direct owner of the package, but it does.

`PackageActions.UploadNewVersion` and its uses should be rewritten to reflect this better.

src/NuGetGallery/Services/AccountActions.cs
scottbommarito: Discussed offline. There's confusion between `PackageActions.UploadNewVersion` and `AccountActions.UploadNewVersionOnBehalfOf`. Most uses of `PackageActions.UploadNewVersion` should be replaced with a check that checks that the user can do the action on behalf of the owners of the package.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: https://github.com/NuGet/NuGetGallery/issues/5043

src/NuGetGallery/Controllers/PackagesController.cs

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: fixed

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Our package upload flows require both the owner of the package and the current user, so it must consider both cases.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: This flow needs to stay the same, regardless--this is not the method to check ownership of the package registration.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: actually, adding this back in because i removed a call somewhere else, and not having this here makes the code very ugly

src/NuGetGallery/RequestModels/EditPackageVersionRequest.cs
scottbommarito: moved it to `VerifyPackageRequest` and removed any changes to the edit flow

src/NuGetGallery/Strings.resx
scottbommarito: fixed

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -361,13 +358,11 @@ public virtual JsonResult UploadPackageProgress()
                 {
                     _telemetryService.TraceException(ex);
 
-                    TempData["Message"] = ex.GetUserSafeMessage();
```
skofman1: are you sure TempData["Message"] is not used?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1125,9 +1120,9 @@ public virtual Task<ActionResult> RejectPendingOwnershipRequest(string id, strin
             {
                 return HttpNotFound();
             }
-            
+
             var user = _userService.FindByUsername(username);
-            if (!PermissionsService.IsActionAllowed(user, GetCurrentUser(), AccountActions.AcceptPackageOwnershipOnBehalfOf))
+            if (!PermissionsService.IsActionAllowed(user, GetCurrentUser(), AccountActions.ManagePackageOwnersOnBehalfOf))
```
skofman1: Shouldn't this be part of ownership management PR?

scottbommarito: it is...haven't rebased this yet

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: yup, it was a bug it was here before

ideally i should have removed it with the verify package UI fix i merged on Friday, but I don't think it's too risky to merge with this PR

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1618,6 +1661,44 @@ protected internal virtual PackageArchiveReader CreatePackage(Stream stream)
...
 
+        /// <summary>
+        /// Determines the possible owners for a package that is being uploaded by the current user.
+        /// Assumes the current user has permissions to upload the package.
+        /// </summary>
+        /// <param name="existingPackageRegistration">The package registration being uploaded to.</param>
+        internal IEnumerable<User> GetPossibleOwnersForUpload(string packageId)
+        {
+            IEnumerable<User> possibleOwners;
+            var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageId);
```
skofman1: you are still calling FindPackageRegistrationById twice for UploadPackage(). This can be easily fixed by adding an another method GetPossibleOwnersForUpload(Package package) that will be called by GetPossibleOwnersForUpload(string packageId)

src/NuGetGallery/RequestModels/EditPackageVersionRequest.cs
```diff
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
```
skofman1: there are no changes in this file so you can revert it

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
```diff
@@ -31,6 +31,9 @@ public VerifyPackageRequest(PackageMetadata packageMetadata)
             Dependencies = new DependencySetsViewModel(packageMetadata.GetDependencyGroups().AsPackageDependencyEnumerable());
             DevelopmentDependency = packageMetadata.GetValueFromMetadata("developmentDependency");
             Edit = new EditPackageVersionRequest(packageMetadata);
+
+            PossibleOwners = possibleOwners.Select(u => u.Username);
+            Owner = PossibleOwners.First();
```
skofman1: there are multiple enumerations over possibleOwners. Please change the type to List/Array.

src/NuGetGallery/Services/PackageActions.cs
```diff
@@ -11,32 +11,29 @@ public static class PackageActions
...
         /// <summary>
-        /// The user can upload new versions of the package.
+        /// The user can upload new versions of the package from the UI.
         /// </summary>
-        public static PermissionLevel UploadNewVersion = 
-            PermissionLevel.Owner | 
-            PermissionLevel.OrganizationAdmin | 
-            PermissionLevel.SiteAdmin | 
-            PermissionLevel.OrganizationCollaborator;
+        public static PermissionLevel UploadNewVersion =
```
skofman1: There is a discrepancy between permissions for upload (only owner) and unlist/edit (owner+org admin, etc.). Should we fix this? (in a different PR)

src/NuGetGallery/Services/ReservedNamespaceActions.cs
skofman1: Please remove SiteAdmin here

src/NuGetGallery/Services/PackageActions.cs
scottbommarito: Potentially yes.

The difference is that unlist/edit don't need to be done on behalf of the owner in the UI. We could rewrite this to so that unlist and edit have "on behalf of" permissions and the value of the owner we're doing the action on behalf of is just unused.

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
scottbommarito: this is an enumeration of `PossibleOwners` not `possibleOwners`

I will add a `ToList` to `PossibleOwners` because it is also enumerated by the UI.

src/NuGetGallery/Services/ReservedNamespaceActions.cs
scottbommarito: done

src/NuGetGallery/RequestModels/EditPackageVersionRequest.cs
scottbommarito: done

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Talked offline--it's not that simple. Either this method needs to accept a `string` and a `PackageRegistration` or it needs to accept a `string` and make a DB request for the `PackageRegistration`. I would vastly prefer to do the latter. I did some investigations and unfortunately I can't determine whether or not EF is caching the request.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -468,6 +468,7 @@ public virtual Task<ActionResult> CreatePackagePost()
                             id,
                             packageToPush,
                             packageStreamMetadata,
+                            user,
                             user);
```
chenriksson: nit: parameter labels would be useful here

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -273,32 +273,27 @@ public virtual JsonResult UploadPackageProgress()
                 }
 
                 var id = nuspec.GetId();
-                var packageRegistration = _packageService.FindPackageRegistrationById(id);
+                var existingPackageRegistration = _packageService.FindPackageRegistrationById(id);
+
                 // For a new package id verify if the user is allowed to use it.
-                if (packageRegistration == null)
+                if (existingPackageRegistration == null && !_reservedNamespaceService.IsPushAllowedOnBehalfOfOwners(id, currentUser, out var matchingNamespaces))
```
chenriksson: nit: wrap long lines (does the team have a recommended column width - 120?)

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: Anything preventing you from rebasing now? Would be useful for reviewers if we're seeing changes not in the PR.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -160,7 +160,8 @@ public virtual JsonResult UploadPackageProgress()
 
                     model.IsUploadInProgress = true;
 
-                    var verifyRequest = new VerifyPackageRequest(packageMetadata);
+                    var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageMetadata.Id);
```
ryuyu: Should we check for null here? Is it even possible?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1397,13 +1393,66 @@ private string GetSecurePushPolicyDescriptions()
...
+                    if (!_reservedNamespaceService.IsPushAllowed(packageMetadata.Id, owner, out var userOwnerdMatchingNamespaces))
+                    {
+                        // The owner specified in the form is not allowed to push to a reserved namespace matching the new ID
+                        var version = packageMetadata.Version.ToNormalizedString();
+                        _telemetryService.TrackPackagePushNamespaceConflictEvent(packageMetadata.Id, version, currentUser, User.Identity);
+
+                        var message = string.Format(CultureInfo.CurrentCulture, Strings.UploadPackage_IdNamespaceConflict);
+                        return Json(409, new string[] { message });
+                    }
+                }
```
chenriksson: nit: method is getting long... could these permissions checks be extracted into a separate method?

src/NuGetGallery/Controllers/PackagesController.cs
ryuyu: It was initially left in place to minimize server side changes that were being made with the UI refresh. If we were able to verify that it is not used, it should be safe to remove.

scottbommarito: When I did the work on the Verify Package UI I determined that all these messages are now correctly piped through the UI without `TempData["Message"]`, should be safe to remove now.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1397,13 +1393,66 @@ private string GetSecurePushPolicyDescriptions()
...
+                    if (!PermissionsService.IsActionAllowed(owner, currentUser, AccountActions.UploadNewIdOnBehalfOf))
+                    {
+                        // The user is not allowed to upload a new ID on behalf of the owner specified in the form
+                        var message = string.Format(CultureInfo.CurrentCulture,
+                            Strings.UploadPackage_NewIdOnBehalfOfUserNotAllowed,
+                            currentUser.Username, owner.Username);
+                        return Json(400, new[] { message });
+                    }
+
+                    if (!_reservedNamespaceService.IsPushAllowed(packageMetadata.Id, owner, out var userOwnerdMatchingNamespaces))
```
chenriksson: This wasn't called in VerifyPackage before; only from UploadPackage?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -160,7 +160,8 @@ public virtual JsonResult UploadPackageProgress()
 
                     model.IsUploadInProgress = true;
 
-                    var verifyRequest = new VerifyPackageRequest(packageMetadata);
+                    var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageMetadata.Id);
+                    var verifyRequest = new VerifyPackageRequest(packageMetadata, GetPossibleOwnersForUpload(packageMetadata.Id, existingPackageRegistration));
```
chenriksson: Can you provide a screenshot of the UI changes?

I'm trying to understand if/why the possible owners are necessary before the user has selected the file upload. Should the drop-down only appear after the file upload finishes?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1618,6 +1664,45 @@ protected internal virtual PackageArchiveReader CreatePackage(Stream stream)
...
+
+            if (existingPackageRegistration != null)
+            {
+                possibleOwners = existingPackageRegistration.Owners
+                    .Where(u => PermissionsService.IsActionAllowed(u, currentUser, AccountActions.UploadNewVersionOnBehalfOf))
+                    .Where(u => PermissionsService.IsActionAllowed(existingPackageRegistration, u, PackageActions.UploadNewVersion));
+                if (!possibleOwners.Any())
+                {
+                    // If the user has the right to upload to the package but they are not able to upload as any of the existing owners, allow the user to upload as themselves.
+                    possibleOwners = new User[] { currentUser };
```
chenriksson: When would this happen? Is it a scenario like SiteAdmin being able to upload on behalf of?

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: `GetPossibleOwnersForUpload` checks it

src/NuGetGallery/Services/ReservedNamespaceActions.cs
chenriksson: SiteAdmin is still here? Change not pushed yet?

src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -365,16 +365,33 @@ public static void ValidateNamespace(string value)
         }
 
         public bool IsPushAllowed(string id, User user, out IReadOnlyCollection<ReservedNamespace> userOwnedMatchingNamespaces)
+        {
+            return IsPushAllowedInternal(
+                id,
+                user,
+                rn => PermissionsService.IsActionAllowed(rn, user, ReservedNamespaceActions.PushToReservedNamespace),
```
chenriksson: Why does the namespace service need to do a permissions check? Isn't this already covered by PackagesController?

src/NuGetGallery/Strings.resx
```diff
@@ -618,4 +618,19 @@ For more information, please contact '{2}'.</value>
   <data name="AccountDelete_Success" xml:space="preserve">
     <value>The account:{0} was deleted succesfully.</value>
   </data>
+  <data name="UploadPackage_NewIdOnBehalfOfUserNotAllowed" xml:space="preserve">
+    <value>User '{0}' does not have the rights to upload a package with a new ID on behalf of user '{1}'.</value>
```
chenriksson: Relationship 'on behalf of' is only applicable for organizations. Should this say ... 'on behalf of organization X'

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1397,13 +1393,66 @@ private string GetSecurePushPolicyDescriptions()
...
+                    }
+
+                    if (!PermissionsService.IsActionAllowed(existingPackageRegistration, owner, PackageActions.UploadNewVersion))
+                    {
+                        // The owner specified in the form is not allowed to upload a new version of the package
+                        var message = string.Format(CultureInfo.CurrentCulture,
+                            Strings.VerifyPackage_OwnerInvalid,
+                            owner.Username, existingPackageRegistration.Id);
+                        return Json(400, new[] { message });
+                    }
```
chenriksson: Can `AccountActions.UploadNewVersionOnBehalfOf` and `PackageActions.UploadNewVersion` not be combined?

I think callers would look cleaner with a single API like:
`PermissionsService.IsActionAllowed(package, owner, currentUser, actions)`
OR
`PermissionsService.IsActionAllowed(package, ownerProxy, actions)`

chenriksson: Also, why does 'push new version' have 2 checks but 'push new' (below) only has one?

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Yes. We took out the `SiteAdmin` scenario, but this code is still necessary in case we add it back in or add a similar similar scenario. This method is not a permissions check.

src/NuGetGallery/Services/ReservedNamespaceService.cs
scottbommarito: `ReservedNamespaceService.IsPushAllowed` was an existing permissions check that checked if the current user is able to push to the reserved namespace. I refactored it to use the `PermissionsService` and also added `IsPushAllowedOnBehalfOfOwners` which checks if the current user is able to push to the reserved namespace on behalf of any of its owners.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: yup, but it's necessary now because of the ability to change the owner through the form.

Also note that this is `IsPushAllowed` and not `IsPushAllowedOnBehalfOf`, which is called in `UploadPackage`.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: It only appears after the file upload finishes. This is the path for pending upload that wasn't finished.

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: done

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Combining the two methods would be cleaner, but ultimately would create additional syntax issues because we want to show different messages depending on which failure it is.

Push new ID only has one check because we don't need to ask the package registration if the owner can push to it--it doesn't exist.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: yup

src/NuGetGallery/Services/ReservedNamespaceActions.cs
scottbommarito: my bad, didn't push until now

src/NuGetGallery/Strings.resx
scottbommarito: This is the error message for any user provided in the upload form. Somebody could put any user in the form and the error message would say it was a organization.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I think this is the only change shared between the two.

I wanted to fix all the pr comments before i rebase because i'm concerned rebasing could be unpleasant. I will do it soon tho.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Note: this message used to state `SiteAdmin` specifically, but no longer does because it's not true.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: added screenshots

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: this line changed so no longer a concern

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1542,13 +1538,29 @@ private string GetSecurePushPolicyDescriptions()
...
                 };
 
+                // Check that the owner specified in the form is valid
+                var owner = _userService.FindByUsername(formData.Owner);
+
+                if (owner == null)
+                {
+                    var message = string.Format(CultureInfo.CurrentCulture, Strings.VerifyPackage_UserNonExistent, formData.Owner);
+                    return Json(400, new[] { message });
+                }
```
skofman1: if we move this check to the beginning of the method we can fail much faster (and cheaper)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1770,6 +1833,45 @@ protected internal virtual PackageArchiveReader CreatePackage(Stream stream)
...
+            var currentUser = GetCurrentUser();
+
+            if (existingPackageRegistration != null)
+            {
+                possibleOwners = existingPackageRegistration.Owners
+                    .Where(u => PermissionsService.IsActionAllowed(u, currentUser, AccountActions.UploadNewVersionOnBehalfOf))
+                    .Where(u => PermissionsService.IsActionAllowed(existingPackageRegistration, u, PackageActions.UploadNewVersion));
+                if (!possibleOwners.Any())
+                {
+                    // If the user has the right to upload to the package but they are not able to upload as any of the existing owners, allow the user to upload as themselves.
```
skofman1: lets add a comment to replace this with a throw after we have cancel logic (and reference the issue opened)

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
```diff
@@ -44,6 +47,10 @@ public VerifyPackageRequest(PackageMetadata packageMetadata)
         /// The non-normalized, unmodified, original version as defined in the nuspec.
         /// </summary>
         public string OriginalVersion { get; set; }
+
+        public string Owner { get; set; }
+        public IEnumerable<string> PossibleOwners { get; set; }
```
skofman1: change this to IList

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
```diff
@@ -31,6 +31,9 @@ public VerifyPackageRequest(PackageMetadata packageMetadata)
             Dependencies = new DependencySetsViewModel(packageMetadata.GetDependencyGroups().AsPackageDependencyEnumerable());
             DevelopmentDependency = packageMetadata.GetValueFromMetadata("developmentDependency");
             Edit = new EditPackageVersionRequest(packageMetadata);
+
+            PossibleOwners = possibleOwners.Select(u => u.Username).ToList();
+            Owner = PossibleOwners.First();
```
skofman1: do we need Owner? Can't we use PossibleOwners everywhere?

scottbommarito: I refactored the code to use `PossibleOwners` more, but `Owner` is needed for the actual submission of the form (`VerifyPackageRequest` is both the model of the view and the model of the request).

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
scottbommarito: done

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: rewrote the comment to better reflect the scenario--this method return doesn't "allow" the user to upload as themselves, it just shows the user as a possible owner in the form.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: will do

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: after determining that EF does not cache, made it accept a `string` and a `PackageRegistration`

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: After discussing offline, we planned on removing this but realized that it is still used for pending uploads that are no longer valid (the ID/version has been snatched by someone else).

scottbommarito:
Closing this because it will be redone with the permissions service 2.0

Issue #4995, opened 2017-11-13
Push via ApiController should ensure that ApiKey scope is still valid


chenriksson:
In addition to validating scopes, it should also use the permission service to verify that:
- Owner scope still matches a package owner
- Organization owner still has permissions for the action (i.e., if push new, should still be an org admin)

Issue #4994, opened 2017-11-13
Push new from ApiController should use owner scope as registration owner


chenriksson:
Owner scope is evaluated, but it should also be used to set the package owner for new registrations.

scottbommarito:
duplicate of https://github.com/NuGet/NuGetGallery/issues/4892

PR #5003, opened 2017-11-14
[Organizations] Support for API upload for organizations

scottbommarito:
The changes to allow organization members to upload packages on behalf of their organization from the API.

(mostly ported from https://github.com/NuGet/NuGetGallery/pull/4963)

### DIFF
--- a/src/NuGetGallery/Authentication/NuGetScopes.cs
+++ b/src/NuGetGallery/Authentication/NuGetScopes.cs
@@ -34,11 +34,6 @@ namespace NuGetGallery.Authentication
 
         public static bool IsActionAllowedOnSubjectByOwner(PackageRegistration packageRegistration, User owner, params string[] requestedActions)
         {
-            if (packageRegistration == null)
-            {
-                return true;
-            }
-
             return requestedActions
                 .Select(GetRequiredPackageAction)
                 .Any(p => PermissionsService.IsActionAllowed(packageRegistration, owner, p));
--- a/src/NuGetGallery/Authentication/NuGetScopes.cs
+++ b/src/NuGetGallery/Authentication/NuGetScopes.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Linq;
+
 namespace NuGetGallery.Authentication
 {
     public static class NuGetScopes
@@ -28,6 +30,62 @@ namespace NuGetGallery.Authentication
             }
 
             return Strings.ScopeDescription_Unknown;
+        }
+
+        public static bool IsActionAllowedOnSubjectByOwner(PackageRegistration packageRegistration, User owner, params string[] requestedActions)
+        {
+            if (packageRegistration == null)
+            {
+                return true;
+            }
+
+            return requestedActions
+                .Select(GetRequiredPackageAction)
+                .Any(p => PermissionsService.IsActionAllowed(packageRegistration, owner, p));
+        }
+
+        private static PermissionLevel GetRequiredPackageAction(string scope)
+        {
+            switch (scope.ToLowerInvariant())
+            {
+                case All:
+                    return PackageActions.ApiAll;
+                case PackagePush:
+                case PackagePushVersion:
+                    return PackageActions.ApiPush;
+                case PackageUnlist:
+                    return PackageActions.ApiUnlist;
+                case PackageVerify:
+                    return PackageActions.ApiVerify;
+                default:
+                    return PermissionLevel.Anonymous;
+            }
+        }
+
+        public static bool IsActionAllowedOnOwnerByCurrentUser(User owner, User currentUser, params string[] requestedActions)
+        {
+            return requestedActions
+                .Select(GetRequiredAccountAction)
+                .Any(p => PermissionsService.IsActionAllowed(owner, currentUser, p));
+        }
+
+        private static PermissionLevel GetRequiredAccountAction(string scope)
+        {
+            switch (scope.ToLowerInvariant())
+            {
+                case All:
+                    return AccountActions.ApiAllOnBehalfOf;
+                case PackagePush:
+                    return AccountActions.ApiPushOnBehalfOf;
+                case PackagePushVersion:
+                    return AccountActions.ApiPushVersionOnBehalfOf;
+                case PackageUnlist:
+                    return AccountActions.ApiUnlistOnBehalfOf;
+                case PackageVerify:
+                    return AccountActions.ApiVerifyOnBehalfOf;
+                default:
+                    return PermissionLevel.Anonymous;
+            }
         }
     }
 }
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -294,9 +294,16 @@ namespace NuGetGallery
                 new PackageAuditRecord(package, AuditedPackageAction.Verify));
 
             User owner;
-            var apiScopeEvaluationResult = CredentialTypes.IsPackageVerificationApiKey(credential.Type) ?
-                EvaluateApiScopeOnExisting(package.PackageRegistration, out owner, NuGetScopes.PackageVerify) :
-                EvaluateApiScopeOnExisting(package.PackageRegistration, out owner, NuGetScopes.PackagePush, NuGetScopes.PackagePushVersion);
+            ApiScopeEvaluationResult apiScopeEvaluationResult;
+            if (CredentialTypes.IsPackageVerificationApiKey(credential.Type))
+            {
+                apiScopeEvaluationResult = EvaluateApiScopeOnExisting(package.PackageRegistration, out owner, NuGetScopes.PackageVerify);
+            }
+            else
+            {
+                apiScopeEvaluationResult = EvaluateApiScopeOnExisting(package.PackageRegistration, out owner, NuGetScopes.PackagePush, NuGetScopes.PackagePushVersion);
+            }
+
             if (apiScopeEvaluationResult != ApiScopeEvaluationResult.Success)
             {
                 return GetHttpResultFromFailedApiScopeEvaluation(apiScopeEvaluationResult, id, version);
@@ -803,7 +810,7 @@ namespace NuGetGallery
             {
                 // Legacy V1 API key without scopes.
                 // Evaluate it as if it has an unlimited scope.
-                scopes = new[] { new Scope(null, NuGetPackagePattern.AllInclusivePattern, NuGetScopes.All) };
+                scopes = new[] { new Scope(ownerKey: null, subject: NuGetPackagePattern.AllInclusivePattern, allowedAction: NuGetScopes.All) };
             }
 
             var failedApiScopeEvaluationResults = new List<ApiScopeEvaluationResult>();
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -776,7 +776,7 @@ namespace NuGetGallery
         /// <returns>A <see cref="ApiScopeEvaluationResult"/> that describes the evaluation of the .</returns>
         private ApiScopeEvaluationResult EvaluateApiScopeOnExisting(PackageRegistration packageRegistration, out User owner, params string[] requestedActions)
         {
-            return EvaluateApiScope(new ScopeSubject(packageRegistration), out owner, requestedActions);
+            return EvaluateApiScope(new ExistingScopeSubject(packageRegistration), out owner, requestedActions);
         }
 
         /// <summary>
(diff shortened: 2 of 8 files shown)


src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -293,21 +293,13 @@ public ActionResult HealthProbe()
             await AuditingService.SaveAuditRecordAsync(
                 new PackageAuditRecord(package, AuditedPackageAction.Verify));
 
-            if (CredentialTypes.IsPackageVerificationApiKey(credential.Type))
+            User owner;
+            var apiScopeEvaluationResult = CredentialTypes.IsPackageVerificationApiKey(credential.Type) ?
```
joelverhagen: nit: too long to be ternary, this is hindering readability

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -706,42 +692,163 @@ protected internal virtual Stream ReadPackageFromRequest()
...
+            var currentUser = GetCurrentUser();
+            IEnumerable<Scope> scopes = User.Identity.GetScopesFromClaim();
+
+            if (scopes == null || !scopes.Any())
             {
-                // Return true if scope allows action for subject (package).
-                return scopes.Any(s => s.AllowsActions(NuGetScopes.PackagePush) && s.AllowsSubject(packageId));
+                // Legacy V1 API key without scopes.
+                // Evaluate it as if it has an unlimited scope.
+                scopes = new[] { new Scope(null, NuGetPackagePattern.AllInclusivePattern, NuGetScopes.All) };
```
joelverhagen: name the null param, this is null owner right? We should call out the meaning of null owner.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -706,42 +692,163 @@ protected internal virtual Stream ReadPackageFromRequest()
...
+                    continue;
+                }
+
+                owner = ownerInScope;
+                return ApiScopeEvaluationResult.Success;
             }
 
-            // Legacy V1 API key (no scopes).
-            return true;
+            return failedApiScopeEvaluationResults.Max();
```
joelverhagen: This is a little tricky. Order of enum values is now significant. Also, if you are just taking the max at the end, why not just track the max as you go?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -706,42 +692,163 @@ protected internal virtual Stream ReadPackageFromRequest()
...
+                case ApiScopeEvaluationResult.ConflictReservedNamespace:
+                    TelemetryService.TrackPackagePushNamespaceConflictEvent(id, version.ToNormalizedString(), GetCurrentUser(), User.Identity);
+                    return new HttpStatusCodeWithBodyResult(HttpStatusCode.Conflict, Strings.UploadPackage_IdNamespaceConflict);
+
+                default:
+                    throw new ArgumentException("Unsupported evaluation result!", nameof(evaluationResult));
+            }
+        }
+
+        private ApiScopeEvaluationResult EvaluateApiScopeOnExisting(Package package, out User owner, params string[] requestedActions)
```
joelverhagen: Would it be possible to extract all of this logic into another class? `ApiController` is already huge.

chenriksson:
@scottbommarito I'm a bit confused. Is this the refactoring you wanted to do to ApiController, or one of the issues that you were concerned about (#4994 or #4995), or all in one? If you're driving this, should I assign those issues to you?

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: Yep it's tricky, but it seemed like any easy way to do what I wanted without introducing a huge amount of complexity. Any idea on how to communicate that enum ordering now matters?

Also good catch on maintaining the max.

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: `ApiScopeEvaluationService`?

src/NuGetGallery/Authentication/NuGetScopes.cs
```diff
@@ -29,5 +31,61 @@ public static string Describe(string scope)
 
             return Strings.ScopeDescription_Unknown;
         }
+
+        public static bool IsActionAllowedOnSubjectByOwner(PackageRegistration packageRegistration, User owner, params string[] requestedActions)
```
chenriksson: Why not put this new logic in the PermissionsService class?

src/NuGetGallery/Authentication/NuGetScopes.cs
```diff
@@ -29,5 +31,61 @@ public static string Describe(string scope)
...
+        {
+            if (packageRegistration == null)
+            {
+                return true;
```
chenriksson: Why true, instead of throwing ArgNull?

src/NuGetGallery/Authentication/NuGetScopes.cs
```diff
@@ -29,5 +31,61 @@ public static string Describe(string scope)
...
+        }
+
+        private static PermissionLevel GetRequiredAccountAction(string scope)
+        {
+            switch (scope.ToLowerInvariant())
+            {
+                case All:
+                    return AccountActions.ApiAllOnBehalfOf;
+                case PackagePush:
+                    return AccountActions.ApiPushOnBehalfOf;
```
chenriksson: What is the difference between the AccountActions and PackageActions here? Why does one IsActionAllowed API use AccountActions and the other use PackageActions?

e.g., AccountActions.PushOnBehalfOf and PackageActions.ApiPush

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -706,42 +692,163 @@ protected internal virtual Stream ReadPackageFromRequest()
...
+
+        private HttpStatusCodeWithBodyResult GetHttpResultFromFailedApiScopeEvaluation(ApiScopeEvaluationResult evaluationResult, string id, NuGetVersion version)
+        {
+            switch (evaluationResult)
+            {
+                case ApiScopeEvaluationResult.Success:
+                    throw new ArgumentException($"{nameof(ApiScopeEvaluationResult.Success)} is not a failed evaluation!", nameof(evaluationResult));
+
+                case ApiScopeEvaluationResult.Forbidden:
+                    return new HttpStatusCodeWithBodyResult(HttpStatusCode.Forbidden, Strings.ApiKeyNotAuthorized);
```
chenriksson: This is a change of behavior. Some APIs return Forbidden for ApiKeyNotAuthorized, but push returns Unauthorized.

I think it would be best to keep this logic with the specific action methods.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -706,42 +692,163 @@ protected internal virtual Stream ReadPackageFromRequest()
...
+        {
+            return GetHttpResultFromFailedApiScopeEvaluation(evaluationResult, id, NuGetVersion.Parse(version));
+        }
+
+        private HttpStatusCodeWithBodyResult GetHttpResultFromFailedApiScopeEvaluation(ApiScopeEvaluationResult evaluationResult, string id, NuGetVersion version)
+        {
+            switch (evaluationResult)
+            {
+                case ApiScopeEvaluationResult.Success:
+                    throw new ArgumentException($"{nameof(ApiScopeEvaluationResult.Success)} is not a failed evaluation!", nameof(evaluationResult));
```
chenriksson: This code should never be reached, so why not just combine it with the default case?

src/NuGetGallery/Controllers/ApiController.cs
chenriksson: I thought we discussed extracting the scope evaluation into the PermissionsService?

And if you're folding in the namespace evaluation into permissions, I would consider turning PermissionsService into a service instance that uses NamespaceService directly... no need to have it on this class anymore.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -706,42 +692,163 @@ protected internal virtual Stream ReadPackageFromRequest()
...
+                {
+                    return ApiScopeEvaluationResult.Forbidden;
+                }
+
+                if (_packageRegistration == null && !_reservedNamespaceService.IsPushAllowed(_id, owner, out var userOwnedNamespaces))
                 {
-                    // ApiKeyHandler has already verified that the current user matches the owner scope.
-                    // Do not need to check organization role (IsAdmin) which is covered by the action scope.
-                    return GetCurrentUser().IsOwnerOrMemberOfOrganizationOwner(package);
+                    return ApiScopeEvaluationResult.ConflictReservedNamespace;
```
chenriksson: This method has a weak/confusing contract. The second half is only relevant when ScopeSubject is constructed using a PackageRegistration?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -706,42 +692,163 @@ protected internal virtual Stream ReadPackageFromRequest()
...
+
+            public ScopeSubject(string id, IReservedNamespaceService reservedNamespaceService)
+            {
+                if (string.IsNullOrEmpty(id))
+                {
+                    throw new ArgumentException("Package ID must be non-null!", nameof(id));
+                }
+
+                _id = id;
+                _reservedNamespaceService = reservedNamespaceService ?? throw new ArgumentNullException(nameof(reservedNamespaceService));
```
chenriksson: I don't understand the need for this class, and the contract is confusing. ReservedNamespaceService is required, but only when constructing with an id?

Does this mean there are 2 states for this class... new and existing registrations?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -706,42 +692,163 @@ protected internal virtual Stream ReadPackageFromRequest()
...
+            if (scopes == null || !scopes.Any())
             {
-                // Return true if scope allows action for subject (package).
-                return scopes.Any(s => s.AllowsActions(NuGetScopes.PackagePush) && s.AllowsSubject(packageId));
+                // Legacy V1 API key without scopes.
+                // Evaluate it as if it has an unlimited scope.
+                scopes = new[] { new Scope(null, NuGetPackagePattern.AllInclusivePattern, NuGetScopes.All) };
+            }
+
+            var failedApiScopeEvaluationResults = new List<ApiScopeEvaluationResult>();
```
chenriksson: Why do we collect these? We should expect some scopes to fail even in the success case, right? We only care that any scopes allow it...

src/NuGetGallery/Authentication/NuGetScopes.cs
scottbommarito: `PackageActions` = actions on packages
`AccountActions` = actions on accounts

One method is "does this user have permissions to do this on this package
The other method is "does this user have permissions to do this action on behalf of this user

src/NuGetGallery/Authentication/NuGetScopes.cs
scottbommarito: If there is no existing package registration, any owner can do any action on it.

src/NuGetGallery/Authentication/NuGetScopes.cs
scottbommarito: I agree that it belongs in there or in some other service, but I left it in here because it needs access to the `ReservedNamespaceService` and the `UserService`, which would mean that the `PermissionsService` or whatever other class it is moved to requires access to those services as well. I'm fine with moving it out, but it seems unnecessary to me as nothing else should use it.

scottbommarito: woops, misread this: I put this in here because it's specifically related to `NuGetScopes`. I suppose it's fine in the `PermissionsService`, but I thought it made sense here.

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: I noticed that, but I also noticed that we return `Forbidden` in one of the API authorization attributes, so all of our clients must handle this scenario anyway. I think push returning `Unauthorized` is more of a bug than anything.

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: Yep, new and existing registrations.

This class prevents us from needing duplicated code for new and existing registrations. It may potentially make more sense with better comments or interfacing, but I think it's the proper design.

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: If a scope succeeds, we exit the loop and return success. If no scopes succeed, we pick the greatest error state and return that.

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: I suppose it would make more sense if this was two separate classes that inherit from an interface.

`IScopeSubject`
`ExistingScopeSubject`
`NewScopeSubject`

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: Turning `PermissionsService` into a service that uses the `ReservedNamespaceService` could be tricky...ultimately, some of the code needs to be called statically. Adding this to `PermissionsService` would make it require access to the `ReservedNamespaceService` and the `UserService`. I suppose the existing methods could remain static and then new, non-static methods could be added, but I feel like that would make it more confusing than it's worth when this code is exclusively used inside `ApiController`.

I'm fine with moving it over, but it seems like more trouble than it's worth to me.

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: it's a separate error--this will provide us more debuggability in case this does happen

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: split up the class into 2 as discussed above

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: made the change suggested

scottbommarito:
Closing this because it will be redone with the permissions service 2.0

PR #5004, opened 2017-11-14
[MSA] Encourage MSA logins on NuGet.org

shishirx34:
This PR makes UI changes for encouraging users to login using MSA.
- Moves the login/register using NuGet.org to the second page. 
- Adds telemetry for register accounts.
- Removes register link from the header.

TODO:
- [x] Need to fix the functional tests that will be broken by this change.

Snapshots:

Main Logon Page:
image

Logon with NuGet page:
image


anangaur:
Can this be at the center of the white space between the header (blue part) and footer (other blue part)? I think that will look better.

image

anangaur:
But do not block on this one. We can take this change later.

src/NuGetGallery/Web.config
```diff
@@ -84,9 +84,9 @@
     <add key="Auth.LocalUser.Enabled" value="true" />
     <add key="Auth.ApiKey.Enabled" value="true" />
     
-    <add key="Auth.MicrosoftAccount.Enabled" value="false" />
-    <add key="Auth.MicrosoftAccount.ClientId" value="" />
-    <add key="Auth.MicrosoftAccount.ClientSecret" value="" />
+    <add key="Auth.MicrosoftAccount.Enabled" value="true" />
+    <add key="Auth.MicrosoftAccount.ClientId" value="00000000481B784E" />
+    <add key="Auth.MicrosoftAccount.ClientSecret" value="32XRPDH0x9jrz2mDj1rRMJs" />
```
joelverhagen: You may want to reset these guys. 👀

joelverhagen:
Could we have a bit of whitespace between the Sign in button and the "No account? Create one!" line? They look crowded.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -69,15 +69,30 @@ public virtual ActionResult LogOn(string returnUrl)
...
-                return SafeRedirect(returnUrl);
+                return LoggedInRedirect(returnUrl);
             }
 
             return SignInView(new LogOnViewModel());
         }
 
         /// <summary>
         /// Sign In\Register view
+        /// <summary>
```
skofman1: the comment is duplicated

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -153,6 +157,28 @@ public void TrackVerifyPackageKeyEvent(string packageId, string packageVersion,
             });
         }
 
+        public void TrackNewUserRegistrationEvent(User user, Credential credential)
+        {
+
+            if (user == null)
```
skofman1: nit: extra line

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -153,6 +157,28 @@ public void TrackVerifyPackageKeyEvent(string packageId, string packageVersion,
...
+
+            if (credential == null)
+            {
+                throw new ArgumentNullException(nameof(credential));
+            }
+
+            TrackEvent(Events.NewUserRegistration, properties => {
+                properties.Add(ClientVersion, GetClientVersion());
+                properties.Add(ProtocolVersion, GetProtocolVersion());
+                properties.Add(ClientInformation, GetClientInformation());
```
skofman1: we don't need client information for this event. It will always be a browser.

skofman1:
Looks good! A few comments.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -325,6 +287,7 @@ private void InitCredentialFormatters()
             Entities.Users.Add(newUser);
             await Entities.SaveChangesAsync();
 
+            _telemetryService.TrackNewUserRegistrationEvent(newUser, credential);
```
cristinamanum: Will this log personal data to AI? If yes, would it be possible to move it Geneva audit pipeline?

shishirx34: no it won't, it simply logs the account creation time and the type of account(password/external.microsoft)

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: ah auto merge

PR #5007, opened 2017-11-15
Add initial support for user deletion request

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/921.

The changes were getting big so I stopped and made a PR. Hopefully easier to review.

This PR:

1. Update the "report my package" form to support a new hidden drop-down (delete decision) and confirmation checkbox.
1. Add new notification message so that all owners know a version was deleted.
1. Add method to `IPackageDeleteService` to determine if a package can be deleted. It always returns false right now.
1. Add logic to soft delete the package and auto-close the support request.

TODO:

1. Implement "can I delete this?" logic.
1. Introduce configuration for the aforementioned logic.
1. Functional test (hopefully)

Without these TODOs, the new code paths introduced by this PR will not be available yet.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -19,6 +19,7 @@ using System.Web.Mvc;
 using NuGet.Packaging;
 using NuGet.Versioning;
 using NuGetGallery.Areas.Admin;
+using NuGetGallery.Areas.Admin.Models;
 using NuGetGallery.AsyncFileUpload;
 using NuGetGallery.Auditing;
 using NuGetGallery.Configuration;
@@ -35,6 +36,29 @@ namespace NuGetGallery
     public partial class PackagesController
         : AppController
     {
+        private static readonly IReadOnlyList<ReportPackageReason> ReportAbuseReasons = new[]
+        {
+            ReportPackageReason.ViolatesALicenseIOwn,
+            ReportPackageReason.ContainsMaliciousCode,
+            ReportPackageReason.HasABugOrFailedToInstall,
+            ReportPackageReason.Other
+        };
+
+        private static readonly IReadOnlyList<ReportPackageReason> ReportMyPackageReasons = new[]
+        {
+            ReportPackageReason.ContainsPrivateAndConfidentialData,
+            ReportPackageReason.ReleasedInPublicByAccident,
+            ReportPackageReason.ContainsMaliciousCode,
+            ReportPackageReason.Other
+        };
+
+        private static readonly IReadOnlyList<ReportPackageReason> DeleteReasons = new[]
+        {
+            ReportPackageReason.ContainsPrivateAndConfidentialData,
+            ReportPackageReason.ReleasedInPublicByAccident,
+            ReportPackageReason.ContainsMaliciousCode,
+        };
+
         // TODO: add support for URL-based package submission
         // TODO: add support for uploading logos and screenshots
         // TODO: improve validation summary emphasis
@@ -564,14 +588,6 @@ namespace NuGetGallery
             return View(viewModel);
         }
 
-        // NOTE: Intentionally NOT requiring authentication
-        private static readonly ReportPackageReason[] ReportOtherPackageReasons = new[] {
-            ReportPackageReason.ViolatesALicenseIOwn,
-            ReportPackageReason.ContainsMaliciousCode,
-            ReportPackageReason.HasABugOrFailedToInstall,
-            ReportPackageReason.Other
-        };
-
         [HttpGet]
         public virtual ActionResult ReportAbuse(string id, string version)
         {
@@ -584,7 +600,7 @@ namespace NuGetGallery
 
             var model = new ReportAbuseViewModel
             {
-                ReasonChoices = ReportOtherPackageReasons,
+                ReasonChoices = ReportAbuseReasons,
                 PackageId = id,
                 PackageVersion = package.Version,
                 CopySender = true,
@@ -610,17 +626,10 @@ namespace NuGetGallery
             return View(model);
         }
 
-        private static readonly ReportPackageReason[] ReportMyPackageReasons = {
-            ReportPackageReason.ContainsPrivateAndConfidentialData,
-            ReportPackageReason.ReleasedInPublicByAccident,
-            ReportPackageReason.ContainsMaliciousCode,
-            ReportPackageReason.Other
-        };
-
         [HttpGet]
         [Authorize]
         [RequiresAccountConfirmation("contact support about your package")]
-        public virtual ActionResult ReportMyPackage(string id, string version)
+        public virtual async Task<ActionResult> ReportMyPackage(string id, string version)
         {
             var package = _packageService.FindPackageByIdAndVersionStrict(id, version);
 
@@ -628,20 +637,23 @@ namespace NuGetGallery
             {
                 return HttpNotFound();
             }
-
-            // If user hit this url by constructing it manually but is not the owner, redirect them to ReportAbuse
+            
             if (!PermissionsService.IsActionAllowed(package, User, PackageActions.ReportMyPackage))
             {
-                return RedirectToAction("ReportAbuse", new { id, version });
+                return RedirectToAction(nameof(ReportAbuse), new { id, version });
             }
+
+            var allowDelete = await _packageDeleteService.CanPackageBeDeletedByUserAsync(package);
 
             var model = new ReportMyPackageViewModel
             {
                 ReasonChoices = ReportMyPackageReasons,
+                DeleteReasonChoices = DeleteReasons,
                 ConfirmedUser = GetCurrentUser().Confirmed,
                 PackageId = id,
                 PackageVersion = package.Version,
-                CopySender = true
+                CopySender = true,
+                AllowDelete = allowDelete,
             };
 
             return View(model);
@@ -652,16 +664,17 @@ namespace NuGetGallery
         [ValidateSpamPrevention]
         public virtual async Task<ActionResult> ReportAbuse(string id, string version, ReportAbuseViewModel reportForm)
         {
-            // Html Encode the message
-            reportForm.Message = System.Web.HttpUtility.HtmlEncode(reportForm.Message);
-
-            var modelIsValid = ModelState.IsValid;
-            if (reportForm.Reason == ReportPackageReason.ViolatesALicenseIOwn)
+            reportForm.Message = HttpUtility.HtmlEncode(reportForm.Message);
+            
+            if (reportForm.Reason == ReportPackageReason.ViolatesALicenseIOwn
+                && string.IsNullOrWhiteSpace(reportForm.Signature))
             {
-                modelIsValid = modelIsValid && !string.IsNullOrEmpty(reportForm.Signature);
+                ModelState.AddModelError(
+                    nameof(ReportAbuseViewModel.Signature),
+                    "The signature is required.");
             }
 
-            if (!modelIsValid)
+            if (!ModelState.IsValid)
             {
                 return ReportAbuse(id, version);
             }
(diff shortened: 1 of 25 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,22 +729,124 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
+            {
+                return RedirectToAction(nameof(ReportAbuse), new { id, version });
+            }
+
+            // Enforce the auto-delete rules.
+            var allowDelete = await _packageDeleteService.CanPackageBeDeletedByUserAsync(package);
+            if (!allowDelete
+                && reportForm.DeleteDecision == PackageDeleteDecision.DeletePackage)
+            {
+                reportForm.DeleteDecision = null;
```
chenriksson: Does this mean that lack of permissions is going to result in the same UI validation error for when a delete decision wasn't provided?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,22 +729,124 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
+                        user,
+                        reason,
+                        signature: "(automated)");
+                    deleted = true;
+                }
+                catch (Exception e)
+                {
+                    // Swallow exceptions that occur during the delete process. If this happens, we'll just
+                    // send out the support request as usual and handle it manually.
+                    QuietLog.LogHandledException(e);
```
chenriksson: Will the user receive a message that the support request was sent?

src/NuGetGallery/ExtensionMethods.cs
```diff
@@ -291,6 +291,19 @@ public static HtmlString ShowCheckboxFor<TModel>(this HtmlHelper<TModel> html, E
         public static MvcHtmlString ShowEnumDropDownListFor<TModel, TEnum>(
             this HtmlHelper<TModel> html,
             Expression<Func<TModel, TEnum?>> expression,
+            string emptyItemText)
```
chenriksson: There's also a HtmlExtensions class. Maybe we should move these HtmlHelper extensions there?

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: In this case, the UI will behave as if the delete decision UI was hidden. Really, the user should never get into this case because the UI doesn't allow it. The missing delete decision validation error only occurs if `allowDelete` is true (which is not the case here).

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: `deleted` will still be `false` so we will show the "support request sent" message as usual.

src/NuGetGallery/ExtensionMethods.cs
joelverhagen: Hm, good point. I'll take a look at that.

joelverhagen: It looks like this is a big mess already. I will create an issue to clean this up.
https://github.com/NuGet/NuGetGallery/issues/5015

skofman1:
can you attach screenshots?

joelverhagen:
## Select a reason

image

## After choosing a delete reason

image

## After choosing the delete decision

image

## After choosing the contact support decision

image

## After deleted 

image

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,22 +729,124 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
+                return HttpNotFound();
+            }
+
+            if (!PermissionsService.IsActionAllowed(package, User, PackageActions.ReportMyPackage))
+            {
+                return RedirectToAction(nameof(ReportAbuse), new { id, version });
+            }
+
+            // Enforce the auto-delete rules.
+            var allowDelete = await _packageDeleteService.CanPackageBeDeletedByUserAsync(package);
```
skofman1: do we need to evaluate CanPackageBeDeletedByUserAsync is DeleteDecision is not DeletePackage? If the user just wants to contact support we are doing unnecessary work here.

joelverhagen: Great point, no we don't.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,22 +729,124 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
+                && reportForm.Reason.HasValue
+                && DeleteReasons.Contains(reportForm.Reason.Value)
+                && !reportForm.DeleteDecision.HasValue)
+            {
+                ModelState.AddModelError(
+                    nameof(ReportMyPackageViewModel.DeleteDecision),
+                    "You must choose whether to delete the package or contact support.");
+            }
+
+            // Require the confirmation checkbox if we are performing an auto-delete.
```
skofman1: the comment is confusing. You are requiring the checkbox in all cases, not only auto delete (which is the correct behavior)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,22 +729,124 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
             }
 
-            var package = _packageService.FindPackageByIdAndVersionStrict(id, version);
-            if (package == null)
+            // Override the copy sender and message fields if we are performing an auto-delete.
+            if (reportForm.DeleteDecision == PackageDeleteDecision.DeletePackage)
             {
-                return HttpNotFound();
+                reportForm.CopySender = false;
+                reportForm.Message = "This support request is intended for automatic deletion.";
```
skofman1: automatic package deletion

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,22 +729,124 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
+                        package,
+                        Url.Package(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
+                        Url.ReportPackage(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false));
+
+                    TempData["Message"] =
+                        "We're performing the package delete right now. It may take a while for this " +
+                        "change to propagate through our system.";
+
+                    return Redirect(Url.Package(id, version));
+                }
```
skofman1: this method is to big. Can you split it to smaller methods?

joelverhagen: Sure!

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,22 +729,124 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
+
+            var supportRequest = await _supportRequestService.AddNewSupportRequestAsync(
+                subject,
+                reportForm.Message,
+                from.Address,
+                reason,
+                user,
+                package);
+
+            if (supportRequest != null
```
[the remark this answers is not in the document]
joelverhagen: If the commit to the support request DB fails. This is existing logic in the support request service.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,22 +729,124 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
+                        editedBy: user.Username);
+
+                    _messageService.SendPackageDeletedNotice(
+                        package,
+                        Url.Package(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
+                        Url.ReportPackage(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false));
+
+                    TempData["Message"] =
+                        "We're performing the package delete right now. It may take a while for this " +
+                        "change to propagate through our system.";
```
skofman1: I know this controller doesn't do much use of Strings, but perhaps it time to start?

joelverhagen: Sure!

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,22 +729,124 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
-            reportForm.Message = System.Web.HttpUtility.HtmlEncode(reportForm.Message);
+            reportForm.Message = HttpUtility.HtmlEncode(reportForm.Message);
+
+            var package = _packageService.FindPackageByIdAndVersionStrict(id, version);
+            if (package == null)
+            {
+                return HttpNotFound();
+            }
+
+            if (!PermissionsService.IsActionAllowed(package, User, PackageActions.ReportMyPackage))
```
skofman1: Are the permissions required for reporting a package are the same as for deleting a package? Do we need a special check for the later?

joelverhagen: Same permissions. The keyword here is "My" in "ReportMyPackage".

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: Fixed.

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: Fixed. Good point.

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: Fixed.

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: Fixed.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1935,69 +1936,464 @@ public void FormRedirectsPackageOwnerToReportMyPackage()
...
+                    Strings.UserPackageDeleteConfirmationIsRequired,
+                    _controller
+                        .ModelState
+                        .Values
+                        .SelectMany(x => x.Errors)
+                        .Select(x => x.ErrorMessage));
+            }
+
+            [Theory]
+            [InlineData(ReportPackageReason.ContainsMaliciousCode)]
```
scottbommarito: two thoughts:

1 - we should make this run on all members of `PackagesController.DeleteReasons`
2 - if we are concerned about the members of `PackageController.DeleteReasons` changing, then there should be tests for that

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -652,16 +664,17 @@ public virtual ActionResult ReportMyPackage(string id, string version)
         [ValidateSpamPrevention]
         public virtual async Task<ActionResult> ReportAbuse(string id, string version, ReportAbuseViewModel reportForm)
         {
-            // Html Encode the message
-            reportForm.Message = System.Web.HttpUtility.HtmlEncode(reportForm.Message);
-
-            var modelIsValid = ModelState.IsValid;
-            if (reportForm.Reason == ReportPackageReason.ViolatesALicenseIOwn)
+            reportForm.Message = HttpUtility.HtmlEncode(reportForm.Message);
```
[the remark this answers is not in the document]
joelverhagen: Out of scope of this PR. I think we are encoding it since it makes its way into the email system.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -652,16 +664,17 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
+            
+            if (reportForm.Reason == ReportPackageReason.ViolatesALicenseIOwn
+                && string.IsNullOrWhiteSpace(reportForm.Signature))
             {
-                modelIsValid = modelIsValid && !string.IsNullOrEmpty(reportForm.Signature);
+                ModelState.AddModelError(
+                    nameof(ReportAbuseViewModel.Signature),
+                    "The signature is required.");
```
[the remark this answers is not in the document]
joelverhagen: Yes, that should be in Strings.

The context of the UI negates the need for verbosity. If the message field is visible, it is required.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,23 +729,115 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
 
-            if (!ModelState.IsValid)
+            var user = GetCurrentUser();
+            var from = user.ToMailAddress();
+            var subject = string.Format(
+                Strings.OwnerSupportRequestSubjectFormat,
+                package.PackageRegistration.Id,
+                package.NormalizedVersion);
+            var reason = EnumHelper.GetDescription(reportForm.Reason.Value);
+            var supportRequest = await _supportRequestService.AddNewSupportRequestAsync(
```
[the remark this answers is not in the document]
joelverhagen: This is to track the auto-delete in a way the PMs asked for.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,23 +729,115 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
             }
 
-            var user = GetCurrentUser();
-            MailAddress from = user.ToMailAddress();
+            if (!PermissionsService.IsActionAllowed(package, User, PackageActions.ReportMyPackage))
+            {
+                return RedirectToAction(nameof(ReportAbuse), new { id = package.PackageRegistration.Id, version = package.NormalizedVersion });
+            }
+
+            reportForm.Message = HttpUtility.HtmlEncode(reportForm.Message);
```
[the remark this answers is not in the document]
joelverhagen: See above.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -35,6 +36,29 @@ namespace NuGetGallery
     public partial class PackagesController
         : AppController
     {
+        private static readonly IReadOnlyList<ReportPackageReason> ReportAbuseReasons = new[]
```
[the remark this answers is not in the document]
joelverhagen: Yeah, good idea. Thanks for doing the work for me 😉

src/NuGetGallery/Views/Packages/ReportMyPackage.cshtml
```diff
@@ -26,21 +26,45 @@
...
+                    <p>Please select how you would like to proceed.</p>
+                    @Html.ShowEnumDropDownListFor(m => m.DeleteDecision, "<Choose how to proceed>")
+                    @Html.ShowValidationMessagesFor(m => m.DeleteDecision)
+                </div>
+                <div id="form-field-confirm-delete" class="hidden checkbox">
+                    <label>
+                        @Html.ShowCheckboxFor(m => m.DeleteConfirmation)
+                        I understand that this action cannot be undone. This will permanently delete the package
+                        @Model.PackageId @Model.PackageVersion and make it unavailable for download and package restore.
+                        This does not allow uploading a new package with the same version.
```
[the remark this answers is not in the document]
joelverhagen: Yeah, I like that better too.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1935,69 +1936,464 @@ public void FormRedirectsPackageOwnerToReportMyPackage()
...
+                    .ReturnsAsync(true);
+
+                // Act
+                var result = await _controller.ReportMyPackage(
+                    _package.PackageRegistration.Id,
+                    _package.Version,
+                    _viewModel);
+
+                // Assert
+                Assert.Contains(
```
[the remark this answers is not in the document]
joelverhagen: Sure.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -716,23 +729,115 @@ public virtual ActionResult ReportMyPackage(string id, string version)
...
+            reportForm.Message = HttpUtility.HtmlEncode(reportForm.Message);
+
+            // Enforce the auto-delete rules.
+            var allowDelete = false;
+            if (reportForm.DeleteDecision != PackageDeleteDecision.ContactSupport)
+            {
+                allowDelete = await _packageDeleteService.CanPackageBeDeletedByUserAsync(package);
+                if (!allowDelete)
+                {
+                    reportForm.DeleteDecision = null;
```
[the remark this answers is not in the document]
joelverhagen: > 1 - return an error message that says they package does not meet the criteria

Why show an error message? The only way this can happen is if someone is fiddling with the request or if they get the form while the package can be deleted and only submit it after the package can't be deleted. We shouldn't put extra effort into these cases.

> 2 - keep the deletion state so that the on-call can see it in the support request

The deletion state shouldn't even be persisted in this case. It's a bad request which should be rejected early on.

> Regardless, there should be a comment on this line describing what we are doing.

Sounds good.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
joelverhagen: Kind of out of scope, as there was no pre-existing validation on which reasons were being sent, but I'll add tests for it.

joelverhagen:
@scottbommarito, I will address these comments in a follow-up PR.

PR #5009, opened 2017-11-15
Setting for blocking async validation in Gallery

agr:
Causes some functional tests to fail due to breaking the assumption that package become immediately available in V2 once uploaded, tracked by https://github.com/NuGet/Engineering/issues/946


src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
```diff
@@ -54,8 +54,10 @@ public class AsynchronousPackageValidationInitiator : IPackageValidationInitiato
                 await _enqueuer.StartValidationAsync(data);
             }
 
-            // For now, don't require asynchronous validation before the package is available for consumption.
-            // Related: https://github.com/NuGet/NuGetGallery/issues/4744
+            if (_appConfiguration.BlockingAsynchronousPackageValidationEnabled)
+            {
+                return PackageStatus.Validating;
+            }
             return PackageStatus.Available;
```
joelverhagen: nit: blank line before the return

joelverhagen:
This should be against `dev`.

Issue #5020, opened 2017-11-16
New login UI does not show invalid password errors


joelverhagen:
1. Go here: https://dev.nugettest.org/users/account/LogOnNuGetAccount
2. User a bad username and password

No validation error.

PR #5023, opened 2017-11-17
Fix new Login UI validation errors missing bug and add telemetry

shishirx34:
This PR fixes https://github.com/NuGet/NuGetGallery/issues/5020 where new Login UI wouldn't show validation errors for bad username and password. Also added telemetry for credential added(for getting data on MSA account linking)

### DIFF
--- a/tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
+++ b/tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
@@ -143,7 +143,7 @@ namespace NuGetGallery.Controllers
 
                 ResultAssert.IsView(result, viewName: SignInViewNuGetName);
                 Assert.False(controller.ModelState.IsValid);
-                Assert.Equal(Strings.UsernameAndPasswordNotFound, controller.ModelState[SignInViewNuGetName].Errors[0].ErrorMessage);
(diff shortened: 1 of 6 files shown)


tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
```diff
@@ -143,7 +143,7 @@ public class TheSignInAction : TestContainer
 
                 ResultAssert.IsView(result, viewName: SignInViewNuGetName);
                 Assert.False(controller.ModelState.IsValid);
-                Assert.Equal(Strings.UsernameAndPasswordNotFound, controller.ModelState[SignInViewNuGetName].Errors[0].ErrorMessage);
+                Assert.Equal(Strings.UsernameAndPasswordNotFound, controller.ModelState[SignInViewName].Errors[0].ErrorMessage);
```
scottbommarito: I see you've replaced `SignInViewNuGetName` with `SignInViewName`, but I don't see a change that adds `SignInViewName`. Is `SignInViewName` an existing field?

shishirx34: Yes, it is an existing field. In my previous change I accidentally changed this to the `SignInViewNuGetName`

PR #5029, opened 2017-11-17
Show VerifyPackage errors correctly in UI

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/4787

A couple fixes here:
1 - if starting an edit request fails (due to title/ID collision), return the error message in JSON
2 - keep the form open when verifying fails
3 - clear warnings when showing errors for an upload
4 - clear errors when canceling an upload

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1589,7 +1589,16 @@ namespace NuGetGallery
 
                 if (pendEdit)
                 {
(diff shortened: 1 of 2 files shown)


chenriksson:
@scottbommarito Can you check the code path for "Id conflicts with Title of existing" to make sure it is fixed as well?

Look at references to `PackageNamingConflictValidator.IdConflictsWithExistingPackageTitle` to trace it back to where it gets invoked by PackagesController.VerifyPackages.

scottbommarito:
@chenriksson investigated and fixed!

PR #5031, opened 2017-11-17
ApiKeyV4 implementation

skofman1:
This PR introduces ApiKeyV4 implementation without integrating it into the exiting scenarios. This will be done in a future PR. 
/cc @blowdart

### DIFF
--- a/src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
+++ b/src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
@@ -1,0 +1,139 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Cryptography;
+
+namespace NuGetGallery.Infrastructure.Authentication
+{
+    public class ApiKeyV4
+    {
+        private const int IdPartLengthBytes = 10;
+        private static readonly byte[] IdPrefix = new byte[] { 118, 52 };
+
+        internal const int IdPartBase32Length = 20;
+        internal const int IdAndPasswordLength = 46;
+        internal const int IdAndPasswordEncryptedLength = 124;
+
+        /// <summary>
+        /// Plaintext format of the ApiKey
+        /// </summary>
+        public string PlaintextApiKey { get; private set; }
+
+        /// <summary>
+        /// Encrypted format of the ApiKey. Will be set only if Create() method was used.
+        /// </summary>
+        public string EncryptedApiKey { get; private set; }
+
+        /// <summary>
+        /// Id part of the ApiKey
+        /// </summary>
+        public string IdPart { get; private set; }
+
+        /// <summary>
+        /// Password part of the ApiKey (plaintext)
+        /// </summary>
+        public string PasswordPart { get; private set; }
+
+        private ApiKeyV4()
+        {
+        }
+
+        public static ApiKeyV4 Create()
+        {
+            var apiKey = new ApiKeyV4();
+            apiKey.CreateInternal();
+
+            return apiKey;
+        }
+
+        public static bool TryParse(string plaintextApiKey, out ApiKeyV4 apiKey)
+        {
+            apiKey = new ApiKeyV4();
+            return apiKey.TryParseInternal(plaintextApiKey);
+        }
+
+        public bool Verify(string encryptedApiKey)
+        {
+            if (string.IsNullOrEmpty(encryptedApiKey) || encryptedApiKey.Length != IdAndPasswordEncryptedLength)
+            {
+                return false;
+            }
+
+            string encyptedApiKeyIdPart = encryptedApiKey.Substring(0, IdPartBase32Length);
+            string encryptedApiKeyPasswordPart = encryptedApiKey.Substring(IdPartBase32Length, encryptedApiKey.Length - IdPartBase32Length);
+
+            if (string.Compare(IdPart, Normalize(encyptedApiKeyIdPart)) != 0)
+            {
+                return false;
+            }
+
+            // The verification is not case sensitive. This is to maintain the existing behavior that ApiKey authentication is not case-sensitive.
+            return V3Hasher.VerifyHash(encryptedApiKeyPasswordPart.ToUpper().FromBase32String(), PasswordPart);
+        }
+
+        private void CreateInternal()
+        {
+            // Create Id
+            var randomBytes = new byte[IdPartLengthBytes];
+            using (var rng = new RNGCryptoServiceProvider())
+            {
+                rng.GetNonZeroBytes(randomBytes);
+            }
+
+            byte[] idBytes = new byte[IdPartLengthBytes + IdPrefix.Length];
+            Buffer.BlockCopy(IdPrefix, 0, idBytes, 0, IdPrefix.Length);
+            Buffer.BlockCopy(randomBytes, 0, idBytes, IdPrefix.Length, randomBytes.Length);
+
+            // Convert to Base32 string. The length of the string is APIKeyV4_IdPartBase64Length
+            string idString = idBytes.ToBase32String().RemoveBase32Padding();
+
+            // Create password
+            var passwordString = Guid.NewGuid().ToByteArray().ToBase32String().RemoveBase32Padding();
+            passwordString = Normalize(passwordString);
+
+            // No need to remove padding or normalize here.. it's stored in the DB and doesn't need to be pretty
+            var encryptedPasswordString = V3Hasher.GenerateHashAsBytes(passwordString).ToBase32String();
+
+            IdPart = Normalize(idString);
+            PasswordPart = passwordString;
+            PlaintextApiKey = idString + passwordString;
+            EncryptedApiKey = idString + encryptedPasswordString;
+        }
+
+        private bool TryParseInternal(string plaintextApiKey)
+        {
+            if (string.IsNullOrEmpty(plaintextApiKey) || plaintextApiKey.Length != IdAndPasswordLength)
+            {
+                return false;
+            }
+
+            try
+            {
+                var id = plaintextApiKey.Substring(0, IdPartBase32Length);
+                var idBytes = id.AppendBase32Padding().ToUpper().FromBase32String();
+                bool success = idBytes[0] == IdPrefix[0] && idBytes[1] == IdPrefix[1];
+
+                if (success)
+                {
+                    string password = plaintextApiKey.Substring(IdPartBase32Length, plaintextApiKey.Length - IdPartBase32Length);
+
+                    PlaintextApiKey = plaintextApiKey;
+                    IdPart = Normalize(id);
+                    PasswordPart = Normalize(password);
+                }
+
+                return success;
+            }
+            catch (InvalidOperationException)
+            {
+                return false;
+            }
+        }
+
+        private string Normalize(string input)
+        {
+            return input.ToLower();
+        }
+    }
+}
--- a/src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
+++ b/src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
@@ -3,13 +3,14 @@
 
 using System;
 using System.Security.Cryptography;
+using System.Text;
 
 namespace NuGetGallery.Infrastructure.Authentication
 {
     public class ApiKeyV4
     {
         private const int IdPartLengthBytes = 10;
-        private static readonly byte[] IdPrefix = new byte[] { 118, 52 };
+        private static readonly byte[] IdPrefix = Encoding.ASCII.GetBytes("v4");
 
         internal const int IdPartBase32Length = 20;
         internal const int IdAndPasswordLength = 46;
@@ -55,15 +56,15 @@ namespace NuGetGallery.Infrastructure.Authentication
 
         public bool Verify(string encryptedApiKey)
         {
-            if (string.IsNullOrEmpty(encryptedApiKey) || encryptedApiKey.Length != IdAndPasswordEncryptedLength)
+            if (string.IsNullOrWhiteSpace(encryptedApiKey) || encryptedApiKey.Length != IdAndPasswordEncryptedLength)
             {
                 return false;
             }
 
-            string encyptedApiKeyIdPart = encryptedApiKey.Substring(0, IdPartBase32Length);
-            string encryptedApiKeyPasswordPart = encryptedApiKey.Substring(IdPartBase32Length, encryptedApiKey.Length - IdPartBase32Length);
+            string encryptedApiKeyIdPart = encryptedApiKey.Substring(0, IdPartBase32Length);
+            string encryptedApiKeyPasswordPart = encryptedApiKey.Substring(IdPartBase32Length);
 
-            if (string.Compare(IdPart, Normalize(encyptedApiKeyIdPart)) != 0)
+            if (!string.Equals(IdPart, Normalize(encryptedApiKeyIdPart)))
             {
                 return false;
             }
@@ -82,8 +83,8 @@ namespace NuGetGallery.Infrastructure.Authentication
             }
 
             byte[] idBytes = new byte[IdPartLengthBytes + IdPrefix.Length];
-            Buffer.BlockCopy(IdPrefix, 0, idBytes, 0, IdPrefix.Length);
-            Buffer.BlockCopy(randomBytes, 0, idBytes, IdPrefix.Length, randomBytes.Length);
+            Buffer.BlockCopy(src: IdPrefix, srcOffset: 0, dst: idBytes, dstOffset: 0, count: IdPrefix.Length);
+            Buffer.BlockCopy(src: randomBytes, srcOffset: 0, dst: idBytes, dstOffset: IdPrefix.Length, count: randomBytes.Length);
 
             // Convert to Base32 string. The length of the string is APIKeyV4_IdPartBase64Length
             string idString = idBytes.ToBase32String().RemoveBase32Padding();
@@ -116,7 +117,7 @@ namespace NuGetGallery.Infrastructure.Authentication
 
                 if (success)
                 {
-                    string password = plaintextApiKey.Substring(IdPartBase32Length, plaintextApiKey.Length - IdPartBase32Length);
+                    string password = plaintextApiKey.Substring(IdPartBase32Length);
 
                     PlaintextApiKey = plaintextApiKey;
                     IdPart = Normalize(id);
--- a/src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
+++ b/src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
@@ -14,7 +14,7 @@ namespace NuGetGallery.Infrastructure.Authentication
 
         internal const int IdPartBase32Length = 20;
         internal const int IdAndPasswordLength = 46;
-        internal const int IdAndPasswordEncryptedLength = 124;
+        internal const int IdAndPasswordHashedLength = 124;
 
         /// <summary>
         /// Plaintext format of the ApiKey
@@ -22,9 +22,9 @@ namespace NuGetGallery.Infrastructure.Authentication
         public string PlaintextApiKey { get; private set; }
 
         /// <summary>
-        /// Encrypted format of the ApiKey. Will be set only if Create() method was used.
+        /// Hashed format of the ApiKey. Will be set only if Create() method was used.
         /// </summary>
-        public string EncryptedApiKey { get; private set; }
+        public string HashedApiKey { get; private set; }
 
         /// <summary>
         /// Id part of the ApiKey
@@ -40,6 +40,9 @@ namespace NuGetGallery.Infrastructure.Authentication
         {
         }
 
+        /// <summary>
+        /// Creates a random ApiKey V4.
+        /// </summary>
         public static ApiKeyV4 Create()
         {
             var apiKey = new ApiKeyV4();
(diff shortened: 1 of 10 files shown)


blowdart:
I have 5 books awaiting my attention already!

--
Sent from a tiny device with fat fingers. Please excuse any spelling mistakes.

On Nov 17, 2017, at 15:45, Svetlana Kofman <notifications@github.com<mailto:notifications@github.com>> wrote:


@skofman1<https://github.com/skofman1> requested your review on: NuGet/NuGetGallery#5031<https://github.com/NuGet/NuGetGallery/pull/5031> ApiKeyV4 implementation.

maartenba:
What is a v4 API key?

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
```diff
@@ -0,0 +1,195 @@
...
+        {
+            if (base32String == null)
+            {
+                throw new NullReferenceException(nameof(base32String));
+            }
+
+            // Validate base32 format
+            if ((base32String.Length % 8 != 0) || (base32String.Where(t => !_encodingChars.Contains(t)).Count() != 0))
+            {
+                throw new InvalidOperationException($"{nameof(base32String)} is not a valid base32 encoding");
```
loic-sharma: This should be an `ArgumentException` for `base32String`.

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
```diff
@@ -0,0 +1,195 @@
...
+        /// </summary>
+        public static byte[] Decode(string base32String)
+        {
+            if (base32String == null)
+            {
+                throw new NullReferenceException(nameof(base32String));
+            }
+
+            // Validate base32 format
+            if ((base32String.Length % 8 != 0) || (base32String.Where(t => !_encodingChars.Contains(t)).Count() != 0))
```
loic-sharma: This is going to `base32String.Length` * `_encodingChars.Length` checks. Could we use a regex instead?

chenriksson: Would regex be the most performant... or maybe something like the following?
https://stackoverflow.com/questions/641361/base32-decoding

Maybe it's worth testing each standalone in a large while loop to see which performs better?

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
```diff
@@ -0,0 +1,195 @@
...
+                        int shift2 = 8 - shift1;
+                        byte b = data[byteOffset + 1];
+                        b >>= shift2;
+                        retval |= b;
+                    }
+                } // (shift1 == 0) {  do nothing }
+
+                // Mask to right 5 bits
+                retval &= Bitmask;
+            } /* else {
```
chenriksson: remove dead code?

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
joelverhagen: This can be O(n) with a hash set of valid characters or `if` of valid character ranges. Also, the implementation can short-circuit on first invalid character. `Count()` checks the whole string which is not always necessary in failure cases.

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
```diff
@@ -0,0 +1,139 @@
...
+
+using System;
+using System.Security.Cryptography;
+
+namespace NuGetGallery.Infrastructure.Authentication
+{
+    public class ApiKeyV4
+    {
+        private const int IdPartLengthBytes = 10;
+        private static readonly byte[] IdPrefix = new byte[] { 118, 52 };
```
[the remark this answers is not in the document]
joelverhagen: can we generate this instead of by constants? e.g. `ASCIIEncoding.GetBytes`

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
```diff
@@ -0,0 +1,195 @@
...
+
+using System;
+using System.Linq;
+
+namespace NuGetGallery.Infrastructure.Authentication
+{
+    public static class Base32Encoder
+    {
+        private const byte Bitmask = 0x1F;
+        private static readonly char[] _encodingChars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567=".ToCharArray();
```
joelverhagen: nit: can haz lowercase? Why yell? 😉

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
```diff
@@ -0,0 +1,195 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+
+namespace NuGetGallery.Infrastructure.Authentication
+{
+    public static class Base32Encoder
```
joelverhagen: Why have we written our own? Can we take a dependency here? 3rd party seems fine to me.

This is a lot of code and testing for a very general purpose.

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
```diff
@@ -0,0 +1,195 @@
...
+    {
+        private const byte Bitmask = 0x1F;
+        private static readonly char[] _encodingChars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567=".ToCharArray();
+        private const byte PaddingByte = 0x20;
```
joelverhagen: Isn't this a space?

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
joelverhagen: A couple of integration tests would be fine with an external library, but if we own the code, then we have to test everything. Less code = better.

skofman1:
@maartenba , this is the new format for our ApiKeys. This change doesn't involve any protocol changes and there is no change to experience.

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
skofman1: I didn't write this code 😆 Copied from VSTS... The reason I didn't go for a third party package is because there were non Microsoft approved, and I preferred not to go through the approval process. Since this code is copied from an internal team there's also no licensing issues.

tests/NuGetGallery.Facts/Infrastructure/Authentication/ApiKeyV4Facts.cs
```diff
@@ -0,0 +1,112 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Xunit;
+
+namespace NuGetGallery.Infrastructure.Authentication
+{
+    public class ApiKeyV4Facts
```
joelverhagen: If we must own this implementation, can we use the RFC's test vectors?
https://www.rfc-editor.org/rfc/rfc4648.txt

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
joelverhagen: It's a shame it's not in BCL. It's even more a shame that there's no one package the community has converged upon. Having hundreds of different implementations floating around various code bases indicates a gap in what NuGet is good for.

Either we should adopt a community one (Albireo.Base32 looks good) or we should publish our own so other people don't have to do this same work we are doing.

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
shishirx34: it is the '=' character in _encodingChars, index value '32' => "0x20"

agr: It should be called `PaddingCharacterIndex` or something like that. And no reason for it to be of type `byte` it's just confusing.

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
```diff
@@ -0,0 +1,195 @@
...
+            byte[] output = new byte[GetByteCount(base32String)];
+
+            int bitLocation = 0;
+
+            foreach (Char c in base32String.ToUpperInvariant())
+            {
+                int byteOffset = bitLocation / 8;
+                int bitOffset = bitLocation % 8;
+
+                byte val = (byte)Array.IndexOf(_encodingChars, c);
```
agr: This makes this loop to be O(n*m) where `n = base32String.Length` and m = '_encodingChars.Length'. Could work faster, but probably does not matter until we process megabytes of Base32.

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
skofman1: Created an issue with .NET team with this question: https://github.com/dotnet/corefx/issues/25393
Looked closely at Albireo.Base32 and I don't feel it's widely enough adopted  by the community for us to take dependency. Also didn't see any activity in their GitHub repo. 
Regarding publishing our own package, lets see what the .NET team's recommendation.

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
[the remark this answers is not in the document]
skofman1: @chenriksson , we change encryption very rarely, so don't think we will need double digits. 
@joelverhagen , fixed.

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        public static bool TryParse(string plaintextApiKey, out ApiKeyV4 apiKey)
+        {
+            apiKey = new ApiKeyV4();
+            return apiKey.TryParseInternal(plaintextApiKey);
+        }
+
+        public bool Verify(string encryptedApiKey)
+        {
+            if (string.IsNullOrEmpty(encryptedApiKey) || encryptedApiKey.Length != IdAndPasswordEncryptedLength)
```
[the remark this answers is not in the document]
skofman1: fixed

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                return false;
+            }
+
+            string encyptedApiKeyIdPart = encryptedApiKey.Substring(0, IdPartBase32Length);
```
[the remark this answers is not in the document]
skofman1: fixed

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
```diff
@@ -0,0 +1,139 @@
...
+            string encryptedApiKeyPasswordPart = encryptedApiKey.Substring(IdPartBase32Length, encryptedApiKey.Length - IdPartBase32Length);
```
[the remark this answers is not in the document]
skofman1: fixed

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            if (string.Compare(IdPart, Normalize(encyptedApiKeyIdPart)) != 0)
```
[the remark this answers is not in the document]
skofman1: fixed

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
```diff
@@ -0,0 +1,139 @@
...
+            // Create Id
+            var randomBytes = new byte[IdPartLengthBytes];
+            using (var rng = new RNGCryptoServiceProvider())
+            {
+                rng.GetNonZeroBytes(randomBytes);
+            }
+
+            byte[] idBytes = new byte[IdPartLengthBytes + IdPrefix.Length];
+            Buffer.BlockCopy(IdPrefix, 0, idBytes, 0, IdPrefix.Length);
+            Buffer.BlockCopy(randomBytes, 0, idBytes, IdPrefix.Length, randomBytes.Length);
```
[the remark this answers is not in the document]
skofman1: done

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            try
+            {
+                var id = plaintextApiKey.Substring(0, IdPartBase32Length);
+                var idBytes = id.AppendBase32Padding().ToUpper().FromBase32String();
+                bool success = idBytes[0] == IdPrefix[0] && idBytes[1] == IdPrefix[1];
+
+                if (success)
+                {
+                    string password = plaintextApiKey.Substring(IdPartBase32Length, plaintextApiKey.Length - IdPartBase32Length);
```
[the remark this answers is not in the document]
skofman1: done

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
skofman1: RFC 4648 defines the alphabet as upper case

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
skofman1: fixed

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
skofman1: fixed

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
skofman1: Moved the check to be part of the decoding loop. Should give the best perf. Let me know if this works.

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
skofman1: This will be used for really small strings.. so don't think the perf here is worth additional code complexity.

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
joelverhagen: Good reason. What does the user see?

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
skofman1: I prefer to leave the commented out code and the warning text in the method documentation in case someone is going to modify this class in the future.

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
skofman1: I lowercase and remove '=' in ApiKeyV4 class

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
```diff
@@ -0,0 +1,202 @@
...
+using System;
+using System.Linq;
+
+namespace NuGetGallery.Infrastructure.Authentication
+{
+    public static class Base32Encoder
+    {
+        private const byte Bitmask = 0x1F;
+        private static readonly char[] _encodingChars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567=".ToCharArray();
+        private const byte PaddingCharacterIndex = 32;
```
joelverhagen: `_encodingChars.IndexOf`

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
```diff
@@ -0,0 +1,202 @@
...
+            byte[] output = new byte[GetByteCount(base32String)];
+
+            int bitLocation = 0;
+
+            foreach (Char c in base32String)
+            {
+                int byteOffset = bitLocation / 8;
+                int bitOffset = bitLocation % 8;
+
+                int index = Array.IndexOf(_encodingChars, c);
```
loic-sharma: >int index = Array.IndexOf(_encodingChars, c); [](start = 16, length = 45)

This is still `O(base32String.Length * _encodingChars.Length)`. I would still recommend a regex, or, a hash set like Joel suggested.

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// <summary>
+        /// Password part of the ApiKey (plaintext)
+        /// </summary>
+        public string PasswordPart { get; private set; }
+
+        private ApiKeyV4()
+        {
+        }
+
+        public static ApiKeyV4 Create()
```
scottbommarito: Need some comments on this, `TryParse`, and `Verify`.

```
/// <summary>
/// Randomly generates a new API key.
/// </summary>
public static ApiKeyV4 Create() { ... }
```

```
/// <summary>
/// Attempts to create an <see cref="ApiKeyV4" /> from with <paramref name="plaintextApiKey"/> as the <see cref="ApiKeyV4.PlaintextApiKey"/>.
/// </summary>
public static bool TryParse(string plaintextApiKey, out ApiKeyV4 apiKey) { ... }
```

```
/// <summary>
/// Compares this API key to the given encrypted API key.
/// </summary>
/// <returns>True if the API keys are identical. False otherwise.</returns>
public bool Verify(string encryptedApiKey) { ... }
```

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            string encryptedApiKeyIdPart = encryptedApiKey.Substring(0, IdPartBase32Length);
+            string encryptedApiKeyPasswordPart = encryptedApiKey.Substring(IdPartBase32Length);
+
+            if (!string.Equals(IdPart, Normalize(encryptedApiKeyIdPart)))
+            {
+                return false;
+            }
+
+            // The verification is not case sensitive. This is to maintain the existing behavior that ApiKey authentication is not case-sensitive.
```
scottbommarito: Important to note here that `Base32` is case-insensitive which is why we can do this.

skofman1:
@blowdart 🔔

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
blowdart: No 3rd parties please, didn't we learn anything from PadLeft? :)

tests/NuGetGallery.Facts/Infrastructure/Authentication/ApiKeyV4Facts.cs
blowdart: This is a good idea

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
blowdart: Prefer hashsets to regexes please. If you must go regex ensure there is a timeout.

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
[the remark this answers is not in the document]
blowdart: You can always go to hex after hitting 9 :)

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        /// <summary>
+        /// Plaintext format of the ApiKey
+        /// </summary>
+        public string PlaintextApiKey { get; private set; }
+
+        /// <summary>
+        /// Encrypted format of the ApiKey. Will be set only if Create() method was used.
+        /// </summary>
+        public string EncryptedApiKey { get; private set; }
```
blowdart: Nit : Don't you mean HashedApiKey? Hashing != Encryption

tests/NuGetGallery.Facts/Infrastructure/Authentication/ApiKeyV4Facts.cs
skofman1: done

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
joelverhagen: We control delete.

src/NuGetGallery/Infrastructure/Authentication/Base32Encoder.cs
skofman1: I moved the correctness validation to be part of the loop, so no perf impact at all.

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
skofman1: done

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
skofman1: The comment here is about the plaintext api key provided by the user not being key sensitive

src/NuGetGallery/Infrastructure/Authentication/ApiKeyV4.cs
skofman1: fixed :(

PR #5045, opened 2017-11-21
Fix broken JsonApiControllerFacts

scottbommarito:
Near the end of the PR, I changed some "!" to "." but did not update the related tests.


shishirx34:
none of the other test failures are related to your change, right?

Issue #4984, opened 2017-11-10
Password reset for accounts created with MSA login is broken


agr:
Reported through a support request.

Repro steps:
1. Get a Microsoft Account;
2. Create new account on nuget.org with that MSA;
3. Go to Account settings -> Password Login -> Enable Password login;
4. Click the link in the email;
5. Enter new password;
6. Observe the error:
![image](https://user-images.githubusercontent.com/102933/32680438-8141d020-c61f-11e7-9604-720d00a5ea2f.png)

shishirx34:
It looks like the user tried to enable password login before he/she confirmed their account.

Please have the user click on the account confirmation link first and then again request link for 'Enabling password login'. The correct link will be sent by gallery and then the user should be able to set the password successfully.

I will send out a PR to fix the 'Account' view to not show password login until the account is confirmed, which should prevent users from hitting this issue again.

PR #5046, opened 2017-11-21
Hide enable password login flow until user verification

shishirx34:
Fix for #4984 - Password reset for accounts created with MSA login is broken

Also, added a redundant server side check which would send invalid password reset links. 

use `?w=1` to ignore whitespace changes.

[UPDATE]
Before confirmation:
image

After confirmation:
image

### DIFF
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -575,7 +575,12 @@ namespace NuGetGallery
             if (oldPassword == null)
             {
                 // User is requesting a password set email
-                await _authService.GeneratePasswordResetToken(user, Constants.PasswordResetTokenExpirationHours * 60);
+                var result = await _authService.GeneratePasswordResetToken(user, Constants.PasswordResetTokenExpirationHours * 60);
+                if (result == PasswordResetResultType.UserNotConfirmed)
+                {
+                    ModelState.AddModelError("ChangePassword", Strings.UserIsNotYetConfirmed);
+                    return AccountView(model);
+                }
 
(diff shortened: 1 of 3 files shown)


skofman1:
What will the UI look like for an unconfirmed account? Can you attach a screenshot?

shishirx34:
@skofman1 - see the updated description

src/NuGetGallery/Views/Users/Account.cshtml
```diff
@@ -252,57 +252,64 @@
...
+                                        Cancel
+                                    </a>
+                                </div>
                             </div>
-                        </div>
+                        }
+                    }
+                    else
+                    {
+                        <p>To enable password login, please confirm your account.</p>
```
chenriksson: Do we need to provide any more guidance on how to confirm their account?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -575,7 +575,12 @@ public virtual ActionResult Profiles(string username, int page = 1)
             if (oldPassword == null)
             {
                 // User is requesting a password set email
-                await _authService.GeneratePasswordResetToken(user, Constants.PasswordResetTokenExpirationHours * 60);
+                var resetResultType = await _authService.GeneratePasswordResetToken(user, Constants.PasswordResetTokenExpirationHours * 60);
```
chenriksson: Maybe add a test for this?

src/NuGetGallery/Views/Users/Account.cshtml
scottbommarito: You can link to the confirmation required page, `account/ConfirmationRequired`

PR #5053, opened 2017-11-21
Remove usage of ApiKey Type for distinguishing between Scoped/non-scoped ApiKeys

skofman1:
This is a preparation step for this work: https://github.com/NuGet/Engineering/issues/927

Since we are going to have V3, V4 ApiKeys doesn't make sense using ApiKey type (V1/V2) to distinguish between scoped/non-scopedm keys.

### DIFF
--- a/src/NuGetGallery.Core/CredentialTypes.cs
+++ b/src/NuGetGallery.Core/CredentialTypes.cs
@@ -55,9 +55,9 @@ namespace NuGetGallery
         /// </summary>
         /// <param name="credential"></param>
         /// <returns></returns>
-        public static bool IsSupportedCredential(Credential credential)
+        public static bool IsSupportedCredential(this Credential credential)
         {
-            return IsViewSupportedCredential(credential) || IsPackageVerificationApiKey(credential.Type);
+            return credential.IsViewSupportedCredential() || IsPackageVerificationApiKey(credential.Type);
         }
 
         /// <summary>
@@ -66,10 +66,15 @@ namespace NuGetGallery
         /// </summary>
         /// <param name="credential"></param>
         /// <returns></returns>
-        public static bool IsViewSupportedCredential(Credential credential)
+        public static bool IsViewSupportedCredential(this Credential credential)
         {
             return SupportedCredentialTypes.Any(credType => string.Compare(credential.Type, credType, StringComparison.OrdinalIgnoreCase) == 0)
                     || credential.Type.StartsWith(ExternalPrefix, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public static bool IsScopedApiKey(this Credential credential)
+        {
+            return IsApiKey(credential.Type) && credential.Scopes != null && credential.Scopes.Any();
         }
     }
 }
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -200,8 +200,9 @@ namespace NuGetGallery.Authentication
                     return null;
                 }
 
-                if (matched.Type == CredentialTypes.ApiKey.V1
-                    && !matched.HasBeenUsedInLastDays(_config.ExpirationInDaysForApiKeyV1))
+                if (CredentialTypes.IsApiKey(matched.Type) &&
+                    !matched.IsScopedApiKey() &&
+                    !matched.HasBeenUsedInLastDays(_config.ExpirationInDaysForApiKeyV1))
                 {
                     // API key credential was last used a long, long time ago - expire it
                     await Auditing.SaveAuditRecordAsync(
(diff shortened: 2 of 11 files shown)


src/NuGetGallery.Core/CredentialTypes.cs
```diff
@@ -66,10 +66,15 @@ public static bool IsSupportedCredential(Credential credential)
...
-        public static bool IsViewSupportedCredential(Credential credential)
+        public static bool IsViewSupportedCredential(this Credential credential)
         {
             return SupportedCredentialTypes.Any(credType => string.Compare(credential.Type, credType, StringComparison.OrdinalIgnoreCase) == 0)
                     || credential.Type.StartsWith(ExternalPrefix, StringComparison.OrdinalIgnoreCase);
         }
+
+        public static bool IsScopedApiKey(this Credential credential)
+        {
+            return IsApiKey(credential.Type) && credential.Scopes != null && credential.Scopes.Any();
```
loic-sharma: Can you use `AnySafe` here like in the `AccountViewModel`?

skofman1: AnySafe is not part of GalleryCore. It's located here: https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/ExtensionMethods.cs#L159

I don't want to move this code for a single use case. If we see more of those in the future we should reconsider.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -200,8 +200,9 @@ private void InitCredentialFormatters()
                     return null;
                 }
 
-                if (matched.Type == CredentialTypes.ApiKey.V1
-                    && !matched.HasBeenUsedInLastDays(_config.ExpirationInDaysForApiKeyV1))
+                if (CredentialTypes.IsApiKey(matched.Type) &&
+                    !matched.IsScopedApiKey() &&
+                    !matched.HasBeenUsedInLastDays(_config.ExpirationInDaysForApiKeyV1))
```
cristinamanum: Do we have unit tests?

src/NuGetGallery/ViewModels/AccountViewModel.cs
```diff
@@ -96,11 +96,11 @@ public class CredentialViewModel
         public string Value { get; set; }
         public TimeSpan? ExpirationDuration { get; set; }
 
-        public bool IsNonScopedV1ApiKey
+        public bool IsNonScopedApiKey
```
cristinamanum: Does not a view need to be changed to consume the new Api instead of the old one?

src/NuGetGallery/ViewModels/ApiKeyViewModel.cs
```diff
@@ -54,7 +54,7 @@ public ApiKeyViewModel(CredentialViewModel cred)
             Description = cred.Description;
             Expires = cred.Expires?.ToString("O");
             HasExpired = cred.HasExpired;
-            IsNonScopedV1ApiKey = cred.IsNonScopedV1ApiKey;
+            IsNonScopedV1ApiKey = cred.IsNonScopedApiKey;
```
cristinamanum: Does the IsNonScopeV1ApiKey needs to be called IsNonScopeApiKey?

src/NuGetGallery/ViewModels/AccountViewModel.cs
skofman1: this is actually not used in any view.. but I didn't find some more spots I need to rename

src/NuGetGallery.Core/CredentialTypes.cs
```diff
@@ -66,10 +66,15 @@ public static bool IsSupportedCredential(Credential credential)
         /// </summary>
         /// <param name="credential"></param>
         /// <returns></returns>
-        public static bool IsViewSupportedCredential(Credential credential)
+        public static bool IsViewSupportedCredential(this Credential credential)
```
chenriksson: `IsSupportedCredential` and `IsViewSupportedCredential` are only used by `UserAuditRecord`.

Can we remove these, along with the `SupportedCredentialTypes` collection?

src/NuGetGallery/ViewModels/ApiKeyViewModel.cs
skofman1: done

src/NuGetGallery.Core/CredentialTypes.cs
skofman1: `IsViewSupportedCredential` is used in UsersController: https://github.com/NuGet/NuGetGallery/blob/ac92cd06eb99c86887d459402912d01d90c9d727/src/NuGetGallery/Controllers/UsersController.cs#L925

Regarding `IsSupportedCredential`, from the comment I see it's there for forward compatibility. Meaning in the case that we have ApiKey.V4 in the DB (because we had to rollback Gallery), but the code doesn't know how to read it, we will just ignore it everywhere. Think it's worth keeping.

PR #5055, opened 2017-11-22
Not using SAS tokens for public containers read URIs

agr:
With @joelverhagen's help found a nasty issue with SAS tokens for blob storage: if expiration time is not specified generated SAS token is invalid and causes HTTP 403 when blob is accessed with that token *even* in public container (which is accessible without any tokens).

It is unclear from documentation whether expiration time should be specified or not and the class that is used to pass that information uses `Nullable<DateTimeOffset>` for the expiration time, which assumes that it's optional. It isn't.

### DIFF
--- a/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Net;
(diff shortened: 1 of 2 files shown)


src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -131,30 +156,25 @@ public CloudBlobCoreFileStorageService(ICloudBlobClient client)
...
+            if (KnownPrivateFolders.Contains(folderName))
+            {
+                return false;
             }
 
-            container = await creationTask;
-            _containers[folderName] = container;
-            return container;
+            throw new InvalidOperationException(
+                String.Format(CultureInfo.CurrentCulture, "The folder name {0} is not supported.", folderName));
```
loic-sharma: Nit: `String` -> `string`

skofman1:
why can't we always create a SAS token with expiration for all containers? Why do we need this separation between public and private?

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
agr: Capitalized `String` is used across the whole file, I don't want all those changes in this PR.

Issue #5059, opened 2017-11-22
Consider adding paging to Manage Packages


chenriksson:
There's a lot of scrolling and long (server-side) response for Manage Packages view when there are a lot of packages. We should consider adding paging to the packages lists.

chenriksson:
Set to P2 - Affects performance, and may be higher priority now w/ organization feature.

scottbommarito:
With some performance fixes, the page load is much more tolerable (https://github.com/NuGet/NuGetGallery/pull/6752). Unfortunately, however, the client-side Javascript is now very expensive, because there is no paging on a list of thousands of packages. For `NuGetTestAccount` on DEV, the page freezes. We may still want to investigate paging for a client-side fix in case accounts on PROD have trouble loading the page, but otherwise the paging is not needed for server-side performance reasons.

scottbommarito:
After testing with `NuGetTestAccount` on INT, which has a similar number of packages (but not organization memberships) as the `Microsoft` account on PROD, we probably won't need to add paging--the page loads quickly and the Javascript is slow, but not broken.

ErikEJ:
@joelverhagen what is the url of "manage packages"??

joelverhagen:
@ErikEJ, https://www.nuget.org/account/Packages after signing in.

ErikEJ:
Thanks... let's see how my PR goes. Might have a look later

Issue #4894, opened 2017-10-24
[Organizations] Show packages owned by a user's organizations in their Manage Packages page and allow filtering


scottbommarito:

chenriksson:
Moving to S128, since I won't have time to finish the PR in S127 due to on-call responsibilities.

chenriksson:
Verification: ensure 'Manage Packages' view still shows user packages; can't verify fully until org onboarding.

chenriksson:
Regression #5173 seen when # packages is >3350. Working on a short-term fix to unblock deployment.

chenriksson:
Closing - can't verify until we onboard an organization. We'll have a bug bash including this scenario at that time.

I'll use #5173 to track Manage Packages performance updates and to verify that there is no regression.

Issue #4862, opened 2017-10-18
It takes 27.41 seconds to load this NuGet page...


bchavez:
![chrome_707](https://user-images.githubusercontent.com/478118/31702141-c670a6d8-b389-11e7-82d2-d5daa415663c.png)

/cc @skofman1 ref: https://github.com/NuGet/NuGetGallery/issues/4859#issuecomment-337455878

https://www.nuget.org/packages/FAKE/

```
$(".version-history tr").length
1627
```

Tool used to check page load performance: https://tools.pingdom.com

----

~~Also, my other NuGet issue #4494 is now 77 days old and still a blocking issue to this day. As a reminder, #4494 is a server-side issue (and optionally a client-side issue) that needs to get fixed. I would ***really appreciate*** it if someone at Microsoft can put some effort into getting these issues fixed instead of pushing issues to a **"future"** label.~~

~~Again, as a quick refresher, I dare anyone to install the `Baseline` NuGet package via **Command Line**:~~

![cmd_712](https://user-images.githubusercontent.com/478118/31702709-9f0898b8-b38d-11e7-81d7-4521ff1060ac.png)

Thank you

loic-sharma:
Thanks for reporting! The Gallery should not display thousands of versions on the package display page. An easy fix here would be to display only the latest 100 versions, and displaying a message if there may be more versions.

bchavez:
The actual reason for me to go to the F# FAKE NuGet page was to look for a specific version in the version history that was published 10 months ago. I needed to go back 10 months so I could resolve an build issue and lock down my dependencies to that specific version number. Additionally, I needed to lock down other dependencies around that similar timeframe. In summary, version history is useful.

I think the best idea is not to truncate information from your website but to make it more accessible.

skofman1:
The page is loading slowly even for packages that have ~100 versions. We need to fix this.

joelverhagen:
There is some pretty bad performance on the most popular packages. I tried the top fix on the search results:


ID | Duration (seconds) | Performance Result
-- | -- | --
Newtonsoft.Json | 6.84 | https://tools.pingdom.com/#!/bxyIsD/https://www.nuget.org/packages/Newtonsoft.Json/
NUnit | 4.3 | https://tools.pingdom.com/#!/vLiaw/https://www.nuget.org/packages/NUnit/
[...]

PR #5060, opened 2017-11-22
Organizations: Add owner filter to Manage Packages view

chenriksson:
Issue #4894 - Adds an owner filter to the 'Manage Packages' view. See [spec](https://microsoft.sharepoint.com/:p:/r/teams/NuGet/_layouts/15/Doc.aspx?sourcedoc=%7B3EA306C5-ADCE-4A3D-9455-F87A43293B8B%7D&file=NuGet-Organizations.pptx&action=edit&mobileredirect=true)

Changes include:
- Manage Packages section now uses ko data-binding, based on the pattern in ApiKeys.cshtml
- Cleaned up section headings for consistency, and so it's apparent that the filter only applies to the first section

Screenshot:
image

@joelverhagen Regarding performance, it's not too bad client side. I tested with >1000 and >1500 packages (Microsoft on PROD has ~1500). Server response took 11.38s for 1000, and 21.3s for 1500. Client took ~1s for the initial data-binding, but the filtering was pretty quick. I can share some videos of the response times.
*(Note: I can look into this server-side perf when I work on #4862)*

@scottbommarito Per your suggestion, I think we should add paging.... I opened #5059 for this.

/CC @anangaur


chenriksson:
@joelverhagen I just realized I didn't work on fixing the odd filter drop-down alignment... will try to make it look nicer.

skofman1:
> I tested with >1000 and >1500 packages (Microsoft on PROD has ~1500). Server response took 11.38s for 1000, and 21.3s for 1500
Why is the server side is taking so long? Is the query complex?
Why is the perf not linear? 1000 packages - 11.38 seconds, 1500 - almost twice..

skofman1:
Was accessibility tested?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -284,8 +284,30 @@ public virtual ActionResult Thanks()
...
-                .Select(p => new ListPackageItemViewModel(p)).OrderBy(p => p.Id).ToList();
+
+            var owners = new List<ListPackageOwnerViewModel> {
+                new ListPackageOwnerViewModel
+                {
+                    Username = "All packages"
+                },
+                new ListPackageOwnerViewModel(user)
+                {
+                    CanManagePackageOwners = PermissionsService.IsActionAllowed(user, user, PackageActions.ManagePackageOwners)
```
skofman1: named params will be helpful here

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -1,11 +1,94 @@
...
+﻿(function () {
     'use strict';
+    
+    $(function () {
+        function PackageListItemViewModel(parent, packageItem) {
+            var self = this;
 
-    window.nuget.configureExpanderHeading("packages-Published");
-    window.nuget.configureExpanderHeading("packages-Unlisted");
+            this.Parent = parent;
```
scottbommarito: nit: it may be more readable to rename this to something that makes it clearer what type the parent is

```
this.ParentList = parent;
```

and then in `PackagesListViewModel`

```
this.ParentPage = parent;
```

src/NuGetGallery/Views/Users/_UserPackagesList.cshtml
```diff
@@ -15,7 +15,7 @@
                 <span>My @Model.Name Packages</span>
             </a>
         </h2>
-        <div class="panel-collapse collapse in" id="packages-@Model.Name" aria-expanded="true">
+        <div class="panel-collapse collapse in" id="packages-@Model.Name" aria-expanded="true" data-bind="visible: ListVisible">
```
scottbommarito: is this file used anymore?

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -5,37 +5,218 @@
...
+                        @<text>
+                            @Html.Partial("_OwnerRequestsList", Model.OwnerRequests.Outgoing)
+                        </text>)
+                }
+            </div>
+        </div>
     }
 </section>
 
+<script type="text/html" id="manage-packages">
```
scottbommarito: This file is beginning to get cluttered. We should keep the knockout html stuff for the packages portion of the page in its own partial (`_UserPackagesList.cshtml` perhaps). If possible, we should also keep the knockout js stuff for the packages portion of the page out of `page-manage-packages.js` and into its own file.

chenriksson:
@skofman1 I opened https://github.com/NuGet/NuGetGallery/issues/5120 to optimize the server-side performance that was mentioned in this PR. I'll track this investigation / fix separately.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -284,8 +284,36 @@ public virtual ActionResult Thanks()
         public virtual ActionResult Packages()
         {
             var user = GetCurrentUser();
-            var packages = _packageService.FindPackagesByAnyMatchingOwner(user, includeUnlisted: true)
-                .Select(p => new ListPackageItemViewModel(p)).OrderBy(p => p.Id).ToList();
+
+            var owners = new List<ListPackageOwnerViewModel> {
```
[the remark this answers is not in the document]
chenriksson: I wouldn't know without measuring it... may be worth benchmarking. :-)

src/NuGetGallery/Views/Users/Packages.cshtml
chenriksson: This is consistent with the pattern in ApiKeys.cshtml; ko template in the same file as the view. I'm going to hold off, unless we want to set this as a general team guideline for UI.

/cc @joelverhagen in case he has input here.

src/NuGetGallery/Views/Users/_UserPackagesList.cshtml
chenriksson: Nope

PR #5063, opened 2017-11-22
Add configuration for user package deletes

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/921.

This PR has no functional impact on the gallery but should give a good picture of what the user package delete feature looks like.

NuGet.org's will probably look like this:
```xml
    <add key="PackageDelete.AllowUsersToDeletePackages" value="true" />
    <add key="PackageDelete.MaximumDownloadsForPackageId" value="125000" />
    <add key="PackageDelete.StatisticsUpdateFrequencyInHours" value="24" />
    <add key="PackageDelete.HourLimitWithMaximumDownloads" value="72" />
    <add key="PackageDelete.MaximumDownloadsForPackageVersion" value="100" />
```

The 125,000 is a value that Karan and I agreed upon based off of the number of downloads in the first 24 hours for the top 500 latest packages in the past 6 months. It is a protective measure we can remove later if we find it too cumbersome. I would like to start on the safe side for this feature and do small adjustments later as we build experience with the feature.

I am doing this feature in small PRs since it is a low priority and I can only work on it between other things.

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -72,6 +72,9 @@ namespace NuGetGallery
                .AsSelf()
                .As<FeatureConfiguration>();
 
+            builder.Register(c => configuration.PackageDelete)
+                .As<IPackageDeleteConfiguration>();
+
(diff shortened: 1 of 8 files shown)


joelverhagen:
🔔

PR #5070, opened 2017-11-27
Clean support requests

cristinamanum:
Clean support requests on account delete action.

### DIFF
--- a/src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
+++ b/src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
@@ -326,7 +326,7 @@ namespace NuGetGallery.Areas.Admin
             }
             var userCreatedIssues = GetIssues().Where(i => string.Equals(i.CreatedBy, createdBy)).ToList();
             // Delete all the support requests with exception of the delete account request.
-            // For the delete account support request the user data.
+            // For the DeleteAccount support request clean the user data.
             foreach(var issue in userCreatedIssues.Where(i => !string.Equals(i.IssueTitle, Strings.AccountDelete_SupportRequestTitle)))
             {
                 _supportRequestDbContext.Issues.Remove(issue);
--- a/src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
+++ b/src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
@@ -318,6 +318,32 @@ namespace NuGetGallery.Areas.Admin
             return issue?.Name;
         }
 
+        public async Task<bool> DeleteSupportRequestsAsync(string createdBy)
+        {
+            if(createdBy == null)
+            {
+                throw new ArgumentNullException(nameof(createdBy));
+            }
+            var userCreatedIssues = GetIssues().Where(i => string.Equals(i.CreatedBy, createdBy)).ToList();
+            // Delete all the support requests with exception of the delete account request.
+            // For the delete account support request the user data.
+            foreach(var issue in userCreatedIssues.Where(i => !string.Equals(i.IssueTitle, Strings.AccountDelete_SupportRequestTitle)))
+            {
+                _supportRequestDbContext.Issues.Remove(issue);
+            }
+            foreach(var deletedIssue in userCreatedIssues.Where(i => string.Equals(i.IssueTitle, Strings.AccountDelete_SupportRequestTitle)))
+            {
+                deletedIssue.OwnerEmail = "null";
+                deletedIssue.CreatedBy = null;
+                foreach(var historyEntry in deletedIssue.HistoryEntries)
+                {
+                    historyEntry.EditedBy = null;
+                }
+            }
+            await _supportRequestDbContext.CommitChangesAsync();
+            return true;
+        }
+
         private IQueryable<Issue> GetFilteredIssuesQueryable(int? assignedTo = null, string reason = null, int? issueStatusId = null, string galleryUsername = null)
         {
             IQueryable<Issue> queryable = _supportRequestDbContext.Issues
--- a/src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
+++ b/src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
@@ -324,7 +324,7 @@ namespace NuGetGallery.Areas.Admin
             {
                 throw new ArgumentNullException(nameof(createdBy));
             }
-            var userCreatedIssues = GetIssues().Where(i => string.Equals(i.CreatedBy, createdBy)).ToList();
+            var userCreatedIssues = GetIssues().Where(i => string.Equals(i.CreatedBy, createdBy, StringComparison.InvariantCultureIgnoreCase)).ToList();
             // Delete all the support requests with exception of the delete account request.
             // For the DeleteAccount support request clean the user data.
             foreach(var issue in userCreatedIssues.Where(i => !string.Equals(i.IssueTitle, Strings.AccountDelete_SupportRequestTitle)))
@@ -337,7 +337,10 @@ namespace NuGetGallery.Areas.Admin
                 accountDeletedIssue.CreatedBy = null;
                 foreach(var historyEntry in accountDeletedIssue.HistoryEntries)
                 {
-                    historyEntry.EditedBy = null;
+                    if (string.Equals(historyEntry.EditedBy, createdBy, StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        historyEntry.EditedBy = null;
+                    }
                 }
             }
             await _supportRequestDbContext.CommitChangesAsync();
--- a/src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
+++ b/src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
@@ -318,7 +318,7 @@ namespace NuGetGallery.Areas.Admin
             return issue?.Name;
         }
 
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
```diff
@@ -318,6 +318,32 @@ public string GetIssueStatusNameById(int id)
...
+            var userCreatedIssues = GetIssues().Where(i => string.Equals(i.CreatedBy, createdBy)).ToList();
+            // Delete all the support requests with exception of the delete account request.
+            // For the DeleteAccount support request clean the user data.
+            foreach(var issue in userCreatedIssues.Where(i => !string.Equals(i.IssueTitle, Strings.AccountDelete_SupportRequestTitle)))
+            {
+                _supportRequestDbContext.Issues.Remove(issue);
+            }
+            foreach(var deletedIssue in userCreatedIssues.Where(i => string.Equals(i.IssueTitle, Strings.AccountDelete_SupportRequestTitle)))
+            {
+                deletedIssue.OwnerEmail = "null";
```
scottbommarito: This may be confused with https://www.nuget.org/profiles/null

We should change this to `(deleted)` or something that is very explicit about what happened to the user.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -119,6 +124,7 @@ AuthenticationService authService
                 {
                     await DeleteGalleryUserAccountImplAsync(userToBeDeleted, admin, signature, unlistOrphanPackages);
                 }
+                await RemoveSupportRequests(userToBeDeleted);
```
scottbommarito: I assume that you'd like this to be a part of the commit above. Because of GDPR, it would be bad if we deleted the account but not the support requests.

```
var deletionTask = async () =>
{
    await DeleteGalleryUserAccountImplAsync(userToBeDeleted, admin, signature, unlistOrphanPackages);
    await RemoveSupportRequests(userToBeDeleted);
}

if (commitAsTransaction)
{
    using (var strategy = new SuspendDbExecutionStrategy())
    using (var transaction = _entitiesContext.GetDatabase().BeginTransaction())
    {
        await deletionTask();
        transaction.Commit();
    }
}
else
{
    await deletionTask();
}
```

tests/NuGetGallery.Facts/Services/SupportRequestServiceFacts.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Threading.Tasks;
+using NuGetGallery.Areas.Admin;
+using NuGetGallery.Areas.Admin.Models;
+using NuGetGallery.Configuration;
+using Moq;
+using Xunit;
+
+namespace NuGetGallery.Services
+{
+    public class SupportRequestServiceFacts
```
scottbommarito: Does anything else in the `SupportRequestService` need better tests?

We should file an issue for test cases in this area if so.

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
```diff
@@ -318,6 +318,32 @@ public string GetIssueStatusNameById(int id)
...
+                throw new ArgumentNullException(nameof(createdBy));
+            }
+            var userCreatedIssues = GetIssues().Where(i => string.Equals(i.CreatedBy, createdBy)).ToList();
+            // Delete all the support requests with exception of the delete account request.
+            // For the DeleteAccount support request clean the user data.
+            foreach(var issue in userCreatedIssues.Where(i => !string.Equals(i.IssueTitle, Strings.AccountDelete_SupportRequestTitle)))
+            {
+                _supportRequestDbContext.Issues.Remove(issue);
+            }
+            foreach(var deletedIssue in userCreatedIssues.Where(i => string.Equals(i.IssueTitle, Strings.AccountDelete_SupportRequestTitle)))
```
skofman1: accountDeleteIssues?

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
```diff
@@ -318,6 +318,32 @@ public string GetIssueStatusNameById(int id)
...
+            {
+                deletedIssue.OwnerEmail = "null";
+                deletedIssue.CreatedBy = null;
+                foreach(var historyEntry in deletedIssue.HistoryEntries)
+                {
+                    historyEntry.EditedBy = null;
```
skofman1: EditedBy is the Admin name (our team), no?

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
```diff
@@ -318,6 +318,32 @@ public string GetIssueStatusNameById(int id)
...
+            if(createdBy == null)
+            {
+                throw new ArgumentNullException(nameof(createdBy));
+            }
+            var userCreatedIssues = GetIssues().Where(i => string.Equals(i.CreatedBy, createdBy)).ToList();
+            // Delete all the support requests with exception of the delete account request.
+            // For the DeleteAccount support request clean the user data.
+            foreach(var issue in userCreatedIssues.Where(i => !string.Equals(i.IssueTitle, Strings.AccountDelete_SupportRequestTitle)))
+            {
+                _supportRequestDbContext.Issues.Remove(issue);
```
skofman1: does this delete the history as well?

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
```diff
@@ -318,6 +318,32 @@ public string GetIssueStatusNameById(int id)
...
+            {
+                deletedIssue.OwnerEmail = "null";
+                deletedIssue.CreatedBy = null;
+                foreach(var historyEntry in deletedIssue.HistoryEntries)
+                {
+                    historyEntry.EditedBy = null;
+                }
+            }
+            await _supportRequestDbContext.CommitChangesAsync();
+            return true;
```
skofman1: if the return value is always true, do we need it?

src/NuGetGallery/Services/DeleteAccountService.cs
skofman1: another option is to delete the support requests first, and then if it fails we can still retry the operatrion

---
In reply to: [153327702](https://github.com/NuGet/NuGetGallery/pull/5070#discussion_r153327702) [](ancestors = 153327702)

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
cristinamanum: Makes sense. Updated.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: @scottbommarito Yes it would be desired to be in the same transaction as above. Unfortunately the support is in a different database than the gallery. TransactionScope works when the two databases are on the same server but it does not when on different servers. I am not aware of any other way of achieving distributed transactions. 
@skofman1 If the support will be deleted first the DeletedAccount support request will have the data cleaned as well. This will  have negative impact on the user experience. 
With the current proposal, if the support request clean step fails the admin will need to do an extra manual step that will be documented in the play book.

tests/NuGetGallery.Facts/Services/SupportRequestServiceFacts.cs
cristinamanum: Added an issue to track this work.

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
cristinamanum: It can be the user as well. For example when the issue is created the EditedBy is the username.

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
cristinamanum: Yes.

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
cristinamanum: Updated

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
cristinamanum: No we do do not. Updated.

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
skofman1: can we clear "EditedBy" only if the person is the user? This will leave us the admin information if we want to investigate who worked on the request.

src/NuGetGallery/Services/DeleteAccountService.cs
skofman1: If we delete the support requests first the admin can simply retry the action, which will take only a few seconds, and the chance the user will notice this are very slim..

scottbommarito: @cristinamanum Can you add a comment describing this to the code? It should be explicitly called out.

cristinamanum: Added comments, inversed the order and fixed the user experience by updating the query that searches for active delete support requests.

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
cristinamanum: Added.

PR #5079, opened 2017-11-28
Add the certificates table to enable an approved list of signing certificates

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/935.


src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -288,14 +294,31 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
...
+                .HasRequired(a => a.DeletedBy)
+                .WithMany()
+                .WillCascadeOnDelete(false);
+
+            modelBuilder.Entity<Certificate>()
+                .HasKey(c => c.Key);
 
+            modelBuilder.Entity<Certificate>()
+                .Property(c => c.Thumbprint)
+                .HasMaxLength(256)
```
dtivel: An [earlier comment](https://github.com/NuGet/NuGetGallery/pull/5079/files#diff-394a5ce9800820f0754063e5386fd4feR15) said the values would be limited to 64 characters.

joelverhagen: Good point. I'll clarify this in the model's XML docs.

joelverhagen: Fixed.

PR #5080, opened 2017-11-28
Temporary package size workaround

chenriksson:
Temporary increase to 350MB, based on earlier change here:
https://github.com/NuGet/NuGetGallery/pull/4848

Issue: https://github.com/NuGet/Engineering/issues/984


skofman1:
I remember from Shishir's PR that there were two places he needed to change, one for website upload and one for API upload

chenriksson:
@skofman1 We already had an IIS limit (maxAllowedContentLength) in place, but it didn't restrict the CLI upload. It looks like this was the only change that we made.

see @shishirx34 comment here: https://github.com/NuGet/NuGetGallery/issues/4823

PR #5083, opened 2017-11-29
Remove usernames from the AI urls

cristinamanum:
A set of URLs contain user names. This info is saved AI. With this change this info is removed before saving it in AI.


src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,83 @@
...
+                var currentUserName = GetOwingContext(httpContext).GetCurrentUser().Username;
+                var uriString = telemetryItem.Url.ToString().Replace(currentUserName, DefaultTelemetryUserName);
+                return new Uri(uriString);
+            }
+            return telemetryItem.Url;
+        }
+
+        protected virtual bool IsPIIOperation(string operationName)
+        {
+            string[] piiActions = new string[]{"GET Users/Profiles",
```
joelverhagen: this should be a `static readonly HashSet<string>`

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace NuGetGallery
+{
+    public class ClientTelemetryPIIProcessor : ITelemetryProcessor
+    {
+        /// <summary>
+        /// Default user name that will replace the real user name.
+        /// This value will be saved in AI instead of the real value.
+        /// </summary>
+        private const string DefaultTelemetryUserName = "username";
```
joelverhagen: could we make this look more like a placeholder? "{USERNAME}" or "HIDDEN_USERNAME"

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,83 @@
...
+                "GET Packages/CancelPendingOwnershipRequest"};
+            return piiActions.Contains(operationName);
+        }
+
+        protected virtual HttpContextBase GetHttpContext()
+        {
+            return new HttpContextWrapper(HttpContext.Current);
+        }
+
+        protected virtual IOwinContext GetOwingContext(HttpContextBase httpContext)
```
chenriksson: Spelling: GetOwinContext

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+
+        private Uri GetUri(RequestTelemetry telemetryItem)
+        {
+            if(IsPIIOperation(telemetryItem.Context.Operation.Name))
+            {
+                return new Uri($"{telemetryItem.Url.Scheme}://{telemetryItem.Url.Host}");
+            }
+            var httpContext = GetHttpContext();
```
chenriksson: Per discussion - would remove this default case which could strip username from strings not intended as usernames.

Instead, suggest we add a unit test for Routes that detects if any new routes are configured with usernames. This will enforce us to update the whitelist in this case.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,83 @@
...
+            bool requestIsAuthenticated = httpContext != null ? (httpContext.Request != null ? httpContext.Request.IsAuthenticated : false) : false;
+            if(requestIsAuthenticated)
+            {
+                var currentUserName = GetOwingContext(httpContext).GetCurrentUser().Username;
+                var uriString = telemetryItem.Url.ToString().Replace(currentUserName, DefaultTelemetryUserName);
```
joelverhagen: Usernames are case insensitive

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Updated.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Updated.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Updated - the previous implementation would have removed more than necessary information. For example if package names will have same pattern as user names.

cristinamanum: Added the route test.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: This part was removed.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    public class ClientTelemetryPIIProcessor : ITelemetryProcessor
+    {
+        /// <summary>
+        /// Default user name that will replace the real user name.
+        /// This value will be saved in AI instead of the real value.
+        /// </summary>
+        public const string DefaultTelemetryUserName = "HiddenUserName";
+
+        public static readonly HashSet<string> PiiActions = new HashSet<string>{
```
chenriksson: Can/should this class be internal? Gallery tests should have internals visible to.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+            "Packages/ConfirmPendingOwnershipRequest",
+            "Packages/RejectPendingOwnershipRequest",
+            "Packages/CancelPendingOwnershipRequest",
+            "Users/Confirm",
+            "Users/Delete",
+            "Users/Profiles",
+            "Users/ResetPassword"};
+
+        private ITelemetryProcessor Next { get; set; }
```
chenriksson: Is the setter necessary?

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            ModifyItem(item);
+            this.Next.Process(item);
+        }
+
+        private void ModifyItem(ITelemetry item)
+        {
+            if(item != null && item is RequestTelemetry)
+            {
+                ((RequestTelemetry)item).Url = GetUri((RequestTelemetry)item);
```
chenriksson: Recommend reducing the # of casts here:
```
var telmetryItem = item as RequestTelemetry;
if (telemetryItem != null)
{
    telemetryItem.Url = GetUri(telemetryItem);
}
```

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                return new Uri($"{telemetryItem.Url.Scheme}://{telemetryItem.Url.Host}/{DefaultTelemetryUserName}");
+            }
+            return telemetryItem.Url;
+        }
+
+        protected virtual bool IsPIIOperation(string operationName)
+        {
+            // Remove the verb from the operation name.
+            return PiiActions.Contains(operationName.Split(' ').Last());
```
chenriksson: Will operationName ever be null or empty? If so, should we just return false in that case?

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+            "Packages/RejectPendingOwnershipRequest",
+            "Packages/CancelPendingOwnershipRequest",
+            "Users/Confirm",
+            "Users/Delete",
+            "Users/Profiles",
+            "Users/ResetPassword"};
+
+        private ITelemetryProcessor Next { get; set; }
+
+        public ClientTelemetryPIIProcessor(ITelemetryProcessor next)
```
chenriksson: Is next always non-null?

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                ((RequestTelemetry)item).Url = GetUri((RequestTelemetry)item);
+            }
+        }
+
+        private Uri GetUri(RequestTelemetry telemetryItem)
+        {
+            if(IsPIIOperation(telemetryItem.Context.Operation.Name))
+            {
+                return new Uri($"{telemetryItem.Url.Scheme}://{telemetryItem.Url.Host}/{DefaultTelemetryUserName}");
```
chenriksson: Was debating whether we could get the request route data and just substitute the username part, but maybe that's too complex for each AI call.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
loic-sharma: Alternative:

```
if (item is RequestTelemetry telemetryItem)
{
  telemetryItem.Url = GetUri(telemetryItem);
}
```

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: It should not but I added the check.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: If we will need it we can add this later.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Updated.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Yes - it should net be null.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Not needed - updated.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
skofman1: Can you add a comment with an example of how the processes URL will look

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        private void ModifyItem(ITelemetry item)
+        {
+            if(item != null && item is RequestTelemetry)
+            {
+                ((RequestTelemetry)item).Url = GetUri((RequestTelemetry)item);
+            }
+        }
+
+        private Uri GetUri(RequestTelemetry telemetryItem)
```
skofman1: CleanUri? ModifyUri? Get* sounds like we retrieve a value, but here we perform of modification

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
skofman1: Can you add a comment on what's the expected input and why this split works?

tests/NuGetGallery.Facts/Telemetry/ClientTelemetryPIIProcessorTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public void UrlIsUpdatedOnPIIAction(bool actionIsPII)
+        {
+            // Arange
+            string userName = "user1";
+            var piiProcessor = CreatePIIProcessor(actionIsPII, userName);
```
skofman1: Consider modifying the test to use different input URLs that will cause the processor to either consider them PII or not instead of mocking the behavior using inheritance.

tests/NuGetGallery.Facts/Telemetry/ClientTelemetryPIIProcessorTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            protected override bool IsPIIOperation(string operationName)
+            {
+                return _isPIIOperation;
+            }
+        }
+
+        private List<string> GetPIIOperationsFromRoute()
+        {
+            RouteCollection currentRoutes = new RouteCollection();
```
skofman1: nit:var

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
chenriksson: To clarify - Gallery internals are *already* visible to tests. Just suggesting we reduce public surface area when it's not necessary.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Updated

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Added.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Added.

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Made the hashset and const internal.

tests/NuGetGallery.Facts/Telemetry/ClientTelemetryPIIProcessorTests.cs
cristinamanum: I could not mock the TelemteryContext . I added two more tests to ensure test coverage for IsPIIOperation.

tests/NuGetGallery.Facts/Telemetry/ClientTelemetryPIIProcessorTests.cs
cristinamanum: Updated

PR #5096, opened 2017-11-30
Fix int overflow for user total downloads

chenriksson:


tests/NuGetGallery.Facts/ViewModels/UserProfileModelFacts.cs
```diff
@@ -0,0 +1,39 @@
...
+                var user = new User("theUser");
+                var packageViewModel = new ListPackageItemViewModel(new Package
+                {
+                    PackageRegistration = new PackageRegistration
+                    {
+                        DownloadCount = int.MaxValue
+                    },
+                    Version = "1.0.0"
+                });
+                var packages = new List<ListPackageItemViewModel> { packageViewModel, packageViewModel };
```
joelverhagen: This should really pass two different versions, since having two of the same versions in here is an unexpected state.

chenriksson: Will fix - I was being lazy :-)

chenriksson:
@joelverhagen I opened an issue to track that:
https://github.com/NuGet/Engineering/issues/989

PR #5097, opened 2017-11-30
PermissionsService 2.0: ActionsRequringPermission

scottbommarito:
I'm merging this into a feature branch because I want to add some more unit tests with future PRs, but would like to merge this in now and then start working on unit tests of controller methods.

- Rename old `PermissionsService` into `PermissionsHelpers`
- Rename `PermissionLevel` to `PermissionsRequirement`
- Add `ActionRequiringAccountPermissions`, `IActionRequiringEntityPermissions`, `ActionRequiringEntityPermissions`, `ActionRequiringPackagePermissions` and `ActionRequiringReservedNamespacePermissions` for consolidating permissions logic
- Add static `ActionsRequiringPermissions` to store statically declared instances of the classes described above
-Replace old `PermissionsService` calls with new `ActionsRequiringPermissions` calls.

Old API:
```
if (!PermissionsService.IsActionAllowed(accountOnBehalfOf, HttpContext.User, AccountActions.UploadNewVersionOnBehalfOf))
{
    ...
}

if (!PermissionsService.IsActionAllowed(package, accountOnBehalfOf, PackageActions.UploadNewVersion))
{
    ...
}
```

New API:
```
if (ActionsRequiringPermissions.UploadNewPackageVersion.IsAllowed(currentUser, accountOnBehalfOf, package) != PermissionsFailure.None)
{
    ...
}
```


src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -414,8 +413,7 @@ public virtual Task<ActionResult> CreatePackagePost()
                             }
 
                             // For a new package id verify that the user is allowed to push to the matching namespaces, if any.
-                            var isPushAllowed = ReservedNamespaceService.IsPushAllowed(id, user, out userOwnedNamespaces);
-                            if (!isPushAllowed)
+                            if (!ActionsRequiringPermissions.UploadNewPackageId.TryGetAccountsIsAllowedOnBehalfOf(user, new ActionOnNewPackageContext(id, ReservedNamespaceService), out var accountsAllowedOnBehalfOf))
```
joelverhagen: `TryGetAccountsIsAllowedOnBehalfOf` sounds awkward:
```
TryGetAccountIsAllowedOnBehalfOf
TryGetAccountsAreAllowedOnBehalfOf
```

joelverhagen: How is this `out` parameter used?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -298,26 +298,21 @@ public virtual JsonResult UploadPackageProgress()
 
                 var id = nuspec.GetId();
                 var packageRegistration = _packageService.FindPackageRegistrationById(id);
+                IEnumerable<User> accountsAllowedOnBehalfOf;
                 // For a new package id verify if the user is allowed to use it.
-                if (packageRegistration == null)
+                if (packageRegistration == null && !ActionsRequiringPermissions.UploadNewPackageId.TryGetAccountsIsAllowedOnBehalfOf(currentUser, new ActionOnNewPackageContext(id, _reservedNamespaceService), out accountsAllowedOnBehalfOf))
```
joelverhagen: Wait, why are we passing a dependency as a parameter? I see that this is a static... could it become an injected thing?

joelverhagen: (That would also facilitate unit testing)

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        /// <summary>
+        /// Is <paramref name="currentPrincipal"/> allowed to perform an action with a requirement of <paramref name="permissionsRequirement"/> on the entity owned by <paramref name="entityOwners"/>?
+        /// </summary>
+        public static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, User currentUser, IEnumerable<User> entityOwners)
+        {
+            if (currentUser == null)
+            {
+                return PermissionLevelsIntersect(PermissionsRequirement.None, permissionsRequirement);
```
joelverhagen: The semantics of intersect suggest a constant output, i.e. `A ∩ ∅ = ∅`

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,142 @@
...
+            }
+
+            if (isUserAdmin &&
+                PermissionLevelsIntersect(PermissionsRequirement.SiteAdmin, permissionsRequirement))
+            {
+                return true;
+            }
+
+            var matchingMembers = entityOwners
+                .Where(o => o is Organization)
```
joelverhagen: This is equivalent to `OfType<Organization>`

joelverhagen:
I am still reviewing this. Very large PR 🤓. Could we have built the framework as one PR and integrated in another?

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: not used in this PR--it's used in the upload flow to determine the list of accounts that the current user can upload a package on behalf of

src/NuGetGallery/Helpers/PermissionsHelpers.cs
scottbommarito: TIL, nice catch

src/NuGetGallery/Services/ActionRequiringPackagePermissions.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            return TryGetAccountsIsAllowedOnBehalfOf(currentUser, ConvertPackageToRegistration(package), out accountsAllowedOnBehalfOf);
+        }
+
+        protected override IEnumerable<User> GetOwners(PackageRegistration packageRegistration)
+        {
+            return packageRegistration != null ? packageRegistration.Owners : Enumerable.Empty<User>();
+        }
+
+        private PackageRegistration ConvertPackageToRegistration(Package package)
```
chenriksson: `Get` is more appropriate than `Convert`. Method doesn't look necessary, would just inline.

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
```diff
@@ -0,0 +1,75 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// A collection of all <see cref="ActionRequiringAccountPermissions"/> and <see cref="IActionRequiringEntityPermissions{TEntity}"/>s.
+    /// </summary>
+    public static class ActionsRequiringPermissions
```
chenriksson: Maybe members should be sorted by entity type, possibly using regions?

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,142 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Principal;
+
+namespace NuGetGallery
+{
+    public class PermissionsHelpers
```
chenriksson: APIs look a lot cleaner!

As mentioned, only concern with static class is that it could be more difficult to integrate auditing/telemetry into the permissions helpers. We can figure it out if that becomes necessary.

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: The `out` parameter returns accounts the action is allowed ON BEHALF OF.

For example, if you were a member of two organizations that owned a package and you called this method on them, the `out` parameter would return both, because you can do the action on behalf of both of them.

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -798,14 +799,18 @@
     <Compile Include="Migrations\201711021733062_ApiKeyOwnerScope.Designer.cs">
       <DependentUpon>201711021733062_ApiKeyOwnerScope.cs</DependentUpon>
     </Compile>
+    <Compile Include="Services\ActionRequiringAccountPermissions.cs" />
+    <Compile Include="Services\ActionRequiringEntityPermissions.cs" />
+    <Compile Include="Services\ActionRequiringPackagePermissions.cs" />
+    <Compile Include="Services\ActionRequiringReservedNamespacePermissions.cs" />
+    <Compile Include="Services\ActionsRequiringPermissions.cs" />
```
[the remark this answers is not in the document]
scottbommarito: Two notes:

1 - We don't actually have a `NuGetGallery.Services` namespace, everything is in `NuGetGallery`.
2 - I think they're ok in services, but TBH I'd prefer to split everything up more granularly and add a `Permissions` folder

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
scottbommarito: I wasn't sure how to sort the file, but entity type sounds better.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Linq;
+using System.Security.Principal;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// An action requiring permissions on an entity that can be done on behalf of another <see cref="User"/>.
+    /// </summary>
+    public abstract class ActionRequiringEntityPermissions<TEntity>
+        : IActionRequiringEntityPermissions<TEntity>
```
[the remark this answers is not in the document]
scottbommarito: No--we also allow `ListPackageItemViewModel` to be here.

I was considering making a superclass (`IOwnedEntity`) and making `PackageRegistration`, `ReservedNamespace`, `User`, and `ListPackageItemViewModel` extend it, but I decided it would be better to leave it alone.

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,142 @@
...
+    {
+        /// <summary>
+        /// Is <paramref name="currentPrincipal"/> allowed to perform an action with a requirement of <paramref name="permissionsRequirement"/> on <paramref name="packageRegistration"/>?
```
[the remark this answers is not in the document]
scottbommarito: IMO line length should be something handled by the tooling.

Not sure the best way to fix the line length, but I was under the impression the only long lines are the comments.

Also this file is 90% the same from before besides renaming.

src/NuGetGallery/Controllers/ApiController.cs
cristinamanum: Does the return value of TryGetAccountsIsAllowedOnBehalfOf have any  correlation with the collection count of the out parameter?
If not, would it be better to have two methods instead of one?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -414,12 +409,14 @@ public virtual JsonResult UploadPackageProgress()
...
-            
+
             // Validating packages should be hidden to everyone but the owners and admins.
+            var currentUser = GetCurrentUser();
+            IEnumerable<User> accountsAllowedOnBehalfOf;
             if (package == null
                 || ((package.PackageStatusKey == PackageStatus.Validating
                      || package.PackageStatusKey == PackageStatus.FailedValidation)
-                    && !PermissionsService.IsActionAllowed(package, User, PackageActions.DisplayPrivatePackage)))
+                    && !ActionsRequiringPermissions.DisplayPrivatePackageMetadata.TryGetAccountsIsAllowedOnBehalfOf(currentUser, package, out accountsAllowedOnBehalfOf)))
```
cristinamanum: nit: shorter lines?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -414,12 +409,14 @@ public virtual JsonResult UploadPackageProgress()
             {
                 package = _packageService.FindPackageByIdAndVersion(id, version, SemVerLevelKey.SemVer2);
             }
-            
+
             // Validating packages should be hidden to everyone but the owners and admins.
+            var currentUser = GetCurrentUser();
+            IEnumerable<User> accountsAllowedOnBehalfOf;
```
cristinamanum: Is this used anywhere?

scottbommarito: just to contain the unused `out` parameter

the `out` parameter will be used in future upload flows

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+        
+        protected abstract PermissionsFailure IsAllowedOnEntity(User account, TEntity entity);
+
+        /// <summary>
+        /// Determines whether <paramref name="currentPrincipal"/> can perform this action on <paramref name="entity"/> on behalf of any <see cref="User"/>.
+        /// </summary>
+        /// <param name="accountsAllowedOnBehalfOf">A <see cref="IEnumerable{User}"/> containing all accounts that <paramref name="currentUser"/> can perform this action on <paramref name="entity"/> on behalf of.</param>
+        /// <returns>True if and only if <paramref name="currentPrincipal"/> can perform this action on <paramref name="entity"/> on behalf of any <see cref="User"/>.</returns>
+        public bool TryGetAccountsIsAllowedOnBehalfOf(User currentUser, TEntity entity, out IEnumerable<User> accountsAllowedOnBehalfOf)
```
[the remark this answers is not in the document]
scottbommarito: The out parameter returns accounts the action is allowed ON BEHALF OF.

In other words, the other method (`IsAllowed`) requires you to specify an account you want to perform the action on behalf of. This method figures out if what accounts you can do the action on behalf of.

For example, if you were a member of two organizations that owned a package and you called this method on them, the `out` parameter would return both, because you can do the action on behalf of both of them. If you were also an owner of the package, it would return you as well.

The method as a whole returns whether or not you can do it on behalf of anyone.

As I mentioned above, the primary use case is determined who you can push a package on behalf of.

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,142 @@
...
+                permissionsRequirement,
+                currentUser.IsAdministrator(),
+                u => currentUser.MatchesUser(u),
+                entityOwners);
+        }
+
+        private static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, bool isUserAdmin, Func<User, bool> isUserMatch, IEnumerable<User> entityOwners)
+        {
+            if ((entityOwners == null || !entityOwners.Any()) &&
+                PermissionLevelsIntersect(PermissionsRequirement.None, permissionsRequirement))
```
cristinamanum: Should not always return true on PermissionRequirement.None ?

scottbommarito: good catch, I can restructure this code to make that check first.

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+                return true;
+            }
+
+            if (entityOwners.Any(isUserMatch) &&
+                PermissionLevelsIntersect(PermissionsRequirement.Owner, permissionsRequirement))
```
cristinamanum: Maybe change the order of the operands to not execute Any(isUserMarch) if not the Owner permissionRequirement

src/NuGetGallery/Services/ActionRequiringAccountPermissions.cs
```diff
@@ -0,0 +1,38 @@
...
+        
+        public ActionRequiringAccountPermissions(PermissionsRequirement accountPermissionsRequirement)
+        {
+            AccountPermissionsRequirement = accountPermissionsRequirement;
+        }
+
+        /// <summary>
+        /// Determines whether <paramref name="currentUser"/> can perform this action on <paramref name="account"/>.
+        /// </summary>
+        public PermissionsFailure IsAllowed(User currentUser, User account)
```
cristinamanum: PermissionFailure suggests a failure. Will this call be always a failure? Why not returning bool?

scottbommarito: `PermissionsCheckResult` instead? Almost all of the `enum` results are failures which is why I called it `PermissionsFailure`, but it probably could be clearer if it was not named that.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,85 @@
...
+                return PermissionsFailure.Account;
+            }
+
+            return IsAllowedOnEntity(account, entity);
+        }
+        
+        protected abstract PermissionsFailure IsAllowedOnEntity(User account, TEntity entity);
+
+        /// <summary>
+        /// Determines whether <paramref name="currentPrincipal"/> can perform this action on <paramref name="entity"/> on behalf of any <see cref="User"/>.
```
cristinamanum: currentPrincipal -> currentUser?

src/NuGetGallery/Helpers/PermissionsHelpers.cs
scottbommarito: applying that logic to all of these checks would probably make it substantially faster...nice catch!

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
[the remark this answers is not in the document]
cristinamanum: Why do not remove the the out and return the IEnum and the caller of this method can do the .Any() ?

scottbommarito:
Closing this PR because I'm planning on splitting it up.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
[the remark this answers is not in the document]
scottbommarito: I prefer the `TryGet` model rather than a boolean because
1 - This method is frequently used simply as a boolean, without concern for the result of the `out` parameter and removing it would add a lot of excess `.Any()` calls around the code
2 - The caller would now need to understand that if an action does not have any accounts it can be performed on behalf of, it is not allowed.

src/NuGetGallery/Services/ActionRequiringPackagePermissions.cs
scottbommarito: I like the naming change.

I added this method because I wanted to guarantee that the method for converting between a `Package` and a `PackageRegistration` is consistent.

PR #5098, opened 2017-11-30
CoreMessageService to share mail sending code

agr:
Also, moved most of the UserExtensions to Core library, since it is used by the CoreMessageService.


tests/NuGetGallery.Core.Facts/Extensions/UserExtensionsFacts.cs
```diff
@@ -17,7 +17,7 @@ public void WhenUserIsNull_ThrowsArgumentNullException()
             {
```
joelverhagen: Nit this file should be called `UserExtensionsCoreFacts`

tests/NuGetGallery.Core.Facts/Extensions/UserExtensionsFacts.cs
```diff
@@ -73,10 +73,11 @@ public void ReturnsApiKeyMatchingClaim(string credentialType)
                 user.Credentials.Add(new Credential(CredentialTypes.ApiKey.V2, "A"));
                 user.Credentials.Add(new Credential(credentialType, "B"));
 
-                var identity = AuthenticationService.CreateIdentity(
-                    new User("testuser"),
-                    AuthenticationTypes.ApiKey,
-                    new Claim(NuGetClaims.ApiKey, "B"));
+                var identity = new ClaimsIdentity(
```
joelverhagen: Why are these different?

agr: There is no `AuthenticationService` in NugetGallery.Core and technically, for test we don't need everything happening there, just claim with API key to be present or absent.

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -734,5 +647,23 @@ public void SendPackageDeletedNotice(Package package, string packageUrl, string
                 }
             }
         }
+
+        override protected void SendMessage(MailMessage mailMessage, bool copySender)
```
joelverhagen: nit `protected override`

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Globalization;
+using System.Linq;
+using System.Net.Mail;
+using AnglicanGeek.MarkdownMailer;
+
+namespace NuGetGallery.Services
+{
+    public class CoreMessageService : ICoreMessageService
+    {
+        protected CoreMessageService()
```
joelverhagen: What is this ctor for?

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+        }
+
+        public CoreMessageService(IMailSender mailSender, ICoreMessageServiceConfiguration coreConfiguration)
+        {
+            MailSender = mailSender ?? throw new ArgumentNullException(nameof(mailSender));
+            CoreConfiguration = coreConfiguration ?? throw new ArgumentNullException(nameof(coreConfiguration));
+        }
+
+        public IMailSender MailSender { get; protected set; }
```
joelverhagen: why not read-only?

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -0,0 +1,108 @@
...
+        public ICoreMessageServiceConfiguration CoreConfiguration { get; protected set; }
+
+        public void SendPackageAddedNotice(Package package, string packageUrl, string packageSupportUrl, string emailSettingsUrl)
```
joelverhagen: You can "out of scope" this but this would be a good time to add a single unit test to this method. Big `string.Format` calls can have runtime issues...

src/NuGetGallery.Core/Services/CoreMessageService.cs
agr: `MessageService` has default constructor, so this guys has to have it, too.

joelverhagen: Oh, gross

src/NuGetGallery.Core/Services/CoreMessageService.cs
agr: `TestableMessageService` - a test wrapper wants write access to it. Likely can be rewritten

src/NuGetGallery.Core/Services/CoreMessageService.cs
agr: https://github.com/NuGet/NuGetGallery/blob/d89c2ba56066743a8ff3bc3660b582d4f86242b0/tests/NuGetGallery.Facts/Services/MessageServiceFacts.cs#L751 has tests for it, I decided not to move it to have all mail tests in one place.

joelverhagen: Sounds good!

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -0,0 +1,108 @@
...
+            string subject = "[{0}] Package published - {1} {2}";
+            string body = @"The package [{1} {2}]({3}) was just published on {0}. If this was not intended, please [contact support]({4}).
+
+-----------------------------------------------
+<em style=""font-size: 0.8em;"">
+    To stop receiving emails as an owner of this package, sign in to the {0} and
+    [change your email notification settings]({5}).
+</em>";
+
+            body = String.Format(
```
skofman1: string with lowercase everywhere

src/NuGetGallery.Core/Extensions/UserExtensionsCore.cs
```diff
@@ -0,0 +1,132 @@
...
+                return new MailAddress(user.UnconfirmedEmailAddress, user.Username);
+            }
+
+            return new MailAddress(user.EmailAddress, user.Username);
+        }
+
+        /// <summary>
+        /// Get the current API key credential, if available.
+        /// </summary>
+        public static Credential GetCurrentApiKeyCredential(this User user, IIdentity identity)
```
skofman1: do you need this method as well? Couldn't find usages

src/NuGetGallery.Core/Extensions/UserExtensionsCore.cs
```diff
@@ -0,0 +1,132 @@
...
+            return user.Credentials.FirstOrDefault(c => c.Value == apiKey);
+        }
+
+        /// <summary>
+        /// Determine if user is direct or indirect (organization admin) package owner.
+        /// </summary>
+        /// <param name="user">User to query.</param>
+        /// <param name="package">Package to query.</param>
+        /// <returns>True if direct or indirect package owner.</returns>
+        public static bool IsOwnerOrMemberOfOrganizationOwner(this User user, PackageRegistration package)
```
skofman1: same

src/NuGetGallery.Core/Extensions/UserExtensionsCore.cs
agr: I only need one extension method from the `UserExtensions`, but I see no reason to leave other methods that can be moved to the Core in the Gallery: `User` class is declared in the Core library, extensions should be there unless they contain Gallery-specific references.

skofman1: There is no reason to have extension methods in Core that are used only in the Gallery. If we need them anywhere else we can move them to Core in the future

src/NuGetGallery.Core/Services/CoreMessageService.cs
agr: We probably need to do it once across the whole project and don't do a spot fixups...

tests/NuGetGallery.Core.Facts/Extensions/UserExtensionsFacts.cs
agr: fixed

src/NuGetGallery/Services/MessageService.cs
agr: fixed

src/NuGetGallery.Core/Extensions/UserExtensionsCore.cs
agr: fixed

joelverhagen:
Have we pinned the signed versions on MyGet? Have we saved the .snk into KeyVault?

PR #5099, opened 2017-11-30
 Create&Consume ApiKey V4

skofman1:
New ApiKeys created will look like this: oy2cshhnw5nmevgh36jajq4opv5nyjnutapjuhl7xb623m


src/NuGetGallery.Core/CredentialTypes.cs
```diff
@@ -47,7 +48,7 @@ public static bool IsPackageVerificationApiKey(string type)
             return type.Equals(ApiKey.VerifyV1, StringComparison.OrdinalIgnoreCase);
         }
         
-        internal static List<string> SupportedCredentialTypes = new List<string> { Password.Sha1, Password.Pbkdf2, Password.V3, ApiKey.V1, ApiKey.V2 };
+        internal static List<string> SupportedCredentialTypes = new List<string> { Password.Sha1, Password.Pbkdf2, Password.V3, ApiKey.V1, ApiKey.V2, ApiKey.V4 };
```
scottbommarito: nit: Would this make more sense if it were read-only, e.g. `IEnumerable` or `IReadOnlyCollection`?

src/NuGetGallery.Core/Entities/Credential.cs
```diff
@@ -106,5 +107,24 @@ public bool HasBeenUsedInLastDays(int numberOfDays)
 
             return true;
         }
+
+        public Credential Clone()
+        {
+            // Creates a copy without referenced Entities, to avoid risk of
```
scottbommarito: nit: This comment would probably be better as an xmldoc of the method.

Also then you can reference `IEntity` directly, e.g. `<see cref="IEntity"/>`

Same with Scopes clone and User clone

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -704,28 +704,20 @@ private Credential FindMatchingCredential(Credential credential)
...
+                .Include(u => u.Scopes);
 
-            return ValidateFoundCredentials(results, "ApiKey");
+            var results = _credentialValidator.ValidateApiKeyCredential(allCredentials, apiKeyCredential.Value);
+
+            return ValidateFoundCredentials(results, Strings.CredentialType_ApiKey);
         }
 
-        private Credential ValidateFoundCredentials(List<Credential> results, string credentialType)
+        private Credential ValidateFoundCredentials(IList<Credential> results, string credentialType)
```
scottbommarito: nit: `IEnumerable` instead of `IList`?

src/NuGetGallery/Authentication/Providers/ApiKey/ApiKeyAuthenticationHandler.cs
```diff
@@ -102,7 +102,7 @@ internal void WriteStatus(string message, int statusCode)
                             AuthenticationService.CreateIdentity(
                                 user, 
                                 AuthenticationTypes.ApiKey, 
-                                new Claim(NuGetClaims.ApiKey, apiKey),
+                                new Claim(NuGetClaims.ApiKey, credential.Value), // For apikey.v4 this will hold the hashed value
```
scottbommarito: 1 - This comment feels out of place. Not every credential that you can authenticate with is an api key V4 (e.g. verification key). I can see it getting out of date if we change the implementation of V4 of create V5 because there's no direct reference to `ApiKeyV4` near this.

3 - Where is any caller using the value of the API key from the claims? Is it possible that this change has broken any of the previous callers because it is no longer true that you can reauthenticate with the key here?

src/NuGetGallery/Authentication/AuthenticationService.cs
skofman1: using IEnumerable might cause multiple enumerations

src/NuGetGallery/Authentication/Providers/ApiKey/ApiKeyAuthenticationHandler.cs
skofman1: 1 - Updated the comment
3 - this change is a fix for callers that use the claims api key value to understand which credential is used in the request. If Claims[NuGetClaims.ApiKey] is plaintext and User.Credentials[0].Value is hashed they can't find the credential and fail.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -782,13 +782,25 @@ public virtual ActionResult PasswordChanged()
...
-            var newCredential = _credentialBuilder.CreateApiKey(expiration);
+            var newCredential = _credentialBuilder.CreateApiKey(expiration, out string plaintextApiKey);
             newCredential.Description = description;
             newCredential.Scopes = scopes;
 
             await _authService.AddCredential(user, newCredential);
 
-            return newCredential;
+            // Replace the encrypted value with the plaintext value for presenting to the user
+            var viewableCredential = CloneCredential(newCredential);
```
joelverhagen: Why isn't this a view model? Clone is a code smell.

joelverhagen: By this I mean that there are now two kinds of "Credential" and "User" instances. One that is "viewable" (not _exactly_ sure what that means) and one that is tracked by an EF context. There is no way for me to tell which is which unless I look around in the code. A new type (like a view model) would be better here.

tests/NuGetGallery.Core.Facts/Entities/CredentialFacts.cs
```diff
@@ -0,0 +1,46 @@
...
+                ExpirationTicks = 10,
+                LastUsed = DateTime.Now - TimeSpan.FromDays(1)
+            };
+
+
+            // Act
+            var clone = credential.Clone();
+
+            // Assert
+            Assert.Equal(credential.Key, clone.Key);
```
joelverhagen: nit clone methods shouldn't refer to the original values, they should refer to constants since it's possible for Clone() to mutate the original.

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -685,6 +685,18 @@ public class TheGenerateApiKeyAction : TestContainer
                     .Setup(u => u.FindByUsername(orgUser.Username))
                     .Returns(orgUser);
 
+                GetMock<AuthenticationService>()
+                .Setup(u => u.AddCredential(
```
joelverhagen: nit indent

joelverhagen:
> oy2cshhnw5nmevgh36jajq4opv5nyjnutapjuhl7xb623m

Beautiful 🥇

src/NuGetGallery/Authentication/AuthenticationService.cs
joelverhagen: Nit `IReadOnlyList` since the method isn't supposed to mutate the list

src/NuGetGallery/Infrastructure/Authentication/CredentialBuilder.cs
```diff
@@ -19,11 +18,15 @@ public Credential CreatePasswordCredential(string plaintextPassword)
                 V3Hasher.GenerateHash(plaintextPassword));
         }
 
-        public Credential CreateApiKey(TimeSpan? expiration)
+        public Credential CreateApiKey(TimeSpan? expiration, out string plaintextApiKey)
```
joelverhagen: Isn't expiration time required these days?

tests/NuGetGallery.Core.Facts/Entities/ScopeFacts.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace NuGetGallery.Entities
+{
+    public class ScopeFacts
+    {
+        public static IEnumerable<object[]> CloneSucceeds_Input
+        {
+            get
+            {
+                return new[]
```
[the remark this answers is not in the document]
joelverhagen: (and make adding more test cases easier  since you don't have to think about the trailing comma)

joelverhagen:
So when will we be disabling the copy button? Future PR right?

src/NuGetGallery.Core/Entities/Scope.cs
```diff
@@ -64,5 +64,10 @@ public Scope(string subject, string allowedAction)
             Subject = subject;
             AllowedAction = allowedAction;
         }
+
+        public Scope Clone()
```
cristinamanum: If a Clone will have the same properties as the original. this is not really a Clone.

src/NuGetGallery.Core/Entities/Credential.cs
```diff
@@ -106,5 +107,25 @@ public bool HasBeenUsedInLastDays(int numberOfDays)
 
             return true;
         }
+
+        /// <summary>
+        /// Creates a copy without referenced <see cref="IEntity"/> to avoid risk of circular clones.
+        /// </summary>
+        public Credential Clone()
```
cristinamanum: Is the goal as Clone.Equals(Original) == true?

src/NuGetGallery/Infrastructure/Authentication/CredentialValidator.cs
```diff
@@ -27,5 +28,26 @@ public bool ValidatePasswordCredential(Credential credential, string providedPas
 
             return validator(providedPassword, credential);
         }
+
+        public IList<Credential> ValidateApiKeyCredential(IQueryable<Credential> allCredentials, string providedApiKey)
```
cristinamanum: Maybe a different name "GetValidCredentialsFromApiKey"? ValidateApiKeyCredential sounds like it should return a bool (is valid or not).

src/NuGetGallery/Authentication/AuthenticationService.cs
skofman1: ValidateApiKeyCredential returns an IList (as it should), and IList is not convertible  to IReadOnlyList. So will leave as is.

src/NuGetGallery/Controllers/UsersController.cs
skofman1: Great feedback! Changed to return a view model. Looks much cleaner now.

tests/NuGetGallery.Core.Facts/Entities/CredentialFacts.cs
skofman1: remove this code

src/NuGetGallery/Infrastructure/Authentication/CredentialBuilder.cs
skofman1: https://github.com/NuGet/NuGetGallery/issues/5107

src/NuGetGallery.Core/Entities/Scope.cs
skofman1: removed the method

src/NuGetGallery.Core/Entities/Credential.cs
skofman1: removed the method

src/NuGetGallery/Authentication/AuthenticationService.cs
scottbommarito: https://msdn.microsoft.com/en-us/library/e78dcd75(v=vs.110).aspx

not sure if u can call this on an `IList` but it exists for `List`

skofman1: It doesn't

skofman1:
🔔 Please review again

src/NuGetGallery/Authentication/AuthenticationService.cs
joelverhagen: `List` implements more that `IList`. @skofman1, my bad.

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -20,12 +20,11 @@ protected MessageService()
         {
         }
 
-        public MessageService(IMailSender mailSender, IAppConfiguration config, AuthenticationService authService)
+        public MessageService(IMailSender mailSender, IAppConfiguration config)
             : this()
         {
             MailSender = mailSender;
             Config = config;
-            AuthService = authService;
```
agr: `AuthService` property should be removed, too?

agr: And from the `TestableMessageService`...

PR #5100, opened 2017-12-01
Added SendPackageUploadedNotice method to MessageService.

agr:
Replaced calls to SendPackageAddedNotice with SendPackageUploadedNotice.
Added tests for the new method.

### DIFF
--- a/src/NuGetGallery/Services/MessageService.cs
+++ b/src/NuGetGallery/Services/MessageService.cs
@@ -647,6 +647,46 @@ The {3} Team";
             }
         }
 
+        public void SendPackageUploadedNotice(Package package, string packageUrl, string packageSupportUrl, string emailSettingsUrl)
+        {
+            string subject = "[{0}] Package uploaded - {1} {2}";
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Services/MessageService.cs
```diff
@@ -614,6 +614,46 @@ public void SendPackageAddedNotice(Package package, string packageUrl, string pa
...
+                CultureInfo.CurrentCulture,
+                body,
+                Config.GalleryOwner.DisplayName,
+                package.PackageRegistration.Id,
+                package.Version,
+                packageUrl,
+                packageSupportUrl,
+                emailSettingsUrl);
+
+            subject = String.Format(CultureInfo.CurrentCulture, subject, Config.GalleryOwner.DisplayName, package.PackageRegistration.Id, package.Version);
```
skofman1: string lowercase

agr: Won't fix: this will be inconsistent with the rest of the file.
Added Engineering issue to fix it globally.

PR #5106, opened 2017-12-01
Update Xunit version

genlu:
This should fix the VS feedback issue reported here:
https://developercommunity.visualstudio.com/content/problem/157928/test-explorer-trying-to-debug-one-test-but-debugs.html


genlu:
FYI @skofman1

skofman1:
Thanks @genlu !

PR #5109, opened 2017-12-04
Package immutability: wave 1

xavierdecoster:
This PR merges the feature branch into dev, and removes the capability to edit package metadata from the edit package page and the verify upload page.

It still relies on the `HandlePackageEdits` job for editing Documentation, which is to be removed in a subsequent wave.


skofman1:
@xavierdecoster , consider moving Documentation editing from `HandlePackageEdits` job to Gallery as part of Wave 1. Should be a very simple fix (save in a different folder), and will leave us with only cleanup work. 
This suggestion of course shouldn't block this merge.

xavierdecoster:
@skofman1 that's tracked as part of https://github.com/NuGet/Engineering/issues/1003.

You say it's *just changing save location*, but please note that all that the Gallery does currently, is saving a `PackageEdit` entity. Moving this logic from the `HandlePackageEdits` job to the gallery may be simple, yet take a little more time than changing a config value somewhere.

PR #5116, opened 2017-12-04
Issue 4258 -  ContactOwners e-mails contain the package version numer next to package id

dvkudr:


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -901,19 +901,19 @@ private void NotifyReportMyPackageSupportRequest(ReportMyPackageViewModel report
...
         [Authorize]
         [RequiresAccountConfirmation("contact package owners")]
-        public virtual ActionResult ContactOwners(string id)
+        public virtual ActionResult ContactOwners(string id, string version)
         {
-            var package = _packageService.FindPackageByIdAndVersion(id, version: null);
-
-            if (package == null || package.PackageRegistration == null)
+            var package = _packageService.FindPackageByIdAndVersionStrict(id, version);
+            if (package == null)
```
scottbommarito: Any reason why you removed the `package.PackageRegistration` check?

It's probably redundant, but we do access `package.PackageRegistration.Owners` directly below, so if `package.PackageRegistration` is null, this will throw.

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -131,30 +131,31 @@ public void ReportMyPackage(ReportPackageRequest request)
             }
         }
 
-        public void SendContactOwnersMessage(MailAddress fromAddress, PackageRegistration packageRegistration, string message, string emailSettingsUrl, bool copySender)
+        public void SendContactOwnersMessage(MailAddress fromAddress, Package package, string message, string emailSettingsUrl, bool copySender)
         {
             string subject = "[{0}] Message for owners of the package '{1}'";
-            string body = @"_User {0} &lt;{1}&gt; sends the following message to the owners of Package '{2}'._
+            string body = @"_User {0} &lt;{1}&gt; sends the following message to the owners of Package '{2}' {3}._
```
scottbommarito: `Package` shouldn't be capitalized. Also, it's a little weird to have the ID in quotes but the version not.

E.g.
```
Package 'Newtonsoft.Json' 6.0.3
```

I would suggest either wrapping the entire thing in quotes.

```
...sends the following message to the owners of package '{2} {3}'.
```

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1209,16 +1209,17 @@ public void OnlyShowsOwnersWhoAllowReceivingEmails()
...
                                 new User { Username = "helpful2", EmailAllowed = true }
                             }
-                    }
+                    },
+                    Version = "1.0.0"
                 };
 
                 var packageService = new Mock<IPackageService>();
-                packageService.Setup(p => p.FindPackageByIdAndVersion("pkgid", null, null, true)).Returns(package);
+                packageService.Setup(p => p.FindPackageByIdAndVersionStrict("pkgid", "1.0.0")).Returns(package);
```
scottbommarito: I would suggest saving `pkgid` and `1.0.0` in variables so you don't need to declare them by hand every time you use them.

```
var id = "pkgid";
var version = "1.0.0";
...
packageService.Setup(p => p.FindPackageByIdAndVersionStrict(id, version)).Returns(package);
```

same with the tests below

src/NuGetGallery/Controllers/PackagesController.cs
dvkudr: fixed

src/NuGetGallery/Services/MessageService.cs
dvkudr: fixed

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
dvkudr: fixed & some new tests

dvkudr:
Hello!

skofman1:
Thanks for the PR @dvkudr ! We will merge this PR and deploy to nuget.org this sprint (the next 3 weeks).

scottbommarito:
@dvkudr , sorry for taking so long to take care of this. I've just merged this PR and it'll appear in our next deployment. Thanks again for the contribution!

dvkudr:
Great! Thank you guys!

Issue #5122, opened 2017-12-05
PackageStatusFacts should have [Fact]


joelverhagen:
https://github.com/NuGet/NuGetGallery/blob/master/tests/NuGetGallery.Core.Facts/PackageStatusFacts.cs

Nice catch, @skofman1.

PR #5123, opened 2017-12-05
Mark PackageStatusFacts facts as [Fact] and improve testing

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/5122

### DIFF
--- a/tests/NuGetGallery.Core.Facts/PackageStatusFacts.cs
+++ b/tests/NuGetGallery.Core.Facts/PackageStatusFacts.cs
@@ -1,20 +1,51 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using Xunit;
 
 namespace NuGetGallery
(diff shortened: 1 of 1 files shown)


tests/NuGetGallery.Core.Facts/PackageStatusFacts.cs
```diff
@@ -1,20 +1,50 @@
...
+                Assert.Equal(pair.Value, (int)pair.Key);
+            }
         }
 
-        public void AssertDeletedNotChanged()
+        [Fact]
+        public void AssertAllValuesAreImmutable()
         {
-            Assert.Equal(1, (int)PackageStatus.Deleted);
+            var expected = Enum
```
skofman1: are expected and actual reversed?

joelverhagen: Hmmm... at first I was thinking as we add values to the enum then we actually have to react by updating the `ImmutablePackageStatusValues`. Like in that case the expected set is actually the new values on the enum... but I am not sure...

joelverhagen: Fixed.

PR #5129, opened 2017-12-06
Permissions tests for DisplayPackage

scottbommarito:
This adds some permissions tests to `DisplayPackage`

More specifically, these test that
1 - Packages with pending edits show the pending edit only to users that own them.
2 - Packages in the validating/failed validation state only appear for users that own them.

Note that I caught a change in behavior from before--with the addition of the permissions service, admins could no longer see pending edits on package. This change returns that functionality (which will rewritten with the "Permissions Service 2.0" PR).

### DIFF
--- a/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
+++ b/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
@@ -474,28 +474,28 @@ namespace NuGetGallery
             [MemberData(nameof(EditedPackageOwners))]
             public Task GivenAnOwnedValidPackageWithNoEditsItDisplaysCurrentMetadata(User currentUser, User owner)
             {
-                return CheckValidPackageWithEdits(currentUser, owner, false, false);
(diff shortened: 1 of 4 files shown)


tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -393,57 +422,83 @@ public class TheDisplayPackageMethod
...
+                    };
+                }
+            }
 
-                indexingService.Setup(i => i.GetLastWriteTime()).Returns(Task.FromResult((DateTime?)DateTime.UtcNow));
+            [Theory]
+            [MemberData(nameof(EditedPackageOwners))]
+            public Task GivenAnOwnedValidPackageWithNoEditsItDisplaysCurrentMetadata(User currentUser, User owner)
+            {
+                return CheckValidPackageWithEdits(currentUser, owner, false, false);
```
skofman1: please use named params

scottbommarito: fixed

Issue #5048, opened 2017-11-21
Package page shouldn't show both "Listed" and "Validation Status" for package owners


skofman1:
This UI is confusing, because "Listed" can be "yes" and "Validation status" can be "Validating"

Suggestion: We should have a single column with values: "Validating/Valid/Invalid/Listed/Unlisted"

skofman1:
@anangaur , @karann-msft how do you think we can improve the UI here?

anangaur:
IMO, there should just one column - **Status** with the following values:
* Validating
* Listed
* Unlisted
* Failed validation

@karann-msft What is the user action when the package fails validation? Can the author resubmit the same package version?

anangaur:
@rido-min what happens when a package is uploaded with invalid signature (any of the reasons)? IMO that chack should be a synchronous check unlike the virus scanning check that may take some time.

rido-min:
@anangaur any package that does not satisfy the requirements should be rejected. The plan has been to do the check in the async pipeline, although the user perception should be almost instantaneous.

anangaur:
@rido-min Agree. So will the author see a rejection of the package then and there i.e. on the upload workflow? Without navigating to the package page? If yes then all is good :)
It would be weird that for package signing validation, author is navigated to the package page and then the validation fails thereby confusing the author whether to upload the same version again or a new version needs to be uploaded?

karann-msft:
where is the spec for this?

PR #5136, opened 2017-12-07
Status

skofman1:
Fix for: https://github.com/NuGet/NuGetGallery/issues/5048

Screenshot:
image


joelverhagen:
What does soft deleted look like? I think you may have an extra column.

skofman1:
There is no change of behavior for delete:
image

Verified before sending this PR

joelverhagen:
@skofman1, I think you have an extra column:
https://github.com/NuGet/NuGetGallery/blob/9076a6ff3de28f3d2f69b3ebe93627be6156979b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml#L463-L466

skofman1:
Fixed: https://github.com/NuGet/NuGetGallery/pull/5148

Issue #5120, opened 2017-12-05
Performance: Manage Packages view


chenriksson:
Opened to address server-side performance mentioned in PR #5060 

Potential optimizations:
1.  Packages.cshtml does a lot of UrlHelper.Action calls for Delete DeletePackage and ManagePackageOwners (for each version). With >1500 versions, this accounts for 65.8% inclusive (RouteCollection.GetVirtualPath is expensive).

Fix: Consider single virtual path lookup for DeletePackage, ManagePackageOwners, and Edit package action links then replace route values manually without route configuration lookup.

![image](https://user-images.githubusercontent.com/15364949/33627396-3ff56720-d9b2-11e7-9323-f92f3c3bcbca.png)

2. Packages.cshtml also has a lot of action links for package owner profiles. We should consider the same approach for these.

![image](https://user-images.githubusercontent.com/15364949/33629693-d6238e82-d9b9-11e7-9cee-511b519ec836.png)

chenriksson:
Branch: chenriks-perf-managepackages
Improvement: -32.8% (server) latency (warm)

Before:
- 6.41s (cold)
- 2.96s
- 2.90s
- 2.93s
- Avg (warm) = 2.93

After:
- 4.82s (cold)
- 1.94s
- 2.00s
- 1.98s
- Avg (warm) = 1.97s

chenriksson:
After rebase latest dev, and including Url.Package template.
Improvement: -60.2% (server) latency (warm)

Before:
* 1.63s
* 1.67s
* 1.68s
* Avg (warm) = 1.66s

After:
* 645ms
* 653ms
* 684ms
* Avg (warm) = 660.7ms

chenriksson:
The initial (cold) load of 'Manage Packages' has additional overhead due to EF reflection:

![image](https://user-images.githubusercontent.com/15364949/33694402-bdd7905a-daac-11e7-9c93-61ff3933a6ea.png)

xavierdecoster:
@chenriksson this is in deployed to dev state, is there a way you could verify and label it as such?

chenriksson:
Closing - will verify as part of #5173

PR #5138, opened 2017-12-07
Manage Packages: optimize package link generation

chenriksson:
Seeing up to 60% increase in server-side view performance. See #5120 for details.

### DIFF
--- a/src/NuGetGallery/App_Code/ViewHelpers.cshtml
+++ b/src/NuGetGallery/App_Code/ViewHelpers.cshtml
@@ -246,17 +246,6 @@
     </span>
 }
 
-@helper UserLinkWithIcon(RouteValueTemplate<User> urlTemplate, User user)
-{
-    <span class="ms-noWrap">
-        @if (user is Organization)
-        {
-            <i class="ms-Icon ms-Icon--Group" aria-hidden="true"></i>
-        }
-        <a href="@urlTemplate.Resolve(user)">@user.Username</a>
-    </span>
-}
-
 @helper WriteMeta(string name, string val) {
     if(!String.IsNullOrEmpty(val)) {
         <meta name="deployment-@name" content="@val" />
--- a/src/NuGetGallery/App_Code/ViewHelpers.cshtml
+++ b/src/NuGetGallery/App_Code/ViewHelpers.cshtml
@@ -246,6 +246,17 @@
     </span>
 }
 
+@helper UserLinkWithIcon(RouteValueTemplate<User> urlTemplate, User user)
+{
+    <span class="ms-noWrap">
+        @if (user is Organization)
+        {
+            <i class="ms-Icon ms-Icon--Group" aria-hidden="true"></i>
+        }
+        <a href="@urlTemplate.Resolve(user)">@user.Username</a>
+    </span>
+}
+
 @helper WriteMeta(string name, string val) {
     if(!String.IsNullOrEmpty(val)) {
         <meta name="deployment-@name" content="@val" />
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -246,6 +246,17 @@
     </span>
 }
 
+@helper UserLinkWithIcon(RouteValueTemplate<User> urlTemplate, User user)
```
chenriksson: Was used before PR #5060 to show org icon next to org owners. Will fix in morning to use again.

src/NuGetGallery/Helpers/RouteValueTemplate.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Web;
+using System.Web.Routing;
+
+namespace NuGetGallery.Helpers
+{
+    public class RouteValueTemplate<T>
```
chenriksson: Maybe RouteValuesTemplate is better name? Or ActionLinkTemplate?

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -3,12 +3,13 @@
 
 using NuGetGallery.Configuration;
```
chenriksson: nit: will sort

src/NuGetGallery/Helpers/RouteValueTemplate.cs
```diff
@@ -0,0 +1,55 @@
...
+        public RouteValueTemplate(Func<RouteValueDictionary, string> linkGenerator, IDictionary<string, Func<T, object>> routesGenerator)
+        {
+            RoutesGenerator = routesGenerator;
+
+            var templateValues = GetTemplateValues();
+            var encodedLinkTemplate = linkGenerator(templateValues);
+            LinkTemplate = HttpUtility.UrlDecode(encodedLinkTemplate);
+        }
+
+        private string LinkTemplate { get; set; }
```
skofman1: nit: the private properties can be made fields.

src/NuGetGallery/App_Code/ViewHelpers.cshtml
chenriksson: reverted, will do separately

src/NuGetGallery/Helpers/RouteValueTemplate.cs
chenriksson: Renamed to RouteUrlTemplate

src/NuGetGallery/Helpers/RouteValueTemplate.cs
chenriksson: done

src/NuGetGallery/UrlExtensions.cs
chenriksson: done

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -213,6 +214,25 @@ public static string PackageList(this UrlHelper url, bool relativeUrl = true)
...
+            this UrlHelper url,
+            bool relativeUrl = true)
+        {
+            var routesGenerator = new Dictionary<string, Func<IPackageVersionModel, object>>
+            {
+                { "id", p => p.Id },
+                { "version", p => null }
+            };
+
+            Func<RouteValueDictionary, string> linkGenerator = rv => GetRouteLink(
```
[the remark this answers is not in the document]
chenriksson: Added comments to clarify when to use `Url.PackageRegistrationTemplate` and when to use `Url.Package`.

For code reusability, there's a few ways to go here. Like you said, I could create a shared helper... or I could just always use the template to create the package links. The templated version reduces expensive GetVirtualPath calls, but has an overhead of more allocations. Right now I only use the template per-request, but I could cache it so it's used across requests and also by the Url.Package, etc APIs.

Opened issue #5162 to track... just want to think about it more, and do in separate PR.

PR #5139, opened 2017-12-07
Add Validation Issues to Display Package page

loic-sharma:
Adds validation issues to the display package page.

Example package that has failed with `Unknown` and `PackageIsSigned` issues:

screenshot

Depends on https://github.com/NuGet/NuGet.Jobs/pull/277


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -432,6 +433,13 @@ public virtual JsonResult UploadPackageProgress()
 
             var model = new DisplayPackageViewModel(package, packageHistory);
 
+            var validationIssues = await _validationService.GetLatestValidationIssuesAsync(package);
+
+            model.ValidationIssues = validationIssues
+                                        .Select(i => i.GetMessage())
```
xavierdecoster: nit: indentation seems a bit off?

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -33,5 +42,35 @@ public class ValidationService : IValidationService
...
+                    return validationSet
+                                .PackageValidations
+                                .SelectMany(v => v.PackageValidationIssues)
+                                .OrderBy(i => i.Key)
+                                .Select(i => ValidationIssue.Deserialize(i.IssueCode, i.Data))
+                                .ToList();
+                }
+            }
+
+            return new ValidationIssue[0];
```
xavierdecoster: Not sure if real benefit in our scenario, but `Enumerable.Empty<ValidationIssue>()` would cache the creation of the empty array which you currently recreate on each call to this method for a package that hasn't failed validation. Edge case, I know :)

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -163,6 +163,16 @@
                         validation failed. Please contact <a href="mailto:support@nuget.org">support@nuget.org</a>.
                     </text>
                 )
+
+                if (Model.ValidationIssues.Any())
+                {
```
xavierdecoster: There's no additional work after the foreach when this condition is met, so I think we can just avoid this check completely and remove the call to `.Any()`.

src/NuGetGallery/Controllers/PackagesController.cs
loic-sharma: Is it? It looks fine to me... would you rather have just one indent instead of a hanging intent from the `validationIssues` value?

src/NuGetGallery/Services/ValidationService.cs
loic-sharma: Sounds good, will fix

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -33,5 +42,35 @@ public class ValidationService : IValidationService
...
         }
+
+        public async Task<IEnumerable<ValidationIssue>> GetLatestValidationIssuesAsync(Package package)
+        {
+            // Only query the database for validation issues if the package has failed validation.
+            if (package.PackageStatusKey == PackageStatus.FailedValidation)
+            {
+                // Grab the most recent failed validation set for this package. There should always
+                // be a failed validation set for a package whose validation has failed.
+                var validationSet = await _validationContext
```
skofman1: I'm confused.. where are we filtering the ValidationSets to get only the ones that belong to the package?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -163,6 +163,16 @@
                         validation failed. Please contact <a href="mailto:support@nuget.org">support@nuget.org</a>.
                     </text>
                 )
+
+                if (Model.ValidationIssues.Any())
```
skofman1: screenshot?

src/NuGetGallery/Services/ValidationService.cs
loic-sharma: Oops, that's mistake 😖

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
loic-sharma: Added a screenshot to the main description.

src/NuGetGallery/Services/ValidationService.cs
loic-sharma: Fixed!

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
loic-sharma: Good catch. I used to only show the validation error message if there were no error messages, but I took that out. Fixed! :)

src/NuGetGallery/Services/ValidationService.cs
loic-sharma: Fixed.

loic-sharma:
🔔 Added tests and addressed feedback. Ready for review! :)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -432,6 +433,13 @@ public virtual JsonResult UploadPackageProgress()
 
             var model = new DisplayPackageViewModel(package, packageHistory);
 
+            var validationIssues = _validationService.GetLatestValidationIssues(package);
+
+            model.ValidationIssues = validationIssues
+                                        .Select(i => i.GetMessage())
+                                        .Distinct()
```
agr: Use case insensitive comparator here maybe? Or just dump messages into `HashSet<string>` instead of using explicit `Distinct`?

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -658,6 +660,67 @@ public class TheDisplayPackageMethod
...
+                packageService.Setup(p => p.FindPackageByIdAndVersion(It.Is<string>(s => s == "Foo"), It.Is<string>(s => s == null), It.Is<int>(i => i == SemVerLevelKey.SemVer2), It.Is<bool>(b => b == true)))
+                    .Returns(package);
+
+                indexingService.Setup(i => i.GetLastWriteTime()).Returns(Task.FromResult((DateTime?)DateTime.UtcNow));
+                
+                validationService.Setup(v => v.GetLatestValidationIssues(It.IsAny<Package>()))
+                    .Returns(new[]
+                    {
+                        new TestIssue("This should be deduplicated"),
+                        new TestIssue("This should be deduplicated"),
```
agr: Please, split values apart (put a teapot in the middle) to make sure deduplication works also when same values are not consecutive.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -658,6 +660,67 @@ public class TheDisplayPackageMethod
...
+                    });
+
+                // Act
+                var result = await controller.DisplayPackage("Foo", version: null);
+
+                // Assert
+                var model = ResultAssert.IsView<DisplayPackageViewModel>(result);
+
+                Assert.Equal(2, model.ValidationIssues.Count());
+                Assert.Equal("This should be deduplicated", model.ValidationIssues[0]);
```
agr: Is item order preservation required? Maybe use `Assert.Contains` here instead?

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -33,5 +41,36 @@ public class ValidationService : IValidationService
...
         }
+
+        public IEnumerable<ValidationIssue> GetLatestValidationIssues(Package package)
+        {
+            // Only query the database for validation issues if the package has failed validation.
+            if (package.PackageStatusKey == PackageStatus.FailedValidation)
+            {
+                // Grab the most recent failed validation set for this package. There should always
+                // be a failed validation set for a package whose validation has failed.
+                var validationSet = _validationSets
```
joelverhagen: This is broken in the revalidate case.

joelverhagen: We should take the latest validation set _with errors_.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -658,6 +660,67 @@ public class TheDisplayPackageMethod
...
+                    {
+                        Id = "Foo",
+                        Owners = new List<User>()
+                    },
+                    Version = "01.1.01",
+                    NormalizedVersion = "1.1.1",
+                    Title = "A test package!",
+                };
+
+                packageService.Setup(p => p.FindPackageByIdAndVersion(It.Is<string>(s => s == "Foo"), It.Is<string>(s => s == null), It.Is<int>(i => i == SemVerLevelKey.SemVer2), It.Is<bool>(b => b == true)))
```
joelverhagen: nit each param on its own line

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: `Distinct` uses a set.

I am hesitant to ignore case here since that is more likely hiding bugs...

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -33,5 +41,36 @@ public class ValidationService : IValidationService
         {
             await _initiator.StartValidationAsync(package);
         }
+
+        public IEnumerable<ValidationIssue> GetLatestValidationIssues(Package package)
+        {
+            // Only query the database for validation issues if the package has failed validation.
+            if (package.PackageStatusKey == PackageStatus.FailedValidation)
```
joelverhagen: I personally think this should be `== PackageStatus.FailedValidation || == PackageStatus.Validating` so they get the errors as soon as possible... but if your intuition says otherwise, feel free to won't fix.

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -33,5 +41,36 @@ public class ValidationService : IValidationService
...
+            {
+                // Grab the most recent failed validation set for this package. There should always
+                // be a failed validation set for a package whose validation has failed.
+                var validationSet = _validationSets
+                                        .GetAll()
+                                        .Where(s => s.PackageKey == package.Key)
+                                        .Where(s => s.PackageValidations.Any(v => v.ValidationStatus == ValidationStatus.Failed))
+                                        .Include(s => s.PackageValidations.Select(v => v.PackageValidationIssues))
+                                        .OrderByDescending(s => s.Updated)
+                                        .FirstOrDefault();
```
joelverhagen: Is this query using any indexes at all?

src/NuGetGallery/Controllers/PackagesController.cs
loic-sharma: I agree with Joel here.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
loic-sharma: Yes, item order preservation is an intended feature. We don't want customers seeing a different order of validation issues when they refresh the page. The `ValidationService` has an `OrderBy` to ensure this.

src/NuGetGallery/Services/ValidationService.cs
loic-sharma: We use `PackageValidations`'s `PackageKey` index. We are not using any indexes from `PackageValidation`. I would expect that a single `PackageKey` has a small number of `PackageValidations`/`PackageValidationSets` so I think this is fine. Thoughts?

src/NuGetGallery/Services/ValidationService.cs
loic-sharma: Hm. Taking the latest validation set with validation issues (or "errors") is also incorrect. What if the validation pipeline has a bug, a package is incorrectly validated as having validation issues, we fix the bug, revalidate the package, and now the package has no validation issues? We should show no validation issues.

I think this should be the latest validation set with ANY failed `PackageValidation` OR ALL successful `PackageValidation`. Thoughts? @agr

src/NuGetGallery/Services/ValidationService.cs
loic-sharma: Talked to @joelverhagen offline. The `PackageStatusKey` of `FailedValidation` and `Available` are "terminal" states by the validation pipeline that indicate that the validation is done. `Validating` is an intermediary state where the validation pipeline may be finding more validation issues. I will allow both `FailedValidation` and `Available`.

loic-sharma:
Addressed feedback and tested original PR on DEV. @agr: The ServerCommon dependencies should be updated to v2.6 when they are created.

src/NuGetGallery/Controllers/PackagesController.cs
agr: If it uses a set, then order might not be preserved, right? The [documentation](https://msdn.microsoft.com/en-us/library/bb348436(v=vs.110).aspx) mentions order only in the following sentence:
> The Distinct<TSource>(IEnumerable<TSource>) method returns an unordered sequence that contains no duplicate values.
It is unclear though if the overall order is kept.

agr:
Verified that it still works after the changes I made

src/NuGetGallery/Controllers/PackagesController.cs
loic-sharma: The current implementation of `Distinct` does maintain the enumerable's order, however, there are no documented guarantees about this. I think this is fine for our needs.

loic-sharma:
Validated on DEV. Blocked by https://github.com/NuGet/ServerCommon/pull/111.

src/NuGetGallery/Services/ValidationService.cs
loic-sharma: @joelverhagen: What do you think of my suggestion? It's the current implementation :)

loic-sharma:
Updated with latest Server Common dependencies. @joelverhagen could you sign off? :)

joelverhagen:
@loic-sharma :shipit:

PR #5141, opened 2017-12-07
Fix issue 5140

cristinamanum:
Do not pull in the Support page the support requests for the deleted accounts.

### DIFF
--- a/src/NuGetGallery/Areas/Admin/Controllers/SupportRequestController.cs
+++ b/src/NuGetGallery/Areas/Admin/Controllers/SupportRequestController.cs
@@ -231,7 +231,7 @@ namespace NuGetGallery.Areas.Admin.Controllers
 
             var skip = (pageNumber - 1) * take;
             var galleryUsername = GetLoggedInUser();
-            var issues = _supportRequestService.GetIssues(assignedTo, reason, issueStatusId, galleryUsername);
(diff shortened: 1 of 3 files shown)


joelverhagen:
Can we get a unit test? I see no reason why support requests shouldn't have at least some unit tests (maybe it already does?).

src/NuGetGallery/Areas/Admin/Controllers/SupportRequestController.cs
```diff
@@ -231,7 +231,7 @@ public ActionResult History(int id)
 
             var skip = (pageNumber - 1) * take;
             var galleryUsername = GetLoggedInUser();
-            var issues = _supportRequestService.GetIssues(assignedTo, reason, issueStatusId, galleryUsername);
+            var issues = _supportRequestService.GetIssues(assignedTo, reason, issueStatusId, galleryUsername).Where(i => i.CreatedBy != null);
```
chenriksson: If we always ignore / filter these out, why not delete a user's support requests when their account is deleted?

cristinamanum: The decision to keep these was for eventual future diagnostics.

cristinamanum: Added tests

PR #5146, opened 2017-12-07
ReportAbuse and ReportMyPackage permissions tests

scottbommarito:


tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1842,196 +1842,230 @@ public class TheListPackagesMethod
...
             {
-                var messageService = new Mock<IMessageService>();
-                messageService.Setup(
-                    s => s.ReportAbuse(It.Is<ReportPackageRequest>(r => r.Message == "Mordor took my finger")));
-                var package = new Package
+                get
                 {
-                    PackageRegistration = new PackageRegistration { Id = "mordor" },
-                    Version = "2.0.1"
+                    yield return new object[]
```
chenriksson: Maybe a helper would reduce the amount of code and improve readability?
```
yield return CurrentUserAndOwner(null, Owner)
yield return CurrentUserAndOwner(TestUtility.FakeUser, Owner)
yield return CurrentUserAndOwner(TestUtility.FakeAdminUser, Owner)
```

chenriksson: Also, if these user/owner combinations will be needed across multiple test classes it may be worth extracting into a helper class. Maybe in a future PR, if you come across the same pattern.

skofman1:
>             if (Request.IsAuthenticated)

looks like a confusing experience.. I filled all the details in report abuse form, and when I click 'submit' instead of submitting, I'm getting redirected to a different page..

---
Refers to: src/NuGetGallery/Controllers/PackagesController.cs:690 in 4544302. [](commit_id = 4544302d74de3133676153b0f35de174854b855e, deletion_comment = False)

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
scottbommarito: Not sure how useful the method is...this is a common confusion with XUnit theory tests. I would prefer keeping it like this because it's consistent with the rest of our XUnit testing.

Regarding extracting out the user/owner combinations, the problem with that is that they vary between the flows--e.g. "manage package owners" needs a different set of owners/non-owners than "edit"--so standardizing them would probably be more trouble than it's worth.

scottbommarito:
@skofman1 changed the behavior and updated the tests to accommodate them.

PR #5151, opened 2017-12-07
Permissions tests for Edit, Delete, Manage Package Owners, and SetLicenseReportVisibility (UI)

scottbommarito:
For Edit, SetLicenseReportVisibility, and Delete I parametrized the tests so that they now run with every test user.

For Manage Package Owners I added two simple tests to guarantee it displays the form when expected.

### DIFF
--- a/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
+++ b/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
@@ -1347,9 +1347,9 @@ namespace NuGetGallery
         public class TheDeleteMethod
             : TestContainer
         {
-            private static string _packageId = "CrestedGecko";
-            private static PackageRegistration _packageRegistration;
-            private static Package _package;
+            private string _packageId = "CrestedGecko";
+            private PackageRegistration _packageRegistration;
+            private Package _package;
 
             public TheDeleteMethod()
             {
@@ -1388,6 +1388,7 @@ namespace NuGetGallery
                 
                 Assert.IsType<HttpNotFoundResult>(result);
             }
+
             public static IEnumerable<object[]> NotOwner_Data
             {
                 get
@@ -2028,14 +2029,19 @@ namespace NuGetGallery
         public class TheManagePackageOwnersMethod
             : TestContainer
         {
-            private static string _packageId = "CrestedGecko";
-            private static string _packageVersion = "3.4.2";
+            private string _packageId = "CrestedGecko";
+            private string _packageVersion = "3.4.2";
 
-            private static Package _package = new Package
+            private Package _package;
+
+            public TheManagePackageOwnersMethod()
             {
-                PackageRegistration = new PackageRegistration { Id = _packageId },
-                Version = _packageVersion
-            };
+                _package = new Package
+                {
+                    PackageRegistration = new PackageRegistration { Id = _packageId },
+                    Version = _packageVersion
+                };
+            }
 
             public static IEnumerable<object[]> NotOwner_Data
             {
--- a/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
+++ b/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
@@ -3861,8 +3861,101 @@ namespace NuGetGallery
         public class TheSetLicenseReportVisibilityMethod
             : TestContainer
         {
-            [Fact]
-            public async Task IndexingAndPackageServicesAreUpdated()
+            public static IEnumerable<object[]> NotOwner_Data
+            {
+                get
+                {
+                    foreach (var visible in new[] { true, false })
+                    {
+                        yield return new object[]
+                        {
+                            null,
+                            TestUtility.FakeUser,
+                            visible
+                        };
+
+                        yield return new object[]
+                        {
+                            TestUtility.FakeUser,
+                            new User { Key = 5535 },
+                            visible
+                        };
+                    }
+                }
+            }
+
+            [Theory]
+            [MemberData(nameof(NotOwner_Data))]
+            public async Task Returns401IfNotOwner(User currentUser, User owner, bool visible)
+            {
+                // Arrange
+                var package = new Package
+                {
+                    PackageRegistration = new PackageRegistration { Id = "Foo" },
+                    Version = "1.0",
+                    Listed = true
+                };
+                package.PackageRegistration.Owners.Add(owner);
+
+                var packageService = new Mock<IPackageService>(MockBehavior.Strict);
+                packageService.Setup(svc => svc.FindPackageByIdAndVersionStrict("Foo", "1.0"))
+                    .Returns(package);
+
+                var controller = CreateController(
+                    GetConfigurationService(),
+                    packageService: packageService);
+                controller.SetCurrentUser(currentUser);
+                controller.Url = new UrlHelper(new RequestContext(), new RouteCollection());
+
+                // Act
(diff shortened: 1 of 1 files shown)


tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1674,10 +1842,23 @@ public class TheEditMethod
                 packageFileService.Verify(s => s.DownloadReadMeMdFileAsync(It.IsAny<Package>(), It.IsAny<bool>()), Times.Never);
             }
 
+            public static IEnumerable<object[]> OnPostBackWithReadMe_SavesPending_Data
+            {
+                get
+                {
+                    foreach (var ownerData in Owner_Data)
```
agr: Might look better in test explorer if rewritten as LINQ query.

agr: (well, probably all of those would have to be rewritten as expression bodied properties, probably not worth it)

scottbommarito: I didn't think it made much of a difference--I thought it just serialized all of the properties of the object regardless of how it was enumerated.

Either case, I'm going to leave this because as you said, it would have to become expression bodied properties.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1367,53 +1367,100 @@ public void DisplaysFullVersionStringAndUsesNormalizedVersionsInUrlsInSelectList
...
-                var packageDelete = new PackageDelete
+            [Fact]
+            public void Returns404IfPackageNotFound()
+            {
+                var controller = CreateController(GetConfigurationService());
+
+                var result = controller.Delete(_packageRegistration.Id, _package.Version);
+                
+                Assert.IsType<HttpNotFoundResult>(result);
+            }
```
chenriksson: nit: space between methods

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1347,18 +1347,18 @@ public void CallsSendContactOwnersMessageWithUserInfo()
         public class TheDeleteMethod
             : TestContainer
         {
-            [Fact]
-            public void DisplaysFullVersionStringAndUsesNormalizedVersionsInUrlsInSelectList()
+            private static string _packageId = "CrestedGecko";
+            private static PackageRegistration _packageRegistration;
+            private static Package _package;
```
chenriksson: I would beware of reusing the same instance of these test objects across tests... in case changes in state could potentially lead to intermittent test failures.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1367,53 +1367,100 @@ public void DisplaysFullVersionStringAndUsesNormalizedVersionsInUrlsInSelectList
...
+            {
+                get
                 {
-                    DeletedBy = user,
-                    DeletedByKey = user.Key,
-                    Packages = new[] { package },
-                    Reason = "Other",
-                    Signature = "John Doe"
-                };
+                    yield return new object[]
```
chenriksson: This has a similar pattern to other tests you added that take (currentUser, owner). Did we add a helper in that case to simplify the code?
```
yield return MemberDataHelpers.UserOnBehalfOfAccount(TestUtility.FakeUser, new User { Key = 5535 });
```

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1839,6 +2025,114 @@ public class TheListPackagesMethod
             }
         }
 
+        public class TheManagePackageOwnersMethod
+            : TestContainer
+        {
+            private static string _packageId = "CrestedGecko";
+            private static string _packageVersion = "3.4.2";
+
+            private static Package _package = new Package
```
chenriksson: Beware of reusing this instance across tests. Consider if the test cases run in parallel, or depend on state which changes.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
scottbommarito: Helper was added, but because this is a feature branch and I don't want to deal with any annoying rebasing, I was planning on integrating it with the tests that don't have it in a later PR.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
scottbommarito: I know some of the other tests had this, but they might not have been static. Will check.

scottbommarito: removed static

scottbommarito:
@chenriksson I fixed your comments!

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
chenriksson: Thanks - assuming new instance of test class is created for each test case? i.e., class variables won't be shared?

scottbommarito: I've seen some tests with this existing pattern (`ReportMyPackage` for example). Also, if there was any sort of concurrency issue here, I would expect my tests to have failed when I tested them.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -3567,8 +3867,101 @@ public void WillReturnCorrectResultForKnownUser()
...
+                            new User { Key = 5535 },
+                            visible
+                        };
+                    }
+                }
+            }
+
+            [Theory]
+            [MemberData(nameof(NotOwner_Data))]
+            public async Task Returns401IfNotOwner(User currentUser, User owner, bool visible)
```
joelverhagen: How is `visible` used?

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -3567,8 +3867,101 @@ public void WillReturnCorrectResultForKnownUser()
...
+                    GetConfigurationService(),
+                    packageService: packageService);
+                controller.SetCurrentUser(currentUser);
+                controller.Url = new UrlHelper(new RequestContext(), new RouteCollection());
+
+                // Act
+                var result = await controller.SetLicenseReportVisibility("Foo", "1.0", visible: false, urlFactory: (pkg, relativeUrl) => @"~\Bar.cshtml");
+
+                // Assert
+                Assert.IsType<HttpStatusCodeResult>(result);
```
joelverhagen: When there is a 401 assertion, we should also verify the dependency (e.g. DB) is not being changed. Applies to all tests here.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
scottbommarito: should be passed into `SetLicenseReportVisibility`--updated to do this

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
scottbommarito: Upon further investigation, in this particular test, it is already verified because the package service is a strict mock. I do need to analyze the rest of the tests however.

PR #5153, opened 2017-12-08
Permissions tests for API Key list and create actions

scottbommarito:


tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -623,6 +623,60 @@ public class TheConfirmAction : TestContainer
...
+            : TestContainer
+        {
+            [Fact]
+            public void CurrentUserIsInPackageOwners()
+            {
+                var currentUser = TestUtility.FakeUser;
+
+                var model = GetModelForApiKeys(currentUser);
+                
+                Assert.True(model.PackageOwners.Any(o => o.Owner == currentUser.Username && o.CanPushNew));
```
chenriksson: Shouldn't this test that PackageOwners contains a single entry for the current user? Maybe test name could be something like `PackageOwners_ContainsSingleCurrentUserEntryWithPushNew`?

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -623,6 +623,60 @@ public class TheConfirmAction : TestContainer
...
+            }
+
+            [Fact]
+            public void AdminIsInPackageOwners()
+            {
+                var currentUser = TestUtility.FakeAdminUser;
+
+                var model = GetModelForApiKeys(currentUser);
+
+                Assert.True(model.PackageOwners.All(o => o.Owner == currentUser.Username && o.CanPushNew));
```
chenriksson: Isn't this testing same as previous, but for site admin? Is *All* used instead of *Any* only because the FakeAdminUser isn't in an organization?

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -623,6 +623,60 @@ public class TheConfirmAction : TestContainer
...
+            [InlineData(false)]
+            public void OrganizationIsInPackageOwners(bool isAdmin)
+            {
+                var currentUser = isAdmin ? TestUtility.FakeOrganizationAdmin : TestUtility.FakeOrganizationCollaborator;
+                var organization = TestUtility.FakeOrganization;
+
+                var model = GetModelForApiKeys(currentUser);
+
+                Assert.True(model.PackageOwners.Any(o => o.Owner == currentUser.Username && o.CanPushNew));
+                Assert.True(model.PackageOwners.Any(o => o.Owner == organization.Username && o.CanPushNew == isAdmin));
```
chenriksson: The first assert is the same as the first test... current user should be the first entry. I could combine w/ that.

For the second assert, I would change it to test that PackageOwners contains a single entry for the Org account at index 2. Makes it a stronger test, and we'll catch if we have a bug with duplicate entries or not listing the current user first.

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -648,12 +702,32 @@ public class TheGenerateApiKeyAction : TestContainer
...
+
+            [Fact]
+            public Task WhenScopeOwnerDoesNotMatchAsAdmin_ReturnsBadRequest()
+            {
+                // Arrange 
+                var fakes = new Fakes();
+                var user = fakes.Admin;
+                var otherUser = fakes.User;
+
+                return WhenScopeOwnerDoesNotMatch_ReturnsBadRequest(user, otherUser);
```
chenriksson: `fakes.Admin` is site admin?

Can this be combined w/ previous test, and just take 2 Users as Theory args?

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
scottbommarito: yes and yes

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
scottbommarito: Yes, I'll combine with the others

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
scottbommarito: yes and yes

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -623,6 +623,100 @@ public class TheConfirmAction : TestContainer
...
+            public void OrganizationIsInPackageOwnersIfMember(bool isAdmin)
+            {
+                var currentUser = isAdmin ? TestUtility.FakeOrganizationAdmin : TestUtility.FakeOrganizationCollaborator;
+                var organization = TestUtility.FakeOrganization;
+
+                var model = GetModelForApiKeys(currentUser);
+                
+                Assert.Equal(1, model.PackageOwners.Count(o => o.Owner == organization.Username && o.CanPushNew == isAdmin));
+            }
+            public static IEnumerable<object[]> OrganizationIsNotInPackageOwnersIfNotMember_Data
```
chenriksson: nit: space between

scottbommarito:
@chenriksson Fixed the tests as requested.

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -623,6 +623,100 @@ public class TheConfirmAction : TestContainer
...
         }
 
+        public class TheApiKeysAction
+            : TestContainer
+        {
+            public static IEnumerable<object[]> CurrentUserIsInPackageOwnersWithPushNew_Data
+            {
+                get
+                {
+                    foreach (var currentUser in
```
chenriksson: Can this just be:
```
yield return new object[] { TestUtility.FakeUser };
yield return new object[] { TestUtility.FakeAdminUser };
```

chenriksson: If we have a lot of test code converting entities to object arrays, could consider a MemberDataHelper class:
```
public static class MemberDataHelper {
    object[] AsMemberData(this User user) { return new object[] { user }; }
}

public class TestClass {
    public static IEnumerable<object[]> TestData {
        get {
            yield return TestUtility.FakeUser.AsMemberData();
            yield return TestUtility.FakeAdminUser.AsMemberData();
        }
    }
    [Theory]
    [MemberData(nameof(TestData))]
    public void TestMethod(User user) { ... }
}
```

scottbommarito: How about

```
public static class MemberDataHelper {
    public static object[] AsData(params object[] objects)
    {
        return objects;
    }
}

public class TestClass {
    public static IEnumerable<object[]> TestData {
        get {
            yield return MemberDataHelper.AsData(TestUtility.FakeUser);
            yield return MemberDataHelper.AsData(TestUtility.FakeAdminUser);
        }
    }
    [Theory]
    [MemberData(nameof(TestData))]
    public void TestMethod(User user) { ... }
}
```

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -623,6 +623,95 @@ public class TheConfirmAction : TestContainer
...
+
+            [Theory]
+            [MemberData(nameof(OrganizationIsNotInPackageOwnersIfNotMember_Data))]
+            public void OrganizationIsNotInPackageOwnersIfNotMember(User currentUser)
+            {
+                var organization = TestUtility.FakeOrganization;
+
+                var model = GetModelForApiKeys(currentUser);
+
+                Assert.True(!model.PackageOwners.Any(o => o.Owner == organization.Username));
```
joelverhagen: Asserting count instead of `Any` would provide a better message.

PR #5156, opened 2017-12-08
Make ObfuscatedAction collection case insensitive

cristinamanum:
Address one comment missed in previous PR.

### DIFF
--- a/src/NuGetGallery/Telemetry/Obfuscator.cs
+++ b/src/NuGetGallery/Telemetry/Obfuscator.cs
@@ -15,7 +15,8 @@ namespace NuGetGallery
         /// </summary>
         internal const string DefaultTelemetryUserName = "ObfuscatedUserName";
 
-        internal static readonly HashSet<string> ObfuscatedActions = new HashSet<string>(StringComparer.InvariantCultureIgnoreCase){
+        internal static readonly HashSet<string> ObfuscatedActions = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        {
             "Packages/ConfirmPendingOwnershipRequest",
             "Packages/RejectPendingOwnershipRequest",
             "Packages/CancelPendingOwnershipRequest",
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Telemetry/Obfuscator.cs
```diff
@@ -15,7 +15,7 @@ internal static class Obfuscator
         /// </summary>
         internal const string DefaultTelemetryUserName = "ObfuscatedUserName";
 
-        internal static readonly HashSet<string> ObfuscatedActions = new HashSet<string>{
+        internal static readonly HashSet<string> ObfuscatedActions = new HashSet<string>(StringComparer.InvariantCultureIgnoreCase){
```
joelverhagen: nit: curly on next line

tests/NuGetGallery.Facts/Telemetry/ObfuscatorFacts.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace NuGetGallery.Telemetry
+{
+    public class ObfuscatorFacts
+    {
+        [Fact]
+        public void ObfuscatedActionsAreNotCaseSensitive()
```
joelverhagen: Use turkey "i" to assert the Invariant is used

src/NuGetGallery/Telemetry/Obfuscator.cs
joelverhagen: `OrdinalIgnoreCase` is supposed to be faster

tests/NuGetGallery.Facts/Telemetry/ObfuscatorFacts.cs
joelverhagen: This is a crazy edge case. You can ignore this if you feel like it's not worth testing.

src/NuGetGallery/Telemetry/Obfuscator.cs
cristinamanum: Updated

tests/NuGetGallery.Facts/Telemetry/ObfuscatorFacts.cs
cristinamanum: As we create the routes, I do not think that is worth adding.

PR #5158, opened 2017-12-09
PermissionsService 2.0 (just implementation)

scottbommarito:
This is the implementation of the new version of the Permissions Service, without changes to the existing code.

You may have seen this code before in https://github.com/NuGet/NuGetGallery/pull/5097

### DIFF
--- a/src/NuGetGallery/Helpers/PermissionsHelpers.cs
+++ b/src/NuGetGallery/Helpers/PermissionsHelpers.cs
@@ -118,8 +118,7 @@ namespace NuGetGallery
                 .OfType<Organization>()
                 .SelectMany(o => o.Members)
                 .Where(m => isUserMatch(m.Member))
-                .ToList()
-                .AsReadOnly();
+                .ToList();
 
             if (WouldSatisfy(PermissionsRequirement.OrganizationAdmin, permissionsRequirement) &&
                 matchingMembers.Any(m => m.IsAdmin))
--- a/src/NuGetGallery/Helpers/PermissionsHelpers.cs
+++ b/src/NuGetGallery/Helpers/PermissionsHelpers.cs
@@ -39,7 +39,7 @@ namespace NuGetGallery
         /// </summary>
         public static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, IPrincipal currentPrincipal, IEnumerable<User> entityOwners)
         {
-            if (PermissionLevelsIntersect(PermissionsRequirement.None, permissionsRequirement))
+            if (WouldSatisfy(PermissionsRequirement.None, permissionsRequirement))
             {
                 return true;
             }
@@ -85,7 +85,7 @@ namespace NuGetGallery
         /// </summary>
         public static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, User currentUser, IEnumerable<User> entityOwners)
         {
-            if (PermissionLevelsIntersect(PermissionsRequirement.None, permissionsRequirement))
+            if (WouldSatisfy(PermissionsRequirement.None, permissionsRequirement))
             {
                 return true;
             }
@@ -104,7 +104,7 @@ namespace NuGetGallery
 
         private static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, bool isUserAdmin, Func<User, bool> isUserMatch, IEnumerable<User> entityOwners)
         {
-            if (PermissionLevelsIntersect(PermissionsRequirement.None, permissionsRequirement))
+            if (WouldSatisfy(PermissionsRequirement.None, permissionsRequirement))
             {
                 return true;
             }
@@ -114,13 +114,13 @@ namespace NuGetGallery
                 return false;
             }
 
-            if (PermissionLevelsIntersect(PermissionsRequirement.Owner, permissionsRequirement) &&
+            if (WouldSatisfy(PermissionsRequirement.Owner, permissionsRequirement) &&
                 entityOwners.Any(isUserMatch))
             {
                 return true;
             }
 
-            if (PermissionLevelsIntersect(PermissionsRequirement.SiteAdmin, permissionsRequirement) &&
+            if (WouldSatisfy(PermissionsRequirement.SiteAdmin, permissionsRequirement) &&
                 isUserAdmin)
             {
                 return true;
@@ -132,13 +132,13 @@ namespace NuGetGallery
                 .Where(m => isUserMatch(m.Member))
                 .ToArray();
 
-            if (PermissionLevelsIntersect(PermissionsRequirement.OrganizationAdmin, permissionsRequirement) &&
+            if (WouldSatisfy(PermissionsRequirement.OrganizationAdmin, permissionsRequirement) &&
                 matchingMembers.Any(m => m.IsAdmin))
             {
                 return true;
             }
 
-            if (PermissionLevelsIntersect(PermissionsRequirement.OrganizationCollaborator, permissionsRequirement) &&
+            if (WouldSatisfy(PermissionsRequirement.OrganizationCollaborator, permissionsRequirement) &&
                 matchingMembers.Any())
             {
                 return true;
@@ -147,9 +147,12 @@ namespace NuGetGallery
             return false;
         }
 
-        private static bool PermissionLevelsIntersect(PermissionsRequirement first, PermissionsRequirement second)
+        /// <summary>
+        /// Returns true if and only if satisfying <paramref name="permissionsRequirementToCheck"/> would also satisfy <paramref name="permissionsRequirementToSatisfy"/>?
+        /// </summary>
+        private static bool WouldSatisfy(PermissionsRequirement permissionsRequirementToCheck, PermissionsRequirement permissionsRequirementToSatisfy)
         {
-            return (first & second) > 0;
+            return (permissionsRequirementToCheck & permissionsRequirementToSatisfy) > 0;
         }
     }
 }
--- a/src/NuGetGallery/Helpers/PermissionsHelpers.cs
+++ b/src/NuGetGallery/Helpers/PermissionsHelpers.cs
@@ -109,6 +109,12 @@ namespace NuGetGallery
                 return true;
             }
 
+            if (WouldSatisfy(PermissionsRequirement.SiteAdmin, permissionsRequirement) &&
+                isUserAdmin)
+            {
+                return true;
+            }
+
             if (entityOwners == null || !entityOwners.Any())
             {
                 return false;
@@ -116,12 +122,6 @@ namespace NuGetGallery
 
             if (WouldSatisfy(PermissionsRequirement.Owner, permissionsRequirement) &&
                 entityOwners.Any(isUserMatch))
-            {
-                return true;
-            }
-
-            if (WouldSatisfy(PermissionsRequirement.SiteAdmin, permissionsRequirement) &&
-                isUserAdmin)
             {
                 return true;
             }
@@ -139,7 +139,7 @@ namespace NuGetGallery
             }
 
             if (WouldSatisfy(PermissionsRequirement.OrganizationCollaborator, permissionsRequirement) &&
-                matchingMembers.Any())
+                matchingMembers.Any(m => !m.IsAdmin))
             {
                 return true;
             }
--- a/src/NuGetGallery/Helpers/PermissionsHelpers.cs
+++ b/src/NuGetGallery/Helpers/PermissionsHelpers.cs
@@ -37,16 +37,12 @@ namespace NuGetGallery
         /// <summary>
         /// Is <paramref name="currentPrincipal"/> allowed to perform an action with a requirement of <paramref name="permissionsRequirement"/> on the entity owned by <paramref name="entityOwners"/>?
         /// </summary>
-        public static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, IPrincipal currentPrincipal, IEnumerable<User> entityOwners)
+        public static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, IPrincipal currentPrincipal, ICollection<User> entityOwners)
         {
-            if (WouldSatisfy(PermissionsRequirement.None, permissionsRequirement))
-            {
-                return true;
-            }
-
             if (currentPrincipal == null)
             {
-                return false;
+                /// If the current principal is logged out, only <see cref="PermissionsRequirement.None"/> is satisfied.
+                return WouldSatisfy(PermissionsRequirement.None, permissionsRequirement);
             }
 
             return IsRequirementSatisfied(
@@ -83,16 +79,12 @@ namespace NuGetGallery
         /// <summary>
         /// Is <paramref name="currentPrincipal"/> allowed to perform an action with a requirement of <paramref name="permissionsRequirement"/> on the entity owned by <paramref name="entityOwners"/>?
         /// </summary>
-        public static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, User currentUser, IEnumerable<User> entityOwners)
+        public static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, User currentUser, ICollection<User> entityOwners)
         {
-            if (WouldSatisfy(PermissionsRequirement.None, permissionsRequirement))
-            {
-                return true;
-            }
-
             if (currentUser == null)
             {
-                return false;
+                /// If the current user is logged out, only <see cref="PermissionsRequirement.None"/> is satisfied.
+                return WouldSatisfy(PermissionsRequirement.None, permissionsRequirement);
             }
 
             return IsRequirementSatisfied(
@@ -102,22 +94,18 @@ namespace NuGetGallery
                 entityOwners);
         }
 
-        private static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, bool isUserAdmin, Func<User, bool> isUserMatch, IEnumerable<User> entityOwners)
+        private static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, bool isUserAdmin, Func<User, bool> isUserMatch, ICollection<User> entityOwners)
         {
-            if (WouldSatisfy(PermissionsRequirement.None, permissionsRequirement))
-            {
-                return true;
-            }
-
-            if (WouldSatisfy(PermissionsRequirement.SiteAdmin, permissionsRequirement) &&
-                isUserAdmin)
+            if (isUserAdmin && 
+                WouldSatisfy(PermissionsRequirement.SiteAdmin, permissionsRequirement))
             {
                 return true;
             }
 
             if (entityOwners == null || !entityOwners.Any())
             {
-                return false;
+                /// If there are no owners of the entity and the current user is not a site admin, only <see cref="PermissionsRequirement.None"/> is satisfied.
+                return WouldSatisfy(PermissionsRequirement.None, permissionsRequirement);
             }
 
             if (WouldSatisfy(PermissionsRequirement.Owner, permissionsRequirement) &&
@@ -130,7 +118,8 @@ namespace NuGetGallery
                 .OfType<Organization>()
                 .SelectMany(o => o.Members)
                 .Where(m => isUserMatch(m.Member))
-                .ToArray();
+                .ToList()
+                .AsReadOnly();
 
             if (WouldSatisfy(PermissionsRequirement.OrganizationAdmin, permissionsRequirement) &&
                 matchingMembers.Any(m => m.IsAdmin))
@@ -144,7 +133,8 @@ namespace NuGetGallery
                 return true;
             }
 
-            return false;
+            /// If the current user is not related to the entity in any way and is not a site admin, only <see cref="PermissionsRequirement.None"/> is satisfied.
+            return WouldSatisfy(PermissionsRequirement.None, permissionsRequirement);
         }
 
         /// <summary>
@@ -152,7 +142,7 @@ namespace NuGetGallery
         /// </summary>
         private static bool WouldSatisfy(PermissionsRequirement permissionsRequirementToCheck, PermissionsRequirement permissionsRequirementToSatisfy)
         {
-            return (permissionsRequirementToCheck & permissionsRequirementToSatisfy) > 0;
+            return (permissionsRequirementToCheck & permissionsRequirementToSatisfy) != PermissionsRequirement.Unsatisfiable;
         }
     }
 }
--- a/src/NuGetGallery/Helpers/PermissionsHelpers.cs
+++ b/src/NuGetGallery/Helpers/PermissionsHelpers.cs
@@ -8,7 +8,7 @@ using System.Security.Principal;
 
 namespace NuGetGallery
 {
-    public class PermissionsHelpers
+    public static class PermissionsHelpers
     {
         /// <summary>
         /// Is <paramref name="currentPrincipal"/> allowed to perform an action with a requirement of <paramref name="permissionsRequirement"/> on <paramref name="packageRegistration"/>?
@@ -39,9 +39,14 @@ namespace NuGetGallery
         /// </summary>
         public static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, IPrincipal currentPrincipal, IEnumerable<User> entityOwners)
         {
+            if (PermissionLevelsIntersect(PermissionsRequirement.None, permissionsRequirement))
+            {
+                return true;
+            }
+
             if (currentPrincipal == null)
             {
-                return PermissionLevelsIntersect(PermissionsRequirement.None, permissionsRequirement);
+                return false;
             }
 
             return IsRequirementSatisfied(
@@ -80,9 +85,14 @@ namespace NuGetGallery
         /// </summary>
         public static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, User currentUser, IEnumerable<User> entityOwners)
         {
+            if (PermissionLevelsIntersect(PermissionsRequirement.None, permissionsRequirement))
+            {
+                return true;
+            }
+
             if (currentUser == null)
             {
-                return PermissionLevelsIntersect(PermissionsRequirement.None, permissionsRequirement);
+                return false;
             }
 
             return IsRequirementSatisfied(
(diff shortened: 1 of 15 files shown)


src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,155 @@
...
+using System.Security.Principal;
+
+namespace NuGetGallery
+{
+    public static class PermissionsHelpers
+    {
+        /// <summary>
+        /// Is <paramref name="currentPrincipal"/> allowed to perform an action with a requirement of <paramref name="permissionsRequirement"/> on <paramref name="packageRegistration"/>?
+        /// </summary>
+        public static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, IPrincipal currentPrincipal, PackageRegistration packageRegistration)
```
scottbommarito: This file is mostly identical to the existent `PermissionsService` except:
1 - `IsActionAllowed` is now `IsRequirementSatisfied`
2 - Added a method for `ReservedNamespace`s
3 - Made some performance checks as suggested by @cristinamanum (specifically, we now check if having a specific kind of ownership would satisfy the requirement before checking if we have that ownership)

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
```diff
@@ -0,0 +1,75 @@
...
+                PermissionsRequirement.Owner | PermissionsRequirement.OrganizationAdmin,
+                PermissionsRequirement.Owner);
+
+        /// <summary>
+        /// The action of uploading a new version of an existing package ID.
+        /// </summary>
+        public static ActionRequiringPackagePermissions UploadNewPackageVersion =
+            new ActionRequiringPackagePermissions(
+                PermissionsRequirement.Owner | PermissionsRequirement.OrganizationAdmin | PermissionsRequirement.OrganizationCollaborator,
+                PermissionsRequirement.Owner | PermissionsRequirement.SiteAdmin);
```
joelverhagen: >PermissionsRequirement.SiteAdmin [](start = 47, length = 32)

This one is wrong. Admins shouldn't be able to upload new versions.

src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
```diff
@@ -0,0 +1,77 @@
...
+            PackageId = packageId;
+            ReservedNamespaceService = reservedNamespaceService;
+        }
+    }
+
+    /// <summary>
+    /// An action requiring permissions on a <see cref="ReservedNamespace"/> that can be done on behalf of another <see cref="User"/>.
+    /// </summary>
+    public class ActionRequiringReservedNamespacePermissions
+        : ActionRequiringEntityPermissions<IEnumerable<ReservedNamespace>>, IActionRequiringEntityPermissions<ActionOnNewPackageContext>
```
joelverhagen: >IEnumerable<ReservedNamespace> [](start = 43, length = 30)

Might be worth leaving a comment about why this is multiple namespaces, i.e. multiple namespaces could apply to a single ID.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// </summary>
+        /// <param name="accountsAllowedOnBehalfOf">A <see cref="IEnumerable{User}"/> containing all accounts that <paramref name="currentUser"/> can perform this action on <paramref name="entity"/> on behalf of.</param>
+        /// <returns>True if and only if <paramref name="currentPrincipal"/> can perform this action on <paramref name="entity"/> on behalf of any <see cref="User"/>.</returns>
+        public bool TryGetAccountsIsAllowedOnBehalfOf(User currentUser, TEntity entity, out IEnumerable<User> accountsAllowedOnBehalfOf)
+        {
+            accountsAllowedOnBehalfOf = Enumerable.Empty<User>();
+
+            var possibleAccountsOnBehalfOf = 
+                new[] { currentUser }
+                    .Union(GetOwners(entity));
```
joelverhagen: >Union [](start = 21, length = 5)

Are set semantics necessary here? Are we sure comparison of user instances is working? Concat may be simpler.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,85 @@
...
+                    possibleAccountsOnBehalfOf
+                        .Union(currentUser.Organizations.Select(o => o.Organization));
+            }
+
+            foreach (var accountOnBehalfOf in possibleAccountsOnBehalfOf)
+            {
+                var failure = IsAllowed(currentUser, accountOnBehalfOf, entity);
+                if (failure == PermissionsCheckResult.Allowed)
+                {
+                    accountsAllowedOnBehalfOf = accountsAllowedOnBehalfOf.Concat(new[] { accountOnBehalfOf });
```
joelverhagen: >Concat [](start = 74, length = 6)

Use `List`, don't need LINQ for everything

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            return false;
+        }
+
+        /// <summary>
+        /// Returns true if and only if satisfying <paramref name="permissionsRequirementToCheck"/> would also satisfy <paramref name="permissionsRequirementToSatisfy"/>?
+        /// </summary>
+        private static bool WouldSatisfy(PermissionsRequirement permissionsRequirementToCheck, PermissionsRequirement permissionsRequirementToSatisfy)
+        {
+            return (permissionsRequirementToCheck & permissionsRequirementToSatisfy) > 0;
```
joelverhagen: >> [](start = 85, length = 1)

`!=` unless this is an unsigned flags or if you have tests asserting never negative flag values

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,158 @@
...
+        {
+            if (WouldSatisfy(PermissionsRequirement.None, permissionsRequirement))
+            {
+                return true;
+            }
+
+            if (entityOwners == null || !entityOwners.Any())
+            {
+                return false;
+            }
```
joelverhagen: This breaks admin permissions (on things they would otherwise do)

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,158 @@
...
+                isUserAdmin)
+            {
+                return true;
+            }
+
+            var matchingMembers = entityOwners
+                .OfType<Organization>()
+                .SelectMany(o => o.Members)
+                .Where(m => isUserMatch(m.Member))
+                .ToArray();
```
joelverhagen: >ToArray [](start = 17, length = 7)

ToList requires fewer allocations and is a better default for these one-off lists

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,158 @@
...
+                return true;
+            }
+
+            if (WouldSatisfy(PermissionsRequirement.OrganizationCollaborator, permissionsRequirement) &&
+                matchingMembers.Any())
+            {
+                return true;
+            }
+
+            return false;
```
joelverhagen: >false [](start = 19, length = 5)

This is such a generic code path, I have a hard time thinking of cases that would fall through here. Might be worth commenting about that.

src/NuGetGallery/Helpers/PermissionsHelpers.cs
scottbommarito: `==` and `!=` work with `Flags`?

I'm using the int logic here because the second parameter is a combination of multiple enum values, so I want to compare that they just have at least one value in common.

src/NuGetGallery/Helpers/PermissionsHelpers.cs
scottbommarito: fixed--it was not only a logic issue but a tests issues

src/NuGetGallery/Helpers/PermissionsHelpers.cs
scottbommarito: i did some restructuring that should make this clearer to understand

src/NuGetGallery/Helpers/PermissionsHelpers.cs
scottbommarito: done

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: makes sense

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: I wanted to get rid of duplicates. I will replace this with `Concat` and `Distinct`

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        private static bool IsRequirementSatisfied(PermissionsRequirement permissionsRequirement, bool isUserAdmin, Func<User, bool> isUserMatch, IEnumerable<User> entityOwners)
+        {
+            if (WouldSatisfy(PermissionsRequirement.None, permissionsRequirement))
+            {
+                return true;
+            }
+
+            if (WouldSatisfy(PermissionsRequirement.SiteAdmin, permissionsRequirement) &&
+                isUserAdmin)
```
skofman1: perf nit: always put the cheaper condition first so the 'and' operation will fail if it's false

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
joelverhagen: Do you really care about duplicates here? Like most of the time it will be unique right? Distinct uses default comparer which is likely reference equals (or did we implement `GetHashCode` and `Equals`?)

src/NuGetGallery/Helpers/PermissionsHelpers.cs
joelverhagen: `[Flags]` is just an attribute. The important details is that the `enum` is an `int` (unless you changed that...)

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// Determines whether <paramref name="currentPrincipal"/> can perform this action on <paramref name="entity"/> on behalf of any <see cref="User"/>.
+        /// </summary>
+        /// <param name="accountsAllowedOnBehalfOf">A <see cref="IEnumerable{User}"/> containing all accounts that <paramref name="currentUser"/> can perform this action on <paramref name="entity"/> on behalf of.</param>
+        /// <returns>True if and only if <paramref name="currentPrincipal"/> can perform this action on <paramref name="entity"/> on behalf of any <see cref="User"/>.</returns>
+        public bool TryGetAccountsIsAllowedOnBehalfOf(User currentUser, TEntity entity, out IEnumerable<User> accountsAllowedOnBehalfOf)
+        {
+            accountsAllowedOnBehalfOf = Enumerable.Empty<User>();
+
+            var possibleAccountsOnBehalfOf = 
+                new[] { currentUser }
```
cristinamanum: Is it ok to add the currentUser if the currentUser == null ?

src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        protected override PermissionsCheckResult IsAllowedOnEntity(User account, IEnumerable<ReservedNamespace> reservedNamespaces)
+        {
+            if (!reservedNamespaces.Any())
+            {
+                return PermissionsCheckResult.Allowed;
+            }
+
+            return reservedNamespaces.Any(rn => PermissionsHelpers.IsRequirementSatisfied(ReservedNamespacePermissionsRequirement, account, rn)) ?
+                PermissionsCheckResult.Allowed : PermissionsCheckResult.ReservedNamespaceFailure;
```
cristinamanum: Why is it fine to have Any? Should not be permissions allowed for All the namespaces?

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: yes, the values calculated here will be piped directly to the UI for upload and API key scenarios.

src/NuGetGallery/Helpers/PermissionsHelpers.cs
scottbommarito: No what I mean is we'll be comparing 0b0010 and 0b1010 and expect it to return true. I didn't think `enum` or `Flags` made this comparison return true.

src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
scottbommarito: You only need one--see https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Services/ReservedNamespaceService.cs#L381

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: yup, it's necessary--if the current user is logged out but still able to perform the action, we still want to return true

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
scottbommarito: fixed

src/NuGetGallery/Helpers/PermissionsHelpers.cs
scottbommarito: nice catch

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                return true;
+            }
+
+            if (entityOwners == null || !entityOwners.Any())
```
agr: There is a multiple enumeration of a `IEnumerable` (between here and lines 124 and 129). Since it is unlikely could be fixed, please use something else (`ICollection` at least).

scottbommarito: switched all methods in this class to use `ICollection`

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public ActionRequiringEntityPermissions(PermissionsRequirement accountOnBehalfOfPermissionsRequirement)
+        {
+            AccountOnBehalfOfPermissionsRequirement = accountOnBehalfOfPermissionsRequirement;
+        }
+
+        /// <summary>
+        /// Determines whether <paramref name="currentUser"/> can perform this action on <paramref name="entity"/> on behalf of <paramref name="account"/>.
+        /// </summary>
+        public PermissionsCheckResult IsAllowed(User currentUser, User account, TEntity entity)
```
agr: If it's not returning `bool` it shouldn't be called `IsAllowed`, same for overload below

src/NuGetGallery/Services/ActionRequiringPackagePermissions.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public ActionRequiringPackagePermissions(
+            PermissionsRequirement accountOnBehalfOfPermissionsRequirement,
+            PermissionsRequirement packageRegistrationPermissionsRequirement)
+            : base(accountOnBehalfOfPermissionsRequirement)
+        {
+            PackageRegistrationPermissionsRequirement = packageRegistrationPermissionsRequirement;
+        }
+
+        public PermissionsCheckResult IsAllowed(User currentUser, User account, Package package)
```
agr: Same comment as above: method name should be different.

src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public PermissionsCheckResult IsAllowed(IPrincipal currentPrincipal, User account, ActionOnNewPackageContext newPackageContext)
+        {
+            return IsAllowed(currentPrincipal, account, GetReservedNamespaces(newPackageContext));
+        }
+
+        protected override PermissionsCheckResult IsAllowedOnEntity(User account, IEnumerable<ReservedNamespace> reservedNamespaces)
+        {
+            if (!reservedNamespaces.Any())
```
agr: Multiple enumeration of `IEnumerable`

src/NuGetGallery/Services/PermissionsRequirement.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        /// <summary>
+        /// No user can satisfy the requirement.
+        /// </summary>
+        Unsatisfiable = 0,
+
+        /// <summary>
+        /// No requirement--any user can satisfy the requirement.
+        /// </summary>
+        None = 1,
```
agr: Replace values with `1 << 0`, `1 << 1`, `1 << 2`, etc. maybe to emphasize bit masking?

tests/NuGetGallery.Facts/Services/PermissionsHelpersFacts.cs
```diff
@@ -0,0 +1,138 @@
...
+                        {
+                            _stateValues.Where(s => Includes(i, s)).ToList()
+                        };
+                    }
+                }
+            }
+
+            private static bool Includes(int i, ReturnsSatisfiedRequirementWhenExpected_State state)
+            {
+                return (i & (int)state) > 0;
```
agr: Should be `!=` instead of `>`, strictly speaking, but wouldn't be an issue unless we have more stuff in the `ReturnsSatisfiedRequirementWhenExpected_State`

src/NuGetGallery/Helpers/PermissionsHelpers.cs
scottbommarito: Oh I see now

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,158 @@
...
+        {
+            if (WouldSatisfy(PermissionsRequirement.None, permissionsRequirement))
+            {
+                return true;
+            }
+
+            if (currentUser == null)
+            {
+                return false;
+            }
```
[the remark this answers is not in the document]
scottbommarito: done

src/NuGetGallery/Services/ActionRequiringPackagePermissions.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Linq;
+using System.Security.Principal;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// An action requiring permissions on a <see cref="PackageRegistration"/> or <see cref="Package"/> that can be done on behalf of another <see cref="User"/>.
+    /// </summary>
+    public class ActionRequiringPackagePermissions
+        : ActionRequiringEntityPermissions<PackageRegistration>, IActionRequiringEntityPermissions<Package>
```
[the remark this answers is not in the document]
scottbommarito: base class extends the interface for `PackageRegistration` and provides the actual implementation

by extending the interface for `Package` here we allow users to pass in `Package` instead of `PackageRegistration` for convenience

src/NuGetGallery/Services/PermissionsRequirement.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <summary>
+    /// Represents a permissions restriction applied to a <see cref="User"/> who is trying to do an operation <see cref="User"/> has with a resource such as a <see cref="Package"/> or account (e.g. another <see cref="User"/>).
+    /// </summary>
+    [Flags]
+    public enum PermissionsRequirement
+    {
+        /// <summary>
+        /// No user can satisfy the requirement.
+        /// </summary>
+        Unsatisfiable = 0,
```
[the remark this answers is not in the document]
scottbommarito: exists for syntactical cleanliness

src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    public class ActionOnNewPackageContext
+    {
+        public string PackageId { get; }
+        public IReservedNamespaceService ReservedNamespaceService { get; }
+
+        public ActionOnNewPackageContext(string packageId, IReservedNamespaceService reservedNamespaceService)
+        {
+            PackageId = packageId;
+            ReservedNamespaceService = reservedNamespaceService;
```
skofman1: null check

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
```diff
@@ -0,0 +1,75 @@
...
+    public static class ActionsRequiringPermissions
+    {
+        /// <summary>
+        /// The action of seeing private metadata about a package.
+        /// For example, if a package is validating, only users who can perform this action can see the metadata of the package.
+        /// </summary>
+        public static ActionRequiringPackagePermissions DisplayPrivatePackageMetadata =
+            new ActionRequiringPackagePermissions(
+                PermissionsRequirement.Owner | PermissionsRequirement.OrganizationAdmin | PermissionsRequirement.OrganizationCollaborator,
+                PermissionsRequirement.Owner | PermissionsRequirement.SiteAdmin);
```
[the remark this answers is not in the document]
scottbommarito: I like both of these--will do both

src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
scottbommarito: good catch

src/NuGetGallery/Services/PermissionsRequirement.cs
scottbommarito: done

src/NuGetGallery/Services/ActionRequiringPackagePermissions.cs
scottbommarito: renamed to `CheckPermissions`

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: renamed to `CheckPermissions`

src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
scottbommarito: done

scottbommarito:
@joelverhagen @chenriksson @skofman1 @agr 

All feedback was addressed.

src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Principal;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Context object for checking permissions of an action involving a new package ID.
+    /// </summary>
+    public class ActionOnNewPackageContext
```
joelverhagen: one class per file. makes github search by file much easier

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        /// <summary>
+        /// Determines whether <paramref name="currentPrincipal"/> can perform this action on <paramref name="entity"/> on behalf of any <see cref="User"/>.
+        /// </summary>
+        /// <param name="accountsAllowedOnBehalfOf">A <see cref="IEnumerable{User}"/> containing all accounts that <paramref name="currentUser"/> can perform this action on <paramref name="entity"/> on behalf of.</param>
+        /// <returns>True if and only if <paramref name="currentPrincipal"/> can perform this action on <paramref name="entity"/> on behalf of any <see cref="User"/>.</returns>
+        public bool TryGetAccountsIsAllowedOnBehalfOf(User currentUser, TEntity entity, out IEnumerable<User> accountsAllowedOnBehalfOf)
+        {
+            var accountsAllowedOnBehalfOfList = new List<User>();
+            accountsAllowedOnBehalfOf = accountsAllowedOnBehalfOfList;
```
chenriksson: why do you need the local variable?

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,102 @@
...
+                    possibleAccountsOnBehalfOf
+                        .Concat(currentUser.Organizations.Select(o => o.Organization));
+            }
+
+            possibleAccountsOnBehalfOf = possibleAccountsOnBehalfOf.Distinct(new UserEqualityComparer());
+
+            foreach (var accountOnBehalfOf in possibleAccountsOnBehalfOf)
+            {
+                var failure = CheckPermissions(currentUser, accountOnBehalfOf, entity);
+                if (failure == PermissionsCheckResult.Allowed)
```
chenriksson: nit: failure sounds odd; maybe permissionsResult?

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,102 @@
...
+                    .Concat(GetOwners(entity));
+
+            if (currentUser != null)
+            {
+                possibleAccountsOnBehalfOf = 
+                    possibleAccountsOnBehalfOf
+                        .Concat(currentUser.Organizations.Select(o => o.Organization));
+            }
+
+            possibleAccountsOnBehalfOf = possibleAccountsOnBehalfOf.Distinct(new UserEqualityComparer());
```
chenriksson: Why do we need Distinct here?

The membership table has PK of (OrganizationKey, MemberKey), so I don't think a User can have multiple memberships in the same Organization.

src/NuGetGallery/Services/IActionRequiringEntityPermissions.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Linq;
+using System.Security.Principal;
+
+namespace NuGetGallery
+{
+    public interface IActionRequiringEntityPermissions<TEntity>
+    {
+        PermissionsCheckResult CheckPermissions(User currentUser, User account, TEntity entity);
+        PermissionsCheckResult CheckPermissions(IPrincipal currentPrincipal, User account, TEntity entity);
+        bool TryGetAccountsIsAllowedOnBehalfOf(User currentUser, TEntity entity, out IEnumerable<User> accountsAllowedOnBehalfOf);
```
chenriksson: Should this be `TryGetAccountsAllowedOnBehalfOf`?

chenriksson:
@scottbommarito My only remaining feedback is with the distinct call... please look at that, then I'm good.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: This is a distinct on not only the user's organizations, but also the owners of the entity we are checking permissions of (e.g. the owners of the package). Therefore, there can be duplicates.

src/NuGetGallery/Helpers/PermissionsHelpers.cs
```diff
@@ -0,0 +1,148 @@
...
+            {
+                return true;
+            }
+
+            var matchingMembers = entityOwners
+                .OfType<Organization>()
+                .SelectMany(o => o.Members)
+                .Where(m => isUserMatch(m.Member))
+                .ToList()
+                .AsReadOnly();
```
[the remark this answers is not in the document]
scottbommarito: I was thinking it might be cleaner if we guaranteed this was read-only but I suppose it doesn't really matter.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: the `out` parameter is an `IEnumerable<User>` and not a `List<User>`, so I need to either store this local variable or cast the out parameter.

I'll switch to casting the `out` parameter instead of having another local variable, it looks cleaner.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: the enum used to be named `PermissionsFailure` and was renamed `PermissionsCheckResult` but this variable was never updated. It is now updated to `result`.

Issue #4604, opened 2017-08-31
Creating an API key for an account with lots of packages freezes the UI


pranavkm:
Our account owns about 1400 packages. Creating a new API key for this API freezes up the UI for a significant period. It looks like something's attempting to sort items and that's locking up the UI:

![api](https://user-images.githubusercontent.com/174281/29939111-a627da96-8e3f-11e7-9ac4-1cf8881879ec.gif)

joelverhagen:
Very helpful gif! We'll take a look. Thanks for letting us know.

chenriksson:
I think I managed to repro running Gallery against DEV and using test account which has a lot of packages (>8K registrations when I tested). Here's what I saw on clicking 'Create' (ajax success callback):
- 17s to reach ApiKeyViewModel.CancelEdit (page-api-keys.js)
- 7s more to reach window.nuget.resetFormValidation (common.js)
- hangs at validator.showErrors (common.js)
- In defaultShowErrors, hangs on for loop to showLabel for successList
  - successList had >8K entries, roughly same as # package registrations

chenriksson:
The primary cause for the hang is resolved, so marking code complete. There is still more we can do, but I can open separate issues.

chenriksson:
Verifying with test account, based on F12 JS profile

| Scenario | Baseline Env | Baseline Result | Env | Result | %Diff |
| --- | --- | --- | --- | --- | --- |
| GenerateApiKey-success | PROD0 (2460 pkgs) | 156928.4ms | INT0 (2980 pkgs) | 4019.5ms | -97.4% |

chenriksson:
@pranavkm The UI should be more responsive now after creation of an API key.

PR #5159, opened 2017-12-09
[Perf][ApiKeys]Fix hang due to resetFormValidation including all fields

chenriksson:
Issue #4604 - Fixes client side hang during ApiKey creation callback due to resetFormValidation including all package checkboxes. Tested with test account on DEV, which had >8K package registrations.

Note, there are more perf issues on this page - may open separate issues when I have more time.

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/common.js
+++ b/src/NuGetGallery/Scripts/gallery/common.js
@@ -210,8 +210,10 @@
 
     nuget.resetFormValidation = function (formElement) {
         var validator = $(formElement).validate();
-        $(formElement).find("*[name]").each(function () {
-            validator.successList.push(this);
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -210,8 +210,10 @@
 
     nuget.resetFormValidation = function (formElement) {
         var validator = $(formElement).validate();
-        $(formElement).find("*[name]").each(function () {
-            validator.successList.push(this);
+        $(formElement).find("*[name][data-val]").each(function () {
```
agr: `[data-val=true]` should work instead of explicit filtering in the callback.

PR #5164, opened 2017-12-11
Clean azure storage audit data.

cristinamanum:
Only the packages audit records will be saved in Azure Storage.
The packages audit data will have removed all the user specific data.


src/NuGetGallery.Core/Auditing/Obfuscator.cs
```diff
@@ -9,6 +9,8 @@ namespace NuGetGallery.Auditing
 {
     public static class Obfuscator
     {
+        internal const string ObfuscatedUserName = "ObfuscatedUserName";
```
scottbommarito: I might have missed a discussion, but why not use some sort of hash on the username instead of setting all usernames to the same value?

src/NuGetGallery.Core/Auditing/PackageAuditRecord.cs
```diff
@@ -71,5 +71,62 @@ public override string GetPath()
...
+
+        public override AuditRecord Obfuscate()
+        {
+            var obfuscatedPackage = CreateObfuscatedPackage();
+            var obfuscatedAuditedPackage = AuditedPackage.CreateFrom(obfuscatedPackage);
+
+            return new PackageAuditRecord(Id, Version, Hash, obfuscatedAuditedPackage, RegistrationRecord, Action, Reason );
+        }
+
+        private Package CreateObfuscatedPackage()
```
scottbommarito: A couple comments here:

1 - `obfuscatedPacakge` -> `obfuscatedPackage`
2 - This is impossible to maintain--if any fields in `Package` are added, this will immediately break in an invisible way.
3 - Is there a way we could rewrite this so that only fields that are changed appear here? At the bare minimum, we could move this into a clone method on `Package` which could potentially be useful outside of this flow and then obfuscate just the fields we care about in this method.
4 - Could we use what I suggest above (property setters) to force obfuscation in such a way that it's no longer necessary to have an `Obfuscate` method?

tests/NuGetGallery.Core.Facts/Auditing/CloudAuditingServiceTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            // Act 
+            var auditActorToBePersisted = await service.GetCloudAuditActorAsync();
+
+            // Assert
+            Assert.Equal<string>("ObfuscatedUserName", auditActorToBePersisted.UserName);
+            Assert.Equal(true, IsObfuscatedIp(auditActorToBePersisted.MachineIP));
+        }
+
+        public bool IsObfuscatedIp(string ip)
```
scottbommarito: This logic is slightly more complex than what I'd like to see in unit tests--is there any reason we aren't just comparing an unobfuscated IP that we pass in with an obfuscated IP generated using the method from `Obfuscator`?

src/NuGetGallery.Core/Auditing/PackageAuditRecord.cs
```diff
@@ -71,5 +71,62 @@ public override string GetPath()
...
+        {
+            Package obfuscatedPacakge = new Package();
+            obfuscatedPacakge.PackageRegistrationKey = PackageRecord.PackageRegistrationKey;
+            obfuscatedPacakge.Copyright = PackageRecord.Copyright;
+            obfuscatedPacakge.Created = PackageRecord.Created;
+            obfuscatedPacakge.Description = PackageRecord.Description;
+            obfuscatedPacakge.ReleaseNotes = PackageRecord.ReleaseNotes;
+            obfuscatedPacakge.DownloadCount = PackageRecord.DownloadCount;
+            obfuscatedPacakge.HashAlgorithm = PackageRecord.HashAlgorithm;
```
skofman1: >e [](start = 28, length = 1)

It will be simpler to obfuscate AuditedPackage directly then to clone the 'Package' itself

src/NuGetGallery.Core/Auditing/AuditActor.cs
```diff
@@ -109,6 +109,21 @@ public static Task<AuditActor> GetCurrentMachineActorAsync()
...
         }
 
+        public static AuditActor Obfuscate(AuditActor actor)
+        {
+            if(actor == null )
+            {
+                return null;
+            }
+            return new AuditActor(actor.MachineName,
+                Obfuscator.ObfuscateIp(actor.MachineIP),
```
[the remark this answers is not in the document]
cristinamanum: There is unit testing that will catch if the data is not obfuscated. I do not see the risk here.

src/NuGetGallery.Core/Auditing/AuditingService.cs
```diff
@@ -88,6 +89,16 @@ protected virtual Task<AuditActor> GetActorAsync()
             return AuditActor.GetCurrentMachineActorAsync();
         }
 
+        /// <summary>
+        /// Used for cases when the AuditService needs to do more processing of the audit record before saving the data to the storage.
+        /// </summary>
+        /// <param name="record">The current record.</param>
+        /// <returns>The record to be saved.</returns>
+        protected virtual AuditRecord GetRecordToPersist(AuditRecord record)
```
skofman1: a method name that implies "pre-processing" will be clearer

src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
```diff
@@ -43,11 +55,21 @@ public CloudAuditingService(string instanceId, string localIP, CloudBlobContaine
...
+        }
+
+        protected override AuditRecord GetRecordToPersist(AuditRecord record)
+        {
+            return record.Obfuscate();
         }
 
         protected override async Task SaveAuditRecordAsync(string auditData, string resourceType, string filePath, string action, DateTime timestamp)
         {
+            if (!cloudAuditPersistedTypes.Contains(resourceType.ToLowerInvariant()))
```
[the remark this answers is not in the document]
cristinamanum: Added.

src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
```diff
@@ -17,12 +18,23 @@ namespace NuGetGallery.Auditing
...
         private Func<Task<AuditActor>> _getOnBehalfOf;
 
+        /// <summary>
+        /// To be used for testing only.
+        /// </summary>
+        public CloudAuditingService(string instanceId, string localIP, Func<Task<AuditActor>> getOnBehalfOf)
+        {
+            _instanceId = instanceId;
+            _localIP = localIP;
+            _getOnBehalfOf = getOnBehalfOf;
```
skofman1: null check

src/NuGetGallery.Core/Auditing/Obfuscator.cs
cristinamanum: How the hash will be used?  What scenario will address. The obfuscation means that there is not any way back to the customer identification.

scottbommarito: A one-way hash will allow us to be able to distinguish users from each other in audit logs without being able to determine which user actually performed the action.

If we obfuscate every username in the audit logs as the same value, it will be impossible to obfuscate different users from each other, which could create issues for us when using the audit logs.

src/NuGetGallery.Core/Auditing/AuditActor.cs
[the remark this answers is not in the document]
scottbommarito: I think it's possible, with your implementation, that if someone adds new auditing scenario that they could forget to call `Obfuscate` and forget to add unit tests. I think my design is easier to maintain and easier to read, but it's up to you.

src/NuGetGallery.Core/Auditing/PackageAuditRecord.cs
cristinamanum: Removed the Package creation and use only the Package Audit. If new filed will be added to the package they will not be audited unless explicit action is taken to add them to auditing.

src/NuGetGallery.Core/Auditing/AuditingService.cs
shishirx34: `PreProcessRecord` perhaps?

src/NuGetGallery.Core/Auditing/Obfuscator.cs
shishirx34: That could potentially be considered a PII where you are able to distinguish the users(even if hashed) coupled with the MachineIP(even obfuscated), I am no expert but I would be cautious of using hash without getting an opinion from the privacy contact.

scottbommarito: I would reach out to the privacy contact then--if all usernames are obfuscated to be identical, using audit logs to trace down some malicious behavior might be difficult if not impossible.

src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
cristinamanum: Updated.

src/NuGetGallery.Core/Auditing/PackageAuditRecord.cs
cristinamanum: Now the package is not cloned but the AuditedPackage is obfuscated.

tests/NuGetGallery.Core.Facts/Auditing/CloudAuditingServiceTests.cs
cristinamanum: Updated

src/NuGetGallery.Core/Auditing/Obfuscator.cs
cristinamanum: If this become an real scenario it can be done later.  Audit data is still available to be used as it was before .

chenriksson:
@cristinamanum Per discussion, just wondering if it would be better to separate out the obfuscation code into a custom Json.Net converter that only the CloudAuditingService uses. We could have attributes on audit record properties to indicate what this custom converter obfuscates.

```
class AuditedUserRecord {
    [Obfuscate]
    string Username { get; }
}
class CloudAuditingService {
    override SerializeObject(AuditEntry entry) {
        JsonConvert.SerializeObject(entry, AuditRecordSerializerSettings, new ObfuscationConverter())
    }
}
```

Take this with a grain of salt... not sure if this would be feasible or an improvement.

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedPackage.cs
```diff
@@ -91,5 +91,49 @@ public static AuditedPackage CreateFrom(Package package)
                 PackageStatusKey = (int)package.PackageStatusKey,
             };
         }
+
+        public static AuditedPackage CreateObfuscatedAuditPackage(AuditedPackage auditPackage)
+        {
+            return new AuditedPackage
+            {
+                PackageRegistrationKey = auditPackage.PackageRegistrationKey,
```
scottbommarito: Unfortunately, this is only slightly better than the previous implementation, and still has the majority of the issues.

If a new field is added to `AuditPackage`, this implementation will not throw if it is not added here, so we could lose fields we intend to audit. It is better that we are using `AuditPackage` instead of `Package` for this purpose because at least any additional fields would be added to this same file, but it could still be easy to miss.

Additionally, as I mentioned before, it is difficult to see which fields are actually changed between the original `auditPackage` and the obfuscated one. As far as I can tell, it appears to only be `UserKey` and `FlattenedAuthors`, although I am confused why `FlattenedAuthors` is obfuscated.

I would really like to see an implementation that does not require enumerating through all of the fields of an object or calling an `Obfuscate` method. I think @chenriksson's suggestion of using custom JSON serialization is perfect for this.

cristinamanum: There are multiple audit services. Some of them will obfuscate data, some of them will not. I updated the code to use Json Converter but the types cannot be adorned with attributes because the data needs to be serialized differently. The converters are used by each Auditing Service to control how data will be serialized.

The flatten authors needs to be obfuscated because they contain author names.

The AuditedPackage contains two methods:
The CreatedFrom - the old method , and a new method CreateObfuscatedAuditPackage that will obfuscate the required fileds. 
If a new property is added to the AuditedPackage the risk to not really audit is 0 because the obfuscation path applies only to the Azure storage. The real audit path is not changed. 
However if Feed2Catalog or V3Monitoring job will need a new added property on AuditedPackage the CreateObfuscatedAuditPackage and the CreatedFrom will need to be updated.

src/NuGetGallery.Core/Auditing/AuditActor.cs
```diff
@@ -109,6 +109,21 @@ public static Task<AuditActor> GetCurrentMachineActorAsync()
                 onBehalfOf);
         }
 
+        public static AuditActor Obfuscate(AuditActor actor)
+        {
+            if(actor == null )
```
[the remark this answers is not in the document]
cristinamanum: Method removed

src/NuGetGallery.Core/Auditing/PackageAuditRecord.cs
```diff
@@ -71,5 +71,62 @@ public override string GetPath()
             return $"{Id}/{NuGetVersionFormatter.Normalize(Version)}"
                 .ToLowerInvariant();
         }
+
+        public override AuditRecord Obfuscate()
+        {
+            var obfuscatedPackage = CreateObfuscatedPackage();
+            var obfuscatedAuditedPackage = AuditedPackage.CreateFrom(obfuscatedPackage);
+
+            return new PackageAuditRecord(Id, Version, Hash, obfuscatedAuditedPackage, RegistrationRecord, Action, Reason );
```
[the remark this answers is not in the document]
cristinamanum: Source code was updated.

src/NuGetGallery.Core/Auditing/AuditingService.cs
cristinamanum: Method removed

src/NuGetGallery.Core/Auditing/Obfuscator.cs
cristinamanum: The audit logs are still present as they were before. The obfuscation applies only to the CloudAudit service that exists for two reasons: Feed2Catalog and v3Monitoring.
None of them needs the user information.

src/NuGetGallery.Core/Auditing/AuditActor.cs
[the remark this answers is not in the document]
cristinamanum: The code was updated. Each AuditService can subscribe to different json converters that will control the structure of the data being persisted.

chenriksson:
@cristinamanum See the following branch for a potential implementation that is type-agnostic:
https://github.com/NuGet/NuGetGallery/compare/dev...chenriks-obfuscation?expand=1

src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
```diff
@@ -125,5 +133,22 @@ public Task<bool> IsAvailableAsync()
...
+
+        public override string RenderAuditEntry(AuditEntry entry)
+        {
+            if (entry == null)
+            {
+                throw new ArgumentNullException(nameof(entry));
+            }
+
+            var settings = GetJsonSerializerSettings();
+            settings.Converters.Add(new ObfuscatorJsonConverter(entry));
```
chenriksson: Is this modifying the static instance that is shared with the other auditing service? Would double-check this...

src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
```diff
@@ -125,5 +133,22 @@ public Task<bool> IsAvailableAsync()
...
+            return JsonConvert.SerializeObject(entry, settings);
+        }
+
+        public override bool RecordWillBePersisted(AuditRecord auditRecord)
+        {
+            return auditRecord.GetType() == typeof(PackageAuditRecord);
```
chenriksson: We'll only write package audit records to cloud auditing?

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <param name="instance"></param>
+        public ObfuscatorJsonConverter(object instance)
+        {
+            Instance = instance ?? throw new ArgumentNullException(nameof(instance));
+        }
+
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(string) ||
+                   objectType == typeof(int?);
```
chenriksson: Good for now; if we need to expand to more value types, we could do objectType.IsValueType or nullable of value type.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        /// <summary>
+        /// Gets the <see cref="PropertyInfo"/> from a full property path.
+        /// </summary>
+        /// <param name="path"></param>
+        /// <returns>The <see cref="PropertyInfo"/> from the <paramref name="path"/>.</returns>
+        protected virtual PropertyInfo ResolvePath(string path)
```
chenriksson: Can probably make private, unless there's need to extend this class.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            var property = ResolvePath(writer.Path);
+            if (property != null)
+            {
+                var obfuscatorAttribute = property.GetCustomAttribute(typeof(ObfuscateAttribute));
+
+                if (obfuscatorAttribute != null)
+                {
+                    var obfuscationType = (ObfuscationType)typeof(ObfuscateAttribute).GetProperty("ObfuscationType").GetValue(obfuscatorAttribute);
```
chenriksson: I don't think you need reflection to get the attribute property. `GetCustomAttribute` should provide the attribute instance.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+                case ObfuscationType.IP:
+                    return Obfuscator.ObfuscateIp(value.ToString());
+                default:
+                    throw new ArgumentException(nameof(obfuscationType));
+            }
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            return serializer.Deserialize(reader, objectType);
```
chenriksson: Did you verify that this works... I remember accidentally getting infinite recursion when calling serializer.Serialize(write, ....) from WriteJson.

chenriksson:
@cristinamanum Looks mostly good to me... just a few comments.

src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
cristinamanum: Yes

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+                case ObfuscationType.UserKey:
+                    return "-1";
+                case ObfuscationType.IP:
+                    return Obfuscator.ObfuscateIp(value.ToString());
+                default:
+                    throw new ArgumentException(nameof(obfuscationType));
+            }
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
```
scottbommarito: Any reason we need to implement deserialization here? If it's any trouble to get working I'd just `throw new NotImplementedException()`.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
cristinamanum: I think that having it protected will make the type more the extensible and it does not have any downside.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
cristinamanum: Not needed for now.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
cristinamanum: Make read scenario not supported.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
cristinamanum: Updated to get the value from the attribute.

src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
cristinamanum: No it gets a new one.

Issue #4659, opened 2017-09-11
Send failed validation email from orchestrator to customer


joelverhagen:
Also move successful push email notification to orchestrator.

joelverhagen:
Email body:

> The package [foo 1.0.0](https://www.nuget.org/packages/foo/1.0.0) failed validation and was therefore not published on NuGet.org gallery. Note that the package will not be available for consumption and you will not be able to push the same package ID and version until further action is taken. Please [contact support](mailto:support@nuget.org) for next steps.

We also need to move the existing package push notification from the Gallery to the orchestrator. The gallery still needs the logic for package push notification for when async validation is disabled, so this should probably be put in NuGetGallery.Core.

PR #5165, opened 2017-12-11
Reverting back to "package published" email

agr:
Conditional on no blocking async validation.

Related to https://github.com/NuGet/NuGetGallery/issues/4659 and https://github.com/NuGet/Engineering/issues/998 - Gallery side update.

### DIFF
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -499,11 +499,14 @@ namespace NuGetGallery
                         await AuditingService.SaveAuditRecordAsync(
                             new PackageAuditRecord(package, AuditedPackageAction.Create, PackageCreatedVia.Api));
 
-                        // Notify user of push
-                        MessageService.SendPackageUploadedNotice(package,
-                            Url.Package(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
-                            Url.ReportPackage(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
-                            Url.AccountSettings(relativeUrl: false));
+                        if (!(ConfigurationService.Current.AsynchronousPackageValidationEnabled && ConfigurationService.Current.BlockingAsynchronousPackageValidationEnabled))
+                        {
+                            // Notify user of push unelss async validation in blocking mode is used
+                            MessageService.SendPackageAddedNotice(package,
+                                Url.Package(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
+                                Url.ReportPackage(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
+                                Url.AccountSettings(relativeUrl: false));
+                        }
 
                         TelemetryService.TrackPackagePushEvent(package, user, User.Identity);
 
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -501,7 +501,7 @@ namespace NuGetGallery
 
                         if (!(ConfigurationService.Current.AsynchronousPackageValidationEnabled && ConfigurationService.Current.BlockingAsynchronousPackageValidationEnabled))
                         {
-                            // Notify user of push unelss async validation in blocking mode is used
+                            // Notify user of push unless async validation in blocking mode is used
                             MessageService.SendPackageAddedNotice(package,
                                 Url.Package(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
                                 Url.ReportPackage(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -499,11 +499,14 @@ public virtual Task<ActionResult> CreatePackagePost()
...
                             new PackageAuditRecord(package, AuditedPackageAction.Create, PackageCreatedVia.Api));
 
-                        // Notify user of push
-                        MessageService.SendPackageUploadedNotice(package,
-                            Url.Package(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
-                            Url.ReportPackage(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
-                            Url.AccountSettings(relativeUrl: false));
+                        if (!(ConfigurationService.Current.AsynchronousPackageValidationEnabled && ConfigurationService.Current.BlockingAsynchronousPackageValidationEnabled))
+                        {
+                            // Notify user of push unelss async validation in blocking mode is used
```
scottbommarito: `unelss` -> `unless`

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -499,11 +499,14 @@ public virtual Task<ActionResult> CreatePackagePost()
                         await AuditingService.SaveAuditRecordAsync(
                             new PackageAuditRecord(package, AuditedPackageAction.Create, PackageCreatedVia.Api));
 
-                        // Notify user of push
-                        MessageService.SendPackageUploadedNotice(package,
-                            Url.Package(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
-                            Url.ReportPackage(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
-                            Url.AccountSettings(relativeUrl: false));
+                        if (!(ConfigurationService.Current.AsynchronousPackageValidationEnabled && ConfigurationService.Current.BlockingAsynchronousPackageValidationEnabled))
```
scottbommarito: What happens when `BlockingAsynchronousPackageValidationEnabled` is enabled but `AsynchronousPackageValidationEnabled` is disabled? Does `BlockingAsynchronousPackageValidationEnabled` have no effect?

tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
```diff
@@ -184,8 +184,12 @@ public class TheCreatePackageAction
...
             }
 
-            [Fact]
-            public async Task CreatePackageWillSendPackageAddedNotice()
+            [Theory]
+            [InlineData(false, false, 1)]
+            [InlineData( true, false, 1)]
+            [InlineData(false,  true, 1)]
+            [InlineData( true,  true, 0)]
+            public async Task CreatePackageWillSendPackageAddedNotice(bool asyncValidationEnabled, bool blockingValidationEnabled, int expectedNumCalls)
```
scottbommarito: nit: this would probably more readable if `int expectedNumCalls` was instead `bool isPackageAddedNoticeSent`

```
controller.MockMessageService
    .Verify(ms => ms.SendPackageAddedNotice(package, It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>()),
    Times.Exactly(isPackageAddedNoticeSent ? 1 : 0));
```

Same with the similar test in `PackagesControllerFacts`

src/NuGetGallery/Controllers/ApiController.cs
agr: Nothing happens. `BlockingAsynchronousPackageValidationEnabled` should be ignored if `AsynchronousPackageValidationEnabled` is `false`.

src/NuGetGallery/Controllers/ApiController.cs
agr: fixed

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -3402,6 +3402,57 @@ public static IEnumerable<object[]> WillApplyEdits_Data
...
+                        messageService: fakeMessageService);
+
+                    controller.SetCurrentUser(TestUtility.FakeUser);
+
+                    // Act
+                    await controller.VerifyPackage(new VerifyPackageRequest { Listed = true, Edit = null });
+
+                    // Assert
+                    fakeMessageService
+                        .Verify(ms => ms.SendPackageAddedNotice(fakePackage, It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>()),
```
chenriksson: Maybe it should verify specific package id/version strings, not just `It.IsAny<string>`?

agr: Both instances of `SendPackageAddedNotice` calling use `UrlHelper` extensions to generate URLs, so expecting certain strings to be passed there would introduce unnecessary dependencies for the test (if routing settings change, test would start to fail)

tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
agr: Updated.
```
            [InlineData(false, false,  true)]
            [InlineData( true, false,  true)]
            [InlineData(false,  true,  true)]
            [InlineData( true,  true, false)]
```
Does not look very readable though :)

PR #5186, opened 2017-12-13
Hide Documentation when it's empty

skofman1:
Fix for: https://github.com/NuGet/NuGetGallery/issues/5184

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -263,7 +263,7 @@
 
             @if (!Model.Deleted)
             {
-                if (Model.ReadMeHtml != null)
+                if (!string.IsNullOrEmpty(Model.ReadMeHtml))
                 {
                     <div id="readme-collapser-container">
                         <h2>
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -263,7 +263,7 @@
 
             @if (!Model.Deleted)
             {
-                if (Model.ReadMeHtml != null)
+                if (!string.IsNullOrEmpty(Model.ReadMeHtml))
```
joelverhagen: Do packages without documentation have empty string readme? That is weird, considering there is no .md file at all for 99% of packages.

joelverhagen: Said another way, why is our null so weak?

skofman1: ReadMeService.GetReadMeHtmlAsync returns string.Empty for the "doesn't exist" case. Agree that it would be better for it to return null.

joelverhagen: Can we do that change there instead? We basically have two code paths here, the null case and the empty case which seems to mean the same thing. Ideally if someone is removing their documentation by setting it to empty string, we should detect that case, delete the .md, and later return null here.

Issue #5173, opened 2017-12-12
ManagePackages: InvalidOperationException due to serialization max length limit


chenriksson:
The following 500 error is seen with testaccount@gmail.com on the DEV environment:

![image](https://user-images.githubusercontent.com/15364949/33897656-e829e426-df1a-11e7-9bbc-e0ea05d75df4.png)

chenriksson:
Verifying no limit with test account, and tracking performance improvements:
- Reduced RouteCollection.GetVirtualPath calls (was 5+ per package, now 5+)
- Switched to Json.NET serializer

| Scenario | Baseline Env | Baseline Result | Env | Result | %Diff |
| --- | --- | --- | --- | --- | --- |
| Packages | PROD0 (1247pkgs) | 30.01s | INT0 (1524 pkgs) | 1.52s | -94.94% |

PR #5187, opened 2017-12-13
Switch to Json.NET serializer which has no max length limit and better perf

chenriksson:
Fixes issue #5173

### DIFF
--- a/src/NuGetGallery/Views/Users/Packages.cshtml
+++ b/src/NuGetGallery/Views/Users/Packages.cshtml
@@ -1,4 +1,5 @@
 ﻿@model ManagePackagesViewModel
+@using Newtonsoft.Json;
 @{
     ViewBag.Title = "Manage My Package";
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -222,7 +223,7 @@
 
 @section BottomScripts {
     <script type="text/javascript">
-        var initialData = @Html.Raw(Json.Encode(new
+        var initialData = @Html.Raw(JsonConvert.SerializeObject(new
```
joelverhagen: Spoke offline. We should track an issue for applying this change elsewhere + evaluating serialization differences. Also consider making a unit test that breaks with Json.Encode and succeeds with JsonConvert (to catch any future regressions when we update newtonsoft.json).

chenriksson: Opened #5188

PR #5189, opened 2017-12-13
Lock UI upload/edit/unlist

skofman1:
Message example from list/unlist:
image

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -317,12 +317,20 @@ namespace NuGetGallery
                 }
 
                 // For existing package id verify if it is owned by the current user
-                if (packageRegistration != null && !PermissionsService.IsActionAllowed(packageRegistration, currentUser, PackageActions.UploadNewVersion))
+                if (packageRegistration != null)
                 {
-                    ModelState.AddModelError(
-                        string.Empty, string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, packageRegistration.Id));
+                    if (!PermissionsService.IsActionAllowed(packageRegistration, currentUser, PackageActions.UploadNewVersion))
+                    {
+                        ModelState.AddModelError(
+                          string.Empty, string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, packageRegistration.Id));
 
-                    return Json(409, new[] { string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, packageRegistration.Id) });
+                        return Json(409, new[] { string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, packageRegistration.Id) });
+                    }
+
+                    if (packageRegistration.IsLocked)
+                    {
+                        return Json(403, new[] { string.Format(CultureInfo.CurrentCulture, Strings.PackageIsLocked, packageRegistration.Id) });
+                    }
                 }
 
                 var nuspecVersion = nuspec.GetVersion();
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -18,62 +18,77 @@
...
-                                 })
-                </div>
+             @if (Model.IsLocked && !User.IsAdministrator())
+             {
+            <div class="validation-error-message-list">
+                @ViewHelpers.AlertDanger(
+                    @<text>
+                        <span>
+                            Package '@Model.Id' has been locked. This means you cannot publish a new version or change the listing status of a published package version. Please contact
+                            <a href="mailto:support@nuget.org?subject=Requesting support for a locked package - '@Model.Id'">support@nuget.org</a>
```
joelverhagen: what are the url encoding requirements for mailto?

skofman1: Didn't see any url encoding for other mailto in our code. What do u mean?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1160,29 +1168,34 @@ public virtual ActionResult Delete(string id, string version)
...
                 Version = package.NormalizedVersion,
-                PackageVersions = package.PackageRegistration.Packages
-                    .OrderByDescending(p => new NuGetVersion(p.Version), Comparer<NuGetVersion>.Create((a, b) => a.CompareTo(b)))
-                    .ToList()
+                IsLocked = package.PackageRegistration.IsLocked,
             };
 
-            // Create version selection.
-            model.VersionSelectList = new SelectList(model.PackageVersions.Select(e => new
+            if (!model.IsLocked)
```
joelverhagen: Why fork on IsLocked at all here? couldn't this be a view concern?

src/NuGetGallery/Views/Packages/Delete.cshtml
joelverhagen: That may be wrong. Like for example, spaces are usually encoded with `%20` in URLs. This is a URL with some non-url safe characters

src/NuGetGallery/Controllers/PackagesController.cs
skofman1: Micro optimization... no need to init those if locked.

joelverhagen: i would rather keep the code simple until we see a perf impact

src/NuGetGallery/Views/Packages/Delete.cshtml
skofman1: I tested the URL. The behavior was correct. My e-mail app opened and the To and Title were correct. Is that enough?

src/NuGetGallery/Controllers/PackagesController.cs
skofman1: I dont think it adds complexity to the code, and we save at least a blob storage call because we dont download a readme.md

src/NuGetGallery/Views/Packages/Delete.cshtml
joelverhagen: We should either adhere to the spec (https://tools.ietf.org/html/rfc2368 -- perhaps I am misreading it) or test with a variety of browsers and email clients. I think it would be easier just to adhere to the spec.

joelverhagen:
Could someone craft a request to the following `PackagesController` action (the verify endpoint) to bypass this?
```csharp
public virtual async Task<JsonResult> VerifyPackage(VerifyPackageRequest formData)
```

joelverhagen:
Although it's not currently enabled, it might be worth adding the check to `DeletePackageOnBehalfOfUserAsync`.

joelverhagen:
Also consider `SetLicenseReportVisibility`, although I have no clue where it is used.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Why not:

```
if (model.IsLocked)
{
    return View(model);
}

model.PackageVersions = ...
...

return View(model);
```

src/NuGetGallery/ViewModels/DeletePackageViewModel.cs
```diff
@@ -26,10 +26,14 @@ public DeletePackageViewModel(Package package, IReadOnlyList<ReportPackageReason
...
             };
+
+            IsLocked = package.PackageRegistration.IsLocked;
         }
 
         public SelectList VersionSelectList { get; set; }
 
         public DeletePackagesRequest DeletePackagesRequest { get; set; }
+
+        public bool IsLocked { get; set; }
```
scottbommarito: This shouldn't need to have `set` if you plan to exclusively initialize it from the constructor.

src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -18,62 +18,77 @@
...
-
-                    @Html.DropDownList("version-selection", Model.VersionSelectList,
-                                 new
-                                 {
-                                     @class = "form-control",
-                                     title = "Selection A Version...",
-                                     id = "input-select-version"
-                                 })
-                </div>
+             @if (Model.IsLocked && !User.IsAdministrator())
```
scottbommarito: We should still show a message stating the package is locked to admins, it just shouldn't hide the form.

src/NuGetGallery/Views/Packages/Delete.cshtml
skofman1: Fixed: Using HttpUtility.UrlPathEncode to encode the subject.

src/NuGetGallery/Controllers/PackagesController.cs
skofman1: Don't see a substantial difference in readability.

skofman1:
@joelverhagen you asked: 
Could someone craft a request to the following `PackagesController` action (the verify endpoint) to bypass this?
```csharp
public virtual async Task<JsonResult> VerifyPackage(VerifyPackageRequest formData)
```
The answer is that you will need to upload a package before you can verify it. And upload is blocked for locked packages.

skofman1:
@joelverhagen you commented:
Although it's not currently enabled, it might be worth adding the check to `DeletePackageOnBehalfOfUserAsync`.

Added a note on the feature tracking issue: https://github.com/NuGet/Engineering/issues/921

skofman1:
@joelverhagen you wrote:

> Also consider `SetLicenseReportVisibility`, although I have no clue where it is used.

This is dead code. Opened an issue to track deletion: https://github.com/NuGet/NuGetGallery/issues/5206

src/NuGetGallery/ViewModels/DeletePackageViewModel.cs
skofman1: removed.

src/NuGetGallery/Views/Packages/Delete.cshtml
skofman1: done

Issue #4900, opened 2017-10-24
API key expiry notification mail sends incorrect URL


anangaur:
Currently the following mail gets sent:

--------
Hi <user>,
We wanted to inform you that the following API key(s) on NuGet Gallery will expire soon:
• \<API key name> – expires in \<duration> day(s).
Visit https://www.nuget.org/account to generate a new API key(s) so that you can continue pushing packages using them.

Best regards,
NuGet Gallery

--------

Note the URL is https://www.nuget.org/account. 

It should be: https://www.nuget.org/account/ApiKeys

ryuyu:
There is a branch where the message that the gallery sends back in the header is fixed (https://github.com/NuGet/NuGetGallery/tree/ryuyu-fix-apikeyexpirymessage). However, i believe this will need a corresponding fix in the job that sends out the api key expiry emails.

scottbommarito:
@ryuyu the change to the job is just configuration--it's set in octopus

scottbommarito:
I have updated the configuration.

chenriksson:
@loic-sharma and I verified the header URL in PROD. Will check with Scott if CredentialExpiration job w/ config change was deployed.

loic-sharma:
Scott confirmed that the job was deployed with the config changes.

PR #5201, opened 2017-12-18
Fix API expiry notification email to use the correct URL

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/4900

When UI redesign happened, we started sending the wrong address because we changed the URL that the API keys could be set from.

By using the `UrlHelper` extension for navigating to the API keys page we should guarantee that the address is always correct.


skofman1:
UT then :shipit:

scottbommarito:
@skofman1 Tried to add unit tests but had trouble due to difficulties mocking `UrlHelper` and getting response headers. I've spent 1-2 hours on it and I've decided to timebox it because I don't think the testing is critical. Will file a bug.

skofman1:
fair enough @scottbommarito ! 🚢

PR #5203, opened 2017-12-18
Add API to check for AAD credential

chenriksson:
Need the `CredentialTypes.IsAzureActiveDirectoryCredential` API to enforce an organization policy of requiring that members have linked with AAD. Design review will be tomorrow.

Most of this PR is renaming, or refactoring to reuse the string consts in tests.


chenriksson:
Closing without merge, as I think this will change now that we're storing TenantId with Credential.

PR #5204, opened 2017-12-18
Schema for organization membership requests

chenriksson:
New membership request table will be used for confirming membership in organizations. Request can result from an organization adding new members, or from the initial migration where the first admin account is added. Design meeting scheduled for tomorrow will include this.

Also adds tenantId column to Credential, which can be used to enforce organization policy that AAD tenant ids match.

/CC: @anangaur


scottbommarito:
Let me know when to review this--I assume schema is going to change slightly after the discussion we had today.

chenriksson:
@scottbommarito Per our discussion, I'll see if I can persist the `Organization.TenantId` as security policy instead of a DB column. Thanks!

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -22,6 +22,8 @@ namespace NuGetGallery.Authentication
 {
     public class AuthenticationService
     {
+        private const string tenantIdClaimType = "http://schemas.microsoft.com/identity/claims/tenantid";
```
shishirx34: I believe the claim type will change with my AAD work, this is fine for now.

chenriksson:
@shishirx34 Do you know if tenantId would ever change?

shishirx34:
I don't think the tenantid would change.

chenriksson:
@scottbommarito This is ready for review when you have a chance. I removed TenantId column from the Organizations table, and will instead do security policies in a separate PR.

src/NuGetGallery.Core/Entities/Organization.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
```
scottbommarito: nit: is this needed?

src/NuGetGallery.Core/Entities/User.cs
```diff
@@ -32,10 +32,22 @@ public User(string username)
...
         /// <summary>
-        /// Organization memberships for a non-organization <see cref="User"/> account.
+        /// Organization memberships, for a non-organization <see cref="User"/> account.
         /// </summary>
         public virtual ICollection<Membership> Organizations { get; set; }
 
+        /// <summary>
+        /// Organization membership requests, for a non-organization <see cref="User"/> account.
+        /// </summary>
+        public virtual ICollection<MembershipRequest> OrganizationRequests { get; set; }
```
scottbommarito: These are requests for a user to become an organization, with an admin specified by the request, correct?

I would explicitly state that. Perhaps name this field `OrganizationMigrationRequest` or something.

src/NuGetGallery.Core/Entities/User.cs
```diff
@@ -32,10 +32,22 @@ public User(string username)
...
+
+        /// <summary>
+        /// Organization membership requests, for an organization <see cref="Organization"/> account.
+        /// 
+        /// Relationship is on the User table because organization transformation may not be complete.
+        /// </summary>
+        public virtual ICollection<MembershipRequest> MemberRequests { get; set; }
```
scottbommarito: The more I look at this the more concerned I am that keeping `MembershipRequest` a single class for both organization migration and adding members to existing organizations instead of two separate classes is ultimately more confusing to use. There should only be one `OrganizationMigrationRequest` for a user at once and `MembershipRequest`s should only be connected to an existing `Organization`. Given this API, neither of those things are guaranteed, and the code will have to constantly check that those conditions are not violated.

chenriksson: Going to split into 2 separate request tables.

src/NuGetGallery.Core/Entities/User.cs
chenriksson: `OrganizationRequests` are requests to join an organization.

Now that there are separate tables, I have `OrganizationMigrationRequest` below for the migration. Another bonus I see for the separate tables is that I'm also able to constrain this to an optional 1:1 relation, so there's only ever 1 migration request for an account... I like that.

chenriksson: There's also `Organization.MemberRequests` now, which are requests to join an organization - from the Organization direction.

src/NuGetGallery.Core/Entities/Organization.cs
```diff
@@ -29,5 +29,10 @@ public Organization(string name) : base(name)
         /// Organization Memberships to this organization.
         /// </summary>
         public virtual ICollection<Membership> Members { get; set; }
+
+        /// <summary>
+        /// Organization membership requests, for an organization <see cref="Organization"/> account.
```
scottbommarito: `for an organization` is now redundant given this is on the `Organization` object.

src/NuGetGallery.Core/Entities/User.cs
scottbommarito: Oh so these are requests to join an organization AFTER it's migrated?

Issue #5209, opened 2017-12-19
Consider caching AuditActor.GetLocalIpAddressAsync


chenriksson:
Scenario: tested push locally with our shim auditing enabled (both ifX and fileStorage).

Profile shows 15.5% inclusive in SaveAuditRecordAsync (36.1% of ApiController::CreatePackageInternal stack). The cause is the AuditActor's call to `NetworkInterface.GetIsNetworkAvailable` which accounts for 14.6% of the 15.5%.

![image](https://user-images.githubusercontent.com/15364949/34176682-b7847908-e4b5-11e7-95dd-c236223b0de7.png)

Caching AuditActor.GetLocalIpAddressAsync seems to reduce api push server latency by 18.5%.

| Run |  Baseline (avg) | Result (avg) | %Diff |
| --- |  --- | --- | --- |
| 1 |  336.61ms | 274.04ms | -18.5% |
| 2 |  309.71ms | 252.58ms | -18.5% |

Note that this optimization will apply to all auditing, not just api push.

chenriksson:
Didn't get a baseline DEV0 run before optimization was deployed, but captured an INT0 baseline.

| Scenario | Baseline Env | Baseline Result | Env | Result | % Diff |
| --- | --- | --- | --- | --- | --- |
| ApiPush | n/a | n/a | DEV0_1220 | 12669.2ms | n/a |
| ApiPush | INT0_1220 | 4249.3ms | INT0_X | n/a | n/a |

chenriksson:
Scratch the previous chart... having a hard time quantifying the optimization in the live environment. Verified that there are no functional issues and that performance looks at least comparable to before.

PR #5210, opened 2017-12-19
Cache AuditActor.GetLocalIpAddressAsync

chenriksson:
Issue #5209 - estimated to improve api push by 18.5%

### DIFF
--- a/src/NuGetGallery.Core/Auditing/AuditActor.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditActor.cs
@@ -17,6 +17,10 @@ namespace NuGetGallery.Auditing
 {
     public class AuditActor
     {
+        private static string _localIpAddress;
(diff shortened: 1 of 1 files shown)


src/NuGetGallery.Core/Auditing/AuditActor.cs
```diff
@@ -114,20 +114,32 @@ public static Task<AuditActor> GetCurrentMachineActorAsync()
                 onBehalfOf);
         }
 
+        private static string _localIpAddress;
+        private static DateTime _localIpAddressExpiration;
+        private const int _localIpAddressExpirationInMinutes = 5;
```
shishirx34: perhaps set expiration to 10mins? Like you said it shouldn't change frequently and most of our timeouts are around 10mins.

PR #5192, opened 2017-12-15
[Organizations] Fixes and unit tests for UI upload for organizations

scottbommarito:
This PR allows organization members to upload on behalf of their organizations using the UI.

Issue #5043, opened 2017-11-20
Pending uploads that become no longer valid are not obvious to the uploader until they submit the verify form


scottbommarito:
1. begin to upload package A as user Z but do not submit the verify form
2. upload package A as user Y or reserve namespace "A" to user Y
3. navigate to the upload page as user Z

EXPECTED:

We should tell the user that the package they were uploading is no longer valid and cancel the existing upload.

ACTUAL:

The form looks exactly the same, but submission fails.

chenriksson:
When testing this scenario, I find the submission failed with no error displayed to the user. I had to find the error in the network trace. Adding screenshot:

![image](https://user-images.githubusercontent.com/15364949/34696780-ef51c604-f484-11e7-9195-87be5f6b486f.png)

PR #5223, opened 2018-01-02
[Organizations] Fixes and unit tests for UI upload for organizations

scottbommarito:
This PR allows organization members to upload on behalf of their organizations using the UI.

(had some git issues--same as https://github.com/NuGet/NuGetGallery/pull/5192)

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -242,8 +242,6 @@ namespace NuGetGallery
                 return Json(400, new[] { Strings.UploadFileMustBeNuGetPackage });
             }
 
-            PackageRegistration existingPackageRegistration;
-
             // If the current user cannot upload the package on behalf of any of the existing owners, show the current user as the only possible owner in the upload form.
             // If the current user doesn't have the rights to upload the package, the package upload will be rejected by submitting the form.
             // Related: https://github.com/NuGet/NuGetGallery/issues/5043
@@ -325,7 +323,7 @@ namespace NuGetGallery
                 }
 
                 var id = nuspec.GetId();
-                existingPackageRegistration = _packageService.FindPackageRegistrationById(id);
+                var existingPackageRegistration = _packageService.FindPackageRegistrationById(id);
                 // For a new package id verify if the user is allowed to use it.
                 if (existingPackageRegistration == null && ActionsRequiringPermissions.UploadNewPackageId.CheckPermissionsOnBehalfOfAnyAccount(currentUser, new ActionOnNewPackageContext(id, _reservedNamespaceService)) != PermissionsCheckResult.Allowed)
                 {
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1673,7 +1673,7 @@ namespace NuGetGallery
                 catch (InvalidPackageException ex)
                 {
                     _telemetryService.TraceException(ex);
-                    
+
                     return Json(400, new[] { ex.Message });
                 }
 
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -325,7 +325,9 @@ namespace NuGetGallery
                 var id = nuspec.GetId();
                 var existingPackageRegistration = _packageService.FindPackageRegistrationById(id);
                 // For a new package id verify if the user is allowed to use it.
-                if (existingPackageRegistration == null && ActionsRequiringPermissions.UploadNewPackageId.CheckPermissionsOnBehalfOfAnyAccount(currentUser, new ActionOnNewPackageContext(id, _reservedNamespaceService)) != PermissionsCheckResult.Allowed)
+                if (existingPackageRegistration == null && 
+                    ActionsRequiringPermissions.UploadNewPackageId.CheckPermissionsOnBehalfOfAnyAccount(
+                        currentUser, new ActionOnNewPackageContext(id, _reservedNamespaceService)) != PermissionsCheckResult.Allowed)
                 {
                     ModelState.AddModelError(
                         string.Empty, string.Format(CultureInfo.CurrentCulture, Strings.UploadPackage_IdNamespaceConflict));
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -185,7 +185,27 @@ namespace NuGetGallery
 
                     model.IsUploadInProgress = true;
 
-                    var verifyRequest = new VerifyPackageRequest(packageMetadata);
+                    var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageMetadata.Id);
+                    bool isAllowed;
+                    IEnumerable<User> accountsAllowedOnBehalfOf = Enumerable.Empty<User>();
+                    if (existingPackageRegistration == null)
+                    {
+                        isAllowed = ActionsRequiringPermissions.UploadNewPackageId.CheckPermissionsOnBehalfOfAnyAccount(currentUser, new ActionOnNewPackageContext(packageMetadata.Id, _reservedNamespaceService), out accountsAllowedOnBehalfOf) == PermissionsCheckResult.Allowed;
+                    }
+                    else
+                    {
+                        isAllowed = ActionsRequiringPermissions.UploadNewPackageVersion.CheckPermissionsOnBehalfOfAnyAccount(currentUser, existingPackageRegistration, out accountsAllowedOnBehalfOf) == PermissionsCheckResult.Allowed;
+                    }
+
+                    if (!isAllowed)
+                    {
+                        // If the current user cannot upload the package on behalf of any of the existing owners, show the current user as the only possible owner in the upload form.
+                        // The package upload will be rejected by submitting the form.
+                        // Related: https://github.com/NuGet/NuGetGallery/issues/5043
+                        accountsAllowedOnBehalfOf = new[] { currentUser };
+                    }
+
+                    var verifyRequest = new VerifyPackageRequest(packageMetadata, accountsAllowedOnBehalfOf);
 
                     model.InProgressUpload = verifyRequest;
                 }
@@ -221,6 +241,13 @@ namespace NuGetGallery
                 ModelState.AddModelError(String.Empty, Strings.UploadFileMustBeNuGetPackage);
                 return Json(400, new[] { Strings.UploadFileMustBeNuGetPackage });
             }
+
+            PackageRegistration existingPackageRegistration;
+
+            // If the current user cannot upload the package on behalf of any of the existing owners, show the current user as the only possible owner in the upload form.
+            // If the current user doesn't have the rights to upload the package, the package upload will be rejected by submitting the form.
+            // Related: https://github.com/NuGet/NuGetGallery/issues/5043
+            IEnumerable<User> accountsAllowedOnBehalfOf = new[] { currentUser };
 
             using (var uploadStream = uploadFile.InputStream)
             {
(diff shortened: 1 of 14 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -184,7 +184,27 @@ public virtual JsonResult UploadPackageProgress()
 
                     model.IsUploadInProgress = true;
 
-                    var verifyRequest = new VerifyPackageRequest(packageMetadata);
+                    var existingPackageRegistration = _packageService.FindPackageRegistrationById(packageMetadata.Id);
+                    bool isAllowed;
+                    IEnumerable<User> accountsAllowedOnBehalfOf = Enumerable.Empty<User>();
+                    if (existingPackageRegistration == null)
+                    {
+                        isAllowed = ActionsRequiringPermissions.UploadNewPackageId.CheckPermissionsOnBehalfOfAnyAccount(currentUser, new ActionOnNewPackageContext(packageMetadata.Id, _reservedNamespaceService), out accountsAllowedOnBehalfOf) == PermissionsCheckResult.Allowed;
```
chenriksson: nit: wrap long lines

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -184,7 +184,27 @@ public virtual JsonResult UploadPackageProgress()
...
+                    {
+                        isAllowed = ActionsRequiringPermissions.UploadNewPackageVersion.CheckPermissionsOnBehalfOfAnyAccount(currentUser, existingPackageRegistration, out accountsAllowedOnBehalfOf) == PermissionsCheckResult.Allowed;
+                    }
+
+                    if (!isAllowed)
+                    {
+                        // If the current user cannot upload the package on behalf of any of the existing owners, show the current user as the only possible owner in the upload form.
+                        // The package upload will be rejected by submitting the form.
+                        // Related: https://github.com/NuGet/NuGetGallery/issues/5043
+                        accountsAllowedOnBehalfOf = new[] { currentUser };
```
chenriksson: Why is this block necessary? Shouldn't `CheckPermissionsOnBehalfOfAnyAccount` always return the current user?

It looks like the permission check is new to `UploadPackage()`, and was only done in `UploadPackage(HttpPostedFileBase)` before. Is it needed to populate a drop-down of on-behalf-of accounts?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -309,14 +336,14 @@ public virtual JsonResult UploadPackageProgress()
 
                     return Json(409, new string[] { string.Format(CultureInfo.CurrentCulture, Strings.UploadPackage_IdNamespaceConflict) });
                 }
-
+
```
chenriksson: nit: whitespace?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1492,13 +1518,20 @@ private string GetSecurePushPolicyDescriptions()
...
+                var message = string.Format(CultureInfo.CurrentCulture, Strings.VerifyPackage_UserNonExistent, formData.Owner);
+                return Json(400, new[] { message });
+            }
+
             Package package;
             using (Stream uploadFile = await _uploadFileService.GetUploadFileAsync(currentUser.Key))
             {
                 if (uploadFile == null)
                 {
-                    TempData["Message"] = Strings.VerifyPackage_UploadNotFound;
```
chenriksson: Is this a bug fix? Do we not use TempData anymore?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1535,13 +1566,73 @@ private string GetSecurePushPolicyDescriptions()
...
+                            var message = string.Format(CultureInfo.CurrentCulture,
+                                Strings.VerifyPackage_OwnerInvalid,
+                                owner.Username, existingPackageRegistration.Id);
+                            return Json(400, new[] { message });
+                        }
+
+                        // An unknown error occurred.
+                        return Json(400, new[] { Strings.VerifyPackage_UnexpectedError });
+                    }
+                }
```
chenriksson: This is all new validation that we never did? Was validation done just in UploadPackage(HttpPostedFileBase) before? If so, why do we need it here now?

If we need it, could it be moved to a helper method? This action method is really long.

src/NuGetGallery/Strings.resx
```diff
@@ -688,4 +688,19 @@ For more information, please contact '{2}'.</value>
...
+  <data name="UploadPackage_NewVersionOnBehalfOfUserNotAllowed" xml:space="preserve">
+    <value>User '{0}' does not have the rights to upload a new version of an existing package on behalf of user '{1}'.</value>
+  </data>
+  <data name="VerifyPackage_OwnerInvalid" xml:space="preserve">
+    <value>User '{0}' does not have the rights to upload new versions of package '{1}'.</value>
+  </data>
+  <data name="VerifyPackage_UserInvalid" xml:space="preserve">
+    <value>You do not have permission to upload packages as user '{0}'.</value>
+  </data>
+  <data name="VerifyPackage_UserNonExistent" xml:space="preserve">
```
chenriksson: nit, would use verbs and more clarification:
* `UserNonExistent` => `UserDoesNotExist`
* `UserInvalid` => `UserDoesNotHavePermission`
* `OwnerInvalid` => `UserDoesNotHavePermissionOnBehalfOf`

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: This is for pending uploads. Consider the following:

1 - User 1 begins upload of Package A but does not complete it
2 - User 2 begins and completes upload of Package A
3 - User 1 navigates back to the pending upload of Package A

`CheckPermissionsOnBehalfOfAnyAccount` will return false because User 1 doesn't have the right to upload Package A on behalf of User 2. The out parameter is also empty in this case.

Therefore, to populate the dropdown on the page (because actually canceling the invalid pending upload is out of the scope of this PR), we need to ignore the fact that the action is unallowed and just show the current user in the dropdown anyway.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: It's a bug fix. Methods that return `JsonResult`s should never use `TempData` as well.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: With the ability to specify the owner of the package in the form, we now need validation to check that the user provided in the form is correct. This is that validation.

src/NuGetGallery/Strings.resx
scottbommarito: done

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: I think I understand... for the scenario above, before we would just fail with conflict when the user clicked 'Submit'. Now, because we have the drop-down we need to populate it with users that can be uploaded on behalf of.

If this is the case, shouldn't we indicate an immediate error that the ID is no longer available and cancel the upload? Why populate the current user (line 206)?

scottbommarito: I had the behavior in a previous iteration of this PR and was asked to change it because it was outside of the scope of this PR.

There was an issue filed: https://github.com/NuGet/NuGetGallery/pull/5002

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -298,9 +325,9 @@ public virtual JsonResult UploadPackageProgress()
                 }
 
                 var id = nuspec.GetId();
-                var packageRegistration = _packageService.FindPackageRegistrationById(id);
+                existingPackageRegistration = _packageService.FindPackageRegistrationById(id);
```
ryuyu: Why was the definition of this moved to line 245? I don't see it being used before this point.

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: Ok, seems harmless enough to let it continue and fail on submit for now. Hopefully this will be removed as part of the #5043 fix.

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: Maybe we should check our code to make sure we don't have other instances of this bug?

src/NuGetGallery/Controllers/PackagesController.cs
chenriksson: The validation logic is very similar to that for UploadPackage(HttpPostedFileBase). Questions:
- Can you reuse the code, and move to a helper?
- Do we really need validation in both UploadPackage(HttpPostedFileBase) and VerifyPackage(VerifyPackageRequest)? It looks like some validation is/was only done in UploadPackage(HttpPostedFileBase), which is why I wonder if we can rely on that already having been done.

chenriksson: e.g., Checks only done in UploadPackage: packageService.EnsureValid, minClientVersion, package locking, etc

It kind of looks like we intended to only validate on the page load (UploadPackage), and that was good enough for the VerifyPackage callback.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I don't think this bug could occur anywhere else than this method and the edit method. I'll check edit.

scottbommarito: None in edit.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I can do some work making the code reusable, but it will likely require passing in a delegate of some sort (one use calls `CheckPermissions` and one uses `CheckPermissionsOnBehalfOfAnyAccount`).

All the checks in VerifyPackage are for the validity of the form. With the addition of specifying an owner through the form, we must now validate that in this callback as well.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: good catch, will fix

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Adding a helper method was too much work--due to subtle differences between the two, it created more work than it solved.

chenriksson: It would be good to understand why we sometimes only add validation to UploadPackage(HttpPostedFileBase), and sometimes to both. If UploadPackage(HttpPostedFileBase) always precedes VerifyPackage, can we safely limit the validation to there?

Maybe check with @skofman1 for the package locking check, and/or @shishirx34 for the namespace conflict check to see what their take on it was?

It would be great to have either consistent validation, or a comment that clarifies why certain validation is only necessary from UploadPackage. I'll leave it at that, since I'm not sure what the right approach is here. :-)

chenriksson:
@scottbommarito My main concern is the inconsistency of validation between UploadPackage and VerifyPackage. My impression is that you'd like full validation in both, but others are happy with validation only on initial upload completion. I'm also a little concerned with how large the action method is getting after adding the validation logic.

Signing off, but could you check into the inconsistencies, see if some code could be extracted into helpers to reduce the method size, and determine whether code reuse is at all possible? I realize this isn't limited to your code, but an overall approach across the team.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: The reason why the validation between `UploadPackage(HttpPostedFileBase)` (the upload of the nupkg) and `VerifyPackage(VerifyPackageRequest formData)` (the submission of the verify form) differ is because some validation is only necessary/possible to do in one of the two.

For example, the package we upload (the nupkg itself) cannot change with the submission of the verify form, so it only needs to be validated when it is uploaded. On the other hand, the package's lock state could change between when the nupkg is uploaded and when the verify form is submitted, so we must validate it in both.

I'll check through our validation but I believe we follow this rule consistently.

scottbommarito: It seems that package locking is not checked in `VerifyPackage(VerifyPackageRequest formData)`. This suggests that if someone began a pending upload of a package and then that package was locked, they could still complete the upload. This seems like a clear bug, but given as it's an existing behavior, I will file an issue to be fixed in a separate PR.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -298,9 +323,9 @@ public virtual JsonResult UploadPackageProgress()
                 }
 
                 var id = nuspec.GetId();
-                var packageRegistration = _packageService.FindPackageRegistrationById(id);
+                var existingPackageRegistration = _packageService.FindPackageRegistrationById(id);
                 // For a new package id verify if the user is allowed to use it.
-                if (packageRegistration == null && ActionsRequiringPermissions.UploadNewPackageId.CheckPermissionsOnBehalfOfAnyAccount(currentUser, new ActionOnNewPackageContext(id, _reservedNamespaceService)) != PermissionsCheckResult.Allowed)
+                if (existingPackageRegistration == null && ActionsRequiringPermissions.UploadNewPackageId.CheckPermissionsOnBehalfOfAnyAccount(currentUser, new ActionOnNewPackageContext(id, _reservedNamespaceService)) != PermissionsCheckResult.Allowed)
```
skofman1: nit: VERY long line

Issue #5112, opened 2017-12-04
Organizations: UI for new Org (migration)


chenriksson:
Includes:
- Create 'Manage Organizations' view
- Add 'Transform your account to an organization' link
- Add 'Transform account' view

Out of scope: AAD integration

Consider: feature flag, initially disabled until we've finished testing

@anangaur will send updated spec.
Based on our discussion, migration will:
-UI accepts single admin
-Logs out org account
-Prompt for admin login
-Redirect edit org page
-Banner on site to integrate w/ AAD or transform to org, shown to subset based on config

PR #5228, opened 2018-01-03
Organizations: transform account on confirmation

chenriksson:
- Added embedded resource SQL script for account migration
- Added UsersController confirmation action
- Added UserService migration API
- Added IDatabase for test mocks

Screenshot of error, if transform fails. Will confirm messages with @anangaur before merging.
image


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -98,6 +99,60 @@ public virtual ActionResult Account()
         {
             return AccountView(new AccountViewModel());
         }
+        
+        [HttpGet]
+        [Authorize]
+        public virtual async Task<ActionResult> ConfirmTransform(string accountName, string token)
```
scottbommarito: `accountName` should probably be `accountToTransform` for clarity

src/NuGetGallery/Infrastructure/AddMigrateToOrganization.Up.sql
```diff
@@ -0,0 +1,47 @@
+﻿-- Copyright (c) .NET Foundation. All rights reserved.
```
scottbommarito: Why is this a script and not done through EntityFramework?

I can't see anything here that would make it necessary to use a script. We've done transactions through EF before.

src/NuGetGallery/Infrastructure/AddMigrateToOrganization.Up.sql
```diff
@@ -0,0 +1,47 @@
...
+	SELECT @reqCount = COUNT(*)
+	FROM [dbo].[OrganizationMigrationRequests]
+	WHERE NewOrganizationKey = @orgKey
+		AND AdminUserKey = @adminKey
+		AND ConfirmationToken = @token
+	IF @reqCount = 0 RETURN
+
+	BEGIN TRANSACTION
+	BEGIN TRY
+		-- Ensure Organizations do not have credentials or memberships
```
scottbommarito: This isn't "ensuring", this is wiping any of these existing entities if they exist.

IMO we should fail the migration if the user has any memberships--that would suggest that the user was already an organization, which is not something we should just ignore.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -98,6 +99,60 @@ public virtual ActionResult Account()
...
+            catch (TransformAccountException e)
+            {
+                TempData["TransformError"] = e.AsUserSafeException().GetUserSafeMessage();
+                return View("AccountTransformFailed");
+            }
+        }
+        
+        private bool CanTransformIntoOrganization(User user)
+        {
+            if (!user.Confirmed || user.IsAdministrator())
```
scottbommarito: I would assume if that user was already an organization, we would also want to reject the transformation. Any reason this condition isn't also filtered here?

src/NuGetGallery/Strings.resx
```diff
@@ -688,4 +688,25 @@ For more information, please contact '{2}'.</value>
   <data name="PackageIsLocked" xml:space="preserve">
     <value>Package '{0}' has been locked. This means you cannot publish a new version or change the listing status of a published package version. Please contact support@nuget.org.</value>
   </data>
+  <data name="TransformAccount_AdminDoesNotHaveTenantId" xml:space="preserve">
+    <value>Failed to transform the account because you do not have an Azure AD sign-in with tenant ID.</value>
```
scottbommarito: How about: `You account must have an Azure AD sign-in with a tenant ID.`

src/NuGetGallery/Strings.resx
```diff
@@ -688,4 +688,25 @@ For more information, please contact '{2}'.</value>
...
     <value>Package '{0}' has been locked. This means you cannot publish a new version or change the listing status of a published package version. Please contact support@nuget.org.</value>
   </data>
+  <data name="TransformAccount_AdminDoesNotHaveTenantId" xml:space="preserve">
+    <value>Failed to transform the account because you do not have an Azure AD sign-in with tenant ID.</value>
+  </data>
+  <data name="TransformAccount_AdminNotConfirmed" xml:space="preserve">
+    <value>Your account must be confirmed before you can create an organization.</value>
+  </data>
+  <data name="TransformAccount_DatabaseError" xml:space="preserve">
+    <value>Failed to transform the account due to a database error.</value>
```
scottbommarito: "Database error" is a little bit cryptic to the average user.

How about: `There was an unexpected error in transforming the account.`

You can also throw in `Contact support` or something like that if this is something they should contact us about.

src/NuGetGallery/Views/Users/AccountTransformFailed.cshtml
```diff
@@ -0,0 +1,17 @@
...
+    ViewBag.MdPageColumns = Constants.ColumnsFormMd;
+    Layout = "~/Views/Shared/Gallery/Layout.cshtml";
+}
+
+<section role="main" class="container main-container page-api-keys">
+    <div class="row">
+        <h1>
+            Transform Account: Failed
+        </h1>
+        @if (TempData.ContainsKey("TransformError"))
```
scottbommarito: If there isn't a `TransformError`, we should probably should a default message of some sort.

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -2229,6 +2231,190 @@ public class TheRequestAccountDeletionAction : TestContainer
...
         }
+
+        public class TheConfirmTransformAction : TestContainer
+        {
+            [Fact]
+            public async Task WhenAdminIsNotConfirmed_ShowsError()
+            {
+                // Arrange
+                var controller = GetController<UsersController>();
+                var currentUser = new User() { UnconfirmedEmailAddress = "unconfirmed@example.com" };
```
scottbommarito: nit: might be easier to use users from Fakes for this

E.g.
```
var fakes = Get<Fakes>();
var currentUser = fakes.User;
```

The advantages are clearer when you need an Admin user below:
```
var fakes = Get<Fakes>();
var adminUser = fakes.AdminUser;
```

src/NuGetGallery/Controllers/UsersController.cs
chenriksson: I could add additional validation here to guard against it.

src/NuGetGallery/Infrastructure/AddMigrateToOrganization.Up.sql
chenriksson: EF has a limitation with inheritance in that it doesn't support changing an object's type. It must be done with SQL. I can add a comment in UserService.TransformToOrganizationAccount.

src/NuGetGallery/Infrastructure/AddMigrateToOrganization.Up.sql
chenriksson: Sounds reasonable... I'll add a SELECT outside the transaction and fail fast if the account has any memberships.

Gallery-side, we should also check for memberships before allowing migration. Account would need to be removed from organization(s) to be eligible for migration.

src/NuGetGallery/Strings.resx
chenriksson: This would be shown on confirmation (admin sign-in), so it's not 'Your account'. 

/cc @anangaur 

Anand, let me know if you want to provide the exact error messages for these migration failure scenarios. Spec didn't go into details on this.

src/NuGetGallery/Strings.resx
chenriksson: This would happen if the migration request didn't exist, a DB constraint prevented the update, etc. Probably not very likely. I'm ok with 'unexpected error'.

src/NuGetGallery/Views/Users/AccountTransformFailed.cshtml
chenriksson: On migration success, this should redirect to the (future) 'Manage Organization' page. For now, I redirect to the admin account page.

I only added this view because I needed a way of displaying an error to the user.

src/NuGetGallery/Infrastructure/AddMigrateToOrganization.Up.sql
chenriksson: Regarding transactions, note that by using a sproc here I also don't need a custom transaction and suspended execution strategy in the Gallery code.

The disadvantage of suspending execution strategy in Gallery code is that I believe it disables SqlAzure strategy defaults such as retry logic.

src/NuGetGallery/Web.config
```diff
@@ -41,6 +41,7 @@
 
     <add key="Gallery.AsynchronousPackageValidationEnabled" value="false" />
     <add key="Gallery.BlockingAsynchronousPackageValidationEnabled" value="false" />
+    <add key="Gallery.OrganizationsEnabledForDomains" value="microsoft.com" />
```
skofman1: microsoft.com should be set in our private configs, not the default ones

chenriksson: Forgot to remove it - will do

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -98,6 +99,60 @@ public virtual ActionResult Account()
...
+                
+                TempData["Message"] = String.Format(CultureInfo.CurrentCulture,
+                    Strings.TransformAccount_Success, accountNameToTransform);
+
+                // todo: redirect to ManageOrganization (future work)
+                return RedirectToAction("Account");
+            }
+            catch (TransformAccountException e)
+            {
+                TempData["TransformError"] = e.AsUserSafeException().GetUserSafeMessage();
```
skofman1: TransformAccountException doesn't inherit from UserSafeException, so we will always show a generic message here. Is this the intention?

src/NuGetGallery/Services/UserService.cs
```diff
@@ -151,5 +157,52 @@ public virtual User FindByUsername(string username)
...
+        /// new EF contexts created after the transaction is committed (i.e., future requests).
+        /// </summary>
+        public async Task TransformToOrganizationAccount(User accountToTransform, User adminUser, string token)
+        {
+            accountToTransform = accountToTransform ?? throw new ArgumentNullException(nameof(accountToTransform));
+            adminUser = adminUser?? throw new ArgumentNullException(nameof(adminUser));
+            
+            if (string.IsNullOrWhiteSpace(token))
+            {
+                throw new ArgumentNullException(nameof(token));
```
skofman1: ArgumentException since the value isn't necessarily null

src/NuGetGallery/Strings.resx
```diff
@@ -688,4 +688,22 @@ For more information, please contact '{2}'.</value>
   <data name="PackageIsLocked" xml:space="preserve">
     <value>Package '{0}' has been locked. This means you cannot publish a new version or change the listing status of a published package version. Please contact support@nuget.org.</value>
   </data>
+  <data name="TransformAccount_AdminDoesNotHaveTenantId" xml:space="preserve">
+    <value>Failed to transform the account because you do not have an Azure AD sign-in with tenant ID. Contact support@nuget.org for more details.</value>
```
skofman1: I don't think the average user knows what a tenant id is. Please go over the messages with Anand :)

chenriksson:
Per discussion in stand-up, I'm going to see if I can create the migration SQL script as an embedded resource and just invoke it inline from the Gallery. Having DB tests sounds nice, but this sounds like less work.

/cc @skofman1 @joelverhagen @scottbommarito

src/NuGetGallery/Services/UserService.cs
```diff
@@ -151,5 +157,52 @@ public virtual User FindByUsername(string username)
...
+                if (result <= 0)
+                {
+                    // Stored procedure check failed (i.e., migration request didn't match, or membership existed).
+                    throw new TransformAccountException(Strings.TransformAccount_DatabaseError);
+                }
+            }
+            catch (Exception ex) when (ex is SqlException || ex is DataException)
+            {
+                // EF exception when saving account transformation to the database.
+                throw new TransformAccountException(Strings.TransformAccount_DatabaseError);
```
skofman1: The exceptions you throw here are swallowed in UsersController, so we have no log of them. Please add a trace (somewhere)

skofman1:
@chenriksson , we will need DB tests anyhow, since if the script breaks, there's no way to know about it.

chenriksson:
@skofman1 You're right - we still need the DB tests.

I've reverted the migration with sproc, and instead execute SQL via embedded resource.

src/NuGetGallery.Core/Entities/DatabaseWrapper.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            string sqlCommand;
+
+            var assembly = Assembly.GetExecutingAssembly();
+            using (var reader = new StreamReader(assembly.GetManifestResourceStream(name)))
+            {
+                sqlCommand = await reader.ReadToEndAsync();
+            }
+
+            if (!string.IsNullOrEmpty(sqlCommand))
```
joelverhagen: Couldn't execute an empty command be indicative of a bug? I think we should throw here -- or let `ExecuteSqlCommandAsync` do its thing.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -151,5 +155,44 @@ public virtual User FindByUsername(string username)
...
+            {
+                errorReason = Strings.TransformAccount_FailedReasonNotInDomainWhitelist;
+            }
+
+            return errorReason == null;
+        }
+
+        public async Task<bool> TransformUserToOrganization(User accountToTransform, User adminUser, string token)
+        {
+            // todo: check for tenantId and add organization policy to enforce this (future work, with manage organization)
```
joelverhagen: Track TODO unless it's gonna happen very soon

anangaur:
Proposal:

# Failed to transform to an organization
'Microsoft' is already an organization on NuGet.org.

src/NuGetGallery/Services/UserService.cs
chenriksson: It's my next work item. I've been tracking all migration with #5112, but can split this into smaller tasks.

chenriksson:
@anangaur Thanks! Here's updated screenshot:

image

Can you confirm these other messages? Note that most of these would be detected before the user can create the migration request (PR #5241), but I also safe guard against it here.
image

src/NuGetGallery/Services/UserService.cs
```diff
@@ -151,5 +156,50 @@ public virtual User FindByUsername(string username)
...
+            
+            try
+            {
+                return await EntitiesContext.TransformUserToOrganization(accountToTransform, adminUser, token);
+            }
+            catch (Exception ex) when (ex is SqlException || ex is DataException)
+            {
+                // todo: log exception
```
skofman1: log?

skofman1:
@chenriksson , looks good! Please add a log and then :shipit:

src/NuGetGallery/Services/UserService.cs
chenriksson: Removed error handling, which seems more consistent with how we handle SQL exceptions in other places.

PR #5235, opened 2018-01-04
Functional test for package locking

skofman1:
The test requires a locked package to exist on the feed, hence it will be disabled by default.

### DIFF
--- a/tests/NuGetGallery.FunctionalTests.Core/EnvironmentSettings.cs
+++ b/tests/NuGetGallery.FunctionalTests.Core/EnvironmentSettings.cs
@@ -29,6 +29,7 @@ namespace NuGetGallery.FunctionalTests
         private static string _testEmailServerHost;
         private static List<string> _trustedHttpsCertificates;
         private static bool? _defaultSecurityPoliciesEnforced;
+        private static bool? _testPackageLock;
 
         /// <summary>
         /// The environment against which the test has to be run. The value would be picked from env variable.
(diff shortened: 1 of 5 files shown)


tests/NuGetGallery.FunctionalTests.Core/EnvironmentSettings.cs
```diff
@@ -331,6 +332,28 @@ public static bool DefaultSecurityPoliciesEnforced
...
+                    }
+                    else
+                    {
+                        _testPackageLock = false;
+                    }
+                }
+
+                return _testPackageLock.Value;
+            }
+        }
```
chenriksson: Is config necessary because locking isn't deployed yet, or are there cases where we'd normally want this disabled?

scottbommarito: Yeah, I think if we deploy it in the right way, we can guarantee that this locked package will always exist.

E.g.

1 - deploy gallery with new locking package DB schema with no functional changes
2 - upload `LockedPackageId` and set `IsLocked` on the package to true (which has no affect on current gallery because it doesn't handle `IsLocked`)
3 - deploy gallery with functional changes for package locking

The functional test will succeed because the package exists, and we won't need the extra configuration.

skofman1: Users that deploy the Gallery should be able to run Functional tests and have them pass. 
In this case, unless the locked package exists in Gallery, the test will fail. Hence, I added a config value to turn it off by default.

Issue #4038, opened 2017-06-01
All regular expressions must have a MatchTimeout property set or be removed


agr:
Regex matching should be time limited, otherwise it opens up the possibility to craft the input data in a way that would cause the server to consume too much resources while checking the match.

See the [docs](https://msdn.microsoft.com/en-us/library/system.text.regularexpressions.regex.matchtimeout%28v=vs.110%29.aspx).

skofman1:
Issue moved to [NuGet/Engineering #1278](https://github.com/NuGet/Engineering/issues/1278) via [**ZenHub**](https://www.zenhub.com/)

PR #5236, opened 2018-01-04
Fix issue 1043: Obfuscate AI data on request redirect

cristinamanum:
The AI data can contain user names. The user names are in general on the request url field but in request redirect case it can be found on the name and operation_Name fields as well.

With this change:
1. The redirect requests data will be obfuscated 
A name like: GET /profiles/cristinamanu will be changed in GET /profiles/ObfuscatedUserName 
An url like: http://localhost/profiles/cristinamanu will be changed in http://localhost/profiles/ObfuscatedUserName
A operation_Name like: GET /profiles/cristinamanu will be changed in GET /profiles/ObfuscatedUserName 
2. The not redirected requests will be obfuscated and kept in sync with the above structure.  A snapshot of the obfuscated values:
image


src/NuGetGallery/Telemetry/Obfuscator.cs
```diff
@@ -25,9 +51,55 @@ internal static class Obfuscator
...
+            string obfuscatedValue = Regex.Replace(value,
+                                             ObfuscatedTemplates[obfuscatedTemplateKey].ObfuscateTemplate,
+                                             ObfuscatedTemplates[obfuscatedTemplateKey].ObfuscateValue);
+            return obfuscatedValue;
+        }
+
+        internal static bool NeedsObfuscation(string valueToBeVerified, out string obfuscatedTemplateKey)
+        {
+            obfuscatedTemplateKey = string.Empty;
+            var match = ObfuscatedTemplates.Where(template => Regex.IsMatch(valueToBeVerified.ToLower(), template.Key)).FirstOrDefault();
```
joelverhagen: Please add an `OrderBy` to the static so that order is defined. Dictionaries don't have a defined order AFAIK

src/NuGetGallery/Telemetry/Obfuscator.cs
```diff
@@ -25,9 +51,55 @@ internal static class Obfuscator
             "Users/Profiles",
             "Users/ResetPassword"};
 
+        internal static readonly Dictionary<string, ObfuscateMetadata> ObfuscatedTemplates = new Dictionary<string, ObfuscateMetadata>
+        {
+            {@"/packages/(.+)/owners/(.+)/confirm/(.+)", new ObfuscateMetadata(@"/owners/(.+)/confirm", $"/owners/{DefaultTelemetryUserName}/confirm")},
```
joelverhagen: We should probably compile these regular expressions.

src/NuGetGallery/Telemetry/Obfuscator.cs
```diff
@@ -25,9 +51,55 @@ internal static class Obfuscator
...
+            {@"/packages/(.+)/owners/(.+)/reject/(.+)", new ObfuscateMetadata(@"/owners/(.+)/reject", $"/owners/{DefaultTelemetryUserName}/reject") },
+            {@"/packages/(.+)/owners/(.+)/cancel/(.+)", new ObfuscateMetadata(@"/owners/(.+)/cancel", $"/owners/{DefaultTelemetryUserName}/cancel") },
+            {@"/account/confirm/(.+)/(.+)", new ObfuscateMetadata(@"/account/confirm/(.+)/", $"/account/confirm/{DefaultTelemetryUserName}/") },
+            {@"/account/delete/(.+)", new ObfuscateMetadata(@"/account/delete/(.+)", $"/account/delete/{DefaultTelemetryUserName}") },
+            {@"/profiles/(.+)", new ObfuscateMetadata(@"/profiles/(.+)", $"/profiles/{DefaultTelemetryUserName}") },
+            {@"/account/setpassword/(.+)/(.+)", new ObfuscateMetadata(@"/setpassword/(.+)/", $"/setpassword/{DefaultTelemetryUserName}/") },
+            {@"/account/forgotpassword/(.+)/(.+)", new ObfuscateMetadata(@"/forgotpassword/(.+)/", $"/forgotpassword/{DefaultTelemetryUserName}/") }
```
joelverhagen: Why `@`?

src/NuGetGallery/Telemetry/Obfuscator.cs
```diff
@@ -25,9 +51,55 @@ internal static class Obfuscator
...
+        }
+
+        internal static string Obfuscate(string value)
+        {
+            string obfuscatedTemplateKey = null;
+            if (value == null || !NeedsObfuscation(value, out obfuscatedTemplateKey))
+            {
+                return value;
+            }
+            string obfuscatedValue = Regex.Replace(value,
```
joelverhagen: Is this case sensitive? I think our URLs are:
https://www.nuget.org/PROFILES/jamesnk

src/NuGetGallery/Telemetry/Obfuscator.cs
```diff
@@ -25,9 +51,55 @@ internal static class Obfuscator
...
             return url == null ? string.Empty : $"{url.Scheme}://{url.Host}/{DefaultTelemetryUserName}";
         }
+
+        internal static Uri ObfuscateUrl(Uri url)
+        {
+            if (url == null)
+            {
+                return url;
+            }
+            return new Uri($"{url.Scheme}://{url.Host}{Obfuscate(url.AbsolutePath)}");
```
joelverhagen: Do we lose port?

joelverhagen: (e.g. if a customer is running gallery on a non-standard port)

src/NuGetGallery/Telemetry/Obfuscator.cs
joelverhagen: Use case insensitive regex instead of `ToLower()`

chenriksson:
@cristinamanum string regex for obfuscating urls seems fragile and error-prone. Would it be possible instead to just use the existing request's RouteData? Maybe we could even inject logic when mapping routes to specify what should be obfuscated?

Very rough idea of what I'm thinking...
```
class Routes {
    routes.MapRoute(
        name: RouteName.Profile,
        url: "profiles/{username}",
        defaults: new { controller = "Users", action = "Profiles" },
        obfuscations: new { username = 'ObfuscatedUsername' }
}

static class NuGetRouteExtensions {

    public static Dictionary<string, Dictionary<string, string>> UrlObfuscations;

    // using Dictionary, or could use object ViewBag to be consistent with MVC routing APIs
    public static Route MapRoute(this RouteCollection routes, string name, string url, object defaults, Dictionary<string, string> obfuscations) {
        if (obfuscations != null) {
            Obfuscator.UrlObfuscations[url] = obfuscations;
        }
    }

    public static string ObfuscateUri(this Route route, string actualUri) {
        if (UrlObfuscations.ContainsKey(route.Uri) {
            var obfuscations = UrlObfuscations[route.Uri];
            var parts = routeUri.Split('/');
            var actualParts = actualUri.Split('/');
            for (int i = 0; i < parts.length; i++) {
                if (obfuscations.ContainsKey(parts[i])) {
                    actualParts[i] = obfuscations[parts[i]];
                }
            }
            return actualParts.Join('/');
        }
    }
}
```

src/NuGetGallery/Telemetry/Obfuscator.cs
loic-sharma: Do we already use compiled regex expressions in the gallery? .NET only caches up to 15 compiled regex expressions by default (source [here](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.cachesize?view=netcore-2.0#System_Text_RegularExpressions_Regex_CacheSize)), and compiling a regex is rather expensive.

TLDR: Compiling regex may result in worse performance, we should due our due diligence before turning it on.

src/NuGetGallery/Telemetry/Obfuscator.cs
skofman1: Add timeouts to all regex matches: https://github.com/NuGet/NuGetGallery/issues/4038

src/NuGetGallery/Telemetry/Obfuscator.cs
joelverhagen: Nice read, good point.

src/NuGetGallery/Telemetry/Obfuscator.cs
cristinamanum: Updated to use case insensitive. 
Why to orderby? The return is only one element.

src/NuGetGallery/Telemetry/Obfuscator.cs
cristinamanum: I did simple perf tests with: static regex and compiled instance regex . The results show similar perf characteristics. 
For invocation of 1000 https://localhost/profiles/cristinamanu
        Static - 525.6479932 seconds
	Instance Compiled - 525.0719887 seconds

src/NuGetGallery/Telemetry/Obfuscator.cs
cristinamanum: I tend to use verbatim string when working with regex expressions. For these regex there is no difference.

src/NuGetGallery/Telemetry/Obfuscator.cs
cristinamanum: I added case insensitive option.

src/NuGetGallery/Telemetry/Obfuscator.cs
cristinamanum: Added the port.

src/NuGetGallery/Telemetry/Obfuscator.cs
cristinamanum: Added the timeouts.

src/NuGetGallery/Extensions/RouteExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public static void MapRoute(this RouteCollection routes, string name, string url, object defaults, ObfuscatedMetadata obfuscationMetadata)
+        {
+            routes.MapRoute(name, url, defaults);
+            if (!ObfuscatedRouteMap.ContainsKey(url)) { ObfuscatedRouteMap.Add(url, obfuscationMetadata); }
+        }
+
+        public static string ObfuscateUrlPath(this Route route, string urlPath)
+        {
+            var path = route.Url;
```
chenriksson: nit: extra single whitespace?

src/NuGetGallery/Extensions/RouteExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Web.Mvc;
+using System.Web.Routing;
+
+namespace NuGetGallery
+{
+    public static class RouteExtensions
```
chenriksson: nit: class could be internal (tests already have internalsVisibleTo)

src/NuGetGallery/Extensions/RouteExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+            { get; }
+
+            public ObfuscatedMetadata(int obfuscatedSegment, string obfuscateValue)
+            {
+                ObfuscatedSegment = obfuscatedSegment;
+                ObfuscateValue = obfuscateValue;
+            }
+        }
+
+        internal static Dictionary<string, ObfuscatedMetadata> ObfuscatedRouteMap = new Dictionary<string, ObfuscatedMetadata>();
```
chenriksson: nit: maybe name `RoutePartsToObfuscateByUrl` is clearer?

src/NuGetGallery/Extensions/RouteExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            var path = route.Url;
+           if (!ObfuscatedRouteMap.ContainsKey(path))
+            {
+                return null;
+            }
+            var metadata = ObfuscatedRouteMap[path];
+            string[] segments = urlPath.Split('/');
+            segments[metadata.ObfuscatedSegment] = metadata.ObfuscateValue;
+            return segments.Aggregate((x, y) => { return x + "/" + y; });
```
chenriksson: Is this equivalent to `String.Join('/', segments)`? Is Join simpler?

src/NuGetGallery/Helpers/RegexEx.cs
```diff
@@ -8,7 +8,7 @@ namespace NuGetGallery.Helpers
 {
     public static class RegexEx
     {
-        private static readonly TimeSpan Timeout = TimeSpan.FromSeconds(2);
+        internal static readonly TimeSpan Timeout = TimeSpan.FromSeconds(2);
```
chenriksson: still necessary?

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -30,29 +30,32 @@ private void ModifyItem(ITelemetry item)
...
 
-        private Uri ObfuscateUri(RequestTelemetry telemetryItem)
+        public virtual Route Route
         {
-            if(IsPIIOperation(telemetryItem.Context.Operation.Name))
+            get
             {
-                // The new url form will be: https://nuget.org/ObfuscatedUserName
-                return new Uri(Obfuscator.DefaultObfuscatedUrl(telemetryItem.Url));
+                return RouteTable.Routes.GetRouteData(new System.Web.HttpContextWrapper(HttpContext.Current)).Route as Route;
```
chenriksson: nit: namespace unnecessary since 'System.Web' is imported?
nit: maybe this is complex enough to be method like `GetCurrentRouteData`?

src/NuGetGallery/Telemetry/Obfuscator.cs
```diff
@@ -2,8 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using NuGetGallery.Helpers;
```
chenriksson: nit: can revert?

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -30,29 +30,32 @@ private void ModifyItem(ITelemetry item)
...
         }
 
-        private Uri ObfuscateUri(RequestTelemetry telemetryItem)
+        public virtual Route Route
         {
-            if(IsPIIOperation(telemetryItem.Context.Operation.Name))
+            get
             {
-                // The new url form will be: https://nuget.org/ObfuscatedUserName
-                return new Uri(Obfuscator.DefaultObfuscatedUrl(telemetryItem.Url));
```
chenriksson: DefaultObfuscatedUrl was also called by QuietLog.GetObfuscatedServerVariables. Can this use the same technique?

chenriksson:
@cristinamanum Looks really nice!

src/NuGetGallery/Extensions/RouteExtensions.cs
cristinamanum: I think I like more the current name.

src/NuGetGallery/Extensions/RouteExtensions.cs
cristinamanum: Yep - join has better perf than aggregate - updated.

src/NuGetGallery/Extensions/RouteExtensions.cs
cristinamanum: updated

src/NuGetGallery/Helpers/RegexEx.cs
cristinamanum: no - updated

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: Comment 1 : it was not needed - updated 
Comment 2 : I agree method is better suited - updated

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: It will be updated if needed in a different change.

src/NuGetGallery/Telemetry/Obfuscator.cs
cristinamanum: yes - updated

src/NuGetGallery/Extensions/RouteExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+                ObfuscateValue = obfuscateValue;
+            }
+        }
+
+        internal static Dictionary<string, ObfuscatedMetadata> ObfuscatedRouteMap = new Dictionary<string, ObfuscatedMetadata>();
+
+        public static void MapRoute(this RouteCollection routes, string name, string url, object defaults, ObfuscatedMetadata obfuscationMetadata)
+        {
+            routes.MapRoute(name, url, defaults);
+            if (!ObfuscatedRouteMap.ContainsKey(url)) { ObfuscatedRouteMap.Add(url, obfuscationMetadata); }
```
skofman1: nit: { should be in a new line

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -30,29 +30,29 @@ private void ModifyItem(ITelemetry item)
...
             if(requestTelemetryItem != null)
             {
-                requestTelemetryItem.Url = ObfuscateUri(requestTelemetryItem);
+                var route = GetCurrentRoute();
+                if(route == null)
+                {
+                    return;
+                }
+                // Removes the first /
+                var requestPath = requestTelemetryItem.Url.AbsolutePath.Remove(0,1);
```
shishirx34: probably a good idea to test if the first char is '/' before removal?

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -30,29 +30,29 @@ private void ModifyItem(ITelemetry item)
...
+                var obfuscatedPath = route.ObfuscateUrlPath(requestPath);
+                if(obfuscatedPath != null)
+                {
+                    requestTelemetryItem.Url = new Uri(requestTelemetryItem.Url.ToString().Replace(requestPath, obfuscatedPath));
+                    requestTelemetryItem.Name = requestTelemetryItem.Name.Replace(requestPath, obfuscatedPath);
+                    if(requestTelemetryItem.Context.Operation != null && requestTelemetryItem.Context.Operation.Name != null)
```
shishirx34: you can simplify this: `if (requestTelemetryItem.Context.Operation?.Name != null)`

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
chenriksson: Maybe use `requestTelemetryItem.Url.AbsolutePath.TrimStart('/')`?

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: nicer - updated

src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
cristinamanum: I tested
updated  with TrimStart

PR #5241, opened 2018-01-05
[Organizations]: UI for creating migration request

chenriksson:
Screenshot, when request already created (will overwrite):
image

Screenshot, when admin is not found:
image


agr:
Was using the word "Transform" discussed? I have a feeling "Convert" would be more appropriate here. As in "Convert to organization account".

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -99,7 +99,65 @@ public virtual ActionResult Account()
         {
             return AccountView(new AccountViewModel());
         }
-        
+
+        [HttpGet]
+        [Authorize]
+        public virtual ActionResult Transform()
+        {
+            var accountToTransform = GetCurrentUser();
```
cristinamanum: null check for accountToTransform?

src/NuGetGallery/Services/UserService.cs
```diff
@@ -155,6 +155,25 @@ public virtual User FindByUsername(string username)
             await UserRepository.CommitChangesAsync();
             return true;
         }
+        
+        public async Task RequestTransformToOrganizationAccount(User accountToTransform, User adminUser)
+        {
+            accountToTransform = accountToTransform ?? throw new ArgumentNullException(nameof(accountToTransform));
```
cristinamanum: Will the proposed admin requested to accept the request of becoming an admin?

cristinamanum: Do we handle the cases when the accountToTransform is already an organization ?

src/NuGetGallery/Services/UserService.cs
```diff
@@ -155,6 +155,25 @@ public virtual User FindByUsername(string username)
...
+            {
+                accountToTransform.OrganizationMigrationRequest = new OrganizationMigrationRequest();
+            };
+
+            accountToTransform.OrganizationMigrationRequest.NewOrganization = accountToTransform;
+            accountToTransform.OrganizationMigrationRequest.AdminUser = adminUser;
+            accountToTransform.OrganizationMigrationRequest.ConfirmationToken = Crypto.GenerateToken();
+            accountToTransform.OrganizationMigrationRequest.RequestDate = DateTime.UtcNow;
+
+            await UserRepository.CommitChangesAsync();
```
cristinamanum: Does it make sense as the requests to have a Status? Will they be ever completed?

cristinamanum: Do we want to add auditing for this scenario?

src/NuGetGallery/Controllers/UsersController.cs
chenriksson: The Authorize attribute should ensure that there is an authenticated user.

src/NuGetGallery/Services/UserService.cs
chenriksson: Yes, it's handled in [UserService.CanTransformToOrganization](https://github.com/NuGet/NuGetGallery/blob/0e0ad2dd81c226e8c4a2963f20276fd62c2cb85b/src/NuGetGallery/Services/UserService.cs#L186)

chenriksson: see PR #5228 for the confirmation

The proposed/pending organization admin must log in and confirm using the generated confirmationToken.

src/NuGetGallery/Services/UserService.cs
chenriksson: No, the confirmation logic is very similar to email confirmation. The user must be authenticated and verify against the token we generate. Nothing fancy is needed for status.

We considered whether to expire the confirmation after 24hr, but chose to do the same as email confirmation which doesn't expire. We do store RequestDate in case we want to change this.

chenriksson: Yes, there's a separate work item to track auditing which will be done at lower priority.

chenriksson:
@agr "Transform" comes from the [PM spec](https://microsoft.sharepoint.com/:p:/r/teams/NuGet/_layouts/15/Doc.aspx?sourcedoc=%7Bdcd4d64f-c131-4273-81ae-1546f71dde87%7D&action=edit), and also matches the terminology used by GitHub.

src/NuGetGallery/Strings.resx
```diff
@@ -715,4 +715,13 @@ For more information, please contact '{2}'.</value>
   <data name="TransformAccount_FailedReasonIsOrganization" xml:space="preserve">
     <value>account is already an organization</value>
   </data>
+  <data name="TransformAccount_AdminAccountDoesNotExist" xml:space="preserve">
+    <value>Administrator account '{0}' does not exist.</value>
+  </data>
+  <data name="TransformAccount_AdminAccountNotConfirmed" xml:space="preserve">
+    <value>Administrator account '{0}' has not validated their email address.</value>
```
scottbommarito: we should use "confirmed" instead of "validated" for this message because that's how it appears elsewhere on the site

src/NuGetGallery/Views/Users/Transform.cshtml
```diff
@@ -0,0 +1,57 @@
...
+        <div class="@ViewHelpers.GetColumnClasses(ViewBag)">
+            <h1 class="text-center">Transform Account</h1>
+            <div class="text-center ms-font-xxl">
+                <a href="@Url.User(CurrentUser)">@CurrentUser.Username</a>
+            </div>
+
+            <div>
+                <aside class="col-md-3 col-md-push-9">
+                    @Html.Label("Logo")
+                    @ViewHelpers.GravatarImage(CurrentUser.EmailAddress, CurrentUser.Username, 332, responsive: true)
```
scottbommarito: A couple thoughts--
1 - I noticed `332` was what we also used on the profile page. Can we put this in a constant of some sort?
2 - Make sure you check how this looks on smaller screens.

src/NuGetGallery/Views/Users/Transform.cshtml
```diff
@@ -0,0 +1,57 @@
...
+                </aside>
+
+                <div class="row form-group col-md-9 col-md-pull-3">
+                    @Html.Label("Organization")
+                    @Html.ShowTextBoxFor(m => CurrentUser.Username, enabled: false)
```
scottbommarito: These look odd as disabled textboxes, especially given that it makes the text harder to see. Can we do what we have in the verify form for these uneditable fields?

E.g.
image

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -2232,6 +2231,125 @@ public class TheRequestAccountDeletionAction : TestContainer
...
+                Assert.NotNull(result);
+                Assert.Equal(
+                    String.Format(CultureInfo.CurrentCulture,
+                        Strings.TransformAccount_FailedWithReason, "account",
+                        "error"),
+                    controller.TempData["TransformError"]);
+            }
+
+            [Fact]
+            public async Task Post_CanTransformReturnsFalse_ShowsError()
```
scottbommarito: i recommend splitting the `Get` and `Post` into separate classes and creating a base for the shared logic.

E.g.
```
public class TheTransformActions : TestContainer
{
    protected UsersController CreateController(string accountToTransform, string canTransformErrorReason = "")
    {
        ...
    }
}

public class TheGetTransformAction : TheTransformActions
{
    ...
}

public class ThePostTransformAction : TheTransformActions
{
    ...
}
```

tests/NuGetGallery.Facts/Services/UserServiceFacts.cs
```diff
@@ -447,6 +447,87 @@ public void WhenAccountIsInWhitelist_ReturnsTrue()
...
+                        tenantId: "zyx987"));
+
+                service.MockUserRepository.Setup(r => r.CommitChangesAsync()).Returns(Task.CompletedTask).Verifiable();
+                
+                DateTime? requestDate = null;
+                string requestToken = null;
+                for (int i = 0; i < (testOverwrite ? 2 : 1); i++)
+                {
+                    // Act
+                    await service.RequestTransformToOrganizationAccount(account, admin);
```
scottbommarito: it may make this test stronger if, for the rewrite test, a different admin user was used

for example, if there was a bug that didn't set the admin of the request to the new admin but still succeeded the request, this test would return a false negative

src/NuGetGallery/Strings.resx
chenriksson: For some reason I chose 'validated' instead of 'confirmed' because I thought it was consistent with other strings. Will try to be consistent and run any messages by @anangaur.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -99,7 +99,65 @@ public virtual ActionResult Account()
         {
             return AccountView(new AccountViewModel());
         }
-        
+
+        [HttpGet]
+        [Authorize]
+        public virtual ActionResult Transform()
```
scottbommarito: One additional thing I noticed--I think `TransformToOrganization` would be more descriptive and potentially a better name for this action.

tests/NuGetGallery.Facts/Services/UserServiceFacts.cs
chenriksson: done

chenriksson:
@scottbommarito new screenshot below -

image

scottbommarito:
@chenriksson Looks much better! Great work!

Issue #5093, opened 2017-11-30
Delete account workflow should unlist all versions of the package


karann-msft:
Currently, we only unlist the latest version. We should unlist all versions.

karann9 was deleted who was the sole owner of https://www.nuget.org/packages/net461_lib
1.0.2 was unlisted
1.0.0 is still listed

karann-msft:
we should fix this soon, ideally this sprint, before more requests start coming in.

PR #5242, opened 2018-01-07
Fix issues 5093: Delete account workflow should unlist all versions of the package

cristinamanum:
Issue https://github.com/NuGet/NuGetGallery/issues/5093


loic-sharma:
Could we get a better title for this pull request?

src/NuGetGallery/Services/IPackageService.cs
```diff
@@ -29,7 +29,7 @@ public interface IPackageService : ICorePackageService
         Package FindPackageByIdAndVersion(string id, string version, int? semVerLevelKey = null, bool allowPrerelease = true);
 
         Package FindAbsoluteLatestPackageById(string id, int? semVerLevelKey);
-        IEnumerable<Package> FindPackagesByAnyMatchingOwner(User user, bool includeUnlisted);
+        IEnumerable<Package> FindPackagesByAnyMatchingOwner(User user, bool includeUnlisted, bool includeVersions = false);
```
chenriksson: Maybe this should just be separate API? e.g., `FindAllPackagesByAnyMatchingOwner` or `FindPackageVersionsByAnyMatchingOwner`?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -205,11 +205,11 @@ ISupportRequestService supportRequestService
             }
         }
 
-        private async Task RemoveOwnership(User user, User admin, bool unsignOrphanPackages, List<Package> packages)
+        private async Task RemoveOwnership(User user, User admin, bool unlistOrphanPackages, List<Package> packages)
         {
             foreach (var package in packages)
             {
-                if (unsignOrphanPackages && package.PackageRegistration.Owners.Count() <= 1)
+                if (unlistOrphanPackages && _packageService.GetPackageUserAccountOwners(package).Count() <= 1)
```
chenriksson: Just verifying... GetPackageUserAccountOwners flattens and removes duplicates, which is why `<= 1` is ok?

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -313,6 +318,44 @@ private void MergeLatestPackagesByOwner(List<int> ownerKeys, bool includeUnliste
...
+            return package.PackageRegistration.Owners
+                .SelectMany((owner) =>
+                {
+                    if (owner is Organization)
+                    {
+                        return OrganizationExtensions.GetUserAccountMembers((Organization)owner);
+                    }
+                    return new List<User> { owner };
+                })
+                .Distinct();
```
scottbommarito: This may not work as expected. I don't think `User` implements the things that are necessary for `Distinct` to work without an `IEqualityComparer`.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Answer for  @chenriksson - Yes.
Answer for @loic-sharma - updated the title

tests/NuGetGallery.Facts/ViewModels/ListPackageItemViewModelFacts.cs
```diff
@@ -272,5 +272,248 @@ public void UseVersionIfLatestSemVer2AndStableSemVer2NotSame()
...
+
+                // Same user in organization and as individual account
+                var packageRegistration6 = CreatePackageRegistration(6);
+                var user61 = new User() { Username = "user6.1", Key = 61 };
+                packageRegistration6.Owners.Add(new Organization()
+                {
+                    Username = "userOrg6",
+                    Key = 60,
+                    Members = new List<Membership>
+                                                        {
```
scottbommarito: nit here and below: spacing

src/NuGetGallery/Services/IPackageService.cs
cristinamanum: I went back an forth and I went with this option because I think that it is easy to discoverer this method with more options instead of adding a new method. If strong concerns it can be can changed.

src/NuGetGallery/Extensions/OrganizationExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            Queue<Organization> organizations = new Queue<Organization>();
+            organizations.Enqueue(organization);
+
+            while (organizations.Any())
+            {
+                var currentOrganization = organizations.Dequeue();
+                foreach (var membership in currentOrganization.Members)
+                {
+                    if (membership.Member is Organization)
```
chenriksson: Like @scottbommarito said, we don't allow Organizations to be members of Organizations (Gallery-side constraint).

Given that, I would just do the following to flatten and de-dupe:
```
organization.SelectMany(m => m.Members
    .Select(m => m.Member))
    .Distinct(u => u.Key);
```

Or if we want only admins:
```
organization.SelectMany(m => m.Members)
    .Where(m => m.IsAdmin)
    .Select(m => m.Member))
    .Distinct(u => u.Key);
```

src/NuGetGallery/Services/IPackageService.cs
chenriksson: I think separate APIs would have a stronger contract since they would have a more specific job, but don't feel strong either way.

src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs
```diff
@@ -55,12 +55,23 @@ public bool HasSingleOwner
         {
             get
             {
-                var organization = Owners.FirstOrDefault() as Organization;
+                var userAccountOwners = Owners.Where(o => !(o is Organization)).ToList();
+                if (userAccountOwners.Count() > 1)
```
chenriksson: Are these the flattened and de-duped owners? If yes, why check where ! is Organization above? If no, is the `>1` check valid?

src/NuGetGallery/Extensions/OrganizationExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery
+{
+    public static class OrganizationExtensions
+    {
+        /// <summary>
+        /// Returns all the user accounts that are members of an organization.
+        /// If the orgnaization has other child organizations their members will be retuned as well. The result does not filter duplicate elements.
```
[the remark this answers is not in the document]
cristinamanum: Fixed first comment.
Second comment: I felt that is more general to have all the data and let the callers filter. However from the comments it seems that de-duping is the preferred way so I de-duped.

src/NuGetGallery/Extensions/OrganizationExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// </summary>
+        /// <param name="organization">The organization.</param>
+        /// <returns>The <see cref="IEnumerable{User}"/> of users that are not <see cref="Organization"/> and are members of <paramref name="organization"/>.</returns>
+        public static IEnumerable<User> GetUserAccountMembers(this Organization organization)
```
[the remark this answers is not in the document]
cristinamanum: I would like to let the method handle nested organizations. Nothing will be lost and the support is there when the nested organizations will be supported.

src/NuGetGallery/Services/PackageService.cs
cristinamanum: EntityFramework should return the same object for the same db entry. https://msdn.microsoft.com/en-us/library/jj573936(v=vs.113).aspx, https://stackoverflow.com/questions/6542700/do-entity-framework-object-references-equal-for-same-database-objects, following this Distinct will work on the default reference equality.
To not rely on this behavior I implemented IEquatable on User.

tests/NuGetGallery.Facts/ViewModels/ListPackageItemViewModelFacts.cs
cristinamanum: Fixed.

src/NuGetGallery/Extensions/OrganizationExtensions.cs
cristinamanum: Please see the comment above. I de-duped but I prefer to keep the method supporting nested orgs.

src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs
cristinamanum: The Owners collection is like user1, user2, user3, org1, org2. The expectation is the they should be Distinct. For this reason no distinct operator was used.  However I added an explicit Distinct

src/NuGetGallery/Extensions/OrganizationExtensions.cs
chenriksson: If you keep support for nested orgs then add a comment to clarify. Otherwise it's confusing when some parts of our code support the nesting and others don't.

cristinamanum: I changed the comment on the method to mention the support for nested organizations.
I think that Gallery side support and the API support are different. On the API side there is no constraint for not supporting Organizations. A comment that the nested Org are not supported on the Gallery side will need to be updated when that support will be unblocked.

src/NuGetGallery.Core/Entities/User.cs
```diff
@@ -15,7 +15,7 @@ namespace NuGetGallery
     /// future if we want to add constraints for user accounts or user-only settings.
     /// </summary>
     /// <see href="https://weblogs.asp.net/manavi/inheritance-mapping-strategies-with-entity-framework-code-first-ctp5-part-2-table-per-type-tpt" />
-    public class User : IEntity
+    public class User : IEntity, IEquatable<User>
```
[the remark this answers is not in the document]
cristinamanum: I agree with this. I think that it is out of scope for this PR. I added a task to track the issue. 
https://github.com/NuGet/NuGetGallery/issues/5254

src/NuGetGallery/Extensions/OrganizationExtensions.cs
chenriksson: The comment implies that nested organizations are supported. I was thinking something more like:
```
/// Nested organizations (teams) are not supported in the Gallery yet, but this method allows for it in case we lift that constraint.
```

src/NuGetGallery.Core/Entities/User.cs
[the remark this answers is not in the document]
chenriksson: Would prefer we either bring in the more generic approach for all IEntity types, or revert and just use `Distinct(u => u.Key)` for now.

chenriksson:
@cristinamanum Added my 2 cents, feel free to review but it's not blocking

src/NuGetGallery/Extensions/OrganizationExtensions.cs
cristinamanum: I updated with the comment.

src/NuGetGallery.Core/Entities/User.cs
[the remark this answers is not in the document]
cristinamanum: Distinct and Union use the GetHashCode. If not implementing the IEquatable and not relying on the EF object ref equality support the operators will need to have IEqualityComparer implemented. 
The general solution will increase the amount of testing. Will stay with this for now and extend later.

PR #5249, opened 2018-01-08
[Organizations] Fixes and unit tests for API upload for organizations

scottbommarito:
This PR allows organization members to upload on behalf of their organizations using the API with API keys scoped to the organization user.


src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -729,42 +727,48 @@ protected internal virtual Stream ReadPackageFromRequest()
...
-            if (scopes != null)
+            switch (evaluationResult)
             {
-                // Return true if scope allows action for subject (package).
-                return scopes.Any(s => s.AllowsActions(NuGetScopes.PackagePush) && s.AllowsSubject(packageId));
+                case ApiScopeEvaluationResult.Success:
+                    throw new ArgumentException($"{nameof(ApiScopeEvaluationResult.Success)} is not a failed evaluation!", nameof(evaluationResult));
+
+                case ApiScopeEvaluationResult.Forbidden:
+                    return new HttpStatusCodeWithBodyResult(HttpStatusCode.Forbidden, Strings.ApiKeyNotAuthorized);
```
chenriksson: Push / CreatePackageInternal was returning HttpStatusCode.Unauthorized, while other APIs returned Forbidden. Even though it's inconsistent, I think we need to keep existing behavior.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,113 @@
...
+                // Legacy V1 API key without scopes.
+                // Evaluate it as if it has an unlimited scope.
+                scopes = new[] { new Scope(ownerKey: null, subject: NuGetPackagePattern.AllInclusivePattern, allowedAction: NuGetScopes.All) };
+            }
+
+            var failureResult = ApiScopeEvaluationResult.Unknown;
+
+            foreach (var scope in scopes)
+            {
+                if (!scope.AllowsSubject(GetScopeSubjectConverter<TEntity>().ConvertToScopeSubject(entity)))
```
chenriksson: Would we ever use this for any TEntity other than Package? Scopes are specific to certain entities only.

Just wondering if we need the added complexity of making this generic. Before it was simply `scope.AllowsSubject(package.Id)`.

skofman1:
why is the PR not against DEV branch?

scottbommarito:
@skofman1 relies on the other PR. Will be rebased when the first PR is merged (soon).

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: `ReservedNamespace[]` and `ActionOnNewPackageContext` are also used as `TEntity`

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: I noticed this initially and was going to change it, but then I saw that `Forbidden` was already returned by the `ApiScopeRequiredAttribute` which is used by `CreatePackageInternal`. As such, all clients using this API must already handle `Unauthorized` and `Forbidden`; therefore, making this change will improve the consistency of our API and should not impact our clients.

scottbommarito:
@skofman1 now rebased against dev

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
skofman1: Agree with Christy. This code is overly complex and mixing scope evaluation with namespace validation in one class is confusing. Lets discuss ways to simplify.

---
In reply to: [160562721](https://github.com/NuGet/NuGetGallery/pull/5249#discussion_r160562721) [](ancestors = 160562721)

src/NuGetGallery/Authentication/IScopeSubjectEvaluator.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NuGetGallery.Authentication
+{
+    public interface IScopeSubjectConverter<TEntity>
```
skofman1: dont create interfaces for classes with a single implementation

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -139,7 +143,7 @@ protected ApiController()
             // Ideally shouldn't be necessary?
             if (!PackageIdValidator.IsValidPackageId(id ?? string.Empty))
             {
-                return new HttpStatusCodeWithBodyResult(HttpStatusCode.BadRequest, "The format of the package id is invalid");
+                return new HttpStatusCodeWithBodyResult(HttpStatusCode.BadRequest, "The format of the package ID is invalid");
```
skofman1: >I [](start = 110, length = 1)

this change is not required. Please undo

src/NuGetGallery/Authentication/IScopeSubjectEvaluator.cs
scottbommarito: We have many many classes in the gallery that do this for the sake of easy mocking and DI. Additionally, there is a second implementation of this that is used by the unit testing.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: I simplified this code by removing the interface and using a custom TypeConverter instead. This should make it more obvious regarding what's going on.

scottbommarito:
@skofman1 @chenriksson I have finished resolving ur feedback

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
```diff
@@ -40,6 +40,14 @@ public static class ActionsRequiringPermissions
                 accountOnBehalfOfPermissionsRequirement: RequireOwnerOrOrganizationMember,
                 packageRegistrationPermissionsRequirement: PermissionsRequirement.Owner);
 
+        /// <summary>
+        /// The action of verify a package verification key.
+        /// </summary>
+        public static ActionRequiringPackagePermissions VerifyPackage =
+            new ActionRequiringPackagePermissions(
```
joelverhagen: >ActionRequiringPackagePermissions [](start = 16, length = 33)

name params. Also, use const RequireOwnerOrOrganizationMember

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,123 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+
+namespace NuGetGallery.Authentication
+{
+    public class ApiScopeEvaluator : IApiScopeEvaluator
+    {
+        private TypeConverter Converter = new ScopeSubjectTypeConverter();
```
joelverhagen: >Converter [](start = 30, length = 9)

Static readonly

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        private IUserService UserService { get; }
+
+        public ApiScopeEvaluator(IUserService userService)
+        {
+            UserService = userService;
+        }
+
+        // Unit test constructor
```
joelverhagen: >Unit test constructor [](start = 11, length = 21)

This can be XML docs.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,123 @@
...
+                var ownerInScope = scope.HasOwnerScope() ? UserService.FindByKey(scope.OwnerKey.Value) : currentUser;
+
+                var isActionAllowed = action.CheckPermissions(currentUser, ownerInScope, entity);
+                if (isActionAllowed != PermissionsCheckResult.Allowed)
+                {
+                    // Current user cannot do the action on behalf of the owner in the scope or owner in the scope is not allowed to do the action.
+                    var currentFailureResult = ApiScopeEvaluationResult.Forbidden;
+                    if (isActionAllowed == PermissionsCheckResult.ReservedNamespaceFailure)
+                    {
+                        currentFailureResult = ApiScopeEvaluationResult.ConflictReservedNamespace;
```
skofman1: Handling reserved namespace conflicts as part of this code makes it very complex. It forces you to return an enum instead of a true false value. I recommend removing all namespaces handling from here, and move it to APIController, even if it will require calling PermissionsService twice in the case of a new package id.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,123 @@
...
+                    }
+
+                    failureResult = ChooseFailureResult(failureResult, currentFailureResult);
```
joelverhagen: Isn't this just a max of three things? ChooseFailureResult can just be params ApiScopeEvaluationResult[]

src/NuGetGallery/Authentication/ApiScopeEvaluationResult.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+
+    /// <summary>
+    /// The result of evaluating the current user's scopes by using <see cref="ApiScopeEvaluator"/>.
+    /// </summary>
+    /// <remarks>
+    /// When an current user's scopes are evaluated and none evaluate with <see cref="Success"/>, 
+    /// the failed result to return is determined by <see cref="ApiScopeEvaluator.ChooseFailureResult(ApiScopeEvaluationResult, ApiScopeEvaluationResult)"/>.
+    /// </remarks>
+    public enum ApiScopeEvaluationResult
```
joelverhagen: >ApiScopeEvaluationResult [](start = 16, length = 24)

These need explicit values if you are depending on integer comparison. It's sneaky otherwise.

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
```diff
@@ -40,6 +40,14 @@ public static class ActionsRequiringPermissions
...
+                PermissionsRequirement.Owner | PermissionsRequirement.OrganizationAdmin | PermissionsRequirement.OrganizationCollaborator,
+                PermissionsRequirement.Owner);
```
chenriksson: Would add named args here like we did in other permissions fields.

src/NuGetGallery/Authentication/ScopeSubjectTypeConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Globalization;
+using System.Linq;
+
+namespace NuGetGallery.Authentication
+{
+    public class ScopeSubjectTypeConverter : TypeConverter
```
joelverhagen: >ScopeSubjectTypeConverter [](start = 17, length = 25)

This seems to convert type T to a package ID. This should be more clear in the class name or at least XML docs.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
skofman1: Also, removing the namespace check will allow you to stop handling the case when TEntity is ActionOnNewPackageContext, which will simplify the code further, since you can remove generics, and have two overloaded methods: one for PackageRegistration and one for Package.
In general, we should avoid using generics and reflection if possible. They make the code complex and not readable.

---
In reply to: [162116057](https://github.com/NuGet/NuGetGallery/pull/5249#discussion_r162116057) [](ancestors = 162116057)

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <param name="currentUser">The current user attempting to do the action with the given <paramref name="scopes"/>.</param>
+        /// <param name="scopes">The scopes being evaluated.</param>
+        /// <param name="action">The action that the scopes being evaluated are checked for permission to do.</param>
+        /// <param name="entity">The entity that the scopes being evaluated are checked for permission to <paramref name="action"/> on.</param>
+        /// <param name="owner">
+        /// The <see cref="User"/> specified by the <see cref="Scope.OwnerKey"/> of the <see cref="Scope"/> that evaluated with <see cref="ApiScopeEvaluationResult.Success"/>.
+        /// If no <see cref="Scope"/>s evaluate to <see cref="ApiScopeEvaluationResult.Success"/>, this will be null.
+        /// </param>
+        /// <returns>A <see cref="ApiScopeEvaluationResult"/> that describes the evaluation of the .</returns>
+        public ApiScopeEvaluationResult Evaluate<TEntity>(
```
joelverhagen: >TEntity [](start = 49, length = 7)

TEntity without any type constraints is a bit deceptive don't you think? Only three types are really support as far as I know so perhaps three overloads would be easier.

joelverhagen: Right now, you can only discover the allowed TEntity types by looking at implementation or at runtime.

---
In reply to: [162119204](https://github.com/NuGet/NuGetGallery/pull/5249#discussion_r162119204) [](ancestors = 162119204)

src/NuGetGallery/Controllers/ApiController.cs
skofman1: Changing return values of CreatePackageInternal is not part of this PR. Please create an issue and we will discuss it as a team.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
chenriksson: nit: TActionEntity?

src/NuGetGallery/Authentication/ScopeSubjectTypeConverter.cs
chenriksson: This class seems unnecessary. Is it possible to just replace with lambda? - this may require function overloads. i.e.,
```
Evaluator {
  Evaluate(..., Func<string, TActionEntity>, ...)
}
Controller {
  EvaluateApiKeyScope(..., Func<string, TActionEntity>, ...)
  {
      evaluator.Evaluate(...)
  }
  EvaluateApiKeyScopeForPackageRegistration(...) {
      return EvaluateApiKeyScope<PackageRegistration>(..., p => p.Id);
  }
  EvaluateApiKeyScopeForPackageRegistration(...) {
      return EvaluateApiKeyScopeForPackageRegistration(..., p.PackageRegistration, ...);
  }
}
```

src/NuGetGallery/Controllers/ApiController.cs
joelverhagen: Agree with @chenriksson and @skofman1, we shouldn't change API unless we have a good reason. And even if we did it seems out of scope for this PR.

src/NuGetGallery/Authentication/ApiScopeEvaluationResult.cs
joelverhagen: Also, we need to assert ordering in a unit test if you are depending on that.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// </summary>
+        /// <param name="currentUser">The current user attempting to do the action with the given <paramref name="scopes"/>.</param>
+        /// <param name="scopes">The scopes being evaluated.</param>
+        /// <param name="action">The action that the scopes being evaluated are checked for permission to do.</param>
+        /// <param name="entity">The entity that the scopes being evaluated are checked for permission to <paramref name="action"/> on.</param>
+        /// <param name="owner">
+        /// The <see cref="User"/> specified by the <see cref="Scope.OwnerKey"/> of the <see cref="Scope"/> that evaluated with <see cref="ApiScopeEvaluationResult.Success"/>.
+        /// If no <see cref="Scope"/>s evaluate to <see cref="ApiScopeEvaluationResult.Success"/>, this will be null.
+        /// </param>
+        /// <returns>A <see cref="ApiScopeEvaluationResult"/> that describes the evaluation of the .</returns>
```
cristinamanum: nit: ... that describes the evaluation of the . ???

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// Determines the <see cref="ApiScopeEvaluationResult"/> to return from <see cref="EvaluateApiScope(IScopeSubject, out User, string[])"/> when no <see cref="Scope"/>s return <see cref="ApiScopeEvaluationResult.Success"/>.
+        /// </summary>
+        /// <param name="last">The result of the <see cref="Scope"/>s that have been evaluated so far.</param>
+        /// <param name="next">The result of the <see cref="Scope"/> that was just evaluated.</param>
+        private ApiScopeEvaluationResult ChooseFailureResult(ApiScopeEvaluationResult last, ApiScopeEvaluationResult next)
+        {
+            return new[] { last, next }.Max();
+        }
+
+        private string ConvertToScopeSubject(object value)
```
cristinamanum: Suggestion: This looks as a ToString() on specific types. Would it make any sense to go on that route?

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: I agree that it's deceptive, but it's necessary because these types don't share any sort of interface. I wasn't sure how adding some sort of empty interface to all of these types would impact them so I decided not to. If you could suggest some other way to do this I'm open.

src/NuGetGallery/Authentication/ScopeSubjectTypeConverter.cs
scottbommarito: I tried the lambda approach as you suggested and came to the conclusion that this is cleaner. What we really want here is a way to convert any valid object from one type to a string (e.g. scope subject).

Requiring this code to be defined on the controller side seems like an incorrect place for it as it
1) doesn't centralize the conversion logic, making it more difficult to maintain
2) requires each call to redefine the conversion, when there's only a small set of possible conversions

I then took the approach of defining multiple methods on the interface, one for each type. This wasn't ideal because it
1) cluttered the interface, making it less maintainable--I had the same comment on each method with small changes
2) seemed inefficient when I could just define the conversion INSIDE the evaluator

Ultimately, I settled with this approach of using a custom `TypeConverter` that can convert between the states because it
1) centralizes the conversion logic, making it easy to add/remove conversions
2) is easily testable to guarantee that all possible types are handled correctly
3) simplifies the controller's code

scottbommarito: It's not EXACTLY a package ID--it's the subject of a scope. Basically this call converts a `T` into a string that can be used to perform a `scope.IsSubjectAllowed(subject)` call. Given that I'm familiar with this code, it seems somewhat confusing to call this a `PackageIdTypeConverter`, but adding more XML docs seems perfectly fine to me.

src/NuGetGallery/Authentication/ApiScopeEvaluationResult.cs
scottbommarito: I'll add a unit test to assert ordering and explicit values.

skofman1: With the suggestions above, sorting will not be required (since the method will return true/false).

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: Can't be `static` (because of unit tests) but it can be `readonly`

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
scottbommarito: fixed

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: removed this class

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: fixed...thanks haha

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: fixed

scottbommarito:
@skofman1 @joelverhagen @chenriksson finished going through the feedback from our discussion yesterday. let me know if i missed anything

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,132 @@
...
+
+namespace NuGetGallery.Authentication
+{
+    public class ApiScopeEvaluator : IApiScopeEvaluator
+    {
+        private IUserService UserService { get; }
+
+        public ApiScopeEvaluator(IUserService userService)
+        {
+            UserService = userService;
```
skofman1: null check

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,132 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery.Authentication
+{
+    public class ApiScopeEvaluator : IApiScopeEvaluator
+    {
+        private IUserService UserService { get; }
```
skofman1: nit: since it's private you can use a field

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            return aggregateResult;
+        }
+
+        private ApiScopeEvaluationResult ChooseFailureResult(params ApiScopeEvaluationResult[] results)
+        {
+            return results.Max();
+        }
+
+        public ApiScopeEvaluationResult Evaluate(
```
skofman1: nit: public methods should be above private methods

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,132 @@
...
+                {
+                    // Action scopes does not match.
+                    aggregateResult = ChooseFailureResult(aggregateResult, new ApiScopeEvaluationResult(false, PermissionsCheckResult.Unknown, null));
+                    continue;
+                }
+
+                // Get the owner from the scope.
+                // If the scope has no owner, use the current user.
+                int ownerInScopeKey = scope.HasOwnerScope() ? scope.OwnerKey.Value : currentUser.Key;
+                if (ownerInScope == null)
```
skofman1: Great perf improvement!

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,132 @@
...
+                {
+                    ownerInScope = UserService.FindByKey(ownerInScopeKey);
+                }
+                
+                if (ownerInScopeKey != ownerInScope.Key)
+                {
+                    // The set of scopes contains multiple owners. This should not be possible.
+                    aggregateResult = ChooseFailureResult(aggregateResult, new ApiScopeEvaluationResult(false, PermissionsCheckResult.Unknown, null));
```
skofman1: The expected behavior in this case is a 500, because it indicates a server error, not a user error. Throw an exception.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,132 @@
...
+                    continue;
+                }
+
+                var isActionAllowed = action.CheckPermissions(currentUser, ownerInScope, entity);
+                var result = new ApiScopeEvaluationResult(true, isActionAllowed, ownerInScope);
```
skofman1: use named params

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,132 @@
...
+                aggregateResult = ChooseFailureResult(aggregateResult, result);
+                if (isActionAllowed != PermissionsCheckResult.Allowed)
+                {
+                    // Current user cannot do the action on behalf of the owner in the scope or owner in the scope is not allowed to do the action.
+                    continue;
+                }
+
```
skofman1: Since there will be only a single scope that matches, we can move the PermissionsService call out of the loop.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -729,42 +729,69 @@ protected internal virtual Stream ReadPackageFromRequest()
...
+            return GetHttpResultFromFailedApiScopeEvaluationHelper(result, id, version, HttpStatusCode.Forbidden);
+        }
 
-            // Legacy V1 API key (no scopes), or Legacy V2 API key (no owner scope).
-            // Must verify that the current user is the package owner or admin for an organization owner.
-            return GetCurrentUser().IsOwnerOrMemberOfOrganizationOwner(package);
+        /// <remarks>
+        /// Push returns <see cref="HttpStatusCode.Unauthorized"/> instead of <see cref="HttpStatusCode.Forbidden"/> for failures not related to reserved namespaces.
+        /// </remarks>
+        private HttpStatusCodeWithBodyResult GetHttpResultFromFailedApiScopeEvaluationForPush(ApiScopeEvaluationResult result, string id, NuGetVersion version)
```
skofman1: consider removing those methods and using GetHttpResultFromFailedApiScopeEvaluationHelper directly at the call sites.
This will make the code simpler and more readable. The way it's written now, it takes an extra step (or two) to see that the actual return code in case of error is.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: fixed

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: yup

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: IMO calling `GetHttpResultFromFailedApiScopeEvaluationHelper` directly is much more confusing, because the user will pass in `HttpStatusCode.Forbidden` and sometimes get `HttpStatusCode.Conflict`. This structure defines the situation much better--everything but push uses `HttpStatusCode.Forbidden`.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: done

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: done

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,121 @@
...
+            // Check that all scopes provided have the same owner scope.
+            var ownerScopes = scopes.Select(s => s.OwnerKey);
+            var firstOwnerScope = ownerScopes.FirstOrDefault();
+            if (ownerScopes.Any(o => o != firstOwnerScope))
+            {
+                throw new ArgumentException("All scopes provided must have the same owner scope.");
+            }
+
+            foreach (var scope in scopes)
+            {
```
skofman1: Consider using linq:

var matchingScope = scopes.FirstOrDefault(scope.AllowsSubject(getSubjectFromEntity(entity)) && scope.AllowsActions(requestedActions));

if (scope == null)
{
 return new ApiScopeEvaluationResult(ownerInScope, PermissionsCheckResult.Unknown, scopesAreValid: false);
}

ownerInScope = matchingScope .HasOwnerScope() ? _userService.FindByKey(matchingScope .OwnerKey.Value) : currentUser;
var isActionAllowed = action.CheckPermissions(currentUser, ownerInScope, entity);
return new ApiScopeEvaluationResult(ownerInScope, isActionAllowed, scopesAreValid: true);

}

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        public ApiScopeEvaluationResult Evaluate(
+            User currentUser, 
+            IEnumerable<Scope> scopes, 
+            IActionRequiringEntityPermissions<PackageRegistration> action, 
+            PackageRegistration packageRegistration, 
+            params string[] requestedActions)
+        {
+            return Evaluate(currentUser, scopes, action, packageRegistration, (pr) => pr.Id, requestedActions);
```
joelverhagen: nit: you don't need the parens around `(pr)`

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public ApiScopeEvaluationResult Evaluate(
+            User currentUser, 
+            IEnumerable<Scope> scopes, 
+            IActionRequiringEntityPermissions<Package> action, 
+            Package package, 
+            params string[] requestedActions)
+        {
+            return Evaluate(currentUser, scopes, action, package, (p) => p.PackageRegistration.Id, requestedActions);
+        }
```
chenriksson: Do you need the `IActionRequiringEntityPermissions<Package>` overloads? Or can callers just pass the registration?

```
var apiScopeEvaluationResult = EvaluateApiScope(
    ActionsRequiringPermissions.VerifyPackage,
    package.PackageRegistration,
    requestedActions);
```
and 
```
var apiScopeEvaluationResult = EvaluateApiScope(
    ActionsRequiringPermissions.UnlistOrRelistPackage,
    package.PackageRegistration,
    NuGetScopes.PackageUnlist);
```

src/NuGetGallery/Authentication/IApiScopeEvaluator.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// The <see cref="User"/> specified by the <see cref="Scope.OwnerKey"/> of the <see cref="Scope"/> that evaluated with <see cref="ApiScopeEvaluationResult.Success"/>.
+        /// If no <see cref="Scope"/>s evaluate to <see cref="ApiScopeEvaluationResult.Success"/>, this will be null.
+        /// </param>
+        /// <returns>A <see cref="ApiScopeEvaluationResult"/> that describes the evaluation of the <see cref="Scope"/>s.</returns>
+        ApiScopeEvaluationResult Evaluate(
+            User currentUser,
+            IEnumerable<Scope> scopes,
+            IActionRequiringEntityPermissions<Package> action,
+            Package package,
+            params string[] requestedActions);
```
chenriksson: see comment above - do you need this overload?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -729,42 +729,70 @@ protected internal virtual Stream ReadPackageFromRequest()
...
+                TelemetryService.TrackPackagePushNamespaceConflictEvent(id, version.ToNormalizedString(), GetCurrentUser(), User.Identity);
+                return new HttpStatusCodeWithBodyResult(HttpStatusCode.Conflict, Strings.UploadPackage_IdNamespaceConflict);
             }
 
-            // Legacy V1 API key (no scopes).
-            return true;
+            return new HttpStatusCodeWithBodyResult(statusCodeOnFailure, Strings.ApiKeyNotAuthorized);
+        }
+
+        private ApiScopeEvaluationResult EvaluateApiScope(IActionRequiringEntityPermissions<Package> action, Package package, params string[] requestedActions)
```
chenriksson: see comment above - do you need this overload?

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            bool success = false;
+
+            // Check that all scopes provided have the same owner scope.
+            var ownerScopes = scopes.Select(s => s.OwnerKey);
+            var firstOwnerScope = ownerScopes.FirstOrDefault();
+            if (ownerScopes.Any(o => o != firstOwnerScope))
+            {
+                throw new ArgumentException("All scopes provided must have the same owner scope.");
+            }
```
chenriksson: Since this would be a bug, could we just do SingleOrDefault and let it throw the InvalidOperationException? I'm assuming the exception would go to AI so we could investigate.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: They could, but the original intention of the design was for it to be easy to call regardless of what object you have.

E.g. `ActionRequiringPackagePermissions` implements both `IActionRequiringEntityPermissions<Package>` and `IActionRequiringEntityPermissions<PackageRegistration>`.

Unfortunately, by making `IApiScopeEvaluator` only support specific types, it now requires a separate overload for each, which complicates that.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: I prefer it as is because it is very clear about the error and the expectations of the code. The error is not an `InvalidOperationException`, it is an `ArgumentException` from being given invalid arguments.

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: fixed. nice catch!

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
scottbommarito: Removed it as requested.

scottbommarito:
@skofman1 @chenriksson applied your feedback, please review again

src/NuGetGallery/Authentication/IApiScopeEvaluator.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Evaluates the whether or not an action is allowed given a set of <paramref name="scopes"/>, an <paramref name="action"/>, and the <paramref name="requestedActions"/>.
+        /// </summary>
+        /// <param name="currentUser">The current user attempting to do the action with the given <paramref name="scopes"/>.</param>
+        /// <param name="scopes">The scopes being evaluated.</param>
+        /// <param name="action">The action that the scopes being evaluated are checked for permission to do.</param>
+        /// <param name="context">The <see cref="ActionOnNewPackageContext"/> that the scopes being evaluated are checked for permission to <paramref name="action"/> on.</param>
+        /// <param name="owner">
```
loic-sharma: >owner [](start = 25, length = 5)

The xml docs don't seem right here. "owner" isn't a param and "requestedActions" has no docs.

scottbommarito: LOL forgot to fix this thanks

src/NuGetGallery/Authentication/ApiScopeEvaluator.cs
chenriksson: Can you double check this before merge?

scottbommarito: oh wait this was the comment on `ApiScopeEvaluator` and not `IApiScopeEvaluator`, my bad

PR #5250, opened 2018-01-08
Fix GetCurrentUser for unauthenticated scenarios

scottbommarito:
`GetCurrentUser` had a bug where if `self.Request.User` was non-null (e.g. with a default `IPrincipal` that was unauthenticated) it would throw a 401 error.


joelverhagen:
Unit test?

scottbommarito:
@joelverhagen added

PR #5257, opened 2018-01-10
[2FA] Add AAD login support for NuGet Gallery

shishirx34:
This PR adds AAD support to NuGet gallery. The new login will work just like the personal MSA. You can login with your AAD account and link it to an existing nuget.org account or create a new account. This is configured as a new provider, so that we can transition away from previous way of logging in. 
I will add new configuration values in deployment PR. This PR uses `Microsoft.Owin.Security.OpenIdConnect` middleware which can now login with a common endpoint that facilitates single sign on flow for Personal/Work microsoft accounts. 

Also, the code takes care of maintaining backwards compatibility with existing MSA accounts and translates the account information into older format. Also added tests for the new provider and some missing coverage in `AuthenticationService`.


src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -538,51 +546,47 @@ public virtual CredentialViewModel DescribeCredential(Credential credential)
...
-            var tenantIdClaim = result.Identity.FindFirst(tenantIdClaimType);
-
-            Authenticator auther;
-            string authenticationType = idClaim.Issuer;
-            if (!Authenticators.TryGetValue(idClaim.Issuer, out auther))
+            try
             {
-                foreach (var authenticator in Authenticators.Values)
+                var userInfo = author.GetAuthInformation(externalIdentity);
+                var emailSuffix = userInfo.Email == null ? String.Empty : (" <" + userInfo.Email + ">");
```
xavierdecoster: Very much a nit and pretty sure the gallery code base is inconsistent, but should we prefer `string` over `String`?

src/NuGetGallery/Authentication/Providers/AuthInformation.cs
```diff
@@ -0,0 +1,29 @@
...
+        public string Email { get; private set; }
+
+        public string TenantId { get; private set; }
+
+        public string AuthenticationType { get; private set; }
+
+        public AuthInformation() { }
+
+        public AuthInformation(string identifier, string name, string email, string authenticationType, string tenantId = null)
+        {
```
xavierdecoster: Do we need argument null checks here?

src/NuGetGallery/Authentication/Providers/AuthInformation.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public string Name { get; private set; }
+
+        public string Email { get; private set; }
+
+        public string TenantId { get; private set; }
+
+        public string AuthenticationType { get; private set; }
+
+        public AuthInformation() { }
```
xavierdecoster: Empty constructor and only read-only property access? Smells weird..

src/NuGetGallery/Authentication/Providers/Authenticator.cs
```diff
@@ -94,16 +95,32 @@ public virtual ActionResult Challenge(string redirectUrl)
...
+
+        /// <summary>
+        /// The providers will override this method to extract the user information
+        /// from the returned claims by the identity authentication.
+        /// </summary>
+        /// <param name="claimsIdentity">The claims identity returned by the identity</param>
+        /// <returns><see cref="AuthInformation"/></returns>
+        public virtual AuthInformation GetAuthInformation(ClaimsIdentity claimsIdentity)
+        {
+            return new AuthInformation();
```
xavierdecoster: Is there any use in returning this `new AuthInformation()`? Should this become an `abstract` method instead?

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -474,12 +473,21 @@ public virtual ActionResult Challenge(string providerName, string redirectUrl)
         public virtual CredentialViewModel DescribeCredential(Credential credential)
         {
             var kind = GetCredentialKind(credential.Type);
-            Authenticator auther = null;
+            Authenticator author = null;
```
chenriksson: I think `auther` was intentional... as in, the one doing the "auth".

Agree it's a bad name, but maybe `authenticator` is a better one?

src/NuGetGallery/Authentication/Providers/Authenticator.cs
```diff
@@ -94,16 +95,32 @@ public virtual ActionResult Challenge(string redirectUrl)
             return new HttpUnauthorizedResult();
         }
 
-        public virtual bool TryMapIssuerToAuthenticationType(string issuer, out string authenticationType)
+        /// <summary>
+        /// Override this method to provide confirmation on identity author.
+        /// </summary>
+        /// <param name="claimsIdentity">The claims identity returned by the identity</param>
+        /// <returns>Returns true if this provider is the author for the identity, false otherwise</returns>
+        public virtual bool IsAuthorForIdentity(ClaimsIdentity claimsIdentity)
```
chenriksson: `Author` and `Auther` are confusing me.

Can we rename `auther` to `authenticator`, and maybe change this signature to `bool IsProviderForIdentity(ClaimsIdentity)`?

src/NuGetGallery/Authentication/Providers/Authenticator.cs
```diff
@@ -94,16 +95,32 @@ public virtual ActionResult Challenge(string redirectUrl)
...
         {
-            if (string.Equals(issuer, BaseConfig.AuthenticationType, StringComparison.OrdinalIgnoreCase))
+            // If the issuer of the claims identity is same as that of the authentication type then this is the author.
+            Claim firstClaim = claimsIdentity?.Claims?.FirstOrDefault();
```
chenriksson: nit: use `var`

src/NuGetGallery/Authentication/Providers/Authenticator.cs
```diff
@@ -94,16 +95,32 @@ public virtual ActionResult Challenge(string redirectUrl)
...
+            if (firstClaim == null)
             {
-                authenticationType = BaseConfig.AuthenticationType;
-                return true;
+                return false;
             }
 
-            authenticationType = null;
-            return false;
+            return string.Equals(firstClaim.Issuer, BaseConfig.AuthenticationType, StringComparison.OrdinalIgnoreCase);
```
chenriksson: If `BaseConfig.AuthenticationType` were guaranteed to be non-null, we could remove the block above and just do:
```
string.Equals(firstClaim?.Issuer, BaseConfig.AuthenticationType, StringComparison.OrdinalIgnoreCase);
```
I don't think AuthenticationConfiguration enforces this though. :-(

src/NuGetGallery/Authentication/Providers/AzureActiveDirectory/AzureActiveDirectoryAuthenticator.cs
```diff
@@ -66,15 +68,9 @@ public override ActionResult Challenge(string redirectUrl)
...
+        public override AuthInformation GetAuthInformation(ClaimsIdentity claimsIdentity)
         {
-            if (string.Equals(issuer, Config.Issuer, StringComparison.OrdinalIgnoreCase))
-            {
-                authenticationType = Config.AuthenticationType;
-                return true;
-            }
-
-            return base.TryMapIssuerToAuthenticationType(issuer, out authenticationType);
+            return MSAIdentityExtractor.GetAuthInformation(claimsIdentity, DefaultAuthenticationType);
```
chenriksson: MSAIdentityExtractor is for both MSA and AAD?

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
```diff
@@ -0,0 +1,145 @@
...
+
+using System;
+using System.Security.Claims;
+using System.Web.Mvc;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.Owin.Security.OpenIdConnect;
+using NuGetGallery.Configuration;
+using Owin;
+
+namespace NuGetGallery.Authentication.Providers.CommonAuth
```
chenriksson: Is `CommonAuth` a known terminology?

How about namespace `Providers.OpenIdConnect` , and class name `OpenIdConnectAuthenticator`?

chenriksson: Since OpenIdConnect extends OAuth 2.0, could also consider whether some functionality belongs in an OAuth2Authenticator base class, in case we ever support OAuth2... or even just to conceptualize the relation between OAuth2 and OpenIdConnect protocols.

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
```diff
@@ -0,0 +1,145 @@
...
+            public const string Issuer = "iss";
+        }
+
+        public static class AuthenticationType
+        {
+            public const string MicrosoftAccount = "MicrosoftAccount";
+            public const string AzureActiveDirectory = "AzureActiveDirectory";
+        }
+
+        public static readonly string DefaultAuthenticationType = "CommonAuth";
```
chenriksson: Should we ever use this default? Or should it always be clarified as "MicrosoftAccount" or "AzureActiveDirectory" which both happen to use common auth?

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
```diff
@@ -0,0 +1,145 @@
...
+        public static readonly string PersonalMSATenant = "9188040d-6c67-4c5b-b112-36a304b66dad";
```
chenriksson: Looks like this is personal (non-tenant) token id? Does that imply MSA, or can still be AAD?

Just looking at:
https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols-oauth-code

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
```diff
@@ -0,0 +1,145 @@
...
+        protected override void AttachToOwinApp(IGalleryConfigurationService config, IAppBuilder app)
+        {
+            // Fetch site root from configuration
+            var siteRoot = config.Current.SiteRoot.TrimEnd('/') + "/";
+            
+            // We *always* require SSL for Authentication
+            if (siteRoot.StartsWith("http://", StringComparison.OrdinalIgnoreCase)) 
+            {
+                siteRoot = siteRoot.Replace("http://", "https://");
+            }
```
chenriksson: Is this necessary - don't we already have redirect configuration for this?

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            if (string.Equals(tenantId, PersonalMSATenant, StringComparison.OrdinalIgnoreCase))
+            {
+                authenticationType = AuthenticationType.MicrosoftAccount;
+
+                // The MSA v2 authentication identifier is returned as 32 character alphanumeric value(padded with 0 and -), 
+                // where as the existing Microsoft account identifiers are 16 character wide.
+                // For e.g old format: 0ae45d63e22e4a60, newer format: 00000000-0000-0000-0AE4-5D63-E22E4A60
+                // We need to format the values into the older format for backwards compatibility
+                identifier = idClaim.Value.Replace("-", "").Substring(16).ToLowerInvariant();
```
chenriksson: Do you know why they left pad with zeros? Just wondering if we'll ever need those digits... would it be useful to have 2 versions of the Id claim, for 16- and 32- characters?

src/NuGetGallery/Authentication/Providers/Utils/MSAIdentityExtractor.cs
```diff
@@ -0,0 +1,29 @@
...
+            var nameClaim = claimsIdentity.FindFirst(ClaimTypes.Name);
+            if (nameClaim == null)
+            {
+                throw new ArgumentException($"External Authentication is missing required claim: {ClaimTypes.Name}");
+            }
+
+            var emailClaim = claimsIdentity.FindFirst(ClaimTypes.Email);
+            return new AuthInformation(identifierClaim.Value, nameClaim.Value, emailClaim?.Value, authType, tenantId: null);
+        }
+    }
```
chenriksson: Is there anything MSA specific in here? Could this just be put into a generic `ClaimsExtensions` class?

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -349,11 +349,23 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
...
-                var name = result
-                    .ExternalIdentity
-                    .GetClaimOrDefault(ClaimTypes.Name);
+                try
+                {
+                    var userInfo = result.Authenticator.GetAuthInformation(result.ExternalIdentity);
+                    name = userInfo.Name;
+                    email = userInfo.Email;
+                }
+                catch (Exception)
```
chenriksson: Is there a more specific Exception type we expect to see? What's the scenario?
 
Should we add trace log?

src/NuGetGallery/Authentication/Providers/AuthInformation.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Authentication.Providers
+{
+    public class AuthInformation
```
skofman1: Please use full words (AuthenticationInformation)

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -538,51 +546,47 @@ public virtual CredentialViewModel DescribeCredential(Credential credential)
...
+            foreach (var authenticator in Authenticators.Values)
             {
-                _trace.Error("External Authentication is missing required claim: " + ClaimTypes.NameIdentifier);
-                return new AuthenticateExternalLoginResult();
+                if (authenticator.IsAuthorForIdentity(externalIdentity))
+                {
+                    author = authenticator;
+                    break;
+                }
             }
```
skofman1: you can retwrite this using FirstOrDefault

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: Yes, I just copied this over from previous place. I usually use `string`. Will fix it.

src/NuGetGallery/Authentication/Providers/AzureActiveDirectory/AzureActiveDirectoryAuthenticator.cs
shishirx34: Yes, for the previous schema MSA and AAD used the same claims to get values. This is an interim thing until we get rid of old providers.

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
skofman1: Please use Aythentication and not Auth everywhere

shishirx34: That's reasonable. I couldn't think of a better name to represent the provider. However, `OpenIdConnect` can be implemented by any provider, this is specific to the AzureAD. Will try to think of something better.

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
shishirx34: This is the tenant for Personal MSA accounts, under the v2 protocol it is returned as the tenant id for  Personal MSA accounts, same as the previous `external.MicrosoftAccount` credentials validated using the MSA flow.

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
shishirx34: This value is passed to the Azure AD to redirect with the claims after authentication. This is added as a query string and validated on the AD side before redirection, it needs to be `https`

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
```diff
@@ -0,0 +1,145 @@
...
+            else
+            {
+                authenticationType = AuthenticationType.AzureActiveDirectory;
+                identifier = idClaim.Value;
+            }
+
+            var nameClaim = claimsIdentity.FindFirst(V2Claims.Name);
+            if (nameClaim == null)
+            {
+                throw new ArgumentException($"External Authentication is missing required claim: '{V2Claims.Name}'");
```
skofman1: why do we require Name claim? Why would it be missing?

src/NuGetGallery/Authentication/Providers/Utils/MSAIdentityExtractor.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace NuGetGallery.Authentication.Providers.Utils
+{
+    public static class MSAIdentityExtractor
+    {
+        public static AuthInformation GetAuthInformation(ClaimsIdentity claimsIdentity, string authType)
+        {
+            var identifierClaim = claimsIdentity.FindFirst(ClaimTypes.NameIdentifier);
+            if (identifierClaim == null)
+            {
+                throw new ArgumentException("External Authentication is missing required claim: " + ClaimTypes.NameIdentifier);
```
skofman1: nit: the 2 exception strings are formatted differently

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -474,12 +473,21 @@ public virtual ActionResult Challenge(string providerName, string redirectUrl)
...
-            Authenticator auther = null;
+            Authenticator author = null;
 
             if (kind == CredentialKind.External)
             {
-                string providerName = credential.Type.Split('.')[1];
-                Authenticators.TryGetValue(providerName, out auther);
+                if (string.IsNullOrEmpty(credential.TenantId))
+                {
+                    string providerName = credential.Type.Split('.')[1];
```
cristinamanum: Will the providerName always be the second ?

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
```diff
@@ -0,0 +1,145 @@
...
+            if (issuer == null || tenant == null)
+            {
+                return false;
+            }
+
+            var expectedIssuer = string.Format(Authority, tenant.Value);
+            return string.Equals(issuer.Value, expectedIssuer, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public override AuthInformation GetAuthInformation(ClaimsIdentity claimsIdentity)
```
shishirx34: May be combine this and above method as `TryGet`

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: Yes.

src/NuGetGallery/Authentication/Providers/AuthInformation.cs
shishirx34: renamed to `IdentityInformation` which seems more accurate.

src/NuGetGallery/Authentication/Providers/AuthInformation.cs
shishirx34: Nope, validation can be done in the service where needed.

src/NuGetGallery/Authentication/Providers/Authenticator.cs
shishirx34: You are right, it is not enforced. I would rather have the check to prevent unseen issues later.

src/NuGetGallery/Authentication/Providers/Authenticator.cs
shishirx34: I agree, its better.

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
shishirx34: The credential type will always be 'MicrosoftAccount' or 'AzureActiveDirectory', it is useful for the provider name comparison for now. I will likely be replacing it with the mapping of provider => CredentialType in my subsequent PRs, will leave it as is for now.

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
shishirx34: removed the name claim, since it is not a necessity.

src/NuGetGallery/Authentication/Providers/Utils/MSAIdentityExtractor.cs
shishirx34: Nothing specific for MSA, renamed it to `ClaimsExtensions`, this will be removed eventually when we remove older providers.

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
shishirx34: I looked into this, and I think its better to throw the exception here, given that we want to disallow missing emails going forward. In subsequent PRs I will clean it up to throw a well defined exception that we can surface to the user as missing email.

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: Before this change, when we found a user trying to register a new account directly with MSA, we would show `email` we got from the auth and allow the user to edit. Going forward, we will not allow this to happen, however, for backwards compatibility(more so especially for tests) we get one or the other claim. I haven't yet modified those tests as we need to maintain backwards compatibility. Hence, for now we will swallow the exceptions. In future we will surface the error to the view model and show appropriate error to the user. No point in adding trace here yet.

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
shishirx34: I checked with `Identity` folks and it is safe to use the 16 character string for MSA accounts.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -474,12 +470,21 @@ public virtual ActionResult Challenge(string providerName, string redirectUrl)
...
         {
             var kind = GetCredentialKind(credential.Type);
-            Authenticator auther = null;
+            Authenticator authenticator = null;
 
             if (kind == CredentialKind.External)
             {
-                string providerName = credential.Type.Split('.')[1];
-                Authenticators.TryGetValue(providerName, out auther);
+                if (string.IsNullOrEmpty(credential.TenantId))
```
chenriksson: So, we'll switch over to the new APIs (OpenIdConnect) once the SQL is run to populate tenantId for existing credentials?

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -474,12 +470,21 @@ public virtual ActionResult Challenge(string providerName, string redirectUrl)
...
+                {
+                    string providerName = credential.Type.Split('.')[1];
+                    Authenticators.TryGetValue(providerName, out authenticator);
+                }
+                else
+                {
+                    authenticator = Authenticators
+                        .Values
+                        .First(provider => provider.Name.Equals(AzureActiveDirectoryV2Authenticator.DefaultAuthenticationType, StringComparison.OrdinalIgnoreCase));
```
chenriksson: Why not use `TryGetValue` here or `FirstOrDefault` like you do for identity lookup. Just wondering about consistency.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -538,51 +543,41 @@ public virtual CredentialViewModel DescribeCredential(Credential credential)
...
-            if (idClaim == null)
-            {
-                _trace.Error("External Authentication is missing required claim: " + ClaimTypes.NameIdentifier);
-                return new AuthenticateExternalLoginResult();
-            }
 
-            var nameClaim = result.Identity.FindFirst(ClaimTypes.Name);
-            if (nameClaim == null)
+            var externalIdentity = result.Identity;
+            Authenticator author = Authenticators
```
chenriksson: Would rename `author` => `authenticator`

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -687,13 +682,13 @@ private string FormatCredentialType(string credentialType)
 
         private string FormatExternalCredentialType(string externalType)
         {
-            Authenticator auther;
-            if (!Authenticators.TryGetValue(externalType, out auther))
+            Authenticator author;
```
chenriksson: Would rename `author` => `authenticator`

src/NuGetGallery/Authentication/Providers/Authenticator.cs
```diff
@@ -94,16 +95,32 @@ public virtual ActionResult Challenge(string redirectUrl)
...
+        }
+
+        /// <summary>
+        /// The providers will override this method to extract the user information
+        /// from the returned claims by the identity authentication.
+        /// </summary>
+        /// <param name="claimsIdentity">The claims identity returned by the identity</param>
+        /// <returns><see cref="IdentityInformation"/></returns>
+        public virtual IdentityInformation GetIdentityInformation(ClaimsIdentity claimsIdentity)
```
chenriksson: Why not leave this abstract? Do any children actually use this?

src/NuGetGallery/Authentication/Providers/CommonAuth/CommonAuthAuthenticator.cs
chenriksson: Maybe clarify in comment why this redirect is necessary for AAD, so nobody mistakenly removes it?

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2AuthenticatorConfiguration.cs
```diff
@@ -0,0 +1,57 @@
...
+        public AzureActiveDirectoryV2AuthenticatorConfiguration()
+        {
+            AuthenticationType = AzureActiveDirectoryV2Authenticator.DefaultAuthenticationType;
+        }
+
+        public override void ApplyToOwinSecurityOptions(AuthenticationOptions options)
+        {
+            base.ApplyToOwinSecurityOptions(options);
+
+            var opts = options as OpenIdConnectAuthenticationOptions;
```
chenriksson: nit: would rename to reflect more specific cast, like `openIdOptions`

tests/NuGetGallery.Facts/Authentication/Providers/CommonAuth/AzureActiveDirectoryV2AuthenticatorFacts.cs
```diff
@@ -0,0 +1,236 @@
...
+namespace NuGetGallery.Authentication.Providers.AzureActiveDirectoryV2
+{
+    public class AzureActiveDirectoryV2AuthenticatorFacts
+    {
+        public class TheGetUIMethod
+        {
+            [Fact]
+            public void GetUIReturnsCorrectValues()
+            {
+                //Arrange
```
chenriksson: nit: space after `//`

src/NuGetGallery/Authentication/Providers/Authenticator.cs
shishirx34: Yes, there are other children where we need default value.

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: Yes, and after we enable the config values for new provider.

PR #5271, opened 2018-01-11
Move validation issue messages into a gallery view to ease revisions

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/1105.
Progress on https://github.com/NuGet/Engineering/issues/785.

This updates to the latest build of NuGet.Services.Validation.Issues and friends. The main changes here are:

1. Two new issue types for the extract and validate job.
1. Move message test and HTML to gallery. Previously there was HTML embedded into a string. HTML and presentation of validation messages is a view concern that only gallery should care about.

To support this, I have added some testing that assert that `_ValidationIssue.cshtml` handles are issue types. This will allow us to catch issue types with no messages at CI time.

There is a slight change here is dedupes messages on their issue code and data now, instead of their final message. I wanted to keep the deduping logic out of the view so that it can more easily be tested. Today all messages use all data available to them so there is no functional difference.

The look and feel on the display package page is the same as before.

I will bump to a stable version of `ServerCommon` after this PR is approved.


joelverhagen:
image

src/NuGetGallery/Views/Packages/_ValidationIssue.cshtml
```diff
@@ -0,0 +1,37 @@
...
+                This signed package was rejected since it has @typedIssue.Count signatures. A signed package must have
+                exactly one signature.
+            </text>
+        }
+        break;
+    default:
+        <text>
+            Package validation failed due to an unknown error.
+        </text>
+        break;
```
chenriksson: Do we ever put view strings in the regx, or just controller messages?

tests/NuGetGallery.Facts/Framework/LooseMocksRegistrationSource.cs
```diff
@@ -16,7 +16,9 @@ internal class LooseMocksRegistrationSource
         : IRegistrationSource
     {
         private readonly bool _onlyInterfaces;
+#pragma warning disable 0618
         private readonly MockFactory _repository;
+#pragma warning restore 0618
```
chenriksson: This moves the `NoWarn` from csproj to the specific location in code that triggered the warning? What's the warning?

chenriksson: Wow, there's lots of these... do we plan to clean them up?

src/NuGetGallery/Views/Packages/_ValidationIssue.cshtml
joelverhagen: Not as far as I know. Just strings found in C# files.

tests/NuGetGallery.Facts/Framework/LooseMocksRegistrationSource.cs
joelverhagen: The warning is Obsolete. `CS0618`:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0618

No plan to clean them up. We could eventually delete the authors array which is all but one of these cases.

tests/NuGetGallery.Facts/Views/Packages/ValidationIssueFacts.cs
```diff
@@ -0,0 +1,118 @@
...
+                    .Trim();
+
+                _output.WriteLine($"HTML:");
+                _output.WriteLine(html);
+
+                return html;
+            }   
+        }
+
+        public static IEnumerable<ValidationIssue> KnownValidationIssues
```
loic-sharma: >KnownValidationIssues [](start = 51, length = 21)

Should we have a unit test test that uses reflection to ensure that all `ValidationIssueCode` are "known" or "unknown" (aka, purposefully not included in view)?

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -62,16 +64,20 @@ public IEnumerable<ValidationIssue> GetLatestValidationIssues(Package package)
...
+                        .OrderBy(i => i.Key)
+                        .Select(i => ValidationIssue.Deserialize(i.IssueCode, i.Data))
+                        .ToList();
                 }
             }
 
-            return Enumerable.Empty<ValidationIssue>();
+            // Filter out unknown issues and deduplicate the issues by code and data.
+            return issues
+                .GroupBy(x => new { x.IssueCode, Data = x.Serialize() })
```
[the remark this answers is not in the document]
joelverhagen: Why? If there is extraneous data or a non-breaking change in serialization (e.g. indented vs. non-indented) the thing won't be deduplicated. Additionally, this de-dupes all unknown issues to one (even cases where the persisted code is not Unknown).

tests/NuGetGallery.Facts/Views/Packages/ValidationIssueFacts.cs
joelverhagen: See `AllIssueCodesAreHandledTestData`

joelverhagen:
> I forgot to mention - this message should only be shown on vcs/unknown failures. This message should not be shown along side "actionable" issues, like package is signed.

Done.

src/NuGetGallery/Services/ValidationService.cs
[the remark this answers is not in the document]
loic-sharma: Ah, I see, that makes sense. This is non obvious behavior, I would recommend adding these details to a comment and possibly adding a unit test for this behavior.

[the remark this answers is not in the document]
joelverhagen: Already have unit test, But I'll add a comment.

PR #5275, opened 2018-01-16
SQL fix and addressing feedback from demo

chenriksson:
Includes the following, with screenshots:
- Update SQL to return and not migrate when no request match

- Add informational hint on transform page to indicate what is expected for administrator:

image

- Add informational message when prompting for new admin sign in:

image

- Adjust transformation error formatting to align with header:

image

//cc @anangaur , @karann-msft

### DIFF
--- a/src/NuGetGallery/Controllers/AuthenticationController.cs
+++ b/src/NuGetGallery/Controllers/AuthenticationController.cs
@@ -62,10 +62,11 @@ namespace NuGetGallery
         /// Sign In\Register view
         /// </summary>
         [HttpGet]
-        public virtual ActionResult LogOn(string returnUrl)
+        public virtual ActionResult LogOn(string returnUrl, string returnUrlMessage = "")
         {
             // I think it should be obvious why we don't want the current URL to be the return URL here ;)
             ViewData[Constants.ReturnUrlViewDataKey] = returnUrl;
+            ViewData[Constants.ReturnUrlMessageViewDataKey] = returnUrlMessage;
 
             if (Request.IsAuthenticated)
             {
--- a/src/NuGetGallery/Controllers/AuthenticationController.cs
+++ b/src/NuGetGallery/Controllers/AuthenticationController.cs
@@ -62,11 +62,15 @@ namespace NuGetGallery
         /// Sign In\Register view
         /// </summary>
         [HttpGet]
-        public virtual ActionResult LogOn(string returnUrl, string returnUrlMessage = "")
+        public virtual ActionResult LogOn(string returnUrl)
         {
             // I think it should be obvious why we don't want the current URL to be the return URL here ;)
             ViewData[Constants.ReturnUrlViewDataKey] = returnUrl;
-            ViewData[Constants.ReturnUrlMessageViewDataKey] = returnUrlMessage;
+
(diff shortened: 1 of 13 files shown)


src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -62,10 +62,11 @@ public partial class AuthenticationController
         /// Sign In\Register view
         /// </summary>
         [HttpGet]
-        public virtual ActionResult LogOn(string returnUrl)
+        public virtual ActionResult LogOn(string returnUrl, string returnUrlMessage = "")
```
chenriksson: Just realized I'm putting this in url; and should probably move to a header. Will update in a few minutes...

joelverhagen: Why not `TempData`? Seems a bit weird to encode an entire message in the URL.

joelverhagen: Actually, in light of `returnUrl` it kind of makes sense... I'm torn 😄

src/NuGetGallery/Views/Users/TransformFailed.cshtml
```diff
@@ -4,15 +4,15 @@
...
+    <div class="row col-sm-8">
+        @if (TempData.ContainsKey("TransformError"))
+        {
+            <h1>Failed to transform to an organization</h1>
             <hr />
-            @TempData["TransformError"]
-        </div>
-    }
+            <div>
+                @TempData["TransformError"]
```
joelverhagen: Why is this `TempData` and not part of a view model?

src/NuGetGallery/Controllers/AuthenticationController.cs
chenriksson: Agree it should not have been in URL. Fixed by moving to TempData.

src/NuGetGallery/Views/Users/TransformFailed.cshtml
chenriksson: Laziness... fixed by creating strong type viewModel for error message.

chenriksson:
@joelverhagen Addressed feedback. Also, reminder to review the screenshots... in case you haven't.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -153,9 +151,14 @@ public virtual ActionResult TransformToOrganization()
             
             await _userService.RequestTransformToOrganizationAccount(accountToTransform, adminUser);
 
-            // prompt for admin sign-on to confirm transformation
+
```
scottbommarito: nit: 2 lines whitespace here instead of one?

src/NuGetGallery/Strings.resx
```diff
@@ -734,6 +734,9 @@ For more information, please contact '{2}'.</value>
     <value>Administrator account '{0}' has not confirmed their email address.</value>
   </data>
   <data name="TransformAccount_RequestExists" xml:space="preserve">
-    <value>Another tranform request was created on {0} with organization admin '{1}'.</value>
+    <value>Another tranform request was created on {0} for administrator '{1}'. Note that a new request will overwrite this.</value>
```
shishirx34: ... this request.

PR #5278, opened 2018-01-17
Update copyright for NuGet Gallery

### LINKED ISSUE #5233: Update copyright year in NuGet gallery footer
The year in the footer should be updated to 2018!

image

shishirx34:
Added support to update date dynamically from the app for the formatted branding message. Sending config updates to NuGet Deployment.
Fixes #5233

### DIFF
--- a/src/NuGetGallery/Configuration/AppConfiguration.cs
+++ b/src/NuGetGallery/Configuration/AppConfiguration.cs
@@ -10,6 +10,8 @@ namespace NuGetGallery.Configuration
 {
     public class AppConfiguration : IAppConfiguration
     {
+        private string _ExternalBrandingMessage;
+
         [DefaultValue(Constants.DevelopmentEnvironment)]
         public string Environment { get; set; }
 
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Configuration/AppConfiguration.cs
```diff
@@ -279,7 +281,16 @@ public class AppConfiguration : IAppConfiguration
...
         /// </summary>
-        public string ExternalBrandingMessage { get; set; }
+        public string ExternalBrandingMessage {
+            get {
+                return _ExternalBrandingMessage;
+            }
+
+            set
+            {
+                _ExternalBrandingMessage = string.Format(value, DateTime.Now.Year);
```
joelverhagen: `DateTime.UtcNow` - better than guessing what timezone the server is running under

src/NuGetGallery/Configuration/AppConfiguration.cs
```diff
@@ -279,7 +281,16 @@ public class AppConfiguration : IAppConfiguration
         /// <summary>
         /// Gets/sets a string that is brand string to display in the footer
         /// </summary>
-        public string ExternalBrandingMessage { get; set; }
+        public string ExternalBrandingMessage {
```
joelverhagen: I personally feel that config classes should be as dumb as possible. Could we perhaps make this a view concern (do it in .cshtml)?

joelverhagen: Also, we should add XML docs stating that this accepts a single `{0}` string format token which is replaced with the UTC year.

shishirx34: There are several places where the branding string is being checked for existence and assignment. I personally feel doing it once at the source of the truth is perhaps better, so that if/when we drag this along to another view we do not miss it.

joelverhagen: Sounds fine!

PR #5280, opened 2018-01-17
Fix issue: [DEV] Images missing from Dev environment

cristinamanum:
Due to null ref exception in ClientTelemetryPIIProcessor.

### DIFF
--- a/src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
+++ b/src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
@@ -52,7 +52,15 @@ namespace NuGetGallery
 
         public virtual Route GetCurrentRoute()
         {
-            return RouteTable.Routes.GetRouteData(new HttpContextWrapper(HttpContext.Current)).Route as Route;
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Telemetry/ClientTelemetryPIIProcessor.cs
```diff
@@ -52,7 +52,12 @@ private void ModifyItem(ITelemetry item)
 
         public virtual Route GetCurrentRoute()
         {
-            return RouteTable.Routes.GetRouteData(new HttpContextWrapper(HttpContext.Current)).Route as Route;
+            if (HttpContext.Current == null)
+            {
+                return null;
```
chenriksson: ModifyItem returns w/o obfuscation when GetCurrentRoute returns null. Is this ok? Is HttpContext.Current always available for urls we need to obfuscate?

cristinamanum: The HttpContext should always  be available. From the current data it is ok to not obfuscate if the route is null.

PR #5291, opened 2018-01-18
Fix branding message formatting in functional test

shishirx34:


shishirx34:
The fix is already in the dev branch, which resulted in this test break.

PR #5292, opened 2018-01-18
[Organizations] Manage Organizations page

scottbommarito:
image

image

Note that:
- There is no button to add an organization. That is because this feature is not implemented yet.
- The edit link for each org is a placeholder for when the edit page is implemented.


src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -116,6 +116,7 @@
                                         <li role="presentation"><a href="@Url.AccountSettings()" role="menuitem">Account Settings</a></li>
                                         <li role="presentation"><a href="@Url.ManageMyApiKeys()" role="menuitem">API Keys</a></li>
                                         <li class="divider"></li>
+                                        <li role="presentation"><a href="@Url.ManageMyOrganizations()" role="menuitem">Manage Organizations</a></li>
```
scottbommarito: @anangaur  should it be

```
Manage Organizations
Manage Packages
```
or
```
Manage Packages
Manage Organizations
```

I feel like the latter would be better, because I expect that users are more likely to want to manage their packages than their organizations.

skofman1: On the other hand, if they have a bunch of packages (like Microsoft), they will never see the organization editing option

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -728,6 +728,11 @@ public static string ManageMyApiKeys(this UrlHelper url, bool relativeUrl = true
             return GetActionLink(url, "ApiKeys", "Users", relativeUrl);
         }
 
+        public static string ManageMyOrganizations(this UrlHelper url, bool relativeUrl = true)
```
chenriksson: ListOrganizations?

src/NuGetGallery/ViewModels/ListOrganizationsItemViewModel.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace NuGetGallery
+{
+    public class ListOrganizationsItemViewModel
+    {
+        public string Username { get; }
+        public string EmailAddress { get; }
+        public bool IsAdmin { get; }
+        public int MemberCount { get; }
+        public int PackagesCount { get; }
```
chenriksson: ViewModel props are a bit confusing since some refer to CurrentUser (`IsAdmin`) and others to the Organization.

The following is probably a little clearer:
```
class ListOrganizationsItemViewModel {
    string AccountName { get; }
    string EmailAddress { get; }
    int MembersCount { get; }
    int PackagesCount { get; }
    bool CurrentUserIsAdmin { get; }
}
```

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
chenriksson: Could we hide the menu item if user doesn't have any memberships? I think it's a better experience to hide complexity that's not relevant to the user.

src/NuGetGallery/Views/Users/Organizations.cshtml
```diff
@@ -0,0 +1,65 @@
...
+@{
+    ViewBag.Title = "Manage My Organizations";
+    Layout = "~/Views/Shared/Gallery/Layout.cshtml";
+}
+
+<section role="main" class="container main-container manage-packages">
+    <div class="row">
+        <div class="col-md-12">
+            <h1>Manage Organizations</h1>
+            @if (Model.Organizations.Any())
```
chenriksson: I think we should hide the menu item if the user has no memberships. If that's the case, I wonder if we should also not serve this page... maybe redirect to home and set temp message?

scottbommarito: I would rather show a page with a clear message than redirect to homepage with temp message. I agree that we should hide the option in the context menu however.

src/NuGetGallery/UrlExtensions.cs
scottbommarito: IMO `ManageOrganizations` or `ManageMyOrganizations` is clearer because it aligns with `Manage Packages` and the menu item.

chenriksson: Then I would rename the view models to match, i.e. `ManageOrganizationsViewModel`. Looks like Manage Packages and Manage Package Owners view model names are consistent with this.

scottbommarito: agreed

src/NuGetGallery/ViewModels/ListOrganizationsItemViewModel.cs
scottbommarito: done

src/NuGetGallery/UrlExtensions.cs
scottbommarito: done

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
scottbommarito: done!

PR #5293, opened 2018-01-19
Remove usernames and emails from AI traces.

cristinamanum:
Remove usernames and emails from AI traces.

### DIFF
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -755,7 +755,7 @@ namespace NuGetGallery.Authentication
                 else
                 {
                     // If multiple matches, leave it null to signal no unique email address
-                    _trace.Warning("Multiple user accounts with a single email address were found. Count: " + String.Join(", ", allMatches.Select(u => u.Username)));
+                    _trace.Warning($"Multiple user accounts with a single email address were found. Count: {allMatches.Count}");
                 }
             }
             return user;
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -756,7 +755,7 @@ private User FindByUserNameOrEmail(string userNameOrEmail)
                 else
                 {
                     // If multiple matches, leave it null to signal no unique email address
-                    _trace.Warning("Multiple user accounts with email address: " + userNameOrEmail + " found: " + String.Join(", ", allMatches.Select(u => u.Username)));
+                    _trace.Warning("Multiple user accounts with a single email address were found. Count: " + String.Join(", ", allMatches.Select(u => u.Username)));
```
loic-sharma: I'm guessing `String.Join(", ", allMatches.Select(u => u.Username))` should be `allMatches.Count()`?

chenriksson:
SecurePushSubscription may have `_diagnostics.Information` including username.

Client IP is ok in AI logs, right?

src/NuGetGallery/Authentication/AuthenticationService.cs
cristinamanum: Right, good catch. Fixed.

cristinamanum: @chenriksson - First comment - thank you - fixed. 
Second comment - plain Client IP is not ok - needs to be obfuscated.

cristinamanum: Verified the traces for cookie compliance and do not seem to be IPs in the messages.

Issue #4990, opened 2017-11-13
Organizations: Edit Organization view


chenriksson:
Includes:
- Update the 'Manage Organizations' page to show organizations
- Add 'Edit Organization' view, linked off 'Manage Organizations' to allow editing members

Out of scope: AAD integration

chenriksson:
Tracking as part of #5409

PR #5295, opened 2018-01-19
Create derived UserAccountViewModel for User (non-Organization) accounts

chenriksson:
Preparation for #4990.

Refactoring to create base AccountViewModel, to be used by both UserAccountViewModel and OrganizationAccountViewModel (coming w/ Manage Organization).

### DIFF
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -348,7 +348,7 @@ namespace NuGetGallery
         [Authorize]
         [HttpPost]
         [ValidateAntiForgeryToken]
-        public virtual async Task<ActionResult> ChangeEmailSubscription(AccountViewModel model)
+        public virtual async Task<ActionResult> ChangeEmailSubscription(UserAccountViewModel model)
         {
             var user = GetCurrentUser();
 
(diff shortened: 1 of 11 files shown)


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -22,6 +22,21 @@ namespace NuGetGallery
     public partial class UsersController
         : AppController
     {
+        public static class Actions
```
scottbommarito: I like this change but
1) It is necessary for this PR? I can't see a huge reason to do it right now, and it clutters up the actual changes you're making here.
2) Could we use these constants in `Route.cs` too?

chenriksson: Completely true, I just got sick of seeing so many duplicated strings. :-)

If I want to keep this, will split into separate PR.

scottbommarito: Feel free to go ahead and keep it--it isn't too much of a change. I was just wondering if there was a huge reason to do it.

chenriksson: Separate PR makes sense, especially if I want to extend it to Routes.cs like you suggest.

Issue #5317, opened 2018-01-22
[Bug Bash]: Transform account unconfirmed admin error formatted incorrectly


chenriksson:
Scenario: Transform / Error displayed if new organization Administrator is not confirmed:

`Administrator account '{0}' has not confirmed their email account.`

Issue #5311, opened 2018-01-22
[Organizations] Transform an account to an organization with same admin org account


cristinamanum:
Use user mitica1 to transform it to organization
In the transform page use as admin the same account name as the organization

Result: no error and the account is transformed to organization.

Issue #5309, opened 2018-01-22
[BugBash]: Transform error not displayed when adding Org as admin


chenriksson:
Scenario: Transform Account / error displayed if admin account is already an Organization

Issue #5308, opened 2018-01-22
[Organizations] Better error message on the account/transform


cristinamanum:
The current error message is:
_Failed to transform to an organization
Account 'cristinamanu' should not belong to any organizations._

The failure happened because the user is already a member of an organization but the message in not explanatory.

chenriksson:
@anangaur I'll need you to do a pass over error messages like this.

Please see:
https://github.com/NuGet/NuGetGallery/blob/dev/src/NuGetGallery/Strings.resx#L706

PR #5329, opened 2018-01-22
Bug bash: fix transform issues

chenriksson:
Fixes for the following:
- #5311: Transform should fail if admin is same as transformed account
- #5309: Transform should fail if admin is an organization
- #5317: Transform error not formatted correctly for unconfirmed admin

**Update 1/23**: Also incorporating feedback from @anangaur, including:
- #5308 and other transform error messages
- UI: Change admin info message to tooltip
- UI: Add transform confirmation
- UI: Formatting of sign-in returnUrl message

### DIFF
--- a/src/NuGetGallery/Services/UserService.cs
+++ b/src/NuGetGallery/Services/UserService.cs
@@ -201,8 +201,7 @@ namespace NuGetGallery
             }
             else if (accountToTransform.Organizations.Any() || accountToTransform.OrganizationRequests.Any())
             {
-                errorReason = String.Format(CultureInfo.CurrentCulture,
-                    Strings.TransformAccount_FailedReasonHasMemberships, accountToTransform.Username);
+                errorReason = Strings.TransformAccount_FailedReasonHasMemberships;
             }
             else if (enabledDomains == null ||
                 !enabledDomains.Contains(accountToTransform.ToMailAddress().Host, StringComparer.OrdinalIgnoreCase))
@@ -240,7 +239,7 @@ namespace NuGetGallery
             return errorReason == null;
         }
 
-            public async Task<bool> TransformUserToOrganization(User accountToTransform, User adminUser, string token)
+        public async Task<bool> TransformUserToOrganization(User accountToTransform, User adminUser, string token)
         {
             // todo: check for tenantId and add organization policy to enforce this (future work, with manage organization)
 
--- a/src/NuGetGallery/Services/UserService.cs
+++ b/src/NuGetGallery/Services/UserService.cs
@@ -214,7 +214,33 @@ namespace NuGetGallery
             return errorReason == null;
         }
 
-        public async Task<bool> TransformUserToOrganization(User accountToTransform, User adminUser, string token)
+        public bool CanTransformUserToOrganization(User accountToTransform, User adminUser, out string errorReason)
+        {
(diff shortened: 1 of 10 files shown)


src/NuGetGallery/Services/UserService.cs
```diff
@@ -214,7 +214,33 @@ public bool CanTransformUserToOrganization(User accountToTransform, out string e
...
+            else if (adminUser is Organization)
+            {
+                errorReason = String.Format(CultureInfo.CurrentCulture,
+                    Strings.TransformAccount_AdminAccountIsOrganization, adminUser.Username);
+            }
+
+            return errorReason == null;
+        }
+
+            public async Task<bool> TransformUserToOrganization(User accountToTransform, User adminUser, string token)
```
loic-sharma: Spacing seems off here

chenriksson: Fixed

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -151,7 +151,7 @@
     <div class="alert-transient @(ViewBag.HasJumbotron == true ? "alert-transient-jumbotron" : string.Empty)">
         <div class="container">
             <div class="row">
-                <div class="@ViewHelpers.GetColumnClasses(ViewBag)" role="alert" aria-live="assertive">
+                <div class="col-sm-12" role="alert" aria-live="assertive">
```
joelverhagen: Why this change?

joelverhagen: If I recall, not all pages take up all columns (i.e. some page content uses less than the 12 columns at which point we wanted the temp message to be that smaller width).

src/NuGetGallery/Views/Users/Transform.cshtml
```diff
@@ -60,4 +61,19 @@
...
     </div>
-</section>
\ No newline at end of file
+</section>
+
+@section bottomScripts {
+    <script type="text/javascript">
+        $(function ()
+        {
+           var confirmation = "@Html.Raw(transformConfirmation)";
```
joelverhagen: You can JSON encode the string instead of doing raw.

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
chenriksson: Per @anangaur 's feedback that the temp message should stretch the width of the header. We were noticing the temp message width vary by page, which seems consistent with what you're saying.

I can circle back w/ him to verify.

joelverhagen: Okay, if this is the case please verify we don't have dead code. I don't feel strongly about it but when we worked with @jonwchu we decided to make the transient message width match the page content width. I don't feel strongly.

src/NuGetGallery/Views/Users/Transform.cshtml
chenriksson: I did `@Html.Raw` to prevent encoding.

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
chenriksson: Reverting, will open issue and can sort out later at lower pri.

Issue #5325, opened 2018-01-22
Non-org, collaborator API key returns 400 when pushing new version of org package


joelverhagen:
400 is weird for an auth issue. Also, the message is confusing.

Package: NuGet.Common3
Owner: jver2 (org)
User: jver4 (collaborator in jver2)
API key is scoped to jver4.

API key:

![image](https://user-images.githubusercontent.com/94054/35242064-61e2fb8c-ff6d-11e7-8ba2-3fdd641d16f7.png)

```
Pushing package.nupkg to 'https://dev.nugettest.org/api/v2/package'...
  PUT https://dev.nugettest.org/api/v2/package/
  BadRequest https://dev.nugettest.org/api/v2/package/ 790ms
Response status code does not indicate success: 400 (The NuGet package is invalid. The error encountered was:'The package ID 'NuGet.Common3' is not available.'. Correct the error and try again.).
````

Issue #5324, opened 2018-01-22
nuget.exe push with collaborator API key is able to push new version of non-org package


joelverhagen:
user: jver4
org: jver2
API key: scopes to jver2

Pushing a new version of NuGet.Common2, which is only owned by jver4. This should fail. Instead it succeeded.

![image](https://user-images.githubusercontent.com/94054/35241783-79da20a4-ff6c-11e7-840a-abd86e3c2728.png)

```
Pushing package.nupkg to 'https://dev.nugettest.org/api/v2/package'...
  PUT https://dev.nugettest.org/api/v2/package/
  Created https://dev.nugettest.org/api/v2/package/ 13041ms
Your package was pushed.
```

joelverhagen:
Also, unlist is allowed when it should not be.

Issue #5323, opened 2018-01-22
nuget.exe push with collaborator API key cannot push a new version


joelverhagen:
user: jver4
org: jver2 
role: collaborator
API key: can push new versions to jver2

![image](https://user-images.githubusercontent.com/94054/35241660-f8ada62c-ff6b-11e7-8745-56a9a42f12d3.png)

```
Pushing package.nupkg to 'https://dev.nugettest.org/api/v2/package'...
  PUT https://dev.nugettest.org/api/v2/package/
  BadRequest https://dev.nugettest.org/api/v2/package/ 785ms
Response status code does not indicate success: 400 (The NuGet package is invalid. The error encountered was:'The package ID 'NuGet.Common3' is not available.'. Correct the error and try again.).
```

This should work. Also, if 409 should be returned if the ID conflicts.

joelverhagen:
Unlist works just fine.

```
nuget.common3 4.0.0-zlocal-52037 will be deleted from the 'https://dev.nugettest.org/api/v2/package'. Would you like to continue? (y/N) y
WARNING: Deleting nuget.common3 4.0.0-zlocal-52037 from the 'https://dev.nugettest.org/api/v2/package'.
  DELETE https://dev.nugettest.org/api/v2/package/nuget.common3/4.0.0-zlocal-52037
  OK https://dev.nugettest.org/api/v2/package/nuget.common3/4.0.0-zlocal-52037 453ms
nuget.common3 4.0.0-zlocal-52037 was deleted successfully.
```

Issue #5316, opened 2018-01-22
[Organizations] After beginning an upload, the "Owners" dropdown in the Verify form always shows current user


loic-sharma:
Logged in as "sharma.loic", a collaborator on Microsoft organization and site admin. Uploaded a new version of "Microsoft.AspNet.WebApi.Core", which only the Microsoft organization owns. The "Owner" field on the upload verification page incorrectly showed "sharma.loic" in text (it should've been "Microsoft"). Submitting the form fails. Reloading the page or manually changing the owner value to "Microsoft" allows the form to be submitted.

scottbommarito:
Changed title to reflect problem better.

PR #5330, opened 2018-01-22
[Organizations] Fix bugs with upload scenarios

scottbommarito:
Fixes to:

https://github.com/NuGet/NuGetGallery/issues/5314
https://github.com/NuGet/NuGetGallery/issues/5316
https://github.com/NuGet/NuGetGallery/issues/5323
https://github.com/NuGet/NuGetGallery/issues/5324
https://github.com/NuGet/NuGetGallery/issues/5325
https://github.com/NuGet/NuGetGallery/issues/5334

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -23,6 +23,7 @@ using NuGetGallery.Areas.Admin;
 using NuGetGallery.Areas.Admin.Models;
 using NuGetGallery.Areas.Admin.Services;
 using NuGetGallery.Auditing;
+using NuGetGallery.Authentication;
 using NuGetGallery.Configuration;
 using NuGetGallery.Configuration.SecretReader;
 using NuGetGallery.Cookies;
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -488,12 +488,7 @@ where VersionRange.Parse(d.VersionSpec).Satisfies(packageVersion)
         private PackageRegistration CreateOrGetPackageRegistration(User owner, User currentUser, PackageMetadata packageMetadata, bool isVerified)
         {
             var packageRegistration = FindPackageRegistrationById(packageMetadata.Id);
-
-            if (packageRegistration != null && ActionsRequiringPermissions.EditPackage.CheckPermissions(currentUser, currentUser, packageRegistration) != PermissionsCheckResult.Allowed)
```
scottbommarito: I took this check out because
1 - this is only ever called by the upload flow which checks the ownership of the package before this point
2 - there is not a corresponding check in this method for ownership of reserved namespaces when we are uploading a new id

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -241,6 +242,11 @@ protected override void Load(ContainerBuilder builder)
                 .As<IReadMeService>()
                 .InstancePerLifetimeScope();
 
+            builder.RegisterType<ApiScopeEvaluator>()
+                .AsSelf()
+                .As<IApiScopeEvaluator>()
+                .InstancePerLifetimeScope();
```
chenriksson: Guess this busted API push? Looks like we should retest these scenarios in the final bug bash as well.

scottbommarito: yup, i was seeing some really weird behavior in the bug bash that resolved with this

PR #5333, opened 2018-01-23
Rewrite functionality that gets latest package versions for package owners

xavierdecoster:
See [this comment](https://github.com/NuGet/NuGetGallery/issues/4782#issuecomment-359814380)

### DIFF
--- a/src/NuGetGallery/Services/PackageService.cs
+++ b/src/NuGetGallery/Services/PackageService.cs
@@ -271,7 +271,7 @@ namespace NuGetGallery
 
                 // Like DisplayPackage we should prefer to show you information from the latest stable version,
                 // but show you the latest version (potentially latest UNLISTED version) otherwise.
-                yield return packagesByRegistration.OrderByDescending(p => p.Version).First();
+                yield return packagesByRegistration.OrderByDescending(p => NuGetVersion.Parse(p.Version)).First();
             }
         }
 
--- a/src/NuGetGallery/Services/PackageService.cs
+++ b/src/NuGetGallery/Services/PackageService.cs
@@ -226,11 +226,11 @@ namespace NuGetGallery
             }
             else
             {
-                return GetLatestPackagesForOwners(packages.ToList());
+                return GetLatestPackageForEachRegistration(packages.ToList());
             }
         }
 
-        private IEnumerable<Package> GetLatestPackagesForOwners(IReadOnlyCollection<Package> packages)
+        private IEnumerable<Package> GetLatestPackageForEachRegistration(IReadOnlyCollection<Package> packages)
         {
             // This method uses First() and FirstOrDefault() instead of Single() or SingleOrDefault() to get the latest package, 
             // in case there are multiple latest due to concurrency issue
--- a/src/NuGetGallery/Services/PackageService.cs
+++ b/src/NuGetGallery/Services/PackageService.cs
@@ -216,123 +216,80 @@ namespace NuGetGallery
         /// </summary>
         public IEnumerable<Package> FindPackagesByAnyMatchingOwner(User user, bool includeUnlisted, bool includeVersions = false)
         {
-            // Like DisplayPackage we should prefer to show you information from the latest stable version,
-            // but show you the latest version (potentially latest UNLISTED version) otherwise.
-            
             var ownerKeys = user.Organizations.Select(org => org.OrganizationKey).ToList();
             ownerKeys.Insert(0, user.Key);
 
-            var mergedResults = new Dictionary<string, Package>(StringComparer.OrdinalIgnoreCase);
+            var packages = GetPackagesForOwners(ownerKeys, includeUnlisted);
             if (includeVersions)
             {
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -216,123 +216,80 @@ public virtual Package FindAbsoluteLatestPackageById(string id, int? semVerLevel
...
+                if (latest != null)
                 {
-                    latestPackage = latestPackagesById.First();
+                    yield return latest;
+                    continue;
                 }
-                else
+
+                // Return latest stable non-SemVer2 version if available
+                latest = packagesByRegistration.Where(p => p.IsLatestStable).FirstOrDefault();
```
joelverhagen: How could you have a latest stable non-SemVer2 but no latest stable SemVer2? SemVer2-compatible (i.e. SemVer 1.0.0 and SemVer 2.0.0) is a superset of non-SemVer2 (i.e. just SemVer 1.0.0)

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -216,123 +216,80 @@ public virtual Package FindAbsoluteLatestPackageById(string id, int? semVerLevel
...
+                if (latest != null)
                 {
-                    mergedResults.Add(latestPackage.PackageRegistration.Id, latestPackage);
+                    yield return latest;
+                    continue;
                 }
+
+                // Like DisplayPackage we should prefer to show you information from the latest stable version,
+                // but show you the latest version (potentially latest UNLISTED version) otherwise.
+                yield return packagesByRegistration.OrderByDescending(p => p.Version).First();
```
joelverhagen: Is `p.Version` a `string` or a `NuGetVersion`?

xavierdecoster: It's the `string Version` property on the `Package` entity. I sense this might need to be parsed as a `NuGetVersion` before ordering :)

joelverhagen: Haha, yup!

src/NuGetGallery/Services/PackageService.cs
xavierdecoster: It shouldn't happen, but just in case, by taking this "fallback" approach, the code is a bit more resilient against possible concurrency issues.

src/NuGetGallery/Services/PackageService.cs
xavierdecoster: Fixed. Good catch!

src/NuGetGallery/Services/PackageService.cs
joelverhagen: Alright, makes sense.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -216,123 +216,80 @@ public virtual Package FindAbsoluteLatestPackageById(string id, int? semVerLevel
...
             {
-                Package latestPackage;
-                if (includeUnlisted)
+                // Return latest stable SemVer2 version if available
+                var latest = packagesByRegistration.Where(p => p.IsLatestStableSemVer2).FirstOrDefault();
+                if (latest != null)
                 {
-                    latestPackage = latestPackagesById.First();
+                    yield return latest;
+                    continue;
```
cristinamanum: Why is the continue needed?

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -216,123 +216,80 @@ public virtual Package FindAbsoluteLatestPackageById(string id, int? semVerLevel
...
-
-                mergedResults[latestStablePackage.PackageRegistration.Id] = latestStablePackage;
-            }
-        }
-
-        /// <summary>
-        /// Merge latest packages by owner, including organization owners that the user belongs to.
-        /// </summary>
-        private void MergeLatestPackagesByOwner(List<int> ownerKeys, bool includeUnlisted, Dictionary<string, Package> mergedResults)
+        private IEnumerable<Package> GetLatestPackagesForOwners(IReadOnlyCollection<Package> packages)
```
cristinamanum: This method is not related with an user, would a a method name without be better?

src/NuGetGallery/Services/PackageService.cs
cristinamanum: I think we would need unit tests to verify the logic described in the issue.

xavierdecoster: The `continue` is needed to exit the current iteration in the `foreach` loop and return only a single package per id.

xavierdecoster: We have existing unit tests covering this (current) logic.

src/NuGetGallery/Services/PackageService.cs
xavierdecoster: Renamed. It used to depend on the owners list, but forgot to rename after removing that parameter.

Issue #5310, opened 2018-01-22
[BugBash] Layout and functionality issues with Manage packages page for Organizations


shishirx34:
Couple of things:
1. The drop down box should be placed a little higher on par with "My packages"
2. The filter should be applicable to all packages on the page not just 'listed' 'unlisted' sections
3. 'My Reserved namespaces' section seems misplaced on this page, perhaps it should be moved to accounts page based on @anangaur's feedback. @diverdan92 - for visibility.
4. 'My reserved namespaces' section doesn't show the reserved namespaces for the organization(I was a collaborator)
4. The 'Pending ownership request' doesn't show incoming requests for the organization account to the collaborators.

scottbommarito:
Regarding number 5: this is intentional, because collaborators cannot accept incoming requests for the organization account.

Note that it does not show incoming requests for the organization account to admins, which it probably should.

chenriksson:
Work remaining to populate organization data and allow filtering for namespaces and owner requests sections.

Issue #5307, opened 2018-01-22
Consider better visual separation of items in the owners column on my packages page


agr:
Check the following:

![image](https://user-images.githubusercontent.com/102933/35238687-1914fc98-ff63-11e7-85a9-28f67172d5f3.png)

Is it a list? Is there one item? Two? Three?
We probably need some better separation of accounts and grouping of elements that relate to the same account (organization icon is different color, has a space between itself and account name and is not part of the link to the organization).

Issue #5306, opened 2018-01-22
Weird account link underline on my packages page


agr:
On the [my packages page](https://dev.nugettest.org/account/Packages) the organization account link extends over to white space between accounts (at least in Firefox):
![image](https://user-images.githubusercontent.com/102933/35238607-cdc5b5b6-ff62-11e7-9633-f33bb42cc911.png)

Issue #5304, opened 2018-01-22
Manage packages package has flash of empty containers


joelverhagen:
![image](https://user-images.githubusercontent.com/94054/35238314-f1e0d684-ff61-11e7-8cda-c1bf1b6fd538.png)

Ideally you don't see skeletons as the page is loading.

PR #5335, opened 2018-01-23
Organization bug bash: Manage Packages page

chenriksson:
Fixes the following bug bash issues:
- #5299 comma separators for package and download counts
- #5304 flash of empty containers
- #5310 Layout issues (except binding filter to namespaces / requests will be separate PR)
- #5298 Org icon too high
- #5306 Owner link whitespace fix
- #5307 Visual separation between owners

Page Screenshot:

image

Package row w/ owners screenshot:

image

### DIFF
--- a/src/NuGetGallery/Views/Users/Packages.cshtml
+++ b/src/NuGetGallery/Views/Users/Packages.cshtml
@@ -4,92 +4,82 @@
     ViewBag.Title = "Manage My Package";
     ViewBag.Tab = "Packages";
     Layout = "~/Views/Shared/Gallery/Layout.cshtml";
+
+    var publishedPackagesCount = Model.ListedPackages.Count();
+    var unlistedPackagesCount = Model.UnlistedPackages.Count();
+    var namespacesCount = Model.ReservedNamespaces.ReservedNamespaces.Count();
+    var ownerRequestsIncomingCount = Model.OwnerRequests.Incoming.RequestItems.Count();
+    var ownerRequestsOutgoingCount = Model.OwnerRequests.Outgoing.RequestItems.Count();
 }
 
-<section role="main" class="container main-container page-manage-packages page-package-details">
+<section role="main" class="container main-container page-account-settings">
     <div class="row">
-        <div class="col-xs-12">
-                <h1>My Packages</h1>
-                <div class="col-xs-3 text-right form-group pull-right">
+        <div class="@ViewHelpers.GetColumnClasses(ViewBag)">
+            <div class="clearfix">
+                <div style="float: left">
+                    <h1 style="text-align: right">Manage Packages</h1>
+                </div>
+                <span class="form-section-data">
                     <select id="ownerFilter" class="form-control pull-right hidden"
                             data-bind="options: Owners, value: OwnerFilter, optionsText: 'Username'"></select>
-                </div>
-            @ViewHelpers.Section(this, "listed", "Published",
+                </span>
+            </div>
+
+            @ViewHelpers.Section(this, "listed", @<text>Published Packages</text>,
+                @<text>
+                    <span id="listed-data" data-bind="text: ListedPackages.VisiblePackagesHeading()"></span>
+                </text>,
                 @<text>
                     <div class="row user-package-list">
                         <div data-bind="template: { name: 'manage-packages', data: ListedPackages }"></div>
                     </div>
-                </text>)
+                </text>, expanded: false)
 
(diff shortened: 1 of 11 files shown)


src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -4,92 +4,82 @@
...
-                    @ViewHelpers.Section(this,
-                        "requests-incoming",
-                        "Incoming",
-                        @<text>
-                            @Html.Partial("_OwnerRequestsList", Model.OwnerRequests.Incoming)
-                        </text>)
-                }
+            @if (Model.OwnerRequests.Incoming.RequestItems.Any())
+            {
+                @ViewHelpers.Section(this, "requests-incoming",
```
scottbommarito: You could probably pretty easily make a helper function to reuse for both `Incoming` and `Outgoing`.

```
@helper OwnerRequestsSection(OwnerRequestListModel model, string name)
{
    @ViewHelpers.Section(this, "requests-" + name",
        @<text>Ownership Requests (@name)</text>,
        @<text>@Html.Raw(model.RequestItems.Count() + " request" + (ownerRequestsIncomingCount == 1 ? "" : "s"))</text>,
        ...
}
```

You could potentially add this into `_OwnerRequestsList` itself as well.

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -4,92 +4,80 @@
     ViewBag.Title = "Manage My Package";
     ViewBag.Tab = "Packages";
     Layout = "~/Views/Shared/Gallery/Layout.cshtml";
+    
+    var namespacesCount = Model.ReservedNamespaces.ReservedNamespaces.Count();
+    var ownerRequestsIncomingCount = Model.OwnerRequests.Incoming.RequestItems.Count();
```
scottbommarito: I really like the change from "Incoming" to "Received" and from "Outgoing" to "Sent", but ideally we should update the field names on the class to match this naming as well.

chenriksson: Fixed

src/NuGetGallery/Views/Users/Packages.cshtml
chenriksson: Will consider in next PR. For #5310 I need to update the namespaces and requests sections to also include organizations and support the account filter at the top.

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -114,14 +102,12 @@
...
-                                <a data-bind="attr: { href: ProfileUrl }">
-                                    <span data-bind="text: Username"></span>
-                                </a>
+                                <i class="ms-Icon ms-Icon--Group" aria-hidden="true" data-bind="visible: IsOrganization" style="vertical-align: -2px"></i>
+                                <a data-bind="attr: { href: ProfileUrl }"><span data-bind="text: Username"></span></a><!-- ko if: ($index() < ($parent.Owners.length - 1)) -->,&nbsp;<!-- /ko -->
                             </span>
                         </td>
                         <td class="align-middle text-nowrap">
-                            <span data-bind="text: DownloadCount"></span>
+                            <span data-bind="text: ko.unwrap(DownloadCount).toLocaleString()"></span>
```
joelverhagen: Seems like this should be a FormattedDownloadCount computed

joelverhagen:
The slash between the package count and the download count seems weird. A comma seems like a better fit.

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -114,14 +102,12 @@
                         </td>
                         <td class="align-middle">
                             <span class="ms-noWrap" data-bind="foreach: Owners">
-                                <i class="ms-Icon ms-Icon--Group" aria-hidden="true" data-bind="visible: IsOrganization"></i>
-                                <a data-bind="attr: { href: ProfileUrl }">
-                                    <span data-bind="text: Username"></span>
-                                </a>
+                                <i class="ms-Icon ms-Icon--Group" aria-hidden="true" data-bind="visible: IsOrganization" style="vertical-align: -2px"></i>
```
joelverhagen: Try to avoid inline css. Other icons are doing this.

chenriksson:
> The slash between the package count and the download count seems weird. A comma seems like a better fit.

@joelverhagen Missed this, but will incorporate in a follow-up PR

Issue #5318, opened 2018-01-22
[Bug Bash]: Manage Owners / 'your organization' displayed for non-member


chenriksson:
Scenario: Organization admins can accept package ownership on behalf of their Organization.

Repro:
- Logged in as 'chenriksgm' (non-member)
- Added 'microsoft' as owner of a package I own
- Saw 'your organization', when chenriksgm is not member

![image](https://user-images.githubusercontent.com/15364949/35240886-aeb2a722-ff69-11e7-971b-eaa0f034efcd.png)

Issue #5301, opened 2018-01-22
Manage package owners does an infinite redirect if you are a collaborator


joelverhagen:
E.g. https://dev.nugettest.org/packages/AjaxMin/ManagePackageOwners

![image](https://user-images.githubusercontent.com/94054/35237972-fc4fc6c6-ff60-11e7-96b5-85430a70517b.png)

PR #5336, opened 2018-01-23
[Organizations] Fix Manage Package Owners issues

scottbommarito:
- https://github.com/NuGet/NuGetGallery/issues/5301

We have configured our MVC so that if a 401 is returned by an endpoint, the app will redirect you to the sign in page.

If you are signed in when this happens, the sign-in page will redirect you back to the page you were initially on that gave you the 401. This leads to an infinite redirect loop.

To fix this, we can return 403 instead. Edit already does this.

We should potentially monitor our other endpoints to see if this is possible elsewhere--any endpoint that returns 401 when the user is logged in will have this behavior.

- https://github.com/NuGet/NuGetGallery/issues/5318

Simple fix that I added unit tests to.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1843,7 +1843,7 @@ namespace NuGetGallery
             return !String.Equals(posted.Replace("\r", ""), package.Replace("\r", ""), StringComparison.Ordinal);
         }
 
-        private HttpStatusCodeResult HttpForbidden()
+        private static HttpStatusCodeResult HttpForbidden()
         {
             return new HttpStatusCodeResult(HttpStatusCode.Forbidden, Strings.Unauthorized);
         }
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -987,7 +987,7 @@ namespace NuGetGallery
 
             if (ActionsRequiringPermissions.ManagePackageOwnership.CheckPermissionsOnBehalfOfAnyAccount(GetCurrentUser(), package) != PermissionsCheckResult.Allowed)
             {
-                return new HttpStatusCodeResult(401, "Unauthorized");
+                return HttpForbidden();
             }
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1842,5 +1842,10 @@ private static bool IsDifferent(string posted, string package)
             // Ignore those pesky '\r' characters which screw up comparisons.
             return !String.Equals(posted.Replace("\r", ""), package.Replace("\r", ""), StringComparison.Ordinal);
         }
+
+        private HttpStatusCodeResult HttpForbidden()
```
skofman1: static

src/NuGetGallery/ViewModels/PackageOwnersResultViewModel.cs
```diff
@@ -32,7 +32,7 @@ public PackageOwnersResultViewModel(User user, User currentUser, PackageRegistra
             ProfileUrl = url.User(user, relativeUrl: false);
             ImageUrl = GravatarHelper.Url(user.EmailAddress, size: Constants.GravatarImageSize);
             GrantsCurrentUserAccess = ActionsRequiringPermissions.ManagePackageOwnership.CheckPermissions(currentUser, user, packageRegistration) == PermissionsCheckResult.Allowed;
-            IsCurrentUserAdminOfOrganization = user is Organization && (user as Organization).Members.Where(m => m.IsAdmin).Any();
+            IsCurrentUserAdminOfOrganization = user is Organization org && org.Members.Where(m => m.Member.MatchesUser(currentUser) && m.IsAdmin).Any();
```
chenriksson: Add helper to simplify this line?
```
class UserExtensions {
    public static bool GetMembership(this Organization organization, User user) {
        organization.Members.FirstOrDefault(m => m.Member.Matches(user));
    }
}

PackageOwnersResultViewModel() {
    IsCurrentUserAdminOfOrganization = (account as Organization)?.GetMembership(currentUser)?.IsAdmin;
}
```

Issue #5281, opened 2018-01-17
AAD / Sign in flow change


chenriksson:
see Sign-In Flow change in [OneNote](https://microsoft.sharepoint.com/teams/NuGet/_layouts/OneNote.aspx?id=%2Fteams%2FNuGet%2FTeam%2FNugetServer%2FNugetServerTeamNote&wd=target%28Security.one%7CD36A0B17-E33C-43DE-B487-A40CAFAAA9E4%2FMicrosoft%20Logins%20%28MSA%5C%29%20and%20Nuget%20Gallery%7C4DD50740-0935-4235-8ADC-9F9A9CF013F1%2F%29)

Work includes:
- create new page for rejecting some logins

skofman1:
will be tested as part of https://github.com/NuGet/NuGetGallery/issues/5399

PR #5338, opened 2018-01-24
Update External Account Linking Flow

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5281

Sign-in Flow change
- Check if the email received links to existing account
    - Existing NuGet.org account
        - Found the matching MSA/AAD login - all good, login successful.
        - Found but has other external logins(MSA/AAD)? - Reject Slide 14
        - No external logins? - Link to Microsoft account - Slide 15
            - Disallow email change in link external account flow.
    - No accounts with specified email address.
        - Remove the "Create account page" - Slide 16
        - Show the create a new account page flow - Slide 17


src/NuGetGallery/Strings.resx
```diff
@@ -736,4 +736,13 @@ For more information, please contact '{2}'.</value>
   <data name="TransformAccount_RequestExists" xml:space="preserve">
     <value>Another tranform request was created on {0} with organization admin '{1}'.</value>
   </data>
+  <data name="LinkingMultipleExternalAccountsUnsupported" xml:space="preserve">
```
skofman1: LinkingMultipleExternalAccountsUnsupported, LinkingMultipleExternalAccountsUnsupportedFailure - the names are too similar. 
Can you rename to something clearer?

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -373,12 +382,37 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
...
+                var existingUserCanBeLinked = true;
+                string existingUserLinkingError = null;
+
+                if (existingUser != null && existingUser.Credentials.Any(c => c.IsExternal()))
+                {
+                    existingUserCanBeLinked = false;
+                    existingUserLinkingError = Strings.LinkingMultipleExternalAccountsUnsupported;
+                }
+
+                if (existingUser is Organization)
```
skofman1: else if

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -373,12 +382,37 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
...
                 }
+                
+                var existingUserCanBeLinked = true;
+                string existingUserLinkingError = null;
+
+                if (existingUser != null && existingUser.Credentials.Any(c => c.IsExternal()))
+                {
+                    existingUserCanBeLinked = false;
+                    existingUserLinkingError = Strings.LinkingMultipleExternalAccountsUnsupported;
+                }
```
skofman1: else if

src/NuGetGallery.Core/Extensions/CredentialExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGetGallery
+{
+    public static class CredentialExtensions
```
skofman1: There's a class called CredentialTypes where we have similar methods. Please put those there

src/NuGetGallery/Strings.resx
scottbommarito: couldn't come up with better names--they're the same issue, just one is a validation error for a form and one is a validation error for a page...any ideas?

scottbommarito:
@skofman1 You can actually test the entire MSA/AAD integration very easily by following some instructions in the OneNote. @shishirx34 showed them to me when I asked him.

src/NuGetGallery/Strings.resx
scottbommarito: `LinkingMultipleExternalAccountsUnsupported` -> `AccountIsLinkedToAnotherExternalAccount`
`LinkingMultipleExternalAccountsUnsupportedFailure` -> `LinkingMultipleExternalAccountsUnsupported`

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -373,12 +382,37 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
                 {
                     existingUser = _userService.FindByEmailAddress(email);
                 }
+                
+                var existingUserCanBeLinked = true;
+                string existingUserLinkingError = null;
+
+                if (existingUser != null && existingUser.Credentials.Any(c => c.IsExternal()))
```
chenriksson: Could save the null check to a local and reuse when initializing the viewModel below. Also consider wrapping the if checks below in the same block.

```
var existingFound = existingUser != null;
if (existingFound) {
    if (existingUser.Credentials.Any(...) { ... }
    if (existingUser is Organization) { ... }
}

var external = new AssociateExternalAccountViewModel {
    FoundExisting = foundExisting;
    ...
}
```

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -144,18 +144,22 @@ public virtual ActionResult SignUp(string returnUrl)
...
-            var user = authenticationResult.AuthenticatedUser;
+            var authenticatedUser = authenticationResult.AuthenticatedUser;
             
             if (linkingAccount)
             {
+                // Verify account has no other external accounts
+                if (authenticatedUser.User.Credentials.Any(c => c.IsExternal()))
+                {
+                    return SignInFailure(model, linkingAccount, Strings.LinkingMultipleExternalAccountsUnsupported);
+                }
```
chenriksson: Do you need this check here? Isn't this failure covered in `LinkExternalAccount` below?

Why 2 different error messages for these checks?

src/NuGetGallery/Views/Authentication/LinkExternal.cshtml
```diff
@@ -30,82 +18,37 @@
...
+                        We found an account with this email address. Sign in with your password to link with this Microsoft account.
+                        Please <a href="mailto:@Config.Current.GalleryOwner.Address">contact support</a> if you need more assistance.
+                        </text>
+
+                        @Html.Partial("_SignIn", Model);
                     }
                     else
                     {
-                        @:the @Model.External.ProviderAccountNoun '@Model.External.AccountName'.
+                        if (!string.IsNullOrEmpty(Model.External.ExistingUserLinkingError))
```
chenriksson: Don't need this check if you have ExistingUserCanBeLinked just return `string.IsNullOrEmpty(ExistingUserLinkingError)`.

src/NuGetGallery/Views/Authentication/_SignIn.cshtml
```diff
@@ -21,7 +21,15 @@
...
+        @if (!string.IsNullOrEmpty(Model.SignIn.UserNameOrEmail))
+        {
+            @Html.HiddenFor(m => m.SignIn.UserNameOrEmail)
+            <p>@Model.SignIn.UserNameOrEmail</p>
+        }
+        else
+        {
+            @Html.ShowTextBoxFor(m => m.SignIn.UserNameOrEmail)
+        }
         @Html.ShowValidationMessagesFor(m => m.SignIn.UserNameOrEmail)
```
chenriksson: `@Html.ShowValidationMessagesFor(...)` should go with `@Html.ShowTextBoxfor(...)`, right?

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
```diff
@@ -16,6 +16,7 @@
 using NuGetGallery.Infrastructure.Authentication;
 using Xunit;
 using System.Web;
+using System.Globalization;
```
chenriksson: nit: sort usings

src/NuGetGallery/Views/Authentication/_Register.cshtml
```diff
@@ -33,7 +33,15 @@
...
+        @if (!string.IsNullOrEmpty(Model.Register.EmailAddress))
+        {
+            @Html.HiddenFor(m => m.Register.EmailAddress)
+            <p>@Model.Register.EmailAddress</p>
+        }
+        else
+        {
+            @Html.ShowEmailBoxFor(m => m.Register.EmailAddress)
+        }
         @Html.ShowValidationMessagesFor(m => m.Register.EmailAddress)
```
chenriksson: `@Html.ShowValidationMessagesFor(...)` should go with `@Html.ShowEmailBoxfor(...)`, right?

scottbommarito: I don't think so--if MVC returns errors for the validation of the `EmailAddress` field, but the field is uneditable (with `ShowHiddenFor`), we still want to see the errors. The user seeing the errors can send the error to us so we can figure out the issue.

src/NuGetGallery/Views/Authentication/_SignIn.cshtml
scottbommarito: I don't think so--if MVC returns errors for the validation of the `UserNameOrEmail` field, but the field is uneditable (with `ShowHiddenFor`), we still want to see the errors. The user seeing the errors can send the error to us so we can figure out the issue.

src/NuGetGallery/Strings.resx
```diff
@@ -736,4 +736,13 @@ For more information, please contact '{2}'.</value>
   <data name="TransformAccount_RequestExists" xml:space="preserve">
     <value>Another tranform request was created on {0} with organization admin '{1}'.</value>
   </data>
+  <data name="AccountIsLinkedToAnotherExternalAccount" xml:space="preserve">
+    <value>We found an account with the email address associated with this external account ({0}), but linked to a different external account.</value>
+  </data>
+  <data name="LinkingMultipleExternalAccountsUnsupported" xml:space="preserve">
+    <value>You cannot link your NuGet.org account to multiple external accounts.</value>
```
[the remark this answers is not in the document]
scottbommarito: @anangaur was working on the messaging--not sure the status of it. I suggested doing something similar to him or at least a message explaining how to do so.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -373,12 +382,37 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
...
+                    existingUserLinkingError = Strings.LinkingOrganizationUnsupported;
+                }
+
+                if (existingUserLinkingError != null)
+                {
+                    existingUserLinkingError = string.Format(
+                        CultureInfo.CurrentCulture, 
+                        existingUserLinkingError, 
+                        email);
+                }
```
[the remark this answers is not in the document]
scottbommarito: When we implement more error messages with different format strings we can do that, but right now I don't see a reason to do so.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -373,12 +382,37 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
...
+                {
+                    existingUserCanBeLinked = false;
+                    existingUserLinkingError = Strings.AccountIsLinkedToAnotherExternalAccount;
+                }
+
+                if (existingUser is Organization)
+                {
+                    existingUserCanBeLinked = false;
+                    existingUserLinkingError = Strings.LinkingOrganizationUnsupported;
+                }
```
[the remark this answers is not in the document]
scottbommarito: done

src/NuGetGallery/Controllers/AuthenticationController.cs
scottbommarito: done

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -373,12 +382,37 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
                 {
                     existingUser = _userService.FindByEmailAddress(email);
                 }
+                
+                var existingUserCanBeLinked = true;
```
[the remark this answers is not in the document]
scottbommarito: done

src/NuGetGallery/Controllers/AuthenticationController.cs
scottbommarito: This is for the form POST, not the initial callback. There needs to be verification here in case somebody hacks the form to put the wrong value in.

The messages are basically the same, but adapted for the context in which they appear.

src/NuGetGallery/ViewModels/LogOnViewModel.cs
```diff
@@ -33,6 +33,13 @@ public class AssociateExternalAccountViewModel
         public string ProviderAccountNoun { get; set; }
         public string AccountName { get; set; }
         public bool FoundExistingUser { get; set; }
+        public bool ExistingUserCanBeLinked { get; set; }
```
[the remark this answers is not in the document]
scottbommarito: done

src/NuGetGallery/Views/Authentication/LinkExternal.cshtml
scottbommarito: done

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
scottbommarito: done

src/NuGetGallery.Core/Extensions/CredentialExtensions.cs
scottbommarito: moved

src/NuGetGallery/Controllers/AuthenticationController.cs
scottbommarito: done

src/NuGetGallery/Controllers/AuthenticationController.cs
scottbommarito: done

scottbommarito:
@skofman1 @chenriksson 

Addressed all feedback. Also tested with MSA/AAD (I had only tested with AAD2) and got all working.

src/NuGetGallery.Core/CredentialTypes.cs
```diff
@@ -69,13 +74,49 @@ public static bool IsSupportedCredential(this Credential credential)
...
+        {
+            if (actualType == null)
+            {
+                throw new ArgumentNullException(nameof(actualType));
+            }
+
+            return actualType.Equals(expectedType, StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static bool IsSubType(string actualType, string expectedTypePrefix)
```
skofman1: @scottbommarito , I understand you wanted to make the methods use shared code, but in this case, since it's only a sinlr line (credential.Type.StartsWith()) the result is overly complex. For example: IsPassword(Credential), instead of having a single line, now calls into IsPassword(type) => IsSubType() => string.StartsWith()
It's hard to read and not necessary.
Please undo this change.

src/NuGetGallery/Authentication/Providers/AzureActiveDirectory/AzureActiveDirectoryAuthenticator.cs
```diff
@@ -83,7 +83,10 @@ public override ActionResult Challenge(string redirectUrl)
 
         public override IdentityInformation GetIdentityInformation(ClaimsIdentity claimsIdentity)
         {
-            return ClaimsExtentions.GetIdentityInformation(claimsIdentity, DefaultAuthenticationType);
+            var identityInfo = ClaimsExtentions.GetIdentityInformation(claimsIdentity, DefaultAuthenticationType);
+            // The claims returned by AzureActiveDirectory have the email as the name claim are missing the email claim.
+            // Copy the object returned by the method but set the email as the name.
+            return new IdentityInformation(identityInfo.Identifier, identityInfo.Name, identityInfo.Name, identityInfo.AuthenticationType, identityInfo.TenantId);
```
chenriksson: nit: long line

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -374,11 +383,35 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
...
+                        existingUserLinkingError = Strings.LinkingOrganizationUnsupported;
+                    }
+                    else if (existingUser.Credentials.Any(c => c.IsExternal()))
+                    {
+                        existingUserLinkingError = Strings.AccountIsLinkedToAnotherExternalAccount;
+                    }
+
+                    if (existingUserLinkingError != null)
+                    {
+                        existingUserLinkingError = string.Format(
```
chenriksson: Would put the `string.Format` inline. This doesn't improve code, and inline would be more maintainable if the resource string args ever diverge.

src/NuGetGallery/Views/Authentication/_Register.cshtml
chenriksson: I disagree here - validation errors are for user input, right? Validation errors help users take corrective actions.

If EmailAddress is not editable and fails validation, wouldn't it indicate bad data on our side. I would not show a validation error in this case, but would log and direct the user to contact support.

chenriksson:
Please see my comment on validation, but otherwise changes look good to me.

src/NuGetGallery/Views/Authentication/_Register.cshtml
scottbommarito: It could indicate bad data (honestly more likely that the user hacked the form with F12), but given that

1) it would be nasty to have to predict from the form submission endpoint whether or not the error was on an uneditable field or not and then log the model errors
2) this would happen very rarely
3) we can debug the issue easier if the user experiencing it can tell us the error message
4) the user will likely be more annoyed if the form refuses to submit but there's no visible message

I would rather keep this like this.

Ultimately this isn't that big of a deal, which is part of why I'm suggesting this easy solution that allows any users that encounter it to approach us and doesn't require any nontrivial code changes.

src/NuGetGallery/Controllers/AuthenticationController.cs
scottbommarito: done

src/NuGetGallery/Authentication/Providers/AzureActiveDirectory/AzureActiveDirectoryAuthenticator.cs
```diff
@@ -83,7 +83,16 @@ public override ActionResult Challenge(string redirectUrl)
 
         public override IdentityInformation GetIdentityInformation(ClaimsIdentity claimsIdentity)
         {
-            return ClaimsExtentions.GetIdentityInformation(claimsIdentity, DefaultAuthenticationType);
+            var identityInfo = ClaimsExtentions.GetIdentityInformation(claimsIdentity, DefaultAuthenticationType);
+
+            // The claims returned by AzureActiveDirectory have the email as the name claim are missing the email claim.
+            // Copy the object returned by the method but set the email as the name.
+            return new IdentityInformation(
```
shishirx34: I think the correct fix here would be to extract the name claim correctly from the `claimsIdentity` Otherwise the identity here would be `email <email>`

src/NuGetGallery/Strings.resx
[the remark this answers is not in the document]
shishirx34: Please run through the flow for strings with @anangaur once before merging.

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
```diff
@@ -243,6 +243,134 @@ public class TheSignInAction : TestContainer
                 GetMock<AuthenticationService>().VerifyAll();
             }
 
+            public async Task WhenAttemptingToLinkExternalToExistingAccountWithNoExternalAccounts_AllowsLinking()
```
shishirx34: Also add the test case for register? when the email doesn't match any account?

src/NuGetGallery/Authentication/Providers/AzureActiveDirectory/AzureActiveDirectoryAuthenticator.cs
scottbommarito: Fixed it. All of the AAD on dev (and probably also int and prod as well) have incorrect identities as well.

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
scottbommarito: There were existing tests for that flow--I didn't change the behavior there at all.

src/NuGetGallery.Core/CredentialTypes.cs
scottbommarito: I initially agreed with you and was going to make this change, but then I realized that despite the fact that it does create deeper call stacks, it also substantially reduces the code because there is no longer the need for every method to have an `if (type == null) throw new ArgumentNullException` in addition to the fact that it unifies all of these `StartsWith` and `Equals` calls.

@shishirx34's PR, that came after this one, also touches this code and creates about 20 lines for something that could be closer to 10 with this restructuring. (https://github.com/NuGet/NuGetGallery/pull/5355/files#diff-624219da9e2d31ff3446f6c5be30fe20R43)

shishirx34: I think the point shouldn't always be to reduce the number of lines if it makes the code unreadable. If few lines to a non-often changing code are repetitive its fine if they are duplicated. I agree with @skofman1 and feel like this is an unnecessary refactoring leading to hard to read code, please undo these changes and keep the code easy to read.

scottbommarito: changed to be one-liners that return false if credential or type is null

Issue #5322, opened 2018-01-22
Last horizontal rule on API keys page is ugly


joelverhagen:
![image](https://user-images.githubusercontent.com/94054/35241413-46e155ec-ff6b-11e7-81fd-fd73e4471fd2.png)

The last one should not be shown

Issue #5320, opened 2018-01-22
API key package has pointless nested container


joelverhagen:
![image](https://user-images.githubusercontent.com/94054/35240964-e423fa3c-ff69-11e7-8ae8-30b512fa8656.png)

Border inside a border? Why?

Issue #5315, opened 2018-01-22
Edit organization's API key lists own packages instead of org packages


agr:
Repro:

1. Create an API key for an organization
2. Reload api key page.
2. Click the "Edit" link for a newly created API key
3. Notice, that the packages listed are that account's own packages, not an organization packages.

Packages list should be org's packages, not own.

Setting the checkboxes and saving, updated the key with access to own packages.

Issue #5313, opened 2018-01-22
Partial form validation on API keys page


agr:
Repro steps:
1. Go to [API keys](https://dev.nugettest.org/account/ApiKeys) page as a collaborator on Microsoft org;
2. Switch owner to Microsoft org;
3. Notice that "Either a glob pattern must be specified or at least one package ID must be selected" validation error popped up under the "Select packages".
4. Notice no other validation errors;
5. Check and then uncheck "Push" checkbox;
6. Notice that there are two validation errors now (the new one is "At least one scope must be selected" under the "Select Scopes").

Validation error reporting should be consistent: either no, or both errors should show up.

PR #5341, opened 2018-01-24
Fix ApiKeys page issues from bug bash

chenriksson:
Fixes the following issues:
- #5170: JS exception from ApiKeys page (edit)
- #5320: pointless nested container
- #5322: horizontal rule is ugly
- #5313: partial validation
- #5315: Edit doesn't show org packages

Screenshot of no nested container, partial validation gone:
image

Screenshot of removed bottom border at end:
image

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-api-keys.js
+++ b/src/NuGetGallery/Scripts/gallery/page-api-keys.js
@@ -101,8 +101,7 @@
                             return owner.Owner.toUpperCase() == data.Owner.toUpperCase()
                         });
                     this.PackageOwner(existingOwner);
-                }
-                else {
+                } else {
                     this.PackageOwner(this.PackageOwners[0]);
                 }
             };
--- a/src/NuGetGallery/Scripts/gallery/page-api-keys.js
+++ b/src/NuGetGallery/Scripts/gallery/page-api-keys.js
@@ -92,7 +92,19 @@
                 this.Scopes(data.Scopes || []);
                 this.Packages(data.Packages || []);
                 this.GlobPattern(data.GlobPattern || "");
-                this._SetPackageSelection(this.Packages())
+                this._SetPackageSelection(this.Packages());
+
+                if (this.Owner()) {
+                    var existingOwner = ko.utils.arrayFirst(
+                        this.PackageOwners,
+                        function (owner) {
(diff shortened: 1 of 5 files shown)


src/NuGetGallery/Scripts/gallery/page-api-keys.js
```diff
@@ -92,7 +92,19 @@
...
+                this._SetPackageSelection(this.Packages());
+
+                if (this.Owner()) {
+                    var existingOwner = ko.utils.arrayFirst(
+                        this.PackageOwners,
+                        function (owner) {
+                            return owner.Owner.toUpperCase() == data.Owner.toUpperCase()
+                        });
+                    this.PackageOwner(existingOwner);
+                }
```
scottbommarito: nit: `} else {`

src/NuGetGallery/Views/Users/ApiKeys.cshtml
```diff
@@ -120,7 +118,7 @@
 </script>
 
 <script type="text/html" id="api-key-details">
-    <div class="api-key-details">
+    <div class="api-key-details" data-bind="css: { 'api-key-details-append': $index() > 0 }">
```
scottbommarito: There are selectors that allow you to do this in CSS without having to touch the knockout. Take a look at `page-statistics-most-downloaded.less` and https://stackoverflow.com/questions/5065766/css3-selector-question-for-all-but-first-select.

If reformatting this to use the selectors is too difficult, this is ok as is.

src/NuGetGallery/Views/Users/ApiKeys.cshtml
```diff
@@ -44,10 +44,8 @@
                     "create",
                     "Create",
                     @<text>
-                        <div class="panel panel-default upsert-api-key">
-                            <div class="panel-body"
-                                 data-bind="template: { name: 'upsert-api-key', data: NewApiKey }">
-                            </div>
+                        <div class="upsert-api-key">
```
shishirx34: unrelated to PR, what is an 'upsert'? I see that class all over the place.

chenriksson: Apparently folks much younger than I invented the [word](https://www.urbandictionary.com/define.php?term=upsert) to mean update or insert. :-)

Issue #5340, opened 2018-01-24
Profile page show "Edit", "Manage owners", "Delete" links for not signed in


skofman1:
![image](https://user-images.githubusercontent.com/16807822/35351338-6d897b94-00f5-11e8-885f-59d5140997e3.png)

PR #5342, opened 2018-01-24
Profile page should only show "Edit", "Manage Package Owners", and "Delete" links when the current user has permission

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5340

Was accidentally passing in the owner of the page instead of the current user.

### DIFF
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -208,27 +208,27 @@ namespace NuGetGallery
         [Authorize]
         public virtual ActionResult DeleteRequest()
         {
-            var user = GetCurrentUser();
+            var currentUser = GetCurrentUser();
 
-            if (user == null || user.IsDeleted)
+            if (currentUser == null || currentUser.IsDeleted)
             {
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -606,7 +608,7 @@ public virtual ActionResult Profiles(string username, int page = 1)
 
             var packages = _packageService.FindPackagesByAnyMatchingOwner(user, includeUnlisted: false)
                 .OrderByDescending(p => p.PackageRegistration.DownloadCount)
-                .Select(p => new ListPackageItemViewModel(p, user)
+                .Select(p => new ListPackageItemViewModel(p, currentUser)
```
chenriksson: Maybe add a test for this?

scottbommarito: There aren't any existing tests for the endpoint so I wasn't sure if I had the time to invest in adding a whole suite but I think I'll add a couple tests?

scottbommarito: added

PR #5343, opened 2018-01-24
Add messages for new validation issue codes

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/785 and https://github.com/NuGet/Engineering/issues/1047.

The look n' feel of the message is exactly like the existing issues:
https://github.com/NuGet/NuGetGallery/pull/5271#issuecomment-357089131


joelverhagen:
I have sent an email to @rido-min, @karann-msft, and @anangaur concerning the specific message text.

PR #5349, opened 2018-01-25
Do not show organization packages on profile

### LINKED ISSUE #5328: Profile page of organization's collaborator shows organization's packages
See this profile: https://dev.nugettest.org/profiles/sharma.loic

The account "sharma.loic" is a collaborator of the "Microsoft" organization. Packages owned by Microsoft are appearing on "sharma.loic"'s profile page.

chenriksson:
Fixes #5328

### DIFF
--- a/tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
+++ b/tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
@@ -901,43 +901,178 @@ namespace NuGetGallery
             }
         }
 
-        public class TheFindPackagesByAnyMatchingOwnerMethod : TestContainer
+        public class TheFindPackagesByOwnerMethod : TheFindPackagesByOwnersMethodsBase
         {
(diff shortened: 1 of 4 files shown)


tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -901,43 +901,178 @@ public void ReturnsTheMostRecentVersionWhenSemVerLevel2()
...
+
+            protected IPackageService PackageService
+            {
+                get
+                {
+                    return GetService<PackageService>();
                 }
             }
 
             [Theory]
```
scottbommarito: Are these supposed to have `[Theory]` on them if they're not meant to be run directly, but instead by the subclasses?

chenriksson: Yes - verified that the test cases run for both subclasses. Unfortunately `MemberData` requires statics, so I need to override and apply that attribute in the subclasses... couldn't find a way around that.

Issue #5312, opened 2018-01-22
Owner dropdown appears upload UI when package ID already exists


joelverhagen:
![image](https://user-images.githubusercontent.com/94054/35239833-6d915fb6-ff66-11e7-9612-5446599184a3.png)

This makes no sense since the owner set is already fixed.

scottbommarito:
@anangaur This is more of an experience bug than a code bug. What do you suggest we show for owner in this scenario? Should we show the dropdown at all?

anangaur:
No dropdown. Just a list of owners should be fine. Not showing the owners here is fine too.

PR #5351, opened 2018-01-25
Only show "Owner" dropdown in upload flow for new packages

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5312

image

(same behavior as before for new package IDs)
image

### DIFF
--- a/src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
+++ b/src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
@@ -12,7 +12,7 @@ namespace NuGetGallery
     {
         public VerifyPackageRequest() { }
 
-        public VerifyPackageRequest(PackageMetadata packageMetadata, IEnumerable<User> possibleOwners)
+        public VerifyPackageRequest(PackageMetadata packageMetadata, IEnumerable<User> possibleOwners, PackageRegistration existingPackageRegistration)
         {
             Id = packageMetadata.Id;
             Version = packageMetadata.Version.ToFullStringSafe();
@@ -39,12 +39,16 @@ namespace NuGetGallery
             Summary = packageMetadata.Summary;
             Tags = PackageHelper.ParseTags(packageMetadata.Tags);
             Title = packageMetadata.Title;
+            IsNewId = existingPackageRegistration == null;
+            if (!IsNewId)
+            {
+                ExistingOwners = string.Join(", ", ParseUserList(existingPackageRegistration.Owners));
+            }
 
             // Editable server-state
             Listed = true;
             Edit = new EditPackageVersionReadMeRequest();
-
-            PossibleOwners = possibleOwners.Select(u => u.Username).ToList();
+            PossibleOwners = ParseUserList(possibleOwners);
         }
 
         public string Id { get; set; }
@@ -60,9 +64,21 @@ namespace NuGetGallery
         public string OriginalVersion { get; set; }
 
         /// <summary>
+        /// This is a new ID.
+        /// There are no existing packages with this ID.
+        /// </summary>
+        public bool IsNewId { get; set; }
+
+        /// <summary>
         /// The username of the <see cref="User"/> to upload the package as.
         /// </summary>
         public string Owner { get; set; }
+
+        /// <summary>
+        /// The <see cref="User"/>s that own the existing package registration that this package will be added to in a string.
+        /// E.g. "alice, bob, chad".
+        /// </summary>
+        public string ExistingOwners { get; set; }
 
         /// <summary>
         /// The usernames of the <see cref="User"/>s that the current user can upload the package as.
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
```diff
@@ -94,5 +110,10 @@ public VerifyPackageRequest(PackageMetadata packageMetadata, IEnumerable<User> p
         public string Summary { get; set; }
         public string Tags { get; set; }
         public string Title { get; set; }
+
+        private static IReadOnlyCollection<string> ParseUserList(IEnumerable<User> users)
+        {
+            return users.Select(u => u.Username).ToList();
+        }
```
chenriksson: Why do we need all of this? Could you just verify in the action method that there's at least 1 matching owner, and return error message instead? This would simplify the code.

scottbommarito: The user MUST be able to specify the owner of the package because we cannot arbitrarily determine who should be the owner of the package.

The form needs to support 3 separate states:

1) For new versions of an existing package, an uneditable owners section with the list of existing owners
2) For new IDs for which the current user has multiple options for owner, an editable owner section with all the users that the current user can upload new on behalf of
3) For new IDs for which only one user is a possible owner, an uneditable owner section that displays the only possible owner (a simplification of the previous option to prevent showing users a dropdown box with only one option)

I'm ok with simplifying the code to show nothing for the first case (make the "Owners" section of the form completely missing), but given that I've already implemented showing the list of users and it's a much better user experience to show them, I'd prefer to keep it.

chenriksson: I was referring to # 1, with not displaying existing owners. If you do display existing, you still don't need IsNewId... that can be determined by whether there are existing owners.

Agree for # 2 (are admin for 1+ organizations) that you need the ability to specify an owner. In this case, the only thing necessary is the possible owners.

chenriksson: I would only add a prop for the ExistingRegistration and keep it simple.

The view can get the owners from ExistingRegistration, and should probably be handling the formatting itself. For instance, if we want to change them to links or use the org icons.

scottbommarito: I was concerned about adding `ExistingRegistration` to the view because the code serializes it, which could potentially lead to some shenanigans. Correct me if I'm wrong, but won't passing in the entire `PackageRegistration` cause it to serialize not just the fields on the object but any objects that have a foreign key in the `PackageRegistration`, which could be very expensive given some IDs have thousands of versions? I think just passing in the necessary fields is much safer and more performant.

chenriksson: Looks like we are serializing the entire `VerifyPackageRequest` model for this page - per `@Html.Raw(Json.Encode(Model.InProgressUpload));` at bottom of UploadPackage.cshtml.

Some of our views use LINQ to only encode a subset of the model.

scottbommarito: Yes, but the size of a serialized `VerifyPackageRequest` is substantially smaller than a serialized `PackageRegistration` if it has a lot of packages because it will deserialize every object touched by that package as well. The `VerifyPackageRequest` only contains objects that are basic types or view models.

src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml
```diff
@@ -60,14 +60,19 @@
 
                 <!-- ko if: $data.PossibleOwners -->
                 <div class="verify-package-field form-group editable">
-                    <label for="Owner" class="verify-package-field-heading">Owner</label>
-                    <!-- ko if: $data.PossibleOwners.length > 1 -->
+                    <label for="Owner" class="verify-package-field-heading" data-bind="text: $data.IsNewId ? 'Owner' : 'Owners'">Owner</label>
```
chenriksson: Remove `>Owner</` now that we're data-binding the text?

scottbommarito: good catch haha

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
```diff
@@ -59,11 +63,23 @@ public VerifyPackageRequest(PackageMetadata packageMetadata, IEnumerable<User> p
         /// </summary>
         public string OriginalVersion { get; set; }
 
+        /// <summary>
+        /// This is a new ID.
+        /// There are no existing packages with this ID.
+        /// </summary>
+        public bool IsNewId { get; set; }
```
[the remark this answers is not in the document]
scottbommarito: I could probably remove it...I added it for consistency with the other fields.

PR #5352, opened 2018-01-25
[MSA] Add support for deprecating password logins - schema changes

shishirx34:
With the new changes coming in for AADv2, we will be deprecating user password logins. Since this will be a phased roll out, we need to mark certain accounts with deprecated passwords, for the accounts that will be migrating to the AADv2. 

Update:
Going forward, we will be deprecating password login for the users who have migrated (linked) account to MSA/AAD. These users will not be able to login with the password anymore. The UI flow in account settings, for Enabling/disabling passwords will be removed[note that this will only be for users whose password login has been deprecated by linking to MSA/AAD] and shown that these users have to use the Microsoft account to login. We cannot deprecate a user's password login until they have linked MSA/AAD account to the NuGet account. Once they have done it then we will block them from enabling/disabling the password login. This is as per @anangaur's specs. This will discourage regular users from using password login.


chenriksson:
@shishirx34 Deprecation is for @ microsoft.com? Could we use the same config whitelist as organization migration as opposed to a DB schema change?

shishirx34:
It will be for all the users who have linked their MSA/AAD accounts after the fact, not just the @microsoft.com accounts. We can't deprecate users who haven't linked their accounts. We need the schema change.

skofman1:
@shishirx34 , can you provide more details about the user experience we are implementing here?

scottbommarito:
Should we also add a column here to store a token for when a user whose password was deprecated tries to regain access to their account?

scottbommarito:
Spoke with @shishirx34 offline, it's not necessary.

shishirx34:
@skofman1 - added more details to the description.

shishirx34:
I had discussion with @anangaur and @scottbommarito, and we will be simplifying this some more. We don't really need this PR anymore. The way to go around would be to remove the password login completely for the users who change/link to the MSA/AAD account and from the account settings page, remove the node for password login. The way to recover a password login would be forgot password flow.

PR #5355, opened 2018-01-26
[MSA/AAD] Support for changing linked Microsoft accounts with MSA/AAD accounts.

shishirx34:
Lot of things/changes in here but all related.

- Adds support for allowing to link to a new MSA/AAD account from the account page with no linked 
Microsoft accounts.
- Allows changing of personal Microsoft account to any personal Microsoft account or AAD account.
- Does not allow changing AAD account to personal account. This flow will be enabled after the Organizations flow is in place.
- Upon linking/Changing Microsoft account removes the password login(part of deprecating password logins) in the backend.
- UI changes on Accounts page:
  - The account caption changed to 'Login Account'
  - Added support to link to Microsoft account.
  - Removed password login enable/disable flow from UI for 'disabled' passwords.
  - Updated password login node with the 'to be deprecated' message.
- Added tests for Authentication service and Users, Authentication controller changes.
- Shows appropriate errors/warnings/messages on user actions for MSA flows.
I will be adding deprecation warnings to 'forgot password' flow in a separate PR.

Screenshots:
- Link a new Microsoft account to NuGet.org login, password login enabled.
image

- Change Microsoft account for linked accounts, no password login(disabled).
image

### DIFF
--- a/src/NuGetGallery.Core/CredentialTypes.cs
+++ b/src/NuGetGallery.Core/CredentialTypes.cs
@@ -22,6 +22,7 @@ namespace NuGetGallery
             public const string Prefix = "apikey.";
             public const string V1 = Prefix + "v1";
             public const string V2 = Prefix + "v2";
+            public const string V3 = Prefix + "v3";
             public const string V4 = Prefix + "v4";
             public const string VerifyV1 = Prefix + "verify.v1";
         }
@@ -63,7 +64,16 @@ namespace NuGetGallery
             return type.Equals(ApiKey.VerifyV1, StringComparison.OrdinalIgnoreCase);
         }
         
-        internal static IReadOnlyList<string> SupportedCredentialTypes = new List<string> { Password.Sha1, Password.Pbkdf2, Password.V3, ApiKey.V1, ApiKey.V2, ApiKey.V4 };
+        internal static IReadOnlyList<string> SupportedCredentialTypes = new List<string>
+            {
+                Password.Sha1,
+                Password.Pbkdf2,
+                Password.V3,
+                ApiKey.V1,
+                ApiKey.V2,
+                ApiKey.V3,
+                ApiKey.V4
+            };
 
         /// <summary>
         /// Determines whether a credential is supported (internal or from the UI). For forward compatibility,
--- a/src/NuGetGallery.Core/CredentialTypes.cs
+++ b/src/NuGetGallery.Core/CredentialTypes.cs
@@ -27,6 +27,8 @@ namespace NuGetGallery
         }
 
         public const string ExternalPrefix = "external.";
+        public const string MicrosoftAccount = "MicrosoftAccount";
+        public const string AzureActiveDirectoryAccount = "AzureActiveDirectory";
 
         public static bool IsPassword(string type)
         {
@@ -36,6 +38,26 @@ namespace NuGetGallery
             }
 
             return type.StartsWith(Password.Prefix, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public static bool IsMicrosoftAccount(string type)
+        {
+            if (type == null)
+            {
+                throw new ArgumentNullException(nameof(type));
+            }
+
+            return type.Equals(ExternalPrefix + MicrosoftAccount, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public static bool IsAzureActiveDirectoryAccount(string type)
+        {
+            if (type == null)
+            {
+                throw new ArgumentNullException(nameof(type));
+            }
+
+            return type.Equals(ExternalPrefix + AzureActiveDirectoryAccount, StringComparison.OrdinalIgnoreCase);
         }
 
         public static bool IsApiKey(string type)
--- a/src/NuGetGallery.Core/CredentialTypes.cs
+++ b/src/NuGetGallery.Core/CredentialTypes.cs
@@ -34,46 +34,36 @@ namespace NuGetGallery
             public const string AzureActiveDirectoryAccount = Prefix + "AzureActiveDirectory";
         }
 
-        public static bool IsPassword(string type)
+        public static bool IsPassword(this Credential c)
         {
-            if (type == null)
-            {
-                throw new ArgumentNullException(nameof(type));
-            }
-
-            return type.StartsWith(Password.Prefix, StringComparison.OrdinalIgnoreCase);
+            return c?.Type?.StartsWith(Password.Prefix, StringComparison.OrdinalIgnoreCase) ?? false;
         }
 
-        public static bool IsMicrosoftAccount(string type)
+        public static bool IsExternal(this Credential c)
         {
-            return type.Equals(External.MicrosoftAccount, StringComparison.OrdinalIgnoreCase);
+            return c?.Type?.StartsWith(External.Prefix, StringComparison.OrdinalIgnoreCase) ?? false;
         }
 
-        public static bool IsAzureActiveDirectoryAccount(string type)
+        public static bool IsApiKey(this Credential c)
         {
-            return type.Equals(External.AzureActiveDirectoryAccount, StringComparison.OrdinalIgnoreCase);
+            return c?.Type?.StartsWith(ApiKey.Prefix, StringComparison.OrdinalIgnoreCase) ?? false;
         }
 
-        public static bool IsApiKey(string type)
+        public static bool IsType(this Credential c, string type)
         {
-            return type.StartsWith(ApiKey.Prefix, StringComparison.OrdinalIgnoreCase);
+            return c?.Type?.Equals(type, StringComparison.OrdinalIgnoreCase) ?? false;
         }
-
-        public static bool IsPackageVerificationApiKey(string type)
-        {
-            return type.Equals(ApiKey.VerifyV1, StringComparison.OrdinalIgnoreCase);
-        }
-        
+      
         internal static IReadOnlyList<string> SupportedCredentialTypes = new List<string>
-            {
-                Password.Sha1,
-                Password.Pbkdf2,
-                Password.V3,
-                ApiKey.V1,
-                ApiKey.V2,
-                ApiKey.V3,
-                ApiKey.V4
-            };
+        {
+            Password.Sha1,
+            Password.Pbkdf2,
+            Password.V3,
+            ApiKey.V1,
+            ApiKey.V2,
+            ApiKey.V3,
+            ApiKey.V4
+        };
 
         /// <summary>
         /// Determines whether a credential is supported (internal or from the UI). For forward compatibility,
@@ -94,13 +84,38 @@ namespace NuGetGallery
         /// <returns></returns>
         public static bool IsViewSupportedCredential(this Credential credential)
         {
-            return SupportedCredentialTypes.Any(credType => string.Compare(credential.Type, credType, StringComparison.OrdinalIgnoreCase) == 0)
-                    || credential.Type.StartsWith(External.Prefix, StringComparison.OrdinalIgnoreCase);
+            return
+                SupportedCredentialTypes.Any(credType => credential.IsType(credType)) ||
+                credential.IsExternal();
         }
 
         public static bool IsScopedApiKey(this Credential credential)
         {
             return IsApiKey(credential.Type) && credential.Scopes != null && credential.Scopes.Any();
+        }
+
+        public static bool IsPassword(string type)
+        {
+            return type?.StartsWith(Password.Prefix, StringComparison.OrdinalIgnoreCase) ?? false;
+        }
+
+        public static bool IsApiKey(string type)
+        {
+            return type?.StartsWith(ApiKey.Prefix, StringComparison.OrdinalIgnoreCase) ?? false;
+        }
+        public static bool IsMicrosoftAccount(string type)
+        {
+            return type?.Equals(External.MicrosoftAccount, StringComparison.OrdinalIgnoreCase) ?? false;
+        }
+
+        public static bool IsAzureActiveDirectoryAccount(string type)
+        {
+            return type?.Equals(External.AzureActiveDirectoryAccount, StringComparison.OrdinalIgnoreCase) ?? false;
+        }
+
+        public static bool IsPackageVerificationApiKey(string type)
+        {
+            return type?.Equals(ApiKey.VerifyV1, StringComparison.OrdinalIgnoreCase) ?? false;
         }
     }
 }
--- a/src/NuGetGallery.Core/CredentialTypes.cs
+++ b/src/NuGetGallery.Core/CredentialTypes.cs
@@ -26,9 +26,12 @@ namespace NuGetGallery
             public const string VerifyV1 = Prefix + "verify.v1";
         }
 
-        public const string ExternalPrefix = "external.";
-        public const string MicrosoftAccount = "MicrosoftAccount";
-        public const string AzureActiveDirectoryAccount = "AzureActiveDirectory";
+        public static class External
+        {
+            public const string Prefix = "external.";
+            public const string MicrosoftAccount = Prefix + "MicrosoftAccount";
+            public const string AzureActiveDirectoryAccount = Prefix + "AzureActiveDirectory";
+        }
 
         public static bool IsPassword(string type)
         {
@@ -42,22 +45,12 @@ namespace NuGetGallery
 
         public static bool IsMicrosoftAccount(string type)
         {
-            if (type == null)
-            {
-                throw new ArgumentNullException(nameof(type));
-            }
-
-            return type.Equals(ExternalPrefix + MicrosoftAccount, StringComparison.OrdinalIgnoreCase);
+            return type.Equals(External.MicrosoftAccount, StringComparison.OrdinalIgnoreCase);
         }
 
         public static bool IsAzureActiveDirectoryAccount(string type)
         {
-            if (type == null)
-            {
-                throw new ArgumentNullException(nameof(type));
-            }
-
-            return type.Equals(ExternalPrefix + AzureActiveDirectoryAccount, StringComparison.OrdinalIgnoreCase);
+            return type.Equals(External.AzureActiveDirectoryAccount, StringComparison.OrdinalIgnoreCase);
         }
 
         public static bool IsApiKey(string type)
@@ -92,7 +85,7 @@ namespace NuGetGallery
         public static bool IsViewSupportedCredential(this Credential credential)
         {
             return SupportedCredentialTypes.Any(credType => string.Compare(credential.Type, credType, StringComparison.OrdinalIgnoreCase) == 0)
-                    || credential.Type.StartsWith(ExternalPrefix, StringComparison.OrdinalIgnoreCase);
+                    || credential.Type.StartsWith(External.Prefix, StringComparison.OrdinalIgnoreCase);
         }
 
         public static bool IsScopedApiKey(this Credential credential)
--- a/src/NuGetGallery/App_Start/Routes.cs
+++ b/src/NuGetGallery/App_Start/Routes.cs
@@ -259,11 +259,6 @@ namespace NuGetGallery
                 new { controller = "Users", action = "RemoveCredential" });
 
             routes.MapRoute(
-                RouteName.ChangeExternalCredential,
-                "account/ChangeExternalCredetial/{credentialType}",
-                new { controller = "Users", action = "ChangeExternalCredential" });
-
-            routes.MapRoute(
                 RouteName.PasswordReset,
                 "account/forgotpassword/{username}/{token}",
                 new { controller = "Users", action = "ResetPassword", forgot = true },
--- a/src/NuGetGallery/App_Start/Routes.cs
+++ b/src/NuGetGallery/App_Start/Routes.cs
@@ -259,6 +259,11 @@ namespace NuGetGallery
                 new { controller = "Users", action = "RemoveCredential" });
 
             routes.MapRoute(
+                RouteName.ChangeExternalCredential,
+                "account/ChangeExternalCredetial/{credentialType}",
+                new { controller = "Users", action = "ChangeExternalCredential" });
+
+            routes.MapRoute(
                 RouteName.PasswordReset,
                 "account/forgotpassword/{username}/{token}",
                 new { controller = "Users", action = "ResetPassword", forgot = true },
@@ -275,7 +280,7 @@ namespace NuGetGallery
                 "account/confirm/{username}/{token}",
                 new { controller = "Users", action = "Confirm" },
                 new RouteExtensions.ObfuscatedMetadata(2, Obfuscator.DefaultTelemetryUserName));
-            
+
             routes.MapRoute(
                 RouteName.ChangeEmailSubscription,
                 "account/subscription/change",
--- a/src/NuGetGallery/App_Start/Routes.cs
+++ b/src/NuGetGallery/App_Start/Routes.cs
@@ -275,7 +275,7 @@ namespace NuGetGallery
                 "account/confirm/{username}/{token}",
                 new { controller = "Users", action = "Confirm" },
                 new RouteExtensions.ObfuscatedMetadata(2, Obfuscator.DefaultTelemetryUserName));
-
+            
             routes.MapRoute(
                 RouteName.ChangeEmailSubscription,
                 "account/subscription/change",
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -313,6 +313,7 @@ namespace NuGetGallery.Authentication
                 return false;
             }
 
+            // Check user credentials for existing cred for optimization to avoid expensive DB query
             if (UserHasCredential(user, credential) || FindMatchingCredential(credential) != null)
             {
                 // Existing credential for a registered account
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -305,6 +305,23 @@ namespace NuGetGallery.Authentication
             return ReplaceCredential(user, credential);
         }
 
+        public virtual async Task<bool> TryReplaceCredential(User user, Credential credential)
+        {
+            if (user == null || credential == null)
+            {
+                return false;
+            }
+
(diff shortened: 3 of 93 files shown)


src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -275,7 +275,7 @@ public static void RegisterUIRoutes(RouteCollection routes)
                 "account/confirm/{username}/{token}",
                 new { controller = "Users", action = "Confirm" },
                 new RouteExtensions.ObfuscatedMetadata(2, Obfuscator.DefaultTelemetryUserName));
-            
+
             routes.MapRoute(
                 RouteName.ChangeEmailSubscription,
                 "account/subscription/change",
```
skofman1: is this the only change to this file? if so, please undo.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -305,6 +305,23 @@ public virtual Task ReplaceCredential(string username, Credential credential)
...
         }
 
+        public virtual async Task<bool> TryReplaceCredential(User user, Credential credential)
+        {
+            if (user == null || credential == null)
+            {
+                return false;
+            }
+
+            if (FindMatchingCredential(credential) != null)
```
skofman1: This performs a DB call. Is it necessary? the user parameter should already have Credentials list loaded, you can use that for cheaper comparison.

src/NuGetGallery.Core/CredentialTypes.cs
```diff
@@ -38,6 +40,26 @@ public static bool IsPassword(string type)
             return type.StartsWith(Password.Prefix, StringComparison.OrdinalIgnoreCase);
         }
 
+        public static bool IsMicrosoftAccount(string type)
```
scottbommarito: When I merge in my changes you can reduce this code by a lot by using the methods I introduced.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -305,6 +305,23 @@ public virtual Task ReplaceCredential(string username, Credential credential)
...
+            {
+                // Existing credential for a registered account
+                return false;
+            }
+
+            await ReplaceCredential(user, credential);
```
scottbommarito: You need to `try-catch` this to prevent it from throwing exceptions.

src/NuGetGallery/Authentication/AuthenticationService.cs
scottbommarito: Maybe add a `FindMatchingCredentialForUser(user, credential)` that doesn't make DB calls?

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -622,14 +639,22 @@ public static ClaimsIdentity CreateIdentity(User user, string authenticationType
...
-            // Find the credentials we're replacing, if any
+            // For replacing external or password credentials, replace all such credentials.
+            var replacePrefixes = new string[] {
+                CredentialTypes.Password.Prefix,
+                CredentialTypes.ExternalPrefix
+            };
+
+            Func<Credential, bool> replacingPredicate =
+                cred => replacePrefixes
+                    .Any(mp => credential.Type.StartsWith(mp, StringComparison.OrdinalIgnoreCase)
```
scottbommarito: Can also reduce this using the new methods from my PR.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -622,14 +639,22 @@ public static ClaimsIdentity CreateIdentity(User user, string authenticationType
...
+                        && cred.Type.StartsWith(mp, StringComparison.OrdinalIgnoreCase))
+                    || cred.Type == credential.Type;
```
skofman1: > cred.Type == credential.Type; [](start = 22, length = 30)

why do we need cred.Type == credential.Type? Won't 
credential.Type.StartsWith(mp, StringComparison.OrdinalIgnoreCase)
                        && cred.Type.StartsWith(mp, StringComparison.OrdinalIgnoreCase) cover all cases we need to remove?

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -307,21 +311,100 @@ public virtual ActionResult LogOff(string returnUrl)
...
+        }
+
+        [ActionName("AuthenticateExternal")]
+        [HttpGet]
+        public virtual ActionResult AuthenticateExternal(string returnUrl)
+        {
+            // Get list of all enabled providers
+            var providers = GetProviders();
+
+            // Prioritize the external authentication mechanism.
```
scottbommarito: What does prioritizing mean here? We will prefer to use the `AzureActiveDirectoryV2Authenticator` instead of the `MicrosoftAccountAuthenticator` if both can be used? I would explain this better.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -307,21 +311,100 @@ public virtual ActionResult LogOff(string returnUrl)
...
+
+            // Prioritize the external authentication mechanism. 
+            // This is for backwards compatibility with MicrosoftAccount provider. 
+            var authPriority = new string[] {
+                Authenticator.GetName(typeof(AzureActiveDirectoryV2Authenticator)),
+                Authenticator.GetName(typeof(MicrosoftAccountAuthenticator))
+            };
+
+            string externalAuthProvider = null;
+            if (authPriority.Count() > 0)
```
skofman1: >if (authPriority.Count() > 0) [](start = 12, length = 29)

why do you need the if? foreach will do nothing if there are no elements..

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: I just moved the code, this was already there, but I don't think it covers all the cases. The `startswith`  code only checks the prefixes `password.` or `external.` it won't compare or find the exact match for types like `apikey.v3` and all.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -307,21 +311,100 @@ public virtual ActionResult LogOff(string returnUrl)
...
+            {
+                foreach (string authenticator in authPriority)
```
skofman1: > foreach (string authenticator in authPriority) [](start = 15, length = 47)

cvonsider replacing foreach will FirstOrDefault

src/NuGetGallery/Controllers/AuthenticationController.cs
scottbommarito: This `if` doesn't seem to actually do anything--`authPriority` is declared right above and is never empty. Also none of the code inside this will actually fail if `authPriority` is empty. I would get rid of this `if` statement.

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: This is needed, we need to find the credential in all credentials, it won't necessarily belong to the same user(account).

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -307,21 +311,100 @@ public virtual ActionResult LogOff(string returnUrl)
...
+                }
+
+                TempData["Message"] = Strings.ChangeCredential_Success;
+            }
+            else
+            {
+                TempData["ErrorMessage"] = string.Format(Strings.ChangeCredential_ExistingCredential, HttpUtility.UrlEncode(newCredential.Identity));
+            }
+
+            return SafeRedirect(returnUrl);
```
scottbommarito: It appears that it will still redirect to the `returnUrl` if linking fails and the user is not authenticated.

This doesn't sound ideal because the `returnUrl` will likely redirect back to authentication when it sees a user is not logged in.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -307,21 +311,100 @@ public virtual ActionResult LogOff(string returnUrl)
...
         }
 
         [NonAction]
         public ActionResult ChallengeAuthentication(string returnUrl, string provider)
         {
-            return _authService.Challenge(provider, Url.LinkExternalAccount(returnUrl));
+            return _authService.Challenge(provider, returnUrl);
+        }
+
+        public virtual async Task<ActionResult> LinkOrChangeExternalCredential(string returnUrl)
```
skofman1: If the method is only used internally change from public to protected/private

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -307,21 +311,100 @@ public virtual ActionResult LogOff(string returnUrl)
...
+            var newCredential = result.Credential;
+
+            if (await _authService.TryReplaceCredential(user, newCredential))
+            {
+                // Authenticate with the new credential after successful replacement
+                var authenticatedUser = await _authService.Authenticate(newCredential);
+                await _authService.CreateSessionAsync(OwinContext, authenticatedUser);
+
+                // Remove the password credential after linking to external account.
+                var passwordCred = user.Credentials.SingleOrDefault(
```
skofman1: ReplaceCredential removes password credentials, no?

src/NuGetGallery/Infrastructure/CookieTempDataProvider.cs
```diff
@@ -66,8 +67,9 @@ void ITempDataProvider.SaveTempData(ControllerContext controllerContext, IDictio
                     continue;
                 }
 
-                dictionary[key] = cookie[key];
+                dictionary[key] = HttpUtility.UrlDecode(cookie[key]);
```
skofman1: can u explain this change?

src/NuGetGallery.Core/CredentialTypes.cs
```diff
@@ -38,6 +40,26 @@ public static bool IsPassword(string type)
...
+        }
+
+        public static bool IsAzureActiveDirectoryAccount(string type)
+        {
+            if (type == null)
+            {
+                throw new ArgumentNullException(nameof(type));
+            }
+
+            return type.Equals(ExternalPrefix + AzureActiveDirectoryAccount, StringComparison.OrdinalIgnoreCase);
```
skofman1: >ExternalPrefix + AzureActiveDirectoryAccount [](start = 31, length = 44)

Create the same structure for External accounts as ApiKey and Password. i.e., static class with prefix and types.

src/NuGetGallery.Core/CredentialTypes.cs
```diff
@@ -38,6 +40,26 @@ public static bool IsPassword(string type)
...
+
+            return type.Equals(ExternalPrefix + MicrosoftAccount, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public static bool IsAzureActiveDirectoryAccount(string type)
+        {
+            if (type == null)
+            {
+                throw new ArgumentNullException(nameof(type));
+            }
```
skofman1: do we need a null check here? In what scenario will type be null?

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: It removes only the related credentials, if password is to be replaced it will replace password. If external is to be replaced it will remove all external credentials. When removing external it won't remove password credentials.

src/NuGetGallery/Infrastructure/CookieTempDataProvider.cs
shishirx34: The `TempData` that we pass to the view from controller is via cookie. We store this data in the cookie. from My change to provide `identity` information for a failed login contains characters like `<` `>`, the framework doesn't like these type of characters in the cookie, hence these need to be encoded before putting into the cookie. That requires us to decode when reading the data, that is what is done here. See the encoding [here](https://github.com/NuGet/NuGetGallery/pull/5355/files#diff-4c84f0f15dae655d9296beba3e548e29R404)

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: it is public, this method gets invoked after the external authentication from AAD.

src/NuGetGallery.Core/CredentialTypes.cs
shishirx34: I added it for consistency with the `IsPassword` methods above, I don't think we should ever see these.

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: I think that is exactly you would want. In my case it will return to the `Account` page. You wouldn't be calling into this action if the user isn't already logged in.

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: its removed. simplified it.

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: yup, waiting for your PR to go in :)

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: I thought 'Prioritize' means exactly that :D

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: Good point, although I don't see how we would get into this situation. This method throws exception only when the user is an `Organization`, which means the route should never be invoked. But for thoroughness add a catch for this method in controller to bubble the error message to the view.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -307,21 +311,100 @@ public virtual ActionResult LogOff(string returnUrl)
...
+            // Prioritize the external authentication mechanism. Select one provider to
+            // authenticate for linking when multiple external providers are enabled.
+            // This is for backwards compatibility with MicrosoftAccount provider. 
+            var authPriority = new string[] {
+                Authenticator.GetName(typeof(AzureActiveDirectoryV2Authenticator)),
+                Authenticator.GetName(typeof(MicrosoftAccountAuthenticator))
+            };
+
+            string externalAuthProvider = authPriority
+                .FirstOrDefault(authenticator => providers.Any(p => p.ProviderName.Equals(authenticator, StringComparison.OrdinalIgnoreCase)));
```
chenriksson: Is `AzureActiveDirectoryV2Authenticator` always going to be selected over `MicrosoftAccountAuthenticator`? When will it use the latter? Just wondering whether the priority list is even necessary.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -307,21 +311,100 @@ public virtual ActionResult LogOff(string returnUrl)
...
+        [HttpGet]
+        public virtual ActionResult AuthenticateExternal(string returnUrl)
+        {
+            // Get list of all enabled providers
+            var providers = GetProviders();
+
+            // Prioritize the external authentication mechanism. Select one provider to
+            // authenticate for linking when multiple external providers are enabled.
+            // This is for backwards compatibility with MicrosoftAccount provider. 
+            var authPriority = new string[] {
```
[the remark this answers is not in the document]
chenriksson: Would the `externalAuthProvider` lookup be better placed as a static method on Authenticator?

src/NuGetGallery/Controllers/AuthenticationController.cs
chenriksson: Would add doc comments that this method is the challenge response return url (or whatever you call it), so it's clear that it needs to remain public.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -307,21 +311,100 @@ public virtual ActionResult LogOff(string returnUrl)
...
+            {
+                if (await _authService.TryReplaceCredential(user, newCredential))
+                {
+                    // Authenticate with the new credential after successful replacement
+                    var authenticatedUser = await _authService.Authenticate(newCredential);
+                    await _authService.CreateSessionAsync(OwinContext, authenticatedUser);
+
+                    // Remove the password credential after linking to external account.
+                    var passwordCred = user.Credentials.SingleOrDefault(
+                        c => c.Type.StartsWith(CredentialTypes.Password.Prefix, StringComparison.OrdinalIgnoreCase));
```
chenriksson: Do we also remove other external linked credentials? Does this happen somewhere else?

src/NuGetGallery/Strings.resx
```diff
@@ -742,4 +742,19 @@ For more information, please contact '{2}'.</value>
   <data name="TransformAccount_AdminMustBeDifferentAccount" xml:space="preserve">
     <value>Administrator account '{0}' cannot be the same account as the one being transformed.</value>
   </data>
+  <data name="ChangeCredential_ExistingCredential" xml:space="preserve">
+    <value>Failed to update the Microsoft account with '{0}' as it is already linked to another NuGet account.</value>
```
chenriksson: Is this the only scenario where TryReplaceCredential could return false? Maybe this needs to be more generic and refer the customer to support?

Maybe something like `Failed to update the Microsoft account with '{0}'. This could happen if it is already linked to another NuGet account. Contact support for more information.`?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -759,6 +759,23 @@ public virtual Task<ActionResult> RemovePassword()
...
+        [ValidateAntiForgeryToken]
+        public virtual ActionResult LinkOrChangeExternalCredential()
+        {
+            var user = GetCurrentUser();
+            var userHasAADCredential = user.Credentials.Any(c => CredentialTypes.IsAzureActiveDirectoryAccount(c.Type));
+
+            if (userHasAADCredential)
+            {
+                TempData["WarningMessage"] = Strings.ChangeCredential_NotAllowed;
+                return RedirectToAction("Account");
```
chenriksson: Will we eventually support this? Or do we support unlinking the AAD credential?

What if someone leaves their company, but wants to keep the same NuGet account... like people do on GitHub? Seems like a bad experience if we let the company essentially own the account.

src/NuGetGallery/Infrastructure/CookieTempDataProvider.cs
chenriksson: This line? Is newCredential.Identity a Url? Can you give an example of the value that gets encoded?

```
TempData["ErrorMessage"] = string.Format(Strings.ChangeCredential_ExistingCredential, HttpUtility.UrlEncode(newCredential.Identity));
```

src/NuGetGallery/Views/Users/Account.cshtml
```diff
@@ -8,6 +8,16 @@
...
         <div class="@ViewHelpers.GetColumnClasses(ViewBag)">
+            @if (TempData.ContainsKey("WarningMessage"))
+            {
+                @ViewHelpers.AlertWarning(@<text>@((string)TempData["WarningMessage"])</text>);
+            }
+
+            @if (TempData.ContainsKey("ErrorMessage"))
+            {
+                @ViewHelpers.AlertDanger(@<text>@((string)TempData["ErrorMessage"])</text>);
+            }
```
chenriksson: Maybe this should be built into Header.cshtml, like `TempData["Message"]` is?

If we do it there, it'll be easier to keep the formatting consistent. Right now `Message` uses the page column width, but I think @anangaur wants to change it to be the header width (full 12 columns).

src/NuGetGallery/Infrastructure/CookieTempDataProvider.cs
shishirx34: It is not a url, however, I found url encoding be the common encoding pattern for values in the cookie. The other encoding patterns like `HtmlUtility.Encode` contains the characters(like `;`) which isn't compliant with the cookie data.  The example of identiy: 'FirstName LastName <emailaddress>'. Note the `<` `>` characters are not cookie compliant.

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: Most cases it will be either/or. For on premise solutions of gallery we have customers using hte `MicrosoftAccountAuthenticator`, in that case that will be the likely solution used. However, it is possible to have both the authenticators enabled(local development/on prem solution), in which case we want to choose one, hence the priority.

src/NuGetGallery/Controllers/UsersController.cs
shishirx34: Yes we will support this eventually. This is an interim blocking solution. We will check if the user is a member of organization, ask them to leave it and then allow them to replace AAD with MSA.

src/NuGetGallery/Views/Users/Account.cshtml
shishirx34: I wasn't sure about the implications of moving to the header. If TempData["ErrorMessage"] has been used in other places that could possibly duplicate the message, but I see that it isn't being used anywhere else.

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: Yes, `TryReplaceCredential` does the removal while replacing external linked credentials.

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: I'll add the optimization check, but definitely don't want to be replacing the credential of a different user's account, which could possibly lead to orphaned accounts without any login mechanism.

src/NuGetGallery/Controllers/AuthenticationController.cs
scottbommarito: I just had to fix an infinite loop due to a bug in sign-in redirects.

The infinite loop was caused by an action that, despite the user being logged in, would redirect to sign-in when the user didn't have permission to perform the action. Because the user was logged in, the sign-in flow would immediately redirect back to the action, but the action would continue to redirect it back.

I'm sure this code is fine, I just want you to be aware of a potential issue.

src/NuGetGallery/Infrastructure/CookieTempDataProvider.cs
shishirx34: added a comment as well.

src/NuGetGallery/Views/Users/Account.cshtml
shishirx34: Moving this and 'WarningMessage' to header. Won't address the header width change since its unrelated to this PR.

src/NuGetGallery.Core/CredentialTypes.cs
scottbommarito: @shishirx34 it's merged now!

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -302,9 +302,36 @@ public virtual Task ReplaceCredential(string username, Credential credential)
...
             }
+
             return ReplaceCredential(user, credential);
         }
 
+        public virtual async Task<bool> TryReplaceCredential(User user, Credential credential)
+        {
+            if (user == null || credential == null)
+            {
+                return false;
```
chenriksson: Why not `ArgumentNullException`?

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -302,9 +302,36 @@ public virtual Task ReplaceCredential(string username, Credential credential)
...
+            }
+
+            if (UserHasCredential(user, credential) || FindMatchingCredential(credential) != null)
```
chenriksson: nit: would add comment that checking user credentials first is optimization to avoid expensive DB query

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -2,15 +2,19 @@
...
+using System.Globalization;
 using System.Threading.Tasks;
-using System.Web.Mvc;
+using System.Security.Claims;
+using System.Collections.Generic;
 using NuGetGallery.Authentication;
+using NuGetGallery.Authentication.Providers;
 using NuGetGallery.Infrastructure.Authentication;
+using NuGetGallery.Authentication.Providers.MicrosoftAccount;
+using NuGetGallery.Authentication.Providers.AzureActiveDirectoryV2;
```
chenriksson: nit: sort usings

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -322,21 +332,92 @@ public virtual ActionResult LogOff(string returnUrl)
...
+
+            if (await _authService.TryReplaceCredential(user, newCredential))
+            {
+                // Authenticate with the new credential after successful replacement
+                var authenticatedUser = await _authService.Authenticate(newCredential);
+                await _authService.CreateSessionAsync(OwinContext, authenticatedUser);
+
+                // Remove the password credential after linking to external account.
+                var passwordCred = user.Credentials.SingleOrDefault(
+                    c => c.Type.StartsWith(CredentialTypes.Password.Prefix, StringComparison.OrdinalIgnoreCase));
```
chenriksson: nit: could use `c => CredentialTypes.IsPassword(c)`

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -2,19 +2,19 @@
...
+using System.Collections.Generic;
+using NuGetGallery.Filters;
 using NuGetGallery.Areas.Admin;
+using NuGetGallery.Configuration;
+using NuGetGallery.Authentication;
 using NuGetGallery.Areas.Admin.Models;
 using NuGetGallery.Areas.Admin.ViewModels;
-using NuGetGallery.Authentication;
-using NuGetGallery.Configuration;
-using NuGetGallery.Filters;
```
chenriksson: nit: sort usings

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: same reason we didn't want the TrySomething method throwing exception. I think returning false is fine here.

PR #5357, opened 2018-01-26
Attempt to gracefully handle pending requests during WebRole.OnStop event

xavierdecoster:


src/NuGetGallery/WebRole.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.WindowsAzure.ServiceRuntime;
+
+namespace NuGetGallery
+{
+    public class WebRole
+        : RoleEntryPoint
+    {
+        public override void OnStop()
+        {
+            // Once the Stopping event was raised, the load balancer stopped sending requests to the WebRole.
```
scottbommarito: `Once the Stopping event is raised, the load balancer stops sending requests to the WebRole.`

src/NuGetGallery/WebRole.cs
```diff
@@ -0,0 +1,44 @@
...
+            // We'll attempt to gracefully handle any pending requests (e.g. package uploads) to this instance before stopping the role.
+            // There's only a 5 mins window before the role will restart either way.
+            // The Trace data from the OnStop method will never appear in WADLogsTable and is only accessible using DebugView (unless a tricky On-Demand Transfer is performed).
+            // See also: https://azure.microsoft.com/en-us/blog/the-right-way-to-handle-azure-onstop-events/
+            Trace.TraceInformation("OnStop called in WebRole");
+
+            var perfCounter = new PerformanceCounter(
```
scottbommarito: Might be clearer if you rename this something like `currentRequestsPerfCounter` so that it's clear what the counter is returning.

src/NuGetGallery/WebRole.cs
```diff
@@ -0,0 +1,44 @@
...
+                Trace.TraceInformation($"ASP.NET Requests Current = {value}");
+
+                if (value <= 0)
+                {
+                    // No more pending requests: the role can be safely stopped.
+                    break;
+                }
+
+                // Check again in a second.
+                Task.Delay(1000);
```
loic-sharma: You're not awaiting this `Task.Delay`. Also, this method is synchronous, so you should probably use `Thread.Sleep`.

src/NuGetGallery/WebRole.cs
loic-sharma: Silly nit: `PerformanceCounter` should be disposed. I guess it doesn't matter much though since this is on shutdown 😄

xavierdecoster:
@skofman1 replied on the email thread, but I suspect we need both to reach our goals.

src/NuGetGallery/WebRole.cs
xavierdecoster: Fixed.

src/NuGetGallery/WebRole.cs
xavierdecoster: Fixed.

src/NuGetGallery/WebRole.cs
xavierdecoster: Fixed.

src/NuGetGallery/WebRole.cs
```diff
@@ -0,0 +1,45 @@
...
+                    Trace.TraceInformation($"ASP.NET Requests Current = {value}");
+
+                    if (value <= 0)
+                    {
+                        // No more pending requests: the role can be safely stopped.
+                        break;
+                    }
+
+                    // Check again in a second.
+                    Thread.Sleep(1000);
```
loic-sharma: Super nit: `Thread.Sleep(TimeSpan.FromSeconds(1))` is a wee bit more readable.

xavierdecoster: Fixed

PR #5374, opened 2018-01-30
[MSA/AAD] Deprecate password warnings

shishirx34:
Add deprecate warning on the forgot password page, also update the forgot password email with the warning message.

image


shishirx34:
@skofman1 - I will check with Anand, but the warning here was for during recovery because the password will be disabled upon linking to the MSA and no way to 'enable password' from the account page anymore.

PR #5380, opened 2018-02-01
Block users with a discontinued password login from using the site

scottbommarito:
1) Adds a claim (`NuGetClaims.DiscontinuedPassword`) to the session if the user has a discontinued password login
2) Created the `BlockDiscontinuedPasswordAuthorize` attribute to redirect users to the homepage if they have the `NuGetClaims.DiscontinuedPassword` claim.
3) Applied the attribute to all actions except `TransformOrganization`.
4) Added unit tests to all code impacted.

The next step after this PR is to add a prompt on the homepage if the discontinued password login is detected.


src/NuGetGallery.Core/Authentication/NuGetClaims.cs
```diff
@@ -5,13 +5,31 @@ namespace NuGetGallery.Authentication
...
+
+        /// <summary>
+        /// The claim url for the claim that stores whether or not the user is authenticated with a discontinued <see cref="CredentialTypes.Password"/>.
+        /// </summary>
+        public const string DiscontinuedPassword = ClaimsDomain + "discontinuedpassword";
+
+        /// <summary>
+        /// The value of <see cref="DiscontinuedPassword"/> when the user is authenticated with a discontinued <see cref="CredentialTypes.Password"/>.
+        /// </summary>
+        public const string DiscontinuedPasswordValue = "true";
```
chenriksson: Can we just use `true` inline`? A bool string constant seems odd.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -244,6 +247,13 @@ private void InitCredentialFormatters()
                 new UserAuditRecord(user.User, AuditedUserAction.Login, user.CredentialUsed));
         }
 
+        private Claim[] GetPasswordDiscontinuedClaims(AuthenticatedUser user)
+        {
+            return user.CredentialUsed.IsPassword() && _userService.IsOrganizationsEnabledForAccount(user.User) ?
```
chenriksson: White list is still in config? AuthenticationService already uses IAppConfig, so it seems strange to require UserService to read from the config it already has. Could we put `IsOrganizationsEnabledForAccount` somewhere that only requires config?

src/NuGetGallery/Filters/BlockDiscontinuedPasswordAuthorizeAttribute.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Security.Claims;
+using System.Web.Mvc;
+using System.Web.Routing;
+using NuGetGallery.Authentication;
+using AuthenticationTypes = NuGetGallery.Authentication.AuthenticationTypes;
+using AuthorizationContext = System.Web.Mvc.AuthorizationContext;
+
+namespace NuGetGallery.Filters
+{
+    public sealed class BlockDiscontinuedPasswordAuthorizeAttribute : AuthorizeAttribute
```
chenriksson: I recommend making a more generic authorize attribute for *all* Gallery/UI actions. Then I would only mark the ones that we allow with password deprecation.

This way, we're more likely to leave the attributes in-place as-is even if we entirely remove password auth / deprecation from the Gallery. Plus, we could reuse the UI authorize attribute for other features.

```
class UsersController {
    [UIAuthorize]
    public ActionResult Account();

    [UIAuthorize(AllowDeprecatedAuth=true)]
    public ActionResult TransformToOrganization();
}
```

src/NuGetGallery/Filters/BlockDiscontinuedPasswordAuthorizeAttribute.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Claims;
+using System.Web.Mvc;
+using System.Web.Routing;
+using NuGetGallery.Authentication;
+using AuthenticationTypes = NuGetGallery.Authentication.AuthenticationTypes;
+using AuthorizationContext = System.Web.Mvc.AuthorizationContext;
```
chenriksson: nit: sort usings

src/NuGetGallery/Filters/BlockDiscontinuedPasswordAuthorizeAttribute.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace NuGetGallery.Filters
+{
+    public sealed class BlockDiscontinuedPasswordAuthorizeAttribute : AuthorizeAttribute
+    {
+        
+        public BlockDiscontinuedPasswordAuthorizeAttribute()
+        {}
```
chenriksson: nit: Remove default ctor

src/NuGetGallery/Services/IUserService.cs
```diff
@@ -30,6 +30,8 @@ public interface IUserService
 
         bool CanTransformUserToOrganization(User accountToTransform, out string errorReason);
 
+        bool IsOrganizationsEnabledForAccount(User account);
```
chenriksson: See other comment - I wouldn't put this on UserService, since it's just a config check.

src/NuGetGallery/Authentication/AuthenticationService.cs
scottbommarito: I'll see if I can move it to `IAppConfiguration`.

src/NuGetGallery.Core/Authentication/NuGetClaims.cs
scottbommarito: By putting it in a constant I can guarantee the getter and setter are using identical values, e.g.
```
NuGetClaims.DiscontinuedPasswordValue.Equals(discontinuedPasswordClaim, StringComparison.OrdinalIgnoreCase)
```
and
```
new Claim(NuGetClaims.DiscontinuedPassword, NuGetClaims.DiscontinuedPasswordValue)
```
and also in the unit tests.

I agree it's odd, but I think it makes the code more maintainable.

src/NuGetGallery/Authentication/AuthenticationService.cs
scottbommarito: I think I'll actually punt this to the PR where I allow configuration via json file.

src/NuGetGallery/Filters/BlockDiscontinuedPasswordAuthorizeAttribute.cs
scottbommarito: Done, nice suggestion

src/NuGetGallery/Filters/BlockDiscontinuedPasswordAuthorizeAttribute.cs
scottbommarito: Done (because the constructor now has arguments)

src/NuGetGallery.Core/Authentication/NuGetClaims.cs
shishirx34: Yeah it is odd, but I agree with @scottbommarito, the claims usually have a defined schema type and value type, since we aren't really supporting a full fledged claims format lets keep all values as string.

src/NuGetGallery.Core/Authentication/NuGetClaims.cs
```diff
@@ -5,13 +5,31 @@ namespace NuGetGallery.Authentication
...
+        
+        /// <summary>
+        /// The claim url for the claim that stores the <see cref="Credential.Key"/> of the <see cref="Credential"/> used to authenticate the request.
+        /// </summary>
+        public const string CredentialKey = ClaimsDomain + "credentialkey";
+
+        /// <summary>
+        /// The claim url for the claim that stores whether or not the user is authenticated with a discontinued <see cref="CredentialTypes.Password"/>.
+        /// </summary>
+        public const string DiscontinuedPassword = ClaimsDomain + "discontinuedpassword";
```
shishirx34: can we have a better name, 'DiscontinuedPassword' doesn't really sound right. May be `DeprecatedLogin` or `DeprecatedPasswordLogin`. The action can be defined as discontinued(or whatever we want to do it) based off of this property.

src/NuGetGallery/Areas/Admin/Controllers/AdminControllerBase.cs
```diff
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Web.Mvc;
+using NuGetGallery.Filters;
 
 namespace NuGetGallery.Areas.Admin.Controllers
 {
-    [Authorize(Roles="Admins")]
+    [UiAuthorize(Roles="Admins")]
```
shishirx34: make it `UIAuthorize` 'i' capital.

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: can you create a task for the work to do, so that we can keep track of the work remaining.

src/NuGetGallery/Filters/UiAuthorizeAttribute.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Security.Claims;
+using System.Web.Mvc;
+using System.Web.Routing;
+using NuGetGallery.Authentication;
+using AuthenticationTypes = NuGetGallery.Authentication.AuthenticationTypes;
+using AuthorizationContext = System.Web.Mvc.AuthorizationContext;
+
+namespace NuGetGallery.Filters
+{
+    public sealed class UiAuthorizeAttribute : AuthorizeAttribute
```
shishirx34: nit: `UIAuthorizeAttribute`

src/NuGetGallery/Filters/UiAuthorizeAttribute.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        public override void OnAuthorization(AuthorizationContext filterContext)
+        {
+            // If a password credential was used, and the user has a discontinued password claim, redirect them to the homepage.
+            var identity = filterContext.HttpContext.User.Identity as ClaimsIdentity;
+            if (!AllowDiscontinuedPassword &&
+                identity != null && 
+                identity.IsAuthenticated && 
+                identity.AuthenticationType == AuthenticationTypes.LocalUser)
```
shishirx34: I think this authentication type is set to 'LocalUser' for all logins even the external logins in `CreateSessionAsync` I think its a bug, but you don't need to add that check in here. Remove the last check and test for `DeprecatedPasswordLoginClaims`

src/NuGetGallery/Authentication/AuthenticationService.cs
scottbommarito: Added https://github.com/NuGet/NuGetGallery/issues/5386

src/NuGetGallery.Core/Authentication/NuGetClaims.cs
scottbommarito: Deprecated doesn't mean it's discontinued, but I think changing it to `DiscontinuedLogin` is fine.

src/NuGetGallery/Areas/Admin/Controllers/AdminControllerBase.cs
scottbommarito: TIL https://msdn.microsoft.com/en-us/library/ms229043%28v=vs.100%29.aspx
```
Do capitalize both characters of two-character acronyms, except the first word of a camel-cased identifier.
```

src/NuGetGallery/Filters/UiAuthorizeAttribute.cs
scottbommarito: If I remember correctly it's only set to `LocalUser` for password. Doesn't matter because of the claim check, so I'll remove it.

src/NuGetGallery/Areas/Admin/Controllers/AdminControllerBase.cs
shishirx34: Yup 😆

src/NuGetGallery/Filters/UiAuthorizeAttribute.cs
scottbommarito: done

scottbommarito:
@shishirx34 @chenriksson handled all feedback

shishirx34:
Looks like a test is broken, also merge changes from dev.

shishirx34:
Would this have any distinctive impact on performance?

scottbommarito:
@shishirx34 shouldn't have any performance disregard, will check the test--didn't fail locally

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -244,6 +247,13 @@ private void InitCredentialFormatters()
                 new UserAuditRecord(user.User, AuditedUserAction.Login, user.CredentialUsed));
         }
 
+        private Claim[] GetDiscontinuedLoginClaims(AuthenticatedUser user)
+        {
+            return user.CredentialUsed.IsPassword() && _userService.IsOrganizationsEnabledForAccount(user.User) ?
+                new[] { new Claim(NuGetClaims.DiscontinuedLogin, NuGetClaims.DiscontinuedLoginValue) } :
+                new Claim[0];
```
chenriksson: nit: return null instead of allocation?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -72,14 +72,14 @@ protected override User GetAccount(string accountName)
...
-        [Authorize]
+        [UIAuthorize]
         public virtual ActionResult Account()
         {
             return AccountView(GetCurrentUser());
         }
 
         [HttpGet]
-        [Authorize]
+        [UIAuthorize(true)]
```
chenriksson: Would use named arg for bool.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -213,6 +211,13 @@ public bool CanTransformUserToOrganization(User accountToTransform, out string e
             return errorReason == null;
         }
 
+        public bool IsOrganizationsEnabledForAccount(User account)
```
chenriksson: nit: `AreOrganizationsEnabledForAccount`?

tests/NuGetGallery.Facts/Filters/UiAuthorizeAttributeFacts.cs
```diff
@@ -0,0 +1,147 @@
...
+                        AllowsDiscontinuedLogins_Data,
+                        AuthTypes.Select(t => MemberDataHelper.AsData(t, false)))
+                    .Concat(
+                        AuthTypes.Select(t => MemberDataHelper.AsData(true, t, true)));
+
+            [Theory]
+            [MemberData(nameof(SucceedsForAuthenticatedUserWithoutDiscontinuedLogin_Data))]
+            public void SucceedsForAuthenticatedUserWithoutDiscontinuedLogin(bool allowsDiscontinuedLogin, string authType, bool hasDiscontinuedLoginClaim)
+            {
+                var context = BuildAuthorizationContext(BuildClaimsIdentity(authType, authenticated: true, hasDiscontinuedLoginClaim: hasDiscontinuedLoginClaim).Object).Object;
```
chenriksson: nit: wrap the long lines in this file

tests/NuGetGallery.Facts/Controllers/ControllerTests.cs
```diff
@@ -86,20 +85,67 @@ public void AllActionsHaveAntiForgeryTokenIfNotGet()
...
-                .Where(m => !(m.GetCustomAttributes()
-                    .Any(a => a.GetType() == typeof(ValidateAntiForgeryTokenAttribute))));
-            
+                .Where(m => !(m.GetCustomAttributes().Any(a => a.GetType() == typeof(ValidateAntiForgeryTokenAttribute))));
+
             Assert.Empty(actionsMissingAntiForgeryToken);
         }
+
+        [Fact]
+        public void AllActionsHaveUIAuthorizeAttribute()
```
chenriksson: nice test idea!

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1750,7 +1750,7 @@ private string GetSecurePushPolicyDescriptions()
             return Json(null);
         }
 
-        [Authorize]
+        [UIAuthorize]
         [HttpPost]
         [ValidateAntiForgeryToken]
         public virtual async Task<JsonResult> PreviewReadMe(ReadMeRequest formData)
```
chenriksson: Have you tested the behavior of redirect on these JsonResult async callbacks? What's going to happen?

chenriksson:
@scottbommarito Mostly looks goood to me, just wondering what the behavior is going to be for the async callback actions where we now redirect instead of returning JsonResult? Have you tried this?

src/NuGetGallery/Authentication/AuthenticationService.cs
scottbommarito: I believe that would behave differently...if nothing is supplied for a `params` argument I'm pretty sure the value of the parameter is an empty array, not `null`.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I don't think it's any different than the existing behavior of `[Authorize]` when the user isn't logged in, but I will test it myself.

scottbommarito: Confirmed that it behaves as a regular redirect.

chenriksson: I guess the main page would redirect anyways :-)

PR #5382, opened 2018-02-02
Remove password credential from user when they link an external account

scottbommarito:
Short addition to https://github.com/NuGet/NuGetGallery/pull/5338


src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -536,6 +536,12 @@ private ActionResult RedirectFromRegister(string returnUrl)
 
             await _authService.AddCredential(user.User, result.Credential);
 
+            var passwordCredential = user.User.Credentials.SingleOrDefault(c => c.IsPassword());
```
chenriksson: Extension method?

```
public static CredentialTypes {
    public static GetPassword(this IEnumerable<Credential> credentials);
}
```

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -536,6 +536,12 @@ private ActionResult RedirectFromRegister(string returnUrl)
...
+            if (passwordCredential != null)
+            {
+                await _authService.RemoveCredential(user.User, passwordCredential);
```
chenriksson: Do we want this in the same transaction as AddCredential? Should the APIs support deferring the SaveChangesAsync?

scottbommarito: I think if removing the password fails, we should still proceed with the request, just leave the credential there--no need to remove a perfectly valid external credential.

scottbommarito: This is probably also a problem with @shishirx34's link and replace flow, where I based this code off of.

src/NuGetGallery/Controllers/AuthenticationController.cs
scottbommarito: done, although I put it on `User` itself because I can't imagine anything else having a list of credentials

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -401,7 +401,7 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
                 await _authService.CreateSessionAsync(OwinContext, authenticatedUser);
 
                 // Remove the password credential after linking to external account.
-                var passwordCred = user.Credentials.SingleOrDefault(c => c.IsPassword());
+                var passwordCred = user.GetPassword();
```
shishirx34: nit: perhaps `GetPasswordCredential` be more appropriate? GetPassword seems like we are getting the password value. Don't care though.

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: I agree it would be much better to have an API for `AddExternalAndRemovePasswordCredential` but I think its fine, for now, to leave it as is. I won't say it be high priority right now.

PR #5383, opened 2018-02-02
Add/remove members UI

chenriksson:
Issue #4990 
- Adds Members section on Manage Organization view
- Can add, remove, or edit (toggle admin/collaborator)

Screenshot (UPDATED per demo feedback):
image

see screen capture here: \\\\sollentuna\Public\5383_AddRemoveMembers

### DIFF
--- a/src/NuGetGallery/Controllers/OrganizationsController.cs
+++ b/src/NuGetGallery/Controllers/OrganizationsController.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
 using System.Web.Mvc;
 using NuGetGallery.Authentication;
 
@@ -37,6 +39,60 @@ namespace NuGetGallery
             var account = GetAccount(accountName);
 
             return AccountView(account);
+        }
+
+        [HttpPost]
+        [Authorize]
+        [ValidateAntiForgeryToken]
+        public async Task<JsonResult> AddOrUpdateMember(string accountName, string memberName, bool isAdmin)
+        {
+            var account = GetAccount(accountName);
+
+            if (account == null
+                || ActionsRequiringPermissions.ManageAccount.CheckPermissions(GetCurrentUser(), account)
+                    != PermissionsCheckResult.Allowed)
+            {
+                Response.StatusCode = (int)HttpStatusCode.Forbidden;
+                return Json(Strings.Unauthorized);
+            }
+
+            try
+            {
+                var membership = await UserService.AddOrUpdateMemberAsync(account, memberName, isAdmin);
+                return Json(new OrganizationMemberViewModel(membership));
+            }
+            catch (EntityException e)
+            {
+                Response.StatusCode = (int)HttpStatusCode.BadRequest;
+                return Json(e.Message);
+            }
+        }
+
+        [HttpPost]
+        [Authorize]
+        [ValidateAntiForgeryToken]
+        public async Task<JsonResult> DeleteMember(string accountName, string memberName)
+        {
+            var account = GetAccount(accountName);
+
+            if (account == null
+                || ActionsRequiringPermissions.ManageAccount.CheckPermissions(GetCurrentUser(), account)
+                    != PermissionsCheckResult.Allowed)
+            {
+                Response.StatusCode = (int)HttpStatusCode.Forbidden;
+                return Json(Strings.Unauthorized);
+            }
+
+            try
+            {
+                await UserService.DeleteMemberAsync(account, memberName);
+                return Json(Strings.DeleteMember_Success);
+            }
+            catch (EntityException e)
+            {
+                Response.StatusCode = (int)HttpStatusCode.BadRequest;
+                return Json(e.Message);
+            }
         }
 
         protected override void UpdateAccountViewModel(Organization account, OrganizationAccountViewModel model)
--- a/src/NuGetGallery/Controllers/OrganizationsController.cs
+++ b/src/NuGetGallery/Controllers/OrganizationsController.cs
@@ -44,7 +44,7 @@ namespace NuGetGallery
         [HttpPost]
         [Authorize]
         [ValidateAntiForgeryToken]
-        public async Task<JsonResult> AddOrUpdateMember(string accountName, string memberName, bool isAdmin)
+        public async Task<JsonResult> AddMember(string accountName, string memberName, bool isAdmin)
         {
             var account = GetAccount(accountName);
 
@@ -52,19 +52,42 @@ namespace NuGetGallery
                 || ActionsRequiringPermissions.ManageAccount.CheckPermissions(GetCurrentUser(), account)
                     != PermissionsCheckResult.Allowed)
             {
-                Response.StatusCode = (int)HttpStatusCode.Forbidden;
-                return Json(Strings.Unauthorized);
+                return Json((int)HttpStatusCode.Forbidden, Strings.Unauthorized);
             }
 
             try
             {
-                var membership = await UserService.AddOrUpdateMemberAsync(account, memberName, isAdmin);
+                var membership = await UserService.AddMemberAsync(account, memberName, isAdmin);
                 return Json(new OrganizationMemberViewModel(membership));
             }
             catch (EntityException e)
             {
-                Response.StatusCode = (int)HttpStatusCode.BadRequest;
-                return Json(e.Message);
+                return Json((int)HttpStatusCode.BadRequest, e.Message);
+            }
+        }
+
+        [HttpPost]
+        [Authorize]
+        [ValidateAntiForgeryToken]
+        public async Task<JsonResult> UpdateMember(string accountName, string memberName, bool isAdmin)
+        {
+            var account = GetAccount(accountName);
+
+            if (account == null
+                || ActionsRequiringPermissions.ManageAccount.CheckPermissions(GetCurrentUser(), account)
+                    != PermissionsCheckResult.Allowed)
+            {
+                return Json((int)HttpStatusCode.Forbidden, Strings.Unauthorized);
+            }
+
+            try
+            {
+                var membership = await UserService.UpdateMemberAsync(account, memberName, isAdmin);
+                return Json(new OrganizationMemberViewModel(membership));
+            }
+            catch (EntityException e)
+            {
+                return Json((int)HttpStatusCode.BadRequest, e.Message);
             }
         }
 
@@ -79,8 +102,7 @@ namespace NuGetGallery
                 || ActionsRequiringPermissions.ManageAccount.CheckPermissions(GetCurrentUser(), account)
                     != PermissionsCheckResult.Allowed)
             {
-                Response.StatusCode = (int)HttpStatusCode.Forbidden;
-                return Json(Strings.Unauthorized);
+                return Json((int)HttpStatusCode.Forbidden, Strings.Unauthorized);
             }
 
             try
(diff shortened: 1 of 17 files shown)


src/NuGetGallery/Scripts/gallery/page-manage-organization.js
```diff
@@ -0,0 +1,168 @@
...
+                };
+                addAntiForgeryToken(data);
+
+                // Send the request.
+                $.ajax({
+                    url: self.OrganizationViewModel.DeleteMemberUrl,
+                    type: 'POST',
+                    dataType: 'json',
+                    data: data,
+                    success: function () {
```
scottbommarito: Using `DeleteMember` and `ToggleIsAdmin`, a user can remove their access to this page, correct?

We should
1 - check if the user is attempting to remove themselves or un-admin themselves
2 - if they are, prompt them, telling them they are giving up their ability to manage the organization
3 - if they accept, navigate them off the page

Basically, the same thing manage package owners does.

This is fine for now though, should probably file an issue.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -40,6 +40,83 @@ public class UserService : IUserService
             EntitiesContext = entitiesContext;
         }
 
+        public async Task<Membership> AddOrUpdateMemberAsync(Organization organization, string memberName, bool isAdmin)
+        {
+            organization = organization ?? throw new ArgumentNullException(nameof(organization));
+            if (string.IsNullOrEmpty(memberName))
+            {
+                throw new ArgumentException(Strings.OrganizationMemberNameIsRequired, nameof(memberName));
```
scottbommarito: `ArgumentException` isn't an `EntityException`, so I believe this will cause a 500 in the controller. Same with `DeleteMemberAsync`. Should be a 400.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -40,6 +40,83 @@ public class UserService : IUserService
             EntitiesContext = entitiesContext;
         }
 
+        public async Task<Membership> AddOrUpdateMemberAsync(Organization organization, string memberName, bool isAdmin)
```
scottbommarito: Using the same method for both the add and update methods could obscure actions by users at the same time.

Admin A, Admin B, and Collaborator C are all members of an org. A and B load the manage page at the same time. B removes C from the org, while, at the same time, A upgrades C to an Admin

Depending on which one gets executed first, C will either be removed or an admin, but both A and B will walk away happy, which is misleading.

This isn't a big deal and it's perfectly fine to have it like this for now, but it's something to think about. May want to file an issue.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -40,6 +40,83 @@ public class UserService : IUserService
...
+                }
+
+                membership = new Membership()
+                {
+                    Member = member,
+                    IsAdmin = isAdmin
+                };
+                organization.Members.Add(membership);
+
+                await EntitiesContext.SaveChangesAsync();
```
scottbommarito: put this after the `if { ... } else if { ... }` and u can remove it from both branches

src/NuGetGallery/Services/UserService.cs
```diff
@@ -40,6 +40,83 @@ public class UserService : IUserService
...
+            }
+            
+            var match = organization.Members
+                .Where(m => m.Member.Username.Equals(memberName, StringComparison.OrdinalIgnoreCase))
+                .SingleOrDefault();
+
+            if (match == null)
+            {
+                throw new EntityException(string.Format(CultureInfo.CurrentCulture,
+                    Strings.UpdateMember_MemberNotFound, memberName));
```
scottbommarito: nit: this is used by both `UpdateMember` and `DeleteMember` but the string's name is `UpdateMember_MemberNotFound`

tests/NuGetGallery.Facts/Controllers/OrganizationsControllerFacts.cs
```diff
@@ -310,5 +311,240 @@ protected override User GetCurrentUser(OrganizationsController controller)
                 Assert.False(model.SuccessfulConfirmation);
             }
         }
+
+        public class TheAddOrUpdateMemberAction : AccountsControllerTestContainer
+        {
+            private const string defaultMemberName = "member";
+
+            protected override User GetCurrentUser(OrganizationsController controller)
```
scottbommarito: I think this would be a little clearer if you instead just added `currentUser` as a parameter to `InvokeAddOrUpdateMember` that defaults to `Fakes.OrganizationAdmin`

```
private Task<JsonResult> InvokeAddOrUpdateMember(..., User currentUser = null)
{
    controller.SetCurrentUser(currentUser ?? Fakes.OrganizationAdmin);
    ...
}
```

tests/NuGetGallery.Facts/Controllers/OrganizationsControllerFacts.cs
```diff
@@ -310,5 +311,240 @@ protected override User GetCurrentUser(OrganizationsController controller)
...
+            {
+                return controller.GetCurrentUser() ?? Fakes.OrganizationAdmin;
+            }
+
+            [Fact]
+            public async Task WhenUserIsCollaborator_ReturnsNonSuccess()
+            {
+                // Arrange
+                var controller = GetController();
```
scottbommarito: `WhenUserIsCollaborator_ReturnsNonSuccess` and `WhenUserIsNotMember_ReturnsNonSuccess` have identical bodies, except the current user is different. You should be able to rewrite this as a theory that takes in the current user (or a `Func<Fakes, User>`).

tests/NuGetGallery.Facts/Controllers/OrganizationsControllerFacts.cs
```diff
@@ -310,5 +311,240 @@ protected override User GetCurrentUser(OrganizationsController controller)
...
+                Assert.IsType<JsonResult>(result);
+                Assert.Equal("error", ((JsonResult)result).Data);
+
+                GetMock<IUserService>().Verify(s => s.AddOrUpdateMemberAsync(account, defaultMemberName, isAdmin), Times.Once);
+            }
+
+            [Theory]
+            [InlineData(true)]
+            [InlineData(false)]
+            public async Task WhenSuccess_ReturnsSuccess(bool isAdmin)
```
scottbommarito: lol this test name

something along the lines of `WhenValidRequest_ReturnsSuccess` is probably better

chenriksson:
Feedback from demo:
- change edit icon to admin/collaborator drop-down
- add confirmation dialog for add/remove member
- expand message banners across full header (separate PR?)

src/NuGetGallery/Scripts/gallery/page-manage-organization.js
chenriksson: No, I'm blocking the current user from changing their own membership. It makes the code simpler - prevents the removal of the last admin, and no need to navigate them away from the page. I also check whether the last owner is being removed in the back end just in case, but that code shouldn't be reached.

I figure we can file an issue if we want to add this later... but I vote for the simplest approach.

src/NuGetGallery/Services/UserService.cs
chenriksson: It's done this way so that `SaveChangesAsync` isn't called if there is no change.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -39,6 +41,60 @@ public virtual ActionResult ManageOrganization(string accountName)
...
+        [ValidateAntiForgeryToken]
+        public async Task<JsonResult> AddOrUpdateMember(string accountName, string memberName, bool isAdmin)
+        {
+            var account = GetAccount(accountName);
+
+            if (account == null
+                || ActionsRequiringPermissions.ManageAccount.CheckPermissions(GetCurrentUser(), account)
+                    != PermissionsCheckResult.Allowed)
+            {
+                Response.StatusCode = (int)HttpStatusCode.Forbidden;
```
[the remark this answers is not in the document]
chenriksson: done

src/NuGetGallery/Services/UserService.cs
chenriksson: split

src/NuGetGallery/Services/UserService.cs
chenriksson: Removed memberName check, will let it throw the EntityException below when no user is  found.

src/NuGetGallery/Services/UserService.cs
chenriksson: updated

tests/NuGetGallery.Facts/Controllers/OrganizationsControllerFacts.cs
chenriksson: `GetCurrentUser` virtual was added so I could reuse the same tests for `UsersController` (self) and `OrganizationsController` (on behalf of). Even though these tests are only specific to `OrganizationsController`, I prefer to keep the same pattern as the shared tests.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -39,6 +41,81 @@ public virtual ActionResult ManageOrganization(string accountName)
...
+                || ActionsRequiringPermissions.ManageAccount.CheckPermissions(GetCurrentUser(), account)
+                    != PermissionsCheckResult.Allowed)
+            {
+                return Json((int)HttpStatusCode.Forbidden, Strings.Unauthorized);
+            }
+
+            try
+            {
+                await UserService.DeleteMemberAsync(account, memberName);
+                return Json(Strings.DeleteMember_Success);
```
cristinamanum: The verification of the last admin will be done in UserService? How the information will displayed to the UI?

src/NuGetGallery/Scripts/gallery/page-manage-organization.js
```diff
@@ -0,0 +1,183 @@
...
+                        self.Error(error);
+                    }
+                });
+            };
+        }
+
+        // Set up the data binding.
+        var manageOrganizationViewModel = new ManageOrganizationViewModel(initialData);
+        ko.applyBindings(manageOrganizationViewModel, document.body);
+    });
```
cristinamanum: I guess this will need manual testing?

src/NuGetGallery/Services/UserService.cs
```diff
@@ -40,6 +40,98 @@ public class UserService : IUserService
...
+
+            var membership = FindMembershipByUsername(organization, memberName);
+            if (membership == null)
+            {
+                throw new EntityException(string.Format(CultureInfo.CurrentCulture,
+                    Strings.UpdateOrDeleteMember_MemberNotFound, memberName));
+            }
+
+            // block removal of last admin
+            if (membership.IsAdmin && organization.Members.Count(m => m.IsAdmin) == 1)
```
cristinamanum: Would be possible due to race conditions to still remove the last admin? The two admins remove each other and both pass this line.

src/NuGetGallery/Controllers/OrganizationsController.cs
chenriksson: I'm preventing removal of last admin in UserService as safe-guard, but this code shouldn't be reached. In the UI, I'm blocking the current user from removing themselves as an admin for simplicity's sake.

src/NuGetGallery/Scripts/gallery/page-manage-organization.js
chenriksson: All Gallery front-end requires manual testing... not sure if we've ever attempted JS unit tests.

src/NuGetGallery/Services/UserService.cs
chenriksson: I'm also blocking the current user from removing themselves. I'm sure it could still be possible if 2 admins simultaneously deleted each other, but that's unlikely.

chenriksson:
@cristinamanum Thanks for the review!

chenriksson:
Per discussion with @anangaur , will add the following in another PR:
- Allow current user to downgrade/remove self, with confirmations for both
- Add confirmation when upgrading user to Administrator (similar to package owners)

PR #5384, opened 2018-02-02
[Organization/AAD] Add UI support for transforming accounts or linking AAD for certain accounts

shishirx34:
This PR adds support to show the UI Modal dialog for transforming **certain** accounts. Most of the changes are UI related and some refactoring for linking AAD. Added tests for the refactoring.

This still needs to be integrated with @scottbommarito's backend changes to identify accounts to be transformed as per whitelist(domains or accounts).

Modal dialog:

image


src/NuGetGallery/Views/Pages/Home.cshtml
```diff
@@ -1,11 +1,40 @@
...
 }
 
+@if (Model.ShowTransformModal)
+{
+    @TransformToOrgOrLinkMSAScript()
+}
+
+@helper TransformToOrgOrLinkMSAScript()
+{
+    <script src="~/Scripts/jquery-1.11.0.min.js"></script>
```
scottbommarito: I believe both of these are already pulled in through the default scripts bundle, `site.min.js`, which is included in the layout, `Views/Shared/Gallery/Layout.cshtml`. You should be able to remove these includes.

src/NuGetGallery/Views/Pages/_TransformOrLink.cshtml
```diff
@@ -0,0 +1,37 @@
...
+        @if (TempData.ContainsKey("ErrorMessage"))
+        {
+            <div class="error-message">
+                @ViewHelpers.AlertDanger(@<text>@((string)TempData["ErrorMessage"])</text>)
+            </div>
+        }
+        <div class="row signin-node">
+            <a role="button" 
+               class="btn btn-default btn-block provider-button" 
+               href="@Url.AuthenticateExternal("/", allowAnyEmailLinking: false)">
```
scottbommarito: When you/me/someone integrates this with my backend, we can make it so the returnUrl gets properly piped in here and the user gets redirected to the page they were initially trying to visit.

src/NuGetGallery/Views/Pages/_TransformOrLink.cshtml
```diff
@@ -0,0 +1,37 @@
...
+                <img height="24" width="24" alt="" aria-hidden="true"
+                     src="@Href("~/Content/gallery/img/microsoft-account.svg")"
+                     @(!string.IsNullOrEmpty("~/Content/gallery/img/microsoft-account-24x24.png") ? (IHtmlString)ViewHelpers.ImageFallback(Url.Absolute("~/Content/gallery/img/microsoft-account-24x24.png")) : new HtmlString(string.Empty)) />
+                <span>Sign in with Microsoft</span>
+            </a>
+        </div>
+        <p>If this is not an individual account and is based on a team or group email address</p>
+        <div class="transform-node">
+            <a href="@Url.TransformAccount()">Transform this account to an organization on NuGet.org</a>
+        </div>
```
scottbommarito: As discussed offline, we should probably include a "dismiss" or at least "sign out" button because otherwise this modal dialog will block users from doing anything else on the site, even logging into another account and migrating that one first. @anangaur any thoughts?

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
```diff
@@ -920,6 +922,109 @@ public class TheLinkOrChangeExternalCredentialAction : TestContainer
                 Assert.Equal(string.Format(Strings.ChangeCredential_Failed, identity), controller.TempData["ErrorMessage"]);
             }
 
+            [Fact]
+            public async Task GivenDisallowAnyEmailLinking_ItSafeRedirectsToReturnUrlWithErrorMessageForWrongEmail()
+            {
+                // Arrange
+                GetMock<AuthenticationService>(); // Force a mock to be created
```
scottbommarito: you can probably share a lot of code between these tests with a helper method that accepts the email to return in `x.GetIdentityInformation`

src/Bootstrap/less/theme/base.less
```diff
@@ -75,6 +75,69 @@ body {
...
+      }
+
+      .transform-node {
+        text-align: center; 
+        margin-bottom: 10px;
+      }
+    }
+  }
+}
+
```
chenriksson: Could we separate into a `less\theme\modals.less` file, similar to how bootstrap does?

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -392,8 +392,17 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
...
-            var newCredential = result.Credential;
+            if (!allowAnyEmailLinking)
+            {
+                var userInformation = result.Authenticator?.GetIdentityInformation(result.ExternalIdentity);
+                if (userInformation == null || !user.EmailAddress.Equals(userInformation.Email, StringComparison.OrdinalIgnoreCase))
+                {
+                    TempData["ErrorMessage"] = string.Format(Strings.ChangeCredential_EmailAddressMismatched, user.EmailAddress);
+                    return SafeRedirect(returnUrl);
+                }
+            }
```
chenriksson: Does `allowAnyEmailLinking` require that the AAD email matches the Gallery account?

I would rename for clarify, and so the inline check doesn't have to be negated. i.e., `requireEmailMatch`.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -345,7 +345,7 @@ public virtual ActionResult AuthenticatePost(string returnUrl, string provider)
 
         [ActionName("AuthenticateExternal")]
         [HttpGet]
-        public virtual ActionResult AuthenticateExternal(string returnUrl)
+        public virtual ActionResult AuthenticateExternal(string returnUrl, bool allowAnyEmailLinking = true)
```
chenriksson: Can the user modify this in the request, so they can control whether the check is done?

src/NuGetGallery/ViewModels/GalleryHomeViewModel.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGetGallery
+{
+    public class GalleryHomeViewModel
+    {
+        public bool ShowTransformModal { get; set; }
```
chenriksson: Where does this get set to true?

src/NuGetGallery/ViewModels/GalleryHomeViewModel.cs
```diff
@@ -0,0 +1,19 @@
...
+
+        public IList<AuthenticationProviderViewModel> Providers { get; set; }
```
chenriksson: Where is this used?

src/NuGetGallery/Views/Pages/Home.cshtml
chenriksson: `site.min.js` also uses a newer version of jquery, 1.12.4, since the redesign. `bootstrap.js` below should also be included with the `site.min.js` bundle.

src/NuGetGallery/Views/Pages/Home.cshtml
```diff
@@ -1,11 +1,40 @@
...
+        });
+
+        $(document).on('ready', function (e) {
+            $("#popUpModal").modal({
+                show: true
+            });
+        })
+
+        $(this).focus();
+    </script>
```
chenriksson: This should probably go into `page-home.js`

src/NuGetGallery/Views/Pages/_TransformOrLink.cshtml
```diff
@@ -0,0 +1,37 @@
...
+    <div class="modal-body">
+        <div class="tag-node">
+            <span>Sign in using your Microsoft corporate account</span>
+        </div>
+        @if (TempData.ContainsKey("ErrorMessage"))
+        {
+            <div class="error-message">
+                @ViewHelpers.AlertDanger(@<text>@((string)TempData["ErrorMessage"])</text>)
+            </div>
+        }
```
chenriksson: `TempData["ErrorMessage"]` is already displayed by `header.cshtml`. Is this displaying it twice?

src/NuGetGallery/Views/Pages/_TransformOrLink.cshtml
```diff
@@ -0,0 +1,37 @@
...
+        <div class="row signin-node">
+            <a role="button" 
+               class="btn btn-default btn-block provider-button" 
+               href="@Url.AuthenticateExternal("/", allowAnyEmailLinking: false)">
+                <img height="24" width="24" alt="" aria-hidden="true"
+                     src="@Href("~/Content/gallery/img/microsoft-account.svg")"
+                     @(!string.IsNullOrEmpty("~/Content/gallery/img/microsoft-account-24x24.png") ? (IHtmlString)ViewHelpers.ImageFallback(Url.Absolute("~/Content/gallery/img/microsoft-account-24x24.png")) : new HtmlString(string.Empty)) />
```
chenriksson: nit: maybe we should start declaring constants for these images.

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
```diff
@@ -16,6 +16,8 @@
 using NuGetGallery.Authentication.Providers.MicrosoftAccount;
 using NuGetGallery.Infrastructure.Authentication;
 using Xunit;
+using NuGetGallery.Authentication.Providers.AzureActiveDirectoryV2;
+using NuGetGallery.Authentication.Providers;
```
chenriksson: nit: sort

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
chenriksson: These test methods are also huge... extracting functionality into helpers would also help readability.

src/NuGetGallery/Views/Pages/_TransformOrLink.cshtml
shishirx34: Yes that is the plan. This is a placeholder for now.

src/NuGetGallery/ViewModels/GalleryHomeViewModel.cs
shishirx34: This will be set upon @scottbommarito changes to validated the whitelisted accounts that need migration, I will add as part of integration work in next PR.

src/NuGetGallery/ViewModels/GalleryHomeViewModel.cs
shishirx34: This would ideally be passed to the view to show the 'Sign in with Microsoft' button. For now I have hardcoded it in the dialog, I will see where we can pass this information from to populate the transform modal dialog in my next PR.

src/NuGetGallery/Views/Pages/Home.cshtml
shishirx34: couldn't really remove it as it is, since the order of imports for scripts matters and rendering in the script here needed these imports. It looks like moving this entire script code to `page-home.js` will get rid of these script imports. 
- done.

src/NuGetGallery/Views/Pages/Home.cshtml
shishirx34: Good idea. Done.

src/NuGetGallery/Views/Pages/_TransformOrLink.cshtml
shishirx34: Yes for now, I couldn't really think of a better place to show any kind of errors upon transform flow, and I couldn't test it well without @scottbommarito's changes. I will remove this for now and address it in next PR.

src/NuGetGallery/Views/Pages/_TransformOrLink.cshtml
shishirx34: These are placeholders. In my next PR I will look into passing these as providers from the controllers

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: Yes, you are right, it can be modified. For now the point of concern will be during transform. I can't test it fully until I integrate it with @scottbommarito's changes. I think in all likelihood I will have to create a new route for linking/registering MSA/AAD from tranform route because the functionality is a mix of Changing/Registering flow. I will leave it as is for now and address this in my subsequent PR. Made a note of it.

shishirx34: On second thoughts, I will undo the changes to this file since it will need to be rewritten anyways, keeping this PR for just the UI transform dialog changes.

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: Undoing all these changes.

PR #5402, opened 2018-02-05
SendValidationTakingTooLongNotice method with tests

agr:
Added a method to `CoreMessageService` that sends notification about validation taking too much time.

Part of the changes for the https://github.com/NuGet/Engineering/issues/1144


karann-msft:
@agr what is "too much time"? where is that defined and how easy is to modify that?

agr:
@karann-msft, it's what previously was "failure timeout" in the orchestrator configuration file:

https://github.com/NuGet/NuGet.Jobs/blob/0cf53771820b35306f12ddd9e730aff12a30a338/src/NuGet.Services.Validation.Orchestrator/settings.json#L6

agr:
@karann-msft modification would consist of configuration update and service redeployment.

skofman1:
@agr, what will be this value configured to? 1 hour? 24 hours? 
IMO 1 hour is the right direction. @karann-msft ?

agr:
Right now it "inherited" the previous "failAfter" value of 1 day, reducing it seems like a right thing. Would need to update our alerting, too, once we settle on a value.

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -125,6 +125,45 @@ public void SendSignedPackageNotAllowedNotice(Package package, string packageUrl
             }
         }
 
+        public void SendValidationTakingTooLongNotice(Package package, string packageUrl, string packageSupportUrl)
+        {
+            string subject = "[{0}] Package validation taking longer than expected - {1} {2}";
+            string body = "It is taking longer than expected for your package [{1} {2}]({3}) to get published.\n\n" +
+                "We are looking into it and there is no action on you at this time. We’ll send you an email notification when your package has been published.\n\n" +
+                "Thank you for your patience.\n\n" +
+                "NuGet Team";
```
chenriksson: Should we use something like `GalleryOwner`, instead of `NuGet Team`?

karann-msft:
I agree with @skofman1, 1 hour seems right. In addition, to sending the email, we should create an alert for us to investigate.
cc: @anangaur

src/NuGetGallery.Core/Services/CoreMessageService.cs
agr: You are right. Good catch!

agr:
We already have an alert

src/NuGetGallery.Core/Services/CoreMessageService.cs
agr: Removed the signature completely as we don't have it in other emails.

Issue #5261, opened 2018-01-10
Organizations: AAD tenant policy


chenriksson:
Includes the following:
- Create new organization security policy that requires members to belong to the same AAD tenant
- UserService.TransformUserToOrganization should add this policy to new organizations
  - Requires admin to have AAD credential w/ tenant ID, which is used to set the policy
- Edit Organization UI should enforce this policy when adding/removing members

PR #5413, opened 2018-02-06
Add organization tenant policy

chenriksson:
Addresses issue #5261
- Create organization policy requiring AAD credential with matching tenant
- On transform: adds policy to organization, using tenant from initial admin's AAD credential
- On add member: evaluates organization policy for new member

Will also require a small shims change for the new security policy action... separate PR coming.

### DIFF
--- a/src/NuGetGallery.Core/CredentialTypes.cs
+++ b/src/NuGetGallery.Core/CredentialTypes.cs
@@ -117,5 +117,10 @@ namespace NuGetGallery
         {
             return type?.Equals(ApiKey.VerifyV1, StringComparison.OrdinalIgnoreCase) ?? false;
         }
+
+        public static Credential GetOrganizationCredential(this ICollection<Credential> credentials)
+        {
+            return credentials.SingleOrDefault(c => IsAzureActiveDirectoryAccount(c.Type));
+        }
     }
 }
--- a/src/NuGetGallery.Core/CredentialTypes.cs
+++ b/src/NuGetGallery.Core/CredentialTypes.cs
@@ -118,7 +118,7 @@ namespace NuGetGallery
             return type?.Equals(ApiKey.VerifyV1, StringComparison.OrdinalIgnoreCase) ?? false;
         }
 
-        public static Credential GetOrganizationCredential(this ICollection<Credential> credentials)
+        public static Credential GetAzureActiveDirectoryCredential(this ICollection<Credential> credentials)
         {
             return credentials.SingleOrDefault(c => IsAzureActiveDirectoryAccount(c.Type));
         }
--- a/src/NuGetGallery/Security/SecurityPolicyService.cs
+++ b/src/NuGetGallery/Security/SecurityPolicyService.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using System.Web;
-using NuGet.Versioning;
 using NuGetGallery.Auditing;
 using NuGetGallery.Configuration;
 using NuGetGallery.Diagnostics;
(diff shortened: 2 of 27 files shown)


src/NuGetGallery/Security/UserSecurityPolicyEvaluationContext.cs
```diff
@@ -12,20 +12,71 @@ namespace NuGetGallery.Security
...
+        public User SourceAccount
+        {
+            get
+            {
+                return _sourceAccount ?? CurrentUser;
+            }
+        }
+
+        /// <summary>
+        /// Account under policy evaluation.
```
skofman1: >ccount under policy evaluation. [](start = 13, length = 31)

the comment is the same for the 3 properties

src/NuGetGallery.Core/CredentialTypes.cs
```diff
@@ -117,5 +117,10 @@ public static bool IsPackageVerificationApiKey(string type)
         {
             return type?.Equals(ApiKey.VerifyV1, StringComparison.OrdinalIgnoreCase) ?? false;
         }
+
+        public static Credential GetOrganizationCredential(this ICollection<Credential> credentials)
```
skofman1: GetAzureActiveDirectoryCredential?

shishirx34:
> On transform: adds policy to organization, using tenant from initial admin's AAD credential

Would this require that we modify the `TenantId` values for existing personal Microsoft accounts? I have done this on 'Dev' database, but perhaps something that might be required in INT and PROD?

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -83,32 +82,57 @@ public virtual IEnumerable<IUserSecurityPolicySubscription> UserSubscriptions
...
-            }
-
             // Evaluate default policies
             if (Configuration.EnforceDefaultSecurityPolicies)
             {
                 var defaultPolicies = DefaultSubscription.Policies;
 
-                var result = await EvaluateInternalAsync(defaultPolicies, httpContext, action, auditSuccess: false);
-                
+                var result = await EvaluateUserPoliciesInternalAsync(action, httpContext, defaultPolicies, auditSuccess: false);
```
shishirx34: any concerns with the `ArgumentNullException` thrown by this method?

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -83,32 +82,57 @@ public virtual IEnumerable<IUserSecurityPolicySubscription> UserSubscriptions
...
+
+        /// <summary>
+        ///  Look up and evaluation of organization security policies for the specified action.
+        /// </summary>
+        public Task<SecurityPolicyResult> EvaluateOrganizationPoliciesAsync(
+            SecurityPolicyAction action,
+            Organization organization,
+            User account)
+        {
+            return EvaluateInternalAsync(action, organization.SecurityPolicies, organization, account, auditSuccess: true);
```
shishirx34: its a bit counter intuitive to term `organization` as the 'source account' and `account` as the 'target account' with respect to `EvaluateInternalAsync` method, the action is `JoinOrganization` which would mean `account` is joining `organization` which would mean `account` should be source account and `organization` to be target account. Don't know if its just me though.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -327,14 +343,33 @@ public bool CanTransformUserToOrganization(User accountToTransform, User adminUs
...
         public async Task<bool> TransformUserToOrganization(User accountToTransform, User adminUser, string token)
         {
-            // todo: check for tenantId and add organization policy to enforce this (future work, with manage organization)
+            var tenantId = adminUser.Credentials.GetOrganizationCredential()?.TenantId;
+            if (string.IsNullOrWhiteSpace(tenantId))
+            {
+                return false;
+            }
 
+            var tenantPolicy = RequireOrganizationTenantPolicy.Create(accountToTransform.Key, tenantId);
```
shishirx34: would it make sense to have mapping of `Username => tenantId` for organization rather than the `Key`. Wouldn't make the change if its too big, just wondering.

src/NuGetGallery/Security/SecurityPolicyService.cs
chenriksson: Previous user security policies (secure push) required the `HttpContext` for getting the current user and request. Organization policies require different context (organization, member) that doesn't come from `HttpContext`. This is managed in the `UserSecurityPolicyEvaluationContext`, but the higher level APIs (`EvaluateUserPolicies`, `EvaluateOrganizationPolicies`) just take what they require. So, not concerned with the exception.

Now that secure push is default, we don't have any user policies... so this code path is pretty dead anyways. I can add some comments in the code.

src/NuGetGallery/Security/SecurityPolicyService.cs
chenriksson: I used `source` for the account where the policy came from, and `target` for the account being evaluated. I can add doc comments to make this clearer.

src/NuGetGallery/Security/SecurityPolicyService.cs
chenriksson: Just noticed I left a null default for context, which doesn't make sense here. Will remove.

src/NuGetGallery/Services/UserService.cs
chenriksson: I can remove `Key` altogether. The policy itself already has a FK to the Users table.

chenriksson:
Thanks @skofman1 , @shishirx34 !

PR #5414, opened 2018-02-07
Add configuration list to determine which users have a discontinued password login

scottbommarito:
This PR adds the `ILoginDeprecationService` which has the method `IsLoginDiscontinuedAsync` that can be used to determine if an `AuthenticatedUser` is using a discontinued login.

The implementation currently checks if the user is on a whitelist loaded from a json configuration file stored alongside the existing Content files.


src/NuGetGallery/App_Data/Files/Content/Password-Login-Discontinuation-Configuration.json
```diff
@@ -0,0 +1,8 @@
+{
+    "DiscontinuedForDomains": [
+        "cannotUsePassword.com"
+    ],
+    "ExceptionsForEmailAddresses": [
+        "exception@cannotUsePassword.com"
+    ]
+}
```
chenriksson: Why do we need this dummy file in Gallery?

src/NuGetGallery/Services/LoginDeprecationService.cs
```diff
@@ -0,0 +1,62 @@
...
+        public async Task<bool> IsLoginDiscontinuedAsync(AuthenticatedUser authenticatedUser)
+        {
+            return 
+                authenticatedUser.CredentialUsed.IsPassword() && 
+                (await GetPasswordDiscontinuationConfigurationAsync()).IsPasswordLoginDiscontinuedForUser(authenticatedUser.User);
+        }
+
+        private async Task<PasswordLoginDiscontinuationConfiguration> GetPasswordDiscontinuationConfigurationAsync()
+        {
+            var configString = (await _contentService.GetContentItemAsync(Constants.ContentNames.PasswordLoginDiscontinuationConfiguration, TimeSpan.FromHours(1))).ToString();
```
chenriksson: This will use blob storage in PROD? Which container?

src/NuGetGallery/Services/LoginDeprecationService.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            _contentService = contentService;
+        }
+
+        public async Task<bool> IsLoginDiscontinuedAsync(AuthenticatedUser authenticatedUser)
+        {
+            return 
+                authenticatedUser.CredentialUsed.IsPassword() && 
+                (await GetPasswordDiscontinuationConfigurationAsync()).IsPasswordLoginDiscontinuedForUser(authenticatedUser.User);
+        }
```
chenriksson: This config should be cached, with expiration.

chenriksson: This config should also replace `AppConfig.OrganizationsEnabledForDomains` and be used as the white list for organization transformation. Can that be added to the PR?

src/NuGetGallery/App_Data/Files/Content/Password-Login-Discontinuation-Configuration.json
scottbommarito: Just for local testing purposes. By creating an account with an email matching this config you can easily test the flow yourself.

src/NuGetGallery/Services/LoginDeprecationService.cs
scottbommarito: The `_contentService` handles the caching, although it still serializes and deserializes the object on every call. I didn't think this was too big of a concern given that the configuration file will be short (one domain and a small number of exceptions) and it will only be used for a month or so.

src/NuGetGallery/Services/LoginDeprecationService.cs
```diff
@@ -0,0 +1,62 @@
...
+            var configString = (await _contentService.GetContentItemAsync(Constants.ContentNames.PasswordLoginDiscontinuationConfiguration, TimeSpan.FromHours(1))).ToString();
+            if (string.IsNullOrEmpty(configString))
+            {
+                return new PasswordLoginDiscontinuationConfiguration(Enumerable.Empty<string>(), Enumerable.Empty<string>());
+            }
+
+            return JsonConvert.DeserializeObject<PasswordLoginDiscontinuationConfiguration>(configString);
+        }
+
+        public class PasswordLoginDiscontinuationConfiguration
```
loic-sharma: Should `PasswordLoginDiscontinuationConfiguration` be private instead? No one else seems to reference it.

scottbommarito: used by tests i believe, probably could be internal then

src/NuGetGallery/Services/LoginDeprecationService.cs
loic-sharma: I would cache a reference to the object and only deserialize once an hour too.

src/NuGetGallery/Services/LoginDeprecationService.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            public IEnumerable<string> DiscontinuedForDomains { get; }
```
loic-sharma: I would consider turning these into `HashSet`s for better performance.

scottbommarito: done

src/NuGetGallery/Services/LoginDiscontinuationAndMigrationConfiguration.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json;
+using NuGetGallery.Authentication;
+
+namespace NuGetGallery
+{
+    public class LoginDiscontinuationAndMigrationConfiguration : ILoginDiscontinuationAndMigrationConfiguration
```
chenriksson: nit: name is really long... how about `DiscontinuedLoginConfiguration` to match the `allowDiscontinuedLogin` on `UIAuthorize`?

src/NuGetGallery/Services/LoginDiscontinuationAndMigrationConfiguration.cs
```diff
@@ -0,0 +1,47 @@
...
+            return
+                authUser.CredentialUsed.IsPassword() &&
+                AreOrganizationsSupportedForUser(authUser.User) &&
+                !ExceptionsForEmailAddresses.Contains(email.Address);
+        }
+
+        public bool AreOrganizationsSupportedForUser(User user)
+        {
+            var email = user.ToMailAddress();
+            return DiscontinuedForDomains.Contains(email.Host, StringComparer.OrdinalIgnoreCase);
```
chenriksson: The link/transform prompt and transform account should use the same whitelist (domains and emails)... they should be consistent.

Could we keep the config simple, and just have a single, shared API? i.e.:
```
public interface IDiscontinuedLoginConfig
{
    HashSet<string> DiscontinuedForDomains { get; }
    HashSet<string> DiscontinuedForEmailAddresses { get; }
}
public class ContentObjectService : IContentObjectService
{
    public IDiscontinuedLoginConfig DiscontinuedLoginConfig { get; }
}
public class AuthenticationService
{
    // called by both transform and AAD link/transform prompt
    public bool IsLoginDiscontinued(User user);
}
```

tests/NuGetGallery.Facts/Services/ContentObjectServiceFacts.cs
```diff
@@ -0,0 +1,54 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Moq;
+using Newtonsoft.Json;
+using NuGetGallery.Framework;
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Web;
+using Xunit;
```
chenriksson: nit: sort usings

src/NuGetGallery/Services/LoginDiscontinuationAndMigrationConfiguration.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    public interface ILoginDiscontinuationAndMigrationConfiguration
```
chenriksson: nit: separate file per interface / class

tests/NuGetGallery.Facts/Services/LoginDiscontinuationAndMigrationConfigurationFacts.cs
```diff
@@ -0,0 +1,103 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGetGallery.Authentication;
+using NuGetGallery.Framework;
+using System.Collections.Generic;
+using Xunit;
```
chenriksson: nit: sort usings

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -1977,6 +1980,7 @@
     <Content Include="Areas\Admin\Views\Delete\_ReflowBulk.cshtml" />
     <Content Include="Areas\Admin\Views\Delete\_ReflowBulkConfirm.cshtml" />
     <Content Include="Areas\Admin\Views\LockPackage\Index.cshtml" />
+    <Content Include="App_Data\Files\Content\Login-Discontinuation-Configuration.json" />
```
chenriksson: What's the default for CopyToOutputDirectory? Is this getting copied (I recommend we don't)?

src/NuGetGallery/Services/LoginDiscontinuationAndMigrationConfiguration.cs
scottbommarito: Makes sense.

src/NuGetGallery/Services/LoginDiscontinuationAndMigrationConfiguration.cs
scottbommarito: Are you sure that users that are given exceptions should be prevented from transforming into an organization? I don't see a reason to prevent them to.

Also IMO it seems odd to call `IsLoginDiscontinued` for an organization support check.

chenriksson: @anangaur said he wanted to test organizations with 3 team accounts before rolling it out to others. My take on that is that we should only allow the prompt and onboarding for those 3, and can open both to the full domain once @anangaur gives us the go-ahead.

chenriksson: The organizations feature is required for disabling passwords. If it's the same check, I would keep a single API.

If we plan to open organizations to everyone, but not deprecate passwords for everyone... then it may be useful to have separate APIs and separate white lists for each. I would check with @anangaur .

src/NuGetGallery/NuGetGallery.csproj
scottbommarito: I don't see this in the output directory so I think it's fine. Also all our other content files are included in the same way and don't get copied either AFAIK.

src/NuGetGallery/Services/ContentObjectService.cs
```diff
@@ -0,0 +1,44 @@
...
+        public async Task Refresh()
+        {
+            LoginDiscontinuationConfiguration = 
+                await Refresh<LoginDiscontinuationConfiguration>(Constants.ContentNames.LoginDiscontinuationConfiguration);
+        }
+
+        private async Task<T> Refresh<T>(string contentName) 
+            where T : class
+        {
+            var configString = (await _contentService.GetContentItemAsync(contentName, TimeSpan.FromHours(1))).ToString();
```
chenriksson: 1 hour is cache expiration? Would move to top of file, i.e., `const string RefreshIntervalInHours = 1`

tests/NuGetGallery.Facts/Services/ContentObjectServiceFacts.cs
chenriksson: System first

src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
```diff
@@ -0,0 +1,52 @@
...
+            ExceptionsForEmailAddresses = new HashSet<string>(exceptionsForEmailAddresses);
+        }
+
+        public bool IsLoginDiscontinued(AuthenticatedUser authUser)
+        {
+            var email = authUser.User.ToMailAddress();
+            return
+                authUser.CredentialUsed.IsPassword() &&
+                AreOrganizationsSupportedForUser(authUser.User) &&
+                !ExceptionsForEmailAddresses.Contains(email.Address);
```
shishirx34: nit: perhaps check for exception list before checking for blacklist. just an optimization.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -30,14 +30,16 @@ public class UserService : IUserService
...
+            IEntitiesContext entitiesContext,
+            IContentObjectService loginDeprecationService)
             : this()
         {
             Config = config;
             UserRepository = userRepository;
             CredentialRepository = credentialRepository;
             Auditing = auditing;
             EntitiesContext = entitiesContext;
+            ContentObjectService = loginDeprecationService;
```
shishirx34: I would keep the naming consistent, use either `contentObjectService` for field and parameter name.

src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public bool AreOrganizationsSupportedForUser(User user)
+        {
+            var email = user.ToMailAddress();
+            return
+                DiscontinuedForEmailAddresses.Contains(email.Address) ||
+                DiscontinuedForDomains.Contains(email.Host, StringComparer.OrdinalIgnoreCase);
```
shishirx34: flip this `or` check, so that the domains are validated first, then email address. nit optimization.

src/NuGetGallery/Configuration/IAppConfiguration.cs
```diff
@@ -93,11 +93,6 @@ public interface IAppConfiguration : ICoreMessageServiceConfiguration
         /// </summary>
         bool BlockingAsynchronousPackageValidationEnabled { get; set; }
 
-        /// <summary>
-        /// Whitelist of domains for which the Organizations feature is enabled.
-        /// </summary>
-        string[] OrganizationsEnabledForDomains { get; set; }
```
shishirx34: Delete this one from `web.config` as well? Also, please create a task to remove this value from all the deployment configs as well.

src/NuGetGallery/Services/UserService.cs
scottbommarito: woops, this slipped by me after a couple naming changes

src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
scottbommarito: yup good catch

src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
scottbommarito: i think the exception list will actually be larger than the other two lists so checking the exception list last is better IMO, but all three lists will be very small so i don't think it matters

src/NuGetGallery/Configuration/IAppConfiguration.cs
scottbommarito: done

PR #5415, opened 2018-02-07
Add logic for respecting the package auto-delete configuration

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/921.

Could I get some crazy nitpick 👀s on this PR? There is risk associated with this change as it allows users to delete their own packages in some cases.

I check both the DB and the statistics reports for download counts. This is to improve resiliency as separate jobs write these two artifacts.

### DIFF
--- a/src/NuGetGallery/Configuration/IPackageDeleteConfiguration.cs
+++ b/src/NuGetGallery/Configuration/IPackageDeleteConfiguration.cs
@@ -27,7 +27,9 @@ namespace NuGetGallery.Configuration
         /// regardless of the number of downloads. Since the download counts for packages are not updated immediately,
         /// there is a window of time when a new package won't have any recorded downloads. Note that
         /// <see cref="AllowUsersToDeletePackages"/> and <see cref="MaximumDownloadsForPackageVersion"/> take
-        /// precedence over this option.
+        /// precedence over this option. The jobs that imply this value are the Stats.CreateAzureCdnWarehouseReports and
+        /// Stats.AggregateCdnDownloadsInGallery jobs. Note that this configuration value may be higher than the actual
+        /// update frequency. This is the more like the maximum expected update frequency.
         /// </summary>
         int? StatisticsUpdateFrequencyInHours { get; }
 
--- a/src/NuGetGallery/Services/PackageDeleteService.cs
+++ b/src/NuGetGallery/Services/PackageDeleteService.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Data.Entity;
 using System.Data.SqlClient;
 using System.Linq;
 using System.Threading.Tasks;
@@ -42,6 +41,8 @@ namespace NuGetGallery
         private readonly IPackageFileService _packageFileService;
         private readonly IAuditingService _auditingService;
         private readonly IPackageDeleteConfiguration _config;
+        private readonly IStatisticsService _statisticsService;
+        private readonly ITelemetryService _telemetryService;
 
         public PackageDeleteService(
             IEntityRepository<Package> packageRepository,
@@ -52,17 +53,21 @@ namespace NuGetGallery
             IIndexingService indexingService,
             IPackageFileService packageFileService,
             IAuditingService auditingService,
-            IPackageDeleteConfiguration config)
+            IPackageDeleteConfiguration config,
+            IStatisticsService statisticsService,
+            ITelemetryService telemetryService)
         {
-            _packageRepository = packageRepository;
-            _packageRegistrationRepository = packageRegistrationRepository;
-            _packageDeletesRepository = packageDeletesRepository;
-            _entitiesContext = entitiesContext;
-            _packageService = packageService;
-            _indexingService = indexingService;
-            _packageFileService = packageFileService;
-            _auditingService = auditingService;
-            _config = config;
+            _packageRepository = packageRepository ?? throw new ArgumentNullException(nameof(packageRepository));
+            _packageRegistrationRepository = packageRegistrationRepository ?? throw new ArgumentNullException(nameof(packageRegistrationRepository));
+            _packageDeletesRepository = packageDeletesRepository ?? throw new ArgumentNullException(nameof(packageDeletesRepository));
+            _entitiesContext = entitiesContext ?? throw new ArgumentNullException(nameof(entitiesContext));
+            _packageService = packageService ?? throw new ArgumentNullException(nameof(packageService));
+            _indexingService = indexingService ?? throw new ArgumentNullException(nameof(indexingService));
+            _packageFileService = packageFileService ?? throw new ArgumentNullException(nameof(packageFileService));
+            _auditingService = auditingService ?? throw new ArgumentNullException(nameof(auditingService));
+            _config = config ?? throw new ArgumentNullException(nameof(config));
+            _statisticsService = statisticsService ?? throw new ArgumentNullException(nameof(statisticsService));
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
 
             if (config.HourLimitWithMaximumDownloads.HasValue
                 && config.StatisticsUpdateFrequencyInHours.HasValue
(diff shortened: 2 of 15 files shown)


src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -74,16 +76,100 @@ public class PackageDeleteService
...
-            return Task.FromResult(false);
+            var now = DateTime.UtcNow;
+            var sinceCreated = now - package.Created;
+
+            // Always refresh the statistics.
+            await _statisticsService.Refresh();
+
+            // Handle the "early" delete case, where the package version download count is not considered but total
+            // download count on the entire ID (all versions) is considered.
+            if (_config.StatisticsUpdateFrequencyInHours.HasValue
```
scottbommarito: IMO the name `StatisticsUpdateFrequencyInHours` is a bit misleading...potentially something along the lines of `ExpectedStatisticsUpdateFrequencyInHours` would be more accurately convey that stats are not GUARANTEED to be updated every X hours.

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -74,16 +76,100 @@ public class PackageDeleteService
...
             }
 
-            // For now, don't allow user's to delete their packages.
-            // https://github.com/NuGet/Engineering/issues/921
-            return Task.FromResult(false);
+            var now = DateTime.UtcNow;
+            var sinceCreated = now - package.Created;
+
+            // Always refresh the statistics.
+            await _statisticsService.Refresh();
```
scottbommarito: So we will refresh stats every time someone goes to the `ReportMyPackage` page or submit the `ReportMyPackage` form?

I'm not sure about the logistics of this method (does it always refresh or is it just if the cache is stale?) but if this refresh is not cached, somebody could create a lot of turmoil on the site by hitting either of these two methods repeatedly.

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -74,16 +76,100 @@ public class PackageDeleteService
...
+
+                return true;
+            }
+
+            // If no time ranges are configured, allow downloads any time.
+            return !_config.HourLimitWithMaximumDownloads.HasValue
+                && !_config.StatisticsUpdateFrequencyInHours.HasValue;
+        }
+
+        private bool AreStatisticsStale(DateTime now)
```
scottbommarito: This sounds like a method that should be on the `StatisticsService` itself, but if it's too much work to move it there it's fine here.

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -74,16 +76,100 @@ public class PackageDeleteService
...
+            if (_config.StatisticsUpdateFrequencyInHours.HasValue
+                && sinceCreated < TimeSpan.FromHours(_config.StatisticsUpdateFrequencyInHours.Value))
+            {
+                if (_config.MaximumDownloadsForPackageId.HasValue)
+                {
+                    // Do not allow a delete of a package version if the package registration record has too many downloads.
+                    if (package.PackageRegistration.DownloadCount > _config.MaximumDownloadsForPackageId.Value)
+                    {
+                        return false;
+                    }
```
skofman1: There are cases when we can avoid _statisticsService.Refresh() which is expensive. Restructure the code to be cheaper.

joelverhagen: True, I really only need it before the stale check.

src/NuGetGallery/Services/PackageDeleteService.cs
skofman1: >_statisticsService.Refresh() [](start = 18, length = 28)

We will want to analyze the success of the feature later on and understand how to improve it. Add traces that help understand why we returned false. Consider structuring if in a way that will allow creating a table of input properties and result.

joelverhagen: Great idea!

joelverhagen: The refresh is cached and the cache is a singleton.

src/NuGetGallery/Services/PackageDeleteService.cs
joelverhagen: `Expected` is good. I'll add that.

src/NuGetGallery/Services/PackageDeleteService.cs
joelverhagen: Stale is a relative thing. Different scenarios may or may not have different "stale" definitions. For now the concept only exists in the auto-delete workflow.

src/NuGetGallery/Services/PackageDeleteService.cs
joelverhagen: Fixed.

src/NuGetGallery/Services/PackageDeleteService.cs
joelverhagen: Fixed.

src/NuGetGallery/Services/PackageDeleteService.cs
joelverhagen: Fixed.

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -52,38 +53,176 @@ public class PackageDeleteService
...
             {
-                throw new ArgumentException($"{nameof(_config.StatisticsUpdateFrequencyInHours)} must be less than " +
+                throw new ArgumentException($"{nameof(_config.ExpectedStatisticsUpdateFrequencyInHours)} must be less than " +
                     $"{nameof(_config.HourLimitWithMaximumDownloads)}.",
                     nameof(config));
             }
         }
 
-        public Task<bool> CanPackageBeDeletedByUserAsync(Package package)
+        public async Task<bool> CanPackageBeDeletedByUserAsync(Package package)
```
joelverhagen: Thoughts on only emitting telemetry when the support request form is actually submitted?

This would reduce the amount of de-duping we would have to do on the telemetry -- but not eliminate all duplicates, since a user could support multiple support requests for a package. I figured it would just be simpler to submit telemetry whenever this method is called and de-dupe later when we are analyzing the data. 

As it stands today, telemetry is emitted before the support request form is loaded and after the form is submitted with a deletion request.

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -52,38 +53,176 @@ public class PackageDeleteService
...
+
+            if (package.PackageRegistration.IsLocked)
+            {
+                return IsAccepted(details, UserPackageDeleteOutcome.LockedRegistration);
+            }
+
+            // Handle the "early" delete case, where the package version download count is not considered but total
+            // download count on the entire ID (all versions) is considered.
+            if (_config.ExpectedStatisticsUpdateFrequencyInHours.HasValue
+                && details.SinceCreated < TimeSpan.FromHours(_config.ExpectedStatisticsUpdateFrequencyInHours.Value))
```
chenriksson: Does this mean owners of popular packages that want to delete will need to unlist immediately, then can delete after the hour limit has expired?

src/NuGetGallery/Configuration/IPackageDeleteConfiguration.cs
```diff
@@ -29,23 +29,23 @@ public interface IPackageDeleteConfiguration
...
         /// <summary>
         /// The hour threshold for the "late" delete criteria. If a package was pushed less than this many hours ago
         /// and if the package version has less than <see cref="MaximumDownloadsForPackageVersion"/> downloads, the
         /// package can be deleted. If this value is null, the package cannot be deleted after
-        /// <see cref="StatisticsUpdateFrequencyInHours"/> hours. Also, if the package has not had its download
-        /// statistics updated in the last <see cref="StatisticsUpdateFrequencyInHours"/> hours, the delete is not
-        /// allowed since the download data is too stale. If <see cref="StatisticsUpdateFrequencyInHours"/> is null,
+        /// <see cref="ExpectedStatisticsUpdateFrequencyInHours"/> hours. Also, if the package has not had its download
+        /// statistics updated in the last <see cref="ExpectedStatisticsUpdateFrequencyInHours"/> hours, the delete is not
+        /// allowed since the download data is too stale. If <see cref="ExpectedStatisticsUpdateFrequencyInHours"/> is null,
```
chenriksson: Should this be configured based on the frequency of the stats jobs? Should we have a specific recommendation in comments, based on job?

src/NuGetGallery/Telemetry/UserPackageDeleteOutcome.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        Accepted,
+        AlreadyDeleted,
+        LockedRegistration,
+        TooManyIdDatabaseDownloads,
+        TooManyIdReportDownloads,
+        StaleStatistics,
+        TooManyVersionDatabaseDownloads,
+        TooManyVersionReportDownloads,
+        TooLate,
```
chenriksson: Will the UI always show an option to delete, but present the error message if the criteria aren't met?

Or do we check before showing a link, and after / on submit?

src/NuGetGallery/Services/PackageDeleteService.cs
chenriksson: Sounds like we'll do the check before, and only show the link if it's possible? I think we should only submit telemetry on submit, not view rendering.

joelverhagen: > Sounds like we'll do the check before, and only show the link if it's possible?

Correct.

>  I think we should only submit telemetry on submit, not view rendering.

Yeah, that's where I'm leaning too.

src/NuGetGallery/Configuration/IPackageDeleteConfiguration.cs
joelverhagen: > Should this be configured based on the frequency of the stats jobs?

That's not too necessary. As mentioned in the XML doc, this value is used for a stale check (so the value can be larger than the average frequency).

> Should we have a specific recommendation in comments, based on job?

Yeah, I'll try to improve this.

src/NuGetGallery/Services/PackageDeleteService.cs
joelverhagen: Unlisted is not a factor here.

Owners of popular packages (IDs > 125,000 downloads) cannot delete their packages at all.

src/NuGetGallery/Telemetry/UserPackageDeleteOutcome.cs
joelverhagen: > Will the UI always show an option to delete, but present the error message if the criteria aren't met?

Nope, we only show the option if at the time of form load their package is eligible. 

> Or do we check before showing a link, and after / on submit?

Correct. And if the eligibility has changed since the form was presented, we just treat it as a normal support request.

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -52,38 +53,176 @@ public class PackageDeleteService
...
+                idReportDownloads,
+                package.DownloadCount,
+                versionReportDownloads);
+
+            return details;
+        }
+
+        private bool HasVersion(StatisticsFact fact, string version)
+        {
+            if (fact != null
```
loic-sharma: >if [](start = 12, length = 2)

Consider `return xyz;` instead of `if (xyz) { return true } return false;`,

src/NuGetGallery/Services/PackageDeleteService.cs
joelverhagen: Actually, I decided to not make this change as it would require a config change in NuGetDeployment. This is not worth the effort.

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -52,38 +53,176 @@ public class PackageDeleteService
...
+                    {
+                        return IsAccepted(details, UserPackageDeleteOutcome.TooManyVersionReportDownloads);
+                    }
+                }
+
+                return IsAccepted(details, UserPackageDeleteOutcome.Accepted);
+            }
+
+            // If no time ranges are configured, allow downloads any time.
+            if (_config.HourLimitWithMaximumDownloads.HasValue
```
loic-sharma: >if [](start = 12, length = 2)

Could we convert the chains of `if` to `else if`s? It'd be a little more clear I think.

src/NuGetGallery/Services/PackageDeleteService.cs
joelverhagen: Sure

src/NuGetGallery/Services/TelemetryClientWrapper.cs
```diff
@@ -52,5 +52,17 @@ public void TrackException(Exception exception, IDictionary<string, string> prop
...
+
+        public void TrackMetric(string metricName, double value, IDictionary<string, string> properties = null)
+        {
+            try
+            {
+                UnderlyingClient.TrackMetric(metricName, value, properties);
+            }
+            catch
+            {
+                // logging failed, don't allow exception to escape
```
loic-sharma: >// logging failed, don't allow exception to escape [](start = 16, length = 50)

Should we maybe log these exceptions?

joelverhagen: Out of scope.

joelverhagen: https://github.com/NuGet/NuGetGallery/issues/5419

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -250,6 +258,27 @@ private void TrackPackageForEvent(string eventValue, string packageId, string pa
...
+        public void TrackUserPackageDelete(UserPackageDeleteEvent details, UserPackageDeleteOutcome outcome)
+        {
+            if (details == null)
+            {
+                throw new ArgumentNullException(nameof(details));
+            }
+
+            var hours = details.SinceCreated.TotalHours;
+
+            TrackMetric(Events.UserPackageDeleteAfterHours, hours, properties => {
```
loic-sharma: >properties [](start = 67, length = 10)

The action here seems unnecessary. I think this just creating a dictionary and using its initializer syntax would be better.

joelverhagen: Existing pattern.

joelverhagen: (it saves the initialization of the dictionary)

src/NuGetGallery/Services/PackageDeleteService.cs
joelverhagen: Sure

PR #5416, opened 2018-02-07
Tool to hash active api key credentials

skofman1:


src/GalleryTools/Program.cs
```diff
@@ -0,0 +1,84 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using System;
+using System.Threading.Tasks;
+using NuGetGallery;
+using NuGetGallery.Infrastructure.Authentication;
+using Microsoft.Extensions.CommandLineUtils;
```
chenriksson: nit: sort

src/GalleryTools/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            var hash = commandLineApplication.Command("hash", config =>
+            {
+                config.Description = "Hash V1/V2 ApiKeys in Gallery DB";
+                CommandOption connectionString = config.Option("-db", "DB connection string. User should have write permission.", CommandOptionType.SingleValue);
+                CommandOption whatIf = config.Option("-w | --whatif", "Don't apply changes to DB.", CommandOptionType.NoValue);
+                config.HelpOption("-? | -h | --help");
+                config.OnExecute(() =>
+                {
+                    Hash(connectionString.Value(), whatIf.HasValue()).Wait();
```
joelverhagen: `GetAwaiter().GetResult()` provides better exceptions

joelverhagen: Also, we should catch all exceptions and log the error rather than crashing.

src/GalleryTools/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+                    return 0;
+                });
+            });
+
+            commandLineApplication.Execute(args);
+        }
+
+        private static async Task Hash(string connectionString, bool whatIf)
+        {
+            IEntitiesContext context = new EntitiesContext(connectionString, readOnly: whatIf);
```
joelverhagen: `using`

src/GalleryTools/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+            var allCredentials = (IQueryable<Credential>)context.Set<Credential>();
```
joelverhagen: Can we use the `IEntityRepository` abstraction here? This would make it look more like gallery code.

src/GalleryTools/GalleryTools.csproj
```diff
@@ -0,0 +1,74 @@
...
+    <ProjectReference Include="..\NuGetGallery.Core\NuGetGallery.Core.csproj">
+      <Project>{097b2cdd-9623-4c34-93c2-d373d51f5b4e}</Project>
+      <Name>NuGetGallery.Core</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\NuGetGallery\NuGetGallery.csproj">
+      <Project>{1dacf781-5cd0-4123-8bac-cd385d864be5}</Project>
+      <Name>NuGetGallery</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
```
joelverhagen: Where are we executing this from? We may need to strong name sign the tool... which means gallery will need strong name signing (unless we can move all deps to NuGetGallery.Core).

src/GalleryTools/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+            // V1/V2 credentials that have a future expiration date
+
+            var validLastUsed = DateTime.UtcNow - TimeSpan.FromDays(365);
+            var activeCredentials = 
+                allCredentials.Where(x => (x.Type == CredentialTypes.ApiKey.V1 || x.Type == CredentialTypes.ApiKey.V2) &&
+                                          ((x.Expires == null && x.LastUsed != null && x.LastUsed > validLastUsed) ||
+                                           (x.Expires != null && x.Expires > DateTime.UtcNow)))
+                                          .ToList();
+
+            Console.WriteLine($"Found {activeCredentials.Count} active V1/V2 ApiKeys. Ids: " +
```
joelverhagen: super-nit: `IDs`

src/GalleryTools/Program.cs
skofman1: ha?

src/GalleryTools/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+                    
+                    Console.WriteLine($"Key: {v1v2ApiKey.Key} Type: {v1v2ApiKey.Type} Old value: {v1v2ApiKey.Value} New value: {hashedApiKey.HashedApiKey}");
+
+                    v1v2ApiKey.Type = CredentialTypes.ApiKey.V3;
+                    v1v2ApiKey.Value = hashedApiKey.HashedApiKey;
+                }
+                catch (ArgumentException e)
+                {
+                    failures = true;
+                    Console.WriteLine($"Failed to hash key: {v1v2ApiKey.Key} Type: {v1v2ApiKey.Type} Old value: {v1v2ApiKey.Value}. Reason: {e.Message}");
```
loic-sharma: Important info may be in the `InnerException`, you should log the full exception instead of just its message

src/GalleryTools/GalleryTools.csproj
skofman1: We will execute this the same way we execute DB migrations..

src/GalleryTools/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+                }
+            }
+
+            if (activeCredentials.Count > 0 && !failures && !whatIf)
+            {
+                await context.SaveChangesAsync();
```
loic-sharma: Should we maybe limit how many credentials can be updated in a single save? Maybe we should have chunks of 100 credential updates?

src/GalleryTools/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="EntityFramework" version="6.2.0" targetFramework="net46" />
```
loic-sharma: Could we use PackageReference instead? :)

src/GalleryTools/Program.cs
skofman1: Added batching!

src/GalleryTools/packages.config
skofman1: Thought about.. prefer not to spend the time

src/GalleryTools/Program.cs
skofman1: don't see a point... it will add additional code and since we don't need UTs for the tool, there's no benefit.

src/GalleryTools/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                var allCredentials = (IQueryable<Credential>)context.Set<Credential>();
+
+                // Get all V1/V2 credentials that are active: 
+                // V1 credentials that have no expiration date, but were used in the last year
+                // V1/V2 credentials that have a future expiration date
+
+                var validLastUsed = DateTime.UtcNow - TimeSpan.FromDays(365);
+                Expression<Func<Credential, bool>> predicate = x => (x.Type == CredentialTypes.ApiKey.V1 || x.Type == CredentialTypes.ApiKey.V2)/* &&
+                                                                ((x.Expires == null && x.LastUsed != null && x.LastUsed > validLastUsed) ||
```
joelverhagen: Is that meant to be commented out?

skofman1: LOL... good catch.. that would have hashed 90K extra values

src/GalleryTools/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+                                v1v2ApiKey.Value = hashedApiKey.HashedApiKey;
+                            }
+                            catch (ArgumentException e)
+                            {
+                                failures = true;
+                                Console.WriteLine($"Failed to hash key: {v1v2ApiKey.Key} Type: {v1v2ApiKey.Type} Old value: {v1v2ApiKey.Value}. Reason: {e}");
+                            }
+                        }
+
+                        if (batch.Count > 0 && !failures && !whatIf)
```
loic-sharma: >batch.Count [](start = 28, length = 11)

`batch.Count` cannot be 0 here, right? If so, this check is unnecessary.

PR #5420, opened 2018-02-08
Remove SecurePushSubscription

skofman1:
https://github.com/NuGet/Engineering/issues/800

Cleaning dead code:
Removing SecurePushSubscription and RequireSecurePushForCoOwnersPolicy.
I removed only the backend components. Didn't touch UI and e-mail that support sending policy messages.


skofman1:
@chenriksson , can you take another look? I merged latest changes from DEV. Regarding the UI, I left it in case we will add new policies later on. I can hide it if you prefer.

chenriksson:
Could remove `policyMessage` altogether (i.e., from `page-manage-owners.js` ajax calls)... but LGTM either way.

Issue #5412, opened 2018-02-06
[Org/AAD] Add close option for dialog


shishirx34:
Allow to close the dialog so that the user can sign out.

Issue #5387, opened 2018-02-03
[Org/AAD] Integrate Tranform dialog with the whitelisted users.


shishirx34:
- Remove the 'allowLinkingAnyEmail' from GET route and create a new route for MSA/AAD from transform route.

PR #5421, opened 2018-02-08
[Org/AAD] Transform Dialog integration for certain accounts

shishirx34:
This PR addresses #5387 and #5412 

- Allows popup dialog dismissal.
- Shows the transform popup dialog for whitelisted accounts.
- Refactored code for reusability with signing/linking MSA/AAD accounts.

image

I will add tests as I go along.

### DIFF
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -324,7 +324,7 @@ namespace NuGetGallery.Authentication
             }
 
             // Check user credentials for existing cred for optimization to avoid expensive DB query
-            if (UserHasCredential(user, credential) || FindMatchingCredential(credential) != null)
+            if (!user.HasCredential(credential) && FindMatchingCredential(credential) != null)
             {
                 // Existing credential for a registered account
                 return false;
@@ -332,7 +332,16 @@ namespace NuGetGallery.Authentication
 
             try
             {
-                await ReplaceCredential(user, credential);
+                if (user.IsAdministrator())
+                {
+                    // For administrators add the credential do not replace until we aggregate AAD providers for admins
+                    await AddCredential(user, credential);
+                }
+                else
+                {
+                    await ReplaceCredential(user, credential);
+                }
+
                 return true;
             }
             catch (InvalidOperationException)
@@ -698,13 +707,6 @@ namespace NuGetGallery.Authentication
 
             await Auditing.SaveAuditRecordAsync(new UserAuditRecord(
                 user, AuditedUserAction.AddCredential, credential));
-        }
-
-        private static bool UserHasCredential(User user, Credential credential)
-        {
-            return user.Credentials.Any(cred =>
-                cred.Type.Equals(credential.Type, StringComparison.OrdinalIgnoreCase)
-                && cred.Value == credential.Value);
         }
 
         private static CredentialKind GetCredentialKind(string type)
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -332,16 +332,7 @@ namespace NuGetGallery.Authentication
 
             try
             {
-                if (user.IsAdministrator())
-                {
-                    // For administrators add the credential do not replace until we aggregate AAD providers for admins
-                    await AddCredential(user, credential);
-                }
-                else
-                {
-                    await ReplaceCredential(user, credential);
-                }
-
+                await ReplaceCredential(user, credential);
                 return true;
             }
             catch (InvalidOperationException)
--- a/src/NuGetGallery/Authentication/Providers/AzureActiveDirectory/AzureActiveDirectoryAuthenticator.cs
+++ b/src/NuGetGallery/Authentication/Providers/AzureActiveDirectory/AzureActiveDirectoryAuthenticator.cs
@@ -8,7 +8,6 @@ using System.Threading.Tasks;
 using System.Web.Mvc;
 using Microsoft.IdentityModel.Protocols;
 using Microsoft.Owin.Security.OpenIdConnect;
-using NuGetGallery.Authentication.Providers.Utils;
 using NuGetGallery.Configuration;
 using Owin;
 
@@ -84,7 +83,7 @@ namespace NuGetGallery.Authentication.Providers.AzureActiveDirectory
 
         public override IdentityInformation GetIdentityInformation(ClaimsIdentity claimsIdentity)
         {
-            return ClaimsExtentions.GetIdentityInformation(claimsIdentity, DefaultAuthenticationType, ClaimTypes.NameIdentifier, ClaimTypeName, ClaimTypes.Name);
+            return ClaimsExtensions.GetIdentityInformation(claimsIdentity, DefaultAuthenticationType, ClaimTypes.NameIdentifier, ClaimTypeName, ClaimTypes.Name);
         }
     }
 }
--- a/src/NuGetGallery/Authentication/Providers/AzureActiveDirectory/AzureActiveDirectoryAuthenticator.cs
+++ b/src/NuGetGallery/Authentication/Providers/AzureActiveDirectory/AzureActiveDirectoryAuthenticator.cs
@@ -22,9 +22,9 @@ namespace NuGetGallery.Authentication.Providers.AzureActiveDirectory
         {
             // Fetch site root from configuration
             var siteRoot = config.Current.SiteRoot.TrimEnd('/') + "/";
-            
+
             // We *always* require SSL for Azure Active Directory
-            if (siteRoot.StartsWith("http://", StringComparison.OrdinalIgnoreCase)) 
+            if (siteRoot.StartsWith("http://", StringComparison.OrdinalIgnoreCase))
             {
                 siteRoot = siteRoot.Replace("http://", "https://");
             }
@@ -45,14 +45,14 @@ namespace NuGetGallery.Authentication.Providers.AzureActiveDirectory
                         }
 
                         return Task.FromResult(0);
-                    } 
+                    }
                 }
             };
             Config.ApplyToOwinSecurityOptions(options);
 
             app.UseOpenIdConnectAuthentication(options);
         }
-        
+
         public override AuthenticatorUI GetUI()
         {
             return new AuthenticatorUI(
@@ -83,7 +83,12 @@ namespace NuGetGallery.Authentication.Providers.AzureActiveDirectory
 
         public override IdentityInformation GetIdentityInformation(ClaimsIdentity claimsIdentity)
         {
-            return ClaimsExtensions.GetIdentityInformation(claimsIdentity, DefaultAuthenticationType, ClaimTypes.NameIdentifier, ClaimTypeName, ClaimTypes.Name);
+            return ClaimsExtensions.GetIdentityInformation(
+                claimsIdentity,
+                DefaultAuthenticationType,
+                ClaimTypes.NameIdentifier,
+                ClaimTypeName,
+                ClaimTypes.Name);
         }
     }
 }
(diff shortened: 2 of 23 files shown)


src/NuGetGallery/Authentication/Providers/Utils/ClaimsExtentions.cs
```diff
@@ -30,5 +30,17 @@ public static IdentityInformation GetIdentityInformation(ClaimsIdentity claimsId
...
         }
+
+        public static bool HasDiscontinuedLoginCLaims(ClaimsIdentity identity)
+        {
+            if (identity == null || !identity.IsAuthenticated)
+            {
+                return false;
+            }
+
+            var discontinuedLoginClaim = identity.GetClaimOrDefault(NuGetClaims.DiscontinuedLogin);
```
chenriksson: Should we add this claim if AAD/MSA are not enabled in web.config? I noticed when testing locally that I still had my whitelist enabled, but hadn't enabled AAD/MSA so Gallery became non-functional with the UIAuthorize redirects.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -15,6 +15,7 @@
 using NuGetGallery.Authentication.Providers.AzureActiveDirectoryV2;
 using NuGetGallery.Authentication.Providers.MicrosoftAccount;
 using NuGetGallery.Infrastructure.Authentication;
+using NuGetGallery.Authentication.Providers.Utils;
```
chenriksson: nit: sort usings

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -381,8 +376,9 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
         /// after external authentication.
         /// </summary>
         /// <param name="returnUrl">The url to return upon credential replacement</param>
+        /// <param name="isTransform">Set to true if this is a transform route for authentication</param>
```
chenriksson: What does `isTransform` mean (transform from what to what)? Sounds like it's auth related, not org related, right?

I would describe what the transform flow is in more detail, it's not clear to me.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -561,6 +569,39 @@ where p.BaseConfig.Enabled
...
+            // Get list of all enabled providers
+            var providers = GetProviders();
+
+            // Select one provider to authenticate for linking when multiple external providers are enabled.
+            // This is for backwards compatibility with MicrosoftAccount provider. 
+            return ExternalAuthenticationPriority
+                .FirstOrDefault(authenticator => providers.Any(p => p.ProviderName.Equals(authenticator, StringComparison.OrdinalIgnoreCase)));
+        }
+
+        private static bool UserHasCredential(User user, Credential credential)
```
chenriksson: nit: would move to UserExtensions in case we need this check elsewhere

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -561,6 +569,39 @@ where p.BaseConfig.Enabled
...
+
+        private string GetEmailAddressFromExternalLoginResult(AuthenticateExternalLoginResult result, out string errorReason)
+        {
+            try
+            {
+                var userInformation = result.Authenticator?.GetIdentityInformation(result.ExternalIdentity);
+                errorReason = null;
+                return userInformation.Email;
+            }
+            catch (Exception ex)
```
chenriksson: What kind of exceptions do we expect here? Is there a specific exception we can handle instead of all?

src/NuGetGallery/Filters/UiAuthorizeAttribute.cs
```diff
@@ -1,44 +1,36 @@
...
 
-using System;
-using System.Security.Claims;
 using System.Web.Mvc;
 using System.Web.Routing;
-using NuGetGallery.Authentication;
-using AuthenticationTypes = NuGetGallery.Authentication.AuthenticationTypes;
+using System.Security.Claims;
+using NuGetGallery.Authentication.Providers.Utils;
 using AuthorizationContext = System.Web.Mvc.AuthorizationContext;
```
chenriksson: nit: sort usings

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -754,6 +754,21 @@ private static string GetAuthenticationRoutes(this UrlHelper url, string action,
                 interceptReturnUrl: false);
         }
 
+        private static string GetAuthenticationRoutes(this UrlHelper url, string action, string returnUrl, bool isTransform = false, bool relativeUrl = true)
```
chenriksson: Is this returning just 1 route? Why use plural?

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
```diff
@@ -16,6 +16,7 @@
 using NuGetGallery.Authentication.Providers.MicrosoftAccount;
 using NuGetGallery.Infrastructure.Authentication;
 using Xunit;
+using NuGetGallery.Authentication.Providers;
```
chenriksson: nit: sort usings

src/NuGetGallery/Authentication/Providers/Utils/ClaimsExtentions.cs
scottbommarito: I mean the plan is for us to move to AAD/MSA logins exclusively, so this will be how it is in the future regardless. I don't know if we want to support passwords for local debugging or not.

src/NuGetGallery/Controllers/AuthenticationController.cs
scottbommarito: IMO there should be an extension on `Credential` that is basically an equality operator like we've done with `User` with `User.MatchesUser`. `UserHasCredential` can be an extension on `User` that uses this extension...

```
user.Credentials.Any(cred => cred.Matches(credential))
```

src/NuGetGallery/Filters/UiAuthorizeAttribute.cs
```diff
@@ -1,44 +1,36 @@
...
+
         public UIAuthorizeAttribute(bool allowDiscontinuedLogins = false)
         {
             AllowDiscontinuedLogins = allowDiscontinuedLogins;
         }
 
         public override void OnAuthorization(AuthorizationContext filterContext)
         {
-            // If a password credential was used, and the user has a discontinued password claim, redirect them to the homepage.
+            // If a password credential was used, and the user has a discontinued password claim, redirect them to the homepage with modal dialog
```
scottbommarito: We should remove any reference to "password" here, because this claim doesn't explicitly imply passwords are the only form of auth that can be discontinued.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -394,21 +390,33 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider)
 
             var newCredential = result.Credential;
 
-            if (await _authService.TryReplaceCredential(user, newCredential))
+            // For transform flow, check if the user already has the linked external account credential, allow the user to login with this credential.
+            var ignoreCredentialReplacement = isTransform && UserHasCredential(user, newCredential);
+            if (ignoreCredentialReplacement || await _authService.TryReplaceCredential(user, newCredential))
```
scottbommarito: It seems to me like treating the transform scenario differently here isn't necessary. If someone goes through the change credential flow using an account they've already linked, wouldn't we expect it to behave the same as someone who tries to log into their linked account when prompted that their password is no longer viable?

How does `TryReplaceCredential` behave for credentials that the user already has?

src/NuGetGallery/Controllers/PagesController.cs
```diff
@@ -11,6 +11,9 @@
 using System.Web.Mvc;
 using NuGetGallery.Areas.Admin;
 using NuGetGallery.Filters;
+using System.Security.Claims;
+using NuGetGallery.Authentication;
+using NuGetGallery.Authentication.Providers.Utils;
```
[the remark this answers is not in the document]
shishirx34: we have kind of inconsistencies of sorting usings, alphabetical vs aesthetic. I don't have any issue with either, will sort alphabetical.

src/NuGetGallery/Authentication/Providers/Utils/ClaimsExtentions.cs
```diff
@@ -30,5 +30,17 @@ public static IdentityInformation GetIdentityInformation(ClaimsIdentity claimsId
             var emailClaim = claimsIdentity.FindFirst(emailClaimType);
```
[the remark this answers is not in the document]
scottbommarito: Didn't even notice, how funny.

src/NuGetGallery/Controllers/PagesController.cs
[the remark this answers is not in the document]
scottbommarito: I think we agreed to `System` first as a team. I don't really care however.

src/NuGetGallery/Authentication/Providers/Utils/ClaimsExtentions.cs
shishirx34: I think that would be part of the Scott's work when adding the claim, but I think we shouldn't worry about it, simply removing the whitelist should make it work locally as well, right? We will be supporting passwords for local debugging, I have a task to put deprecation behind a config.

src/NuGetGallery/Controllers/PagesController.cs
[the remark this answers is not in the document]
chenriksson: Sorting System first is the common standard internally.

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: Had an offline discussion, and I think it makes sense to allow replacing an existing credential which belongs to the user(same account) this will allow users to replace multiple external credentials with a single credential that is an existing one. This will also simplify the transform flow and we can reuse the same controller for both removing the need for `isTransform` completely.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -324,15 +324,24 @@ public virtual Task ReplaceCredential(string username, Credential credential)
...
                 return false;
             }
 
             try
             {
-                await ReplaceCredential(user, credential);
+                if (user.IsAdministrator())
+                {
+                    // For administrators add the credential do not replace until we aggregate AAD providers for admins
+                    await AddCredential(user, credential);
```
chenriksson: Different behavior for admins... should we note this for the bug bash?

src/NuGetGallery/Authentication/Providers/AzureActiveDirectory/AzureActiveDirectoryAuthenticator.cs
```diff
@@ -84,7 +83,7 @@ public override ActionResult Challenge(string redirectUrl)
 
         public override IdentityInformation GetIdentityInformation(ClaimsIdentity claimsIdentity)
         {
-            return ClaimsExtentions.GetIdentityInformation(claimsIdentity, DefaultAuthenticationType, ClaimTypes.NameIdentifier, ClaimTypeName, ClaimTypes.Name);
+            return ClaimsExtensions.GetIdentityInformation(claimsIdentity, DefaultAuthenticationType, ClaimTypes.NameIdentifier, ClaimTypeName, ClaimTypes.Name);
```
chenriksson: nit: would wrap long lines

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -595,7 +583,7 @@ private string GetEmailAddressFromExternalLoginResult(AuthenticateExternalLoginR
                 errorReason = null;
                 return userInformation.Email;
             }
-            catch (Exception ex)
+            catch (ArgumentException ex)
```
chenriksson: `ArgumentException` implies that the caller is passing invalid data. Can we avoid the `ArgumentException` altogether by validating `result.ExternalIdentity`?

PR #5424, opened 2018-02-09
Search Box Autocomplete

ryuyu:
This PR adds in search box "autocomplete" and a new api endpoint to support this. @anangaur 
On Front Page:
image

On Packages Page:
image

Current interaction scheme:
Arrow keys up and Down while focused in search box will move up and down the autocomplete list.
Tabbing/esc will close the list.
Enter with result selected/focused will navigate to package details page.


anangaur:
Nice!! I like black color better than the blue for the package name. Is it possible to append the download count along with the package ID?

E.g.

MyPackage.Sample (\<downloadIcon>21,345)   Description.

Package ID  - black
Rest - Gray

src/NuGetGallery/Scripts/gallery/autocomplete.js
```diff
@@ -0,0 +1,185 @@
...
+        _resultsCache.results = ko.observable();
+        $(document).keydown(function (e) {
+            if (e.keyCode === 27) {
+                removeOldAutocompleteResults();
+                e.stopPropagation();
+            }
+        });
+
+        $("#autocomplete-results-container").keydown(function (e) {
+            if (e.keyCode === 40) {
```
scottbommarito: I think every keycode should have either a comment or an appropriately named constant. You put comments in the searchBox key which was good, but I don't see comments on all the keycodes.

```
if (e.KeyCode === _downArrow) {
```

or

```
if (e.KeyCode === 40) { // down arrow
```

src/NuGetGallery/Scripts/gallery/autocomplete.js
```diff
@@ -0,0 +1,185 @@
...
+    'use strict';
+    var _autocompleteTimeout = 0;
+    var _autocompleteDelay = 100; //ms
+    var _resultsCache = {};
+    var _maxResults = 9;
+    var _lastIndex = 0;
+
+    function hookAutocomplete(maxResults) {
+        _resultsCache.results = ko.observable();
+        $(document).keydown(function (e) {
```
scottbommarito: May want to add some more comments through this code in general, we're C# devs not JS devs so some of this stuff goes over our heads a little more than the usual knockout stuff does.

Especially for the stuff below with the timeouts, e.g.
```
// We will only begin loading the autocomplete results X seconds after the user has typed into the autocomplete box.
```

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -686,6 +686,26 @@ protected internal virtual Stream ReadPackageFromRequest()
             };
         }
 
+        [HttpGet]
+        [ActionName("PackageDetails")]
+        public virtual async Task<ActionResult> GetPackageDetails(
```
skofman1: >GetPackageDetails [](start = 48, length = 17)

change the name to indicate that this searches

src/NuGetGallery/Scripts/gallery/autocomplete.js
```diff
@@ -0,0 +1,191 @@
...
+        _autocompleteTimeout = 0;
+    }
+
+    function doSearch(maxResults) {
+        var currInput = $("#search").val();
+        if (currInput.length < 1) {
+            return;
+        }
+
+        var requestUrl = "/api/v2/package-ids?partialId=" + currInput + "&semVerLevel=2.0.0";
```
skofman1: please test on DEV to see if this works well on APIM

src/NuGetGallery/Scripts/gallery/autocomplete.js
```diff
@@ -0,0 +1,191 @@
...
+        });
+    }
+
+    function setupAllAuxData(idList) {
+        var requestUrl = "/api/v2/package-details?searchString=";
+        for (var i = 0; i < idList.length; i++) {
+            var tempId = idList[i];
+            var searchData = _resultsCache[safeId(tempId)];
+            if (typeof searchData() == "string") {
+                requestUrl += "packageId:" + idList[i] + " ";
```
joelverhagen: nit: `packageid`

src/NuGetGallery/Scripts/gallery/autocomplete.js
```diff
@@ -0,0 +1,191 @@
...
+                    }
+                }
+
+                setupAllAuxData(data.slice(0, maxResults));
+            }
+        });
+    }
+
+    function setupAllAuxData(idList) {
+        var requestUrl = "/api/v2/package-details?searchString=";
```
joelverhagen: `q`

joelverhagen: (instead of `searchString`

joelverhagen:
Have we done an accessibility pass on this?

Could you ask @anangaur if we should talk to UX exports about this change?

joelverhagen:
Accessibility then :shipit:

maartenba:
A new endpoint under /api/v2. Can it somehow be documented? Do third party servers need it?

joelverhagen:
@maartenba, this endpoint is an implementation detail for the auto-complete box and will not be used by the client. 

> Can it somehow be documented?

@ryuyu, @skofman1 - feel free to chime in here, but seeing as even the parts of the V2 protocol that the client uses are not documented, it doesn't make sense here.

> Do third party servers need it?

Nope. Unless you want to implement an auto-complete search box in the same way... but even that it's an implementation detail.

anangaur:
I am not in favor of documenting any V2 APIs. If required, we can have a work item tracking a corresponding V3 API but not something we can work on, in near term.

maartenba:
Thanks! Not a requirement imo but given it suddenly appears on the API/v2 path I just had to ask.

xavierdecoster:
@anangaur @joelverhagen please bare in mind that properly rolling out APIM in front of the gallery may require "some" v2 API docs going forward.

Reason: APIM efforts mandate we define these API endpoints, and their expected behavior in APIM config rules (e.g. request throttling, redirect to HTTPS, etc). 
Missing an endpoint under `/api/v2` in APIM config rules can have unintended side-effects.
On the other hand, the very act of configuring APIM rules in a way is documenting the API v2 endpoints ;)

PR #5425, opened 2018-02-09
Add tool to bulk reflow packages

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/1094.

Also, refactor the `hash` tool to use DI, app.config, and KeyVault.

### DIFF
--- a/src/GalleryTools/Commands/ReflowCommand.cs
+++ b/src/GalleryTools/Commands/ReflowCommand.cs
@@ -1,0 +1,223 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Autofac;
+using Microsoft.Extensions.CommandLineUtils;
+using NuGet.Packaging.Core;
+using NuGet.Versioning;
+using NuGetGallery;
+
+namespace GalleryTools.Commands
+{
+    public static class ReflowCommand
+    {
+        private const int DefaultBatch = 1;
+        private const int DefaultSleepMultiplier = 60;
+        private const string PackageListShortOption = "-p";
+
+        public static void Configure(CommandLineApplication config)
+        {
+            config.Description = "Bulk reflow many packages";
+            config.HelpOption("-? | -h | --help");
+
+            var packageListOption = config.Option(
+                $"{PackageListShortOption} | --path",
+                "A path to a list of packages, one ID and version per line with a space in between.",
+                CommandOptionType.SingleValue);
+
+            var batchSizeOption = config.Option(
+                "-b | --batch",
+                $"The number of packages to reflow in a batch (default: {DefaultBatch}).",
+                CommandOptionType.SingleValue);
+
+            var sleepDurationOption = config.Option(
+                "-s | --sleep",
+                $"The duration in seconds to sleep between each batch (default: batch size * {DefaultSleepMultiplier}).",
+                CommandOptionType.SingleValue);
+
+            config.OnExecute(() =>
+            {
+                return ExecuteAsync(packageListOption, batchSizeOption, sleepDurationOption).GetAwaiter().GetResult();
+            });
+        }
+
+        private static async Task<int> ExecuteAsync(
+            CommandOption packageListOption,
+            CommandOption batchSizeOption,
+            CommandOption sleepDurationOption)
+        {
+            if (!packageListOption.HasValue())
+            {
+                Console.WriteLine($"The '{PackageListShortOption}' parameter is required.");
+                return 1;
+            }
+
+            var batchSize = DefaultBatch;
+            if (batchSizeOption.HasValue())
+            {
+                batchSize = int.Parse(batchSizeOption.Value());
+
+                if (batchSize <= 0)
+                {
+                    Console.WriteLine("The batch size must be greater than zero.");
+                    return 1;
+                }
+            }
+
+            var sleepDuration = TimeSpan.FromSeconds(batchSize * DefaultSleepMultiplier);
+            if (sleepDurationOption.HasValue())
+            {
+                sleepDuration = TimeSpan.FromSeconds(int.Parse(sleepDurationOption.Value()));
+
+                if (sleepDuration < TimeSpan.Zero)
+                {
+                    Console.WriteLine("The sleep duration but be zero or more seconds.");
+                    return 1;
+                }
+            }
+
+            var packageListPath = packageListOption.Value();
+            var completedListPath = packageListPath + ".progress";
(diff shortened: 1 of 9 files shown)


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -690,9 +690,12 @@ private static void RegisterCookieComplianceService(ContainerBuilder builder, Co
...
                 .SingleInstance();
 
-            // Initialize the service on App_Start to avoid any performance degradation during initial requests.
-            var siteName = configuration.GetSiteRoot(true);
-            HostingEnvironment.QueueBackgroundWorkItem(async cancellationToken => await service.InitializeAsync(siteName, diagnostics, cancellationToken));
+            if (!(service is NullCookieComplianceService))
+            {
+                // Initialize the service on App_Start to avoid any performance degradation during initial requests.
+                var siteName = configuration.GetSiteRoot(true);
+                HostingEnvironment.QueueBackgroundWorkItem(async cancellationToken => await service.InitializeAsync(siteName, diagnostics, cancellationToken));
```
joelverhagen: `HostingEnvironment` doesn't work in a console app context. So turn this off if the cookie compliance service is the null implementation.

src/GalleryTools/Commands/ReflowCommand.cs
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        private static async Task<int> ExecuteAsync(
+            CommandOption packageListOption,
+            CommandOption batchSizeOption,
+            CommandOption sleepDurationOption)
+        {
+            if (!packageListOption.HasValue())
+            {
+                Console.WriteLine("The '-f' parameter is required.");
```
chenriksson: Should '-f' be '-p'?

src/GalleryTools/Commands/ReflowCommand.cs
```diff
@@ -0,0 +1,221 @@
...
+                return 1;
+            }
+
+            var batchSize = DefaultBatch;
+            if (batchSizeOption.HasValue())
+            {
+                batchSize = int.Parse(batchSizeOption.Value());
+
+                if (batchSize <= 0)
```
chenriksson: Should there be an upper limit on batchSize? Could there be potential performance issues if batch size is too large, or sleep to small?

src/GalleryTools/Commands/ReflowCommand.cs
joelverhagen: Yup, nice catch.

src/GalleryTools/Commands/ReflowCommand.cs
joelverhagen: I'd rather leave that to the admin of the tool.

src/GalleryTools/Commands/ReflowCommand.cs
```diff
@@ -0,0 +1,222 @@
...
+                {
+                    Console.WriteLine("The batch size must be greater than zero.");
```
scottbommarito: no `return 1` here?

If batch size is 0 I would imagine it'd get stuck and never finish.

joelverhagen: Yup, nice catch.

src/GalleryTools/Commands/ReflowCommand.cs
```diff
@@ -0,0 +1,222 @@
...
+                    return 1;
+                }
+            }
+
+            var packageListPath = packageListOption.Value();
+            var completedListPath = packageListPath + ".progress";
+            var remainingList = GetRemainingList(packageListPath, completedListPath);
+            if (!remainingList.Any())
+            {
+                return 1;
```
scottbommarito: Might want to add something along the lines of
```
Console.WriteLine("There were no packages to process.")
```
so that it's clear that the tool didn't do anything.

joelverhagen: This is logged in `GetRemainingList`.

src/GalleryTools/Commands/ReflowCommand.cs
```diff
@@ -0,0 +1,222 @@
...
+
+                    if (string.IsNullOrWhiteSpace(line))
+                    {
+                        continue;
+                    }
+
+                    var pieces = line.Split(new[] { ' ' }, 2);
+                    if (pieces.Length != 2)
+                    {
+                        Console.WriteLine($"Line {lineNumber}: Ignoring line without a space separator.");
```
scottbommarito: nit: technically this could be without a space separator, or with too many space separators

src/GalleryTools/Commands/ReflowCommand.cs
```diff
@@ -0,0 +1,222 @@
...
+                    if (string.IsNullOrWhiteSpace(unparsedVersion) || !NuGetVersion.TryParse(unparsedVersion, out version))
+                    {
+                        Console.WriteLine($"Line {lineNumber}: Ignoring invalid version.");
+                        continue;
+                    }
+
+                    var packageIdentity = new PackageIdentity(id, version);
+                    if (!uniquePackageIdentities.Add(packageIdentity))
+                    {
+                        Console.WriteLine($"Line {lineNumber}: ignoring duplicate package.");
```
scottbommarito: nit: the other messages have `Ignoring` and not `ignoring` (casing)

src/GalleryTools/Commands/ReflowCommand.cs
joelverhagen: I don't understand. The second parameter of `Split` means an array of length 1 or 2 will be returned.

src/GalleryTools/Commands/ReflowCommand.cs
joelverhagen: Yeah good point.

joelverhagen: Fixed.

src/GalleryTools/Commands/ReflowCommand.cs
joelverhagen: Fixed.

src/GalleryTools/Commands/ReflowCommand.cs
joelverhagen: Fixed.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
scottbommarito: Can we just detect that this is a console app context otherwise? If the shims are set up correctly in the console app this will have the same issue.

joelverhagen: Out of scope for this PR. The null implementation (even in gallery) does not need to run in the background.

src/GalleryTools/Commands/ReflowCommand.cs
scottbommarito: missed that second parameter, didn't know that was a thing haha

Issue #5263, opened 2018-01-10
Organizations: Notifications


chenriksson:
Integrate notifications in with new organization UI, including:
- Email confirmation link for organization migration request to new admin
- Email confirmation link for membership request to new member
- Email notification of new members to admins

PR #5428, opened 2018-02-10
[Organizations] Notifications for transforming an account into an organization

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5263

This PR adds email notifications for
1 - Asking a user to become the admin of your transformed account
2 - Rejecting a request to become the admin of a transformed account
3 - Cancelling a request for a user to become the admin of your transformed account
4 - Asking a user to join your organization
5 - Accepting a request to join an organization
6 - Rejecting a request to join an organization
7 - Cancelling a request for a user to join your organization
8 - Removing a member from your organization
9 - Updating a member of your organization (changing admin to collaborator and vice-versa)

As a result, like packages, users are no longer automatically added to organizations when an admin adds them--they must accept the request. The UI has been updated to support this, and shows "(pending)" for users that are pending.

NOTE: more than half of this PR is unit tests

### DIFF
--- a/src/NuGetGallery/Controllers/OrganizationsController.cs
+++ b/src/NuGetGallery/Controllers/OrganizationsController.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Linq;
 using System.Net;
 using System.Net.Mail;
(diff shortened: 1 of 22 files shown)


skofman1:
@scottbommarito , please add screenshots

chenriksson:
@scottbommarito @anangaur Should we display organization membership requests on the Manage Organizations view, like we do with package ownership requests on Manage Packages?

src/NuGetGallery/Services/UserService.cs
```diff
@@ -74,21 +84,112 @@ public class UserService : IUserService
...
+                }
+
+                membership = new Membership()
+                {
+                    Member = member,
+                    IsAdmin = request.IsAdmin
+                };
+                organization.Members.Add(membership);
+            }
+            else
```
chenriksson: The `else` case is an update, not add? Shouldn't this throw, since there's a separate `UpdateMemberAsync` API?

We don't add member requests for changing role, do we?

src/NuGetGallery/Services/UserService.cs
```diff
@@ -62,6 +62,16 @@ public class UserService : IUserService
                     Strings.AddMember_AlreadyAMember, memberName));
             }
 
+            var request = FindMembershipRequestByUsername(organization, memberName);
+            if (request != null)
+            {
+                // If there is already an existing request, return it.
+                // If the existing request grants collaborator but we are trying to create a request that grants admin, update the request to grant admin.
```
chenriksson: Does the front-end allow re-adding a pending member?

Would prefer to keep it simple - block the re-add scenario, and just throw here. If we do allow replacing/editing requests, I think we should regenerate the token... since the request context may have changed.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -58,8 +59,96 @@ public virtual ActionResult ManageOrganization(string accountName)
...
+                var member = await UserService.AddMemberAsync(account, GetCurrentUser().Username, confirmationToken);
+                MessageService.SendOrganizationMemberUpdatedNotice(account, member);
+
+                return HandleOrganizationMembershipRequestView(new HandleOrganizationMembershipRequestModel(true, account));
+            }
+            catch (EntityException e)
+            {
+                var failureReason = e.AsUserSafeException().GetUserSafeMessage();
+                return HandleOrganizationMembershipRequestView(new HandleOrganizationMembershipRequestModel(true, account, failureReason));
+            }
```
chenriksson: When we confirm transformation, we redirect the admin to the Manage Organization view.

Do we want to do the same for adding members... since we added read-only mode for collaborators? The success message could be shown in the information banner, like we do on transformation.

I would check w/ @anangaur on this.

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -681,6 +681,266 @@ public void SendPackageDeletedNotice(Package package, string packageUrl, string
             }
         }
 
+        public void SendOrganizationTransformRequest(User accountToTransform, User adminUser, string profileUrl, string confirmationUrl, string rejectionUrl)
```
chenriksson: Would have @anangaur review all these emails.

src/NuGetGallery/Scripts/gallery/page-manage-organization.js
```diff
@@ -123,6 +124,7 @@
 
             this.AccountName = initialData.AccountName;
             this.AddMemberUrl = initialData.AddMemberUrl;
+            this.CancelMemberRequestUrl = initialData.CancelMemberRequestUrl;
```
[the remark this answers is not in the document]
scottbommarito: It does not. The current implementation only sends an email. I like this idea a lot however, and it'd be very easy to implement.

src/NuGetGallery/Services/UserService.cs
scottbommarito: It updates the member request, not the member. `UpdateMemberAsync` is for updating an existing member, not a member request.

src/NuGetGallery/Services/UserService.cs
scottbommarito: I would rather handle this case gracefully rather than throw because the logic is fairly simple.

After thinking about it more I think adding the same user twice will mess up the UI, but I think the logic to fix it should be simple.

src/NuGetGallery/Services/UserService.cs
scottbommarito: Woops, assumed this was a different method.

To get to this point requires there was a valid membership request. I'm not sure how you could have a member and a membership request giving that member greater access privileges without someone manually changing the DB, but I thought that if there's an existing membership request we should ahead and upgrade the member to whatever that request grants while cleaning up the invalid request.

src/NuGetGallery/Controllers/OrganizationsController.cs
scottbommarito: I modeled the page after the corresponding package ownership request page. I'm fine with changing it to `TempData["Message"]` however. Note that we would still need this page for rejecting a request however.

scottbommarito:
Fixed a couple UI bugs regarding pending users and adding them.

src/NuGetGallery/Scripts/gallery/page-manage-organization.js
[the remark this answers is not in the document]
scottbommarito: I have filed an issue to track this: https://github.com/NuGet/NuGetGallery/issues/5550

scottbommarito:
Closing because this was split into two separate PRs:
https://github.com/NuGet/NuGetGallery/pull/5552
https://github.com/NuGet/NuGetGallery/pull/5556

PR #5429, opened 2018-02-10
USSC gallery api/status should use secondary storage.

cristinamanum:
Configure the Azure services to use secondary storage if the primary is not available. For a set of resources this setting was already in place but not for all. One sample is audit storage (that is used for api/stats) .

### DIFF
--- a/src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
@@ -35,7 +35,12 @@ namespace NuGetGallery
 
         public Task<bool> ExistsAsync()
         {
-            return _blobContainer.ExistsAsync();
+            var o = new BlobRequestOptions()
+            {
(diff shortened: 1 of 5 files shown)


src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
```diff
@@ -35,7 +35,12 @@ public ISimpleCloudBlob GetBlobReference(string blobAddressUri)
 
         public Task<bool> ExistsAsync()
         {
-            return _blobContainer.ExistsAsync();
+            var o = new BlobRequestOptions()
+            {
+                LocationMode = Microsoft.WindowsAzure.Storage.RetryPolicies.LocationMode.SecondaryOnly
```
joelverhagen: Primary then secondary?

Could we have some sort of unit/integration test for this?

cristinamanum: Reverted. It was not meant to be persisted. The integration tests are testing this path , however falling back to secondary testing cannot be done unless in a specific environment.

PR #5440, opened 2018-02-13
Add IsAuthenticated property to RequestsTelemetry

skofman1:
Validating a hypothesis around customer browsing behavior.

### DIFF
--- a/src/NuGetGallery/Telemetry/ClientInformationTelemetryEnricher.cs
+++ b/src/NuGetGallery/Telemetry/ClientInformationTelemetryEnricher.cs
@@ -30,6 +30,9 @@ namespace NuGetGallery
                         httpContext.Request.Headers[Constants.NuGetProtocolHeaderName]);
 
                     telemetry.Context.Properties.Add(TelemetryService.ClientInformation, httpContext.GetClientInformation());
+
+                    // Is the user authenticated or this is an anonymous request?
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Telemetry/ClientInformationTelemetryEnricher.cs
```diff
@@ -30,6 +30,9 @@ public void Initialize(ITelemetry telemetry)
                         httpContext.Request.Headers[Constants.NuGetProtocolHeaderName]);
 
                     telemetry.Context.Properties.Add(TelemetryService.ClientInformation, httpContext.GetClientInformation());
+
+                    // Is the user authenticated or this is an anonymous request?
+                    telemetry.Context.Properties.Add(TelemetryService.IsAuthenticated, httpContext.Request.IsAuthenticated.ToString());
```
joelverhagen: Spoke offline.

Worth checking:

1. What does API key do?
1. What about invalid API key? 
1. Have we verified both on devbox and a real environment?
1. Does this work the same for password auth vs. MSA/AAD?

skofman1: Works for API Keys, NuGet.org login, AAD login. Invalid API Key is not considered authenticated. Will test on real environment when it gets to DEV, and fix as needed. It's cheaper then deploying from feature branch (especially for a minor feature like this).

joelverhagen: 👍

Issue #5461, opened 2018-02-13
[BugBash] Manage Packages page slow for account with 60K packages


chenriksson:
Scenario 15, per @skofman1 

I got an account with 60,000 packages. "Manage Packages" page doesn't load.

chenriksson:
Call to User.IsInRole is 11.9% incl, which is excessive for checking if the current user is an admin. I believe this is when performing permissions checks.

![image](https://user-images.githubusercontent.com/15364949/36167913-50315960-10ac-11e8-9e20-3eb676b8ea1c.png)

chenriksson:
Microsoft (PROD) has almost as many package versions as NuGetStagingTest (DEV) has... so I think we should keep this in sprint. Roughly 42K versus 59K versions.

chenriksson:
Plans for optimizations:
1. Improve server query by ordering in DB and only selecting latest versions. For packages that don't have IsLatest set (concurrency issue), we will do minimal work which could be:
    - exclude entirely
    - display registration details only (not latest version)
2. Add paging if serialization and client side performance is still bad. If we do this, we should add a search option for filtering by packageId.

//cc @joelverhagen  @ryuyu

chenriksson:
We should still consider paging for scenarios with many registrations (i.e., test accounts), but that's already tracked by #5059... marking this code complete.

skofman1:
@chenriksson , will your fix help with profile page? For example: https://www.nuget.org/profiles/awscpp
The page loads very slowly despite the number of packages being low and having paging.

chenriksson:
@skofman1 No -- `PackagesService.FindPackagesByAnyMatchingOwner` was only used by "Manage Packages" and the Delete Account workflows. We could do the same w/ profiles if we're ok with the caveat there... being that we may not pull the latest if there's no IsLatest set due to the concurrency issue.

skofman1:
Created a separate issue to track: https://github.com/NuGet/NuGetGallery/issues/5593
If this is a simple fix, we should definitely go for it.

chenriksson:
@anangaur Did 'Manage Packages' performance improve for you after this latest deployment? If so, the priority for adding paging should be lower.

Microsoft has 3923 registrations and 43217 versions, so it should have helped a fair amount.

PR #5470, opened 2018-02-13
Cache IsAdmin check

chenriksson:
Issue #5461 - addresses issue 1 (11.9% incl from isAdmin checks, which increased with permissions helpers)

Server response went from 39.4s to 37.8s.

### DIFF
--- a/src/NuGetGallery.Core/Entities/User.cs
+++ b/src/NuGetGallery.Core/Entities/User.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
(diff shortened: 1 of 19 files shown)


src/NuGetGallery.Core/Entities/User.cs
```diff
@@ -17,6 +18,25 @@ namespace NuGetGallery
     /// <see href="https://weblogs.asp.net/manavi/inheritance-mapping-strategies-with-entity-framework-code-first-ctp5-part-2-table-per-type-tpt" />
     public class User : IEntity, IEquatable<User>
     {
+        #region per-request query cache
+
+        private Nullable<bool> _isAdmin;
```
shishirx34: why not simply use `bool?` type?

PR #5472, opened 2018-02-13
Cache admins and collaborators for permissions

chenriksson:
Issue #5461, optimization 2.

Caches and separates admin/collaborators lookups for current user during permissions checks so duplicate checks don't hit the DB.

Originally saw response go from 37.8s -> 35.1s, but now seeing it go from 59.1s -> 42.5s

### DIFF
--- a/src/NuGetGallery/Services/UserService.cs
+++ b/src/NuGetGallery/Services/UserService.cs
@@ -134,7 +134,7 @@ namespace NuGetGallery
             }
 
             // block removal of last admin
-            if (membership.IsAdmin && organization.Members.Count(m => m.IsAdmin) == 1)
+            if (membership.IsAdmin && organization.Administrators.Count() == 1)
             {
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Services/UserService.cs
```diff
@@ -109,7 +109,7 @@ public class UserService : IUserService
             if (membership.IsAdmin != isAdmin)
             {
                 // block removal of last admin
-                if (membership.IsAdmin && organization.Members.Count(m => m.IsAdmin) == 1)
+                if (membership.IsAdmin && organization.Administrators.Count() == 1)
```
skofman1: >rganization.Administrators.Count() == 1 [](start = 43, length = 39)

nit: use Any()

chenriksson: The check is to see if the last admin is being removed, so I need to check for exactly 1 not >= 1

Issue #5441, opened 2018-02-13
[BugBash] Email missing from message on account link


chenriksson:
see scenario 1, per @joelverhagen 

The email on the link page is empty

shishirx34:
Needs investigation.

scottbommarito:
Not high pri because it doesn't not block the flow because the email address is still persisted in the form.

scottbommarito:
Note: this only appears after submitting the form with incorrect data--the first load is successful.

Issue #5436, opened 2018-02-12
AAD+Orgs: NullReferenceException when whitelist blob is missing


chenriksson:
Exception seen on DEV, when /content/Login-Discontinuation-Configuration.json was missing.

Stack:
"assembly":"NuGetGallery, Version=4.4.4.0, Culture=neutral, PublicKeyToken=null","method":"NuGetGallery.Authentication.AuthenticationService+<GetDiscontinuedLoginClaims>d__28.MoveNext","level":0,"line":254,"fileName":"C:\\vsts-agent\\_work\\12\\s\\src\\Gallery\\submodules\\Gallery\\src\\NuGetGallery\\Authentication\\AuthenticationService.cs"
"assembly":"mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089","method":"System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw","level":1,"line":0
"assembly":"mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089","method":"System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification","level":2,"line":0
"assembly":"NuGetGallery, Version=4.4.4.0, Culture=neutral, PublicKeyToken=null","method":"NuGetGallery.Authentication.AuthenticationService+<CreateSessionAsync>d__27.MoveNext","level":3,"line":239,"fileName":"C:\\vsts-agent\\_work\\12\\s\\src\\Gallery\\submodules\\Gallery\\src\\NuGetGallery\\Authentication\\AuthenticationService.cs"
"assembly":"mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089","method":"System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw","level":4,"line":0
"assembly":"mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089","method":"System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification","level":5,"line":0
[...]

PR #5482, opened 2018-02-15
Small orgs/AAD bug bash fixes

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5441
https://github.com/NuGet/NuGetGallery/issues/5436

And also fixed a NPE for when the user has an old AAD credential without a tenant ID.


src/NuGetGallery/Security/RequireOrganizationTenantPolicy.cs
```diff
@@ -75,7 +75,7 @@ public override SecurityPolicyResult Evaluate(UserSecurityPolicyEvaluationContex
             var targetCredential = targetAccount.Credentials.GetAzureActiveDirectoryCredential();
 
             if (targetCredential == null
-                || !targetCredential.TenantId.Equals(state.Tenant, StringComparison.OrdinalIgnoreCase))
+                || !state.Tenant.Equals(targetCredential.TenantId, StringComparison.OrdinalIgnoreCase))
```
scottbommarito: Sometimes, for old AAD, `targetCredential.TenantId` is null. `state.Tenant` is guaranteed to be non-null, so we should use that to perform this comparison.

src/NuGetGallery/Views/Authentication/LinkExternal.cshtml
```diff
@@ -66,7 +66,7 @@
             else
             {
                 <p class="text-center">
-                    Looks like we don't have an account with this email address (@Model.SignIn.UserNameOrEmail) in our records.
+                    Looks like we don't have an account with this email address (@Model.Register.EmailAddress) in our records.
```
shishirx34: were you able to find out why the `SignIn.UserNameOrEmail` was empty even though the controller set its value in the model?

src/NuGetGallery/Security/RequireOrganizationTenantPolicy.cs
shishirx34: Ideally we should populate the tenant ID values, but it would only be the case for admins that is missing.

src/NuGetGallery/Views/Authentication/LinkExternal.cshtml
scottbommarito: When you submit the form and the submission fails, `SignIn.UserNameOrEmail` is not set.

Issue #5479, opened 2018-02-14
Remove INuGetExeDownloaderService after /nuget.exe redirect is verified in PROD


joelverhagen:
This code is no longer necessary.

Need to remove associated config too.

PR #5483, opened 2018-02-15
Remove NugetExeDownloaderService and related configuration

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/5479


joelverhagen:
> This looks good but I was pretty sure there was a functional test for the /nuget.exe endpoint, has that been checked yet to make sure it still works?

This PR has not functional impact. The change to the endpoint is already in `dev` branch and is passing functional tests. Also, I could not find that functional test, I just found some test infrastructure that downloads from PROD's `/nuget.exe` endpoint as test setup.

PR #5485, opened 2018-02-15
[BugBash] MSA - Add exception handler for access denied

### LINKED ISSUE #5451: [BugBash] 500 errors when cancel on initial auth flow
Scenario 9, per @joelverhagen 

Hitting cancel during initial auth...

Can you also investigate possible 500 here...

Scenario 2, per @skofman1 (Issue 2)

After I finished the linking, and got back to home page, clicked "back" a few times and got a 500...

shishirx34:
The OpenIdConnect library throws exception upon failed authentication which resulted in 500 when access was denied by user on MSA/AAD auth. This fixes #5451 and shows ExternalLinkFailed error.

image


src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -33,6 +35,9 @@ public static class AuthenticationType
         public static readonly string V2CommonTenant = "common";
         public static readonly string Authority = "https://login.microsoftonline.com/{0}/v2.0";
 
+        private string redirectUrl;
```
chenriksson: authenticator instance is per request?

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -136,5 +146,19 @@ public override IdentityInformation GetIdentityInformation(ClaimsIdentity claims
...
+
+        // The OpenIdConnect.<AuthenticateCoreAsync> throws OpenIdConnectProtocolException upon denial of access permissions by the user, 
+        // this could result in an internal server error, catch this exception and continue to the controller where appropriate
+        // error handling is done.
+        private Task AuthenticationFailed(AuthenticationFailedNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions> notification)
+        {
+            if (notification.Exception.Message == ACCESS_DENIED)
+            {
+                notification.HandleResponse();
+                notification.Response.Redirect(redirectUrl);
```
chenriksson: For confirm transform by the org admin, it makes sense to redirect so the user sees the transform failed error page if the AAD linking failed or was cancelled.

For UIAuthorize actions that don't support deprecated password auth, will the user still see the info message if the UIAuthorize redirects them to the homepage for the link/transform prompt?

shishirx34: This will only be shown once when the permission was denied to access permissions, consumed from cookie and is removed so won't repeat upon further UIAuthorize redirects.

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
shishirx34: Ah! good catch, it is a SingleInstance. this won't work.

chenriksson: Can we ignore the redirectUrl in the error case, and just redirect to home with the info message?

shishirx34: I found another way. This is cleaner.

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -136,5 +144,31 @@ public override IdentityInformation GetIdentityInformation(ClaimsIdentity claims
...
+                    .ProtocolMessage
+                    .State
+                    .Split('=');
+                var authenticationProperties = notification
+                    .Options
+                    .StateDataFormat
+                    .Unprotect(authenticationPropertiesEncodedString[1]);
+
+                notification.HandleResponse();
+                notification.Response.Redirect(authenticationProperties.RedirectUri);
```
chenriksson: Is this guaranteed to be set, or do we need to default to redirecting to homepage?

shishirx34: Its guaranteed to be set, see https://github.com/NuGet/NuGetGallery/blob/0c6458164ec897c810baa6cd5f5e388c25a0ebd3/src/NuGetGallery/Infrastructure/ChallengeResult.cs#L26

Issue #5463, opened 2018-02-13
[BugBash] Update message on attempt to remove last org member


chenriksson:
Scenario 21, per @anangaur 

The text needs update...

PR #5490, opened 2018-02-15
Error message for removal of last org admin

chenriksson:
Issue #5463, messaging per @anangaur 

Scenario is current user (org admin) as only admin deleting their membership or changing self to collaborator... which we block.

### DIFF
--- a/src/NuGetGallery/Strings.resx
+++ b/src/NuGetGallery/Strings.resx
@@ -786,8 +786,8 @@ If you wish to update the linked Microsoft account you can do so from the accoun
   <data name="DeleteMember_CannotDeleteLastAdmin" xml:space="preserve">
     <value>Organization member '{0}' is the last admin and cannot be deleted.</value>
   </data>
-  <data name="UpdateOrDeleteMember_CannotRemoveLastAdmin" xml:space="preserve">
-    <value>Organization member '{0}' is the last admin and cannot be changed.</value>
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Strings.resx
```diff
@@ -786,8 +786,8 @@ If you wish to update the linked Microsoft account you can do so from the accoun
   <data name="DeleteMember_CannotDeleteLastAdmin" xml:space="preserve">
     <value>Organization member '{0}' is the last admin and cannot be deleted.</value>
   </data>
-  <data name="UpdateOrDeleteMember_CannotRemoveLastAdmin" xml:space="preserve">
-    <value>Organization member '{0}' is the last admin and cannot be changed.</value>
+  <data name="UpdateMember_CannotRemoveLastAdmin" xml:space="preserve">
+    <value>You can't change your role to collaborator. In order to change, another collaborator must be assigned as an administrator.</value>
```
scottbommarito: I think the phrasing of `another collaborator must be assigned as an administrator` is a bit awkward.

Perhaps `there must be at least one other admin` or something similar would be simpler?

chenriksson: This text came from @anangaur

chenriksson:
@scottbommarito Tests existed, but only verified exception type... added verification of the message.

PR #5491, opened 2018-02-16
[Organizations] Include reserved namespaces and package ownership requests in Manage Packages page filtering

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5310

Previously, only the listed packages and unlisted packages sections were included in the filtering. This change adds the reserved namespaces and package ownership requests sections to the filtering.

A couple notable changes are:
- You can now cancel any package ownership request made for another user to be an owner of a package you own from this page. Previously you had to do it from the Manage Package Owners page.
- The package owner requests table shows all of the package's owners instead of the user who sent the request like it used to. I could show both, but I came to the conclusion that the requestor is pretty irrelevant information.
- If a user tries to accept or reject a package ownership request for a package they are already an owner of, the request is deleted because it's no longer valid.

(the screenshots have the packages sections removed--no actual changes were made to these sections)

All packages:

image

Filtered by self:

image

Filtered by organization that self is collaborator of:

(doesn't show any requests because the current user can't do anything about them)

image

Filtered by organization that self is admin of:

image


chenriksson:
@scottbommarito Per screenshots, looks like we show logo with org owners in the Owners column, but not the New Owner column. Is this easy to fix?

scottbommarito:
@chenriksson nice catch! should be easy to do

scottbommarito:
@chenriksson done!

chenriksson:
@scottbommarito I'm not sure we need a separate cancel owner request icon... UI is simpler without.

Instead, could 'X' mean reject the request if the user is the receiver, but mean cancel the request if the user is the sender (or another admin)? Both essentially close the request, which is similar enough to the end user... IMHO.

scottbommarito:
@chenriksson smart catch--we only need to show either the reject or the cancel icon, because the end result is the same for the user lol

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -368,11 +368,21 @@ public virtual ActionResult Packages()
...
-            var received = _packageOwnerRequestService.GetPackageOwnershipRequests(newOwner: currentUser);
-            var sent = _packageOwnerRequestService.GetPackageOwnershipRequests(requestingOwner: currentUser);
+            var userReceived = _packageOwnerRequestService.GetPackageOwnershipRequests(newOwner: currentUser);
+            var orgReceived = currentUser.Organizations
+                .Where(m => ActionsRequiringPermissions.HandlePackageOwnershipRequest.CheckPermissions(currentUser, m.Organization) == PermissionsCheckResult.Allowed)
+                .SelectMany(m => _packageOwnerRequestService.GetPackageOwnershipRequests(newOwner: m.Organization));
+            var received = userReceived.Union(orgReceived);
+            
+            var sent = _packageService.FindPackagesByAnyMatchingOwner(currentUser, true, false)
+                .SelectMany(p => _packageOwnerRequestService.GetPackageOwnershipRequests(package: p.PackageRegistration));
```
chenriksson: For perf issue #5461, I'm trying to rework the query in FindPackagesByAnyMatchingOwner, which is currently really inefficient. This API fetches the latest version of each registration. Due to our concurrency issue with the IsLatest columns, it ends up fetching all package rows for all matching owners and sorting on NuGetVersion in the client (Gallery).

If possible, can you avoid adding this second call to `FindPackagesByAnyMatchingOwner`? It's going to make #5461 twice as bad.

scottbommarito:
@chenriksson done as well!

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -3,16 +3,16 @@
...
     }
 
     function formatPackagesData(packagesCount, downloadsCount) {
         return packagesCount.toLocaleString()
-            + ' package' + (packagesCount == 1 ? '' : 's')
+            + ' package' + (packagesCount === 1 ? '' : 's')
             + ' / '
             + downloadsCount.toLocaleString()
-            + ' download' + (downloadsCount == 1 ? '' : 's');
+            + ' download' + (downloadsCount === 1 ? '' : 's');
```
chenriksson: awesome - thanks!

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -90,6 +90,141 @@
             }, this);
         }
 
+        function showInitialReservedNamespaceData(dataSelector, namespacesList) {
+            $(dataSelector).text(formatReservedNamespacesData(namespacesList.length));
+        }
+
+        function formatReservedNamespacesData(namespacesCount) {
+            return namespacesCount.toLocaleString() + " namespace" + (namespacesCount === 1 ? '' : 's');
+        }
```
chenriksson: I wonder if we can add a pluralize helper in common.js?

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -105,11 +240,17 @@
...
+            this.RequestsReceived = new OwnerRequestsListViewModel(this, initialData.RequestsReceived, true, false);
+            this.RequestsSent = new OwnerRequestsListViewModel(this, initialData.RequestsSent, false, true);
         }
 
         // Immediately load initial expander data
         showInitialPackagesData("#listed-data", initialData.ListedPackages);
         showInitialPackagesData("#unlisted-data", initialData.UnlistedPackages);
+        showInitialReservedNamespaceData("#namespaces-data", initialData.ReservedNamespaces);
+        showInitialOwnerRequestsData("#requests-received-data", initialData.RequestsReceived);
+        showInitialOwnerRequestsData("#requests-sent-data", initialData.RequestsSent);
```
chenriksson: Looks great - I know this was a lot of work!

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -48,32 +48,45 @@
...
+                        <span id="requests-sent-data" data-bind="text: RequestsSent.VisibleRequestsHeading()"></span>
+                    </text>,
+                    @<text>
+                        <div class="row user-package-list">
+                            <div data-bind="template: { name: 'manage-owner-requests', data: RequestsSent }"></div>
+                        </div>
                     </text>, expanded: false)
             }
         </div>
+    </div>
```
chenriksson: We could still consider moving sections into partials, like I did for Account/ManageOrganizations views. It might be nice to reduce the size of this file.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
scottbommarito: I was tired of those LINT errors haha

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: good catch! I can reuse the results from when it's called earlier.

src/NuGetGallery/Views/Users/Packages.cshtml
scottbommarito: Not sure how to best split this up...I could split the templates into their own partials but there's still gonna be a massive script section at the end.

chenriksson: Feel free to leave as-is, and I can re-evaluate when I do my work for #5461

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -143,35 +152,141 @@
...
+                        <th>Package ID</th>
+                        <th>Owners</th>
+                        <th>New Owner</th>
+                        <th><span class="hidden">Actions</span></th>
+                    </tr>
+                </thead>
+                <tbody data-bind="foreach: Requests">
+                    <tr class="manage-package-listing" data-bind="visible: Visible">
+                        <td class="align-middle hidden-xs">
+                            <img class="package-icon img-responsive" aria-hidden="true" alt=""
```
chenriksson: accessibility requires alt for images, so I assume we should use non-empty here.

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -143,35 +152,141 @@
     </div>
 </script>
 
+<script type="text/html" id="manage-namespaces">
+    <div class="col-md-12">
+        <div class="panel-collapse collapse in" aria-expanded="true">
+            <table class="table">
+                <thead>
+                    <tr class="manage-package-headings">
+                        <th class="hidden-xs"></th>
```
chenriksson: see: [Accessibility PR #5426](https://github.com/NuGet/NuGetGallery/pull/5426/files#diff-e611d875aea9efbcb6f2db8db63b873eR85)

Keros complained about empty header columns, so I put hidden spans with a text description. I wonder if we should do that here too.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
scottbommarito: I'm a little skeptical about adding something like this given that
1 - the code here is (less than) a one-liner
2 - the actual problem of pluralization is very complex (if someone supplies "octopus" they might expect it to be pluralized to "octopi")

I'm fine with adding a function just for this file because it is used 3 times but beyond that scope might be misleading.

src/NuGetGallery/Views/Users/Packages.cshtml
scottbommarito: `alt` was empty for the package icon in the packages lists. Not sure what to put there but I'll figure it out.

src/NuGetGallery/Views/Users/Packages.cshtml
scottbommarito: Fixed!

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1301,6 +1301,8 @@ public virtual Task<ActionResult> RejectPendingOwnershipRequest(string id, strin
 
             if (package.Owners.Any(o => o.MatchesUser(user)))
             {
+                // If the user is already an owner, clean up the invalid request.
+                await _packageOwnershipManagementService.DeletePackageOwnershipRequestAsync(package, user);
```
cristinamanum: Probably this should not throw.

Issue #5454, opened 2018-02-13
[BugBash] Transform account validation errors don't persist long


chenriksson:
Scenario 10, per @skofman1 

transform account page: when entering a admin that doesn't have AAD...

PR #5502, opened 2018-02-16
[BugBash] Prevent added model errors from being reset by JQuery Validate

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5454

Basically, by putting the errors on `string.Empty` instead of `AdminUsername`, they are not erased by JQuery Validate when it checks the input.

Also fixed this in the `ForgotPassword` form because it was easy to do so.

### DIFF
--- a/src/NuGetGallery/Views/Users/Transform.cshtml
+++ b/src/NuGetGallery/Views/Users/Transform.cshtml
@@ -45,6 +45,9 @@
                         @Html.ShowTextBoxFor(m => m.AdminUsername)
                         @Html.ShowValidationMessagesFor(m => m.AdminUsername)
                     </div>
+                    <div class="form-group has-error">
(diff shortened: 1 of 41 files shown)


src/NuGetGallery/Views/Users/Transform.cshtml
```diff
@@ -45,6 +45,9 @@
                         @Html.ShowTextBoxFor(m => m.AdminUsername)
                         @Html.ShowValidationMessagesFor(m => m.AdminUsername)
```
chenriksson: Since we're using `string.Empty`, should we remove `ShowValidationMessagesFor(m => m.AdminUsername)`, `Email`, etc?

scottbommarito: No, because those will show errors with `AdminUsername` not specified by us (e..g. if the field is empty).

Issue #5522, opened 2018-02-21
Unable to reflow packages with invalid dependency version ranges


joelverhagen:
Today, we block invalid dependency version ranges on upload. For example:

```xml
<dependency id="WebActivator" version="" />
<dependency id="jQuery" version="$version$"/>
```

The client APIs and feed2catalog all treat such version ranges as open ranges, `(, )`.

Today we are unable to reflow these packages and must resort to an unlist/relist hack. Exception:

```
NuGet.Packaging.Core.PackagingException: Invalid package version for a dependency with id 'SimpleInjector' in package 'SimpleInjector.AssemblyScanner.1.0.8.37': ''
   at NuGet.Packaging.NuspecReader.GetPackageDependencies(IEnumerable`1 nodes, Boolean useStrictVersionCheck)
   at NuGet.Packaging.NuspecReader.<GetDependencyGroups>d__28.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at NuGetGallery.Packaging.PackageMetadata..ctor(Dictionary`2 metadata, IEnumerable`1 dependencyGroups, IEnumerable`1 frameworkGroups, IEnumerable`1 packageTypes, NuGetVersion minClientVersion) in D:\src\git\github.com\NuGet\NuGetGallery\src\NuGetGallery.Core\Packaging\PackageMetadata.cs:line 27
   at NuGetGallery.Packaging.PackageMetadata.FromNuspecReader(NuspecReader nuspecReader) in D:\src\git\github.com\NuGet\NuGetGallery\src\NuGetGallery.Core\Packaging\PackageMetadata.cs:line 156
   at NuGetGallery.ReflowPackageService.<ReflowAsync>d__4.MoveNext() in D:\src\git\github.com\NuGet\NuGetGallery\src\NuGetGallery\Services\ReflowPackageService.cs:line 53
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
[...]

PR #5524, opened 2018-02-21
Enable reflowing of packages that don't pass the valid version range rule

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/5522

### DIFF
--- a/src/NuGetGallery.Core/Packaging/PackageMetadata.cs
+++ b/src/NuGetGallery.Core/Packaging/PackageMetadata.cs
@@ -147,15 +147,18 @@ namespace NuGetGallery.Packaging
         /// Gets package metadata from a the provided <see cref="NuspecReader"/> instance.
         /// </summary>
         /// <param name="nuspecReader">The <see cref="NuspecReader"/> instance used to read the <see cref="PackageMetadata"/></param>
+        /// <param name="strict">
+        /// Whether or not to be strict when reading the <see cref="NuspecReader"/>. This should be <code>true</code>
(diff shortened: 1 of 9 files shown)


joelverhagen:
🔔

src/NuGetGallery.Core/Packaging/PackageMetadata.cs
```diff
@@ -147,15 +147,18 @@ private Uri GetValue(string key, Uri alternateValue)
...
         /// <param name="nuspecReader">The <see cref="NuspecReader"/> instance used to read the <see cref="PackageMetadata"/></param>
+        /// <param name="strict">
+        /// Whether or not to be strict when reading the <see cref="NuspecReader"/>. This should be <code>true</code>
+        /// on initial ingestion but false when a package that has already been accepted is being processed.</param>
         /// <exception cref="PackagingException">
         /// We default to use a strict version-check on dependency groups. 
         /// When an invalid dependency version range is detected, a <see cref="PackagingException"/> will be thrown.
         /// </exception>
-        public static PackageMetadata FromNuspecReader(NuspecReader nuspecReader)
+        public static PackageMetadata FromNuspecReader(NuspecReader nuspecReader, bool strict)
```
shishirx34: nit: may be rename this parameter to `dependencyStrictVersionCheck` to better reflect the change?

joelverhagen: I was hoping to leave it as `script` so that future validations can be relaxed with this same parameter.

PR #5526, opened 2018-02-21
[Organizations] Add Organization

scottbommarito:
Allows users to create their own organizations.

Manage Organizations page (with "Add" link):
image
Manage Organizations page (with "Add" link and existing orgs):
image
Create Organization page:
image


src/NuGetGallery/Services/IUserService.cs
```diff
@@ -41,5 +41,7 @@ public interface IUserService
         Task RequestTransformToOrganizationAccount(User accountToTransform, User adminUser);
         
         Task<bool> TransformUserToOrganization(User accountToTransform, User adminUser, string token);
+
+        Task<Organization> CreateOrganization(string username, string emailAddress, User adminUser);
```
skofman1: organizationName

src/NuGetGallery/Services/UserService.cs
```diff
@@ -357,19 +366,85 @@ public bool CanTransformUserToOrganization(User accountToTransform, User adminUs
...
+            var existingUsers = EntitiesContext.Users;
+            if (existingUsers.Any(u => u.Username == organizationName))
+            {
+                throw new EntityException(Strings.UsernameNotAvailable, organizationName);
+            }
+            
+            if (existingUsers.Any(u => u.EmailAddress == emailAddress))
+            {
+                throw new EntityException(Strings.EmailAddressBeingUsed, emailAddress);
+            }
```
skofman1: This will perform two DB queries. The code to create a use uses one query:
[https://github.com/NuGet/NuGetGallery/blob/7774fd305f3825ba31cf288fe43504863cfdb456/src/NuGetGallery/Authentication/AuthenticationService.cs#L266](https://github.com/NuGet/NuGetGallery/blob/7774fd305f3825ba31cf288fe43504863cfdb456/src/NuGetGallery/Authentication/AuthenticationService.cs#L266)

scottbommarito: I saw that code but thought this was syntactically prettier. Didn't realize that this requires two queries however so I'll revert it back.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -357,19 +366,85 @@ public bool CanTransformUserToOrganization(User accountToTransform, User adminUs
...
+            };
+
+            var membership = new Membership { Organization = organization, Member = adminUser, IsAdmin = true };
+
+            organization.Members.Add(membership);
+            adminUser.Organizations.Add(membership);
+
+            OrganizationRepository.InsertOnCommit(organization);
+
+            if (string.IsNullOrEmpty(GetAzureActiveDirectoryCredentialTenant(adminUser)))
```
skofman1: why do we need this check? I thoughs LoginDiscontinuationConfiguration takes care of this.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -357,19 +366,85 @@ public bool CanTransformUserToOrganization(User accountToTransform, User adminUs
...
+            {
+                throw new EntityException(String.Format(CultureInfo.CurrentCulture,
+                        Strings.Organizations_AdminAccountDoesNotHaveTenant, adminUser.Username));
+            }
+
+            // SubscribeOrganizationToTenantPolicy will commit
+            if (!await SubscribeOrganizationToTenantPolicy(organization, adminUser))
```
skofman1: Consider not counting on SecurityPolicyService.SubscribeAsync to commit, but passing it a flag commit=false, and committing here. This pattern is used a lot in PackageService. Makes the code more predictable and readable.

skofman1:
Looks good! I wonder if the UI is friendly enough. Perhaps more text with guidelines, instructions, etc. @anangaur ?

src/NuGetGallery/Services/UserService.cs
scottbommarito: Nope, `LoginDiscontinuationConfiguration` only covers checking the whitelists and has nothing to do with AAD tenants.

src/NuGetGallery/Services/IUserService.cs
scottbommarito: I noticed the `UserService` method has `organizationName` however so I will align it with here.

src/NuGetGallery/Services/UserService.cs
skofman1: I'm confused, I thought "Add new organization" flow was open for everybody, not just the configured list, and not just AAD but also MSA.

src/NuGetGallery/Services/UserService.cs
scottbommarito: makes sense, am fixing

anangaur:
Some observations:
1. Use **Add** instead of **Create**
2. There is no need to show the Administrator line item.
3. Can you bring in logo section on the right and suggest that email is used to look up the logo from gravatar as suggested in the storyboard:
image
2. There is a different page when no Organizations are present:
image

scottbommarito:
@anangaur Alright, will use add!

src/NuGetGallery/Services/UserService.cs
scottbommarito: I thought "add new organization" would be scoped to the same users as the other org features. @anangaur should "add new organization" be open to the public?

src/NuGetGallery/Scripts/gallery/page-add-organization.js
```diff
@@ -0,0 +1,41 @@
...
+
+    function updateGravatar() {
+        var src = defaultImage;
+
+        var email = emailBox.val();
+        if (email.match(/^([0-9a-zA-Z]([-.\w]*[0-9a-zA-Z])*@([0-9a-zA-Z][-\w]*[0-9a-zA-Z]\.)+[a-zA-Z]{2,9})$/)) {
```
joelverhagen: this should be simpler or come from server side

src/NuGetGallery/Scripts/gallery/page-add-organization.js
```diff
@@ -0,0 +1,41 @@
...
...C[P + 10], j, 3200236656); Y = D(Y, X, W, V, C[P + 13], o, 681279174); V = D(V, Y, X, W, C[P + 0], m, 3936430074); W = D(W, V, Y, X, C[P + 3], l, 3572445317); X = D(X, W, V, Y, C[P + 6], j, 76029189); Y = D(Y, X, W, V, C[P + 9], o, 3654602809); V = D(V, Y, X, W, C[P + 12], m, 3873151461); W = D(W, V, Y, X, C[P + 15], l, 530742520); X = D(X, W, V, Y, C[P + 2], j, 3299628645); Y = t(Y, X, W, V, C[P + 0], U, 4096336452); V = t(V, Y, X, W, C[P + 7], T, 1126891415); W = t(W, V, Y, X, C[P + 14], R, 2878612391); X = t(X, W, V, Y, C[P + 5], O, 4237533241); Y = t(Y, X, W, V, C[P + 12], U, 1700485571); V = t(V, Y, X, W, C[P + 3], T, 2399980690); W = t(W, V, Y, X, C[P + 10], R, 4293915773); X = t(X, W, V, Y, C[P + 1], O, 2240044497); Y = t(Y, X, W, V, C[P + 8], U, 1873313359); V = t(V, Y, X, W, C[P + 15], T, 4264355552); W = t(W, V, Y, X, C[P + 6], R, 2734768916); X = t(X, W, V, Y, C[P + 13], O, 1309151649); Y = t(Y, X, W, V, C[P + 4], U, 4149444226); V = t(V, Y, X, W, C[P + 11], T, 3174756917); W = t(W, V, Y, X, C[P + 2], R, 718787259); X = t(X, W, V, Y, C[P + 9], O, 3951481745); Y = K(Y, h); X = K(X, E); W = K(W, v); V = K(V, g) } var i = B(Y) + B(X) + B(W) + B(V); return i.toLowerCase() };
```
joelverhagen: we need source, license sign off on this. 

alternatively, make an endpoint on server side that says "give me a gravatar URL for this email"

src/NuGetGallery/Scripts/gallery/page-add-organization.js
scottbommarito: Found an implementation in `RegisterViewModel` that I have moved to `Constants` and passed in here.

src/NuGetGallery/Scripts/gallery/page-add-organization.js
scottbommarito: Spoke with @skofman1 and found a version of MD5 with source and an MIT license. Have moved it into a separate file (`md5.js`) and added it to the `page-add-organization` bundle.

src/NuGetGallery/Views/Organizations/Add.cshtml
```diff
@@ -0,0 +1,65 @@
...
+        <div class="@ViewHelpers.GetColumnClasses(ViewBag)">
+            <h1 class="text-center">
+                <a href="@Url.ManageMyOrganizations()">Organizations</a>
+                &gt;
+                Add New
+            </h1>
+            <div>
+                <aside class="col-sm-3 col-sm-push-9">
+                    @Html.Label("Logo")
+                    <img src="@Url.Absolute("~/Content/gallery/img/default-package-icon-256x256.png")" class="owner-image img-responsive" height="332" id="gravatar-image" width="332" alt="gravatar" title="Organization logo">
```
joelverhagen: SVG first, fall back to PNG.

joelverhagen: Also, the height and width should match the PNG to avoid distortion

src/NuGetGallery/Views/Users/Organizations.cshtml
```diff
@@ -4,16 +4,25 @@
     Layout = "~/Views/Shared/Gallery/Layout.cshtml";
 }
 
-<section role="main" class="container main-container manage-packages">
+<section role="main" class="container main-container page-manage-packages">
```
chenriksson: Should we introduce `page-manage-organizations`? Or move common logic to something like `common-list-view.less`?

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -47,6 +48,43 @@ protected override void SendEmailChangedConfirmationNotice(User account)
...
+                SendNewAccountEmail(organization);
+                return RedirectToAction(nameof(ManageOrganization), new { accountName = organization.Username });
+            }
+            catch (EntityException e)
+            {
+                errorMessage = e.Message;
+            }
+            catch (Exception e)
+            {
+                errorMessage = e.GetUserSafeMessage();
```
chenriksson: Do we normally handle the general exceptions in our action methods? Is there a specific, known exception we're handling here?

src/NuGetGallery/ViewModels/AddOrganizationViewModel.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.ComponentModel.DataAnnotations;
+
+namespace NuGetGallery
+{
+    public class AddOrganizationViewModel
```
chenriksson: Registering an organization account is very similar to registering a user account... so you could try to reuse code. For instance, why not reuse the RegisterViewModel, and then use inheritance to create a User-specific RegisterUserViewModel that also takes a password.

This might reduce the size of this change too - for instance, regex logic could remain on RegisterViewModel.

src/NuGetGallery/Scripts/gallery/md5.js
```diff
@@ -0,0 +1,251 @@
+﻿/**
+ *
+ *  MD5 (Message-Digest Algorithm)
+ *  http://www.webtoolkit.info/
+ *
+ **/
+
+(function (exports) {
```
chenriksson: This is just for displaying the gravatar at org creation time?

IMHO, I don't think we should do this... at least not as part of the same PR. Organization registration is very similar to User registration, so I think we should reuse as much of the code as possible in this PR. I would add the gravatar in a separate PR, and consider whether we should also do it for User registrations.

src/NuGetGallery/Services/UserService.cs
chenriksson: We need to discuss with @anangaur about how onboarding changes when we open this to everyone.

Initial onboarding required (internal) Organizations to have an AAD policy such that everyone needed an AAD credential with tenant Id that matched the initial admin. I suspect we're going to change to require AAD or MSA for everyone.

I believe MSA now uses AADv2 with the personal/MSA tenant, so we probably just need to check for both AAD and MSA external credential types.

src/NuGetGallery/Views/Users/Organizations.cshtml
scottbommarito: I think given that it's only used in two places it's fine to stay in its existing file. If a third page uses this we should move it to a shared file.

src/NuGetGallery/Services/UserService.cs
scottbommarito: Discussed this with @anangaur previously. I believe we're going to add a checkbox that says "restrict this to your AAD tenant" or something like that that will enable this behavior. For now, however, this is going to be restricted to internal orgs.

src/NuGetGallery/Controllers/OrganizationsController.cs
scottbommarito: Removed this and made the exception thrown if subscribing to the policy fails an `EntityException` instead.

src/NuGetGallery/ViewModels/AddOrganizationViewModel.cs
scottbommarito: That's a good idea but
1) The regex still needs to move because it's used by the script
2) I'm not sure how to deal with the fact that the `Display` annotations for the two different models need to be different (e.g. register account's username shows "username" and not "organization name")

scottbommarito: Also with the introduction of AAD the register flow is going to be very different from this, so I think it makes sense to keep it separate.

src/NuGetGallery/Scripts/gallery/md5.js
scottbommarito: @anangaur asked for it specifically. It shows the Gravatar as you fill in the form.

With the MSA/AAD changes, organization registration flow is going to be a lot different than the register flow on the site--e.g. you don't enter an email, it gets it from the MSA account automatically--so I don't think it's super important to separate these PRs.

src/NuGetGallery/Services/IUserService.cs
```diff
@@ -41,5 +41,7 @@ public interface IUserService
         Task RequestTransformToOrganizationAccount(User accountToTransform, User adminUser);
         
         Task<bool> TransformUserToOrganization(User accountToTransform, User adminUser, string token);
+
+        Task<Organization> AddOrganization(string username, string emailAddress, User adminUser);
```
chenriksson: nit: would name `AddOrganizationAsync`... (I realize I didn't do a good job of this above)

src/NuGetGallery/Services/UserService.cs
chenriksson: Maybe open issue for making AAD tenant policy optional, since it's required before opening this feature to everyone?

PR #5529, opened 2018-02-22
[Orchestrator] Delay the delivery of validation requests

loic-sharma:
Upon receiving a package's validation request, the Orchestrator polls the Gallery database once a minute until it finds the package that should be validated. Unfortunately, the Orchestrator often receives the validation request seconds before the Gallery has had a chance to persist changes to its database. In other words, the Orchestrator often waits a full minute to validate a package, when in reality, it should have only waited a few seconds.

This change aims to drastically reduce the time the Orchestrator waits to start a package's validation by delaying the delivery of the validation request by 5 seconds.

Tracked by https://github.com/NuGet/Engineering/issues/1179
Will be followed by https://github.com/NuGet/Engineering/issues/1180

## Supporting Data

I ran the following Application Insights query:

```
traces
| where message contains "Starting Enqueuing asynchronous package validation"
| join (requests | where name == "POST Packages/VerifyPackage" or name == "PUT Api/PushPackageApi" | project requestTimestamp = timestamp, duration, operation_Id) on operation_Id 
| project requestTimestamp, validationTimestamp = timestamp, duration
| extend timeToEnqueue = ((validationTimestamp - requestTimestamp)/ time(1ms))
| extend timeAfterEnqueue = duration - timeToEnqueue
| summarize percentiles(timeAfterEnqueue, 10, 50, 75, 80, 90, 95, 99), max(timeAfterEnqueue)
```

This `timeAfterEnqueue` estimates the the time upload requests spent after the validation request has been enqueued to persist changes to the database (it also includes the package upload to blob storage). The breakdown for `timeAfterEnqueue` for ALL requests (including failed ones) since December 1st is:

10% | 50% | 75% | 80% | 90% | 95% | 99%
--- | --- | --- | --- | --- | --- | ---
2s | 3s | 3.7s | 4s | 10.3s | 12.4s | 38.6s

This shows that by delaying the delivery of validation requests by 5 seconds, at least 80% of validation requests will begin after waiting 5 seconds.


skofman1:
How future proof is this approach? 
Since it depends on package upload time and DB update time, seems like variations in package size trends/storage performance/DB load/DB scale will impact this 5s magic number.

loic-sharma:
That is correct, the 5 seconds magic number is highly affected by all of the factors you mentioned. If these factors were to change, we may be adding 5 seconds to our total validation time. Considering our current [package validation times](https://blog.nuget.org/20180201/NuGet-package-publishing-workflow-behind-the-scenes.html) (3 minutes at 50%, 13 minutes at 90%), 5 seconds is negligible.

This change is an easy hack that will shave off a minute from validation times. @joelverhagen, @agr, and I have considered other approaches as well, but all of these solutions will be harder to implement.

joelverhagen:
This needs a NuGetDeployment change too.

joelverhagen:
@loic-sharma, could we test this in DEV to make sure the validation durations actually go down as expected? Just push N (maybe like 20) packages before and after the change.

loic-sharma:
[Here's a link](https://jarvis-west.dc.ad.msft.net/dashboard/share/4C3BEC50) to the Orchestrator's time to start validation metric. Today, our 50th percentile is at 58 seconds.

loic-sharma:
Added deployment configuration changes. See [this pull request](https://nuget.visualstudio.com/NuGetMicrosoft/_git/NuGetDeployment/pullrequest/324?_a=overview).

Issue #5378, opened 2018-02-01
Add telemetry which would illustrate spikes in push, unlist, relist


joelverhagen:
We have had a variety of incidents in prod where one or a few package authors execute a massive volume of queries resulting in a big lag in V3 -- due to slow China blob storage queries. (this may be mitigated by APIM)

Additionally, other parts of our system (such as the asynchronous validation pipeline) can be sensitive to big spikes in package pushes.

I think we need a dashboard in Geneva which has the following data points:

1. packages pushed (custom metric from gallery?)
1. packages moving to available state (custom metric from orchestrator)
1. unlisted packages (custom metric from gallery?)
1. relisted packages (custom metric from gallery?)
1. reflow (custom metric from gallery?) -- not strictly necessary but we have in the past done a lot of reflows at once

Items 2 - 4 all cause catalog leaves to be produced and therefore have the potential for causing index lags.

This would be a good place to start when diagnosing cursor/index lags or slow validations.

joelverhagen:
It looks like we already have the `PackagePush` custom event in the gallery. Unfortunately I don't think custom events show up in AI -> geneva continuous export...

loic-sharma:
We should consider adding distributed tracing to our services (like [this](http://opentracing.io/) with [this](https://zipkin.io/)). Today, we rely on logs to determine the breakdown of performance of actions like package pushes. This won't scale as we continue adding more services and validations.

PR #5540, opened 2018-02-23
Add telemetry for unlist, list, delete, reflow, and revalidate

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/5378.

I chose not to pass down `User` and `IIdentity` for now. I wanted to emit telemetry from the service layer and not the controller layer so that we would not need to duplicate emits in `ApiController` vs. `PackagesController` and we would not emit telemetry when the service call no-ops (e.g. unlisting an already unlisted package).

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1052,7 +1052,8 @@ namespace NuGetGallery
             var reflowPackageService = new ReflowPackageService(
                 _entitiesContext,
                 (PackageService)_packageService,
-                _packageFileService);
+                _packageFileService,
+                _telemetryService);
 
             try
             {
--- a/src/NuGetGallery/Services/ITelemetryService.cs
+++ b/src/NuGetGallery/Services/ITelemetryService.cs
@@ -13,6 +13,18 @@ namespace NuGetGallery
 
         void TrackPackagePushEvent(Package package, User user, IIdentity identity);
 
+        void TrackPackageUnlisted(Package package);
+
+        void TrackPackageListed(Package package);
+
+        void TrackPackageDelete(Package package, bool isHardDelete);
+
+        void TrackPackageReflow(Package package);
+
+        void TrackPackageHardDeleteReflow(string packageId, string packageVersion);
+
+        void TrackPackageRevalidate(Package package);
+
         void TrackPackageReadMeChangeEvent(Package package, string readMeSourceType, PackageEditReadMeState readMeState);
 
         void TrackCreatePackageVerificationKeyEvent(string packageId, string packageVersion, User user, IIdentity identity);
--- a/src/NuGetGallery/Services/ReflowPackageService.cs
+++ b/src/NuGetGallery/Services/ReflowPackageService.cs
@@ -14,15 +14,18 @@ namespace NuGetGallery
         private readonly IEntitiesContext _entitiesContext;
         private readonly IPackageService _packageService;
         private readonly IPackageFileService _packageFileService;
+        private readonly ITelemetryService _telemetryService;
 
         public ReflowPackageService(
             IEntitiesContext entitiesContext,
             IPackageService packageService,
-            IPackageFileService packageFileService)
+            IPackageFileService packageFileService,
+            ITelemetryService telemetryService)
         {
-            _entitiesContext = entitiesContext;
-            _packageService = packageService;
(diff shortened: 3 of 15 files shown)


src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -252,7 +351,7 @@ private void TrackPackageForEvent(string eventValue, string packageId, string pa
                 throw new ArgumentNullException(nameof(identity));
             }
 
-            TrackEvent(eventValue, properties => {
+            TrackMetric(metricName, 1, properties => {
```
shishirx34: I don't really understand the need to emit a metric rather than an event. Emitting metric of 1 is equivalent to sending an Event.

The reason we should use `Events` over `metrics` when is Sampling. If we were to turn on sampling, the AI provides the aggregation needed. There is no 'Performance' analysis done by sending a metric of 1.

> If sampling is in operation, the itemCount property shows a value greater than 1. For example itemCount==10 means that of 10 calls to trackEvent(), the sampling process only transmitted one of them. To get a correct count of custom events, you should use therefore use code such as customEvent | summarize sum(itemCount).

From App Insights documentation:

> It is recommended to aggregate metrics before sending them from your app, to reduce bandwidth, cost and to improve performance.

joelverhagen: > I don't really understand the need to emit a metric rather than an event. Emitting metric of 1 is equivalent to sending an Event.

AI to Geneva export does not take events as hot path metrics.

src/NuGetGallery/Services/ReflowPackageService.cs
```diff
@@ -75,7 +78,10 @@ public class ReflowPackageService
                         // 5) Update IsLatest so that reflow can correct concurrent updates (see Gallery #2514)
                         await _packageService.UpdateIsLatestAsync(package.PackageRegistration, commitChanges: false);
 
-                        // 6) Save and profit
+                        // 6) Emit telemetry.
+                        _telemetryService.TrackPackageReflow(package);
+
+                        // 7) Save and profit
                         await _entitiesContext.SaveChangesAsync();
```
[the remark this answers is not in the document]
joelverhagen: In general we should not be worried about off-by-one errors on telemetry. I chose to do this to match how we emit auditing which is also done before commit.

src/NuGetGallery/Services/TelemetryService.cs
shishirx34: What a pain. 😆

src/NuGetGallery/Services/ReflowPackageService.cs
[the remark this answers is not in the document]
joelverhagen: > I disagree on the part about auditing, in most cases we log audit after we commit. 

In all of the cases in my PR, auditing is done before commit.

src/NuGetGallery/Services/ITelemetryService.cs
```diff
@@ -13,6 +13,18 @@ public interface ITelemetryService
...
 
+        void TrackPackageUnlisted(Package package);
+
+        void TrackPackageListed(Package package);
+
+        void TrackPackageDelete(Package package, bool isHardDelete);
+
+        void TrackPackageReflow(Package package);
+
+        void TrackPackageHardDeleteReflow(string packageId, string packageVersion);
```
skofman1: why is this interesting?

joelverhagen: Because it results in V3 work (catalog2registration, catalog2dnx, catalog2lucene)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1052,7 +1052,8 @@ public virtual ActionResult Delete(string id, string version)
             var reflowPackageService = new ReflowPackageService(
```
loic-sharma: >reflowPackageService [](start = 16, length = 20)

I suppose this is out of scope for this PR, but, could we instead share the `reflowPackageService` instance across requests?

joelverhagen: Why? Risk of shared state seems to out way any performance benefit until we have measurements suggesting otherwise.

loic-sharma: The state is already shared by the controller's fields.

PR #5556, opened 2018-02-28
[Organizations] Notifications for Organization membership changes

scottbommarito:
This PR improves the organization membership change flow.

Instead of immediately adding a member to an organization, members are now in a pending state. A request is sent to the email of the user who was added. If they follow a link in the email, they accept the request and will be made a member of the organization.

Much like package ownership requests, I have added the following functionality
- pending owners can reject a request and delete it from the database
- existing owners can cancel a request and delete it from the database

Note: two thirds of this PR are unit tests

Manage Organization Page with pending user:

image

Page for accepting a membership request:

image

Page for rejecting a membership request:

image

Email sent to a user with a request to join an organization:

image

Email sent to the organization after a user joins the organization:

image

Email sent to the organization after a user rejects a request to join:

image

Email sent to a user after their request to join an organization is cancelled:

image

Email sent to a user when their membership is changed (collaborator to admin and vice versa):

image

Email sent to a user when they are removed from an organization:

image


src/NuGetGallery/Scripts/gallery/page-manage-organization.js
```diff
@@ -178,6 +180,20 @@
...
+                    if (member.Username.toLocaleLowerCase() === self.NewMemberUsername().toLocaleLowerCase()) {
+                        memberExists = true;
+                    }
+                });
+
+                if (memberExists) {
+                    var error = "'" + self.NewMemberUsername() + "' is already a member or pending member.";
+                    self.Error(error);
+                    return;
+                }
```
chenriksson: nit: Do we need the client side validation, or can we just have the server return the error?

src/NuGetGallery/Services/UserService.cs
```diff
@@ -83,12 +93,103 @@ public class UserService : IUserService
...
+            var request = FindMembershipRequestByUsername(organization, memberName);
+            if (request == null || request.ConfirmationToken != confirmationToken)
+            {
+                throw new EntityException(string.Format(CultureInfo.CurrentCulture,
+                    Strings.RejectMembershipRequest_NotFound, memberName));
+            }
+
+            organization.MemberRequests.Remove(request);
+            await EntitiesContext.SaveChangesAsync();
+        }
```
chenriksson: nit: `RejectMembershipRequestAsync` is almost identical to `CancelMembershipRequestAsync`. Could it just validate the token and then call cancel?

src/NuGetGallery/Services/UserService.cs
```diff
@@ -83,12 +93,103 @@ public class UserService : IUserService
...
+            {
+                // Ensure that the new member meets the AAD tenant policy for this organization.
+                var policyResult = await SecurityPolicyService.EvaluateOrganizationPoliciesAsync(
+                    SecurityPolicyAction.JoinOrganization, organization, member);
+                if (policyResult != SecurityPolicyResult.SuccessResult)
+                {
+                    // Delete the request before throwing.
+                    await EntitiesContext.SaveChangesAsync();
+
+                    throw new EntityException(policyResult.ErrorMessage);
```
chenriksson: Should we delete the request here? Or maybe we should at least give a more detailed message? Just trying to think of the best user experience...

maybe `Your request cannot be completed, as you no longer meet the requirements for this organization. Contact support for more details.`

src/NuGetGallery/Services/UserService.cs
```diff
@@ -132,6 +233,8 @@ public class UserService : IUserService
                     Strings.UpdateOrDeleteMember_MemberNotFound, memberName));
             }
 
+            var memberToRemove = membership.Member;
```
chenriksson: nit: variable isn't necessary as it's only used as a return value. Maybe just change to `return membership.Member` at end of method?

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -788,6 +788,41 @@ public static string AddOrganizationMember(this UrlHelper url, string accountNam
...
+        {
+            return url.HandleOrganizationMembershipRequest("ConfirmMemberRequest", request, relativeUrl);
+        }
+
+        public static string RejectOrganizationMembershipRequest(this UrlHelper url, MembershipRequest request, bool relativeUrl = true)
+        {
+            return url.HandleOrganizationMembershipRequest("RejectMemberRequest", request, relativeUrl);
+        }
+
+        private static string HandleOrganizationMembershipRequest(this UrlHelper url, string routeName, MembershipRequest request, bool relativeUrl = true)
```
chenriksson: nit: wrap long lines

src/NuGetGallery/ViewModels/HandleOrganizationMembershipRequestModel.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public class HandleOrganizationMembershipRequestModel
+    {
+        /// <summary>
+        /// If true, this is an attempt to confirm a request.
+        /// If false, this is an attempt to reject a request.
+        /// </summary>
+        public bool Confirm { get; }
+        public string OrganizationName { get; }
+        public bool Successful { get; }
```
chenriksson: nit: Successful property could return true/false based on `FailureReason` being set, instead of being set independently.

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -681,6 +681,152 @@ public void SendPackageDeletedNotice(Package package, string packageUrl, string
             }
         }
 
+        public void SendOrganizationMembershipRequest(Organization organization, User newUser, User adminUser, bool isAdmin, string profileUrl, string confirmationUrl, string rejectionUrl)
```
chenriksson: nit: wrap long lines

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -681,6 +681,152 @@ public void SendPackageDeletedNotice(Package package, string packageUrl, string
...
+        {
+            if (!newUser.EmailAllowed)
+            {
+                return;
+            }
+
+            var membershipLevel = isAdmin ? "an admin" : "a collaborator";
+
+            string subject = $"[{Config.GalleryOwner.DisplayName}] The user '{adminUser.Username}' would like you to become {membershipLevel} of their organization, '{organization.Username}'.";
```
chenriksson: Subject is long - is that what we normally do?

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -681,6 +681,152 @@ public void SendPackageDeletedNotice(Package package, string packageUrl, string
...
+        public void SendOrganizationMembershipRequestRejectedNotice(Organization organization, User pendingUser)
+        {
+            if (!organization.EmailAllowed)
+            {
+                return;
+            }
+
+            string subject = $"[{Config.GalleryOwner.DisplayName}] The user '{pendingUser.Username}' has rejected your request for them to become a member of your organization.";
+
+            string body = string.Format(CultureInfo.CurrentCulture, $@"The user '{pendingUser.Username}' has rejected your request for them to become a member of your organization.
```
chenriksson: nit: would use `declined` instead of `rejected`

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -681,6 +681,152 @@ public void SendPackageDeletedNotice(Package package, string packageUrl, string
...
+            {
+                return;
+            }
+
+            var membershipLevel = membership.IsAdmin ? "an admin" : "a collaborator";
+            var member = membership.Member;
+
+            string subject = $"[{Config.GalleryOwner.DisplayName}] The user '{member.Username}' is now {membershipLevel} of organization '{organization.Username}'.";
+
+            string body = string.Format(CultureInfo.CurrentCulture, $@"The user '{member.Username}' is now {membershipLevel} of organization '{organization.Username}'.
```
chenriksson: nit: '...is now *an* administrator' or '...is now *a* collaborator'

src/NuGetGallery/Views/Organizations/HandleOrganizationMembershipRequest.cshtml
```diff
@@ -0,0 +1,74 @@
...
+    </div>
+    <div class="@ViewHelpers.GetColumnClasses(ViewBag)">
+        @if (Model.Confirm)
+        {
+            if (Model.Successful)
+            {
+                <p class="text-center">
+                    You are now a member of the
+                    '<strong><a href="@Url.User(@Model.OrganizationName)" title="The @Model.OrganizationName organization">@Model.OrganizationName</a></strong>' organization!
+                </p>
```
chenriksson: For success cases, should we instead redirect the user to the `Manage Organization` view like we do for transform confirmation?

src/NuGetGallery/Services/UserService.cs
scottbommarito: I think I'll make a helper function that has a token parameter but ignores it if it's null and then make these two both call it--there will be duplication of work and maybe DB calls if `Reject` has to get the request to validate the token.

src/NuGetGallery/Services/UserService.cs
scottbommarito: On second thought, I think not deleting the request and then showing something like the following error message would be best:

```
Your request cannot be completed as you no longer meet the requirements for this organization.
{policyResult.ErrorMessage}
Please contact support for more details.
```

src/NuGetGallery/Services/UserService.cs
scottbommarito: In past experiences, once an EF object is destroyed (e.g. by `organization.Members.Remove(membership)`), its references to other objects are destroyed. If I call `membership.Member` after `SaveChangesAsync` I'm pretty sure it gives me `null` or some other garbage.

src/NuGetGallery/ViewModels/HandleOrganizationMembershipRequestModel.cs
scottbommarito: Good catch

src/NuGetGallery/Views/Organizations/HandleOrganizationMembershipRequest.cshtml
scottbommarito: I was thinking about this, and I think that sounds better.

src/NuGetGallery/Scripts/gallery/page-manage-organization.js
scottbommarito: Server side returns OK and doesn't override a request if one already exists. We need the client side here because otherwise the UI doesn't understand when someone is already added.

E.g.
A requests C is a collaborator and creates request with token "abcd" that grants collaborator
B requests C is an admin and the existing request with token "abcd" is updated to grant admin instead of collaborator

src/NuGetGallery/Services/MessageService.cs
scottbommarito: Almost all of our emails have the same subject as the first sentence in the body.

src/NuGetGallery/Services/MessageService.cs
scottbommarito: What do you mean? `membershipLevel` here is either `an admin` or `a collaborator`.

chenriksson: I'm suggesting adding `a` or `an` in front of the membershipLevel

scottbommarito: Doesn't it already?

```
...is now {membershipLevel} of...
```
becomes
```
...is now an admin of...
```
because `{membershipLevel}` is `an admin` and not just `admin`.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -70,8 +72,99 @@ public virtual ActionResult ManageOrganization(string accountName)
...
             try
             {
-                var membership = await UserService.AddMemberAsync(account, memberName, isAdmin);
-                return Json(new OrganizationMemberViewModel(membership));
+                var request = await UserService.AddMembershipRequestAsync(account, memberName, isAdmin);
+
+                var profileUrl = Url.User(account, relativeUrl: false);
+                var confirmUrl = Url.AcceptOrganizationMembershipRequest(request, relativeUrl: false);
+                var rejectUrl = Url.RejectOrganizationMembershipRequest(request, relativeUrl: false);
+                MessageService.SendOrganizationMembershipRequest(account, request.NewMember, GetCurrentUser(), request.IsAdmin, profileUrl, confirmUrl, rejectUrl);
```
cristinamanum: Will the initiator of this request know if this send email failed? Will it be an exception or will it fail silently? I think that it should not fail silently.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -70,8 +72,99 @@ public virtual ActionResult ManageOrganization(string accountName)
...
+            if (account == null)
+            {
+                return new HttpStatusCodeResult(HttpStatusCode.NotFound);
+            }
+
+            try
+            {
+                var member = GetCurrentUser();
+                await UserService.RejectMembershipRequestAsync(account, member.Username, confirmationToken);
+                MessageService.SendOrganizationMembershipRequestRejectedNotice(account, member);
```
cristinamanum: Will the sender of this "request to join" be notified that its request was rejected?

src/NuGetGallery/Controllers/OrganizationsController.cs
scottbommarito: Even if this call succeeds there's a chance that SendGrid fails or bounces or something. It's not really possible to determine from the Gallery side that the email was successfully sent.

This problem exists with the existing package owner requests flow.

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -681,6 +681,153 @@ public void SendPackageDeletedNotice(Package package, string packageUrl, string
             }
         }
 
+        public void SendOrganizationMembershipRequest(
+            Organization organization, User newUser, User adminUser, bool isAdmin, string profileUrl, string confirmationUrl, string rejectionUrl)
+        {
+            if (!newUser.EmailAllowed)
+            {
+                return;
```
cristinamanum: In this case the sender will never know what happened with his request, is this fine?

src/NuGetGallery/Controllers/OrganizationsController.cs
scottbommarito: It's sent to the organization email, which is likely a DL containing the user that sent the request.

src/NuGetGallery/Services/MessageService.cs
scottbommarito: Yes, because the sender explicitly turned off email notifications.

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -681,6 +681,153 @@ public void SendPackageDeletedNotice(Package package, string packageUrl, string
...
+
+                mailMessage.To.Add(newUser.ToMailAddress());
+                SendMessage(mailMessage);
+            }
+        }
+
+        public void SendOrganizationMembershipRequestRejectedNotice(Organization organization, User pendingUser)
+        {
+            if (!organization.EmailAllowed)
+            {
```
cristinamanum: Should not be enforced as any organization to have an available email?

scottbommarito: `EmailAllowed` is a true or false setting that the account can set regarding whether or not they want email notifications or not. Someone can turn `EmailAllowed` off despite having an email address.

Issue #5551, opened 2018-02-27
In build.ps1 scripts that fetch from ServerCommon, set SSL protocols


joelverhagen:
A lot of machines default to SSL 3.0 and TLS 1.0, which is not compatible with GitHub.

```
System.Net.WebException: The request was aborted: Could not create SSL/TLS secure channel.
   at Microsoft.PowerShell.Commands.WebRequestPSCmdlet.GetResponse(WebRequest request)
   at Microsoft.PowerShell.Commands.WebRequestPSCmdlet.ProcessRecord()
```

Repositories:
- [x] NuGetGallery
- [x] NuGet.Jobs
- [x] NuGet.Server
- [x] NuGet.Services.Metadata
- [x] NuGet.Services.EndToEnd

joelverhagen:
Workaround is to run this on your machine:
```
Set-ItemProperty -Path 'HKLM:\SOFTWARE\Wow6432Node\Microsoft\.NetFramework\v4.0.30319' -Name 'SchUseStrongCrypto' -Value '1' -Type DWord
Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\.NetFramework\v4.0.30319' -Name 'SchUseStrongCrypto' -Value '1' -Type DWord
```

PR #5563, opened 2018-02-28
Enable TLS 1.2 in build.ps1

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/5551.


skofman1:
Shouldn't this go to the common script?

joelverhagen:
> Shouldn't this go to the common script?

Spoke offline. This is a bootstrapping problem. We can't even get the common script.

Issue #5375, opened 2018-01-31
Package that failed validation is visible under "Published Packages" heading


joelverhagen:
If all versions of a package are in the validating or failed validation state, they should not be under a heading called "Published Packages"

## Manage packages page

![image](https://user-images.githubusercontent.com/94054/35599582-3705852a-05de-11e8-95fc-d073a3b249d6.png)
![image](https://user-images.githubusercontent.com/94054/35599590-427750f0-05de-11e8-8f6f-38bfd7cbba62.png)

# Package details page

![image](https://user-images.githubusercontent.com/94054/35599596-486834a2-05de-11e8-9e7e-104557774222.png)

joelverhagen:
The page is not published 😄

xavierdecoster:
This view currently only queries for the "latest" package version and splits on the `Listed` property.

Is it even possible for a package version to be marked as "latest", "listed" and have a status of either "validating", "validation failed", or "deleted", all at the same time?
Should a package registration for which all versions failed validation even have a package marked as "listed" or "latest"?

In other words: is it safe to assume that if the "latest" package doesn't have a status of "Available", all other versions also are unavailable?

xavierdecoster:
@joelverhagen can you please verify whether the Zip64Package is shown under the unlisted packages section on your account page?

PR #5573, opened 2018-03-02
Improve query performance for Manage Packages

chenriksson:
First optimization for #5461, which covers 2 fixes:
1. Regression from d8371bd

In **UsersController**, fix sent to query off the request table instead of the large list of packages. Performance was hard to measure, as it was hanging due to the regression.

2. Optimize packages query to only sort in DB, based on IsLatest columns

In **PackageService**, modified the EF query to sort and select based on IsLatest state in DB. Previously we'd read all rows and sort in memory (on NuGetVersion as last resort) to handle cases where IsLatest wasn't set due to concurrency issues.

The second optimization helps mostly in scenarios with many versions, as opposed to many registrations. Results below:

**Scenario 1:  59K registrations, 59K versions (NuGetStagingTest on DEV)**

  - BEFORE: 48.54, 49.41, 60, 35.74, 49.13 => 45.71s

  - AFTER: 36.12, 49.95, 43.13, 40.32, 46.30 => 41.47 **(-9.28%)**

**Scenario 2: 100 registrations, 52K versions (DEV, various packages)**

  - BEFORE: 126, 144, 90, 108, 78 => 100.5s

  - AFTER: 2.71, 3.15, 2.28, 2.23, 2.26 => 2.37s **(-97.64%)**

### DIFF
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -378,7 +378,7 @@ namespace NuGetGallery
             var orgSent = currentUser.Organizations
                 .Where(m => ActionsRequiringPermissions.HandlePackageOwnershipRequest.CheckPermissions(currentUser, m.Organization) == PermissionsCheckResult.Allowed)
                 .SelectMany(m => _packageOwnerRequestService.GetPackageOwnershipRequests(requestingOwner: m.Organization));
-            var sent = userReceived.Union(orgReceived);
+            var sent = userSent.Union(orgSent);
 
             var ownerRequests = new OwnerRequestsViewModel(received, sent, currentUser, _packageService);
 
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -366,13 +366,19 @@ namespace NuGetGallery
                 .Select(p => new ListPackageItemViewModel(p, currentUser)).OrderBy(p => p.Id)
                 .ToList();
 
+            // find all received ownership requests
             var userReceived = _packageOwnerRequestService.GetPackageOwnershipRequests(newOwner: currentUser);
             var orgReceived = currentUser.Organizations
                 .Where(m => ActionsRequiringPermissions.HandlePackageOwnershipRequest.CheckPermissions(currentUser, m.Organization) == PermissionsCheckResult.Allowed)
                 .SelectMany(m => _packageOwnerRequestService.GetPackageOwnershipRequests(newOwner: m.Organization));
             var received = userReceived.Union(orgReceived);
-            
-            var sent = packages.SelectMany(p => _packageOwnerRequestService.GetPackageOwnershipRequests(package: p.PackageRegistration));
+
+            // find all sent ownership requests
+            var userSent = _packageOwnerRequestService.GetPackageOwnershipRequests(requestingOwner: currentUser);
+            var orgSent = currentUser.Organizations
+                .Where(m => ActionsRequiringPermissions.HandlePackageOwnershipRequest.CheckPermissions(currentUser, m.Organization) == PermissionsCheckResult.Allowed)
+                .SelectMany(m => _packageOwnerRequestService.GetPackageOwnershipRequests(requestingOwner: m.Organization));
+            var sent = userReceived.Union(orgReceived);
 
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -230,16 +230,45 @@ public IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted,
...
+                .Include(p => p.PackageRegistration.Owners)
+                .ToList();
+            }
+
+            // Do a best effort of retrieving the latest version. Note that UpdateIsLatest has had concurrency issues
+            // where sometimes packages no rows with IsLatest set. In this case, we'll just select the last inserted
+            // row (descending [Key]) as opposed to reading all rows into memory and sorting on NuGetVersion.
+            return packages
+                // order booleans desc so that true (1) comes first
+                .OrderByDescending(p => p.IsLatestStableSemVer2)
```
joelverhagen: 👏

chenriksson:
@joelverhagen Can you review my last push (663bceb)? I had a bug where I wasn't getting the correct latest version... EF doesn't preserve the ordering done before groupby like linq-to-objects does.

Also added a fix for #5375: `.ThenByDescending(p => p.Listed)` should order published versions first. UsersController.Packages applies filter `.Where(p => p.Listed)` which should exclude these from the Published list. (I'm assuming `!Listed` when `PackageState != Available`... but correct me if I'm wrong and I can also look at `PackageState`)

**Scenario 2: 100 registrations, 52K versions (re-run morning of 3/5)**
- BEFORE: 72s, 84s, 72s, 84s, 84s => 78s
- AFTER: 4.6s, 4.64s, 4.52s, 4.55s, 4.48s => 4.54s (-94.18%)

joelverhagen:
> EF doesn't preserve the ordering done before groupby like linq-to-objects does.

WOW! That's a serious gotcha.

> `.ThenByDescending(p => p.Listed)`

This doesn't seem 100% critical since this is a package owner view, but seems fine.

> I'm assuming `!Listed` when `PackageState != Available`... 

Invalid assumption. You need to filter out non-Available if you actually want that.

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -1006,9 +1006,10 @@ public override void ReturnsOnlyLatestStableSemVer2PackageIfBothExist(User curre
             public override void ReturnsOnlyLatestStablePackageIfNoLatestStableSemVer2Exist(User currentUser, User packageOwner)
               => base.ReturnsOnlyLatestStablePackageIfNoLatestStableSemVer2Exist(currentUser, packageOwner);
 
-            [MemberData(nameof(TestData_RoleVariants))]
+            [Theory(Skip = "Disabled: FindPackagesByAnyMatchingOwner does not order by NuGetVersion for better performance")]
             public override void ReturnsCorrectLatestVersionForMixedSemVer2AndNonSemVer2PackageVersions_IncludeUnlistedTrue(User currentUser, User packageOwner)
```
skofman1: Can we remove the test? What's the benefit of leaving it here?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -366,13 +366,19 @@ public virtual ActionResult Packages()
...
             var received = userReceived.Union(orgReceived);
-            
-            var sent = packages.SelectMany(p => _packageOwnerRequestService.GetPackageOwnershipRequests(package: p.PackageRegistration));
+
+            // find all sent ownership requests
+            var userSent = _packageOwnerRequestService.GetPackageOwnershipRequests(requestingOwner: currentUser);
+            var orgSent = currentUser.Organizations
+                .Where(m => ActionsRequiringPermissions.HandlePackageOwnershipRequest.CheckPermissions(currentUser, m.Organization) == PermissionsCheckResult.Allowed)
+                .SelectMany(m => _packageOwnerRequestService.GetPackageOwnershipRequests(requestingOwner: m.Organization));
+            var sent = userReceived.Union(orgReceived);
```
cristinamanum: Is this the  same as line 374? Should it be  
var sent = userSent.Union(orgSent)?

chenriksson: Wow, great catch @cristinamanum !

PR #5576, opened 2018-03-05
[Organizations] Minor bug fixes from dev

scottbommarito:
1 - Fetch the image immediately for the text box on the add page if the form is already filled (e.g. if form submission fails, show the image from the email already.
2 - Don't use `string.Equals` in EF queryables because it throws on Azure SQL lol.
3 - Show the `Manage Organizations` page for users who are allowed to create an org.

### DIFF
--- a/src/NuGetGallery/Services/UserService.cs
+++ b/src/NuGetGallery/Services/UserService.cs
@@ -383,9 +383,7 @@ namespace NuGetGallery
             }
             
             var existingUserWithIdentity = EntitiesContext.Users
-                .FirstOrDefault(u => 
-                    u.Username.Equals(username, StringComparison.OrdinalIgnoreCase) || 
-                    string.Equals(u.EmailAddress, emailAddress, StringComparison.OrdinalIgnoreCase));
+                .FirstOrDefault(u => u.Username == username || u.EmailAddress == emailAddress);
             if (existingUserWithIdentity != null)
             {
                 if (existingUserWithIdentity.Username.Equals(username, StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Views/NuGetViewBase.cs
```diff
@@ -29,6 +29,8 @@ public User CurrentUser
             get { return NuGetContext.CurrentUser; }
         }
 
+        public IContentObjectService ContentObjectService => DependencyResolver.Current.GetService<IContentObjectService>();
```
chenriksson: Would use `Lazy<>` on these

src/NuGetGallery/Services/UserService.cs
```diff
@@ -383,9 +383,7 @@ public bool CanTransformUserToOrganization(User accountToTransform, User adminUs
             }
             
             var existingUserWithIdentity = EntitiesContext.Users
-                .FirstOrDefault(u => 
-                    u.Username.Equals(username, StringComparison.OrdinalIgnoreCase) || 
-                    string.Equals(u.EmailAddress, emailAddress, StringComparison.OrdinalIgnoreCase));
+                .FirstOrDefault(u => u.Username == username || u.EmailAddress == emailAddress);
```
joelverhagen: Yeah this is already case-insensitive thanks to our DB collation.

scottbommarito: Yup, was 500'ing before with the `string.Equals`

src/NuGetGallery/Views/NuGetViewBase.cs
scottbommarito: done

Issue #5443, opened 2018-02-13
[BugBash] Fix messaging in link MS account email notification


chenriksson:
Scenario 2, per @skofman1 (Issue 3)

Regarding e-mail text: [NuGet Gallery] Microsoft account... 

![image](https://user-images.githubusercontent.com/15364949/40145977-b6262cc8-5918-11e8-86fc-bf7a4a08aa41.png)

shishirx34:
Not a high priority and an existing issue. Moving to backlog to be addressed later.

PR #5578, opened 2018-03-06
[Organization] Update existing email notifications for organizations

scottbommarito:
1 - If the target of the email is an organization, updated any reference of `you` or `your account` with `your organization`
2 - If the target of a package ownership email is an organization, the email sends to the admins of the organization and not the organization DL (because org admins are the only users who can handle these requests)

Messaging pending review with @anangaur


chenriksson:
Can you take care of this?
https://github.com/NuGet/NuGetGallery/issues/5443

scottbommarito:
@chenriksson looked into #5443, seems hilariously non-trivial, should be fixed in a separate change

PR #5580, opened 2018-03-06
[WebScout] Remove clientaccesspoliy.xml

loic-sharma:
The `clientaccesspolicy.xml` file is supposedly only used by Silverlight clients. This change removes it to remove the WebScout warning.


blowdart:
👍

PR #5581, opened 2018-03-06
Add CopyFileAsync to ICoreFileStorageService

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/1190
Progress on https://github.com/NuGet/Engineering/issues/1203

### DIFF
--- a/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Net;
@@ -16,6 +17,15 @@ namespace NuGetGallery
 {
     public class CloudBlobCoreFileStorageService : ICoreFileStorageService
     {
+        /// <summary>
+        /// This is the maximum duration for <see cref="CopyFileAsync(string, string, string, string)"/> to poll,
+        /// waiting for a package copy to complete. The value picked today is based off of the maximum duration we wait
+        /// when uploading files to Azure China blob storage. Note that in cases when the copy source and destination
+        /// are in the same container, the copy completed immediately and no polling is necessary.
+        /// </summary>
+        private static readonly TimeSpan MaxCopyDuration = TimeSpan.FromMinutes(10);
+        private static readonly TimeSpan CopyPollFrequency = TimeSpan.FromMilliseconds(500);
+
         private static readonly HashSet<string> KnownPublicFolders = new HashSet<string> {
             CoreConstants.PackagesFolderName,
             CoreConstants.PackageBackupsFolderName,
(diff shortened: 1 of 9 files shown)


src/NuGetGallery/Services/FileSystemFileStorageService.cs
```diff
@@ -168,6 +169,34 @@ public Task SaveFileAsync(string folderName, string fileName, Stream packageFile
             return Task.FromResult(0);
         }
 
+        public Task CopyFileAsync(string srcFolderName, string srcFileName, string destFolderName, string destFileName)
```
skofman1: check input

src/NuGetGallery/Services/FileSystemService.cs
```diff
@@ -51,5 +51,10 @@ public IFileReference GetFileReference(string path)
             var info = new FileInfo(path);
             return info.Exists ? new LocalFileReference(info) : null;
         }
+
+        public virtual void Copy(string sourceFileName, string destFileName, bool overwrite)
+        {
+            File.Copy(sourceFileName, destFileName, overwrite);
```
skofman1: can we make this async? [https://stackoverflow.com/questions/882686/asynchronous-file-copy-move-in-c-sharp](https://stackoverflow.com/questions/882686/asynchronous-file-copy-move-in-c-sharp)

src/NuGetGallery/Services/FileSystemFileStorageService.cs
```diff
@@ -168,6 +169,34 @@ public Task SaveFileAsync(string folderName, string fileName, Stream packageFile
...
+        }
+
+        private string HashFile(string filePath)
+        {
+            using (var hashAlgorithm = HashAlgorithm.Create(CoreConstants.Sha512HashAlgorithmId))
+            using (var fileStream = _fileSystemService.OpenRead(filePath))
+            {
+                return Convert.ToBase64String(hashAlgorithm.ComputeHash(fileStream));
+            }
+        }
```
skofman1: What's the perf of this? Please test on files of size 250M

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -100,6 +104,60 @@ public CloudBlobCoreFileStorageService(ICloudBlobClient client)
             }
         }
 
+        public async Task CopyFileAsync(string srcFolderName, string srcFileName, string destFolderName, string destFileName)
```
skofman1: check input

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -16,6 +17,9 @@ namespace NuGetGallery
 {
     public class CloudBlobCoreFileStorageService : ICoreFileStorageService
     {
+        private static readonly TimeSpan MaxCopyDuration = TimeSpan.FromMinutes(10);
+        private static readonly TimeSpan CopyPollFrequency = TimeSpan.FromMilliseconds(100);
```
skofman1: seems too small. What is this number based on?

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -16,6 +17,9 @@ namespace NuGetGallery
 {
     public class CloudBlobCoreFileStorageService : ICoreFileStorageService
     {
+        private static readonly TimeSpan MaxCopyDuration = TimeSpan.FromMinutes(10);
```
skofman1: what is this number based on?

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -100,6 +104,60 @@ public CloudBlobCoreFileStorageService(ICloudBlobClient client)
...
+
+            var stopwatch = Stopwatch.StartNew();
+            while (destBlob.CopyState.Status == CopyStatus.Pending
+                   && stopwatch.Elapsed < MaxCopyDuration)
+            {
+                await destBlob.FetchAttributesAsync();
+                await Task.Delay(CopyPollFrequency);
+            }
+
+            if (destBlob.CopyState.Status == CopyStatus.Pending)
```
skofman1: should we kill the copy operation in those cases?

src/NuGetGallery/Services/FileSystemService.cs
joelverhagen: I'd rather use the high level file system API on `File`. We don't use this in production so I'd rather favor small code. We can switch to async later if the performance will be helped.

src/NuGetGallery/Services/FileSystemFileStorageService.cs
joelverhagen: Perf is bound by disk. Note we don't use the `FileSystemFileStorageService` in production. We can optimize this later if necessary.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
joelverhagen: Agree, I've seen some examples online that use 500ms.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
joelverhagen: This is the max operation timeout for China uploads.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
joelverhagen: I decided against this because we can still recover if the caller calls this `CopyFileAsync` method later. Consider if the timeout occurs, exception is thrown, and orchestrator tries again via service bus message. Things can still work out.

I would rather take this approach than aborting and trying again over and over.

joelverhagen: (since would rather trust blob storage server side to get the job done eventually than us trying to start it every N minutes)

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
joelverhagen: Fixed.

src/NuGetGallery/Services/FileSystemFileStorageService.cs
joelverhagen: Fixed.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
joelverhagen: Fixed.

src/NuGetGallery/Services/FileSystemFileStorageService.cs
joelverhagen: I added this so the method would have similar semantics to the blob storage implementation (which no-ops if the destination file is the same).

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
chenriksson: Comment would be good here

src/NuGetGallery/Services/FileSystemFileStorageService.cs
skofman1: We still need to understand the perf. If the hash takes much longer then the copy itself, doesn't make sense to do it.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
skofman1: what if we trigger a copy operation and the previous copy is still in progress?

joelverhagen: We won't even start. We have a precondition `If-None-Match: *`.

src/NuGetGallery/Services/FileSystemFileStorageService.cs
joelverhagen: Sure, I'll get some numbers on disk and SSD.

joelverhagen: You're right. `File.Copy` is freakishly fast and file comparison is slow.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
joelverhagen: Fixed.

src/NuGetGallery/Services/FileSystemFileStorageService.cs
joelverhagen: Removed the check.

joelverhagen:
🔔 comments adressed.

src/NuGetGallery/Services/FileSystemService.cs
joelverhagen: Did some tests. Performance is worse.

PR #5582, opened 2018-03-06
Fix SemVerLevel propagation in Odata Next links.

ryuyu:
Addresses https://github.com/NuGet/NuGetGallery/issues/5433


src/NuGetGallery/OData/SearchService/SearchAdaptor.cs
```diff
@@ -374,6 +374,14 @@ public static Uri GetNextLink(Uri currentRequestUri, long? totalResultCount, obj
                 queryBuilder.Append("&");
             }
 
+            if (semVerLevelKey != null)
+            {
+                if(semVerLevelKey == SemVerLevelKey.SemVer2)
+                {
+                    queryBuilder.Append("semVerLevel=2.0.0&");
```
joelverhagen: Use const `SemVerLevelKey.SemVerLevel2`.

Issue #5584, opened 2018-03-07
[Organizations] Use a configurable list to determine what users should be asked to transform to an organization with a deprecated password


scottbommarito:
Users are confused whether to transform their account into an organization or to log in with their MSA/AAD account. We should use a configurable list to determine who should be asked to do what.

PR #5588, opened 2018-03-08
[Organizations] Use a configurable list to determine what users should be asked to transform to an organization with a deprecated password

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5584

Added a new field to the existing `LoginDiscontinuationConfiguration` for the accounts that should be asked to transform.

Individual account:
image

Organization account:
image

### DIFF
--- a/src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
+++ b/src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
@@ -11,10 +11,10 @@ namespace NuGetGallery
 {
     public class LoginDiscontinuationConfiguration : ILoginDiscontinuationConfiguration
     {
-        public HashSet<string> DiscontinuedForEmailAddresses { get; }
-        public HashSet<string> DiscontinuedForDomains { get; }
-        public HashSet<string> ExceptionsForEmailAddresses { get; }
-        public HashSet<string> ForceTransformationForEmailAddresses { get; }
+        internal HashSet<string> DiscontinuedForEmailAddresses { get; }
+        internal HashSet<string> DiscontinuedForDomains { get; }
+        internal HashSet<string> ExceptionsForEmailAddresses { get; }
+        internal HashSet<string> ForceTransformationToOrganizationForEmailAddresses { get; }
 
         public LoginDiscontinuationConfiguration()
             : this(Enumerable.Empty<string>(), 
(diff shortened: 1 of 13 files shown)


src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
```diff
@@ -14,25 +14,36 @@ public class LoginDiscontinuationConfiguration : ILoginDiscontinuationConfigurat
         public HashSet<string> DiscontinuedForEmailAddresses { get; }
         public HashSet<string> DiscontinuedForDomains { get; }
         public HashSet<string> ExceptionsForEmailAddresses { get; }
+        public HashSet<string> ForceTransformationForEmailAddresses { get; }
```
shishirx34: `ForceTransformToOrgForEmailAddresses` would be more appropriate.

chenriksson:
@scottbommarito Since this is throw-away work to improve onboarding, maybe we should open an issue to clean this up once we open to everyone.

src/NuGetGallery/ViewModels/GalleryHomeViewModel.cs
```diff
@@ -9,13 +9,16 @@ public class GalleryHomeViewModel
     {
         public bool ShowTransformModal { get; set; }
 
+        public bool ShouldTransform { get; set; }
```
shishirx34: rename this as `TransformToOrg` rather than `ShouldTransform`

src/NuGetGallery/ViewModels/GalleryHomeViewModel.cs
```diff
@@ -9,13 +9,16 @@ public class GalleryHomeViewModel
...
+
         public IList<AuthenticationProviderViewModel> Providers { get; set; }
 
-        public GalleryHomeViewModel() : this(showTransformModal: false) { }
+        public GalleryHomeViewModel() : this(showTransformModal: false, canTransform: false) { }
 
-        public GalleryHomeViewModel(bool showTransformModal)
+        public GalleryHomeViewModel(bool showTransformModal, bool canTransform)
```
shishirx34: rename to `transformToOrg`

src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
cristinamanum: Does it need to be public?

scottbommarito:
@chenriksson issue is filed.

src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
scottbommarito: unit tests use them...made them internal

PR #5589, opened 2018-03-09
[Organizations] Clone existing API keys to org admin on account transform

scottbommarito:
With this PR, when an account is transformed into an organization, its API keys are now cloned to the new admin.

This prevents breaking existing CI scenarios upon organization transformation.

### DIFF
--- a/src/NuGetGallery/Services/UserService.cs
+++ b/src/NuGetGallery/Services/UserService.cs
@@ -10,6 +10,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using NuGetGallery.Auditing;
 using NuGetGallery.Configuration;
+using NuGetGallery.Infrastructure.Authentication;
 using NuGetGallery.Security;
 using Crypto = NuGetGallery.CryptographyService;
 
@@ -46,7 +47,8 @@ namespace NuGetGallery
             IEntitiesContext entitiesContext,
             IContentObjectService contentObjectService,
             ISecurityPolicyService securityPolicyService,
-            IDateTimeProvider dateTimeProvider)
+            IDateTimeProvider dateTimeProvider,
+            ICredentialBuilder credentialBuilder)
             : this()
         {
             Config = config;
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Services/UserService.cs
```diff
@@ -370,10 +372,41 @@ public bool CanTransformUserToOrganization(User accountToTransform, User adminUs
...
+
+            await TransferApiKeysScopedToUser(accountToTransform, adminUser);
             
             return await EntitiesContext.TransformUserToOrganization(accountToTransform, adminUser, token);
         }
 
+        public async Task TransferApiKeysScopedToUser(User userWithKeys, User userToOwnKeys)
+        {
+            var eligibleApiKeys = userWithKeys.Credentials
+                .Where(c => c.IsApiKey() && c.Scopes.All(k => k.Owner == null || k.Owner == userWithKeys)).ToArray();
```
shishirx34: can there be api keys for a user owned by anyone else?

scottbommarito: This is in case the user has API keys scoped to another user or organization.

E.g.
User A is a member of organization B.
A makes API keys scoped to its own packages.
A makes API keys scoped to B's packages.
A leaves B.
A transforms into an organization with user C as admin.
We copy the API keys that A has that are scoped to its own packages but not B's.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -370,10 +372,41 @@ public bool CanTransformUserToOrganization(User accountToTransform, User adminUs
             {
                 return false;
             }
+
+            await TransferApiKeysScopedToUser(accountToTransform, adminUser);
             
             return await EntitiesContext.TransformUserToOrganization(accountToTransform, adminUser, token);
         }
 
+        public async Task TransferApiKeysScopedToUser(User userWithKeys, User userToOwnKeys)
```
chenriksson: You could also do this in `MigrateUserToOrganization.sql`, which woudl allow you to keep it in the same transaction as the transform.

scottbommarito: I was considering that, but given that
1 - I'm less knowledgeable about SQL so this would have been harder for me to write
2 - we're trying to get this change out quickly for the next deployment
3 - if cloning the keys succeeds but transforming the org fails, the worst case scenario is that the admin user has some useless keys
I thought this was fine for now.

PR #5594, opened 2018-03-12
Add destination access condition parameter to CopyFileAsync

joelverhagen:
This is used during revalidate to verify the destination package has not changed.

If a processor has mutated a package, the copy that replaces the destination package should fail if the destination has changed.

Progress on https://github.com/NuGet/Engineering/issues/1190

### DIFF
--- a/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
@@ -110,7 +110,12 @@ namespace NuGetGallery
             }
         }
 
-        public async Task CopyFileAsync(string srcFolderName, string srcFileName, string destFolderName, string destFileName)
+        public async Task<string> CopyFileAsync(
+            string srcFolderName,
+            string srcFileName,
+            string destFolderName,
+            string destFileName,
+            IAccessCondition destAccessCondition)
         {
             if (srcFolderName == null)
             {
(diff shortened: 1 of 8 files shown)


src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -137,36 +142,44 @@ public CloudBlobCoreFileStorageService(ICloudBlobClient client)
...
-            var destAccessCondition = AccessCondition.GenerateIfNotExistsCondition();
             if (await destBlob.ExistsAsync())
             {
                 if (destBlob.CopyState?.Status == CopyStatus.Failed)
                 {
-                    // If the last copy failed, allow this copy to occur.
-                    destAccessCondition = AccessCondition.GenerateIfMatchCondition(destBlob.ETag);
+                    // If the last copy failed, allow this copy to occur no matter what the caller's destination
+                    // condition is. This is because the source blob is preferable over a failed copy.
+                    mappedDestAccessCondition = AccessCondition.GenerateIfMatchCondition(destBlob.ETag);
```
scottbommarito: I think you might want to add to this comment that we're setting the condition to be that the blob hasn't changed since the failed copy. In other words, if we see there's a failed copy, we will allow the copy to occur as long as the failed copy hasn't been replaced by anything else.

joelverhagen: Sure, sounds good.

PR #5597, opened 2018-03-12
Add GenerateEmptyCondition and fix AccessConditionWrapper bug

joelverhagen:
Note to self, add unit tests even on wrapper classes...

### DIFF
--- a/src/NuGetGallery.Core/Services/AccessConditionWrapper.cs
+++ b/src/NuGetGallery.Core/Services/AccessConditionWrapper.cs
@@ -10,12 +10,19 @@ namespace NuGetGallery
         private AccessConditionWrapper(string ifNoneMatchETag, string ifMatchETag)
         {
             IfNoneMatchETag = ifNoneMatchETag;
-            IfMatchETag = IfMatchETag;
+            IfMatchETag = ifMatchETag;
         }
 
         public string IfNoneMatchETag { get; }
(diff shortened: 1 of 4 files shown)


src/NuGetGallery.Core/Services/AccessConditionWrapper.cs
```diff
@@ -10,13 +10,20 @@ public class AccessConditionWrapper : IAccessCondition
         private AccessConditionWrapper(string ifNoneMatchETag, string ifMatchETag)
         {
             IfNoneMatchETag = ifNoneMatchETag;
-            IfMatchETag = IfMatchETag;
+            IfMatchETag = ifMatchETag;
```
joelverhagen: d'oh!

src/NuGetGallery.Core/Services/AccessConditionWrapper.cs
```diff
@@ -10,13 +10,20 @@ public class AccessConditionWrapper : IAccessCondition
...
         }
 
         public string IfNoneMatchETag { get; }
 
         public string IfMatchETag { get; }
 
+        public static IAccessCondition GenerateEmptyCondition()
```
joelverhagen: This is useful because someone you want to copy and blindly replace the destination (e.g. copying the validation set package).

PR #5605, opened 2018-03-14
 Add integration tests for Azure Blob Storage testing CopyFileAsync

joelverhagen:


tests/NuGetGallery.Core.Facts/Services/CloudBlobCoreFileStorageServiceIntegrationTests.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        [BlobStorageFact]
+        public async Task CopyingWithUriWorksWithDifferentStorageAccounts()
+        {
+            await CopyFileWorksAsync(CopyFileWithUriAsync, _prefixA, _targetA, _prefixB, _targetB);
+        }
+
+        [BlobStorageFact]
+        public async Task CopyingWithNamesWorksWithinTheSameStorageAccount()
+        {
```
joelverhagen: Not that it is not possible to copy blobs by names between different accounts.

PR #5607, opened 2018-03-14
Fix migrate apikeys unique index conflict

chenriksson:
Deployment hotfix, which includes:
- Moves the migration of organiation API keys into transform script
- Informs users about key transfer in tooltip and confirmation


skofman1:
Christy, shouldn't the scopes of the API key be changed to the org scope? An I missing the sql that does that?

scottbommarito:
@skofman1 yes, and I don't see that either.

chenriksson:
@skofman1 @scottbommarito I forgot about older keys that didn't already have the owner scope set, but it's updated now.

src/NuGetGallery.Core/Infrastructure/MigrateUserToOrganization.sql
```diff
@@ -22,7 +22,16 @@ BEGIN TRY
 	INSERT INTO [dbo].[Organizations] ([Key]) VALUES (@organizationKey)
 	INSERT INTO [dbo].[Memberships] (OrganizationKey, MemberKey, IsAdmin) VALUES (@organizationKey, @adminKey, 1)
 
-	-- Remove organization credentials
+	-- Reassign organization API keys to the admin user
+	UPDATE [dbo].[Scopes] SET OwnerKey = @organizationKey
+	WHERE CredentialKey IN (
```
scottbommarito: If the account transforming into an organization has any API keys scoped to another user, they'll get reset to be scoped to the organization. This isn't that big of a deal (the situation itself is unlikely) but I assume the fix would be to add a `OwnerKey = null` in the `WHERE`.

skofman1:
Will this work well for full Access API keys? They have no scopes.

scottbommarito:
@skofman1 good point...no it won't.

src/NuGetGallery.Core/Infrastructure/MigrateUserToOrganization.sql
chenriksson: I think we should discuss w/ @anangaur about how to handle keys on leaving an org.

If we want to add the additional check here, it's just:
```
UPDATE [dbo].[Scopes] ...
WHERE (OwnerKey IS NULL OR OwnerKey = @organizationKey)
    AND CredentialKey IN ( ... )
```

chenriksson:
The full access API keys would be transferred to the admin user... which does change the implicit owner scope. We should discuss how to handle these... options:
- Add (broadest) scopes: push|unlist * as org
- Don't transfer, just remove... admin should create scoped keys manually

skofman1:
@chenriksson , we decided to remove them, otherwise we can get weirdness in the DB like API keys that never expire that have scopes. @scottbommarito  is working on a fix.

anangaur:
We shouldn’t remove/delete. Let’s not have majic on some keys that get transferred and others not.
May be make them expire in another 365 days?

skofman1:
@anangaur , full access API keys have no scopes, so they can't be used to access organizations. If we want them to work for organizations, we will need to add scopes, a description, expiration date if it's missing, basically modify the legacy api key to be a scoped api key.
I don't think it's the right way to go. 
We are adding a comment on the transform page about this, IMO it's sufficient.

skofman1:
@anangaur , the screenshot is here: https://github.com/NuGet/NuGetGallery/pull/5611

PR #5612, opened 2018-03-14
Get the package hash from the stream when backing up

joelverhagen:

Since we can now mutate packages via processors, it is possible for the hash in the DB to mismatch the hash of the actual package. Therefore, we should calculate the hash from the backed up stream.

I chose not to change the backup method to use `CopyFileAsync` (i.e. server-side copy) because this would have it's own race conditions and requires a `srcAccessCondition`. We have to download the package anyway to calculate the hash. We can revisit this later if we want to improve the performance.

Also, make the hash algorithm a required parameter to `CryptographyService.GenerateHash`. It's silly for the caller to not care about the hash algorithm, especially since you have to put the hash algorithm name in the database record anyway.

### DIFF
--- a/src/NuGetGallery.Core/Services/CorePackageFileService.cs
+++ b/src/NuGetGallery.Core/Services/CorePackageFileService.cs
@@ -159,10 +159,17 @@ namespace NuGetGallery
                 version = package.NormalizedVersion;
             }
 
+            // Hash the provided stream instead of using the hash on the package. This is to avoid a backup with the
+            // incorrect file name if the hash in the DB does not match the package (a potentially transient issue).
+            var hash = CryptographyService.GenerateHash(
+                packageFile,
+                hashAlgorithmId: CoreConstants.Sha512HashAlgorithmId);
+            packageFile.Position = 0;
+
             var fileName = BuildBackupFileName(
                 package.PackageRegistration.Id,
                 version,
-                package.Hash);
+                hash);
 
             // If the package already exists, don't even bother uploading it. The file name is based off of the hash so
             // we know the upload isn't necessary.
--- a/src/NuGetGallery.Core/Services/CryptographyService.cs
+++ b/src/NuGetGallery.Core/Services/CryptographyService.cs
@@ -12,7 +12,7 @@ namespace NuGetGallery
     {
         public static string GenerateHash(
             Stream input,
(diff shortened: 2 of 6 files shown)


chenriksson:
LGTM... do we also keep the original package in blob storage as backup in case of failure? I'm assuming we'd only access these manually and may need to update the hash in DB if that's the case?

joelverhagen:
> LGTM... do we also keep the original package in blob storage as backup in case of failure?

Yes. The back-up will be used in two flows and in both cases the original is kept.

1. Package delete (existing). This backs up first then deletes the original. If the back-up fails, the delete operation fails.
2. In orchestrator when starting a validation set (new). This backs up first and leaves the original. If the back-up fails, the validation operation fails.

> I'm assuming we'd only access these manually and may need to update the hash in DB if that's the case?

joelverhagen:
> I'm assuming we'd only access these manually and may need to update the hash in DB if that's the case?

This is done at the end of the validation operation in orchestrator at the same time as replacing the package in the packages container.

If there is a mismatch in the delete flow, we don't do a fix-up.

src/NuGetGallery.Core/Services/CorePackageFileService.cs
```diff
@@ -159,10 +159,17 @@ public Task<bool> DoesValidationPackageFileExistAsync(Package package)
                 version = package.NormalizedVersion;
             }
 
+            // Hash the provided stream instead of using the hash on the package. This is to avoid a backup with the
+            // incorrect file name if the hash in the DB does not match the package (a potentially transient issue).
+            var hash = CryptographyService.GenerateHash(
+                packageFile,
+                hashAlgorithmId: CoreConstants.Sha512HashAlgorithmId);
+            packageFile.Position = 0;
```
loic-sharma: >packageFile.Position = 0; [](start = 12, length = 25)

Shouldn't SaveFileAsync do this?

joelverhagen: Out of scope.

joelverhagen: Sorry, to clarify, it does not do this today as the caller is expected to put the stream in the proper state. Since `GenerateHash` moves the cursor to the end, this method (the caller of `SaveFileAsync`) is responsible for this.

Issue #5606, opened 2018-03-14
LocalUser authentication permanently disabled


chenriksson:
Despite being enabled in web.config, I can't run Gallery locally w/ password auth enabled.

```
<add key="Auth.LocalUser.Enabled" value="true" />
```

In SignIn.cshtml, it looks like "Sign in using NuGet.org account" is enclosed in the Model.Providers block, which is only rendered if an AAD or MSA authenticator is enabled.

It would also be nice if we could set up a shared app instance for AADv2 that uses returnUrl of localhost, which we document in our OneNote for everyone's use in local debugging.

Issue #5040, opened 2017-11-20
Gallery instances without AAD/MSA integration require additional click


joelverhagen:
## Step 1:

![image](https://user-images.githubusercontent.com/94054/33032887-6f15c040-cdd7-11e7-85ef-a12ea9d07290.png)

## Step 2:

![image](https://user-images.githubusercontent.com/94054/33032893-78f7830a-cdd7-11e7-9038-5c02f98fde45.png)

Step 1 is unnecessary.

PR #5613, opened 2018-03-14
Fix login page for local debugging.

shishirx34:
Fixes:
- https://github.com/NuGet/NuGetGallery/issues/5606 - Show password login for local login.
- https://github.com/NuGet/NuGetGallery/issues/5040 - Simplify password login when no external providers present.

Now if there is no external provider configured, we will see password login on the first sign in page itself. 

Note: use `?w=1`(https://github.com/NuGet/NuGetGallery/pull/5613/files?w=1) to ignore whitespace changes

### DIFF
--- a/src/NuGetGallery/Views/Authentication/SignIn.cshtml
+++ b/src/NuGetGallery/Views/Authentication/SignIn.cshtml
@@ -25,28 +25,31 @@
             <h1>Sign in</h1>
         </div>
     </div>
-
-    @foreach (var provider in Model.Providers)
+    @if (Model.Providers.Any())
     {
-        <div class="row">
-            <div class="@ViewHelpers.GetColumnClasses(ViewBag) text-center">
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Authentication/SignIn.cshtml
```diff
@@ -25,28 +25,31 @@
             <h1>Sign in</h1>
         </div>
     </div>
-
-    @foreach (var provider in Model.Providers)
+    @if (Model.Providers.Any())
```
chenriksson: Is the NuGet login still in this block? Doesn't that still prevent password logins if no AAD/MSA providers are enabled?

I think Model.Providers only includes Authenticators w/ custom UI, right?

shishirx34: Nope, we have the entire password login signin experience on this page if there is no external provider. No need for the link to second page.

PR #5615, opened 2018-03-14
Hotfix: additional messaging fixes for Organizations

scottbommarito:


src/NuGetGallery/Views/Users/Transform.cshtml
```diff
@@ -62,6 +58,17 @@
...
+                <div class="row">
+                    <div class="col-md-12">
+                        <p>
+                            Administrators have full control over their organizations, including the ability to add and remove other members.
+                        </p>
+                        <p>
+                            Existing API keys owned by this account will be transferred to the administrator on transformation. Full access legacy API keys will not be transferred.
+                        </p>
+                    </div>
+                </div>
```
shishirx34: can you use the `transformNote` and `tranformNote2` here instead of re-writing this?

PR #5620, opened 2018-03-15
Code refactoring

cristinamanum:


src/NuGetGallery.Core/Services/CorePackageService.cs
```diff
@@ -222,6 +222,12 @@ public virtual Package FindPackageByIdAndVersionStrict(string id, string version
             return package;
         }
 
+        public PackageStatus GetStatus(Package package)
+        {
+            return package.PackageStatusKey;
+        }
+
+
```
shishirx34: nit: new line.

what is the need to add new method in refactoring? don't seem to be used. I would add it with the changes that consume it.

shishirx34:
can you provide more details on the context of the need for this refactoring?

cristinamanum:
The change is not going to be persisted. Close the pull request.

PR #5622, opened 2018-03-15
Don't delete Web UI test results and only delete minimal functional test results

joelverhagen:
Today, since we delete the Web UI test results, only the P2 Web UI results are available.

We can depend on the CI build configurations "Clean all build directories" setting to do the clean-up.

### DIFF
--- a/tests/Scripts/RunP0Tests.bat
+++ b/tests/Scripts/RunP0Tests.bat
@@ -16,14 +16,11 @@ set nuget="nuget.exe"
 set mstest="C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\mstest.exe"
 
 REM Clean previous test results
-if exist functionaltests.*.xml (
-    del functionaltests.*.xml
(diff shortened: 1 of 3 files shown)


joelverhagen:
> This is an improvement, but why do we need to delete the functional test results? Can't we depend on "Clean all build directories" for that as well?

I left it for local runs.

scottbommarito:
@joelverhagen sounds good then!

chenriksson:
@joelverhagen Just to confirm, leaving TestResults won't cause disk space issues on the build agents? Do agents remove the cloned repo once the run completes?

joelverhagen:
> @joelverhagen Just to confirm, leaving TestResults won't cause disk space issues on the build agents? Do agents remove the cloned repo once the run completes?

AFAIK, the clean happens at the beginning of the build rather than at the end.

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/1253.

Issue #5484, opened 2018-02-15
Hitting disabled external provider route returns 500


joelverhagen:
https://dev.nugettest.org/users/account/authenticate/MicrosoftAccount

This produces the following exception

```
System.InvalidOperationException: The 'MicrosoftAccount' authentication provider is disabled and cannot be used to authenticate

   at NuGetGallery.Authentication.AuthenticationService.Challenge(String providerName, String redirectUrl) in C:\vsts-agent\_work\12\s\src\Gallery\submodules\Gallery\src\NuGetGallery\Authentication\AuthenticationService.cs:line 486
   at NuGetGallery.AuthenticationController.AuthenticateAndLinkExternal(String returnUrl, String provider) in C:\vsts-agent\_work\12\s\src\Gallery\submodules\Gallery\src\NuGetGallery\Controllers\AuthenticationController.cs:line 363
   at lambda_method(Closure , ControllerBase , Object[] )
   at System.Web.Mvc.ReflectedActionDescriptor.Execute(ControllerContext controllerContext, IDictionary`2 parameters)
   at System.Web.Mvc.ControllerActionInvoker.InvokeActionMethod(ControllerContext controllerContext, ActionDescriptor actionDescriptor, IDictionary`2 parameters)
   at System.Web.Mvc.Async.AsyncControllerActionInvoker.<BeginInvokeSynchronousActionMethod>b__39(IAsyncResult asyncResult, ActionInvocation innerInvokeState)
   at System.Web.Mvc.Async.AsyncResultWrapper.WrappedAsyncResult`2.CallEndDelegate(IAsyncResult asyncResult)
   at System.Web.Mvc.Async.AsyncControllerActionInvoker.EndInvokeActionMethod(IAsyncResult asyncResult)
   at System.Web.Mvc.Async.AsyncControllerActionInvoker.AsyncInvocationWithFilters.<InvokeActionMethodFilterAsynchronouslyRecursive>b__3d()
   at System.Web.Mvc.Async.AsyncControllerActionInvoker.AsyncInvocationWithFilters.<>c__DisplayClass46.<InvokeActionMethodFilterAsynchronouslyRecursive>b__3f()
   at System.Web.Mvc.Async.AsyncControllerActionInvoker.EndInvokeActionMethodWithFilters(IAsyncResult asyncResult)
[...]

Issue #5391, opened 2018-02-03
[MSA] Allow transition from AAD to MSA


shishirx34:
This work is currently blocked on organization membership checks.

Issue #5168, opened 2017-12-11
[2FA] Deprecate NuGet.org logins - Phase1


shishirx34:
Add deprecation banner and point to 'accounts' page for linking MSA.

PR #5628, opened 2018-03-15
[MSA] Deprecate password logins(Phase 1) and other fixes

shishirx34:
- https://github.com/NuGet/NuGetGallery/issues/5484 - Fixes issue with non-existent providers returning 500(now 404)
- https://github.com/NuGet/NuGetGallery/issues/5391 - Added support to allow transition of AAD to MSA account.
- https://github.com/NuGet/NuGetGallery/issues/5168 - Start Deprecation of password logins, several UI changes.

Message upon signing in with accounts having external as well as password logins:

image

Message upon signing in with accounts having only password logins:

image

Warning icons:
image

image

Accounts with no external credential:

image

Linking the external account removes all the warning icons and messages.


src/NuGetGallery/Views/Users/_UserAccountChangePassword.cshtml
```diff
@@ -6,7 +6,13 @@
 @ViewHelpers.Section(parent,
          "change-password",
         @<text>Password Login</text>,
-        @<text><em>@(Model.HasPassword ? "(Enabled)" : "(Disabled)")</em></text>,
+        @<text>
+        @if (this.Config.Current.DeprecateNuGetPasswordLogins)
+        {
+            <span class="ms-Icon ms-Icon--Warning warning-icon" title="Password login is deprecated, please disable this login." aria-hidden="true"></span>
```
scottbommarito: A couple thoughts:

1 - Using "login" twice in the same sentence is too much. Also, you should phrase `this login` in a way that makes it clear that we are referring to the current user's password login.

2 - I would prefer us telling the user that password logins are going to stop working eventually rather than that they need to disable the login themselves. The main concern for a user here is that their password is going to stop working regardless of whether or not they disable it.

src/NuGetGallery/Views/Users/_UserAccountLinkExternalCredential.cshtml
```diff
@@ -1,13 +1,18 @@
...
 }
 
 @ViewHelpers.Section(parent,
     "create-external-credential",
     @<text>Login Account</text>,
-    @<text>Microsoft account</text>,
+    @<text>
+        @if (this.Config.Current.DeprecateNuGetPasswordLogins)
+        {
+           <span class="ms-Icon ms-Icon--Warning warning-icon" title="Password login is deprecated, please link Microsoft account to sign-in to @this.Config.Current.Brand." aria-hidden="true"></span>
```
scottbommarito: We use this string a lot, can we put it in `Strings` and then have some sort of `ViewHelpers` to generate it?

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -390,11 +394,24 @@ public virtual ActionResult AuthenticatePost(string returnUrl, string provider)
...
-            var aadCredential = user?.Credentials.GetAzureActiveDirectoryCredential();
-            if (aadCredential != null)
+            var userOrganizationsWithTenantPolicy = user?
+                .Organizations?
+                .Where(o => o
+                    .Organization
+                    .SecurityPolicies
+                    .Any(policy => policy.Name == nameof(RequireOrganizationTenantPolicy)));
+
+            if (userOrganizationsWithTenantPolicy != null && userOrganizationsWithTenantPolicy.Any())
```
scottbommarito: With this change, are we checking whether or not the user is an admin who MUST authenticate with AAD before allowing them to change credentials?

src/NuGetGallery.Core/Authentication/NuGetClaims.cs
```diff
@@ -28,8 +28,18 @@ public static class NuGetClaims
...
+
+        /// <summary>
+        /// The claim url for the claim that stores whether or not the user has an external login.
+        /// </summary>
+        public const string HasExternalLogin = ClaimsDomain + "externallogin";
+
+        /// <summary>
+        /// The default string value for all the boolean claims in <see cref="NuGetClaims"/>.
+        /// </summary>
+        public const string DefaultValue = "true";
```
chenriksson: Maybe `DefaultBooleanClaimValue` is more specific? Also, consider encapsulating this and instead exposing APIs for working with boolean claims.

```
public static class ClaimsExtensions {
        private const string BooleanClaimDefault = "true";

        public static void AddBooleanClaim(List<Claim> claims, string name)
        {
            claims.Add(new Claim(name, BooleanClaimDefault));
        }

        public static bool HasBooleanClaim(ClaimsIdentity identity, string name)
        {
            return identity.GetClaimOrDefault(name)?
                .Equals(BooleanClaimDefault, StringComparison.OrdinalIgnoreCase)
                ?? false;
        }
}
```

src/NuGetGallery/Extensions/PrincipalExtensions.cs
```diff
@@ -81,6 +81,40 @@ public static bool IsAdministrator(this IPrincipal self)
...
+        public static bool HasPasswordLogin(this IPrincipal self)
+        {
+            if (self == null || self.Identity == null)
+            {
+                return false;
+            }
+
+            var identity = self.Identity as ClaimsIdentity;
+            var passwordClaim = identity.GetClaimOrDefault(NuGetClaims.HasPasswordLogin);
+            return !string.IsNullOrEmpty(passwordClaim);
```
chenriksson: `ClaimsExtensions.HasDiscontinuedLogin` checks both claim existence and value, but only existence is checked in `HasPasswordLogin` and `HasExternalLogin`.

Per earlier comment, if you create a single API (extension method) you can handle boolean claims consistently in single place.

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -188,3 +202,28 @@
         </div>
     </div>
 }
+
+@if (this.Config.Current.DeprecateNuGetPasswordLogins && TempData.ContainsKey("ShowPasswordDeprecationWarning"))
```
chenriksson: constant?

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -188,3 +202,28 @@
...
+{
+    if ((Convert.ToBoolean((string)TempData["ShowPasswordDeprecationWarning"]))
+        && Request.IsAuthenticated
+        && User.HasPasswordLogin())
```
chenriksson: Can the two `if` statements be combined? Or maybe you could add calculated property `ShowPasswordDeprecation` on the model that evaluates these conditions?

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: No need, if the admin changes their credential to AAD they will not get admin access, and will need to convert to AAD to access any admin related things.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -410,13 +427,21 @@ public virtual ActionResult AuthenticateExternal(string returnUrl)
         [NonAction]
         public ActionResult AuthenticateAndLinkExternal(string returnUrl, string provider)
         {
+            TempData["ShowPasswordDeprecationWarning"] = true;
```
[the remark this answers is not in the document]
shishirx34: This is the auth mechanism, the check for if the user has a password is done via `PrincipalExtentions` to keep it simple, this simply notifies the view that the user logged in and we need to show the banner if the user has the password(only once).

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -188,3 +202,28 @@
...
+    {
+    <div class="alert-transient @(ViewBag.HasJumbotron == true ? "alert-transient-jumbotron" : string.Empty)">
+        <div class="container">
+            <div class="row">
+                <div class="@ViewHelpers.GetColumnClasses(ViewBag)" role="alert" aria-live="assertive">
+                    @if (User.HasExternalLogin()) 
+                    {
+                        @ViewHelpers.AlertWarning(@<text>NuGet.org password login is deprecated. Go to <a href="@Url.AccountSettings()">Account Settings</a> to disable the password login and use Microsoft account to sign-in to @(this.Config.Current.Brand).</text>)
```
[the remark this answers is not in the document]
shishirx34: wish I paid attention in grammar class 😆, done 👍

src/NuGetGallery/Views/Users/_UserAccountChangePassword.cshtml
shishirx34: Suggestion for 1, I couldn't think of anything better and precise? For 2 I think deprecation means the same. Note, this is a tooltip on the warning icon.

Issue #5501, opened 2018-02-16
Organizations: Allow collaborators to leave an organization


chenriksson:
Currently we only allow org admins to manage members. We should allow collaborators to delete their own membership / leave an organization.

/cc @anangaur @scottbommarito @shishirx34

anangaur:
Wasn't aware of this :) Should have tested this out in the bugbash. The desired flow is this:

1. Admins can add new members - newer admins or collaborators
2. For Admins, there would be a confirmation before the Admin gets added.
3. The newly added members (Admins/Collaborators) would get a mail notifying the addition to an organization. 
4. [P2] This mail should contain a link to "leave the organization" 
5. Members (Admins/Collaborators) can leave any organization they do not want to be part of. 
6. The option to leave an organization, edit/view should be there on the list organizations page. I get the icons.

PR #5648, opened 2018-03-20
[Organizations] Allow collaborators to leave an organization

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5501

image


src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -145,9 +145,12 @@ public virtual ActionResult ManageOrganization(string accountName)
...
 
-            if (account == null
-                || ActionsRequiringPermissions.ManageAccount.CheckPermissions(GetCurrentUser(), account)
-                    != PermissionsCheckResult.Allowed)
+            var currentUser = GetCurrentUser();
+
+            if (account == null || 
+                (ActionsRequiringPermissions.ManageAccount.CheckPermissions(currentUser, account)
+                    != PermissionsCheckResult.Allowed && 
+                currentUser.Username != memberName))
```
chenriksson: nit: would order the `currentUser` check first, since it's less expensive

src/NuGetGallery/Scripts/gallery/page-manage-organization.js
```diff
@@ -57,8 +57,14 @@
             this.GravatarUrl = member.GravatarUrl;
 
             this.DeleteMember = function () {
-                if (!window.nuget.confirmEvent("Are you sure you want to delete member '" + self.Username + "'?")) {
-                    return;
+                if (self.IsCurrentUser) {
+                    if (!window.nuget.confirmEvent("Are you sure you want to leave this organization? You will no longer be able to manage it or its packages if you do.")) {
```
chenriksson: `no longer be able to manage it` - that's not the case for collaborators, right?

src/NuGetGallery/Views/Organizations/_OrganizationAccountManageMembers.cshtml
```diff
@@ -67,15 +67,21 @@
...
                         {
-                            <td class="text-center align-middle heading-right member-controls col-xs-1">
-                                <span>
-                                    <a class="btn" data-bind="click: DeleteMember, attr: { 'aria-label': 'Delete Member' }">
-                                        <i class="ms-Icon ms-Icon--Cancel" aria-hidden="true"></i>
-                                    </a>
-                                </span>
-                            </td>
+                            // If the user cannot manage the organization, only show the delete member button if they are the member.
+                            <!-- ko if: IsCurrentUser -->
```
chenriksson: Can we also use the `ko if:` binding for the razor check?

```
<!-- ko if: IsCurrentUser || Model.CanManage -->
```

src/NuGetGallery/Controllers/OrganizationsController.cs
shishirx34: wouldn't this allow for the last remaining admin user to leave the organization, essentially leaving the organization orphaned. Is that the intention, or should we prevent it?

scottbommarito: No, this check is in `UserService.DeleteMemberAsync`.

src/NuGetGallery/Scripts/gallery/page-manage-organization.js
scottbommarito: Technically yes--collaborators never had the ability to manage the organization itself--but I think special-casing this based on collaborator/administrator might be a little too much. I can do it though if you think it's better.

src/NuGetGallery/Views/Organizations/_OrganizationAccountManageMembers.cshtml
scottbommarito: probably:
```
<!-- ko if: IsCurrentUser || @(Model.CanManage ? "true" : "false") -->
```

src/NuGetGallery/Controllers/OrganizationsController.cs
scottbommarito: Ordered the `currentUser` check first as requested.

PR #5656, opened 2018-03-21
Add tab for PackageReference

### LINKED ISSUE #5670: Add a tab for PackageReference on the package details page
In addition to the existing "Package Manager", ".NET CLI" and "Paket CLI" tabs, I'd like to see one for PackageReference:

Example

That would make it really easy to copy a PackageReference XML tag for a specific version and paste it directly into your project file.

Questions that needs answers:

1. What should the name of the tab be?
1. Where would the new tab be placed in relation to the others? What's the ordering?

khellang:
Just a suggestion :wink:

This will allow you to easily copy `PackageReference` tags directly from the gallery and paste them into the project file 😄 

Closes #5670

### DIFF
--- a/src/Bootstrap/less/theme/page-display-package.less
+++ b/src/Bootstrap/less/theme/page-display-package.less
@@ -172,6 +172,7 @@
           border-color: @panel-footer-bg;
           border-style: solid;
           border-width: 1px 0 1px 1px;
+          user-select: all;
         }
 
         .copy-button {
--- a/src/Bootstrap/dist/css/bootstrap-theme.css
+++ b/src/Bootstrap/dist/css/bootstrap-theme.css
@@ -996,6 +996,10 @@ img.reserved-indicator-icon {
   border-color: #002440;
   border-style: solid;
   border-width: 1px 0 1px 1px;
+  -webkit-user-select: all;
+  -moz-user-select: all;
+  -ms-user-select: all;
+  user-select: all;
 }
 .page-package-details .install-tabs .tab-content .tab-pane .copy-button {
   display: table-cell;
--- a/src/NuGetGallery/ViewModels/PackageManagerViewModel.cs
+++ b/src/NuGetGallery/ViewModels/PackageManagerViewModel.cs
@@ -8,6 +8,12 @@ namespace NuGetGallery
     /// </summary>
     public class PackageManagerViewModel
     {
+        public PackageManagerViewModel(string name)
+        {
+            Name = name;
+            CopyLabel = string.Format("Copy the {0} command", name);
+        }
+
         /// <summary>
         /// The package manager's name.
         /// </summary>
@@ -37,6 +43,11 @@ namespace NuGetGallery
         /// The level with which the above message will be displayed.
         /// </summary>
         public AlertLevel AlertLevel { get; set; }
+
+        /// <summary>
+        /// The label for the copy button.
+        /// </summary>
+        public string CopyLabel { get; set; }
     }
 
     public enum AlertLevel
--- a/src/NuGetGallery/ViewModels/ThirdPartyPackageManagerViewModel.cs
+++ b/src/NuGetGallery/ViewModels/ThirdPartyPackageManagerViewModel.cs
@@ -15,7 +15,7 @@ namespace NuGetGallery
         /// </summary>
         public string ContactUrl { get; set; }
 
-        public ThirdPartyPackageManagerViewModel(string contactUrl)
+        public ThirdPartyPackageManagerViewModel(string name, string contactUrl) : base(name)
         {
             ContactUrl = contactUrl;
             AlertLevel = AlertLevel.Warning;
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -28,10 +28,9 @@
     {
         packageManagers = new PackageManagerViewModel[]
         {
-            new PackageManagerViewModel()
+            new PackageManagerViewModel(".NET CLI")
             {
                 Id = "dotnet-cli",
-                Name = ".NET CLI",
                 CommandPrefix = "> ",
                 InstallPackageCommand = string.Format("dotnet tool install --global {0} --version {1}", Model.Id, Model.Version),
                 AlertLevel = AlertLevel.Info,
@@ -43,36 +42,33 @@
     {
         packageManagers = new PackageManagerViewModel[]
         {
-            new PackageManagerViewModel()
+            new PackageManagerViewModel("Package Manager")
             {
                 Id = "package-manager",
-                Name = "Package Manager",
                 CommandPrefix = "PM> ",
                 InstallPackageCommand = string.Format("Install-Package {0} -Version {1}", Model.Id, Model.Version)
             },
 
-            new PackageManagerViewModel()
+            new PackageManagerViewModel(".NET CLI")
             {
                 Id = "dotnet-cli",
-                Name = ".NET CLI",
                 CommandPrefix = "> ",
                 InstallPackageCommand = string.Format("dotnet add package {0} --version {1}", Model.Id, Model.Version)
             },
 
-            new PackageManagerViewModel()
+            new PackageManagerViewModel("PackageReference")
             {
                 Id = "package-reference",
-                Name = "PackageReference",
                 InstallPackageCommand = string.Format("<PackageReference Include=\"{0}\" Version=\"{1}\" />", Model.Id, Model.Version),
                 AlertLevel = AlertLevel.Info,
                 AlertMessage = string.Format("For projects that support <a href=\"{0}\">PackageReference</a>, copy this XML node into the project file to reference the package.",
                                              "https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files"),
+                CopyLabel = "Copy the PackageReference XML node",
             },
 
-            new ThirdPartyPackageManagerViewModel("https://fsprojects.github.io/Paket/contact.html")
+            new ThirdPartyPackageManagerViewModel("Paket CLI", "https://fsprojects.github.io/Paket/contact.html")
             {
                 Id = "paket-cli",
-                Name = "Paket CLI",
                 CommandPrefix = "> ",
                 InstallPackageCommand = string.Format("paket add {0} --version {1}", Model.Id, Model.Version),
             }
@@ -136,7 +132,7 @@
             <div class="copy-button">
                 <button id="@packageManager.Id-button" class="btn btn-default btn-warning" type="button"
                         data-toggle="popover" data-placement="bottom" data-content="Copied."
-                        aria-label="Copy the @packageManager.Name command">
+                        aria-label="@packageManager.CopyLabel">
                     <span class="ms-Icon ms-Icon--Copy" aria-hidden="true"></span>
                 </button>
             </div>
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -59,6 +59,16 @@
                 InstallPackageCommand = string.Format("dotnet add package {0} --version {1}", Model.Id, Model.Version)
             },
 
+            new PackageManagerViewModel()
+            {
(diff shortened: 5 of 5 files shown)


joelverhagen:
Could we get a screenshot? Also, could you do a sanity check to see how things look with long IDs or versions? Perhaps out of scope for you but we should understand if some CSS changes should be made before making this public.

For example:
https://www.nuget.org/packages/runtime.opensuse.13.2-x64.runtime.native.System.Security.Cryptography.OpenSsl/4.3.0-preview1-24530-04

https://www.nuget.org/packages/Protogame.Template.PanningCamera/0.0.0-SHAf08bff9a581d8d73d22ec19d279194df4a3c6977

Overall, I love it! 👏 I'll talk to the team.

khellang:
> Could we get a screenshot?

Sure, here you go:

image

image

image

It looks pretty weird with the super long names and versions, but at least it works. You have the same problem with the other tabs:

image

Is "PackageReference" the best name for the tab? I was thinking about "Project File" as well...

khellang:
One alternative is to add

```css
white-space: nowrap;
overflow-x: auto;
```

And get a scrollbar:

image

But that looks pretty horrible as well 😝

khellang:
One potential issue I can see is that the tooling automatically adds `PrivateAssets="All"` for some packages. I'm assuming this is based on the `developmentDependency` flag in the manifest, but I can't see a way to get that metadata from the `Package` entity in the gallery.

rohit21agrawal:
@khellang love the idea of having this on the gallery. Just  a couple of things I would like to see more:

1) The PackageReference tab name doesn't sound great, Project File with a note saying something along the lines `` Copy and paste the xml into your PackageReference based project file within an ItemGroup ``

2) Development Dependencies would have to be : (https://github.com/NuGet/NuGet.Client/pull/1930)
```
<IncludeAssets>compile; build; native; contentfiles; analyzers</IncludeAssets>
```

joelverhagen:
> It looks pretty weird with the super long names and versions, but at least it works. You have the same problem with the other tabs:

Yeah, this is at least parity -- which is the correct scope for your PR 😄.

> And get a scrollbar:
> But that looks pretty horrible as well 😝 

Yeah, I toyed with that when working on that page and thought the line wrap is more usable than the scrollbar.

> One potential issue I can see is that the tooling automatically adds `PrivateAssets="All"` for some packages. I'm assuming this is based on the `developmentDependency` flag in the manifest, but I can't see a way to get that metadata from the `Package` entity in the gallery.

Great point. We don't persist this information today so perhaps this work could be tracked separately (different GitHub issue). I am fine with this gap for now as long as it is tracked on GitHub so other customers can voice how much they care about the gap.

khellang:
> Great point. We don't persist this information today so perhaps this work could be tracked separately (different GitHub issue). I am fine with this gap for now as long as it is tracked on GitHub so other customers can voice how much they care about the gap.

I played around a bit with the code to see if I [could get it to work](https://github.com/khellang/NuGetGallery/compare/patch-1...khellang:developement-dependency-metadata), but because the informatin is persisted at upload-time, it'll only work for packages uploaded going forward. How feasible it is to migrate existing package entries?

skofman1:
@khellang , a great suggestion, thanks for the PR. 
To comply with our process, please create a GitHub issue to track the improvement. @joelverhagen will work with you and getting the UI perfect and merge the PR 😃

joelverhagen:
> I played around a bit with the code to see if I could get it to work, but because the informatin is persisted at upload-time, it'll only work for packages uploaded going forward.

Yeah, when we want to do this, we would turn it on for new packages then write a one-off job to reflow (re-evaluated package metadata) all packages that are development dependencies or do a data fix in the database.

> How feasible it is to migrate existing package entries?

This property is available in our [catalog leaves](https://docs.microsoft.com/en-us/nuget/api/catalog-resource#catalog-leaf), so we could easily scan for all such packages, e.g.
https://api.nuget.org/v3/catalog0/data/2018.03.22.18.58.09/console.host.dataflow.0.0.7.37.json

This is all out of scope for your PR though, so I added https://github.com/NuGet/NuGetGallery/issues/5666 to track this gap.

> @khellang , a great suggestion, thanks for the PR. 
> To comply with our process, please create a GitHub issue to track the improvement. @joelverhagen will work with you and getting the UI perfect and merge the PR 😃 

Yup, if you could take care of that, @khellang, that would help. 

> The PackageReference tab name doesn't sound great, Project File with a note saying something along the lines `` Copy and paste the xml into your PackageReference based project file within an ItemGroup ``

Yes, I agree with this. I will work with @anangaur to get PM sign off on the tab name and what this note should look like. My guess is probably like the "third party package manager" alert but Bootstrap's "info" instead of "warning". And perhaps links to the [PackageReference docs](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files).

khellang:
OK, I've filed #5670. Now we just need a tab name, placement and what the note should say.

joelverhagen:
@khellang, we'll get back to you next week about this additional information. Thanks for your patience.

304NotModified:
That's a long week...

joelverhagen:
@304NotModified, my apologies. I definitely oversold the timing there. I think the latest state of conversation is here: https://github.com/NuGet/NuGetGallery/issues/5670.

xavierdecoster:
Hey @khellang!

Looking into merging your changes, but somehow there's only a few CSS changes.
Where's the code for the new tab? :)

khellang:
Umm... I'm not really sure what's happened 🤷‍♂️ I'll look into it to see if I can find the rest of the code 🤔

xavierdecoster:
Ok, keep me posted :)

xavierdecoster:
@khellang got any update on this?

khellang:
Sorry about the delay @xavierdecoster! Had to go digging through my local reflog to find the commits 😂 Anyway, I think I managed to restore it. Let me know if I need to change anything 😄

khellang:
So after reading through this PR and https://github.com/NuGet/NuGetGallery/issues/5670 (it's been a while 😅), I think there still might be decisions to make here:

 - Is the order of the tab good? I.e. after all tooling-based options, but before 3rd party.
 - Is the name "PackageReference" good? I see @rohit21agrawal was leaning towards "Project File" with an additional message.
 - Regardless of the naming decision above, should the tab have an extra message about how to add a `PackageReference`? Possibly with a link? Like @joelverhagen mentioned. What should the message be? What's the link?

If you folks would just take care of the bike-shedding and come back with the answers, I'll apply the the results 😉

khellang:
Oh, and there's a couple more things to think about:

 - What should happen if the package is a CLI tool? Ref. https://github.com/NuGet/NuGetGallery/pull/5684#issuecomment-376256395
 - Are you thinking this should go in regardless of https://github.com/NuGet/NuGetGallery/issues/5666? I'm thinking yes and that it should cover 80% of cases. Then the last 20% could be covered once https://github.com/NuGet/NuGetGallery/issues/5666 is closed.

skofman1:
@anangaur , can you take a look at the questions above?

anangaur:
> - Is the order of the tab good? I.e. after all tooling-based options, but before 3rd party.

👍 

>  - Is the name "PackageReference" good? I see @rohit21agrawal was leaning towards "Project File" with an additional message.

I too don't like `PackageReference` as that's the mechanics. Isn't a **Manual** way to put the reference in project file? Can we then call it just **Manual**? I am up for more ideas as I do not have a good word that fits perfectly for this scenario. 

> - Regardless of the naming decision above, should the tab have an extra message about how to add a `PackageReference`? Possibly with a link? Like @joelverhagen mentioned. What should the message be? What's the link?

Something like this with links to `PackageReference` and end it with **Learn more.**?
image

>  - What should happen if the package is a CLI tool? Ref. https://github.com/NuGet/NuGetGallery/pull/5684#issuecomment-376256395

I guess that should already be covered as we have a different text for .NET tools. 

 - Are you thinking this should go in regardless of https://github.com/NuGet/NuGetGallery/issues/5666? I'm thinking yes and that it should cover 80% of cases. Then the last 20% could be covered once https://github.com/NuGet/NuGetGallery/issues/5666 is closed.

Good case to cover but can be its own PR 😄.

khellang:
> Something like this with links to `PackageReference` and end it with **Learn more.**?

I'm assuming that **Lear more.** link would be an `aka.ms` link? Who can get one of those?

xavierdecoster:
> I too don't like `PackageReference` as that's the mechanics. Isn't a **Manual** way to put the reference in project file? Can we then call it just **Manual**? I am up for more ideas as I do not have a good word that fits perfectly for this scenario.

@anangaur @khellang @joelverhagen @skofman1 just a suggestion: what if the tab name just stated `MSBuild`?

In a way, that's the _tool_ in use here. It might also have the intended side-effect of steering away people from manually modifying files, unless they really know what they're doing 😆 

One can also modify the old `packages.config` file *manually*... <shivers />. Which is XML. Which could be confusing for some attempting to add a `<PackageReference>` xml node into that file... (yes, I've seen things... 😭)

khellang:
> One can also modify the old `packages.config` file _manually_... . Which is XML. Which could be confusing for some attempting to add a `<PackageReference>` xml node into that file... (yes, I've seen things... 😭)

The message says "project file" though? 🤷‍♂️

xavierdecoster:
> The message says "project file" though? 🤷‍♂️

I think I've just proven some people don't read 😄 I'm assuming people do read the wording of the tab before clicking, but who knows...

Either way, no strong opinion, it was just a suggestion.

khellang:
FWIW, I'm leaning towards "Project File". I'll run a quick Twitter poll 😉

NickCraver:
`PackageReference` is the nomenclature around the feature *everywhere* else, including blogs from Microsoft. Please stick with that here.

On the other suggestions:
- Manual could mean any of the formats, CLI, etc.
- Project file isn't accurate, it can and often is in `Directory.Build.props` for example.

Please, keep the naming `PackageReference`. It's easy to google, it's unambiguous, it's what devs call it, and it's literally the name of the element.

khellang:
The Twitter poll seems pretty conclusive:

Screenshot_20190409_181450

I think naming the tab "PackageReference" with a good explanation below (including a link to docs) is probably fine.

skofman1:
> The Twitter poll seems pretty conclusive

So let's got with that!

anangaur:
Lets leave `Learn more.` for now and go with the text. The `PackageReference` word should be linked to: https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files

khellang:
No aka.ms link? 🤔

anangaur:
For `PAckageReference`, we do have a docs link that we typically do not change and even if when the doc pages move, we put the redirections in place. For "Learn more", I do not have a good content page available for it yet. As we make the page available, we could have this added back here.

khellang:
Added info message. If we've decided the position, name, message and links are good, I think I'm done here ✨

khellang:
Hmm. The help text is a bit verbose... I tested on a couple of different screen sizes and it ends up wrapping on pretty common resolutions:

image

Maybe if we (at least) remove the word 'directly', it would fit better:

image

anangaur:
The shortest version looks good to me. 👍

anangaur:
Once we add "Learn more", this will become better :)

src/Bootstrap/less/theme/page-display-package.less
```diff
@@ -172,6 +172,7 @@
           border-color: @panel-footer-bg;
           border-style: solid;
           border-width: 1px 0 1px 1px;
+          user-select: all;
```
xavierdecoster: Could you please also add the vendor-specific `*-user-select: all` styles from `bootstrap-theme.css` to this LESS file?

(We don't manually modify the files in the `src/Bootstrap/dist/` directory; we generate them by running `grunt` in the `src/Bootstrap` directory.)

khellang: I did it the other way around; adding `user-select: all;` to this less file and generated `bootstrap-theme.css` using the grunt script. What you're seeing is what's generated from this file. I don't think vendor prefixes are necessary in the source (less) files. Just more clutter.

xavierdecoster: Perfect :)

khellang:
Thanks for bringing this in ❤️

PR #5664, opened 2018-03-22
Add GetPriviledgedFileUriAsync to create SAS enabled URLs

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/1207.

Currently we only have `GetFileReadUriAsync` which allows downloading but nothing else. I am adding `GetPriviledgedFileUriAsync` which accepts arbitrary permissions. 

 I am adding this method so that `IProcessor` running in the orchestrator processor can use this URL to both read and delete (for clean up) a temporary .nupkg, a la `FileUriPermissions.Read | FileUriPermissions.Delete`.

### DIFF
--- a/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
@@ -300,7 +300,39 @@ namespace NuGetGallery
             await blob.SetPropertiesAsync();
         }
 
+        public async Task<Uri> GetPriviledgedFileUriAsync(
+            string folderName,
+            string fileName,
+            FileUriPermissions permissions,
+            DateTimeOffset endOfAccess)
+        {
+            var blob = await GetBlobForUriAsync(folderName, fileName, endOfAccess);
+
+            return new Uri(
+                blob.Uri,
+                blob.GetSharedAccessSignature(MapFileUriPermissions(permissions), endOfAccess));
+        }
+
         public async Task<Uri> GetFileReadUriAsync(string folderName, string fileName, DateTimeOffset? endOfAccess)
+        {
+            var blob = await GetBlobForUriAsync(folderName, fileName, endOfAccess);
+
+            if (IsPublicContainer(folderName))
+            {
+                return blob.Uri;
+            }
+
+            return new Uri(
+                blob.Uri,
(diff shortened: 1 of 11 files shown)


src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -300,28 +300,50 @@ private static string Log(AccessCondition accessCondition)
...
+            string folderName,
+            string fileName,
+            FileUriPermissions permissions,
+            DateTimeOffset endOfAccess)
+        {
+            var blob = await GetBlobForUriAsync(folderName, fileName, endOfAccess);
+
+            return new Uri(
+                blob.Uri,
+                blob.GetSharedAccessSignature((SharedAccessBlobPermissions)permissions, endOfAccess));
```
agr: Can we please have a method that converts `FileUriPermissions` to `SharedAccessBlobPermissions`?
I'm ok for it to do that ugly type case inside, just so there would be no temptation to copy it elsewhere.

joelverhagen: Ah, good point.

src/NuGetGallery.Core/Services/FileUriPermissions.cs
```diff
@@ -0,0 +1,23 @@
...
+    public enum FileUriPermissions
+    {
+        None = 0,
+        Read = 1,
+        Write = 2,
+        Delete = 4,
+        List = 8,
+        Add = 16,
+        Create = 32
+    }
```
ryuyu: Let's mark the permissions that we don't explicitly need at this point as [Obsolete] so we know that we _probably_ shouldn't be requesting those. If we need them in the future, we can remove it as part of the PR that needs it so it's clear that we are expanding the scope.

joelverhagen: Yeah, great idea.

joelverhagen: Fixed. `Read`, `Write`, `Delete`, and `Create` are "supported.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
joelverhagen: Fixed.

Issue #5550, opened 2018-02-27
[Organizations] Manage membership requests page


scottbommarito:
We should add a page or add a section to an existing page for organization membership requests similarly to the package owner requests section on the manage packages page.

PR #5665, opened 2018-03-22
[Organizations] Manage membership requests page

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5550

This PR integrates the pending membership/transformation requests that a user has into their list of organizations.

image


src/NuGetGallery.Core/Entities/User.cs
```diff
@@ -67,6 +67,11 @@ public User(string username)
         /// </summary>
         public virtual OrganizationMigrationRequest OrganizationMigrationRequest { get; set; }
 
+        /// <summary>
+        /// Requests for this user to become the admin of a <see cref="User"/> account that was transformed into an <see cref="Organization"/> account.
+        /// </summary>
+        public virtual ICollection<OrganizationMigrationRequest> OrganizationMigrationRequests { get; set; }
```
skofman1: >OrganizationMigrationRequests [](start = 65, length = 29)

Initialize with an empty list like we are doing for other collections.

src/NuGetGallery/Migrations/201803092333364_ConnectUsersToTheirOrganizationMigrationRequests.cs
```diff
@@ -0,0 +1,16 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class ConnectUsersToTheirOrganizationMigrationRequests : DbMigration
+    {
+        public override void Up()
```
skofman1: It's an empty migration... why is it needed?

scottbommarito: I assumed it needed to create an index or something--will revert.

skofman1:
>         public string Username { get; }

OrganizationName

---
Refers to: src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs:10 in 65d72bd. [](commit_id = 65d72bd4dc42c6727070c429185325ba12291adf, deletion_comment = False)

skofman1:
>         public string EmailAddress { get; }

Is e-mail address used?

---
Refers to: src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs:11 in 65d72bd. [](commit_id = 65d72bd4dc42c6727070c429185325ba12291adf, deletion_comment = False)

src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs
```diff
@@ -12,15 +12,46 @@ public class ManageOrganizationsItemViewModel
         public bool CurrentUserIsAdmin { get; }
```
skofman1: >CurrentUserIsAdmin [](start = 20, length = 18)

how is it used?

src/NuGetGallery/Views/Users/Organizations.cshtml
```diff
@@ -59,9 +63,24 @@
...
+                                    else if (organization.Pending)
+                                    {
+                                        var accept = organization.Transform ? Url.ConfirmTransformAccount(organization.Username, organization.ConfirmationToken) : Url.AcceptOrganizationMembershipRequest(organization.Username, organization.ConfirmationToken);
+                                        var reject = organization.Transform ? Url.RejectTransformAccount(organization.Username, organization.ConfirmationToken) : Url.RejectOrganizationMembershipRequest(organization.Username, organization.ConfirmationToken);
+
+                                        <a class="btn" href="@accept" title="Accept Organization Membership Request" aria-label="@("Accept Organization Membership Request for " + organization.Username)">
+                                            <i class="ms-Icon ms-Icon--Accept" aria-hidden="true"></i>
+                                        </a>
+                                        <a class="btn" href="@reject" title="Reject Organization Membership Request" aria-label="@("Reject Organization Membership Request for " + organization.Username)">
```
skofman1: should stay on the same page.
We should use Ajax here to have a good experience. Please open am issue to track this improvement

src/NuGetGallery/Views/Users/Organizations.cshtml
```diff
@@ -59,9 +63,24 @@
...
+                                            <i class="ms-Icon ms-Icon--Edit" aria-hidden="true"></i>
+                                        </a>
+                                    }
+                                    else if (organization.Pending)
+                                    {
+                                        var accept = organization.Transform ? Url.ConfirmTransformAccount(organization.Username, organization.ConfirmationToken) : Url.AcceptOrganizationMembershipRequest(organization.Username, organization.ConfirmationToken);
+                                        var reject = organization.Transform ? Url.RejectTransformAccount(organization.Username, organization.ConfirmationToken) : Url.RejectOrganizationMembershipRequest(organization.Username, organization.ConfirmationToken);
+
+                                        <a class="btn" href="@accept" title="Accept Organization Membership Request" aria-label="@("Accept Organization Membership Request for " + organization.Username)">
```
skofman1: stay on the same page for accept as well until we move to Ajax

src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs
scottbommarito: used to show either admin/collab in UI

src/NuGetGallery/Views/Users/Organizations.cshtml
scottbommarito: https://github.com/NuGet/NuGetGallery/issues/5669

src/NuGetGallery/Views/Users/Organizations.cshtml
scottbommarito: https://github.com/NuGet/NuGetGallery/issues/5669

scottbommarito:
@skofman1 `EmailAddress` is used for Gravatar and `CurrentUserIsAdmin` is used to show either "administrator" or "collaborator"

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -814,25 +814,34 @@ public static string AddOrganizationMember(this UrlHelper url, string accountNam
...
 
         public static string RejectOrganizationMembershipRequest(this UrlHelper url, MembershipRequest request, bool relativeUrl = true)
         {
-            return url.HandleOrganizationMembershipRequest("RejectMemberRequest", request, relativeUrl);
+            return url.RejectOrganizationMembershipRequest(request.Organization.Username, request.ConfirmationToken, relativeUrl);
         }
 
-        private static string HandleOrganizationMembershipRequest(
-            this UrlHelper url, string routeName, MembershipRequest request, bool relativeUrl = true)
+        public static string AcceptOrganizationMembershipRequest(this UrlHelper url, string organizationUsername, string confirmationToken, bool relativeUrl = true)
```
chenriksson: Do you need both overloads of Accept/Reject?

src/NuGetGallery/ViewModels/ManageOrganizationsViewModel.cs
```diff
@@ -12,7 +12,11 @@ public class ManageOrganizationsViewModel
 
         public ManageOrganizationsViewModel(User currentUser, IPackageService packageService)
         {
-            Organizations = currentUser.Organizations.Select(m => new ManageOrganizationsItemViewModel(m, packageService));
+            var organizations = currentUser.Organizations.Select(m => new ManageOrganizationsItemViewModel(m, packageService));
+            var pendingMemberships = currentUser.OrganizationRequests.Select(m => new ManageOrganizationsItemViewModel(m, packageService));
+            var pendingTransformations = currentUser.OrganizationMigrationRequests.Select(m => new ManageOrganizationsItemViewModel(m, packageService));
```
chenriksson: This is requests to transform the user into an organization?

Why include it? Isn't there only 1 possible migration request, and doesn't that mean the user isn't a member of any other organizations?

src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs
```diff
@@ -7,20 +7,51 @@ namespace NuGetGallery
...
+        {
+            Pending = true;
+            ConfirmationToken = request.ConfirmationToken;
+        }
+
+        public ManageOrganizationsItemViewModel(OrganizationMigrationRequest request, IPackageService packageService)
+            : this(request.NewOrganization, true, packageService)
+        {
+            Pending = true;
+            Transform = true;
```
chenriksson: What do Pending and Transform mean?

chenriksson: Pending means current user is pending membership into this org?
Transform means current user has migration (to organization) request?

Could these be more descriptive?

src/NuGetGallery/Views/Users/Organizations.cshtml
```diff
@@ -59,9 +63,24 @@
                                     @organization.PackagesCount
                                 </td>
                                 <td class="text-right align-middle package-controls">
-                                    <a class="btn" href="@Url.ManageMyOrganization(organization.Username)" title="Edit Organization" aria-label="@("Edit Organization " + organization.Username)">
-                                        <i class="ms-Icon ms-Icon--Edit" aria-hidden="true"></i>
-                                    </a>
+                                    @if (!organization.Pending && !organization.Transform)
```
chenriksson: Per comments above, if `Transform` indicates current user is being transformed into an organization... this shouldn't be possible, right?

src/NuGetGallery/ViewModels/ManageOrganizationsViewModel.cs
```diff
@@ -12,7 +12,11 @@ public class ManageOrganizationsViewModel
...
+
+            Organizations = organizations.Concat(pendingMemberships).Concat(pendingTransformations);
```
chenriksson: If the model code is getting more complex, I suggest extracting it into the Controller and adding unit tests.

src/NuGetGallery/ViewModels/ManageOrganizationsViewModel.cs
scottbommarito: no, `OrganizationMigrationRequests` is the list of requests with this user as admin

src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs
scottbommarito: `Pending` = the user has a request to join this organization and is not actually in the organization
`Transform` = this pending request is for a user to become the admin of this organization when it is transformed

not sure how else to name these without being too descriptive. I can add some XML docs if that would help.

src/NuGetGallery/Views/Users/Organizations.cshtml
scottbommarito: no, see comment above

src/NuGetGallery/ViewModels/ManageOrganizationsViewModel.cs
scottbommarito: I don't think it's very complex--it's just extracting some fields and concat-ing a set of models.

src/NuGetGallery/UrlExtensions.cs
scottbommarito: They are both used, one in the controller using the entities themselves and one in the view using the model.

scottbommarito:
@skofman1 @chenriksson I've handled all feedback.

src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs
```diff
@@ -7,20 +7,51 @@ namespace NuGetGallery
...
+            : this(organization as User, isAdmin, packageService)
         {
-            var organization = membership.Organization;
-            Username = organization.Username;
+            OrganizationName = organization.Username;
             EmailAddress = organization.EmailAddress ?? organization.UnconfirmedEmailAddress;
-            CurrentUserIsAdmin = membership.IsAdmin;
+            CurrentUserIsAdmin = isAdmin;
             MemberCount = organization.Members.Count();
-            PackagesCount = packageService.FindPackageRegistrationsByOwner(membership.Organization).Count();
```
chenriksson: We don't use `PackagesCount` for organizations?

src/NuGetGallery/ViewModels/ManageOrganizationsViewModel.cs
chenriksson: My point is that we don't always add unit tests for view models... maybe we should?

src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs
```diff
@@ -7,20 +7,51 @@ namespace NuGetGallery
...
+            IsPendingRequest = true;
+            ConfirmationToken = request.ConfirmationToken;
+        }
+
+        public ManageOrganizationsItemViewModel(OrganizationMigrationRequest request, IPackageService packageService)
+            : this(request.NewOrganization, true, packageService)
+        {
+            IsPendingRequest = true;
+            IsPendingTransformRequest = true;
+            ConfirmationToken = request.ConfirmationToken;
```
chenriksson: Maybe it's useful to add comments for these constructors... the variations and differences between them is a bit confusing.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -162,6 +162,12 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
                 .HasOptional(u => u.OrganizationMigrationRequest)
                 .WithRequired(m => m.NewOrganization);
 
+            modelBuilder.Entity<User>()
+                .HasMany(u => u.OrganizationMigrationRequests)
+                .WithRequired(m => m.AdminUser)
+                .HasForeignKey(m => m.AdminUserKey)
+                .WillCascadeOnDelete(true); // Migration request will be deleted with the Admin account.
```
chenriksson: No schema changes required for this, right?

src/NuGetGallery/UrlExtensions.cs
```diff
@@ -1211,6 +1220,21 @@ public static string RejectTransformAccount(this UrlHelper url, User accountToTr
...
+
+        public static string ConfirmTransformAccount(this UrlHelper url, string accountToTransformUsername, string confirmationToken, bool relativeUrl = true)
+        {
+            return url.HandleTransformAccount(RouteName.TransformToOrganizationConfirmation, accountToTransformUsername, confirmationToken, relativeUrl);
+        }
+
+        public static string RejectTransformAccount(this UrlHelper url, string accountToTransformUsername, string confirmationToken, bool relativeUrl = true)
+        {
+            return url.HandleTransformAccount("RejectTransform", accountToTransformUsername, confirmationToken, relativeUrl);
+        }
```
chenriksson: Add tests in UrlExtensionsFacts?

src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs
chenriksson: Alternatively, maybe subclasses for variants?

src/NuGetGallery/ViewModels/ManageOrganizationsViewModel.cs
scottbommarito: I think we should if the behavior is complex...but I don't think it is here.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
scottbommarito: yup--had an empty migration in this PR earlier and didn't realize it until @skofman1 noticed it haha

src/NuGetGallery/UrlExtensions.cs
scottbommarito: We don't have tests for every new endpoint in `UrlExtensionsFacts`, just for `GetActionLink` itself (which is called by these endpoint functions).

src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs
scottbommarito: This is already done by https://github.com/NuGet/NuGetGallery/pull/5665/files#diff-ab7f759b44970a4b0b7e813110774a34R24 so I removed it from here.

src/NuGetGallery/ViewModels/ManageOrganizationsItemViewModel.cs
scottbommarito: I think more comments would be helpful but I don't think subclasses would.

PR #5668, opened 2018-03-22
[Organizations] Restrict organization membership to AAD tenant only if it matches a whitelist

scottbommarito:
https://github.com/NuGet/Engineering/issues/1270

1) Allows organizations to be created by someone who is not in an AAD tenant.
2) Adds configuration list with pairs of AAD tenants and email domains.
3) If a user is creating an organization, only restrict its membership to an AAD tenant if the email address of the org is on the whitelist and the user has a credential with the corresponding AAD tenant on the list.

UI if organization is restricted on AAD tenant:

image


src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -272,6 +273,8 @@ protected override void UpdateAccountViewModel(Organization account, Organizatio
             model.Members =
                 account.Members.Select(m => new OrganizationMemberViewModel(m))
                 .Concat(account.MemberRequests.Select(m => new OrganizationMemberViewModel(m)));
+
+            model.RequiresTenant = account.SecurityPolicies.Any(sp => sp.Name == nameof(RequireOrganizationTenantPolicy));
```
joelverhagen: Data should only be based off of a type name if there is a clear unit test asserting that the type name never changes. Alternatively (and preferably) you would create a dedicate constant for the policy name.

src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
```diff
@@ -74,12 +79,51 @@ public bool ShouldUserTransformIntoOrganization(User user)
...
+        bool IsTenantIdPolicySupportedForOrganization(string emailAddress, string tenantId);
+    }
+
+    public class OrganizationTenantPair
+    {
+        public string EmailDomain { get; }
+        public string TenantId { get; }
+
+        [JsonConstructor]
+        public OrganizationTenantPair(string emailDomain, string tenantId)
```
joelverhagen: Check parameters or make the comparer below resilient to null.

scottbommarito: done

src/NuGetGallery/Services/UserService.cs
```diff
@@ -525,42 +513,24 @@ public bool CanTransformUserToOrganization(User accountToTransform, User adminUs
...
 
-        private async Task<bool> SubscribeOrganizationToTenantPolicy(User organization, User adminUser, bool commitChanges = true)
+        private async Task SubscribeOrganizationToTenantPolicyIfTenantIdIsSupported(User organization, User adminUser, bool commitChanges = true)
         {
-            var tenantId = GetAzureActiveDirectoryCredentialTenant(adminUser);
-            if (string.IsNullOrWhiteSpace(tenantId))
+            var tenantId = adminUser.Credentials.GetAzureActiveDirectoryCredential()?.TenantId;
+            if (string.IsNullOrWhiteSpace(tenantId) || 
+                !ContentObjectService.LoginDiscontinuationConfiguration.IsTenantIdPolicySupportedForOrganization(organization.EmailAddress ?? organization.UnconfirmedEmailAddress, tenantId))
             {
```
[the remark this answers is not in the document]
scottbommarito: I asked @anangaur, he said only if both match.

Issue #5632, opened 2018-03-17
[Organizations] Fix Package management UI text width


skofman1:
Happens for the Microsoft account
![image](https://user-images.githubusercontent.com/16807822/37549643-68543f5c-293f-11e8-8ef7-359c59c6e8d3.png)

Issue #5303, opened 2018-01-22
[BugBash] Package list table does not fit its borders


agr:
On the [my packages page](https://dev.nugettest.org/account/Packages):
![image](https://user-images.githubusercontent.com/102933/35238156-8bdba846-ff61-11e7-8b61-8a4eb213e909.png)

Either table width should be compacted, or border widened, so there is no overlap

chenriksson:
Removing `Organizations` epic, as this is more of a version formatting issue. Moving to backlog.

PR #5672, opened 2018-03-22
Fix formatting of Manage Packages page with long data

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5632
https://github.com/NuGet/NuGetGallery/issues/5303

Basically
1) only display first two owners
2) abbreviate owners with names too long
3) abbreviate package ID
4) make package ID not wrap
5) abbreviate version even more

This guarantees it will fit within the bounds of the table.

I have highlighted these changes in the screenshot below:
image


anangaur:
What if the first owner name itself is long? In my view there should be a cap on the length of owner name shown else show each name with ellipsis. For more than 2 owners it should not be ellipsis but remaining count.

E.g.
owner1, owner2 + 1

scottbommarito:
@anangaur as requested:

Owner username too long:
image

Too many owners:
image

anangaur:
Nice!

scottbommarito:
@joelverhagen The fix looks great at `lg` and `xs` but unfortunately bad at the sizes in between (in all browsers). Not sure how to improve it more but it's definitely better than the previous iteration and I don't want to spend too much time on it.

Issue #5559, opened 2018-02-28
[Log-in flow] 500 when canceling access to AAD


skofman1:
repro:
1. open an environment where u never logged in before with AAD (for me it was INT)
2. click "sign in with Microsoft"
3. when the app asks permissions to your account, click back
4. Result: 500 page

Issue #5169, opened 2017-12-11
[2FA] Add MSA 2FA support


shishirx34:
Use OIDC with the MFA claims and acr tokens

2FA setting on account. Warn if not set that leads to the setting to be enabled (MSA only)

PR #5681, opened 2018-03-24
[MSA] Add 2FA support for Microsoft logins

shishirx34:
This PR adds the support to challenge the AADv2 provider for enforcing 2FA for accounts that have enabled multi-factor authentication. 

Addresses:
- #5559 - Fix 500 when login with AAD
- #5169 - Add support for 2fA for MSA.

I will follow up the same PR with tests, and a separate PR for enabling 2FA from account management page(UI changes).


src/NuGetGallery/Authentication/AuthenticateExternalLoginResult.cs
```diff
@@ -12,5 +14,19 @@ public class AuthenticateExternalLoginResult
         public ClaimsIdentity ExternalIdentity { get; set; }
         public Authenticator Authenticator { get; set; }
         public Credential Credential { get; set; }
+        public ExternalLoginSessionDetails LoginDetails { get; set; }
+    }
+
+    public class ExternalLoginSessionDetails
+    {
+        public string UsedEmail { get; }
```
scottbommarito: nit: `UsedEmail` sounds like `WasEmailUsed` to me...I would renamed this to `EmailUsed`

src/NuGetGallery/Authentication/Providers/AuthenticationPolicy.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace NuGetGallery.Authentication.Providers
+{
+    public class AuthenticationPolicy
+    {
+        public string Email { get; set; }
+
+        public bool EnforceMfa { get; set; }
```
scottbommarito: nit: `EnforceMultiFactorAuthentication`

MFA isn't super clear and you already type the full phrase above

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -22,6 +24,7 @@ public static class V2Claims
             public const string Email = "preferred_username";
             public const string Name = "name";
             public const string Issuer = "iss";
+            public const string ACR = "http://schemas.microsoft.com/claims/authnclassreference";
```
scottbommarito: `Acr`?

What is this? What does it do? More comments would be nice.

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -35,7 +38,14 @@ public static class AuthenticationType
         public static readonly string V2CommonTenant = "common";
         public static readonly string Authority = "https://login.microsoftonline.com/{0}/v2.0";
 
-        private static readonly string ACCESS_DENIED = "access_denied";
+        private static HashSet<string> ERROR_MESSAGE_LIST = new HashSet<string> { "access_denied", "consent_required" };
+
+        public static class ACR_VALUES
```
scottbommarito: `AcrValues`?

src/NuGetGallery/Authentication/Providers/AuthenticationPolicy.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        private static string ENFORCE_MFA_TOKEN = "enforce_mfa";
```
scottbommarito: Is this how `private static` variables are supposed to be named? I thought it was `_enforceMfaToken` or something like that. Same with some variables below as well.

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -142,33 +153,62 @@ public override IdentityInformation GetIdentityInformation(ClaimsIdentity claims
...
+                var authenticationProperties = GetAuthenticationPropertiesFromProtocolMessage(notification.ProtocolMessage, notification.Options);
 
                 notification.HandleResponse();
                 notification.Response.Redirect(authenticationProperties.RedirectUri);
             }
 
             return Task.FromResult(0);
         }
+
+        // Before redirecting for authentication to the provider, append the properties for Multi-Factor Authentication.
```
scottbommarito: nit: this should probably be inside this method, or an XML doc

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -142,33 +153,62 @@ public override IdentityInformation GetIdentityInformation(ClaimsIdentity claims
...
+        private Task RedirectToIdentityProvider(RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions> notification)
+        {
+            var authenticationProperties = GetAuthenticationPropertiesFromProtocolMessage(notification.ProtocolMessage, notification.Options);
+
+            // AcrValues token control the multi-factor authentication, when supplied with any(which could be default or mfa), the user set policy for 2FA
```
scottbommarito: This honestly feels like a comment that should be spread across `V2Claims.ACR`, `ACR_VALUES`, `ProtocolMessage.AcrValues`, and `ProtocolMessage.LoginHint`. I'm not sure if we control all of these classes or not but a comment saying what a field means on an object is best attached to the field itself.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -749,5 +758,24 @@ private ActionResult AuthenticationView(string viewName, LogOnViewModel existing
...
+
+        private bool ShouldEnforceMultiFactorAuthentication(AuthenticateExternalLoginResult result)
+        {
+            if (result?.Authenticator == null || result.Authentication == null)
+            {
+                return false;
+            }
+
+            // Enforce multi-factor authentication only if:
+            // 1. The authenticator supports multi-factor authentication, otherwise no use.
```
scottbommarito: So, if a user has a credential that doesn't support multi-factor auth but they have an authentication policy that requires multi-factor auth, they can still authenticate without multi-factor auth? This sounds ok with me, but are there protections for a user getting into this state? If they try to add a credential that doesn't support multi-factor auth are they prompted?

src/NuGetGallery/Authentication/Providers/AuthenticationPolicy.cs
shishirx34: earlier had this as `public` changing it to correct convention.

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: For now, we have only single provider for signing in and it supports the 2FA policy. The `2FA` policy is global and will be applicable only to the AADv2 provider. I agree we might need to think about a good user flow for other providers when we add it.

src/NuGetGallery/Authentication/Providers/IdentityInformation.cs
```diff
@@ -15,13 +15,16 @@ public class IdentityInformation
...
+        public bool MultiFactorAuthenticated { get; set; }
+
+        public IdentityInformation(string identifier, string name, string email, string authenticationType, string tenantId = null, bool usedMultiFactorAuth = false)
         {
             Identifier = identifier;
             Name = name;
             Email = email;
             AuthenticationType = authenticationType;
             TenantId = tenantId;
+            MultiFactorAuthenticated = usedMultiFactorAuth;
```
skofman1: UsedMultiFactorAuthentication is a clearer name then MultiFactorAuthenticated

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -142,33 +167,67 @@ public override IdentityInformation GetIdentityInformation(ClaimsIdentity claims
...
+                notification.ProtocolMessage.AcrValues = ACR_VALUES.ANY;
+            }
+
+            return Task.FromResult(0);
+        }
+
+        private AuthenticationProperties GetAuthenticationPropertiesFromProtocolMessage(OpenIdConnectMessage message, OpenIdConnectAuthenticationOptions options)
+        {
+            var authenticationPropertiesEncodedString = message.State.Split('=');
+            return options.StateDataFormat.Unprotect(authenticationPropertiesEncodedString[1]);
```
skofman1: can authenticationPropertiesEncodedString be shorter then 2? If something goes wrong the error message will be unclear

src/NuGetGallery/Infrastructure/ChallengeResult.cs
```diff
@@ -1,29 +1,31 @@
...
 {
     // Borrowed from ASP.Net template in VS 2013 :)
     public class ChallengeResult : HttpUnauthorizedResult
     {
-        public ChallengeResult(string provider, string redirectUri)
+        public ChallengeResult(string provider, string redirectUri, IDictionary<string, string> properties)
         {
             LoginProvider = provider;
             RedirectUri = redirectUri;
+            Properties = properties;
```
skofman1: null check

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
shishirx34: nope, the contract for state is that it will have `=` in it.

Issue #5686, opened 2018-03-26
Login Discontinuation Configuration should be case-insensitive


scottbommarito:
The configuration lists are not case-insensitive, which has led to some odd behavior with certain users. It should be case-insensitive.

PR #5687, opened 2018-03-26
Login Discontinuation Configuration should be case-insensitive

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5686

### DIFF
--- a/src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
+++ b/src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
@@ -118,7 +118,7 @@ namespace NuGetGallery
 
             return
                 string.Equals(x.EmailDomain, y.EmailDomain, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals(x.TenantId, y.TenantId, StringComparison.Ordinal);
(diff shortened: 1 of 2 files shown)


scottbommarito:
@chenriksson not sure--tbh I think transformation could be a useful feature for the general public

src/NuGetGallery/Services/LoginDiscontinuationConfiguration.cs
```diff
@@ -118,7 +118,7 @@ public bool Equals(OrganizationTenantPair x, OrganizationTenantPair y)
 
             return
                 string.Equals(x.EmailDomain, y.EmailDomain, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals(x.TenantId, y.TenantId, StringComparison.OrdinalIgnoreCase);
+                string.Equals(x.TenantId, y.TenantId, StringComparison.Ordinal);
```
joelverhagen: Are tenant IDs case sensitive?

joelverhagen: Let's leave this as case-insensitive to align with DB.

PR #5697, opened 2018-03-27
Delete Account for not confirmed email.

cristinamanum:
The delete account is not allowed for an user with an email that is not confirmed.

With the current change an user can delete its account if it does not have its account confirmed. The self delete will release the account name so it can be reused.


joelverhagen:
We should check this on the server side as well. Someone could spoof a request.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -160,6 +164,54 @@ IAuditingService auditingService
             }
         }
 
+        public async Task<DeleteUserAccountStatus> SelfDeleteGalleryUserAccountAsync(User userToBeDeleted)
```
scottbommarito: As discussed offline, I think it would be best if the delete unconfirmed user flow called into the delete confirmed user flow but with a parameter to "soft-delete" the username. This would guarantee that a deleted user did not have any foreign key relationships to any other tables.

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -2155,6 +2155,47 @@ public class TheRequestAccountDeletionAction : TestContainer
...
                 Assert.Equal<bool>(!successOnSentRequest, tempData);
             }
+
+            /// <summary>
+            /// If the user does not have the email account confirmed it can delete the account.
+            /// </summary>
+            /// <param name="successOnSentRequest"></param>
+            /// <returns></returns>
+            [Fact]
+            public async Task RequestSelfDeleteAccountAsync()
```
scottbommarito: nit: This and the test above could be more descriptively named, e.g.
```
WhenUserIsConfirmedAddsSupportRequest
```
```
WhenUserIsUnconfirmedDeletesAccount
```

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -309,6 +309,18 @@ public virtual ActionResult DeleteRequest()
...
+            if (!user.Confirmed)
+            {
+                // execute the delete request
+                DeleteUserAccountStatus accountDeleteStatus = await _deleteAccountService.SelfDeleteGalleryUserAccountAsync(userToBeDeleted: user);
+                OwinContext.Authentication.SignOut();
+                if (!accountDeleteStatus.Success)
+                {
+                    return Json(Strings.AccountSelfDelete_Fail);
+                }
+                return SafeRedirect(null);
```
chenriksson: `SafeRedirect(null)` seems odd since it ignores the primary argument and only depends on the default. Can we instead add an overload such as `AppController.RedirectHome`?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -160,6 +164,54 @@ IAuditingService auditingService
...
+                        Strings.AccountDelete_AccountAlreadyDeleted,
+                        userToBeDeleted.Username),
+                    AccountName = userToBeDeleted.Username
+                };
+            }
+
+            try
+            {
+                await RemoveUser(userToBeDeleted);
+                await _auditingService.SaveAuditRecordAsync(new DeleteAccountAuditRecord(username: userToBeDeleted.Username,
```
chenriksson: Auditing retention is short enough to allow PII?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -160,6 +164,54 @@ IAuditingService auditingService
             }
         }
 
+        public async Task<DeleteUserAccountStatus> SelfDeleteGalleryUserAccountAsync(User userToBeDeleted)
+        {
+            if (userToBeDeleted == null)
+            {
+                throw new ArgumentNullException(nameof(userToBeDeleted));
+            }
+            if (userToBeDeleted.IsDeleted)
```
chenriksson: Do we even need to check `IsDeleted`? Even for a confirmed account... shouldn't it essentially be a no-op, since relationships were already removed?

src/NuGetGallery/Services/DeleteAccountService.cs
chenriksson: If you reuse the existing flow, can you just reuse the same `DeleteGalleryUserAccountAsync` API?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -309,6 +309,18 @@ public virtual ActionResult DeleteRequest()
...
             }
 
+            if (!user.Confirmed)
+            {
+                // execute the delete request
+                DeleteUserAccountStatus accountDeleteStatus = await _deleteAccountService.SelfDeleteGalleryUserAccountAsync(userToBeDeleted: user);
+                OwinContext.Authentication.SignOut();
+                if (!accountDeleteStatus.Success)
+                {
+                    return Json(Strings.AccountSelfDelete_Fail);
```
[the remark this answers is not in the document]

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
cristinamanum: Updated

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Updated. One flow but if the Use is not confirmed the delete will be hard delete.

cristinamanum: Use the same method.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Yes.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: This code was removed.

src/NuGetGallery/Controllers/UsersController.cs
cristinamanum: Updated to use the Url.Home

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -160,16 +164,29 @@ IAuditingService auditingService
             }
         }
 
-        private async Task DeleteGalleryUserAccountImplAsync(User userToBeDeleted, User admin, string signature, bool unlistOrphanPackages)
+        internal bool HardDeleteUser(User userToBeDeleted)
```
chenriksson: nit: would name with verb `ShouldHardDeleteUser`

src/NuGetGallery/Services/IDeleteAccountService.cs
```diff
@@ -12,11 +12,15 @@ public interface IDeleteAccountService
         /// Deletes an user gallery account.
         /// </summary>
         /// <param name="userToBeDeleted">The user to be deleted.</param>
-        /// <param name="admin">The admin that will execute the delete action.</param>
+        /// <param name="userToExecuteTheDelete">The admin that will execute the delete action.</param>
         /// <param name="signature">The admin signature.</param>
-        /// <param name="unsignOrphanPackages">True if the orphan packages will be unlisted.</param>
+        /// <param name="unlistOrphanPackages:">True if the orphan packages will be unlisted.</param>
```
chenriksson: nit: remove colon?

src/NuGetGallery/Services/IDeleteAccountService.cs
```diff
@@ -12,11 +12,15 @@ public interface IDeleteAccountService
         /// Deletes an user gallery account.
         /// </summary>
         /// <param name="userToBeDeleted">The user to be deleted.</param>
-        /// <param name="admin">The admin that will execute the delete action.</param>
+        /// <param name="userToExecuteTheDelete">The admin that will execute the delete action.</param>
```
chenriksson: why not just `adminUser`?

Issue #4887, opened 2017-10-24
[Organizations] For co-owners, show the user who last pushed/edited each version


scottbommarito:
![image](https://user-images.githubusercontent.com/18014088/31960539-f5a33f42-b8ac-11e7-8f69-89799e26bcb4.png)

If the user that pushed the package was doing so on behalf of an org, it should show the org's name for other co-owners instead of the actual user unless they are in the same org.

anangaur:
We should include the publisher in the emails that we send out - https://github.com/NuGet/NuGetGallery/issues/5807

PR #5698, opened 2018-03-27
[Organizations] For co-owners, show the user who last pushed/edited each version

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/4887

Shows the user who last pushed/edited each version.
Had to abbreviate the version because otherwise the table looked awful.

As organization member:
image

As co-owner:
image


src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs
```diff
@@ -91,5 +97,25 @@ private static bool CanPerformAction(User currentUser, Package package, ActionRe
...
+            var organizationsThatUserPushedByBelongsTo =
+                (package.PackageRegistration?.Owners ?? Enumerable.Empty<User>())
+                    .OfType<Organization>()
+                    .Where(organization => ActionsRequiringPermissions.ViewAccount.CheckPermissions(userPushedBy, organization) == PermissionsCheckResult.Allowed);
+            if (organizationsThatUserPushedByBelongsTo.Any())
+            {
+                return organizationsThatUserPushedByBelongsTo.Any(organization => ActionsRequiringPermissions.ViewAccount.CheckPermissions(currentUser, organization) == PermissionsCheckResult.Allowed) ?
+                    userPushedBy?.Username :
+                    organizationsThatUserPushedByBelongsTo.First().Username;
+            }
```
chenriksson: Looks like this could be expensive - recommend measuring the effect on performance.

scottbommarito: It definitely could get expensive if there are a lot of organizations owning a package and they all have a lot of members. Not sure about any other way to do it though.

scottbommarito: I think in general, however, it should be incredibly rare for us to have packages that fit those conditions. I imagine most organizations will have only a couple users and the packages owned by organizations will only be owned by one or two organizations.

anangaur:
We should implement with the expectation that there can be multiple members in an org and not just two. I agree that with orgs CO ownerships may be less and that’s a fair assumption to make.

scottbommarito:
@anangaur there is no such expectation in this implementation

src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs
chenriksson: The original code already has the push history with `Package.UserKey`.

The new code does the expensive permissions check for every package version, even if the check was the same as a different version. Instead, consider only doing the check 1x for each user in the push history.

chenriksson: Also, I recommend measuring the effect on performance.

1. Define the scenario
- e.g., View package history where 20 versions and 5 different users that pushed
2. Get baseline measurement: what's server latency of this page (browser F12) before these changes?
- Recommend doing 4 iterations, and averaging the last 3 (warm / cached)
3. Get new measurement: what's server latency of this page after these changes?
4. Try the change in query I suggested, and what effect does that have on page latency?

scottbommarito: The new implementation now caches whether or not to show the user who pushed the package for each user. I tested it on DEV on some of the largest packages and the implementation is very fast. I have no concerns about performance anymore.

scottbommarito:
@chenriksson Can you review again with new performance fixes? It performs substantially better now and doesn't appear to have any performance impact on any packages on DEV.

Issue #5689, opened 2018-03-26
[Organizations] Support for deleting organizations


scottbommarito:
Users should be able to delete their organizations, especially if those organizations have no members or packages.

scottbommarito:
UI comments:

- warning should be on one line
- missing colon after "we will"
- accept button has period

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5923

Issue #5025, opened 2017-11-17
Account deletion support for Organization members


chenriksson:
See #5016 

Work remaining includes:
- Delete account views should show what organizations an account will be removed from
- Organizations that would be orphaned should be highlighted, as we do with packages
  - *we discussed organization orphans would be based on last member, not last admin*
- DeleteAccountService needs update to do the following:
  - Unblock delete of organization members
  - Remove deleted account from organizations
  - Unlist packages should include packages orphaned due to orphaned organization

Also consider whether GDPR needs to support deletion of Organization accounts, which also have email.

PR #5701, opened 2018-03-28
[Organizations] Support for deleting organizations

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5689

Administrators of an organization can now delete it if it has no packages that would be orphaned. Unlike user deletion, there is no admin interaction.

Also made some minor modifications to the delete user flow including deleting package ownership requests associated with the user and aligning the package deletion table with the fixes to the manage packages page.

Profile page link:
image

Organization with packages that would be orphaned:
image

Organization without packages that would be orphaned:
image


src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -251,7 +251,7 @@
         {
             <i class="ms-Icon ms-Icon--Group" aria-hidden="true"></i>
         }
-        <a href="@url.User(user.Username)">@user.Username</a>
+        <a href="@url.User(user.Username)">@user.Username.Abbreviate(15)</a>
```
joelverhagen: If we call abbreviate, we should have a `title` attribute with the full value.

joelverhagen:
Feedback:

1. We should either require the admin to remove all other users before deleting the organization (similar to the package ownership transfer rule) or send a notification that the organization has been deleted to all former members.

1. For account deletion @skofman1 / @unniravindranathan / @anangaur agreed to collect data about why people are requesting user deletion before making changes to our process. We should understand why we are collecting data there first and decided whether this "data first" approach is appropriate here.

1. @cristinamanum has a PR for requiring confirmed email before deleting: https://github.com/NuGet/NuGetGallery/pull/5697. How does this fit here?

joelverhagen:
1. We should add a notification to the org email address saying that the org was deleted. Not sure how this should work if the org email is unconfirmed.

scottbommarito:
@joelverhagen Regarding 3, @cristinamanum and I spoke offline. The main reason for the user account confirmation restriction was that we use the support request system there, which could result in an unconfirmed email address being contacted. In this PR, we don't send any emails (although I agree with you that there should be some communication with the members of the organization), so the restriction is not necessary.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -266,6 +274,85 @@ private ActionResult HandleOrganizationMembershipRequestView(HandleOrganizationM
             }
         }
 
+        [HttpGet]
+        [UIAuthorize]
+        public virtual ActionResult Delete(string accountName)
```
chenriksson: Can `Delete` and `ConfirmDelete` be actions on the base AccountsController class, instead of different implementations on UsersController and OrganizationsController... similar to other account actions?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -111,65 +98,70 @@ IAuditingService auditingService
...
+
+            return await RunAccountDeletionTask(
+                () => DeleteGalleryOrganizationAccountImplAsync(organizationToBeDeleted, requestingUser), 
+                organizationToBeDeleted, 
+                requestingUser, 
+                commitAsTransaction);
         }
 
-        private async Task DeleteGalleryUserAccountImplAsync(User userToBeDeleted, User admin, string signature, bool unlistOrphanPackages)
+        private async Task DeleteGalleryOrganizationAccountImplAsync(Organization organizationToBeDeleted, User requestingUser)
```
chenriksson: Maybe we should use a script to revert the `Organization` back to a `User` type first, similar as the transformation flow. This would include: removing memberships, and deleting from the Organizations table.

Once that's done, then it could do the normal delete account flow... sharing the same code.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -111,65 +98,70 @@ IAuditingService auditingService
...
+            var ownedPackages = _packageService.FindPackagesByAnyMatchingOwner(userToBeDeleted, includeUnlisted: true, includeVersions: true).ToList();
+
+            await RemovePackageOwnership(userToBeDeleted, admin, unlistOrphanPackages, ownedPackages);
+            await RemoveReservedNamespaces(userToBeDeleted);
+            await RemoveSecurityPolicies(userToBeDeleted);
+            await RemovePackageOwnershipRequests(userToBeDeleted);
+            await RemoveUserCredentials(userToBeDeleted);
+            await RemoveUserDataInUserTable(userToBeDeleted);
+            await InsertDeleteAccount(userToBeDeleted, admin, signature);
+        }
```
chenriksson: We need to update user account deletion to also remove memberships.

src/NuGetGallery/ViewModels/DeleteOrganizationViewModel.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery
+{
+    public class DeleteOrganizationViewModel
```
chenriksson: If you reuse code by moving actions to the AccountController, then you should reuse the viewModel as well.

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: That is a part of https://github.com/NuGet/NuGetGallery/issues/5025

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: Currently the flow is different--you can only delete empty `Organization`s, but `User`s can be deleted regardless. It's a good suggestion, however.

scottbommarito:
@joelverhagen after discussion with @anangaur 

1 - We have decided to require an organization to only have a single administrator remaining before deletion is allowed

2 - We have deemed data collection is not necessary because we are restricting this flow to orgs with no packages and a single member

3 - As said above, this flow will not send any additional emails, so it does not require an account to be confirmed

scottbommarito:
@chenriksson @cristinamanum I have responded to all feedback and updated the PR with the latest work from DEV, please review.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -277,6 +282,63 @@ private ActionResult HandleOrganizationMembershipRequestView(HandleOrganizationM
...
+        public override async Task<ActionResult> RequestAccountDeletion(string accountName = null)
+        {
+            var account = GetAccount(accountName);
+            var currentUser = GetCurrentUser();
+
+            if (account == null
+                || ActionsRequiringPermissions.ManageAccount.CheckPermissions(GetCurrentUser(), account)
+                    != PermissionsCheckResult.Allowed)
+            {
+                return new HttpNotFoundResult();
```
cristinamanum: Why to return not found when not having permissions?

scottbommarito: It's consistent with what we have in other places for permissions failures. A user who is not able to manage an organization should not know it's an organization, so we return `NotFound` because it's ambiguous whether or not the account is an organization or not.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -266,43 +269,12 @@ private ActionResult TransformToOrganizationFailed(string errorMessage)
...
-        }
-
         [HttpPost]
         [UIAuthorize]
         [ValidateAntiForgeryToken]
-        public virtual async Task<ActionResult> RequestAccountDeletion()
+        public override async Task<ActionResult> RequestAccountDeletion(string accountName = null)
         {
-            var user = GetCurrentUser();
+            var user = GetAccount(accountName);
```
cristinamanum: Is this call needed if the accountName == null?

scottbommarito: Yes, it's a confusing pattern but `UsersController.GetAccount` returns the current user if `accountName` is null or equal to the current user's username.

scottbommarito: Basically, if someone calls `RequestAccountDeletion("notMyUsername")` we want them to know that they are not allowed to edit "notMyUsername"'s account.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -50,21 +51,6 @@ IAuditingService auditingService
...
-        /// 3. Any owned namespaces will be released.
-        /// 4. The user credentials will be cleaned.
-        /// 5. The user data will be cleaned.
-        /// </summary>
-        /// <param name="userToBeDeleted">The user to be deleted.</param>
-        /// <param name="userToExecuteTheDelete">The user that will perform the delete action.</param>
-        /// <param name="signature">The admin signature.</param>
-        /// <param name="unlistOrphanPackages">If the orphaned packages will unlisted.</param>
-        /// <param name="commitAsTransaction">If the data will be persisted as a transaction.</param>
-        /// <returns></returns>
```
cristinamanum: Why removing all the comments?

scottbommarito: The comment for the method belongs on the interface and not the implementation because
1 - having it in both places is redundant and can lead to changing one comment but not the other
2 - you are almost always working with instances of the interface, so when using intellisense, only the comment on the interface will actually appear

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -266,43 +269,12 @@ private ActionResult TransformToOrganizationFailed(string errorMessage)
...
-                return HttpNotFound("User not found.");
-            }
-
-            var listPackageItems = _packageService
-                 .FindPackagesByAnyMatchingOwner(currentUser, includeUnlisted: true)
-                 .Select(p => new ListPackageItemViewModel(p, currentUser))
-                 .ToList();
-
-            bool hasPendingRequest = _supportRequestService.GetIssues().Where((issue) => (issue.UserKey.HasValue && issue.UserKey.Value == currentUser.Key) && 
-                                                                                 string.Equals(issue.IssueTitle, Strings.AccountDelete_SupportRequestTitle) &&
```
cristinamanum: Just to make sure - with the new implementation, if a "Delete support request" is present the user cannot submit a new one?

scottbommarito: Yes, this code was moved to `DeleteUserViewModel`.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -229,18 +216,55 @@ internal bool HardDeleteUser(User userToBeDeleted)
...
+            var requests = _packageOwnershipManagementService.GetPackageOwnershipRequests(newOwner: user).ToList();
+            foreach (var request in requests)
+            {
+                await _packageOwnershipManagementService.DeletePackageOwnershipRequestAsync(request.PackageRegistration, request.NewOwner);
+            }
+        }
+
+        private async Task RemoveMemberships(Organization organization)
+        {
+            foreach (var membership in organization.Members.ToList())
```
cristinamanum: Will the members be notified?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -115,78 +102,78 @@ IAuditingService auditingService
                     AccountName = userToBeDeleted.Username
                 };
             }
+
```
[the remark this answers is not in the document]
scottbommarito: Good catch. This case should still exist because `DeleteGalleryUserAccountAsync` should not be called to delete an `Organization` but the error should not say that deleting an organization is not implemented.

src/NuGetGallery/ViewModels/DeleteAccountViewModel.cs
```diff
@@ -7,22 +7,46 @@
...
         }
 
-        public List<ListPackageItemViewModel> Packages { get; set; }
+        public List<ListPackageItemViewModel> Packages { get; }
 
         public User User { get; set; }
 
-        public string AccountName { get; set; }
-
-        public bool HasPendingRequests { get; set; }
```
cristinamanum: Same comment as above - a user with a pending request should not be able to create a new one.

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: No, but we require that the user deleting the organization is the only member of the organization before calling this method, so this should never delete more than the current user.

src/NuGetGallery/ViewModels/DeleteAccountViewModel.cs
scottbommarito: Yes, this was moved to `DeleteUserViewModel`

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -277,6 +282,63 @@ private ActionResult HandleOrganizationMembershipRequestView(HandleOrganizationM
...
+            }
+
+            var model = GetDeleteOrganizationViewModel(account);
+
+            if (model.HasOrphanPackages)
+            {
+                TempData["ErrorMessage"] = "You cannot delete your organization unless you transfer ownership of all of its packages to another account.";
```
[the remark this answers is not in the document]
scottbommarito: We require this on organizations and not on user accounts because we are required to allow users to delete their accounts due to GDPR but we are not required to allow users to delete their organizations. Therefore, because we want to discourage people from deleting their organizations, we have this restriction.

src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -252,7 +252,7 @@
         {
             <i class="ms-Icon ms-Icon--Group" aria-hidden="true"></i>
         }
-        <a href="@url.User(user.Username)">@user.Username</a>
+        <a href="@url.User(user.Username)" title="@user.Username">@user.Username.Abbreviate(15)</a>
```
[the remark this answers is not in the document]
scottbommarito: My understanding was that the title field is all that's necessary.

src/NuGetGallery/Areas/Admin/ViewModels/DeleteAccountAsAdminViewModel.cs
```diff
@@ -5,20 +5,29 @@
...
+
+        public string AccountName { get; set; }
+
         [Required(ErrorMessage = "Please sign using your name.")]
         [StringLength(1000)]
         [Display(Name = "Signature")]
         public string Signature { get; set; }
 
-        [Display(Name = "Unlist the packages with no other owners.")]
+        [Display(Name = "Unlist any orphaned packages.")]
```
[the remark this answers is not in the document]
scottbommarito: I think we explain what an "orphaned package" is earlier on the page, but I'll check the language and try to make it clearer.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -277,6 +282,63 @@ private ActionResult HandleOrganizationMembershipRequestView(HandleOrganizationM
...
+            }
+
+            if (model.HasAdditionalMembers)
+            {
+                TempData["ErrorMessage"] = "You cannot delete your organization unless you remove all other members.";
+
+                return RedirectToAction(nameof(DeleteRequest));
+            }
+
+            var result = await DeleteAccountService.DeleteGalleryOrganizationAccountAsync(account, currentUser, true);
```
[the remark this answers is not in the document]
scottbommarito: Basically, GDPR requirement as discussed above. We don't want to delete organizations unless there's no reason they shouldn't be deleted.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -257,5 +281,63 @@ internal bool HardDeleteUser(User userToBeDeleted)
             _userRepository.DeleteOnCommit(user);
             await _userRepository.CommitChangesAsync();
         }
+
+        private async Task RemoveOrganization(Organization organization)
+        {
+            _entitiesContext.DeleteOnCommit(organization);
+            await _entitiesContext.SaveChangesAsync();
```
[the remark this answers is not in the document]
scottbommarito: See above, GDPR requirement

src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs
```diff
@@ -81,6 +81,8 @@ public bool HasSingleOwner
             }
         }
 
+        public bool HasSingleOrganizationOwner => Owners.Distinct().Count() < 2;
```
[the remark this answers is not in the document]
scottbommarito: This is used to determine whether or not deleting a user or organization will orphan a package. Basically, because what will "orphan" a package is different depending on whether or not you're deleting a `User` or an `Organization`, we need to introduce this additional field to use in `DeleteOrganizationViewModel`.

For example:
Organization A has user B and user C as members.

Package 1 is owned by A.
If you delete any user that owns it, it is not orphaned.
If you delete any organization that owns it, it IS orphaned.

src/NuGetGallery/Areas/Admin/ViewModels/DeleteAccountAsAdminViewModel.cs
```diff
@@ -5,20 +5,29 @@
 
 namespace NuGetGallery.Areas.Admin.ViewModels
 {
-    public class DeleteUserAccountViewModel : DeleteAccountViewModel
+    public class DeleteAccountAsAdminViewModel
```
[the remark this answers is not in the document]
scottbommarito: The actual inheritance structure is more complicated than that, but I think it's ideal. It's unfortunate that it's so complex but I think any other structure would involve redundant/less maintainable code. If you think there are better names for it then I have no problem renaming them.

`IDeleteAccountViewModel`
contains fields used by `DeleteAccountViewModel` and `DeleteAccountAsAdminViewModel`

`DeleteAccountViewModel` : `IDeleteAccountViewModel`
contains fields that do not vary when deleting a `User` vs deleting an `Organization`
used by `_UserPackagesListForDeletedAccount`, the partial used by both deletion views

`DeleteAccountViewModel<TAccount>` : `DeleteAccountViewModel`
contains a field with a `Account` field of type `TAccount`
exists to make inheritance with `AccountsController` possible

`DeleteUserViewModel` : `DeleteAccountViewModel<User>`
contains fields used when deleting a `User` but not an `Organization`

`DeleteOrganizationViewModel` : `DeleteAccountViewModel<Organization>`
contains fields used when deleting an `Organization` but not a `User`

`DeleteAccountAsAdminViewModel`
model for the form that admins use to delete a package
has a constructor that takes in a `IDeleteAccountViewModel`

src/NuGetGallery/Services/DeleteAccountService.cs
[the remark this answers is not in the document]
chenriksson: Discussed with @cristinamanum - I think we should do the same for both user and organization accounts. I believe it's more than GDPR, but also a security concern with making accounts immediately available to others.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -277,6 +282,63 @@ private ActionResult HandleOrganizationMembershipRequestView(HandleOrganizationM
...
+            {
+                TempData["ErrorMessage"] = "You cannot delete your organization unless you transfer ownership of all of its packages to another account.";
+
+                return RedirectToAction(nameof(DeleteRequest));
+            }
+
+            if (model.HasAdditionalMembers)
+            {
+                TempData["ErrorMessage"] = "You cannot delete your organization unless you remove all other members.";
+
```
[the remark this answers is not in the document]
chenriksson: IMHO, #1 is a better user experience. Members would be notified that the organization was deleted, instead of being notified that they were removed from the organization.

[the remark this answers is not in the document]
scottbommarito: I agree, but it's beyond MVP. I'll file an issue.

scottbommarito:
@chenriksson Deleting organizations now has the same behavior with regards to reserving the name--confirmed organizations are soft-deleted and unconfirmed organizations are hard-deleted

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -229,18 +226,55 @@ internal bool HardDeleteUser(User userToBeDeleted)
             }
         }
 
-        private async Task RemoveOwnership(User user, User admin, bool unlistOrphanPackages, List<Package> packages)
+        private async Task RemovePackageOwnership(User user, User requestingUser, bool unlistOrphanPackages)
         {
-            foreach (var package in packages)
+            foreach (var package in GetPackagesOwnedByUser(user))
             {
                 if (unlistOrphanPackages && _packageService.GetPackageUserAccountOwners(package).Count() <= 1)
```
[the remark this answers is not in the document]
scottbommarito: `GetPackageUserAccountOwners` enumerates all members of organizations that own the package and returns them.

src/NuGetGallery/Services/DeleteAccountService.cs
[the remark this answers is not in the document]
scottbommarito: done

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -229,18 +226,55 @@ internal bool HardDeleteUser(User userToBeDeleted)
...
-                await _packageOwnershipManagementService.RemovePackageOwnerAsync(package.PackageRegistration, admin, user, commitAsTransaction:false);
+                await _packageOwnershipManagementService.RemovePackageOwnerAsync(package.PackageRegistration, requestingUser, user, commitAsTransaction:false);
+            }
+        }
+
+        private async Task RemovePackageOwnership(Organization organization, User requestingUser)
+        {
+            foreach (var package in GetPackagesOwnedByUser(organization))
+            {
+                await _packageOwnershipManagementService.RemovePackageOwnerAsync(package.PackageRegistration, requestingUser, organization, commitAsTransaction: false);
```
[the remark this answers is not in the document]
scottbommarito: combined the methods

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -229,18 +226,55 @@ internal bool HardDeleteUser(User userToBeDeleted)
...
+            {
+                organization.Members.Remove(membership);
+            }
+
+            foreach (var memberRequest in organization.MemberRequests.ToList())
+            {
+                organization.MemberRequests.Remove(memberRequest);
+            }
+
+            await _entitiesContext.SaveChangesAsync();
```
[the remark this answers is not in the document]
scottbommarito: It's called within the transaction scope and so it doesn't matter--this gets committed with the transaction.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -257,5 +291,63 @@ internal bool HardDeleteUser(User userToBeDeleted)
...
+
+        private async Task<DeleteUserAccountStatus> RunAccountDeletionTask(Func<Task> getTask, User userToBeDeleted, User requestingUser, bool commitAsTransaction)
+        {
+            try
+            {
+                // The support requests db and gallery db are different.
+                // TransactionScope can be used for doing transaction actions across db on the same server but not on different servers.
+                // The below code will clean the suppport requests before the gallery data.
+                // The order is important in order to allow the admin the opportunity to execute this step again.
+                await RemoveSupportRequests(userToBeDeleted);
```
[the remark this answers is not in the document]
scottbommarito: `RemoveSupportRequests` is on a completely different database than the other calls. The `RemoveX` calls are all inside the transaction scope.

src/NuGetGallery/Services/DeleteAccountService.cs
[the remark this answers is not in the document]
scottbommarito: combined with the hard delete user logic

src/NuGetGallery/Services/DeleteAccountService.cs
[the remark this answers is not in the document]
chenriksson: OK, looks like same for RemoveUser below. Seems `commitAsTransaction` is just for testing, but wanted to be sure we weren't mocking in different ways.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -277,6 +282,63 @@ private ActionResult HandleOrganizationMembershipRequestView(HandleOrganizationM
...
+                return RedirectToAction(nameof(DeleteRequest));
+            }
+
+            var result = await DeleteAccountService.DeleteGalleryUserAccountAsync(account, currentUser, commitAsTransaction: true);
```
chenriksson: nit: maybe rename to just `DeleteAccountAsync` (org or user)?

scottbommarito: good catch haha

Issue #5691, opened 2018-03-26
[Organizations] Link empty Manage Organizations page to the Transform flow


scottbommarito:
A user who would like to transform their account into an organization cannot do so without knowing the endpoint directly. There should be a visible link on the Manage Organizations page to do so.

scottbommarito:
@anangaur should the Manage Organizations page when a user has organizations have any indication that if a user would like to transform into an organization they need to leave all their organizations?

anangaur:
Not required.

PR #5706, opened 2018-03-28
[Organizations] Link empty Manage Organizations page to the Transform flow

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5691

Also some small UI fixes to the transform page

Manage organizations page:

image

Transform page:

image

### DIFF
--- a/src/NuGetGallery/Views/Users/Organizations.cshtml
+++ b/src/NuGetGallery/Views/Users/Organizations.cshtml
@@ -45,7 +45,7 @@
                                     <img src="@GravatarHelper.Url(organization.EmailAddress, 256)" class="package-icon img-responsive" aria-hidden="true" alt="" />
                                 </td>
                                 <td class="align-middle package-id">
-                                    <a title="View Organization Profile" href="@Url.User(organization.OrganizationName)" >
+                                    <a title="View Organization Profile" href="@Url.User(organization.OrganizationName)">
                                         @organization.OrganizationName
                                     </a>
                                     @if (organization.IsPendingRequest)
@@ -103,9 +103,7 @@
                     </form>
                 </div>
                 <div class="subtitle text-center">
-                    <form action="@Url.TransformAccount()">
-                        <input type="submit" class="btn btn-primary" value="Transform account into organization" />
-                    </form>
+                    <a href="@Url.TransformAccount()">Transform your account into an organization</a>
                 </div>
             }
         </div>
--- a/src/NuGetGallery/Views/Users/Organizations.cshtml
+++ b/src/NuGetGallery/Views/Users/Organizations.cshtml
@@ -102,6 +102,11 @@
                         <input type="submit" class="btn btn-primary" value="Add new organization" />
                     </form>
                 </div>
+                <div class="subtitle text-center">
+                    <form action="@Url.TransformAccount()">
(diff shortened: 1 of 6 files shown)


scottbommarito:
@anangaur let me know what you think would look better on the page

anangaur:
First of all. Not related to this but the breadcrumb looks weird. Too big. The arrow needs to change from the > to another icon.

As for the transform option, we need 3 things:

1. The transform option should not be a button but a linked button:
image

2. There needs to be a confirm dialog as this is an irreversible process:
image

3. There needs to a small fix on the transform page:
image

scottbommarito:
@anangaur what is the change on 3? I don't see anything different

scottbommarito:
@anangaur regarding 2, I'm also concerned that the text you have provided will scare users into thinking that the "transform this account into an organization" button actually performs the transformation. I think the text you're proposing would be better off on the transform form page itself.

anangaur:
Looks like I added wrong picture for 3. Will add the right one soon.

For 2, I am fine with your suggestion. Can you make this a warning banner on the transform page at the top itself?

scottbommarito:
@anangaur I have updated this PR and the screenshots. Any thoughts?

scottbommarito:
@anangaur see updated screenshot

anangaur:
Some fixes required:
1. Left align the breadcrumb
2. Remove the account name
3. Increase the spacing/padding between the warning and the rest of the contents.

The warning seems too big but i do not have any better suggestion as of now. Lets see if people have better inputs in tomorrow' bugbash.

image

scottbommarito:
@anangaur see updated screenshot

src/NuGetGallery/Views/Users/Organizations.cshtml
```diff
@@ -102,6 +102,9 @@
                         <input type="submit" class="btn btn-primary" value="Add new organization" />
                     </form>
                 </div>
+                <div class="subtitle text-center">
+                    <a href="@Url.TransformAccount()">Transform your account into an organization</a>
+                </div>
```
chenriksson: With the new Org flow, we already updated top right menu to always show the Organizations view link?

scottbommarito: Yup!

src/NuGetGallery/Views/Users/TransformToOrganization.cshtml
```diff
@@ -15,9 +15,27 @@
...
+                    </span>
+                    <ul class="transform-alert-list">
+                        <li>
+                            You will no longer be able sign in to @CurrentUser.Username.
+                        </li>
+                        <li>
+                            All permissions to manage @CurrentUser.Username and the packages it owns will be transferred to the administrator you choose here.
+                        </li>
+                        <li>
+                            Existing scoped API keys for this account will be transferred to the administrator. Full access API keys (legacy API keys without expiry) will not be transferred and will be deleted.
```
shishirx34: may be put ... will not be transferred and will be deleted in bold?

scottbommarito: I like that suggestion! Will do.

scottbommarito:
per @anangaur feedback to apply before merging: increase margin between breadcrumb and image in empty page

PR #5708, opened 2018-03-29
Add default regex match timeout (10s)

ryuyu:
Add a domain variable that sets a default time out on RegEx matches. Found [here](https://msdn.microsoft.com/en-us/library/system.text.regularexpressions.regex.matchtimeout%28v=vs.110%29.aspx)

Regex match timeout defaults to Regex.InfiniteMatchTimeout if a timeout is not specified.
Example of nearly unbounded match is shown on the [Best Practices](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices) page.

@blowdart Is this sufficient for protection from Regex match taking unbounded time?

### DIFF
--- a/src/NuGetGallery/App_Start/OwinStartup.cs
+++ b/src/NuGetGallery/App_Start/OwinStartup.cs
@@ -45,6 +45,9 @@ namespace NuGetGallery
             ServicePointManager.SecurityProtocol &= ~SecurityProtocolType.Ssl3;
             ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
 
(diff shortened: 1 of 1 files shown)


blowdart:
Yup. Note when you move to .NET Core that won't work any more. Eventually you'll have to add specific timeouts :)

scottbommarito:
Woops, clicked the update branch button out of habit, hopefully it doesn't upset you @ryuyu lol

skofman1:
Fixes: https://github.com/NuGet/Engineering/issues/1278

Issue #5712, opened 2018-03-29
[Organizations] Unconfirmed orgs should not be able to push new packages or change membership


scottbommarito:
This was skipped initially because the account transform flow required the user account to be confirmed, but when add new organization was implemented, confirmation checking was not implemented.

PR #5713, opened 2018-03-29
[Organizations] Restrict unconfirmed organizations from pushing and changing membership

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5712

### DIFF
--- a/src/NuGetGallery/Authentication/ApiScopeEvaluationResult.cs
+++ b/src/NuGetGallery/Authentication/ApiScopeEvaluationResult.cs
@@ -25,6 +25,8 @@ namespace NuGetGallery.Authentication
         /// </summary>
         public User Owner { get; }
 
+        public bool IsOwnerConfirmed => Owner != null && Owner.Confirmed;
+
         public ApiScopeEvaluationResult(User owner, PermissionsCheckResult permissionsCheckResult, bool scopesAreValid)
         {
             ScopesAreValid = scopesAreValid;
@@ -38,7 +40,7 @@ namespace NuGetGallery.Authentication
         /// </summary>
         public bool IsSuccessful()
         {
-            return ScopesAreValid && PermissionsCheckResult == PermissionsCheckResult.Allowed;
+            return ScopesAreValid && PermissionsCheckResult == PermissionsCheckResult.Allowed && IsOwnerConfirmed;
         }
     }
 }
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -775,8 +775,11 @@ namespace NuGetGallery
                 TelemetryService.TrackPackagePushNamespaceConflictEvent(id, version.ToNormalizedString(), GetCurrentUser(), User.Identity);
                 return new HttpStatusCodeWithBodyResult(HttpStatusCode.Conflict, Strings.UploadPackage_IdNamespaceConflict);
             }
+            
+            var message = result.PermissionsCheckResult == PermissionsCheckResult.Allowed && !result.IsOwnerConfirmed ? 
+                Strings.ApiKeyOwnerUnconfirmed : Strings.ApiKeyNotAuthorized;
 
-            return new HttpStatusCodeWithBodyResult(statusCodeOnFailure, Strings.ApiKeyNotAuthorized);
+            return new HttpStatusCodeWithBodyResult(statusCodeOnFailure, message);
         }
 
         private ApiScopeEvaluationResult EvaluateApiScope(IActionRequiringEntityPermissions<PackageRegistration> action, PackageRegistration packageRegistration, params string[] requestedActions)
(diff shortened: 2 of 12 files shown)


src/NuGetGallery/Authentication/ApiScopeEvaluationResult.cs
```diff
@@ -25,6 +25,8 @@ public class ApiScopeEvaluationResult
         /// </summary>
         public User Owner { get; }
 
+        public bool IsOwnerConfirmed => Owner != null && Owner.Confirmed;
```
chenriksson: If property is invoked multiple times, consider setting value in constructor

src/NuGetGallery/Strings.resx
```diff
@@ -830,4 +830,13 @@ If you wish to update the linked Microsoft account you can do so from the accoun
   <data name="AddMember_Success" xml:space="preserve">
     <value>You are now a member of organization '{0}'.</value>
   </data>
+  <data name="VerifyPackage_OwnerUnconfirmed" xml:space="preserve">
+    <value>User '{0}' is unconfirmed and cannot be the owner of the package.</value>
+  </data>
```
chenriksson: Could we validate `CurrentUser.IsConfirmed` in `PackagesController.UploadPackage` with a more personalized message like "You must confirm your account before you can upload packages.".

If we do that, you know this message happens when the current user's organization is unconfirmed... and could give a more personalized message like "Your organization must be confirmed before any members can upload packages on its behalf."

src/NuGetGallery/Strings.resx
```diff
@@ -830,4 +830,13 @@ If you wish to update the linked Microsoft account you can do so from the accoun
   <data name="AddMember_Success" xml:space="preserve">
     <value>You are now a member of organization '{0}'.</value>
   </data>
+  <data name="VerifyPackage_OwnerUnconfirmed" xml:space="preserve">
+    <value>User '{0}' is unconfirmed and cannot be the owner of the package.</value>
+  </data>
+  <data name="ApiKeyOwnerUnconfirmed" xml:space="preserve">
+    <value>The specified API key is scoped to an owner that is unconfirmed. You must confirm the owner that this API key is scoped to before you can use it.</value>
```
chenriksson: Same feedback as last message... this means either current user or their organization is unconfirmed. Can we personalize the message better and be more specific?

src/NuGetGallery/Strings.resx
scottbommarito: It's not `CurrentUser` that's not confirmed. It's the owner in the form. I chose to make this less personal because there's technically no restrictions in the code for who the owner in the form could be.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -775,8 +775,11 @@ private HttpStatusCodeWithBodyResult GetHttpResultFromFailedApiScopeEvaluationHe
                 TelemetryService.TrackPackagePushNamespaceConflictEvent(id, version.ToNormalizedString(), GetCurrentUser(), User.Identity);
                 return new HttpStatusCodeWithBodyResult(HttpStatusCode.Conflict, Strings.UploadPackage_IdNamespaceConflict);
             }
+            
+            var message = result.PermissionsCheckResult == PermissionsCheckResult.Allowed && !result.IsOwnerConfirmed ? 
+                Strings.ApiKeyOwnerUnconfirmed : Strings.ApiKeyNotAuthorized;
```
shishirx34: nit: move the conditional check `?` to next line

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1457,6 +1457,12 @@ private void SendAddPackageOwnerNotification(PackageRegistration package, User n
                 return Json(400, new[] { message });
             }
 
+            if (!owner.Confirmed)
+            {
+                var message = string.Format(CultureInfo.CurrentCulture, Strings.VerifyPackage_OwnerUnconfirmed, formData.Owner);
+                return Json(400, new[] { message });
```
shishirx34: use `HttpStatusCode.BadRequest`, I know not consistent in this file, but lets start to change it.

shishirx34: bonus if you can change other 400's as well in this file, ignore if too many changes :)

src/NuGetGallery/Authentication/ApiScopeEvaluationResult.cs
scottbommarito: I like this better because
1 - It's a very fast check.
2 - If `Owner` changes after this object is created, `IsOwnerConfirmed` could be wrong.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: There were too many changes but I did it anyway!

Issue #5703, opened 2018-03-28
Infinite loop in RetryingHttpClientWrapper when a search service is down


cristinamanum:
WeightedRandomComparer – it does not implement the total order of two elements  - equality is missing.
This cause the OrderBy to spin infinite loop. 

This is a known bug described at:
https://github.com/dotnet/corefx/issues/20563

cristinamanum:
In order to increase the NuGet search service availability the search service is deployed to two regions US North Central and US south central. The NuGet gallery has implemented a weighted algorithm to switch between the two search services based on the health of the endpoints.

During the outage the USNC went down, this caused the health of USNC to drop. The code path went to
WeightedRandomComparer .Compare to the -1 path (below) 
~~~public int Compare(int x, int y)
            {
                var totalWeight = x + y;
                var randomNumber = _random.Next(0, totalWeight);
                if (randomNumber < x)
                {
                    return 1;
                }
                return -1;
            }
~~~

This caused the call at :RetryingHttpClientWrapper.GetWithRetry

~~~
  var healthyEndpoints = endpointsAsList.OrderByDescending(e =>
            {
                var health = _endpointHealthIndicatorStore.GetHealth(e);
                var order = endpointsAsList.IndexOf(e);

                return (health * health) / ((order + 1) * 80);
            }, HealthComparer).ToList(); 
~~~

To go in infinite loop.

The issue is reproducible outside of the gallery context using the same Comparer and try to order a collection with health values 0 and 62.

cristinamanum:
Could not verify in prod because will take the service down.

PR #5715, opened 2018-03-29
Fix for the infinite loop due to bad comparer

cristinamanum:
Fix for the infinite loop due to bad comparer. 
With this change add the equality in the comparer and add a thread safe random.

### DIFF
--- a/src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper.cs
+++ b/src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper.cs
@@ -215,10 +215,6 @@ namespace NuGet.Services.Search.Client
         public class WeightedRandomComparer
             : IComparer<int>
         {
-            public WeightedRandomComparer()
-            {
-            }
-
             public int Compare(int x, int y)
             {
                 var totalWeight = x + y;
--- a/src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper.cs
+++ b/src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper.cs
@@ -212,7 +212,7 @@ namespace NuGet.Services.Search.Client
             return false;
         }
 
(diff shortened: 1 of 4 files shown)


src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper.cs
```diff
@@ -212,27 +211,31 @@ private static bool ShouldTryOther(HttpResponseMessage response)
...
-        
-        class WeightedRandomComparer
+
+        public class WeightedRandomComparer
             : IComparer<int>
         {
-            private readonly Random _random;
-
-            public WeightedRandomComparer(Random random)
+            public WeightedRandomComparer()
```
skofman1: you can remove the constructor

cristinamanum: Updated.

src/NuGet.Services.Search.Client/Client/ThreadSafeRandom.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    /// <summary>
+    /// A thread safe random implementation 
+    /// https://blogs.msdn.microsoft.com/pfxteam/2009/02/19/getting-random-numbers-in-a-thread-safe-way/
+    /// </summary>
+    public static class ThreadSafeRandom
+    {
+        private static Random _global = new Random();
+        [ThreadStatic]
+        private static Random _local;
```
skofman1: can you edit the code to comply with our coding style? I mean line breaks, curly brackets around lock statement. etc.

cristinamanum: Updated.

skofman1:
Fixes: https://github.com/NuGet/NuGetGallery/issues/5703

Issue #5119, opened 2017-12-05
Duplicate metadata elements in nuspec causes serialization of JSON metadata fields as arrays


scottbommarito:
Consider the following nuspec.

```
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
  <metadata>
    <id>chenriks-350mb-limit-test</id>
    <version>2.2.0</version>
    <title>chenriks-350mb-limit-test</title>
    <description>chenriks-350mb-limit-test</description>
    <authors>Microsoft</authors>
    <owners>Microsoft, CNTKTeam</owners>
    <requireLicenseAcceptance>false</requireLicenseAcceptance>
    <licenseUrl></licenseUrl>
    <projectUrl></projectUrl>
    <iconUrl></iconUrl>
    <description></description>
    <summary></summary>
    <releaseNotes></releaseNotes>
    <copyright>© Microsoft Corporation. All rights reserved</copyright>
    <tags></tags>
  </metadata>
</package>
```

`<description>` is duplicated

Now consider the generated registration JSON:

```
"@id": "https://az635243.vo.msecnd.net/v3-catalog0/data/2017.12.01.20.50.51/chenriks-350mb-limit-test.2.2.0.json",
"@type": "PackageDetails",
"authors": "Microsoft",
"description": [
    "chenriks-350mb-limit-test",
    ""
],
"iconUrl": "",
"id": "chenriks-350mb-limit-test",
"language": "",
"licenseUrl": "",
"listed": false,
"minClientVersion": "",
"packageContent": "https://apidev.nugettest.org/v3-flatcontainer/chenriks-350mb-limit-test/2.2.0/chenriks-350mb-limit-test.2.2.0.nupkg",
"projectUrl": "",
"published": "1900-01-01T00:00:00+00:00",
"requireLicenseAcceptance": false,
"summary": "",
"tags": [
    ""
],
"title": "chenriks-350mb-limit-test",
"version": "2.2.0"
```

`"description"` is serialized as an array instead of a `string`

This is incorrect and causes our monitoring to fail.

The gallery should reject this package and our V3 jobs should either ignore the duplicated field or skip the package.

xavierdecoster:
[...]

PR #5716, opened 2018-03-30
Reject duplicate metadata elements during upload

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/5119.

There will be another PR to make `catalog2lucene` resilient to such catalog leaves.

### DIFF
--- a/src/NuGetGallery.Core/CoreStrings.resx
+++ b/src/NuGetGallery.Core/CoreStrings.resx
@@ -189,4 +189,8 @@
   <data name="PackageIsMissingRequiredData" xml:space="preserve">
     <value>The package is missing required data.</value>
   </data>
+  <data name="Manifest_DuplicateMetadataElements" xml:space="preserve">
+    <value>The package manifest contains duplicate metadata elements: '{0}'</value>
+    <comment>{0} is a comma-delimited list of duplicate metadata element names.</comment>
+  </data>
 </root>
--- a/tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
+++ b/tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
@@ -2,10 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
-using System.Linq;
 using NuGet.Packaging;
 using NuGet.Packaging.Core;
 using NuGet.Versioning;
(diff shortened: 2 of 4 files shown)


src/NuGetGallery.Core/CoreStrings.resx
```diff
@@ -189,4 +189,8 @@
   <data name="PackageIsMissingRequiredData" xml:space="preserve">
     <value>The package is missing required data.</value>
   </data>
+  <data name="Manifest_DuplicateMetadataElements" xml:space="preserve">
+    <value>The package manifest contains duplicate metadata elements: '{0}'</value>
+    <comment>{0} is a comma seperated list of duplicate metadata element names.</comment>
```
shishirx34: separated

scottbommarito: `comma seperated` -> `comma-separated`

bonus points if you go for `comma-delimited`

tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
```diff
@@ -154,5 +228,86 @@ private static Stream CreateTestPackageStreamWithInvalidDependencyVersion()
...
+                        <description>package A description.</description>
+                        <language>en-US</language>
+                        <language>
+
+                        de-DE                        
+
+                        </language>
+                        <foo>1</foo>
+                        <foo>2</foo>
+                        <releaseNotes></releaseNotes>
```
scottbommarito: I'm slightly confused, because this method is called `CreateTestPackageStreamWithDuplicateNonEmptyMetadataElements` but some of these fields have duplicate empty metadata fields.

tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
```diff
@@ -154,5 +228,86 @@ private static Stream CreateTestPackageStreamWithInvalidDependencyVersion()
...
+
+        private static Stream CreateTestPackageStreamWithDuplicateNonEmptyMetadataElements()
+        {
+            var packageStream = new MemoryStream();
+            using (var packageArchive = new ZipArchive(packageStream, ZipArchiveMode.Create, true))
+            {
+                var nuspecEntry = packageArchive.CreateEntry("TestPackage.nuspec", CompressionLevel.Fastest);
+                using (var streamWriter = new StreamWriter(nuspecEntry.Open()))
+                {
+                    streamWriter.WriteLine(@"<?xml version=""1.0""?>
```
scottbommarito: As far as I can tell, these two methods are identical except for the contents of the `streamWriter`. You can write a helper method and reduce the code substantially.

tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
```diff
@@ -154,5 +228,86 @@ private static Stream CreateTestPackageStreamWithInvalidDependencyVersion()
...
+                    <package xmlns=""http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd"">
+                      <metadata>
+                        <id>TestPackage</id>
+                        <version>0.0.0.1</version>
+                        <title>Package A</title>
```
scottbommarito: It seems like this nuspec has a number of problems with it. In general, I prefer to see testing that tests each problem individually as well as combinations of those problems.

E.g. make this a function that takes in `titleIsDuplicated`, `authorsAreDuplicated`, `descriptionIsDuplicated`, etc and then test all combinations using a `Theory`.

tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
joelverhagen: I'll rename this to `EmptyAndNonEmpty`

tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
joelverhagen: Sounds good, I'll give this a try.

src/NuGetGallery.Core/CoreStrings.resx
joelverhagen: Sounds fine.

tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
joelverhagen: Fixed. Nice suggestion.

tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
joelverhagen: Decided against this since it really wouldn't give us any test coverage benefit in this case and would make the test code more complex.

tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
joelverhagen: Fixed.

src/NuGetGallery.Core/CoreStrings.resx
joelverhagen: Fixed.

Issue #5736, opened 2018-04-02
[Organizations] Spamming add member for an organization shows duplicate member in UI


scottbommarito:
1 - go to edit org page
2 - enter in a user to add
3 - press enter many times in quick succession

Observe that there are many copies of the user you meant to add.

This has no functional effects, just broken UI.

NOTE: this is much harder to show locally, because the server response is much faster. It is very evident on DEV.

Issue #5730, opened 2018-04-02
[Organizations] Edit organization page for collaborators should explain why editing fields is locked


scottbommarito:
When they try to edit an organization, collaborators should be told that they cannot edit any fields because they are only a collaborator.

anangaur:
`You are a collaborator. Only Administrators can makes changes to this organization.`

PR #5749, opened 2018-04-02
[Organizations] Edit organization UI fixes

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5730:

image

https://github.com/NuGet/NuGetGallery/issues/5736

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-manage-organization.js
+++ b/src/NuGetGallery/Scripts/gallery/page-manage-organization.js
@@ -230,6 +230,13 @@
                     data: data,
                     success: function (data) {
                         self.Error(null);
+
+                        // Remove any duplicates of this user in the UI.
+                        // This can happen if a user makes multiple add requests before a response is received.
+                        self.Members.remove(function (member) {
+                            return member.Username.toLocaleLowerCase() === data.Username.toLocaleLowerCase();
+                        });
+
                         self.Members.push(new OrganizationMemberViewModel(self, data));
                         self.NewMemberUsername(null);
                     },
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Scripts/gallery/page-manage-organization.js
```diff
@@ -230,6 +230,13 @@
                     data: data,
                     success: function (data) {
                         self.Error(null);
+
+                        // Remove any duplicates of this user in the UI.
+                        // This can happen if a user makes multiple add requests before a response is received.
+                        self.Members.remove(function (member) {
```
skofman1: can we lock the button until we receive a response?

anangaur:
How come this org has 0 admins and 1 collaborator?

anangaur:
There needs to be a profile icon link at the start, here too.

scottbommarito:
@anangaur pls ignore 0 admin, bad data from testing

profile icon link is out of scope for this PR

src/NuGetGallery/Scripts/gallery/page-manage-organization.js
scottbommarito: Doesn't seem intuitive, especially if the user has a bad network connection and they want to add several people in a row without waiting for network response. This is a better approach IMO--we are updating the UI with the most recent update from the server and removing the old data.

src/NuGetGallery/Scripts/gallery/page-manage-organization.js
```diff
@@ -230,6 +230,13 @@
...
+                            return member.Username.toLocaleLowerCase() === data.Username.toLocaleLowerCase();
+                        });
+
                         self.Members.push(new OrganizationMemberViewModel(self, data));
```
chenriksson: Instead of `remove`, why not check `inArray` before doing push?

scottbommarito: The data could have changed (e.g. the user was added as admin instead of collaborator) so we should always replace it with the response from the server.

Issue #5718, opened 2018-03-30
Consider using nuspec XSD during pushed package validation in gallery


joelverhagen:
This would give us much more strict validation and probably cover a lot of oddball cases.

Some open questions:

1. What version of the XSD? Latest version of client?
1. The XSD has some things that only make sense at pack time, e.g. `<files>` element.
1. Will the error messages be good enough for the web UI?
1. How does this fit into PC to PR issues?

/cc @rrelyea @skofman1 @rohit21agrawal

joelverhagen:
From https://github.com/NuGet/NuGetGallery/pull/5750#discussion_r178650284.

> > IMO the correct approach for input validation is to define what is allowed, instead of what is not allowed.
> 
> Yes, I agree. However, we need to analyze all of the metadata currently being pushed by package authors (and is being accepted) to make sure we are not breaking some scenarios we actually may care about.

PR #5750, opened 2018-04-02
Reject invalid boolean values and properties generated by feed2catalog

joelverhagen:
Address on https://github.com/NuGet/Engineering/issues/1308.
Address on https://github.com/NuGet/Engineering/issues/1309.
Address on https://github.com/NuGet/Engineering/issues/1310.

In short, two validations on upload or revalidate are being added:

1. Reject metadata element names that collide with those generated by feed2catalog.
1. Reject non-boolean values for boolean metadata fields.

Instead of overhauling .nuspec/.nupkg validation in gallery with XSD or some other strict approach, I am taking a minimal-ish approach that is least likely to break existing package authors. The goal here is to reject .nuspec metadata child elements that break V3. See the aforementioned issues for what is breaking. I don't want to tackle the problem of rejecting anything but the most perfect .nuspecs 😄. That is work that should be done... but not without more thorough analysis of all .nuspec oddities out there. This broader item is tracked here: https://github.com/NuGet/NuGetGallery/issues/5718.

I am currently analyzing all .nuspecs on NuGet.org checking for these two new validation cases. My expectation is that none of the existing packages are on NuGet.org because they would have caused live site incidents. My analysis should be done within the next couple hours.

References:

- [Client .nuspec XSD used at `pack` time](https://github.com/NuGet/NuGet.Client/blob/deab84db30cbd8cc738cd46ad694aebdbb43a1cf/src/NuGet.Core/NuGet.Packaging/compiler/resources/nuspec.xsd)
- [feed2catalog XSLT](https://github.com/NuGet/NuGet.Services.Metadata/blob/f7a42142d0674fb8a8bc09b4c48e51595e331796/src/Catalog/xslt/nuspec.xslt)
- [Catalog leaf JSON-LD frame](https://github.com/NuGet/NuGet.Services.Metadata/blob/f7a42142d0674fb8a8bc09b4c48e51595e331796/src/Catalog/context/PackageDetails.json)

### DIFF
--- a/src/NuGetGallery.Core/Packaging/PackageMetadata.cs
+++ b/src/NuGetGallery.Core/Packaging/PackageMetadata.cs
@@ -13,6 +13,33 @@ namespace NuGetGallery.Packaging
 {
     public class PackageMetadata
     {
+        /// <summary>
+        /// These are properties generated in the V3 pipeline (feed2catalog job) and could collide if the .nuspec
+        /// itself also contains these properties.
+        /// </summary>
+        private static readonly HashSet<string> RestrictedMetadataElements = new HashSet<string>
+        {
+            "created",
+            "dependencyGroups", // Not to be confused with the valid element "dependencies" with a sub-element "group".
+            "isPrerelease",
+            "lastEdited",
+            "listed",
+            "packageEntries",
+            "packageHash",
+            "packageHashAlgorithm",
+            "packageSize",
+            "published",
+            "supportedFrameworks",
+            "verbatimVersion",
+        };
+
+        private static readonly HashSet<string> BooleanMetadataElements = new HashSet<string>
+        {
+            "developmentDependency",
+            "requireLicenseAcceptance",
+            "serviceable",
+        };
+
         private readonly Dictionary<string, string> _metadata;
         private readonly IReadOnlyCollection<PackageDependencyGroup> _dependencyGroups;
         private readonly IReadOnlyCollection<FrameworkSpecificGroup> _frameworkReferenceGroups;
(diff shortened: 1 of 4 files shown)


src/NuGetGallery.Core/Packaging/PackageMetadata.cs
```diff
@@ -13,6 +13,33 @@ namespace NuGetGallery.Packaging
 {
     public class PackageMetadata
     {
+        /// <summary>
+        /// These are properties generated in the V3 pipeline (feed2catalog job) and could collide if the .nuspec
+        /// itself also contains these properties.
+        /// </summary>
+        private static readonly HashSet<string> RestrictedMetadataElements = new HashSet<string>
+        {
+            "created",
```
skofman1: IMO the correct approach for input validation is to define what is allowed, instead of what is not allowed.
I also understand that this is the cheaper/faster approach to prevent packages that break V3 indexing.

Do we have an issue to track replacing the logic in this PR?

joelverhagen: > IMO the correct approach for input validation is to define what is allowed, instead of what is not allowed.

Yes, I agree. However, we need to analyze all of the metadata currently being pushed by package authors (and is being accepted) to make sure we are not breaking some scenarios we actually may care about.

> Do we have an issue to track replacing the logic in this PR?

https://github.com/NuGet/NuGetGallery/issues/5718

joelverhagen:
> I am currently analyzing all .nuspecs on NuGet.org checking for these two new validation cases. My expectation is that none of the existing packages are on NuGet.org because they would have caused live site incidents. My analysis should be done within the next couple hours.

Done. None of the newly rejected elements are in any of the .nuspecs on NuGet.org.

The only oddities found were an uppercase `False`:
```
<requireLicenseAcceptance>False</requireLicenseAcceptance>
```

Found in these two packages:
https://www.nuget.org/packages/TestPackage19fa75eb-9384-4371-9f82-0f4348c0aad3/1.0.0
https://www.nuget.org/packages/TestPackage19fa75eb-9384-4371-9f82-0f4348c0aad3/1.0.1

This is totally fine though since `False` is parsable by `bool.TryParse`.

Issue #5731, opened 2018-04-02
[Organizations] Organization collaborators should have same permissions on packages as administrators


scottbommarito:
Discussing offline with @skofman1, we should allow collaborators to upload new package IDs and contact support, but not manage package ownership.

(check API key creation flow to make sure it uses permissions service)

anangaur:
Collaborators:
* Manage packages - Upload/Update/Edit/Unlist/Delete

Administrators:
* Same perms for Manage packages as Collaborators +
* Manage ownership
* Manage Organization 
   - Edit metadata
   - Notifications
   - Add/Remove members 
   - ...

Issue #5104, opened 2017-12-01
Cleanup ApiKeys UI to use PermissionsService


chenriksson:
This is work remaining from S127. See:
https://github.com/NuGet/NuGetGallery/issues/4872

PR #5755, opened 2018-04-03
[Organizations] Give organization collaborators the same permissions on packages as administrators

scottbommarito:
We decided that allowing collaborators not to push new IDs and contact support was a weird experience.

This PR includes making that permissions change and also updating the API keys page to use the permissions service correctly.

https://github.com/NuGet/NuGetGallery/issues/5731
https://github.com/nuget/nugetgallery/issues/5104

### DIFF
--- a/src/NuGetGallery/ViewModels/ApiKeyOwnerViewModel.cs
+++ b/src/NuGetGallery/ViewModels/ApiKeyOwnerViewModel.cs
@@ -7,15 +7,19 @@ namespace NuGetGallery
 {
     public class ApiKeyOwnerViewModel
     {
-        public ApiKeyOwnerViewModel(string owner, bool canPushNew, IList<string> packageIds)
+        public ApiKeyOwnerViewModel(string owner, bool canPushNew, bool canPushExisting, bool canUnlist, IList<string> packageIds)
         {
             Owner = owner;
             CanPushNew = canPushNew;
+            CanPushExisting = canPushExisting;
+            CanUnlist = canUnlist;
             PackageIds = packageIds;
         }
         
         public string Owner { get; }
         public bool CanPushNew { get; }
+        public bool CanPushExisting { get; }
+        public bool CanUnlist { get; }
         public IList<string> PackageIds { get; }
     }
 }
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -364,10 +364,10 @@ namespace NuGetGallery
         [UIAuthorize]
         public virtual ActionResult ApiKeys()
         {
-            var user = GetCurrentUser();
+            var currentUser = GetCurrentUser();
 
             // Get API keys
-            if (!GetCredentialGroups(user).TryGetValue(CredentialKind.Token, out List<CredentialViewModel> credentials))
+            if (!GetCredentialGroups(currentUser).TryGetValue(CredentialKind.Token, out List<CredentialViewModel> credentials))
             {
                 credentials = new List<CredentialViewModel>();
             }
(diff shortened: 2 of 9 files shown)


src/NuGetGallery/ViewModels/ApiKeyOwnerViewModel.cs
```diff
@@ -7,15 +7,19 @@ namespace NuGetGallery
 {
     public class ApiKeyOwnerViewModel
     {
-        public ApiKeyOwnerViewModel(string owner, bool canPushNew, IList<string> packageIds)
+        public ApiKeyOwnerViewModel(string owner, bool canPushNew, bool canPushExisting, bool canUnlist, IList<string> packageIds)
         {
             Owner = owner;
             CanPushNew = canPushNew;
+            CanPushExisting = canPushExisting;
```
skofman1: why do we need the split for canPushNew, canPushExisting, canUnlist? Everyone with access to the package can do all of those.

scottbommarito: Because the permissions service makes it a possibility that any user can have the ability to do any subset of those actions. Currently, because of how we've configured the system, they all have the same permissions requirements, but as we've just demonstrated by changing it with this PR, these permissions requirements may change.

skofman1: Right now, since there's no such requirement, this granularity introduces an unnecessary complexity.   I recommend changing the permissions service to have a single level of "write permissions for package", and making this change in all other layers as well.

scottbommarito: After an offline discussion, @skofman1 and I came to the conclusion that removing the permissions service checks would be more difficult than leaving them in.

src/NuGetGallery/ViewModels/ApiKeyOwnerViewModel.cs
```diff
@@ -7,15 +7,19 @@ namespace NuGetGallery
 {
     public class ApiKeyOwnerViewModel
     {
-        public ApiKeyOwnerViewModel(string owner, bool canPushNew, IList<string> packageIds)
+        public ApiKeyOwnerViewModel(string owner, bool canPushNew, bool canPushExisting, bool canUnlist, IList<string> packageIds)
```
shishirx34: Instead of passing all possible booleans, create a class for the possible permissions and pass it in here.
```
public class PackagePermissions
{
    public bool CanPushNew { get; set; }
    public bool CanPushExisting { get; set; }
    public bool CanUnlist { get; set; }
}
```

src/NuGetGallery/ViewModels/ApiKeyOwnerViewModel.cs
```diff
@@ -7,15 +7,19 @@ namespace NuGetGallery
...
             CanPushNew = canPushNew;
+            CanPushExisting = canPushExisting;
+            CanUnlist = canUnlist;
             PackageIds = packageIds;
         }
         
         public string Owner { get; }
         public bool CanPushNew { get; }
+        public bool CanPushExisting { get; }
+        public bool CanUnlist { get; }
```
shishirx34: Use `PackagePermissions` object here as well.

src/NuGetGallery/ViewModels/ApiKeyOwnerViewModel.cs
scottbommarito: I don't see a benefit to this. I don't see this class being reused anywhere else and I don't see this structure adding any additional requirements to the code that improve it.

I think I will change the code such that the user provides a `User` and a `PackageRegistration` and this constructor calculates these values for you, but I don't see abstracting the permissions booleans into a class adding any benefits.

src/NuGetGallery/ViewModels/ApiKeyOwnerViewModel.cs
chenriksson: Consider checking if each CanX permissions check results in a separate DB query. You can do this by enabling logging on the context, and just stepping through the code with the debugger (see below). If each check is a new DB query, then it's worth the effort to reduce the granularity. 

Aside from performance, it may be worth it for the code reduction as well.

```
public EntitiesContext(string connectionString, bool readOnly)
    : base(connectionString)
{
    ReadOnly = readOnly;
    Database.Log = s => System.Diagnostics.Debug.WriteLine(s);
}
```

scottbommarito: I think due to the large amount of coding work required to reduce the granularity of the properties, I don't think the performance concerns here are very pressing, especially given that `ListItemPackageViewModel` has 5-10 properties that would likely have the same behavior in terms of DB calls.

Issue #5725, opened 2018-04-02
[Organizations] Add new organization page improvements


scottbommarito:
- needs learn more link
- org name should appear in test profile link (do this for register flow as well?)
- explain gravatar for org email
- "the username is not available" should appear lower

scottbommarito:
@anangaur what link should this page use for its learn more link?

anangaur:
https://go.microsoft.com/fwlink/?linkid=870439

anangaur:
![image](https://user-images.githubusercontent.com/14800916/38215239-f96c8172-367b-11e8-9e66-f86796ddcc55.png)

Font should be smaller.

PR #5757, opened 2018-04-03
[Organizations] Add new organization page improvements

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5725

(work to fix the breadcrumb at the top will be done in a separate PR with other breadcrumb work)

image

### DIFF
--- a/src/NuGetGallery/Controllers/OrganizationsController.cs
+++ b/src/NuGetGallery/Controllers/OrganizationsController.cs
@@ -64,8 +64,6 @@ namespace NuGetGallery
             var organizationEmailAddress = model.OrganizationEmailAddress;
             var adminUser = GetCurrentUser();
 
-            string errorMessage;
-
             try
             {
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -74,11 +72,9 @@ public ActionResult Add()
             }
             catch (EntityException e)
             {
-                errorMessage = e.Message;
+                TempData["AddOrganizationErrorMessage"] = e.Message;
```
skofman1: UT?

scottbommarito:
@anangaur tool-tips can't have links inside them. I can make the tool-tip itself a link to gravatar.com however.

src/NuGetGallery/Controllers/OrganizationsController.cs
scottbommarito: There was one that was broken by this change that I have now fixed.

scottbommarito:
as per offline discussion with @anangaur, using tool-tip but without link

Issue #5743, opened 2018-04-02
Manage package owners UI fixes


scottbommarito:
- remove button should be properly right-aligned
- no need for confirmation on add if user doesn't have any security policies

Issue #5739, opened 2018-04-02
[Organizations] Pending requests sent by a user who has the right to manage the account the request is being sent to should be accepted automatically


scottbommarito:
User A is admin of org B.
A would like to add B as a package owner of package C.
When A adds B as a package owner, there should be no pending request process because A has the rights to accept the request on behalf of B.

PR #5759, opened 2018-04-03
[Organizations] Manage package owners fixes

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5739
https://github.com/NuGet/NuGetGallery/issues/5743

Basically:
- if you add an organization you own as an owner, there is no longer any need to confirm the request
- removed the AJAX confirmation that is no longer necessary due to the removal of `SecurePushPolicy`
- fixed the right alignment of the "remove" button

image

### DIFF
--- a/src/NuGetGallery/Controllers/JsonApiController.cs
+++ b/src/NuGetGallery/Controllers/JsonApiController.cs
@@ -104,26 +104,6 @@ namespace NuGetGallery
             return Json(result, JsonRequestBehavior.AllowGet);
         }
 
-        [HttpGet]
-        public virtual ActionResult GetAddPackageOwnerConfirmation(string id, string username)
-        {
-            ManagePackageOwnerModel model;
-            if (TryGetManagePackageOwnerModel(id, username, isAddOwner: true, model: out model))
-            {
-                return Json(new
-                {
-                    success = true,
-                    confirmation = string.Format(CultureInfo.CurrentCulture, Strings.AddOwnerConfirmation, username),
-                    policyMessage = string.Empty
-                },
-                JsonRequestBehavior.AllowGet);
-            }
-            else
-            {
-                return Json(new { success = false, message = model.Error }, JsonRequestBehavior.AllowGet);
-            }
-        }
-
         [HttpPost]
         [ValidateAntiForgeryToken]
         public async Task<JsonResult> AddPackageOwner(string id, string username, string message)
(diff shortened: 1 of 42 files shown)


scottbommarito:
feedback from @skofman1: reduce margin between icon and username in current owners

this is likely hard to do given bootstrap and I will not include it in this PR if it would require too much finagling with the CSS.

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -104,64 +104,56 @@ public virtual ActionResult GetPackageOwners(string id, string version)
...
 
-                var confirmationUrl = Url.ConfirmPendingOwnershipRequest(
-                    model.Package.Id,
-                    model.User.Username,
-                    ownerRequest.ConfirmationCode,
-                    relativeUrl: false);
+                    foreach (var owner in model.Package.Owners)
+                    {
+                        _messageService.SendPackageOwnerAddedNotice(owner, model.User, model.Package, packageUrl);
+                    }
```
chenriksson: Per `PackagesController.SendAddPackageOwnerNotification`, this notification isn't intended for the new owner.

Before the new owner would receive the request in email. Now that we auto-confirm, maybe we should send a customized notification to the new owner that sounds similar to the request email... "The user X has added you as an owner of package Y..."

src/NuGetGallery/Scripts/gallery/page-manage-owners.js
```diff
@@ -100,19 +100,6 @@
...
-            }
-
-            var existingOwner = ko.utils.arrayFirst(
-                viewModel.owners(),
-                function (owner) { return owner.name().toUpperCase() == newUsername.toUpperCase() });
-
-            if (existingOwner) {
-                viewModel.message(window.nuget.formatString(strings_AlreadyPending, newUsername));
-                return;
-            }
```
chenriksson: Does validation in addOwner mean it's not needed in confirmAddOwner anymore?

chenriksson: Is confirmAddOwner still used? Looks like it was removed from ManagePackageOwners.cshtml?

scottbommarito: good catch, we can remove the entire confirmAddOwner function

src/NuGetGallery/Controllers/JsonApiController.cs
scottbommarito: I think the only reason why `PackagesController.SendAddPackageOwnerNotification` isn't sent to the new owner is because the user gets the information in the UI when they accept the request. I think the text of the notification (`User 'x' is now an owner of the package 'y'.`) is fine to send to the owner who was added. I don't think a custom email is necessary.

PR #5784, opened 2018-04-09
[Organizations] Allow collaborators to create API keys scoped to an organization with the ability to push new IDs

scottbommarito:
Unfortunately missed this bug when fixing https://github.com/NuGet/NuGetGallery/issues/5731

### DIFF
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -777,10 +777,9 @@ namespace NuGetGallery
                 Response.StatusCode = (int)HttpStatusCode.BadRequest;
                 return Json(Strings.UserNotFound);
             }
-
-            // todo: move validation logic to PermissionsService
+            
             var resolvedScopes = BuildScopes(scopeOwner, scopes, subjects);
-            if (!VerifyScopes(scopeOwner, resolvedScopes))
+            if (!VerifyScopes(resolvedScopes))
             {
                 Response.StatusCode = (int)HttpStatusCode.BadRequest;
                 return Json(Strings.ApiKeyScopesNotAllowed);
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -855,29 +854,57 @@ public virtual ActionResult PasswordChanged()
             return credentialViewModel;
         }
 
-        // todo: integrate verification logic into PermissionsService.
-        private bool VerifyScopes(User scopeOwner, IEnumerable<Scope> scopes)
+        private static IDictionary<string, IActionRequiringEntityPermissions> AllowedActionToActionRequiringEntityPermissions = new Dictionary<string, IActionRequiringEntityPermissions>
```
loic-sharma: Nit: fields normally are defined before methods.

scottbommarito: I thought this was fine given that it's exclusively used for this method haha

loic-sharma: Also, would the name `AllowedScopeToActionRequiringEntityPermissions`?

scottbommarito: No--unfortunately `NuGetScopes` was named poorly and has a misleading name. `NuGetScopes` makes a lot more sense if you think of it as `NuGetScopeActions`, which is why I named this `AllowedActionToActionRequiringEntityPermissions`.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -855,29 +854,57 @@ public virtual ActionResult PasswordChanged()
...
-                if (organization != null)
+                if (string.IsNullOrEmpty(scope.AllowedAction))
+                {
+                    // All scopes must have an allowed action.
+                    return false;
+                }
+                
+                // Get the list of actions allowed by this scope.
+                var actions = new List<IActionRequiringEntityPermissions>();
+                foreach (var allowedAction in AllowedActionToActionRequiringEntityPermissions.Keys)
```
chenriksson: Why three `for` loops? Isn't there 1 `ActionsRequiringPermissions` per `NuGetScopes`?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -855,29 +854,57 @@ public virtual ActionResult PasswordChanged()
...
         {
-            var currentUser = GetCurrentUser();
-
-            // scoped to the user
-            if (currentUser.MatchesUser(scopeOwner))
+            { NuGetScopes.PackagePush, ActionsRequiringPermissions.UploadNewPackageId },
+            { NuGetScopes.PackagePushVersion, ActionsRequiringPermissions.UploadNewPackageVersion },
+            { NuGetScopes.PackageUnlist, ActionsRequiringPermissions.UnlistOrRelistPackage },
+            { NuGetScopes.PackageVerify, ActionsRequiringPermissions.VerifyPackage },
+        };
```
chenriksson: Can you just create a helper on `NuGetScopes` or `ActionsRequiringPermissions` that does the 1:1 conversion?

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: As far as I can tell from the code, the `AllowedAction` could be `NuGetScopes.All` which means we need to check the upload new, upload existing, unlist, and verify actions.

scottbommarito: To clarify, the loop is best explained like this:

For each scope:
 1) Get the actions that it grants the user to perform
 2) Check that each action can be performed by the user

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: I was going to do this if `ApiScopeEvaluator` did any conversion similar to this, but it doesn't seem to be the case. This code doesn't seem to be useful to share outside of this file.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -855,29 +854,57 @@ public virtual ActionResult PasswordChanged()
             return credentialViewModel;
         }
 
-        // todo: integrate verification logic into PermissionsService.
-        private bool VerifyScopes(User scopeOwner, IEnumerable<Scope> scopes)
+        private static IDictionary<string, IActionRequiringEntityPermissions[]> AllowedActionToActionRequiringEntityPermissions = new Dictionary<string, IActionRequiringEntityPermissions[]>
```
skofman1: perhaps call this field something with "map"?

PR #5786, opened 2018-04-09
Use JavaScript module pattern and 'use strict';

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/1313.

- Add module pattern and `use strict;` to closures returning in global state change
  - Calls to `$(...)`
  - Global var, i.e. `var foo =` at the root scope of a .js file
- Remove two unused scripts


joelverhagen:
View this with https://github.com/NuGet/NuGetGallery/pull/5786/files?w=1.

Issue #5790, opened 2018-04-10
Content-Security-Policy has unrecognized directive


joelverhagen:
In latest Chrome web tools console on page load:

```
Unrecognized Content-Security-Policy directive 'frame-ancestor'.
```

Current header value:
```
Content-Security-Policy: frame-ancestor 'none'
```

I think this should be **frame-ancestors** (plural).

joelverhagen:
MDN:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

PR #5793, opened 2018-04-11
[Hotfix]Correct a typo in the CSP header.

ryuyu:
Correct a typo in CSP header value.


joelverhagen:
Addresses https://github.com/NuGet/NuGetGallery/issues/5790.

PR #5795, opened 2018-04-12
Fix accessibility issues identified by tool

scottbommarito:
https://github.com/NuGet/Engineering/issues/1359

This PR fixes a number of small issues with the accessibility of our site as identified by a tool.
Mostly consists of changing roles and adding labels.

### DIFF
--- a/src/Bootstrap/dist/css/bootstrap-theme.css
+++ b/src/Bootstrap/dist/css/bootstrap-theme.css
@@ -93,13 +93,13 @@ body h3 {
 .navbar-logo {
   margin: 8px 20px 0 0;
 }
-.navbar-seperated .seperator {
+.navbar-separated .seperator {
   display: block;
   padding: 11px 0;
   font-weight: 100;
   color: #fff;
 }
-.navbar-seperated .seperator > span::after {
+.navbar-separated .seperator > span::after {
   content: " | ";
 }
 .dropdown-menu {
--- a/src/Bootstrap/dist/css/bootstrap-theme.css
+++ b/src/Bootstrap/dist/css/bootstrap-theme.css
@@ -87,9 +87,6 @@ body h3 {
   height: auto;
   padding-bottom: 75px;
 }
-.main-container .page-subheading {
-  color: #777;
-}
 .navbar-logo {
   margin: 8px 20px 0 0;
 }
--- a/src/NuGetGallery/Scripts/gallery/common.js
+++ b/src/NuGetGallery/Scripts/gallery/common.js
@@ -50,8 +50,43 @@
                 } else {
                     error.insertAfter(element);
                 }
+            },
+            showErrors: function (errorMap, errorList) {
+                this.defaultShowErrors();
+
+                var i;
+                for (i = 0; this.errorList[i]; i++) {
+                    removeInvalidAriaDescribedBy(this.errorList[i].element);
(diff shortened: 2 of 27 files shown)


src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -50,10 +50,45 @@
                 } else {
                     error.insertAfter(element);
                 }
+            },
+            showErrors: function (errorMap, errorList) {
+                this.defaultShowErrors();
+
+                var i;
```
scottbommarito: JQuery Validate adds an `aria-describedby` attribute to every validated field, even if the validation is successful. This is wrong because if there is no error, the `aria-describedby` leads to an empty element, which is invalid. This code removes the unnecessary attribute.

src/Bootstrap/dist/css/bootstrap-theme.css
```diff
@@ -87,19 +87,16 @@ body h3 {
...
   padding-bottom: 75px;
 }
-.main-container .page-subheading {
-  color: #777;
-}
 .navbar-logo {
   margin: 8px 20px 0 0;
 }
-.navbar-seperated .seperator {
+.navbar-separated .seperator {
```
agr: Fix the `seperator` as well, maybe?

scottbommarito: apparently we don't even have a `seperator`, which means the entire `navbar-separated` class is useless

removed it entirely

src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -50,10 +50,48 @@
                 } else {
                     error.insertAfter(element);
                 }
+            },
+            showErrors: function (errorMap, errorList) {
```
chenriksson: How is this used?

scottbommarito: `jquery.validate` calls it whenever the state of our forms has changed. It adds validation error text to our forms when the content isn't correct. I'm using their default process and appending an additional step.

src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -50,10 +50,48 @@
...
+                if (!describedById) {
+                    return false;
+                }
+
+                var describedByElement = $("#" + describedById);
+                return describedByElement && describedByElement.text();
+            });
+
+        if (ids.length) {
+            element.setAttribute("aria-describedby", ids.join(" "));
```
joelverhagen: Only call `setAttribute if the IDs set has changed.

src/NuGetGallery/Views/Authentication/SignIn.cshtml
```diff
@@ -44,9 +44,9 @@
                 </div>
             </div>
             <div class="row text-center create-provider-account">
```
joelverhagen: There should only be one of these... not one per provider. You can put it in a `Model.Providers.Count > 0` `if` depending on what @shishirx34 things.

src/NuGetGallery/Scripts/gallery/common.js
scottbommarito: Is `setAttribute` an expensive operation? I feel like the code is much simpler like this.

PR #5797, opened 2018-04-12
Sb deleteorgmember

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5025

User view:
image

Admin view:
image


tests/NuGetGallery.Facts/ViewModels/DeleteAccountViewModelFacts.cs
```diff
@@ -1,113 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
```
scottbommarito: I removed this file because it seemed redundant--it's only testing a field on `ListPackageItemViewModel` that is passed to `DeleteAccountViewModel`. The field already has its own tests in `ListPackageItemViewModel`.

src/NuGetGallery/Areas/Admin/ViewModels/DeleteUserAccountViewModel.cs
```diff
@@ -5,20 +5,32 @@
 
 namespace NuGetGallery.Areas.Admin.ViewModels
 {
-    public class DeleteUserAccountViewModel : DeleteAccountViewModel
+    public class DeleteUserAccountViewModel
     {
         public DeleteUserAccountViewModel()
         {
+            // This constructor exists so that we can have a form that uses this model.
```
chenriksson: nit: move to doc comment?

We need this and the setters because of the `_DeleteUserAccountForm.cshtml` postback?

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteUserAccount.cshtml
```diff
@@ -1,5 +1,5 @@
 ﻿@using NuGetGallery
-@model NuGetGallery.Areas.Admin.ViewModels.DeleteUserAccountViewModel
+@model DeleteAccountViewModel
```
chenriksson: Why change from `DeleteUserAccountViewModel` to `DeleteAccountViewModel`?

If this is going to be shared for deleting org accounts, should the view also change to `DeleteAccount.cshtml`?

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/_DeleteUserAccountForm.cshtml
```diff
@@ -0,0 +1,39 @@
...
+        @Html.AntiForgeryToken()
+        @Html.ValidationSummary(true)
+        <div>
+            <div class="form-group">
+                @Html.LabelFor(m => m.Signature)
+                @Html.EditorFor(m => m.Signature)
+                @Html.ValidationMessageFor(m => m.Signature)
+            </div>
+
+            @if (Model.HasOrphanPackages)
```
chenriksson: Was the `HasOrphanPackages` already updated to account for packages for the member's organizations?

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/_DeleteUserAccountForm.cshtml
```diff
@@ -0,0 +1,39 @@
...
+            {
+                <div class="form-group">
+                    @Html.EditorFor(m => m.ShouldUnlist)
+                    <label class="checkbox-inline">
+                        <b>Unlist the packages without any user.</b>
```
chenriksson: Maybe `Unlist the packages without any owner` is better?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -224,6 +200,43 @@ IAuditingService auditingService
...
+            }
+
+            var migrationRequest = user.OrganizationMigrationRequest;
+            user.OrganizationMigrationRequest = null;
+            if (migrationRequest != null)
+            {
+                migrationRequest.AdminUser.OrganizationMigrationRequests.Remove(migrationRequest);
+            }
+
+            await _entitiesContext.SaveChangesAsync();
```
chenriksson: There's logic in `DeleteGalleryUserAccountAsync` to `commitAsTransaction`. Given that,
- Should this call `SaveChangesAsync`?
- Are all of the `RemoveX` operations honoring the deferred commit?

src/NuGetGallery/ViewModels/DeleteAccountViewModel.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using NuGetGallery.Areas.Admin;
+using NuGetGallery.Areas.Admin.Models;
```
chenriksson: nit: sort after `System` usings

src/NuGetGallery/ViewModels/DeleteAccountViewModel.cs
```diff
@@ -10,19 +12,36 @@ namespace NuGetGallery
...
+
+            Packages = packageService
+                 .FindPackagesByAnyMatchingOwner(User, includeUnlisted: true)
+                 .Select(p => new ListPackageItemViewModel(p, currentUser))
+                 .ToList();
+
+            HasPendingRequests = supportRequestService.GetIssues()
+                .Where(issue => 
+                    (issue.UserKey.HasValue && issue.UserKey.Value == User.Key) &&
+                    string.Equals(issue.IssueTitle, Strings.AccountDelete_SupportRequestTitle) &&
```
chenriksson: Ignore case?

src/NuGetGallery/ViewModels/DeleteAccountViewModel.cs
```diff
@@ -10,19 +12,36 @@ namespace NuGetGallery
...
             _hasOrphanPackages = new Lazy<bool>(() => Packages.Any(p => p.HasSingleOwner));
+
+            User = userToDelete;
+
+            Packages = packageService
+                 .FindPackagesByAnyMatchingOwner(User, includeUnlisted: true)
+                 .Select(p => new ListPackageItemViewModel(p, currentUser))
+                 .ToList();
+
+            HasPendingRequests = supportRequestService.GetIssues()
```
chenriksson: Aside from checking for support requests, are we also deleting them? Could there be PII in the support requests that necessitate deleting them as part of Delete Account?

src/NuGetGallery/Views/Users/DeleteAccount.cshtml
```diff
@@ -35,6 +35,7 @@
                             <li>Revoke your api key(s).</li>
                             <li>Remove you as the owner for any packages you own.</li>
                             <li>Remove your ownership from any ID prefix reservations and delete any ID prefix reservations that you were the only owner of.</li>
+                            <li>Remove your membership from any organizations you are a member of.</li>
```
chenriksson: nit: would remove redundancy... `Remove your membership from any organizations` or `Remove yourself from any organizations that you are a member of`

src/NuGetGallery/Views/Users/_UserOrganizationsListForDeletedAccount.cshtml
```diff
@@ -0,0 +1,73 @@
...
+                        </tr>
+                    </thead>
+                    <tbody>
+                        @foreach (var organization in Model)
+                        {
+                            <tr class="manage-package-listing">
+                                <td class="align-middle ">
+                                    @if (organization.MemberCount == 1)
+                                    {
+                                        <i class="ms-Icon ms-Icon--StatusErrorFull red-check" title="After account delete the organization will not have any members."></i>
```
chenriksson: I wonder if we should just automatically delete any orphaned organizations... to avoid support requests. We're already unlisting all the packages. This user can't easily request the orphaned org(s) be deleted after their account is deleted.

src/NuGetGallery/Areas/Admin/ViewModels/DeleteUserAccountViewModel.cs
scottbommarito: yup

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteUserAccount.cshtml
scottbommarito: I changed this because

`DeleteUserAccountViewModel` used to inherit from `DeleteAccountViewModel`.
`DeleteUserAccountViewModel` contained the information required by the form, while `DeleteAccountViewModel` had the information required by rendering the view.
I thought having `DeleteUserAccountViewModel` depend on `DeleteAccountViewModel` was confusing because the two objects served very different purposes and the improper inheritance would lead to `DeleteAccountViewModel`s being created with empty fields.

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/_DeleteUserAccountForm.cshtml
scottbommarito: Yes

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: As far as I understand
- `SaveChangesAsync` "enlists in any ambient transactions" e.g. it attaches the changes to the ongoing transaction
- other services that are called also call `SaveChangesAsync`

I was under the impression this was the right way to do things but if it's wrong then we'll need to fix other things as well. In my experience when I was testing this and it failed at this point, it rolled back any changes that it was attempting to make.

src/NuGetGallery/ViewModels/DeleteAccountViewModel.cs
scottbommarito: Deletion is done inside `DeleteAccountService` I believe.

src/NuGetGallery/ViewModels/DeleteAccountViewModel.cs
scottbommarito: Existing code that I moved here. Ignore case is probably correct though.

src/NuGetGallery/Views/Users/_UserOrganizationsListForDeletedAccount.cshtml
scottbommarito: I can't imagine that anyone would be too concerned about their orphaned organization if they were willing to orphan it.
I think leaving it as an organization might make it easier for a real user to take over ownership of the packages (the next PR after this is to allow site admins to change any organization's membership).

anangaur:
Should we show admins instead of members? Because ultimately you would need additional admins to be present for an organization.

In case the user deleting his/her account is the only Admin, we should just delete the orgs as well and make all the packages orphans. We do not want to end up in a situation where there is a dangling organization with only collaborators.

OR

We continue to show members and not just admins.
1. Organizations where the user account is the sole member (Admin) --> Delete the organization and orphan all the packages as if they directly belonged to  the user
1. Organization where the user account is one of the n admins --> User is deleted. Organization exists with other Admins to manage it -- No brainer :)
1. Organization where the user is the sole Admin but there are collaborators --> User is deleted and all the collaborators are promoted as Admins. We should show this message on the page.

scottbommarito:
@anangaur @chenriksson Currently this implementation does NOT delete orphaned organizations. This is because deleting these organizations requires actually implementing the delete flow for organizations. I don't mind doing that--I've already done most of the work required for it--but I don't see any major problems with having orphaned organizations when the main reason for implementing this feature is GDPR concerns.

scottbommarito:
@anangaur @chenriksson @cristinamanum The implementation has been updated to delete any orphaned organizations, please review again.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -216,8 +204,54 @@ private List<Package> GetPackagesOwnedByUser(User user)
...
+
+                if (memberCount < 2)
+                {
+                    // The user we are deleting is the only member of the organization.
+                    // We should delete the entire organization.
+                    await DeleteAccountImplAsync(organization, requestingUser, orphanPackagePolicy, signature);
+                }
+                else if (memberCount - 1 <= collaborators.Count())
+                {
+                    // All other members of this organization are collaborators, so we should promote them to administrators.
```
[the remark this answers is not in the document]
scottbommarito: @anangaur suggested this--promote all collaborators when the last admin deletes their account.

This isn't an orphaned organization, this is an organization whose last admin left that still has collaborators. I think if the owner of this organization doesn't care enough to give administrator rights to another member, I don't think it's worth detailing a fancy policy.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: @scottbommarito is correct if called under a transaction it appends the changes and commit everything at the end. If not under a transaction each SaveChangesAsync will be committed individually

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -216,8 +204,54 @@ private List<Package> GetPackagesOwnedByUser(User user)
...
             }
         }
+        
+        private async Task RemoveMemberships(User user, User requestingUser, AccountDeletionOrphanPackagePolicy orphanPackagePolicy, string signature)
+        {
+            foreach (var membership in user.Organizations.ToArray())
+            {
+                var organization = membership.Organization;
+                var collaborators = organization.Members.Where(m => !m.IsAdmin).ToList();
+                var memberCount = organization.Members.Count();
```
cristinamanum: I think that these two lines do two queries to db? Maybe we can do one query and do the filter and the count on the same collection?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -249,12 +283,6 @@ private List<Package> GetPackagesOwnedByUser(User user)
             await _userRepository.CommitChangesAsync();
         }
 
-        private async Task RemoveUser(Organization organization)
```
cristinamanum: Is this delete action replaced by other call?

scottbommarito: Yeah, `_userRepository.DeleteOnCommit(organization)` as called by `RemoveUser(User user)` works fine for this.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -216,8 +204,54 @@ private List<Package> GetPackagesOwnedByUser(User user)
...
+            foreach (var membershipRequest in user.OrganizationRequests.ToArray())
+            {
+                user.OrganizationRequests.Remove(membershipRequest);
+            }
+
+            foreach (var transformationRequest in user.OrganizationMigrationRequests.ToArray())
+            {
+                user.OrganizationMigrationRequests.Remove(transformationRequest);
+                transformationRequest.NewOrganization.OrganizationMigrationRequest = null;
+            }
```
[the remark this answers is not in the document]
scottbommarito: I wanted to make sure that there's no dependency from either end, e.g. this user has no `OrganizationMigrationRequest`s and there are no users that have `OrganizationMigrationRequest`s that depend on this user. not sure if the cascade was set up correctly.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Discussed offline with @scottbommarito and this method is not needed anymore. The orgs will be deleted as user entities.

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: done

Issue #2871, opened 2016-02-03
Add AWS S3 support for file storage


fandango-babak:
For users that want to deploy their private NuGet Gallery to AWS, they should be able to leverage S3 as their storage.

maartenba:
Feel free to propose a pull request for this. We are very much tied to Azure right now and for our needs this will remain the case, but if a low impact change can add AWS support we will look into it.

fandango-babak:
I have an S3 backed implementation of IFileStorageService working and am working on the statistics now.

maartenba:
Sweet!

JonCubed:
@fandango-babak how did you end up going with this? I'm looking for an option to use S3, if you need help finishing this off let me know

fandango-babak:
We've been using it in production for months with great success. I will try to get a PR prepared as soon as I can.

JonCubed:
@fandango-babak thanks would be very useful

SeanFarrow:
@fandango-babak, Did you ever get a PR for this? I'm just about to deploy a private Gallery to AWS. If you need a hand with this, feel free to ask!

fandango-babak:
I got it working in my fork but had to move on to other projects before I could prepare a PR. I’m still hoping to prepare a PR this month of next month.
Babak Naffas
Staff Engineer, Software Engineering
Fandango LLC / NBCUniversal
12200 W. Olympic Blvd. Suite 400
Los Angeles, CA  90064
E: babak.naffas@fandango.com<mailto:babak.naffas@fandango.com%0d>
T: (310) 954-0278 x274
F: (310) 954-0292

[cid:FBBC0296-0FE1-408C-AC45-24D538714EF0]

SeanFarrow:
Any chance you could give us a pointer to your fork/specific branch?

Texernie:
I know this is an old issue, but I'm sure many people keep looking for AWS S3 support. I know I am. I am preparing a PR for review of an implementation for this. I need to rework it following the Git workflow, but that shouldn't be too hard.

PR #5799, opened 2018-04-13
Issues #2871 & #5839: Added ability to use AWS S3 as a backing store for the package files.

Texernie:
#2871


Texernie:
I tried to follow the FileSystemFileStorageService and FileSystemService as close as I could. I also patterned some stuff after the existing CloudBlobFileStorageService when needed.

Texernie:
#5839

anangaur:
Thanks for your contribution to NuGet. The Gallery code base currently supports two storage backends: local files and Azure. Both are constantly used and tested by us: we deploy nuget.org on Azure, and our development environments use local files. NuGet Gallery doesn’t support additional infrastructures since we can’t reliably test and maintain those integrations. We do believe that users should be able to use the Gallery with any storage backend and suggest adding an extensibility point that will enable those scenarios. 

Gallery already has support for extensions (see [here](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L677)). Currently it's used for Auditing extensions. This can be extended to support Storage extensions as well. The NuGet team will be more than willing to help you out by reviewing the design/code. We have created [an issue](https://github.com/NuGet/NuGetGallery/issues/5982) to track this extensibility approach.

anangaur:
@Texernie as we are exploring extension based options at https://github.com/NuGet/NuGetGallery/issues/5982, I am closing this PR.

Issue #5113, opened 2017-12-04
Organizations: Auditing of create/edit organizations


chenriksson:
Add auditing for:
- Create new organization (from scratch, or migration)
- Edit organization

PR #5805, opened 2018-04-16
[Organizations] Auditing for organizations

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5113


src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
```diff
@@ -17,23 +17,15 @@ public class UserAuditRecord : AuditRecord<AuditedUserAction>
         public CredentialAuditRecord[] Credentials { get; }
         public CredentialAuditRecord[] AffectedCredential { get; }
         public string AffectedEmailAddress { get; }
+        public string AffectedMemberUsername { get; }
+        public bool AffectedMemberIsAdmin { get; }
```
chenriksson: Maybe add comments above properties that are specific to only org (membership) or user (credential) accounts.

src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
```diff
@@ -45,25 +37,40 @@ public UserAuditRecord(User user, AuditedUserAction action, IEnumerable<Credenti
...
+        public UserAuditRecord(User user, AuditedUserAction action, Credential affected)
+            : this(user, action, new[] { affected })
+        {
+        }
+
+        public UserAuditRecord(User user, AuditedUserAction action, IEnumerable<Credential> affected)
+            : this(user, action)
+        {
+            AffectedCredential = affected.Select(c => new CredentialAuditRecord(c, action == AuditedUserAction.RemoveCredential)).ToArray();
         }
```
chenriksson: Why the refactoring?

scottbommarito: It was really awkward before...most of the constructors were passing `Enumerable.Empty()` when they called `this(...)`. This is the proper way for these constructors to be set up.

src/NuGetGallery.Core/Auditing/UserAuditRecord.cs
scottbommarito: Sure. Will also add to the other fields that are only used for some calls.

scottbommarito:
@chenriksson Shims PR coming soon!

Issue #5634, opened 2018-03-19
Certificates:  enable certificate registration in Gallery


dtivel:
* certificate registration should be available from both user and organization account settings
* filter file extensions to .cer and .crt
* limit file upload size to 10 KB and warn if limit exceeded
* generate fingerprints (SHA-1 and SHA-256) server side
* upload certificate to blob storage
* add tests

PR #5809, opened 2018-04-17
Certificates:  add migration for certificate registration

dtivel:
Resolve https://github.com/NuGet/NuGetGallery/issues/5634.


dtivel:
Will send updated PR.

Issue #5722, opened 2018-04-02
Warning icons on menu and account settings are unclear


scottbommarito:
It should be very clear to users what the warning icon next to their profile means. Currently, users have to follow a trail of icons that eventually lead to some tool-tips that are not well explained.

Users should also be guided on exactly how to fix the issue with their account by the tool-tips and other information displayed.

anangaur:
We should expand the tooltip inside the **Password login** section. Same for the **Login Account** section.

In the password login, replace the checkbox with a button so that it aligns with the warning text.

shishirx34:
- [ ] Remove warning icon upon removing Password login

Issue #5673, opened 2018-03-22
[MSA] Add two factor auth management UI


shishirx34:
- [x] UI work
- [x] Update the User policy to enabled 2FA if a user used 2FA and didn't have 2FA enabled after login.
- [x] Add 2FA was used/not to the session for certificate management?

Issue #5531, opened 2018-02-22
[MSA] Registering new account title mismatch


shishirx34:
Title should be "Registering Microsoft account"
![image.png](https://images.zenhubusercontent.com/5805460995e42b6c5baa9e04/4521dba0-1be5-4f95-97ab-7e4fc790585a)

PR #5827, opened 2018-04-19
[2FA] MSA/AAD integration with 2FA and other fixes

shishirx34:
Addresses:
- Register title mismatch - https://github.com/NuGet/NuGetGallery/issues/5531
- Warning icons - https://github.com/NuGet/NuGetGallery/issues/5722
- Add 2FA auth management UI and other work - https://github.com/NuGet/NuGetGallery/issues/5673

Also,

- Added support for application to know if 2FA was used for the session for certificate management using Claims

UI work screenshot:

**Personal MSA accounts**

Warning icons for 2FA disabled

image

Enabling 2FA warnings go away

image

Disabling 2FA warnings come back.

image

Loggin in with MSA account that has 2FA enabled, but not set on nuget profile, enables it:

image

**AAD Accounts**

That have 2FA enabled by administrator

image

That do not have 2FA enabled by administrator

image


anangaur:
Text review:
* Linked Microsoft account -> Microsoft account 
* Do not change the text for 2FA setting for AAD. Instead have an info banner in the box with the following text:
`Two-factor setting is managed by your Azure AD tenant administrator.`

anangaur:
Also lets be consistent everywhere for AAD. My recommendation is to use `Azure AD` as its short and is a generally known and marketed name. For reference: https://docs.microsoft.com/en-us/azure/active-directory/active-directory-whatis

src/Bootstrap/dist/css/bootstrap-theme.css
```diff
@@ -601,6 +601,22 @@ img.reserved-indicator-icon {
 .page-account-settings .form-group + .form-group .btn {
   margin-top: 20px;
 }
+.page-account-settings .login-account-row {
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display:         flex;
```
chenriksson: nit: spacing?

src/Bootstrap/dist/css/bootstrap-theme.css
```diff
@@ -601,6 +601,22 @@ img.reserved-indicator-icon {
...
+  padding-top: 10px;
+
+  -webkit-box-align: center;
+  -webkit-align-items: center;
+  -ms-flex-align: center;
+          align-items: center;
```
chenriksson: nit: is this spacing intended? Same w/ justify-content below?

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -233,10 +233,10 @@ private void InitCredentialFormatters()
             }
         }
 
-        public virtual async Task CreateSessionAsync(IOwinContext owinContext, AuthenticatedUser user)
+        public virtual async Task CreateSessionAsync(IOwinContext owinContext, AuthenticatedUser user, bool WasMutliFactorAuthenticated = false)
```
chenriksson: nit: spelling, same below ("MutliFactor" => "MultiFactor")

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -166,7 +166,7 @@ public virtual ActionResult SignUp(string returnUrl)
             }
 
             var authenticatedUser = authenticationResult.AuthenticatedUser;
-            
+            bool usedMultiFactorAuthentication = false;
```
chenriksson: nit: don't need explicit set since false is default

src/NuGetGallery.Core/Authentication/NuGetClaims.cs
```diff
@@ -36,5 +36,30 @@ public static class NuGetClaims
...
         /// </summary>
         public const string ExternalLogin = ClaimsDomain + "externallogin";
+
+        /// <summary>
+        /// The claim url for the claim that stores whether or not the user has enabled multi-factor authentication.
+        /// </summary>
+        public const string EnabledMultiFactorAuthentication = ClaimsDomain + "enabledmultifactorauthentication";
+
+        /// <summary>
+        /// The claim url for the claim that stores whether or not the user was multi-factor authenticated for the current session.
```
chenriksson: Took a while to figure out how this was originally set... which I think is a claim from the AADv2 authenticator. Maybe worth mentioning that here?

src/Bootstrap/dist/css/bootstrap-theme.css
shishirx34: file is auto-generated by grunt

src/NuGetGallery/Authentication/AuthenticationService.cs
agr: `camelCase` for argument names, not `PascalCase`.

src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -268,6 +268,18 @@ private void InitCredentialFormatters()
                 ClaimsExtensions.AddBooleanClaim(claims, NuGetClaims.ExternalLogin);
             }
 
+            if (user.User.EnableMultiFactorAuthentication)
```
scottbommarito: nit; i would suggest renaming `AuthenticatedUser user` to `AuthenticatedUser authUser` so that this is statement here is less redundant and it's clearer what type `user` is.

src/NuGetGallery.Core/Authentication/NuGetClaims.cs
shishirx34: Ideally this claim is independent of where you get the original value from. You could have a different way of doing a 2-FA(even for nuget logins). I can probably add that information where we add this claim to the session.

src/NuGetGallery/Extensions/PrincipalExtensions.cs
```diff
@@ -97,6 +97,68 @@ public static bool HasPasswordLogin(this IPrincipal self)
             return ClaimsExtensions.HasBooleanClaim(identity, NuGetClaims.PasswordLogin);
         }
 
+        /// <summary>
+        /// Determine if the current user has multi-factor authentication enabled.
+        /// </summary>
+        /// <param name="self">Current user principal.</param>
+        /// <returns>True if user has multi-factor authentication enabled, false otherwise.</returns>
+        public static bool HasMultiFactorAuthenticationEnabled(this IPrincipal self)
```
scottbommarito: You can simplify these into a `HasBooleanClaim` method that takes in an `IPrincipal self` and `string claimUrl`.

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
```diff
@@ -249,7 +249,7 @@ public class TheSignInAction : TestContainer
 
                 var controller = GetController<AuthenticationController>();
                 GetMock<AuthenticationService>()
-                    .Setup(a => a.CreateSessionAsync(controller.OwinContext, authUser))
+                    .Setup(a => a.CreateSessionAsync(controller.OwinContext, authUser, false))
```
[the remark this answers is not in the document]
shishirx34: in the `Setup` method, the named arguments are invalid, since it uses `Expression`.

`An expression tree may not contain named arguments`

src/NuGetGallery/Views/Users/_UserAccountChangeExternalCredential.cshtml
```diff
@@ -8,27 +8,104 @@
...
         @<text>Login Account</text>,
-        @<text>@cred.TypeCaption (@cred.Identity)</text>,
-            @<text>
-                <p>You can use this account to sign in to @(Config.Current.Brand).</p>
-                            
-                @using (Html.BeginForm("LinkOrChangeExternalCredential", "Users", FormMethod.Post, new { aria_label = "Link or change " + cred.TypeCaption }))
+        @<text>@cred.TypeCaption | 
+            @if (CredentialTypes.IsMicrosoftAccount(cred.Type))
+            {
+                if (Model.User.EnableMultiFactorAuthentication)
```
scottbommarito: nit:
```
var is2faEnabled = Model.User.EnableMultiFactorAuthentication;
<span>
    <i class="ms-Icon ms-Icon--@(is2faEnabled ? "CheckMark" : "Warning") @(is2faEnabled ? "checkmark" : "warning")-icon" aria-hidden="true"></i> 2FA @(is2faEnabled ? "enabled" : "disabled")
</span>
```

(similar thing below)

src/NuGetGallery/Views/Users/_UserAccountChangeExternalCredential.cshtml
```diff
@@ -8,27 +8,104 @@
...
+                    @Html.Hidden("enableMultiFactor", !Model.User.EnableMultiFactorAuthentication);
+                    <div class="login-account-row">
+                        @if (Model.User.EnableMultiFactorAuthentication)
+                        {
+                            <span>
+                                Two-factor authentication is currently enabled <i class="ms-Icon ms-Icon--CheckMark checkmark-icon" aria-hidden="true"></i>
+                            </span>
+                            <span class="col-sm-3">
+                                <button class="btn btn-danger btn-block" type="submit"
+                                        data-confirm="This action will disable 2FA making your account less secure. Are you sure you want to disable this setting?">
```
scottbommarito: `will disable 2FA making` -> `will disable 2FA, making`

tests/NuGetGallery.Facts/Authentication/AuthenticationServiceFacts.cs
```diff
@@ -633,6 +633,47 @@ public class TheCreateSessionAsyncMethod : TestContainer
...
+                passwordConfigMock
+                    .Setup(x => x.IsLoginDiscontinued(authUser))
+                    .Returns(false);
+
+                GetMock<IContentObjectService>()
+                    .Setup(x => x.LoginDiscontinuationConfiguration)
+                    .Returns(passwordConfigMock.Object);
+
+                // Act
+                await Get<AuthenticationService>().CreateSessionAsync(context, authUser, WasMutliFactorAuthenticated: true);
```
scottbommarito: should you test `wasMultiFactorAuthenticated` with both `true` and `false`?

src/NuGetGallery/Extensions/PrincipalExtensions.cs
shishirx34: This is meant for calling from cshtml, it would be better to have it descriptive enough.

scottbommarito: I mean make `HasMultiFactorAuthenticationEnabled` call the `HasBooleanClaim` described above.

Gets rid of the redundant null checks in each method and makes this vastly simpler.

src/NuGetGallery/Controllers/AuthenticationController.cs
shishirx34: turns out this is needed, compiler otherwise on the usage of uninitialized variable.

Issue #5779, opened 2018-04-09
TryGetDownloadCountForPackageRegistration can throw a modified enumerator exception


joelverhagen:
This results in an HTTP 500 for the user. This has happened at least 6 times in the last 7 days on PROD.

AI query
```
requests
| where timestamp > ago(90d)
| where name contains "PushPackageApi"  or name contains "verify" 
| where name startswith "PUT" or name startswith "POST"
| summarize sum(itemCount) by bin(timestamp, 1d), tostring(resultCode)

requests
| where timestamp > ago(7d)
| where name contains "PushPackageApi"  or name contains "verify" 
| where name startswith "PUT" or name startswith "POST"
| where resultCode == 500
| join kind=inner (
   exceptions 
) on $left.operation_Id == $right.operation_Id
| where innermostMessage == "Collection was modified; enumeration operation may not execute."
| project problemId
```

This is where the exception is being thrown from:
https://github.com/NuGet/NuGetGallery/blob/4607a1d3d8475c22e1e474fe32c2bfeab39c049b/src/NuGetGallery/Services/CloudDownloadCountService.cs#L53

Probably need some locking.

joelverhagen:
I've merged the fix to dev branch. It is not validated anywhere except my devbox.

PR #5830, opened 2018-04-20
Fix collection modified exception which causes rare HTTP 500s

joelverhagen:
Had some free time so I wanted to fix some intermittent 500s related to getting download counts on an ID. This PR addresses #5779.

Things done:

1. Use `ConcurrentDictionary` in CloudDownloadCountService instead of `Dictionary`. This allows concurrent reads (via HTTP requests) and writes (via a background refresh every 15 minutes)
1. Don't lowercase. Using `StringComparer.OrdinalIgnoreCase`. This should reduce allocations (not measured though).
1. Add unit tests, in particular one that repro'd the race condition on `Dictionary`.

Performance analysis:

Version | Operation | Duration
-- | -- | --
`Dictionary` | `Refresh` | 17.13 seconds
`Dictionary` | `TryGetDownloadCountForPackageRegistration` | 0.0009 ms
`ConcurrentDictionary` | `Refresh` | 17.70 seconds
`ConcurrentDictionary` | `TryGetDownloadCountForPackageRegistration` | 0.0011 ms

For `Refresh`, I copied the `downloads.v1.json` from PROD to my own storage account and ran gallery code from my dev box. 100 iterations. Every 10 iterations I would refresh from scratch.

For `TryGetDownloadCountForPackageRegistration`, I ran all PROD IDs through `TryGetDownloadCountForPackageRegistration`, which 176,157 invocations of the method. 245 iterations. Every 49 iterations I re-initialized the server (new `Refresh`).

I thing the stability of `ConcurrentDictionary` is worth the minor performance decrease. We can consider an alternative approach if you guys think its worth it.

### DIFF
--- a/src/NuGetGallery/Services/CloudDownloadCountService.cs
+++ b/src/NuGetGallery/Services/CloudDownloadCountService.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Services/CloudDownloadCountService.cs
```diff
@@ -127,78 +119,102 @@ public void Refresh()
...
                 {
-                    try
+                    if (blobStream == null)
                     {
-                        jsonReader.Read();
+                        return;
+                    }
 
-                        while (jsonReader.Read())
+                    using (var jsonReader = new JsonTextReader(new StreamReader(blobStream)))
```
skofman1: please make sure all the IDisposables  will work well with each other (nothing missing, no double dispose that throws..)

PR #5833, opened 2018-04-20
Enable certificate registration

dtivel:
Resolve https://github.com/NuGet/NuGetGallery/issues/5634 and https://github.com/NuGet/NuGetGallery/issues/5776.

New Certificates section in both user and organization account settings:

certificates1

Attempting to upload a non .cer file:

noncer

Attempting to upload a PFX renamed to CER:

pfxascer

After registering certificates:

certificates2

Managing required signer for owned packages:

packages

Viewing signing information in private package history view:

versionhistory1

Hovering over the badge icon:

versionhistory2

Requiring 2FA on certain actions ([spec](https://github.com/NuGet/Home/wiki/Register-package-signing-certificates#2-fa-requirement-for-managing-certificates), [task](https://github.com/NuGet/NuGetGallery/issues/5775)) is blocked by @shishirx34's open [PR](https://github.com/NuGet/NuGetGallery/pull/5827).


src/NuGetGallery/Filters/ValidateAjaxAntiForgeryTokenAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Web.Helpers;
+using System.Web.Mvc;
+
+namespace NuGetGallery.Filters
+{
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
+    public sealed class ValidateAjaxAntiForgeryTokenAttribute : FilterAttribute, IAuthorizationFilter
```
skofman1: We were using ValidateAntiForgeryTokenAttribute with Ajax requests already. Why do we need this attribute?

dtivel: The `ValidateAntiForgeryTokenAttribute` only works with [cookies and forms](https://github.com/aspnet/AspNetWebStack/blob/v3.2.3/src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs#L160), and HTML forms appear to be supported formally only with [POST and GET requests](http://www.w3.org/MarkUp/html-spec/html-spec_8.html#SEC8.2).

ASP.NET documentation [recommends a solution](https://docs.microsoft.com/en-us/aspnet/web-api/overview/security/preventing-cross-site-request-forgery-csrf-attacks#anti-csrf-and-ajax) which I [implemented](https://github.com/NuGet/NuGetGallery/pull/5833/files#diff-8603e72ad0bb3773838748c6019f08eaR44) through the `VerifyAjaxAntiForgeryTokenAttribute`.  This solution applies generally to all Ajax requests irrespective of HTTP verb.

I investigated this again and found that the existing `ValidateAntiForgeryTokenAttribute` can work.  If the Ajax `contentType` and `processData` properties are true (default) then the form data (which contains the anti-forgery token) is URL encoded.  This is a bit strange sending form data for a DELETE request, and possibly unsupported, but it works.  Thus, I've removed the `VerifyAjaxAntiForgeryTokenAttribute` and used the existing `ValidateAntiForgeryTokenAttribute` instead.

src/NuGetGallery/Services/CertificateValidator.cs
```diff
@@ -0,0 +1,165 @@
...
+using System;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Web;
+
+namespace NuGetGallery
+{
+    public sealed class CertificateValidator : ICertificateValidator
```
chenriksson: Seems like something we should be able to get from the `System.Security.Cryptography.OpenSsl` package, but not seeing public APIs for this.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -707,5 +706,113 @@ private void ValidatePackageTitle(PackageMetadata packageMetadata)
...
+            {
+                throw new ArgumentNullException(nameof(registration));
+            }
+
+            var isCommitRequired = false;
+
+            string previousRequiredSigner = null;
+            string newRequiredSigner = null;
+
+            if (signer == null)
```
chenriksson: Why does this overload of `SetRequiredSignerAsync` not require signer, but the other does?

Should the API reflect the difference? Or at least a comment to clarify?

src/NuGetGallery/ViewModels/ListPackageItemRequiredSignerViewModel.cs
```diff
@@ -0,0 +1,150 @@
...
+                throw new ArgumentNullException(nameof(securityPolicyService));
+            }
+
+            var requiredSigner = package.PackageRegistration?.RequiredSigners.FirstOrDefault();
+
+            RequiredSigner = Convert(requiredSigner);
+
+            var owners = package.PackageRegistration?.Owners ?? Enumerable.Empty<User>();
+
+            if (HasSingleOwner || !owners.Any())
```
chenriksson: nit: `|| owners.Empty()`?

src/NuGetGallery/ViewModels/ListPackageItemRequiredSignerViewModel.cs
```diff
@@ -0,0 +1,150 @@
...
+            else
+            {
+                var currentUserIsOwner = owners.Contains(currentUser);
+                var currentUserHasControl = securityPolicyService.IsSubscribed(
+                    currentUser,
+                    ControlRequiredSignerPolicy.PolicyName);
+                var noOwnerHasControl = !owners.Any(
+                    owner => securityPolicyService.IsSubscribed(owner, ControlRequiredSignerPolicy.PolicyName));
+
+                CanEditRequiredSigner = currentUserIsOwner && (currentUserHasControl || noOwnerHasControl);
```
chenriksson: Does `currentUserIsOwner` account for allowing admins of an organization owner to edit the required signer?

src/NuGetGallery/ViewModels/ListPackageItemRequiredSignerViewModel.cs
```diff
@@ -0,0 +1,150 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using NuGetGallery.Security;
+
+namespace NuGetGallery
+{
+    public sealed class ListPackageItemRequiredSignerViewModel : ListPackageItemViewModel
```
chenriksson: This adds additional queries to an already expensive page. They don't look too expensive, but maybe it's worth measuring latency overhead for a large account (Microsoft or functional test account)?

Some performance work was done on DB queries, but we stopped short of paging since the query optimization was enough at the time.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -252,8 +265,8 @@ private ActionResult HandleOrganizationMembershipRequestView(HandleOrganizationM
 
             var currentUser = GetCurrentUser();
 
-            if (account == null || 
-                (currentUser.Username != memberName && 
+            if (account == null ||
+                (currentUser.Username != memberName &&
```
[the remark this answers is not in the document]
dtivel: I'll leave it as-is.  I don't have bandwidth to test this change as part of this PR.

src/NuGetGallery.Core/Auditing/CertificateAuditRecord.cs
```diff
@@ -0,0 +1,30 @@
...
+    public sealed class CertificateAuditRecord : AuditRecord<AuditedCertificateAction>
+    {
+        public string Thumbprint { get; }
+        public string HashAlgorithm => "SHA-256";
+
+        public CertificateAuditRecord(AuditedCertificateAction action, string thumbprint)
+            : base(action)
+        {
+            if (string.IsNullOrEmpty(thumbprint) ||
+                thumbprint.Length != 64) // Did the thumbprint hash algorithm change?
```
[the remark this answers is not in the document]
shishirx34: why care about the change of hash algorithm(different length) during auditing? Shouldn't auditing be independent of any implementation details?

src/NuGetGallery.Core/Auditing/CertificateAuditRecord.cs
```diff
@@ -0,0 +1,30 @@
...
+            {
+                throw new ArgumentException(CoreStrings.ArgumentCannotBeNullOrEmpty, nameof(thumbprint));
```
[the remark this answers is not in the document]
shishirx34: throw `ArgumentNullException` and ignore the length check if possible.

src/NuGetGallery/Services/PackageService.cs
dtivel: This overload is called by `AccountsController` when an account registers one certificate and the `AutomaticallyOverwriteRequiredSignerPolicy` is in effect.  The account should automatically overwrite the required signer for all package registrations it is an owner of, so `signer` cannot be null.

The other overload is called in the manage packages page on a per package registration basis.  When `signer` is null there, the required signer for that package registration is nulled out.

src/NuGetGallery/ViewModels/ListPackageItemRequiredSignerViewModel.cs
dtivel: There is no `Empty()` (extension) method available.  `Any()` is built-in, and I don't think it's worth writing and testing an extension method wrapper around `Any()`.

src/NuGetGallery/ViewModels/ListPackageItemRequiredSignerViewModel.cs
dtivel: I've rewritten this code because it doesn't account for the current user being an admin of an organization which is an owner..

src/NuGetGallery.Core/Certificates/CertificateFile.cs
```diff
@@ -0,0 +1,96 @@
...
+                return GetThumbprint(stream, hashAlgorithm);
+            }
+        }
+
+        private static string GetThumbprint(MemoryStream stream, HashAlgorithm hashAlgorithm)
+        {
+            var buffer = stream.GetBuffer();
+
+            hashAlgorithm.TransformBlock(buffer, 0, buffer.Length, outputBuffer: null, outputOffset: 0);
+            hashAlgorithm.TransformFinalBlock(EmptyBuffer, inputOffset: 0, inputCount: 0);
```
shishirx34: can you add a comment on the requirement of `TransformFinalBlock` for posterity?

src/NuGetGallery/Scripts/gallery/certificates.js
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+
+        function addCertificateAsync(data) {
+            clearErrors();
+
+            $.ajax({
+                method: 'POST',
+                url: _addCertificateUrl,
+                data: data,
```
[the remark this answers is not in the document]
dtivel: It does have an anti-forgery token.

See [here](https://github.com/NuGet/NuGetGallery/pull/5833/files#diff-3a276a7bbfe9236dbefd071cb73893beR58) and [here](https://github.com/NuGet/NuGetGallery/pull/5833/files#diff-fab023128f85c20a5525964a903f8697R24).

src/NuGetGallery.Core/Services/CorePackageService.cs
```diff
@@ -222,6 +229,114 @@ public virtual Package FindPackageByIdAndVersionStrict(string id, string version
             return package;
         }
 
+        public virtual PackageRegistration FindPackageRegistrationById(string packageId)
```
shishirx34: I think this is redundant, any particular reason on using `CorePackageService` and not `PackageService`? These methods have been implemented in there, you can expand on the `PackageService` for your stuff. Or better yet, create a separate service for `Certificates` and reference `PackageService` in there with addition of required methods for updating package registrations in the `PackageService`.

src/NuGetGallery/Scripts/gallery/certificates.js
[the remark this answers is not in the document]
scottbommarito: My confusion was that I don't see `window.nuget.addAjaxAntiForgeryToken` here

src/NuGetGallery.Core/Services/ICorePackageService.cs
```diff
@@ -42,5 +42,48 @@ public interface ICorePackageService
...
+        /// Determines if the certificate with specified thumbprint is valid for signing the specified package ID.
+        /// </summary>
+        /// <param name="packageId">A package ID.</param>
+        /// <param name="thumbprint">A certificate thumbprint.</param>
+        /// <returns>A flag indicating whether the certificate is acceptable for signing.</returns>
+        /// <exception cref="ArgumentException">Thrown if <paramref name="packageId" /> is <c>null</c>
+        /// or empty or if the package registration does not exist.</exception>
+        /// <exception cref="ArgumentException">Thrown if <paramref name="thumbprint" /> is <c>null</c>
+        /// or empty.</exception>
+        bool IsAcceptableSigningCertificate(string packageId, string thumbprint);
```
shishirx34: Why not put these in `src\NuGetGallery\Services\IPackageService.cs` and update `PackageService` instead of `ICorePackageService`?

src/NuGetGallery/Scripts/gallery/certificates.js
```diff
@@ -0,0 +1,162 @@
...
+                case "abort":
+                    displayErrors(["The operation was aborted. Please try again."]);
+                    break;
+                default:
+                    displayErrors(model.responseJSON);
+                    break;
+            }
+
+            if (fullResponse.status >= 500) {
+                displayErrors(["There was a server error."]);
```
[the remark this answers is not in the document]
dtivel: We don't [elsewhere](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Scripts/gallery/async-file-upload.js#L182) and the Contact link is already prominently at the bottom of every page.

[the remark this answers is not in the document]
scottbommarito: We definitely have error messages in a couple places around the site that ask the user to contact support but given that package upload has this same message it's fine.

src/NuGetGallery.Core/Auditing/CertificateAuditRecord.cs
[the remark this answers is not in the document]
dtivel: The hash algorithm should be logged with the thumbprint.  A [few lines up](https://github.com/NuGet/NuGetGallery/pull/5833/files#diff-c3c6975f09b0b15a996f2136a6d7c55aR11) you can see that it is.

Ideally, the hash algorithm would be passed as an argument, but because we don't propagate the hash algorithm from the [source](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery.Core/Entities/Certificate.cs#L14), the caller doesn't know it either.  It's not worth doing the work to propagate the hash algorithm now.

src/NuGetGallery.Core/Auditing/CertificateAuditRecord.cs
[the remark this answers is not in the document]
dtivel: @scottbommarito, yes.
@shishirx34, no.  The `IsNullOrEmpty(...)` check is reasonable, as is a sanity check on the thumbprint length.

src/NuGetGallery.Core/Certificates/CertificateFile.cs
dtivel: MSDN does an [excellent job](https://msdn.microsoft.com/en-us/library/system.security.cryptography.hashalgorithm.transformfinalblock(v=vs.110).aspx#Anchor_2) explaining the call `TransformFinalBlock(...)`, and I won't repeat it here in comments for the indolent.

src/NuGetGallery.Core/Services/ICorePackageService.cs
dtivel: Signature validation jobs use the minimal `ICorePackageService` interface vs. the full-featured `IPackageService` interface.  I don't see a good reason to implement this in `IPackageService`.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -57,6 +63,37 @@
...
+                    cache: false,
+                    data: window.nuget.addAjaxAntiForgeryToken({}),
+                    complete: function (xhr, textStatus) {
+                        switch (xhr.status) {
+                            case 200:
+                            case 409:
+                                break;
+
+                            default:
+                                break;
```
shishirx34: switch statements do nothing..

src/NuGetGallery.Core/Services/CorePackageService.cs
```diff
@@ -232,6 +347,16 @@ protected IQueryable<Package> GetPackagesByIdQueryable(string id)
...
         }
 
+        private bool HasAnyCertificate(User user)
+        {
+            return user.UserCertificates.Any();
+        }
+
+        private bool CanUseCertificate(User user, string thumbprint)
+        {
+            return user.UserCertificates.Any(uc => uc.Certificate.Thumbprint == thumbprint);
```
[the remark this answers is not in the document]
loic-sharma: SQL server is case insensitive :)

Issue #5816, opened 2018-04-18
Gallery seems to reuse the same instance of MailSender between multiple threads


agr:
https://github.com/NuGet/NuGetGallery/blob/67373eb24d14207969ecb80f0c9c20a8cc17f7fd/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L268-L309

Here we create a `Lazy<IMailSender>` object (one per process) and pass `c => mailSenderThunk.Value` as a factory for generating `InstancePerLifetimeScope` objects. `mailSenderThunk.Value` on first evaluation will run the big chunk of code that creates `MailSender` and subsequent calls will reuse the instance it created.
This should only be done if we are sure that `MailSender` is thread-safe.

skofman1:
To see logs in AI:
exceptions
| where outerMessage == "An asynchronous call is already in progress. It must be completed or canceled before you can call this method." and timestamp > ago(90d)

cristinamanum:
Verified by checking the AI logs for the mentioned exception.

PR #5835, opened 2018-04-20
Create IMailSender per request and not per instance

skofman1:
Fixes: https://github.com/NuGet/NuGetGallery/issues/5816

There's no "expensive" initializations in the ctors, so no problem in initializing per request.


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -265,8 +265,7 @@ protected override void Load(ContainerBuilder builder)
                     .Refresh());
             }
 
-            var mailSenderThunk = new Lazy<IMailSender>(
-                () =>
+            Func<MailSender> mailSenderFacotry = () =>
```
joelverhagen: `Factory`

agr: C# supports methods declared inside methods (forgot what's the "official" name), should we declare one instead of a delegate?

agr: [Local functions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions)

Issue #5751, opened 2018-04-02
[ML] Recommendations for package


shishirx34:
This is the tracking item for adding support on NuGet gallery package details page some sort of recommendation for similar packages, possible usage using Machine learning.

For an MVP consider generating a static list of recommendation packages per package which will be integrated with the gallery UI.

jamesqo:
Hi Shishir! Sorry for the delay in reply.

I'm a student, so I actually have been really busy with schoolwork for the last couple of weeks and forgot about this; I haven't begun the integration work yet. I also thought this was basically a one-man effort so I wasn't holding anyone up, but I'm really glad that you've brought this up with management and are offering your help (thank you).

The good news is that it should be relatively easy to export the recommendations from the Python program to the gallery, e.g. just export a matrix of package indices in CSV format. There is only one major concern: the ML algorithm builds an `n x n` matrix where `n` is the number of instances. I only ran the algorithm on 18k packages in my app, but considering that there are 100k packages in total hosted on the gallery, it's possible that we may run into space/time issues with a 10b-item matrix. (I think there could be some ways to avert this but I'll have to think some more.)

Regarding telemetry:

- How would I store the data for each interesting event such as a user clicking on a recommended package, or search for a different package while on the page of one package? (Could you link me to the code that stores/retrieves NuGet API metadata from a DB?)
- Can I create a new API (e.g. a new endpoint in the index.json) around this, and then use that API from the code that takes that telemetry into account? I think this would simplify the work by splitting it up into 2 tasks: implementing the telemetry vs. using the telemetry.

Also just to clarify, the recommendations will initially be static (so implementing telemetry comes later). I'll have to think some more about how user behavior would be used to weigh/disweigh recommendations.

jamesqo:
Actually, nvm about the first problem: it may not take that much memory to store a 10b+ element matrix if most of the elements are 0 (a sparse matrix can be used).

scottbommarito:
Hey @jamesqo,

That sounds really cool! I have a couple comments.

The gallery actually has around 1.2 million (listed) packages, not just 100k. There's an additional 300k-500k unlisted packages which I don't think we'd want to include in this (they were unlisted intentionally to make themselves less discoverable). Do you think the additional `12 * 12` factor would increase the size of the matrix too much?

[...]

PR #5841, opened 2018-04-25
Adds machine learning-generated recommendations to each package's page

jamesqo:
Have not built or tested yet, please do not merge.

Contributes to https://github.com/NuGet/NuGetGallery/issues/5751 (since this is hopefully just the initial stage of the feature, I won't write "fixed" just yet :wink:)

/cc @shishirx34 @scottbommarito


src/NuGetGallery/Services/PackageRecommendationService.cs
```diff
@@ -0,0 +1,57 @@
...
+            foreach (byte b in bytes)
+            {
+                sb.Append(b.ToString("x2"));
+            }
+            return sb.ToString();
+        }
+
+        private static RecommendedPackageViewModel CreateViewModel(string recommendationId)
+        {
+            string galleryPageUrl = $"https://www.nuget.org/packages/{recommendationId}";
```
jamesqo: What's a clean way to get the gallery page url / icon url from the package ID here?

src/NuGetGallery/ViewModels/RecommendedPackageViewModel.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery
+{
+    public class RecommendedPackageViewModel
```
jamesqo: Should I create a custom view model for this or use the existing `Package` entity? (If using `Package` would be better, is there a succinct way to get a `Package` object from the id?)

src/NuGetGallery/Services/PackageRecommendationService.cs
scottbommarito: Not necessary to do here. In the view you can use `Url.Package(recommendation.Id)` to get the link. You can remove this field from the model entirely.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -481,6 +481,19 @@
             </div>
         </article>
         <aside class="col-sm-3 package-details-info">
+            <h2>Related</h2>
+            <ul class="list-unstyled">
+                @foreach (var recommendation in Model.RecommendedPackages)
```
scottbommarito: Can we have a picture of how this looks?

src/NuGetGallery/ViewModels/RecommendedPackageViewModel.cs
scottbommarito: An alternative to making a new model for this would be to use the existing `ListPackageItemViewModel`.

You could make your `PackageRecommendationService` use an `IPackageService` to get the `Package` and then create a `ListPackageItemViewModel` using it.

A benefit of this is that the existing model already has everything that you would ever want to display a package, including the `IconUrl`.

src/NuGetGallery/Services/PackageRecommendationService.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            _reportService = reportService;
+        }
+
+        public async Task<IEnumerable<RecommendedPackageViewModel>> GetRecommendedPackagesAsync(Package package)
+        {
+            string encodedId = GetHexadecimalString(Encoding.UTF8.GetBytes(package.Title));
+            string reportName = $"{encodedId}.json";
+
+            var report = await _reportService.Load(reportName);
```
scottbommarito: Because we are now using these `StatisticReport` objects to represent more than just statistics, we should rename the class used here to be just `Report`.

src/NuGetGallery/Services/PackageRecommendationService.cs
```diff
@@ -0,0 +1,57 @@
...
+            var reportJson = JObject.Parse(report.Content);
+
+            string targetId = reportJson["Id"];
+            Debug.Assert(targetId == package.Title);
```
scottbommarito: I think it would be better if it just logged an error and returned an empty list rather than throwing. If this happens on our production environment we want to know it happened.

Also, this isn't a very safe string comparison. We usually use `Equals` with `StringComparison.OrdinalIgnoreCase`.

src/NuGetGallery/Services/PackageRecommendationService.cs
```diff
@@ -0,0 +1,57 @@
...
+            _reportService = reportService;
+        }
+
+        public async Task<IEnumerable<RecommendedPackageViewModel>> GetRecommendedPackagesAsync(Package package)
+        {
+            string encodedId = GetHexadecimalString(Encoding.UTF8.GetBytes(package.Title));
+            string reportName = $"{encodedId}.json";
+
+            var report = await _reportService.Load(reportName);
+            var reportJson = JObject.Parse(report.Content);
```
scottbommarito: Could you create a class with the expected structure and deserialize the content field into that object?

src/NuGetGallery/Services/PackageRecommendationService.cs
scottbommarito: There doesn't appear to be any error handling here. I think ideally if at any point in the process of fetching recommendation blobs there was a failure, we'd just want to log the error and return an empty list.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
scottbommarito: @anangaur any idea on how recommended packages should be shown on the package details page?

jamesqo: @scottbommarito The basic layout I have in mind is it should go right above the "Info" section in an unstyled list, with the icon of each package to the left of the anchor text.

src/NuGetGallery/Services/PackageRecommendationService.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            string targetId = recommendedPackages.Id;
+            Debug.Assert(string.Equals(targetId, package.Title, StringComparison.OrdinalIgnoreCase));
+
+            var recommendationIds = recommendedPackages.Recommendations;
+            return recommendationIds.Select(
+                id => new ListPackageItemViewModel(
+                    _packageService.FindAbsoluteLatestPackageById(id),
+                    // We don't need to know about the user's permissions for any of the recommended packages.
+                    currentUser: null));
```
scottbommarito: I would pass in the actual current user--this parameter is used to display package controls (edit/manage owners/unlist) for owners of the package when it is displayed.

src/NuGetGallery/Services/PackageRecommendationService.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public async Task<IEnumerable<ListPackageItemViewModel>> GetRecommendedPackagesAsync(Package package)
+        {
+            string reportName = GetReportName(package);
+            StatisticsReport report;
+            try
+            {
+                report = await _reportService.Load(reportName);
+            }
+            catch (StatisticsReportNotFoundException ex)
```
scottbommarito: Because we are no longer using this service exclusively for stats, we should rename this exception to just `ReportNotFoundException`.

jamesqo: Should I also rename `StatisticsReport` to `Report`? Seems uncomfortably short for the name of a .NET class.

src/NuGetGallery/Services/PackageRecommendationService.cs
jamesqo: OK

src/NuGetGallery/Services/PackageRecommendationService.cs
scottbommarito: Yeah. I agree that it's short but I can't think of a longer name that would be better.

scottbommarito:
@jamesqo Can you post a screenshot of the new layout?

src/NuGetGallery/Services/ReportNotFoundException.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Exception thrown when a report is not found.
```
scottbommarito: nit: `when a <see cref="ReportBlob"/> is not found by the <see cref="ReportService"/>

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -481,6 +481,17 @@
             </div>
         </article>
         <aside class="col-sm-3 package-details-info">
+            @if (Model.RecommendedPackages.Any())
```
scottbommarito: This is in the `aside`? How does the package list look when it's only 3 columns?

src/NuGetGallery/Services/LocalReportService.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using System.Web;
+
+namespace NuGetGallery
+{
+    public class LocalReportService : IReportService
+    {
+        private static readonly string ProjectDirectory = HttpRuntime.AppDomainAppPath;
+        private static readonly string ReportsDirectory =
```
jamesqo: Not sure if there's a cleaner way to get this path.

src/NuGetGallery/Services/PackageRecommendationService.cs
```diff
@@ -0,0 +1,85 @@
...
+                var sb = new StringBuilder(capacity: bytes.Length * 2);
+                foreach (byte b in bytes)
+                {
+                    sb.Append(b.ToString("x2"));
+                }
+                return sb.ToString();
+            }
+
+            string encodedId = GetHexadecimalString(
+                Encoding.UTF8.GetBytes(package.PackageRegistration.Id));
```
jamesqo: Will disable encoding once the rest of changes are approved

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
jamesqo: > How does the package list look when it's only 3 columns?

For this PR at least, I decided not to limit it to 3 entries with a "+ See more" link as I'd originally wanted to. I'm not sure how to update the HTML in response to the user clicking the link, or how to animate the link sliding down once the user clicks it. For now, I'm just showing all 5 recs.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -481,6 +481,31 @@
...
+            {
+                <h2>Related</h2>
+                <ul class="list-unstyled">
+                    @foreach (var package in Model.RecommendedPackages)
+                    {
+                        <li class="recommendation-item" style="margin-bottom: 10px; position: relative;">
```
jamesqo: @scottbommarito Be warned: A lot of the HTML code from here on is quite messy (I was prototyping so I used inline style attributes everywhere). I'm not sure in which CSS file to put the classes that will hold these styles.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -481,6 +481,31 @@
...
+                               style="display: inline-block; float: left; margin-right: 10px;"
+                               target="_blank"><img aria-hidden="true" alt=""
+                                                    width="28" height="28"
+                                                    class="recommendation-icon"
+                                                    style="display: inline-block;"
+                                                    src="@(PackageHelper.ShouldRenderUrl(package.IconUrl) ? package.IconUrl : Url.Absolute("~/Content/gallery/img/default-package-icon.svg"))"
+                                                    /></a>
+                            <a href="@Url.Package(package)"
+                                class="recommendation-anchor"
+                                style="word-wrap: break-word; position: absolute; bottom: 0;"
```
jamesqo: selfnote: word-wrap may no longer be necessary since the div seems to have stopped wrapping

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -481,6 +481,31 @@
...
+                                target="_blank">@package.Id</a>
+                            <br clear="all" />
```
jamesqo: selfnote: use `:after` trick to remove this from html

jamesqo:
Here's a pic of the UI everyone:

related-ui

jamesqo:
@scottbommarito @anangaur I ran into a tiny quirk with the UI. If the ID of a recommended package is too long, then instead of wrapping it will add some extra whitespace to the left of the page:

related-ui-2

Can you tell from the HTML why it's doing this?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
scottbommarito: I was referring to the fact that this is in a section of the page that occupies a fourth of the full width of the page (3 columns) instead of the main section of the page (9 columns).

I expected to see the recommended packages in the main section, somewhere underneath "Release Notes", using the `_ListPackage` partial.

```
@foreach (var package in Model.RecommendedPackages)
{
    @Html.Partial("_ListPackage", package)
}
```

This looks good as well, but I'd prefer that
1 - it's underneath the Statistics of the aside, instead of first section of the side
2 - [given that your UI is so similar to what we show for the list of owners](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml#L668), that we reuse the code for that, e.g.
```
<ul class="list-unstyled owner-list">
@foreach (var package in Model.RecommendedPackages)
{
    <li>
        <a href="@Url.Package(package.Id)" title="@package.Id">
            <img class="package-icon img-responsive" aria-hidden="true" alt=""
                src="@(PackageHelper.ShouldRenderUrl(Model.IconUrl) ? package.IconUrl : Url.Absolute("~/Content/gallery/img/default-package-icon.svg"))"
                @ViewHelpers.ImageFallback(Url.Absolute("~/Content/gallery/img/default-package-icon-256x256.png")) />
        </a>
        <a href="@Url.Package(package.Id)" title="@package.Id">@package.Id</a>
    </li>
}
</ul>
```

jamesqo: @scottbommarito 

1- Ehhhh... on the one hand I'd agree with you because not all of the recommendations are super top-notch, and I wouldn't want to click on a package and see a bunch of irrelevant packages sitting in the top-left of the package page when it loads. On the other, I don't think the feature would be very successful if it wasn't the first thing people saw. Devs are not going to bother to scroll down to see recommendations, especially if they're not 100% relevant all the time.

2- Good idea. Working on that now.

jamesqo: @scottbommarito Copying-and-pasting the HTML for the owners list didn't quite work out; the anchor text was not wrapping properly for recommendations with long IDs (the package ID text would either be moved to the next line, or it would go off the screen, etc). I finally came up with a solution, but I had to modify the LESS a bit.

jamesqo:
@scottbommarito I addressed all of your feedback. The only thing is that I'm worried this feature might fail to take off if we place the recommendations at the bottom of the page... [more here](https://github.com/NuGet/NuGetGallery/pull/5841#discussion_r189144249).

I finally took care of a word-wrapping issue I was running into with long IDs through a CSS hack, so the recommendations now look like this

nuget-website

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
scottbommarito: I agree that if the feature is too hard to find, it's going to be difficult for people to adopt using it, but I think the Info and Statistics about the package are more important than the recommendations. The info section has important information about the package and links to edit/report it. The statistics section is important because it shows download counts that can be used to determine whether or not the package is popular and well-supported. Additionally, the Info and Statistics section never take up a full page vertically, so the recommendations will still appear without scrolling.

Personally, I would move the recommended packages to the main section of the page and use the `_ListPackage` partial as I suggested, because this side of the page is much less overloaded. Not only will the recommendations always show on the page without scrolling because the description and readme are truncated if they are too long, but you can use the full width to show more information about the package rather than just the icon and ID.

By using the `_ListPackage` partial you will automatically get all of the environment that shows on the package search page, e.g.
image

I think this will be more useful for recommendations, because with just an icon and ID, users will have to click and then read to determine some basic info about the package.

jamesqo:
I changed the layout so that the recommendations are now right below the description/release notes. Here's how it looks like:

layout

Are we OK with having it above the dependencies / version history? If not then I can move it to the bottom.

I just realized that there is a "+ Show more" widget near the bottom of the Version History section, which looks exactly like what I had in mind for limiting the number of recs initially shown to 3. I'll implement that too.

jamesqo:
Now it looks like this with an expandable "+ Show more" widget:

layout-2

jamesqo:
@anangaur Ok, I will move it to the bottom.

> And just have the tip two lines per package. Description and a ruin buttons do not make sense.

I just followed @scottbommarito's suggestion of using the `_ListPackage` partial for each recommendation. To alter the design would probably require duplicating the code from there and involve some more effort... and personally, I think having a description makes sense because it gives a little more context about what the package is about if it has a non-descriptive name like "Antlr", etc.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -518,7 +518,7 @@
                 }
             </div>
 
-            @if (Model.RecommendedPackages.Any())
+            @if (!Model.Deleted && Model.RecommendedPackages.Any())
```
jamesqo: I just noticed that everything except Version History is under one giant `@if (!Model.Deleted) { ... }` block, so I added this here as well.

scottbommarito:
@anangaur - can you clarify what you mean by "just have the tip two lines per package" and "description and a ruin buttons do not make sense"?

I think the description is very useful to have in a list of recommended packages. If there is not sufficient information for the user to determine if the recommended package is useful for them on this page, they'll have to click on every single package to determine its usefulness. Additionally, because the list of recommended packages is at the bottom of the page and is collapsible, it doesn't really matter how much space it takes up. Also, what are "ruin buttons"?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
scottbommarito: Yes, I think that's correct. If the package is deleted, we shouldn't show recommendations.

scottbommarito (dismissed):
Looks great now! We need UI sign off from @anangaur but code looks great to me.

anangaur:
Autocorrect did the trick:). I meant the action buttons like edit, delete etc. 
I still feel the recommendations should not have much details.

jamesqo:
@anangaur I agree too, but the administrative buttons will not show for the vast majority of recommendations (only if the recommended package is owned by the user, which is pretty rare). Maybe a later PR can add an option to the `_ListPackage` partial to not show the administrative buttons?

Or... @scottbomarito maybe we can go back to passing in `currentUser: null` as a quick fix so the adminstrative buttons don't show?

src/NuGetGallery/Services/IPackageRecommendationService.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    public interface IPackageRecommendationService
```
skofman1: Please rename "Recommended" to "Related" everywhere

src/NuGetGallery/Services/IPackageService.cs
```diff
@@ -27,7 +27,7 @@ public interface IPackageService : ICorePackageService
         /// <returns></returns>
         Package FindPackageByIdAndVersion(string id, string version, int? semVerLevelKey = null, bool allowPrerelease = true);
 
-        Package FindAbsoluteLatestPackageById(string id, int? semVerLevelKey);
+        Package FindAbsoluteLatestPackageById(string id, int? semVerLevelKey = null);
```
skofman1: Please undo this change. It makes the code less readable.

src/NuGetGallery/Services/LocalReportService.cs
```diff
@@ -0,0 +1,40 @@
...
+            _fileStorage = fileStorage;
+        }
+
+        public async Task<ReportBlob> Load(string reportName)
+        {
+            using (var stream = await _fileStorage.GetFileAsync(FolderName, fileName: reportName))
+            {
+                if (stream == null)
+                {
+                    throw new ReportNotFoundException();
```
skofman1: Do you expect to always have a file per package? If this is the case, it sounds wasteful since there probably will be many packages with no reports. 
If you don't expect to always have a file per package, throwing an exception is not the correct behavior, since not finding a file is an expected and normal behavior.
I recommend returning an empty report in this case.

skofman1:
>         private const string _statsContainerName = "nuget-cdnstats";

This always loads reports from container "nuget-cdnstats". For "Related packages" we should use a separate container.

---
Refers to: src/NuGetGallery/Services/CloudReportService.cs:13 in 7d0a335. [](commit_id = 7d0a335c37df36c9cbf3d86ab48d48b8e4ef22e9, deletion_comment = False)

src/NuGetGallery/Services/PackageRecommendationService.cs
```diff
@@ -0,0 +1,71 @@
...
+    public class PackageRecommendationService : IPackageRecommendationService
+    {
+        private readonly IPackageService _packageService;
+        private readonly IReportService _reportService;
+
+        public PackageRecommendationService(
+            IPackageService packageService,
+            IReportService reportService)
+        {
+            _packageService = packageService;
```
skofman1: please add null checks

src/NuGetGallery/Services/PackageRecommendationService.cs
skofman1: Why do we need this encoding?

---
In reply to: [187759262](https://github.com/NuGet/NuGetGallery/pull/5841#discussion_r187759262) [](ancestors = 187759262)

src/NuGetGallery/ViewModels/RecommendedPackages.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGetGallery
+{
+    public class RecommendedPackages
```
skofman1: where is this class used? Couldn't find..

src/NuGetGallery/Services/PackageRecommendationService.cs
```diff
@@ -0,0 +1,71 @@
...
+            _reportService = reportService;
```
shishirx34: I believe this won't work with AzureStorage integration, the way it is right now `IReportService` is an instance of `CloudReportService` which `Load`s data from `nuget-cdnstats` container in the cloud. You need to create another service for `RecommendationReportService` which specifies `recommendations` as the container, under which we will upload the generated json blobs for recommendation. Or you could refactor the `CloudReportService` for this change.

src/NuGetGallery/Services/PackageRecommendationService.cs
```diff
@@ -0,0 +1,71 @@
...
+                var sb = new StringBuilder(capacity: bytes.Length * 2);
+                foreach (byte b in bytes)
+                {
+                    sb.Append(b.ToString("x2"));
+                }
+                return sb.ToString();
+            }
+
+            string encodedId = GetHexadecimalString(Encoding.UTF8.GetBytes(packageId));
+            return $"Recommendations/{encodedId}.json";
```
shishirx34: I believe this might be working with your local setup, probably looking under `reports/recommendations/<encodedId>.json` but it won't work with the Cloud storage.. see my comment earlier.

shishirx34: Also, these blob names need to be simply `id.json` for consistency.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -199,7 +199,7 @@ public override PackageRegistration FindPackageRegistrationById(string packageId
             return package;
         }
 
-        public virtual Package FindAbsoluteLatestPackageById(string id, int? semVerLevelKey)
+        public virtual Package FindAbsoluteLatestPackageById(string id, int? semVerLevelKey = null)
```
shishirx34: why default to null?

src/NuGetGallery/Services/ReportNotFoundException.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class ReportNotFoundException : Exception
+    {
+        public ReportNotFoundException() { }
+        public ReportNotFoundException(string message) : base(message) { }
+        public ReportNotFoundException(string message, Exception inner) : base(message, inner) { }
+        protected ReportNotFoundException(
+          SerializationInfo info,
+          StreamingContext context)
+            : base(info, context) { }
+    }
```
shishirx34: nit: new line spacings for consistency please.

src/NuGetGallery/ViewModels/RecommendedPackages.cs
shishirx34: In the recommendation service, it is used to `Deserialize` the data from json.

src/NuGetGallery/Services/PackageRecommendationService.cs
shishirx34: `throw new ArgumentNullException(nameof(packageService));`

src/NuGetGallery/ViewModels/RecommendedPackages.cs
skofman1: so why is it under ViewModels?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -517,6 +517,23 @@
                     </a>
                 }
             </div>
+
+            @if (!Model.Deleted && Model.RecommendedPackages.Any())
```
skofman1: to be able to evaluate the usage of the feature we need Google Analytics events.

src/NuGetGallery/ViewModels/RecommendedPackages.cs
shishirx34: `RequestModels` may be?

shishirx34:
@jamesqo - I will help you out on deploying this to our DEV environment(where we can do the actual hookup to all the recommendations) to validate if everything looks good once the feedback is addressed. We will need to generate the recommendations for the dev environment as well. 

https://dev.nugettest.org/packages/ 
https://apidev.nugettest.org/v3-index/index.json

src/NuGetGallery/Services/LocalReportService.cs
jamesqo: Nah, this is just part of the contract of the method to indicate we didn't find a report-- `CloudReportService` does it too. In `PackageRecService` we catch this exception and appropriately return an empty list.

src/NuGetGallery/Services/CloudReportService.cs
```diff
@@ -26,7 +26,7 @@ public Task<bool> IsAvailableAsync()
             return container.ExistsAsync();
         }
 
-        public async Task<StatisticsReport> Load(string reportName)
+        public async Task<ReportBlob> Load(string reportName)
```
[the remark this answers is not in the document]
jamesqo: Since this is an async method the compiler won't allow `out` parameters. Another option would be to make use of C# 7 and return `Task<(bool, ReportBlob)>` to avoid `out`, but the repo would need to adopt a dependency on the `System.ValueTuple` package.

src/NuGetGallery/Services/IReportService.cs
```diff
@@ -7,6 +7,6 @@ namespace NuGetGallery
 {
     public interface IReportService
     {
-        Task<StatisticsReport> Load(string reportName);
+        Task<IReportContainer> GetContainer(string containerName);
```
jamesqo: @skofman1 @anangaur Do you think this is a good way to refactor it to accept a container name?

src/NuGetGallery/Services/CloudReportService.cs
```diff
@@ -20,43 +20,52 @@ public CloudReportService(string connectionString, bool readAccessGeoRedundant)
             _readAccessGeoRedundant = readAccessGeoRedundant;
         }
 
-        public Task<bool> IsAvailableAsync()
+        public IReportContainer GetContainer(string containerName)
         {
-            var container = GetCloudBlobContainer();
-            return container.ExistsAsync();
+            return new Container(GetCloudBlobContainer(containerName));
```
jamesqo: @shishirx34 @skofman1 Did some refactoring and here are the results. Instead of this:

```cs
var report = await reportService.Load("blob.json");
```

we now do this:

```cs
var reportContainer = reportService.GetContainer("nuget-cdnstats");
var report = await reportContainer.Load("blob.json");
```

Now we're letting the caller specify the container name.

jamesqo: One thing to note is that we're calling `GetCloudBlobContainer` just once when `GetContainer` is called, instead of every time `Load` is called like before. Will this affect behavior at all?

src/NuGetGallery/Services/IPackageService.cs
jamesqo: OK, I'll just assume `SemVerLevelKey.Unknown` is the appropriate value. LMK otherwise

jamesqo: I just noticed the method right above this one also does `int? semVerLevelKey = null`. Are you sure this makes it less readable?

src/NuGetGallery/Services/PackageService.cs
jamesqo: @shishirx34 Also the method right above this defaults to null.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
jamesqo: Is there existing code you or @shishirx34 can point to that does this?

src/NuGetGallery/Services/PackageRecommendationService.cs
jamesqo: Done.

src/NuGetGallery/ViewModels/RecommendedPackages.cs
jamesqo: @skofman1 This is the same folder where `AggregateStats` is located, and that class serves a similar function (it's deserialized from JSON in `JsonAggregateStatsController`).

jamesqo:
Shishir / Svetlana, I made some more changes and responded to all of your feedback. PTAL when you have time 😃

src/NuGetGallery/Services/CloudReportService.cs
shishirx34: Hey @jamesqo, 

I like the split from `ReportService` to `ReportContainer`, however, I think it would be better to generate the container instances(`nuget-cdnstats`, `nuget-relatedpackages`), as part of DI rather than initiating them the way you have done here(in the service). This would make it easy to be consumed by other services in the constructor without having to specify `GetContainer` on each of the consuming service. Also testing is easier with the DI setup without the need to initialize containers. 

I have written up some [refactored code](https://github.com/NuGet/NuGetGallery/compare/dev...ml-cloud-report-generalize?expand=1) in branch `ml-cloud-report-generalize` to give you an idea of what I mean. However, you can integrate that logic with your change to have multiple instances of the registered container(per container name) and the service(`PackageRelatedService` or `JsonStatsService` will consume it as needed). I think this will help in reusability and keep the code clean. So we have the single instance of `CloudReportService` `StatsContainerService` `RelatedPackagesService` etc.

Let me know what you think.

With regards to
> One thing to note is that we're calling GetCloudBlobContainer just once when GetContainer is called, instead of every time Load is called like before. Will this affect behavior at all?

I don't think it should affect it in most cases, however, I think the `_connectionString` value can be updated remotely(we use a separate secret store) and that would mean, if we were to change it, this would break downloading any kind of blobs. So I believe, it should be kept in sync on per request. @scottbommarito - thoughts here?

shishirx34:
@jamesqo - I have reviewed the change, see my comment in the thread. With regards to `Google analytics` events, I am not much familiar either, let me research a bit on the data that we are already collecting and get back to you on if/what changes we might need to make to make sure we capture the `click` data on the recommended packages.

jamesqo:
@shishirx34 Sorry for the delay on this! Been busy with school for the past couple of weeks, but now that summer's here I think I can finally finalize the PR in the coming days.

shishirx34:
@jamesqo - not a problem. I am afraid I haven't been able to look into the Google analytics either. I will try to get the required information next week for you and we can start testing your changes in DEV env.

jamesqo:
@shishirx34 Since you're here, I want to ask you about some of your feedback. I tried your suggestion of making `IReportService` generic. I understand what your intention was (to avoid all the `GetContainer()` calls), but it's turning into quite a mess when I try to implement it. For example, now that `CloudReportService<TContainer> : IReportContainer<TContainer>` is generic, I have to change the Autofac registration to this:

```cs
            builder.RegisterGeneric(typeof(CloudReportService<>))
                .AsSelf()
                .As(typeof(IReportService<>))
                .As<ICloudStorageStatusDependency>()
                .SingleInstance();
```

This code will fail actually, because when it tries to create an instance of `ICloudStorageStatusDependency` it won't know which type parameter to use. So I have to make `ICloudStorageStatusDependency` generic as well and change all of the code that uses that, etc. which is kinda annoying... IMO, I don't think it's worth all this trouble to avoid the `GetContainer()` calls. I've already adapted all of the code + the tests to use `GetContainer()`.

jamesqo:
@skofman1 @shishirx34 I looked into the Google Analytics reference online and registered three different types of events. See here: https://github.com/NuGet/NuGetGallery/pull/5841/commits/9d8140c986306837481bfe333dd25724775de530

I'm not familiar with how Google Analytics works since I've never used it before, but does it let you see the URL of the page the user was on when the event was fired? This is important for generating more targeted recommendations in the future: we need to know "user clicked on package B from package A", not just "user clicked on package B".

jamesqo:
@shishirx34 @skofman1 @scottbommarito Do you guys still want this PR? I took care of all the feedback before, all that needs to be done is fixing the merge conflicts that accumulated over the summer. If so let me know, if not I'll close.

shishirx34:
@jamesqo - my apologies, I had been away and missed your previous responses. Yes we would very much like to merge this PR in, and I really appreciate your contribution. Please do address the merge conflicts. With regards to the other feedback, with `RegisterGenerics` you can ignore it. I will refactor it separately after we get your changes merged in. 

Once you have everything ready, lets test it out on DEV. Could you please run your recommendations algorithm on the packages in DEV env, I will upload the generated recommendation blobs to the azure container and deploy your changes to DEV environment. The endpoint for discovering packages on DEV: 
```
https://apidev.nugettest.org/v3-index/index.json
```

jamesqo:
@shishirx34 Haha I'll have to make sure I don't forget about this myself, this almost got buried in my GitHub notifications. I'll certainly notify you once I've done that and tested it out.

jamesqo:
@skofman1 Yeah. Is it alright if I resolve the conflicts this weekend?

joelverhagen:
I'm going to close this PR. I think there is potential in this space but, given the PR is so stale, I think it's best to close it an revisit the issue later.

If anyone is interested in taking this idea forward, I think it would be best to have a formal proposal as described here:
https://github.com/NuGet/Home/tree/dev/meta#nuget-proposal-process

This would allow us (NuGet team) and the community to discuss the ideas first and then later approach the code.

The https://github.com/NuGet/NuGetGallery/issues/5751 issue still tracks the idea and could be referenced in a formal proposal.

PR #5864, opened 2018-05-01
FindByKey and FindByUsername should not include deleted accounts.

cristinamanum:
The default behavior for FindByKey and FindByUsername should not include deleted accounts.


src/NuGetGallery/Services/UserService.cs
```diff
@@ -352,17 +352,33 @@ public virtual IList<User> FindByUnconfirmedEmailAddress(string unconfirmedEmail
             }
         }
 
-        public virtual User FindByUsername(string username)
+        public virtual User FindByUsername(string username, bool includeDeleteRecord = false)
```
chenriksson: Arg names should match the interface (plural). Also, would rename to simply `includeDeleted` as `Record` is kind of generic data term that we don't use elsewhere in the gallery.

src/NuGetGallery/Services/UserService.cs
```diff
@@ -352,17 +352,33 @@ public virtual IList<User> FindByUnconfirmedEmailAddress(string unconfirmedEmail
...
+                return UserRepository.GetAll()
+                .Include(u => u.Roles)
+                .Include(u => u.Credentials)
+                .SingleOrDefault(u => u.Username == username);
+            }
             return UserRepository.GetAll()
+                .Where(u => !u.IsDeleted)
                 .Include(u => u.Roles)
                 .Include(u => u.Credentials)
                 .SingleOrDefault(u => u.Username == username);
```
chenriksson: nit: can reuse code by wrapping only Where clause in the condition.

```
var users = UserRepository.GetAll();
if (includeDeleted)
{
    users = users..Where(u => u.IsDeleted);
}
return users.Include(...).Include(...).SingleOrDefault(...);
```

chenriksson: Same for below. Also, username should be case insensitive search to match our DB collation.

shishirx34: this equality comparison is case insensitive by default due to db collation, why would it matter if username has any casing?

src/NuGetGallery/Services/UserService.cs
cristinamanum: Updated

src/NuGetGallery/Services/UserService.cs
cristinamanum: Updated (first comment) , second comment @shishirx34 is correct.

Issue #5889, opened 2018-05-04
HTTP 500 error when verifying a new account


kbaley:
We're converting our company username/password account into an organization. Here's what happened leading up to the HTTP 500 error

- In Chrome, I logged into our company nuget account (with a username/password)
- In Firefox, I created a new account for myself using my Microsoft account
- I received the email asking me to verify my account
- I clicked the verify link which launched Chrome (where I was already logged in but to a different account)

At this point, I received the error:

![image](https://user-images.githubusercontent.com/130546/39647691-9673a94a-4fa5-11e8-8fb4-9e16fda7169f.png)

scottbommarito:
Thanks for reporting this issue.

The verify link is intended to be opened while logged in as the account you are verifying. In other words, in your situation, we expected you to open the link from Firefox and not Chrome.  If you try opening the link while logged in as your new account it should work.

Regardless, the server should not have returned 500, and should have instead told you that the verify link you clicked was not for your company's NuGet account.

I tried looking at our exception logs and didn't see your request at a glance.
Could you tell me what your NuGet account usernames are (both your company account and your new account) so I can find the logs more easily?
If you don't want to post this information publicly, feel free to send an email to support@nuget.org mentioning this issue along with the usernames.

scottbommarito:
I was able to find the logs and determine the bug in the code. No need to send your usernames.

skofman1:
Investigation: 
Repro steps:
1. Create account A 
2. Create account B
3. Log in with account A
4. Click on account confirmation link for account B. 

Result: 500 page (NullReferenceException)
Expected: a nice message

The issue is here: https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Controllers/AccountsController.cs#L130

GetAccount returns null when there's a mismatch, while [here](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Controllers/AccountsController.cs#L136) we try to get the username from the null ref

PR #5891, opened 2018-05-04
Fix HTTP 500 error when verifying a new account

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5889

When trying to confirm a different account than your own, it now says:
```
You cannot confirm a different user's email address!
Please log in as '<AccountName>' and then click the link again.
```


src/NuGetGallery/Views/Shared/Confirm.cshtml
```diff
@@ -34,36 +34,28 @@
                     @<text>
                         @if (Model.WrongUsername)
                         {
+                            <span>You cannot confirm a different user's email address!</span>
```
scottbommarito: The alert messages used to be on 3 lines.
```
<alert icon>
You cannot confirm a different user's email address!
Please log in as '@Model.AccountName' and then click the link again.
```

This changes it to:
```
<alert icon> You cannot confirm a different user's email address!
Please log in as '@Model.AccountName' and then click the link again.
```
which is more consistent with the rest of the site.

src/NuGetGallery/ViewModels/ConfirmationViewModel.cs
```diff
@@ -24,6 +24,22 @@ public class ConfirmationViewModel
...
 
+        /// <summary>
+        /// Creates a <see cref="ConfirmationViewModel"/> for an attempt to confirm a different account's email.
+        /// </summary>
+        /// <param name="username">The username of the account whose email was attempted to be confirmed.</param>
+        public ConfirmationViewModel(string username)
+        {
+            AccountName = username;
+            WrongUsername = true;
+            SuccessfulConfirmation = false;
```
chenriksson: It's not apparent from the constructors that one is for an error case and the other isn't.

I recommend continuing to set these from outside the constructor.

scottbommarito: I agree, but I think I'd rather have
```
public static ConfirmationViewModel CreateForWrongUsername(string username)
{
    var model = new ConfirmationViewModel(username);
    ...
}
```
than setting it from outside the constructor.

scottbommarito:
@chenriksson the 500 is that the constructor for `ConfirmationViewModel` expects a non-null `User`, and in the `UsersController`, when a username that's not the current user's username is provided, the `User` passed to `ConfirmationViewModel` is null. This PR introduces an additional constructor for `ConfirmationViewModel` that does not expect a non-null `User`.

src/NuGetGallery/ViewModels/ConfirmationViewModel.cs
scottbommarito: Changed my mind, using your approach.

PR #5894, opened 2018-05-07
Fix nondeterministic test.

cristinamanum:
The test NuGetGallery.SearchClient.RetryingHttpClientWrapperFacts.LoadBalancesBetweenValidUrisForGetAsync uses the random comparer to direct the requests. This makes it to be non-deterministic. Change the Assert to make it deterministic.

### DIFF
--- a/tests/NuGetGallery.Facts/SearchClient/RetryingHttpClientWrapperFacts.cs
+++ b/tests/NuGetGallery.Facts/SearchClient/RetryingHttpClientWrapperFacts.cs
@@ -119,7 +119,7 @@ namespace NuGetGallery.SearchClient
             }
             // Because the algorithm for choosing uri1 or uri2 is random asserting that the one or the other uri is not reliable
             // However the sum of requests between two uri requests needs to be larger or equal with 1
-            Assert.True(hitUri1Count + hitUri2Count >= 1, $"The uri1 had {hitUri1Count} hits and the uri2 had {hitUri2Count} hits. The sum will need to be at least 1." );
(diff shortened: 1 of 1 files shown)


tests/NuGetGallery.Facts/SearchClient/RetryingHttpClientWrapperFacts.cs
```diff
@@ -104,22 +104,22 @@ private RetryingHttpClientWrapper CreateWrapperClient()
...
-
-            Assert.True(hasHitUri1, "The first valid Uri has not been hit within the limit of " + numRequests + " requests.");
-            Assert.True(hasHitUri2, "The second valid Uri has not been hit within the limit of " + numRequests + " requests.");
+            if (inspectingHandler.Requests.Any(r => r.RequestUri == ValidUri2))
+            {
+                hitUri2Count++;
+            }
+            // Because the algorithm for choosing uri1 or uri2 is random asserting that the one or the other uri is not reliable
+            // However the sum of requests between two uri requests needs to be larger or equal with 1
+            Assert.True(hitUri1Count + hitUri2Count >= 1, $"The uri1 had {hitUri1Count} hits and the uri2 had {hitUri2Count} hits. Together will need to be more than 1." );
```
dtivel: Update the message.  It shouldn't be "more than 1" but "at least 1."

    Assert.True(hitUri1Count + hitUri2Count >= 1, $"The uri1 had {hitUri1Count} hits and the uri2 had {hitUri2Count} hits. The combined hit count needs to be at least 1.");

cristinamanum: Updated.

PR #5905, opened 2018-05-09
Run multiple functional test categories in parallel

scottbommarito:
Part of https://github.com/NuGet/Engineering/issues/1213

Before making major changes to the functional test suite, I decided to tackle this issue first for a quick improvement.

When functional tests are run, several "test categories" are ran. Usually, this is P0, P1, and P2.
Before this change, these functional tests are ran in series: P0, then P1, and then P2.
With this change, they are run in parallel.

https://nuget.visualstudio.com/NuGetBuild/_build/index?buildId=31009&_a=summary
Functional tests took 50 minutes instead of 80!

I also rewrote any `bat` scripts into `ps1` and deleted any scripts that are no longer used.

Note: the test categories passed into the script are now `P0;P1;P2` instead of `P0Tests;P1Tests;P2Tests`, so we will need to update our configuration. This was done because it made consolidating the `Run__Tests.bat` into a single file possible.


tests/Scripts/BuildTests.ps1
```diff
@@ -0,0 +1,31 @@
...
+& $nuget "restore" $fullSolutionPath "-NonInteractive"
+if ($LastExitCode) {
+    throw "Failed to restore packages!"
+}
+
+# Build the solution
+Write-Host "Building solution"
+& $msBuild $fullSolutionPath "/p:Configuration=$Config" "/p:Platform=Any CPU" "/p:CodeAnalysis=true" "/m" "/v:M" "/fl" "/flp:LogFile=$rootName\msbuild.log;Verbosity=diagnostic" "/nr:false"
+if ($LastExitCode) {
+    throw "Failed to restore packages!"
```
loic-sharma: The error message here is wrong

tests/Scripts/RunTestScripts.ps1
```diff
@@ -1,35 +1,105 @@
...
 param(
-    [string]$TestCategories
+    [Parameter(Mandatory)][string]$TestCategories
 )
 
 $dividerSymbol = "~"
 
+& "$PSScriptRoot\BuildTests.ps1" | Out-Host
+
+Write-Host ($dividerSymbol * 20)
```
loic-sharma: Can we make two helper variables? Something like `$fullDivider` and `$halfDivider`. Full = 20 characters, half = 10 characters

tests/Scripts/RunTestScripts.ps1
```diff
@@ -1,35 +1,105 @@
...
+        } | Out-Null
+    
+    Write-Host ($dividerSymbol * 20)
+
+    do {
+        Write-Host "Waiting for $($TestCategoriesArray -join ", ")"
+        $jobs = Get-Job -Name $TestCategoriesArray
+        $job = Wait-Job $jobs -Any
+        Output-Job $job | Out-Host
+        $TestCategoriesArray = $TestCategoriesArray | Where-Object { $_ -ne $job.Name }
```
loic-sharma: You should force the result to be an array like this:

```
$TestCategoriesArray = @($TestCategoriesArray | Where-Object { $_ -ne $job.Name })
```

Why? To guess the result of:

```
$a = @('foo', 'bar') | Where-Object { $_.StartsWith('b') }
$b = @('foo', 'bar', 'buzz') | Where-Object { $_.StartsWith('b') }

Write-Host $a[0]
Write-Host $b[0]
```

The answer is:

```
b
bar
```

This happens because the first result of `Where-Object` returns a single element, so Powershell "magically unwraps" it into just a string. You can - and should - avoid this stupid behavior by forcing the result of the `Where-Object` to be an array:

````
$a = @(@('foo', 'bar') | Where-Object { $_.StartsWith('b') })
```

tests/Scripts/RunTestScripts.ps1
```diff
@@ -1,35 +1,107 @@
...
+    [CmdletBinding()]
+    param(
+        [Parameter(ValueFromPipeline)]$job
+    )
+
+    Write-Host $fullDivider
+    $jobName = $job.Name
+    Write-Host "Finished testing test category $jobName."
+    Write-Host $halfDivider
+    Receive-Job $job | Out-Host
```
agr: Will we get job output in logs?

tests/Scripts/RunTests.ps1
```diff
@@ -0,0 +1,59 @@
...
+$xunit = "$rootRootName\packages\xunit.runner.console.2.3.1\tools\net452\xunit.console.exe"
+
+# Test results files
+$functionalTestsResults = "$rootName/functionaltests.$TestCategory.xml"
+$webUITestResults = "$rootName/NuGetGallery.$TestCategory.WebUITests.trx"
+$loadTestResults = "$rootName/NuGetGallery.$TestCategory.LoadTests.trx"
+
+# Clean previous test results
+if (Test-Path $functionalTestsResults) {
+    Remove-Item $functionalTestsResults
```
agr: `Remove-Item $functionalTestsResults -ErrorAction Ignore` ? One line instead of 3 :)

PR #5915, opened 2018-05-11
Refresh content on app start

shishirx34:
There was a bug which would prevent content refresh on app start due to `NullReferenceException`, `DependencyResolver` cannot resolve a dependency in the `Autofac` builder, the resolver is set after the `load` method. Moving the code to OwinStartup fixes this issue.

### DIFF
--- a/src/NuGetGallery/App_Start/OwinStartup.cs
+++ b/src/NuGetGallery/App_Start/OwinStartup.cs
@@ -64,7 +64,7 @@ namespace NuGetGallery
             // Configure machine key for session persistence across slots
             SessionPersistence.Setup(config);
 
-            // Refresh the content
+            // Refresh the content for the ContentObjectService to guarantee it has loaded the latest configuration on startup.
             if (config.Current.IsHosted)
             {
                 var contentObjectService = dependencyResolver.GetService<IContentObjectService>();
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/App_Start/OwinStartup.cs
```diff
@@ -63,6 +64,13 @@ public static void Configuration(IAppBuilder app)
             // Configure machine key for session persistence across slots
             SessionPersistence.Setup(config);
 
+            // Refresh the content
```
chenriksson: Could you replace this comment w/ a useful one about why this needs to be in OwinStartup?

shishirx34:
Its possibly also in AI, @scottbommarito brought it to my attention.

src/NuGetGallery/App_Start/OwinStartup.cs
shishirx34: I think its self explanatory, the bug was it was put in the defaultdependencies where it couldn't be resolved.

scottbommarito: I agree with @chenriksson 

```
Refresh the ContentObjectService to guarantee it has loaded the latest configuration on startup.
```

scottbommarito:
I believe we discovered the exception in AI, not sure though--it's been a known issue for a while.

PR #5922, opened 2018-05-14
Make functional tests reuse packages they have uploaded if possible

scottbommarito:
- deleted Fluent tests, which we haven't run in years
- modified `ClientSdkHelper` to keep track of what packages it has uploaded and return existing packages if they match what is being asked for
- made most tests no longer wait on V3 for upload/unlist
- made other small modifications to the tests

Functional tests now run in ~25 minutes!
https://nuget.visualstudio.com/NuGetBuild/_build/index?buildId=31459&_a=summary


tests/NuGetGallery.FunctionalTests.Core/Helpers/UploadHelper.cs
```diff
@@ -48,23 +50,24 @@ protected PackageToUpload(PackageToUpload package)
         /// <summary>
         /// Gets a unique ID for a package to upload.
         /// </summary>
-        public static string GetUniquePackageId(string test)
+        public static string GetUniquePackageId()
         {
-            if (test == null)
+            lock (UniqueLock)
```
agr: Could you please provide clarifications about why lock is needed? Were there concurrency issues? If `GetUniquePackageId()` is called multiple times from different threads I'd guess there is a good chance of this method generating the same id for multiple calls, since timer resolution is quite low.

scottbommarito: It is frequently called simultaneously by multiple threads. I added the lock to guarantee that it's always separate IDs, regardless of the threading.

tests/NuGetGallery.FunctionalTests.Core/Helpers/ClientSdkHelper.cs
```diff
@@ -227,18 +367,156 @@ public bool CheckIfPackageExistsInSource(string packageId, string sourceUrl)
...
+                    Packages.Add(packageInfo);
+                }
+
+                await packageInfo.PackageIsReady;
+                return packageInfo;
+            }
+
+            public async Task<PackageInfo> PushNewVersion(ClientSdkHelper helper, string uploadApiKey = null)
+            {
+                if (!CanUseApiKeyToPushNewVersion(uploadApiKey))
```
skofman1: this is not clear. Lets discuss offline

tests/NuGetGallery.FunctionalTests.Core/Helpers/ClientSdkHelper.cs
```diff
@@ -227,18 +367,156 @@ public bool CheckIfPackageExistsInSource(string packageId, string sourceUrl)
...
+
+            public bool WasUnlistedByApiKey(string apiKey)
+            {
+                return apiKey == UnlistApiKey;
+            }
+
+            private static string MapApiKeyToOwner(string apiKey)
+            {
+                if (apiKey == null ||
+                    apiKey == EnvironmentSettings.TestAccountApiKey ||
```
skofman1: I assume the config changes are not part of this PR?

tests/NuGetGallery.FunctionalTests.Core/Helpers/UploadHelper.cs
agr: How does lock make "separation" for IDs?

tests/NuGetGallery.FunctionalTests.Core/Helpers/ClientSdkHelper.cs
scottbommarito: https://github.com/NuGet/NuGetGallery/pull/5939

tests/NuGetGallery.FunctionalTests.Core/Helpers/ClientSdkHelper.cs
scottbommarito: Discussed offline--the code here is relatively complex because the behavior we're trying to support (caching packages for later use) is inherently more complex than the behavior the E2E tests have. I think the benefits of this (25 minute runs instead of 40 minute runs) outweigh the complexity concerns.

tests/NuGetGallery.FunctionalTests.Core/Helpers/ClientSdkHelper.cs
```diff
@@ -118,52 +123,187 @@ public bool CheckIfPackageExistsInSource(string packageId, string sourceUrl)
...
-        public async Task UploadNewPackage(string packageId, string version = "1.0.0", string minClientVersion = null,
-            string title = null, string tags = null, string description = null, string licenseUrl = null,
-            string dependencies = null, string apiKey = null, bool success = true)
+        private static string GetApiKeyWithSameOwnerThatCanUpload(string apiKey = null)
         {
+            if (apiKey == null ||
+                apiKey == GalleryConfiguration.Instance.Account.ApiKey ||
+                apiKey == GalleryConfiguration.Instance.Account.ApiKeyPush ||
+                apiKey == GalleryConfiguration.Instance.Account.ApiKeyPushVersion ||
+                apiKey == GalleryConfiguration.Instance.Account.ApiKeyUnlist)
```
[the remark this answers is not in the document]
scottbommarito: If somebody passes in a variable that isn't identical to these values, someone wrote an invalid test or wrote something that should be expected to fail.

tests/NuGetGallery.FunctionalTests.Core/Helpers/ClientSdkHelper.cs
```diff
@@ -118,52 +123,187 @@ public bool CheckIfPackageExistsInSource(string packageId, string sourceUrl)
...
+            {
+                return GalleryConfiguration.Instance.Account.ApiKey;
```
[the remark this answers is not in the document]
scottbommarito: No, this is used for uploading a package if there is no existing package that can be used.

E.g. if someone wants to unlist a package but there is no existing package that this API key can be used to unlist, the framework uploads a package first using this key and then unlists it.

tests/NuGetGallery.FunctionalTests.Core/Helpers/ClientSdkHelper.cs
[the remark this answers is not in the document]
chenriksson: I believe DB collation is case insensitive, but we should do whatever Gallery does.

tests/NuGetGallery.FunctionalTests.Core/Helpers/UploadHelper.cs
scottbommarito: The idea is that acquiring and releasing the lock should take at least 1 ms, thus making the IDs unique.

agr: What's the timer resolution?

PR #5936, opened 2018-05-17
TSA: move to latest d3

chenriksson:
Fix d3 TSA issues by moving to latest version:
- use strict
- syntax errors
  d3 quote: *`To the consternation of some users, 3.x employed Unicode variable names such as λ, φ, τ and π for a   concise representation of mathematical operations. A downside of this approach was that a SyntaxError would occur if you loaded the non-minified D3 using ISO-8859-1 instead of UTF-8. 3.x also used Unicode string literals, such as the SI-prefix µ for 1e-6. 4.0 uses only ASCII variable names and ASCII string literals (see [rollup-plugin-ascii](https://github.com/mbostock/rollup-plugin-ascii)), avoiding encoding problems.`*
- possibly the banned terms ('arguments')

Also fixing 'mimetype' build warnings due to casing.


skofman1:
Issue to track the deployment?

chenriksson:
@skofman1 I'm currently tracking all TSA fixes in https://github.com/NuGet/Engineering/issues/1416, but I'll be separating into separate TSA issues per deployment once I get a complete list.

PR #5938, opened 2018-05-17
Track metric for unexpected failed package pushes

scottbommarito:
https://github.com/NuGet/Engineering/issues/1391

may want to add `?w=1` to the URL to view without whitespace changes


src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -169,6 +170,20 @@ public void TrackPackagePushEvent(Package package, User user, IIdentity identity
...
+        {
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            TrackMetric(Events.PackagePushFailure, 1, properties => {
+                properties.Add(ClientVersion, GetClientVersion());
+                properties.Add(ProtocolVersion, GetProtocolVersion());
+                properties.Add(AccountCreationDate, GetAccountCreationDate(user));
```
loic-sharma: >AccountCreationDate [](start = 31, length = 19)

What's the account creation date for on a push failure?

loic-sharma: Exception type would be a useful dimension

---
In reply to: [189134212](https://github.com/NuGet/NuGetGallery/pull/5938#discussion_r189134212) [](ancestors = 189134212)

skofman1: Please don't log the user info. It's not relevant here. If needed we can get it from auditing.

---
In reply to: [189134212](https://github.com/NuGet/NuGetGallery/pull/5938#discussion_r189134212) [](ancestors = 189134212)

skofman1: Exception can be retrieved by correlation with Exceptions table

---
In reply to: [189135201](https://github.com/NuGet/NuGetGallery/pull/5938#discussion_r189135201) [](ancestors = 189135201,189134212)

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -169,6 +170,20 @@ public void TrackPackagePushEvent(Package package, User user, IIdentity identity
...
         }
 
+        public void TrackPackagePushFailureEvent(User user)
+        {
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            TrackMetric(Events.PackagePushFailure, 1, properties => {
```
skofman1: Package id and version (best effort)

src/NuGetGallery/Services/TelemetryService.cs
scottbommarito: We track account creation date for all package and organization events, probably just because we can. I'd rather leave it in there than put in the effort to remove it.

src/NuGetGallery/Services/TelemetryService.cs
scottbommarito: added

scottbommarito:
@loic-sharma @skofman1 I added the id/version as requested

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -332,217 +332,228 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+            }
+            catch (Exception)
+            {
+                TelemetryService.TrackPackagePushFailureEvent(GetCurrentUser(), id, version);
+                throw;
             }
         }
 
-        private static ActionResult BadRequestForExceptionMessage(Exception ex)
+        private ActionResult HandleExpectedPackagePushFailure(Exception ex)
```
chenriksson: Why remove the `static`?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -332,217 +332,228 @@ public virtual Task<ActionResult> CreatePackagePost()
...
-                    return BadRequestForExceptionMessage(ex);
-                }
-                catch (FrameworkException ex)
-                {
-                    return BadRequestForExceptionMessage(ex);
-                }
+            }
+            catch (Exception)
+            {
+                TelemetryService.TrackPackagePushFailureEvent(GetCurrentUser(), id, version);
```
chenriksson: Client retries by default? Just wondering how accurate the metric will be... maybe just something to keep in mind when monitoring this.

src/NuGetGallery/Services/ITelemetryService.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using NuGet.Versioning;
```
chenriksson: nit: sort System first

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: will revert changes to this method name--originally i was going to put the telemetry service call in it

src/NuGetGallery/Services/TelemetryService.cs
loic-sharma: @skofman1 In Application Insights, yes, but that wouldn't work in Geneva. I guess we can add in the dimension if/when we need it.

@scottbommarito We shouldn't log irrelevant data "just because we can". The account creation date will likely never help us debug an issue, it's just noise.

shishirx34: Yeah `AccountCreationDate` not needed here, in other cases we were using a single method for user related event tracking hence it gets tracked for unrealted events too. Plus we are hitting daily cap limits on AI we should try to avoid unnecessary data.

PR #5939, opened 2018-05-17
Use json configuration files instead of environment variables for functional tests

scottbommarito:
https://github.com/NuGet/Engineering/issues/1258

https://nuget.visualstudio.com/_git/NuGetBuild/pullrequest/438?_a=overview


tests/Scripts/Import-Configuration.ps1
```diff
@@ -0,0 +1,65 @@
...
+param (
+    [string]$Instance,
+    [string]$Project,
+    [string]$PersonalAccessToken,
+    [string]$Repository,
+    [string]$Branch = "master",
+    [ValidateSet("production", "staging")][string]$Slot = "production",
+    [ValidateSet("Dev", "Int", "Prod")][string]$Environment,
+    [ValidateSet("USNC", "USSC")][string]$Region
+)
```
chenriksson: I assume this is based on the following?
https://raw.githubusercontent.com/NuGet/NuGet.Services.EndToEnd/dev/importconfig.ps1

Would the logic be simpler if the config were layed out the same as E2E config, with 1 json per environment?

tests/Scripts/Import-Configuration.ps1
```diff
@@ -0,0 +1,65 @@
+[CmdletBinding()]
+param (
```
agr: `[Parameter(Mandatory=$true)]` needed anywhere?..

scottbommarito: I'll add it, good catch

tests/Scripts/Import-Configuration.ps1
```diff
@@ -0,0 +1,65 @@
...
+)
+
+Write-Host "Importing configuration for Gallery Functional tests on $Environment $Region"
+
+Add-Type -AssemblyName System.IO.Compression.FileSystem
+
+$basicAuth = [Convert]::ToBase64String([Text.Encoding]::ASCII.GetBytes(("{0}:{1}" -f 'PAT', $PersonalAccessToken)))
+$headers = @{ Authorization = ("Basic {0}" -f $basicAuth) }
+
+Function Get-MergedObject {
```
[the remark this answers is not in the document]
scottbommarito: agreed, wasn't familiar with powershell documentation, will add

tests/Scripts/Import-Configuration.ps1
scottbommarito: I like having the multi-layered files here because it makes our config easier to maintain. Additionally, I was able to write this code fairly easily using examples online, and I don't expect us to need to make many changes to it.

tests/NuGetGallery.FunctionalTests.Core/Helpers/UploadHelper.cs
```diff
@@ -111,14 +111,12 @@ private static IEnumerable<WebTestRequest> UploadPackage(WebTest test, PackageTo
             var uploadRequest = AssertAndValidationHelper.GetHttpRequestForUrl(UrlHelper.UploadPageUrl);
             yield return uploadRequest;
 
+            var cancelUploadPostRequest = AssertAndValidationHelper.GetCancelUploadPostRequestForPackage(test);
+            yield return cancelUploadPostRequest;
```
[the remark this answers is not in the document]
scottbommarito: basically this test will fail right now if a package is pending upload

sometimes, when functional tests run and fail, a package will be left in the pending upload state, which will break the next run of the functional tests

i will add a comment

tests/NuGetGallery.FunctionalTests.Core/GalleryConfiguration.cs
```diff
@@ -0,0 +1,83 @@
...
+        [JsonProperty]
+        public OrganizationConfiguration AdminOrganization { get; private set; }
+        [JsonProperty]
+        public OrganizationConfiguration CollaboratorOrganization { get; private set; }
+        [JsonProperty]
+        public BrandingConfiguration Branding { get; private set; }
+
+        static GalleryConfiguration()
+        {
+            var configurationString = File.ReadAllText(EnvironmentSettings.ConfigurationFilePath);
```
shishirx34: This will throw `ArgumentNullException` if parameter is `null`(if not set), should be fine I suppose. I think it would be good to do a null check and throw exception with a helpful message error(suggest running import-configuration.ps1) if it's null.

tests/Scripts/Import-Configuration.ps1
```diff
@@ -0,0 +1,65 @@
...
+                # Add the property to the output object
+                # If the property already exists on the output object, overwrite it
+                $output | Add-Member -MemberType NoteProperty -Name $name -Value $value -Force
+            }
+        }
+}
+
+# Download the config files--common ("All"), per environment, and per region--and merge them into a single file
+$configObject = New-Object PSObject
+"All", $Environment, "$Environment-$Region" | ForEach-Object {
```
agr: Maybe do regular `foreach` loop?
```
foreach ($name in @("All", $Environment, "$Environment-$Region"))
{
    # do stuff with $name instead of $_
}
```
?
Reads better in my opinion

tests/Scripts/Import-Configuration.ps1
```diff
@@ -0,0 +1,65 @@
...
+    # Merge the current file with the last files
+    Get-MergedObject -source $configData -output $configObject
+}
+
+# Add a field to the file determining which slot should be tested
+$configObject | Add-Member -MemberType NoteProperty -Name "Slot" -Value $Slot
+
+# Save the file and set an environment variable to be used by the functional tests
+$configurationFilePath = "$PSScriptRoot\Config-$Environment-$Region.json"
+[Environment]::SetEnvironmentVariable("ConfigurationFilePath", $configurationFilePath)
```
agr: `$env:ConfigurationFilePath = $configurationFilePath`?

scottbommarito: nope--doesn't work because `$env:ConfigurationFilePath` doesn't set the environment variable in a way that can be read by the functional tests

tests/Scripts/Import-Configuration.ps1
```diff
@@ -0,0 +1,94 @@
...
+            $source = New-Object PSObject -Property @{ A = "A", B = New-Object PSObject -Property @{ C = "C" }, D = "D"}
+            $output = New-Object PSObject -Property @{ D = "E", E = "E" }
+
+            Get-MergedObject -Source $source -Output $output
+
+            $output is now @{ A = "A", B = New-Object PSObject -Property @{ C = "C" }, D = "D", E = "E" }
+    #>
+    param(
+        [PSObject]$Source,
+        [PSObject]$Output
```
agr: `Target` would probably be more appropriate name than `Output`.

PR #5958, opened 2018-05-21
Move PagerDuty support rotation to config

chenriksson:
Moving support request duty to secondary on-call to reduce rotations. NuGetDeployment PR coming.


chenriksson:
Change in plans - new PR coming...

PR #5972, opened 2018-05-21
Remove PG integration for support

chenriksson:
Removes PG integration, so new requests will be unassigned.


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -312,7 +312,7 @@ private ActionResult TransformToOrganizationFailed(string errorMessage)
             }
 
             var isSupportRequestCreated = await _supportRequestService.TryAddDeleteSupportRequestAsync(user);
-            if (await _supportRequestService.TryAddDeleteSupportRequestAsync(user))
+            if (isSupportRequestCreated)
```
chenriksson: bug: delete account workflow currently creates 2 requests

src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
```diff
@@ -243,20 +229,7 @@ public int GetIssueCount(int? assignedToId, string reason, int? issueStatusId)
...
-                QuietLog.LogHandledException(sqlException);
-
-                var packageInfo = "N/A";
-                if (package != null)
-                {
-                    packageInfo = $"{package.PackageRegistration.Id} v{package.Version}";
-                }
-
-                var errorMessage = $"Error while submitting support request at {DateTime.UtcNow}. Support reason = {reason ?? "N/A"}. Package info = {packageInfo}";
-                await _pagerDutyClient.TriggerIncidentAsync(errorMessage);
```
chenriksson: Searched PG incidents, and it doesn't look like this is common. Note that all callers also send email, so we'd still get support email in this case.

PR #5981, opened 2018-05-22
Remove all PG references in code and DB

chenriksson:
PR #5972 removed the integration. This removes all references, including deleting the PG username column in the support DB.


src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
```diff
@@ -134,14 +110,14 @@ public int GetIssueCount(int? assignedToId, string reason, int? issueStatusId)
 
                 if (currentIssue.AssignedToId != assignedToId)
                 {
```
cristinamanum: This assigned Id will be taken from galley db?

chenriksson: No, assigned to ID is only from support DB.

Scenario here is that you change who issue is assigned to in the Support Requests admin view. The `if` check just avoids the update if the assignedToId didn't change. This logic is the same.

chenriksson:
Thanks @ryuyu @shishirx34 @cristinamanum !

Issue #5944, opened 2018-05-18
System.InvalidOperationException: Admin of an organization cannot remove the organization as owner of a package


cristinamanum:
The scenario is:
Package A has two owners that are organizations (org1, org2).
These organizations own the namespace for package A
Org1 admin tries to remove the org1 as owner of the package.

Result:
System.InvalidOperationException at NuGetGallery.PackageOwnership
The user '**admin1***' does not have permission to remove the owner 'EntityFramework'.

Exception in:

C:\vsts-agent\_work\30\s\src\Gallery\submodules\Gallery\src\NuGetGallery\Services\PackageOwnershipManagementService.cs
Line 114

AI exception for nugetgallery
 exceptions | where timestamp == todatetime('2018-05-17T23:33:55.623')
| where operation_Id == "8iJqhjsQT3k="

PR #5984, opened 2018-05-22
Allow organization admins to remove their organization's ownership of a package in a reserved namespace

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5944

1 - Organization A owns package B that is in reserved namespace C
2 - User D that is admin of A removes A from ownership of B

Expected:
Organization A is no longer an owner of B and B loses its verification

Actual:
D gets an HTTP 500.
Organization A remains an owner of B.

This PR fixes that behavior.

Also made some general fixes to prevent issues like these in the future.
E.g. we now use `ActionsRequiringPermissions` to determine reserved namespace package verification.

### DIFF
--- a/src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
+++ b/src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
@@ -15,7 +15,7 @@ namespace NuGetGallery
     /// E.g. "JQuery.Extensions.MyCoolExtension" matches both "JQuery.*" and "JQuery.Extensions.*".
     /// </remarks>
     public class ActionRequiringReservedNamespacePermissions
-        : ActionRequiringEntityPermissions<IReadOnlyCollection<ReservedNamespace>>, IActionRequiringEntityPermissions<ActionOnNewPackageContext>
+        : ActionRequiringEntityPermissions<IReadOnlyCollection<ReservedNamespace>>, IActionRequiringEntityPermissions<ActionOnNewPackageContext>, IActionRequiringEntityPermissions<ReservedNamespace>
     {
         public PermissionsRequirement ReservedNamespacePermissionsRequirement { get; }
 
@@ -32,9 +32,19 @@ namespace NuGetGallery
             return CheckPermissions(currentUser, account, GetReservedNamespaces(newPackageContext));
         }
 
+        public PermissionsCheckResult CheckPermissions(User currentUser, User account, ReservedNamespace reservedNamespace)
+        {
+            return CheckPermissions(currentUser, account, GetReservedNamespaces(reservedNamespace));
+        }
+
         public PermissionsCheckResult CheckPermissions(IPrincipal currentPrincipal, User account, ActionOnNewPackageContext newPackageContext)
         {
             return CheckPermissions(currentPrincipal, account, GetReservedNamespaces(newPackageContext));
+        }
+
+        public PermissionsCheckResult CheckPermissions(IPrincipal currentPrincipal, User account, ReservedNamespace reservedNamespace)
+        {
+            return CheckPermissions(currentPrincipal, account, GetReservedNamespaces(reservedNamespace));
         }
 
         protected override PermissionsCheckResult CheckPermissionsForEntity(User account, IReadOnlyCollection<ReservedNamespace> reservedNamespaces)
@@ -54,9 +64,19 @@ namespace NuGetGallery
             return CheckPermissionsOnBehalfOfAnyAccount(currentUser, GetReservedNamespaces(newPackageContext));
         }
 
+        public PermissionsCheckResult CheckPermissionsOnBehalfOfAnyAccount(User currentUser, ReservedNamespace reservedNamespace)
+        {
+            return CheckPermissionsOnBehalfOfAnyAccount(currentUser, GetReservedNamespaces(reservedNamespace));
+        }
+
         public PermissionsCheckResult CheckPermissionsOnBehalfOfAnyAccount(User currentUser, ActionOnNewPackageContext newPackageContext, out IEnumerable<User> accountsAllowedOnBehalfOf)
         {
             return CheckPermissionsOnBehalfOfAnyAccount(currentUser, GetReservedNamespaces(newPackageContext), out accountsAllowedOnBehalfOf);
+        }
+
+        public PermissionsCheckResult CheckPermissionsOnBehalfOfAnyAccount(User currentUser, ReservedNamespace reservedNamespace, out IEnumerable<User> accountsAllowedOnBehalfOf)
+        {
+            return CheckPermissionsOnBehalfOfAnyAccount(currentUser, GetReservedNamespaces(reservedNamespace), out accountsAllowedOnBehalfOf);
         }
 
         protected override IEnumerable<User> GetOwners(IReadOnlyCollection<ReservedNamespace> reservedNamespaces)
@@ -67,6 +87,11 @@ namespace NuGetGallery
         private IReadOnlyCollection<ReservedNamespace> GetReservedNamespaces(ActionOnNewPackageContext newPackageContext)
         {
             return newPackageContext.ReservedNamespaceService.GetReservedNamespacesForId(newPackageContext.PackageId);
+        }
+
+        private IReadOnlyCollection<ReservedNamespace> GetReservedNamespaces(ReservedNamespace reservedNamespace)
+        {
+            return new[] { reservedNamespace };
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
```diff
@@ -15,7 +15,7 @@ namespace NuGetGallery
     /// E.g. "JQuery.Extensions.MyCoolExtension" matches both "JQuery.*" and "JQuery.Extensions.*".
     /// </remarks>
     public class ActionRequiringReservedNamespacePermissions
-        : ActionRequiringEntityPermissions<IReadOnlyCollection<ReservedNamespace>>, IActionRequiringEntityPermissions<ActionOnNewPackageContext>
+        : ActionRequiringEntityPermissions<IReadOnlyCollection<ReservedNamespace>>, IActionRequiringEntityPermissions<ActionOnNewPackageContext>, IActionRequiringEntityPermissions<ReservedNamespace>
```
chenriksson: nit: wrap long lines; makes it easier to review

src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
```diff
@@ -68,5 +88,10 @@ private IReadOnlyCollection<ReservedNamespace> GetReservedNamespaces(ActionOnNew
         {
             return newPackageContext.ReservedNamespaceService.GetReservedNamespacesForId(newPackageContext.PackageId);
         }
+
+        private IReadOnlyCollection<ReservedNamespace> GetReservedNamespaces(ReservedNamespace reservedNamespace)
+        {
+            return new[] { reservedNamespace };
```
chenriksson: Is there another way around this? Would prefer to avoid unnecessary allocations.

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
```diff
@@ -128,5 +128,21 @@ public static class ActionsRequiringPermissions
              new ActionRequiringPackagePermissions(
                 accountOnBehalfOfPermissionsRequirement: RequireOwnerOrOrganizationAdmin,
                 packageRegistrationPermissionsRequirement: RequireOwnerOrOrganizationAdmin);
+
+        /// <summary>
+        /// The action of showing a package as verified by a reserved namespace.
+        /// </summary>
+        public static ActionRequiringReservedNamespacePermissions ShowPackageAsVerifiedByReservedNamespace =
+            new ActionRequiringReservedNamespacePermissions(
+                accountOnBehalfOfPermissionsRequirement: PermissionsRequirement.Owner,
```
chenriksson: At first glance, this readonly permission seemed more restrictive than the Remove permission below... but seems like that's because we only invoke this on package owners. Maybe it's worth calling this out in the comments?

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
```diff
@@ -121,13 +121,13 @@ public IEnumerable<PackageOwnerRequest> GetPackageOwnershipRequests(PackageRegis
...
                     using (var transaction = _entitiesContext.GetDatabase().BeginTransaction())
                     {
-                        await RemovePackageOwnerImplAsync(packageRegistration, requestingOwner, ownerToBeRemoved);
+                        await RemovePackageOwnerImplAsync(packageRegistration, ownerToBeRemoved);
                         transaction.Commit();
                     }
                 }
                 else
                 {
-                    await RemovePackageOwnerImplAsync(packageRegistration, requestingOwner, ownerToBeRemoved);
```
chenriksson: We verify that `requestingOwner` has permissions for `ownerToBeRemoved` (organization) higher in the stack?

src/NuGetGallery/Services/ActionRequiringReservedNamespacePermissions.cs
scottbommarito: no, this system expects an array of reserved namespaces so we have to convert a single namespace into an array

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
scottbommarito: To clarify, this action is
```
If I own a package, I can show that package as verified by a reserved namespace that applies to it.
```

The reserved namespace must be owned by you and exactly you (you cannot show your package as verified on behalf of another user that owns a reserved namespace that applies to it), so the permissions requirements here are both `Owner`. It's a little convoluted but by using `ActionsRequiringPermissions` for this purpose we shouldn't have issues similar to what we saw here in the future.

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
scottbommarito: Yes, we call `OwnerHasPermissionsToRemove` earlier which does the verification.

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
```diff
@@ -195,27 +186,25 @@ public IEnumerable<PackageOwnerRequest> GetPackageOwnershipRequests(PackageRegis
...
+                packageRegistration.Owners.Count(
+                    o => ActionsRequiringPermissions.ShowPackageAsVerifiedByReservedNamespace
+                        .CheckPermissionsOnBehalfOfAnyAccount(o, reservedNamespaces) != PermissionsCheckResult.Allowed);
 
-        private static bool IsUserAnOwnerOfPackageNamespace(PackageRegistration packageRegistration, User user)
-        {
-            return packageRegistration.ReservedNamespaces.Any(rn => rn.Owners.Any(owner => owner == user));
+            // If multiple package owners own a reserved namespace that the package is a member of, then removing an owner will not remove the package from all of its reserved namespaces.
+            // If no package owners own a reserved namespace that the package is a member of, then the package is not in a reserved namespace and the owner can be removed.
+            return packageOwnersThatOwnAReservedNamespaceThatThePackageIsAMemberOf != 1;
```
shishirx34: I think this method is missing a case, you aren't checking the namespace ownership of `ownerToBeRemoved`, if the `requestingOwner` doesn't own the namespace, and neither does the `ownerToBeRemoved`, and there is a third owner who does own the namespace, this method will return `false`; which it shouldn't.

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
```diff
@@ -195,27 +186,25 @@ public IEnumerable<PackageOwnerRequest> GetPackageOwnershipRequests(PackageRegis
...
+            var packageOwnersThatOwnAReservedNamespaceThatThePackageIsAMemberOf =
+                packageRegistration.Owners.Count(
+                    o => ActionsRequiringPermissions.ShowPackageAsVerifiedByReservedNamespace
+                        .CheckPermissionsOnBehalfOfAnyAccount(o, reservedNamespaces) != PermissionsCheckResult.Allowed);
 
-        private static bool IsUserAnOwnerOfPackageNamespace(PackageRegistration packageRegistration, User user)
-        {
-            return packageRegistration.ReservedNamespaces.Any(rn => rn.Owners.Any(owner => owner == user));
+            // If multiple package owners own a reserved namespace that the package is a member of, then removing an owner will not remove the package from all of its reserved namespaces.
+            // If no package owners own a reserved namespace that the package is a member of, then the package is not in a reserved namespace and the owner can be removed.
```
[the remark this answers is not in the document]
scottbommarito: The comment is trying to explain how the number of package owners owning a reserving namespace is related to the package being in a reserved namespace...I think your comment removes this relationship. I'm not sure how to make the connection and also not have a confusing comment here.

src/NuGetGallery/Services/PackageOwnershipManagementService.cs
scottbommarito: Totally correct. The reason why this wasn't broken in my unit tests was because of the `!=` (instead of a `==`) in line 203. I have fixed and simplified the logic.

src/NuGetGallery/Services/ActionsRequiringPermissions.cs
```diff
@@ -128,5 +128,21 @@ public static class ActionsRequiringPermissions
              new ActionRequiringPackagePermissions(
                 accountOnBehalfOfPermissionsRequirement: RequireOwnerOrOrganizationAdmin,
                 packageRegistrationPermissionsRequirement: RequireOwnerOrOrganizationAdmin);
+
+        /// <summary>
+        /// The action of showing a package as verified by a reserved namespace.
+        /// </summary>
+        public static ActionRequiringReservedNamespacePermissions ShowPackageAsVerifiedByReservedNamespace =
```
shishirx34: I think this is a misnomer for the action. Rather rename it as `IsOwnerOfReservedNamespace` or a better action that represents ownership of namespace than the `ShowPackageAsVerified`, its not really the application of this action in the code below.

Issue #5924, opened 2018-05-14
Deleting an organization that has any API keys scoped to it fails


scottbommarito:
The organization deletion flow should delete any API keys scoped to it.

PR #5986, opened 2018-05-22
Delete API keys scoped to a user when that user is deleted

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5924

We must delete these API keys because
- they are no longer valid when the owner that they are scoped to is deleted
- there is a DB constraint that the user in the owner scope exists

### DIFF
--- a/src/NuGetGallery/Services/DeleteAccountService.cs
+++ b/src/NuGetGallery/Services/DeleteAccountService.cs
@@ -154,14 +154,6 @@ namespace NuGetGallery
                 .ToArray();
 
             var credentials = userScopes.Select(s => s.Credential).Distinct().ToArray();
-
-            foreach (var userScope in userScopes)
-            {
-                _scopeRepository.DeleteOnCommit(userScope);
-            }
-
-            await _scopeRepository.CommitChangesAsync();
-
             foreach (var credential in credentials)
             {
                 await _authService.RemoveCredential(credential.User, credential);
--- a/src/NuGetGallery/Services/DeleteAccountService.cs
+++ b/src/NuGetGallery/Services/DeleteAccountService.cs
@@ -24,11 +24,13 @@ namespace NuGetGallery
         private readonly ISecurityPolicyService _securityPolicyService;
         private readonly AuthenticationService _authService;
         private readonly IEntityRepository<User> _userRepository;
+        private readonly IEntityRepository<Scope> _scopeRepository;
         private readonly ISupportRequestService _supportRequestService;
         private readonly IAuditingService _auditingService;
 
         public DeleteAccountService(IEntityRepository<AccountDelete> accountDeleteRepository,
                                     IEntityRepository<User> userRepository,
+                                    IEntityRepository<Scope> scopeRepository,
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -138,26 +141,44 @@ IAuditingService auditingService
...
+                .GetAll()
+                .Where(s => s.OwnerKey == user.Key)
+                .ToArray();
+
+            var credentials = userScopes.Select(s => s.Credential).Distinct().ToArray();
+
+            foreach (var userScope in userScopes)
+            {
+                _scopeRepository.DeleteOnCommit(userScope);
+            }
```
chenriksson: I believe scopes are cascade deleted with the credential. Given this, you probably don't need to explicitly delete the scopes, and may not need the `EntityRepository<Scope>`.

Plus, this will allow them to be removed in the same commit as the credentials... which is safer.

scottbommarito: The fact that we're getting this error suggests they aren't. Perhaps a better approach is to change the DB schema?

scottbommarito: After online discussion, we no longer delete the scopes manually and just delete the credentials for each scope automatically.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -138,26 +141,36 @@ IAuditingService auditingService
...
+            foreach (var uc in user.Credentials.ToArray())
             {
                 await _authService.RemoveCredential(user, uc);
             }
+
+            // Remove any credential scoped to this user.
+            var userScopes = _scopeRepository
+                .GetAll()
+                .Where(s => s.OwnerKey == user.Key)
+                .ToArray();
```
agr: Unnecessary `ToArray` call? Next line does it again.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -138,26 +141,36 @@ IAuditingService auditingService
 
         private async Task RemoveUserCredentials(User user)
         {
-            var copyOfUserCred = user.Credentials.ToArray();
-            foreach (var uc in copyOfUserCred)
+            // Remove any credential owned by this user.
+            foreach (var uc in user.Credentials.ToArray())
```
agr: Why do we need all those `ToArray` calls everywhere? Is there a chance that enumerating the `user` properties would result into query to be run?

scottbommarito: We are deleting from `user.Credentials`, so we must create a copy of the array or else deleting from the array will cause `Collection was modified` exceptions.

chenriksson: Considering adding an `IAuthenticationService.RemoveAllCredentials` API, which would prevent the need for `ToArray` and allow for a single commit. That API could also be smart enough to handle deletion of credentials owned by OR scoped to the account.

PR #6001, opened 2018-05-25
Storage:  enable setting blob metadata

dtivel:
Resolve https://github.com/NuGet/NuGetGallery/issues/6000.


src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -328,6 +328,47 @@ private static string Log(AccessCondition accessCondition)
                 blob.GetSharedAccessSignature(SharedAccessBlobPermissions.Read, endOfAccess));
         }
 
+        public async Task SetMetadataAsync(
```
cristinamanum: I think that comments would useful to set the correlation between the updateMetadataAsync and the SetMetadataAsync.

Issue #5288, opened 2018-01-18
Hosting own Nuget with ActiveDirectory Authentication


stefan-hochi:
Is there an Implementation for Active Directory currently available? In Issue #554 write you are interessted on multiple authentication consists and Issue has been closed
has someone Implement there own Active Directory with Current Release
or is there an workaround available

PR #6003, opened 2018-05-29
Added LDAP support

### LINKED ISSUE #554: NuGet Gallery should support LDAP for on-premise deployments
For those who use NuGet Gallery on-premise (rather than just nuget.org), it'd be a big win for the Gallery to support LDAP for authentication rather than (or in addition to) the NuGet Gallery database authentication store.

Falco20019:
Fixes #554, #5288

Since I also needed it, I added support for LDAP / Windows Active Directory. I was only able to test it within our company with our companies AD. To use it, you have to configure the `Auth.LdapUser` block in the `Web.config` file.

When `Auth.LdapUser.Enabled`, the register form is used to check the given credentials (username and password) and if the credentials are valid, it creates a new account with the username as `ldap.user` credentials instead of saving the password with `password.v3` credentials. No password credentials will be created for this account. If you want to use `LdapUser` with the current password as `LocalUser` fallback, you can just add the credentials to the account in `AuthenticationService.cs` right after the `CreateIdentity` was called (around line 252). This may be something to make configurable.

If you only want to support LDAP without regular passwords (i.e. when you are setting up a new server), I advise you to set `Auth.LocalUser.Enabled` in `Web.config` to `false`. They can coexist, since they both have an own cookie, but in most cases, you really only want to have either LDAP or local users.

Right now, there is no migration path to change existing user credentials from `password.v3` to `ldap.user`. In case the NuGet username already matches the LDAP username, the easiest way would be to update the `dbo.Credentials` entries for `password.v3` and set `Type`= `ldap.user` and `Value` = `(SELECT [Username] FROM [Users] WHERE [Key] = [UserKey])`.

**Tests are currently missing. I didn't want to put in the effort in case that it won't be merged. So I will add them, in case that there are no points speaking against merging.**

### DIFF
--- a/src/NuGetGallery.Core/CredentialTypes.cs
+++ b/src/NuGetGallery.Core/CredentialTypes.cs
@@ -9,6 +9,12 @@ namespace NuGetGallery
 {
     public static class CredentialTypes
     {
+        public static class Ldap
+        {
+            public const string Prefix = "ldap.";
+            public const string User = Prefix + "user";
+        }
+
         public static class Password
         {
             public const string Prefix = "password.";
@@ -32,6 +38,11 @@ namespace NuGetGallery
             public const string Prefix = "external.";
             public const string MicrosoftAccount = Prefix + "MicrosoftAccount";
             public const string AzureActiveDirectoryAccount = Prefix + "AzureActiveDirectory";
+        }
+
+        public static bool IsLdap(this Credential c)
+        {
+            return c?.Type?.StartsWith(Ldap.Prefix, StringComparison.OrdinalIgnoreCase) ?? false;
         }
 
         public static bool IsPassword(this Credential c)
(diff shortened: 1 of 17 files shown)


awickham10:
Do we know when this will be accepted? I could use this as well.

DjArt:
So?

joelverhagen:
Given the age of this PR, I'm going to close it. 

I think the main concern I have about this PR is that it adds complexity to the codebase with a scenario we do not plan to use directly on NuGet.org. If the code can be factored in such a way that the ongoing maintenance costs are low, then we can consider it. In other words, we can certainly accept a contribution like this in the future, but we need to be careful not to add overhead to our team's main objective -- the improvement and maintenance of NuGet.org itself.

I wonder if there is any way to re-use NuGetGallery's existing MSA/AAD integration with LDAP. I don't know how Azure Active Directory and enterprise (Windows?) Active Directory overlap.

If anyone is interested in pushing this change through, I think it would be worth going through the formal proposal as described here: https://github.com/NuGet/Home/tree/dev/meta#nuget-proposal-process

This would allow us to align on the direction first, and later jump into the code.

PR #6006, opened 2018-05-30
Replace PoliteCaptcha

chenriksson:
Replacing old, unmaintained PoliteCaptcha reference with invisible recaptcha v2.

This is required for moving to PackageRefs. I'm prioritizing the move to PackageRefs as part of the SQL AAD work, since I need to touch all dependencies as a result of retargeting projects.

### DIFF
--- a/src/NuGetGallery/Filters/ValidateRecaptchaResponse.cs
+++ b/src/NuGetGallery/Filters/ValidateRecaptchaResponse.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Net.Http;
+using System.Threading.Tasks;
+using System.Web.Mvc;
+using Newtonsoft.Json;
+
+namespace NuGetGallery.Filters
+{
+    public sealed class ValidateRecaptchaResponse : AuthorizeAttribute
+    {
+        private const string RecaptchaResponseId = "g-recaptcha-response";
+        private const string RecaptchaValidationUrl = "https://www.google.com/recaptcha/api/siteverify?secret={0}&response={1}";
+        private static readonly TimeSpan RequestTimeout = TimeSpan.FromSeconds(10);
+
+        public override void OnAuthorization(AuthorizationContext filterContext)
+        {
+            var controller = filterContext.Controller as AppController;
+            var privateKey = controller.NuGetContext.Config.Current.ReCaptchaPrivateKey;
+            
+            if (!string.IsNullOrEmpty(privateKey))
+            {
+                var response = controller.HttpContext.Request.Form[RecaptchaResponseId];
+                if (!string.IsNullOrEmpty(response))
+                {
+                    if (!(Task.Run(() => RecaptchaIsValid(privateKey, response)).Result))
+                    {
+                        controller.TempData["Message"] = Strings.InvalidRecaptchaResponse;
+                        controller.SafeRedirect(controller.Url.Home());
+                    }
+                }
+            }
+
+            base.OnAuthorization(filterContext);
+        }
+
+        private async Task<bool> RecaptchaIsValid(string privateKey, string response)
+        {
+            var validationUrl = string.Format(CultureInfo.InvariantCulture, RecaptchaValidationUrl, privateKey, response);
+            using (var client = new HttpClient() { Timeout = RequestTimeout })
+            {
+                var reply = await client.GetStringAsync(validationUrl);
+                var state = JsonConvert.DeserializeObject<RecaptchaState>(reply);
+                return state.Success;
+            }
+        }
+
+        public class RecaptchaState
+        {
+            [JsonProperty("success")]
+            public bool Success { get; set; }
+
+            [JsonProperty("error-codes")]
+            public List<string> ErrorCodes { get; set; }
+        }
+    }
+}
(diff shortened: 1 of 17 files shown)


skofman1:
Can u explain how recaptch works? I don't see any UI on the "report abuse" page, so confused.. also, I see recaptcha v2 doen't need any packages. What was the original package doing? Why did we need it in the first place?

chenriksson:
@skofman1 
Recaptcha uses a client-side risk analysis engine to determine whether a user is a bot. The risk analysis likely evaluates client-side events such as form interaction, mouse movement and clicks, etc to make its decision. The default v2 recaptcha includes an explicit check, whereas the invisible recaptcha only prompts if the risk analysis doesn't have enough information.

see:
- https://security.googleblog.com/2014/12/are-you-robot-introducing-no-captcha.html
- https://nakedsecurity.sophos.com/2016/12/06/are-you-human-or-a-bot-googles-invisible-recaptcha-will-decide

In my testing, I've sometimes seen the client-side prompt on my initial 'Contact Support' submit and is the mobile-based version:
image

Our original PoliteCaptcha package first checks user agent to determine if the user is a bot, then falls back to recatpcha, v1. However, we're not setting recaptcha keys at deployment so there currently is no fallback.

see:
- https://github.com/NuGet/PoliteCaptcha

Since recaptcha is getting less invasive, and because I didn't think we wanted to keep maintaining PoliteCaptcha I decided to go with this approach. There is a recaptcha v3 now that may be less invasive still, but it may introduce some complexity as well.

chenriksson:
Note that the use of recaptcha does require user notification of data collection via privacy policy. We should be good here, since we only integrate recaptcha on forms that require authentication... meaning, that the user has already given cookie consent by navigating the site and should be aware of our privacy policy.

Found this post useful on the topic:
- https://termsfeed.com/blog/privacy-policy-recaptcha/

src/NuGetGallery/Filters/ValidateRecaptchaResponse.cs
```diff
@@ -0,0 +1,61 @@
...
+                }
+            }
+
+            base.OnAuthorization(filterContext);
+        }
+
+        private async Task<bool> RecaptchaIsValid(string privateKey, string response)
+        {
+            var validationUrl = string.Format(CultureInfo.InvariantCulture, RecaptchaValidationUrl, privateKey, response);
+            using (var client = new HttpClient() { Timeout = RequestTimeout })
```
agr: Should we try to [reuse](https://docs.microsoft.com/en-us/azure/architecture/antipatterns/improper-instantiation/) instances of `HttpClient` instead of instantiating one for each request?

shishirx34:
@chenriksson - report abuse form does not need authentication. Would we need privacy policy update then?

chenriksson:
@skofman1 Updated to add recaptcha to Contact Owners form per https://github.com/NuGet/Engineering/issues/1095

src/NuGetGallery/Filters/ValidateRecaptchaResponse.cs
```diff
@@ -0,0 +1,61 @@
...
+            
+            if (!string.IsNullOrEmpty(privateKey))
+            {
+                var response = controller.HttpContext.Request.Form[RecaptchaResponseId];
+                if (!string.IsNullOrEmpty(response))
+                {
+                    if (!(Task.Run(() => RecaptchaIsValid(privateKey, response)).Result))
+                    {
+                        controller.TempData["Message"] = Strings.InvalidRecaptchaResponse;
+                        controller.SafeRedirect(controller.Url.Home());
```
shishirx34: why not redirect to the origin url?

src/NuGetGallery/Filters/ValidateRecaptchaResponse.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            base.OnAuthorization(filterContext);
+        }
+
+        private async Task<bool> RecaptchaIsValid(string privateKey, string response)
+        {
+            var validationUrl = string.Format(CultureInfo.InvariantCulture, RecaptchaValidationUrl, privateKey, response);
+            using (var client = new HttpClient() { Timeout = RequestTimeout })
+            {
+                var reply = await client.GetStringAsync(validationUrl);
```
shishirx34: `Try..Catch` for `HttpTimeoutException`? Could return `false` in case of exceptions.

src/NuGetGallery/Filters/ValidateRecaptchaResponse.cs
```diff
@@ -0,0 +1,61 @@
...
+                        controller.TempData["Message"] = Strings.InvalidRecaptchaResponse;
+                        controller.SafeRedirect(controller.Url.Home());
+                    }
+                }
+            }
+
+            base.OnAuthorization(filterContext);
+        }
+
+        private async Task<bool> RecaptchaIsValid(string privateKey, string response)
```
shishirx34: Don't see using `response` here, is it needed?

src/NuGetGallery/Strings.resx
```diff
@@ -905,4 +905,7 @@ If you would like to update the linked Microsoft account you can do so from the
   <data name="DiscontinuedLogin_PasswordRemoved" xml:space="preserve">
     <value>The password login is discontinued and has been removed for your account. Please use your Microsoft account to log into {0} going forward.</value>
   </data>
+  <data name="InvalidRecaptchaResponse" xml:space="preserve">
+    <value>Recaptcha response was not valid. Email support if you need assistance.</value>
```
shishirx34: May be suggest, to retry again? `Recaptcha response was not valid. Please retry submitting the form or email support if you need assistance.`.

Issue #5778, opened 2018-04-09
HTTP 500 is thrown when documentation markdown URL is not raw.github.com


joelverhagen:
1. Go to a package you own
1. Edit package
1. Provide a URL to a markdown document that is not on raw.github.com
    - example: https://github.com/adam-p/markdown-here/blob/master/README.md
1. Submit

![image](https://user-images.githubusercontent.com/94054/38510985-803d4cea-3bdb-11e8-82f0-1b7f059e8ae7.png)

PR #6023, opened 2018-06-04
Handle ArgumentNullException and HTTP 500 when readme host is invalid.

xavierdecoster:
Fixes https://github.com/NuGet/NuGetGallery/issues/5778

### DIFF
--- a/tests/NuGetGallery.Facts/Services/ReadMeServiceFacts.cs
+++ b/tests/NuGetGallery.Facts/Services/ReadMeServiceFacts.cs
@@ -151,6 +151,33 @@ namespace NuGetGallery
                     x => x.SaveChangesAsync(),
                     commitChanges ? Times.Once() : Times.Never());
             }
+
+            [Theory]
+            [InlineData(true)]
+            [InlineData(false)]
+            public async Task ThrowsArgumentExceptionWhenReadmeUrlHostInvalid(bool commitChanges)
+            {
+                // Arrange
+                _package.HasReadMe = true;
+                _packageFileService.Setup(m => m.DownloadReadMeMdFileAsync(_package)).ReturnsAsync((string)null);
+                _edit.ReadMe = new ReadMeRequest { SourceUrl = "https://github.com/username/markdown-here/blob/master/README.md", SourceType = "url" };
+                _edit.ReadMeState = PackageEditReadMeState.Changed;
+
+                // Act
+                var saveTask = _target.SaveReadMeMdIfChanged(
+                    _package,
+                    _edit,
+                    _encoding,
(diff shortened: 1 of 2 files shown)


loic-sharma:
There's another bug in https://github.com/NuGet/NuGetGallery/issues/5778. If [the host requirement](https://github.com/NuGet/NuGetGallery/blob/af49d214a2316faeb626f09be1d1995a6a2b4c0d/src/NuGetGallery/Services/ReadMeService.cs#L29) isn't met, an [`ArgumentException` is thrown](https://github.com/NuGet/NuGetGallery/blob/af49d214a2316faeb626f09be1d1995a6a2b4c0d/src/NuGetGallery/Services/ReadMeService.cs#L298) and not caught. I would fix that too, and add tests.

xavierdecoster:
@loic-sharma that's exactly the bug being fixed here. The `ArgumentNullException` was causing the `ArgumentException` not to be thrown.

The `ArgumentException` thrown when url host is invalid is handled in the calling [`PackagesController`](https://github.com/NuGet/NuGetGallery/blob/af49d214a2316faeb626f09be1d1995a6a2b4c0d/src/NuGetGallery/Controllers/PackagesController.cs#L1274).

Added unit test theory to cover this case.

tests/NuGetGallery.Facts/Services/ReadMeServiceFacts.cs
```diff
@@ -151,6 +151,33 @@ public TheSaveReadMeMdIfChangedMethod()
...
+                var saveTask = _target.SaveReadMeMdIfChanged(
+                    _package,
+                    _edit,
+                    _encoding,
+                    commitChanges);
+
+                // Assert
+                var exception = await Assert.ThrowsAsync<ArgumentException>(() => saveTask);
+
+                Assert.True(exception.Message.Contains(Strings.ReadMeUrlHostInvalid));
```
agr: `Assert.Contains`?

loic-sharma: This is what the controller checks, so I think this would be a good assertion. See: https://github.com/NuGet/NuGetGallery/blob/4cc71844cd5dca7edd9af36a569839f354173dca/src/NuGetGallery/Controllers/PackagesController.cs#L1618

loic-sharma:
@xavierdecoster Ah, I see. My bad :)

tests/NuGetGallery.Facts/Services/ReadMeServiceFacts.cs
xavierdecoster: Fixed

Issue #5471, opened 2018-02-13
Show package size in Gallery UI


skofman1:
Similar to this issue: https://github.com/NuGet/NuGetGallery/issues/2957 but only for Gallery

PR #6024, opened 2018-06-04
Show package size in Gallery UI Package Page

xavierdecoster:
Addresses #5471

Preview Screenshot:

image

If there's a better icon let me know.

### DIFF
--- a/src/NuGetGallery/Extensions/NumberExtensions.cs
+++ b/src/NuGetGallery/Extensions/NumberExtensions.cs
@@ -2,10 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.Globalization;
-using System.Linq;
-using System.Web;
 
 namespace NuGetGallery
 {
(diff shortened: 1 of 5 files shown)


anangaur:
I would club this with "Manual download" (renamed to just "Download"). Eg:
image

src/NuGetGallery/Extensions/NumberExtensions.cs
```diff
@@ -36,5 +33,28 @@ public static string ToNuGetNumberString(this long self)
...
+        {
+            const int scale = 1024;
+            string[] orders = { "GB", "MB", "KB", "bytes" };
+            var max = (long)Math.Pow(scale, orders.Length - 1);
+
+            foreach (var order in orders)
+            {
+                if (bytes >= max)
+                {
+                    return string.Format(CultureInfo.CurrentCulture, "{0:##.##} {1}", decimal.Divide(bytes, max), order);
```
agr: Need a special case for `1 byte`?

src/NuGetGallery/Extensions/NumberExtensions.cs
```diff
@@ -36,5 +33,28 @@ public static string ToNuGetNumberString(this long self)
             return self.ToString("n0", CultureInfo.CurrentCulture);
         }
 
+        /// <summary>
+        /// Format the number of bytes into a user-friendly display label.
+        /// </summary>
+        /// <param name="bytes"></param>
+        /// <returns></returns>
+        public static string ToUserFriendlyBytesLabel(this long bytes)
+        {
```
agr: Missing range check for negative numbers.

src/NuGetGallery/Extensions/NumberExtensions.cs
agr: Also, should we format with `CurrentCulture`? The web site is in English-only, why would we format numbers using user-specific locale? That sounds confusing: for example Russian users would see comma as a decimal separator in otherwise English web site.

agr: Why not `{0:F2}` format specifier?

xavierdecoster: Because `{0:##.##}` will strop off trailing zeros, to avoid e.g. `1.00 KB`, `512.00 bytes`, etc.

xavierdecoster: Changed to using `InvariantCulture` instead.

src/NuGetGallery/Extensions/NumberExtensions.cs
xavierdecoster: Added.

src/NuGetGallery/Extensions/NumberExtensions.cs
xavierdecoster: Added.

xavierdecoster:
Added tests for the extension method.

Issue #5728, opened 2018-04-02
Add additional links to upload page


scottbommarito:
On the profile page, there should be a message stating that the user has no packages, and if the current user is that user, a link to the upload page

On the manage packages page, there should be an "upload" button similar to the "add new organization" button

anangaur:
![image](https://user-images.githubusercontent.com/14800916/38215706-86b19198-367d-11e8-93e9-2cdd88c50661.png)


![image](https://user-images.githubusercontent.com/14800916/38215709-895dcccc-367d-11e8-9160-c7025c0ef253.png)

agr:
There is already an upload link on the page (And it's in direct vicinity of the proposed one). Why do we need another one?

loic-sharma:
@anangaur The pages' content looks empty. I think it'd look better if the page's content was some image followed by a "call-to-action" link (like "Upload a new package", or "Create a new organization")

anangaur:
@agr Many users do not expect "Upload" as a menu item and did not discover that its a good idea to show all the CRUD operations at one place on the list objects page :)

anangaur:
@loic-sharma I agree we do not have a blank page filler for the packages page. Created an issue: https://github.com/NuGet/NuGetGallery/issues/6031

PR #6029, opened 2018-06-04
Add a "Add new (upload)" link to the Manage Packages view breadcrumb navigation

xavierdecoster:
Addresses #5728 

image

### DIFF
--- a/src/Bootstrap/less/theme/page-manage-packages.less
+++ b/src/Bootstrap/less/theme/page-manage-packages.less
@@ -1,7 +1,8 @@
 .page-manage-packages {
   @section-margin-top: 40px;
 
-  .organizations-divider {
+  .organizations-divider,
+  .packages-divider {
     color: @gray-lighter;
   }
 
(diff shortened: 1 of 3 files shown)


xavierdecoster:
I have the feeling there's more to it than just adding this link, but happy to make this PR incremental.

@anangaur Can you please provide what are the additional requirements for the below screenshot? 
e.g. noticing a dropdown to the right

image

anangaur:
The drip down to the right is the owner filter that is not inline with the site theme today. The width of the filter drop down should be increased and the drop down button should be different than what is present today. We should use the same icon for dropdiwn as we use for expandable sections on NuGet.org

agr:
Why do we need two links to the same upload page there?

scottbommarito:
@anangaur I think `Upload new` without "add" would look a lot better.
Also, I think dropdown changes should be made in a separate PR--the issue that this PR is targeting is very specifically focused on upload links and not other UI.

anangaur:
I would say just `+ Add new` is sufficient. It will anyway take you to the upload page.

xavierdecoster:
Updated:

image

image

Issue #5850, opened 2018-04-27
Experience Improvement: On package submit failure due to validation pipeline being down


cristinamanum:
In the latest BCDR drill it was observed that when validation service bus is down(it was intentionally deleted) on the submit package page the page does not show any error and it stays in the same view as it was not submitted. 
The user does not have any information about what state its package is in.

We may consider adding information back to the user about the state.

PR #6038, opened 2018-06-07
Minor fixes for verify page.

ryuyu:
This addresses most of https://github.com/NuGet/NuGetGallery/issues/5850

* Adds AI logging for any failure fed back to gallery during validation flow.
* Scrolls error message container into view on Upload page when Submit is pressed (if there are any errors to show). scrollIntoView tested on IE, Edge, Firefox, Chrome.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1638,50 +1638,58 @@ namespace NuGetGallery
 
                 // Commit the package to storage and to the database.
                 uploadFile.Position = 0;
-                var commitResult = await _packageUploadService.CommitPackageAsync(
-                    package,
-                    uploadFile.AsSeekableStream());
-
-                switch (commitResult)
+                try
                 {
-                    case PackageCommitResult.Success:
-                        break;
-                    case PackageCommitResult.Conflict:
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1638,50 +1638,58 @@ private void SendAddPackageOwnerNotification(PackageRegistration package, User n
...
-            });
+                    TempData["Message"] = String.Format(
+                        CultureInfo.CurrentCulture, Strings.SuccessfullyUploadedPackage, package.PackageRegistration.Id, package.Version);
+
+                    return Json(new
+                    {
+                        location = Url.Package(package.PackageRegistration.Id, package.NormalizedVersion)
+                    });
+                }
+                catch (Exception e)
```
shishirx34: is it possible to catch any specific exceptions here before general exception, so that we can give customized message?

loic-sharma: Also, maybe we should propagate the error message out of `UserSafeException`?

ryuyu: It's definitely possible. We just need to enumerate which cases we want specific error messages for.

PR #6041, opened 2018-06-08
Fix DisplayPackage perf regression

chenriksson:
Fixes https://github.com/NuGet/Engineering/issues/1453

### DIFF
--- a/src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
+++ b/src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
@@ -34,6 +34,7 @@ namespace NuGetGallery
                 TotalDaysSinceCreated = Convert.ToInt32(Math.Max(1, Math.Round((DateTime.UtcNow - packageHistory.Min(p => p.Created)).TotalDays)));
                 DownloadsPerDay = TotalDownloadCount / TotalDaysSinceCreated; // for the package
                 DownloadsPerDayLabel = DownloadsPerDay < 1 ? "<1" : DownloadsPerDay.ToNuGetNumberString();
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -34,6 +34,7 @@ public DisplayPackageViewModel(Package package, User currentUser, IOrderedEnumer
                 TotalDaysSinceCreated = Convert.ToInt32(Math.Max(1, Math.Round((DateTime.UtcNow - packageHistory.Min(p => p.Created)).TotalDays)));
                 DownloadsPerDay = TotalDownloadCount / TotalDaysSinceCreated; // for the package
                 DownloadsPerDayLabel = DownloadsPerDay < 1 ? "<1" : DownloadsPerDay.ToNuGetNumberString();
+                IsDotnetToolPackageType = package.PackageTypes.Any(e => e.Name.Equals("DotnetTool", StringComparison.OrdinalIgnoreCase));
```
dtivel: How does this [constructor call](https://github.com/NuGet/NuGetGallery/pull/6041/files#diff-c3acedea891434d6c61d815a4e9c54fdR24), which is called once for each package version, not also cause the perf regression even with your fix?

chenriksson: It's wrapped in `if (packageHistory.Any())`, so only executed for the latest version.

chenriksson: Also, `PackageViewModel` is shared by multiple views, so this regression could have affected other pages. Moving to `DisplayPackageViewModel` limits it to just the `DisplayPackage.cshtml` where it's used.

Issue #6039, opened 2018-06-08
Package remained verified after the organization owning the namespace was removed as owner of the package


cristinamanum:
See PR: https://github.com/NuGet/NuGetGallery/pull/5984

The package is still verified after the organization owning the namespace was removed as the owner of the package. 
Observation: Everything in database seems to be cleaned.
1. The package has only one owner that is not the org or the org admin.
2. The namespace "cmaad" is reserved by the cmanutestorg
3. The ReservedNamespaceRegistrations does not have a reference to the package that is verified. 

Few UI snapshots:
![image](https://user-images.githubusercontent.com/16580006/41133712-640ae41c-6a7d-11e8-8277-302c78f0d286.png)

![image](https://user-images.githubusercontent.com/16580006/41133731-8586e776-6a7d-11e8-8947-21c0414aa193.png)

![image](https://user-images.githubusercontent.com/16580006/41133765-b43b39aa-6a7d-11e8-8727-def1ab7918da.png)

shishirx34:
Looks like the `IsVerified` flag isn't cleared from the `PackageRegistrations` table after removing organization that owns it. @scottbommarito worked on it recently to manage and cleanup registrations for namespace for org owner removal.

shishirx34:
There is a check for removing the verified flag: https://github.com/NuGet/NuGetGallery/blob/4cc71844cd5dca7edd9af36a569839f354173dca/src/NuGetGallery/Services/PackageOwnershipManagementService.cs#L159 are you sure that the package registration doesn't have any other matching/related namespace?

shishirx34:
Oh, I think I see the problem, the issue is on the same line mentioned above, however, the previous call to `RemovePackageRegistrationFromNamespace()` to remove the relevant `PackageRegistration` from `ReservedNamespace` is not `committed`, and its likely the `!packageRegistration.ReservedNamespaces.Any()` is returning `false`, thereby not removing the `IsVerified` flag, I will let @scottbommarito fix the issue, since it is likely this has far reaching effects due to the above said refactoring.

scottbommarito:
Currently `RemovePackageRegistrationFromNamespace` removes the `PackageRegistration` from the `ReservedNamespace` but does not remove the `ReservedNamespace` from the `PackageRegistration`. It needs to do both for the check you mentioned above to work.

The unit tests do not need to change because they already assume `RemovePackageRegistrationFromNamespace` does both.

PR #6043, opened 2018-06-11
Correctly unverify packages when they are removed from their reserved namespaces

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6039

### DIFF
--- a/src/NuGetGallery/Services/ReservedNamespaceService.cs
+++ b/src/NuGetGallery/Services/ReservedNamespaceService.cs
@@ -296,6 +296,7 @@ namespace NuGetGallery
             }
 
             reservedNamespace.PackageRegistrations.Remove(packageRegistration);
+            packageRegistration.ReservedNamespaces.Remove(reservedNamespace);
         }
 
         public virtual ReservedNamespace FindReservedNamespaceForPrefix(string prefix)


src/NuGetGallery/Services/ReservedNamespaceService.cs
```diff
@@ -296,6 +296,7 @@ public void RemovePackageRegistrationFromNamespace(ReservedNamespace reservedNam
             }
 
             reservedNamespace.PackageRegistrations.Remove(packageRegistration);
+            packageRegistration.ReservedNamespaces.Remove(reservedNamespace);
```
chenriksson: Both are necessary?

Do we have other cases where this is broken? e.g.:
https://github.com/NuGet/NuGetGallery/blob/86b2646f255ba2f9f60fa5a1c1e57df2c0983209/src/NuGetGallery/Services/ReservedNamespaceService.cs#L238

scottbommarito: When you have a relation between two objects and you remove the relation using one object, the relation from the other object to this object remains intact until the next commit.

E.g.

```
public class Example1
{
    public ICollection<Example> Examples { get; set; }
}

public class Example2
{
    public ICollection<Example> Examples { get; set; }
}
```
```
modelBuilder.Entity<Example1>()
    .HasMany(e => e.Examples)
    .WithRequired(e => e.Examples);

modelBuilder.Entity<Example2>()
    .HasMany(e => e.Examples)
    .WithRequired(e => e.Examples);
```
```
var a = new Example1();
var b = new Example2();
a.Examples.Add(b);
b.Examples.Add(a);
await CommitChangesAsync(); // a <-> b
a.Examples.Remove(b); // a <- b
await CommitChangesAsync(); // a     b
```

TBH I think Entity Framework should support deleting both sides automatically (and it might in a more recent version).

The code you linked doesn't check the entity relations after the relation is deleted from one side.
My code depends on the relation being deleted from both sides before the commit.

In general, the behavior is only broken if the changes are not committed before the function exits.

scottbommarito:
Closing because we want to open this as a hotfix.

PR #6047, opened 2018-06-11
Correctly unverify packages when they are removed from their reserved namespaces (HOTFIX)

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6039

(also see https://github.com/NuGet/NuGetGallery/pull/6043)


loic-sharma:
Unit test?

scottbommarito:
@loic-sharma done

PR #6064, opened 2018-06-13
Defer open of DB connection to EF context

chenriksson:
Depends on https://github.com/NuGet/ServerCommon/pull/162
Fixes https://github.com/NuGet/Engineering/issues/1472

Alternative to https://github.com/NuGet/NuGetGallery/pull/6063, but I think we should do both:
- Don't connect to validation DB unless feature flag is enabled
- Defer DB connection open to the EF context

### DIFF
--- a/src/NuGetGallery/NuGetGallery.csproj
+++ b/src/NuGetGallery/NuGetGallery.csproj
@@ -2168,7 +2168,7 @@
       <Version>2.26.0-master-33065</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Sql">
-      <Version>2.26.0-master-33065</Version>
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -2168,7 +2168,7 @@
       <Version>2.26.0-master-33065</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Sql">
-      <Version>2.26.0-master-33065</Version>
+      <Version>2.26.0-chenriks-sqlconnection-33142</Version>
```
chenriksson: Will update to master version once ServerCommon PR is merged.

Issue #4678, opened 2017-09-14
Consider adding an "auto delete" feature to handle failed validations


joelverhagen:
Validation orchestrator should have a feature that means failed validations result in package deletion.

skofman1:
We should allow package authors to re-upload a package that failed certain kinds of validation, like signature validation.

rido-min:
Packages failing validation should not appear in `PackageRegistrations` or `Packages` tables. So "auto delete" or "re-upload" gestures are not required. Just allow submitting again.

Related to #6014 #6008, #6009 and #6010

onovotny:
When is this hitting prod?

scottbommarito:
@onovotny - It's live now!

PR #6065, opened 2018-06-13
Allow packages that failed validation to be reuploaded

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/4678


src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -440,22 +442,22 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+                            var existingPackage = PackageService.FindPackageByIdAndVersionStrict(nuspec.GetId(), nuspecVersion.ToStringSafe());
+                            if (existingPackage != null)
                             {
-                                return new HttpStatusCodeWithBodyResult(
-                                    HttpStatusCode.Conflict,
-                                    string.Format(CultureInfo.CurrentCulture, Strings.PackageExistsAndCannotBeModified,
-                                        id, nuspec.GetVersion().ToNormalizedStringSafe()));
+                                if (existingPackage.PackageStatusKey == PackageStatus.FailedValidation)
+                                {
+                                    await PackageDeleteService.HardDeletePackagesAsync(new[] { existingPackage }, currentUser, "", "", false);
```
chenriksson: If this was the initial version, are we still applying the same permissions check (i.e., as if the registration doesn't exist)?

chenriksson: What will the performance impact be?
- Could hard delete cause this request to timeout?
- Will the hard delete complete enough of the V3 pipeline so that there are no conflicts w/ the re-push?

Did you consider if we could treat as revalidate (put new nupkg in validation container and overwrite)? What are the advantages to this option?

scottbommarito: Yes, those checks are done directly above this one.

chenriksson: But for initial versions... the first attempt would have registrationExists=false and retry would have registrationExists=true. Shouldn't we still require org admin to push the retry of initial versions?

scottbommarito: 1 - I doubt it. Hard deleting a package isn't particularly complex--it consists of deleting a couple rows from the DB and then deleting from the packages and validation containers.
2 - Am testing this on dev now, but given that Feed2Catalog commits deletes and then creates there should be no issue.

It's not a revalidate because it could completely replace the package's metadata in the DB. Revalidating doesn't change the package's metadata at all. I believe hard delete much more closely duplicates this scenario.

scottbommarito: Oh I see. A couple comments:
- We no longer require organization admin to push a new ID, so the permissions check is the same.
- I think reuploading a package should be considered identical to uploading a new version of an existing package for permissions purposes. The package registration has already been claimed.

scottbommarito: Verified this fix on DEV. It works with V3 as expected. There is an extra `PackageDelete` entry in the catalog but it is always followed by a `PackageDetails` entry.

After offline discussion I will change `PackageDeleteService` to not emit an auditing record in this case.

chenriksson: Why not audit the reupload? Wouldn't a second upload be as important for security audits, especially if a different user uploads?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -440,22 +442,22 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+                            var existingPackage = PackageService.FindPackageByIdAndVersionStrict(nuspec.GetId(), nuspecVersion.ToStringSafe());
+                            if (existingPackage != null)
                             {
-                                return new HttpStatusCodeWithBodyResult(
-                                    HttpStatusCode.Conflict,
-                                    string.Format(CultureInfo.CurrentCulture, Strings.PackageExistsAndCannotBeModified,
-                                        id, nuspec.GetVersion().ToNormalizedStringSafe()));
+                                if (existingPackage.PackageStatusKey == PackageStatus.FailedValidation)
+                                {
+                                    await PackageDeleteService.HardDeletePackagesAsync(new[] { existingPackage }, currentUser, "", "", deleteEmptyPackageRegistration: false, saveAuditRecord: false);
```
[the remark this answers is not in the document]
shishirx34: I think it would be wise to put this in a `try..catch` and show meaningful error on any exception(likely timeout, db connection issue) suggestion could be to retry.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -360,33 +360,41 @@ public virtual JsonResult UploadPackageProgress()
                 var existingPackage = _packageService.FindPackageByIdAndVersionStrict(nuspec.GetId(), nuspecVersion.ToStringSafe());
                 if (existingPackage != null)
                 {
-                    // Determine if the package versions only differ by metadata, 
-                    // and provide the most optimal the user-facing error message.
-                    var existingPackageVersion = new NuGetVersion(existingPackage.Version);
-                    String message = string.Empty;
-                    if ((existingPackageVersion.HasMetadata || nuspecVersion.HasMetadata)
-                        && !string.Equals(existingPackageVersion.Metadata, nuspecVersion.Metadata))
+                    if (existingPackage.PackageStatusKey == PackageStatus.FailedValidation)
```
shishirx34: This method is a monster, unrelated to your change. I think we should refactor this, guess I will do it with my symbol server changes.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -360,33 +360,41 @@ public virtual JsonResult UploadPackageProgress()
...
                     {
-                        message = string.Format(
-                                CultureInfo.CurrentCulture,
-                                Strings.PackageVersionDiffersOnlyByMetadataAndCannotBeModified,
-                                existingPackage.PackageRegistration.Id,
-                                existingPackage.Version);
+                        // Packages that failed validation can be reuploaded.
+                        await _packageDeleteService.HardDeletePackagesAsync(new[] { existingPackage }, currentUser, "", "", deleteEmptyPackageRegistration: false, saveAuditRecord: false);
```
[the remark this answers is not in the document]
shishirx34: why not save audit record? I think we should still save the audit record.

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -314,7 +314,10 @@ private bool HasVersion(StatisticsFact fact, string version)
                         "DELETE pf FROM PackageFrameworks pf JOIN Packages p ON p.[Key] = pf.Package_Key WHERE p.[Key] = @key",
                         new SqlParameter("@key", package.Key));
 
-                    await _auditingService.SaveAuditRecordAsync(CreateAuditRecord(package, package.PackageRegistration, AuditedPackageAction.Delete, reason));
+                    if (saveAuditRecord)
+                    {
+                        await _auditingService.SaveAuditRecordAsync(CreateAuditRecord(package, package.PackageRegistration, AuditedPackageAction.Delete, reason));
```
shishirx34: we should always save the audit record, I don't understand why skip on auto delete.

src/NuGetGallery/Controllers/PackagesController.cs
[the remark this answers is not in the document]
scottbommarito: We weren't saving the audit record because it results in a package delete catalog entry for a package that was never in V3. After more discussion with @loic-sharma and @agr however, we decided that we should still have this audit record.

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: After offline discussion we are now auditing the hard delete again.

src/NuGetGallery/Strings.resx
```diff
@@ -908,4 +908,10 @@ If you would like to update the linked Microsoft account you can do so from the
   <data name="InvalidRecaptchaResponse" xml:space="preserve">
     <value>Recaptcha response was not valid. Try again or email support if you need assistance.</value>
   </data>
+  <data name="FailedValidationHardDeleteReason" xml:space="preserve">
+    <value>Automatic hard-delete for reupload of package that failed validation</value>
+  </data>
+  <data name="FailedValidationHardDeleteSignature" xml:space="preserve">
+    <value>NuGet Gallery</value>
```
loic-sharma: >Gallery [](start = 17, length = 7)

Could we mention reupload? Maybe `NuGet Gallery (Reupload)`

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -360,33 +360,48 @@ public virtual JsonResult UploadPackageProgress()
...
-                                existingPackage.Version);
+                        _telemetryService.TrackPackageReupload(existingPackage);
+
+                        // Packages that failed validation can be reuploaded.
+                        await _packageDeleteService.HardDeletePackagesAsync(
+                            new[] { existingPackage }, 
+                            currentUser,
+                            Strings.FailedValidationHardDeleteReason,
+                            Strings.FailedValidationHardDeleteSignature, 
+                            deleteEmptyPackageRegistration: false);
```
loic-sharma: >deleteEmptyPackageRegistration: false [](start = 28, length = 37)

Could we add a test that the registration won't be touched?

scottbommarito: Those tests are in `PackageDeleteServiceFacts`.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -440,22 +442,29 @@ public virtual Task<ActionResult> CreatePackagePost()
...
                             {
-                                return new HttpStatusCodeWithBodyResult(
-                                    HttpStatusCode.Conflict,
-                                    string.Format(CultureInfo.CurrentCulture, Strings.PackageExistsAndCannotBeModified,
-                                        id, nuspec.GetVersion().ToNormalizedStringSafe()));
+                                if (existingPackage.PackageStatusKey == PackageStatus.FailedValidation)
+                                {
+                                    TelemetryService.TrackPackageReupload(existingPackage);
+
+                                    await PackageDeleteService.HardDeletePackagesAsync(
```
agr: We don't send any automatic emails on deletion, do we?

scottbommarito: nope

dotMorten:
Thank you !!!

onovotny:
Is this live yet? This just happened to one of my packages.

PR #6073, opened 2018-06-15
Add Zip Entry restriction

ryuyu:
Adds an additional nupkg validation that will fail if a package contains any entries that would result in extraction outside of the target extraction directory.

See https://github.com/NuGet/NuGet.Client/pull/2296 and https://github.com/NuGet/NuGet.Client/pull/2302 for client change details

NuGet Client:
image

NuGet Website:
image

### DIFF
--- a/src/NuGetGallery/Services/PackageService.cs
+++ b/src/NuGetGallery/Services/PackageService.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Data.Entity;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using NuGet.Frameworks;
 using NuGet.Packaging;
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -65,6 +65,9 @@ public void EnsureValid(PackageArchiveReader packageArchiveReader)
                 {
                     ValidateSupportedFrameworks(supportedFrameworks);
                 }
+
+                // This will throw if the package contains an entry which will extract outside of the target extraction directory
+                packageArchiveReader.ValidatePackageEntriesAsync(new System.Threading.CancellationToken()).GetAwaiter().GetResult();
```
loic-sharma: Can we make the method async and await? This can dead lock.

ryuyu: Done!

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -65,6 +65,9 @@ public void EnsureValid(PackageArchiveReader packageArchiveReader)
                 {
                     ValidateSupportedFrameworks(supportedFrameworks);
                 }
+
+                // This will throw if the package contains an entry which will extract outside of the target extraction directory
+                await packageArchiveReader.ValidatePackageEntriesAsync(new System.Threading.CancellationToken());
```
loic-sharma: This can be [`CancellationToken.None`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken.none?view=netframework-4.7.2). Also, please add a `using` for `System.Threading`.

PR #6088, opened 2018-06-21
Failed Validation packages should not be shown in the account page

### LINKED ISSUE #6010: Failed Validation packages should not be shown in the account page
When submitting a package that fails validation (e.g. incorrect signature), the package is listed in the account page /profiles/<user>. 

The profile page should list publicly available packages only (note that unlisted packages are not shown)

xavierdecoster:
Fixes #6010

### DIFF
--- a/tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
+++ b/tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
@@ -1024,17 +1024,45 @@ namespace NuGetGallery
                 };
                 var invalidatedPackage = new Package
                 {
-                    Version = "1.1.1",
+                    Version = "1.0.0",
 
                     PackageRegistration = new PackageRegistration
                     {
-                        Id = "package",
+                        Id = "packageFailedValidation",
                         Owners = new[] { owner },
-                        DownloadCount = 150
+                        DownloadCount = 0
                     },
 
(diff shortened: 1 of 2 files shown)


tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -1019,7 +1019,22 @@ public void ReturnsSinglePackageAsExpected(User currentUser, User owner)
...
+                    DownloadCount = 100,
+                    PackageStatusKey = PackageStatus.Available
+                };
+                var invalidatedPackage = new Package
+                {
+                    Version = "1.1.1",
+
+                    PackageRegistration = new PackageRegistration
+                    {
+                        Id = "package",
```
loic-sharma: >package [](start = 30, length = 7)

Could we use a different package id or version so that we can distinguish different packages? Also, let's add "Deleted" and "Validating" packages so that we know all cases are covered.

scottbommarito: Yeah, I'd enumerate through all of the possible values of the enum and expect that it doesn't appear unless the status is available.

xavierdecoster: Added

PR #6089, opened 2018-06-21
ApiController actions should not throw exceptions when called with an unparseable version

scottbommarito:
If `VerifyPackageKeyAsync`, `DeletePackage`, or `PublishPackage` were called without a valid version string, they could throw because the `NuGetVersion.Parse` used in `GetHttpResultFromFailedApiScopeEvaluation` would throw. This fixes that issue.

### DIFF
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -763,9 +763,17 @@ namespace NuGetGallery
             return new HttpStatusCodeResult(HttpStatusCode.NotFound);
         }
 
-        private HttpStatusCodeWithBodyResult GetHttpResultFromFailedApiScopeEvaluation(ApiScopeEvaluationResult evaluationResult, string id, string version)
+        private HttpStatusCodeWithBodyResult GetHttpResultFromFailedApiScopeEvaluation(ApiScopeEvaluationResult evaluationResult, string id, string versionString)
         {
-            return GetHttpResultFromFailedApiScopeEvaluation(evaluationResult, id, NuGetVersion.Parse(version));
+            if (NuGetVersion.TryParse(versionString, out var version))
+            {
+
+                return GetHttpResultFromFailedApiScopeEvaluation(evaluationResult, id, version);
+            }
+            else
+            {
+                return GetHttpResultFromFailedApiScopeEvaluation(evaluationResult, id, versionString);
+            }
         }
 
         private HttpStatusCodeWithBodyResult GetHttpResultFromFailedApiScopeEvaluation(ApiScopeEvaluationResult result, string id, NuGetVersion version)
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -779,10 +779,15 @@ private HttpStatusCodeWithBodyResult GetHttpResultFromFailedApiScopeEvaluation(A
         /// </remarks>
         private HttpStatusCodeWithBodyResult GetHttpResultFromFailedApiScopeEvaluationForPush(ApiScopeEvaluationResult result, string id, NuGetVersion version)
         {
-            return GetHttpResultFromFailedApiScopeEvaluationHelper(result, id, version, HttpStatusCode.Unauthorized);
+            return GetHttpResultFromFailedApiScopeEvaluationHelper(result, id, ParseNuGetVersionForHttpResultForFailedApiScopeEvaluation(version), HttpStatusCode.Unauthorized);
         }
 
-        private HttpStatusCodeWithBodyResult GetHttpResultFromFailedApiScopeEvaluationHelper(ApiScopeEvaluationResult result, string id, NuGetVersion version, HttpStatusCode statusCodeOnFailure)
+        private string ParseNuGetVersionForHttpResultForFailedApiScopeEvaluation(NuGetVersion version)
```
shishirx34: Unnecessarily long name.. simply name it `GetNormalizedVersion`

skofman1:
please attach an issue and add UTs

scottbommarito:
@skofman1 - unit tests are added, please review

tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
```diff
@@ -917,9 +918,11 @@ public async Task WillNotUnlistThePackageIfScopesInvalid(ApiScopeEvaluationResul
                 var currentUser = fakes.User;
 
                 var id = "theId";
+                var version = "some version"; // We are using an invalid version string to guarantee it is not attempted to be parsed into a NuGetVersion.
```
skofman1: not clear why this test needs to be modified. You should test version parsing in a separate test

skofman1:
>             public async Task WillListThePackageIfScopesInvalid(ApiScopeEvaluationResult evaluationResult, HttpStatusCode expectedStatusCode, string description)

the name is wrong. Should be WillNotListThePackage..

---
Refers to: tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs:1248 in 637537c. [](commit_id = 637537c61bd41b501d9255bdfbc1d91b1c35bcd9, deletion_comment = False)

tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
```diff
@@ -1248,9 +1251,11 @@ public async Task WillListThePackageIfScopesInvalid(ApiScopeEvaluationResult eva
...
```
skofman1: again, no need to modify this test. This test expects method PublishPackage to return failure. If you provide an invalid version string we don't know if it's failed because of the version string or the scopes.

tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
joelverhagen: I agree, "invalid scope" does not necessarily mean invalid version string, right?

scottbommarito: To clarify--the code doesn't parse the version string at all. It behaves the same regardless of whether or not it's a valid version string or an invalid version string. The bug before was that it parsed the version string. I thought that testing that the version is not parsed by supplying an invalid version string was a sufficient change.

I can add the version string as a parameter to this test.

skofman1: If this test fails, it should fail because the scopes are wrong, not because the version string is invalid. If the code changes in the future to parse version always, that might happen and the test will not catch it.

Issue #2923, opened 2016-03-05
More than one migrations configuration type was found in the assembly 'NuGetGallery'


cmcknz77:
I'm setting up Nuget Gallery for first time and am following the steps in this document: https://github.com/NuGet/NuGetGallery/wiki/Hosting-the-NuGet-Gallery-Locally-in-IIS
I have successfully installed IIS and created the Website http://localhost/Gallery
I have also successfully installed SQLExpress2014 and am using the following as my connectionstring for the destination database:
"Data Source=.\SQLEXPRESS;Initial Catalog=master;Integrated Security=True"
 
But when I ran Update-Database command to create the DB I get below error:

PM>  Update-Database -StartUpProjectName NuGetGallery -debug -verbose
Using StartUp project 'NuGetGallery'.
Using NuGet project 'NuGetGallery'.
Specify the '-Verbose' flag to view the SQL statements being applied to the target database.
System.Data.Entity.Migrations.Infrastructure.MigrationsException: More than one migrations configuration type was found in the assembly 'NuGetGallery'. Specify the name of the one to use.
   at System.Data.Entity.Utilities.TypeFinder.FindType(Type baseType, String typeName, Func`2 filter, Func`2 noType, Func`3 multipleTypes, Func`3 noTypeWithName, Func`3 multipleTypesWithName)
   at System.Data.Entity.Migrations.Utilities.MigrationsConfigurationFinder.FindMigrationsConfiguration(Type contextType, String configurationTypeName, Func`2 noType, Func`3 multipleTypes, Func`3 noTypeWithName, Func`3 multipleTypesWithName)
   at System.Data.Entity.Migrations.Design.ToolingFacade.BaseRunner.FindConfiguration()
   at System.Data.Entity.Migrations.Design.ToolingFacade.BaseRunner.GetMigrator()
   at System.Data.Entity.Migrations.Design.ToolingFacade.UpdateRunner.Run()
   at System.AppDomain.DoCallBack(CrossAppDomainDelegate callBackDelegate)
   at System.AppDomain.DoCallBack(CrossAppDomainDelegate callBackDelegate)
   at System.Data.Entity.Migrations.Design.ToolingFacade.Run(BaseRunner runner)
   at System.Data.Entity.Migrations.Design.ToolingFacade.Update(String targetMigration, Boolean force)
   at System.Data.Entity.Migrations.UpdateDatabaseCommand.<>c__DisplayClass2.<.ctor>b__0()
[...]

PR #6092, opened 2018-06-22
Update README.md

matthewjones555:
Fixed the Update-Database command as per the following issue thread (https://github.com/NuGet/NuGetGallery/issues/2923#issuecomment-192675346), multiple configurations have been added since this readme was authored, so you must now explicitly state which configuration you want to use, otherwise you will get the following error:

System.Data.Entity.Migrations.Infrastructure.MigrationsException: More than one migrations configuration type was found in the assembly 'NuGetGallery'.

### DIFF
--- a/docs/Deploying/README.md
+++ b/docs/Deploying/README.md
@@ -66,7 +66,7 @@ Replacing '[servername]', 'sa' and '[sapassword]' with the name of your Azure SQ
 In the Package Manager console, type the following command
 
 ```PowerShell
-Update-Database -ConnectionString "[ConnectionString]" -ConnectionProviderName System.Data.SqlClient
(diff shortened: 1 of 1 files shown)


DoctaJonez:
@skofman1, that explains a few things!  Thanks for the link.  I've now managed to get the gallery up and running, and wanted to leave the docs in a state where others finding them in future will be able to follow them without any problems.

It's a good idea to have a tidy up of the documentation, especially if some of it is old and obsolete.

matthewjones555:
@skofman1, that explains a few things! Thanks for the link. I've now managed to get the gallery up and running, and wanted to leave the docs in a state where others finding them in future will be able to follow them without any problems.

It's a good idea to have a tidy up of the documentation, especially if some of it is old and obsolete.

skofman1:
Thanks! I created an issue to track cleanup: https://github.com/NuGet/NuGetGallery/issues/6142
Closing this PR.

Issue #6098, opened 2018-06-25
Improve user experience for package reupload


scottbommarito:
Both Gallery UI and validation failure email should contain:

1 - link to the NU error documentation
2 - "fix the issue and reupload your package" instructions

PR #6105, opened 2018-06-26
Include failure reasons in email sent when package validation fails

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6098

Example:

The package [MyPackage 1.0.0]({packageUrl}) failed validation because of the following reason(s):

- *NU3007:* Package signatures must have format version 1.
- This package must be signed with a registered certificate. [Read more...](https://aka.ms/nuget-signed-ref)

Your package was not published on NuGet Gallery and is not available for consumption.

Once you've fixed all the issues with your package, you can reupload it with the same ID and version.

### DIFF
--- a/src/NuGetGallery.Core/Extensions/PackageValidationSetExtensions.cs
+++ b/src/NuGetGallery.Core/Extensions/PackageValidationSetExtensions.cs
@@ -1,0 +1,29 @@
+﻿using NuGet.Services.Validation;
+using NuGet.Services.Validation.Issues;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery
+{
+    public static class PackageValidationSetExtensions
+    {
+        public static IReadOnlyList<ValidationIssue> GetValidationIssues(this PackageValidationSet validationSet)
+        {
+            // Get the failed validation set's validation issues. The issues are ordered by their
+            // key so that it appears that issues are appended as more validations fail.
+            var issues = validationSet
+                .PackageValidations
+                .SelectMany(v => v.PackageValidationIssues)
+                .OrderBy(i => i.Key)
+                .Select(i => ValidationIssue.Deserialize(i.IssueCode, i.Data))
+                .ToList();
+
+            // Filter out unknown issues and deduplicate the issues by code and data. This also deduplicates cases
+            // where there is extraneous data in the serialized data field or if the issue code is unknown.
+            return issues
+                .GroupBy(x => new { x.IssueCode, Data = x.Serialize() })
+                .Select(x => x.First())
+                .ToList();
+        }
+    }
+}
--- a/src/NuGetGallery.Core/Extensions/PackageValidationSetExtensions.cs
+++ b/src/NuGetGallery.Core/Extensions/PackageValidationSetExtensions.cs
@@ -9,29 +9,34 @@ namespace NuGetGallery
     {
         public static IReadOnlyList<ValidationIssue> GetValidationIssues(this PackageValidationSet validationSet)
         {
-            // Get the failed validation set's validation issues. The issues are ordered by their
-            // key so that it appears that issues are appended as more validations fail.
-            var issues = validationSet
-                .PackageValidations
-                .SelectMany(v => v.PackageValidationIssues)
-                .OrderBy(i => i.Key)
-                .Select(i => ValidationIssue.Deserialize(i.IssueCode, i.Data))
-                .ToList();
+            IReadOnlyList<ValidationIssue> issues = null;
 
-            // Filter out unknown issues and deduplicate the issues by code and data. This also deduplicates cases
-            // where there is extraneous data in the serialized data field or if the issue code is unknown.
-            IReadOnlyList<ValidationIssue> groupedIssues = issues
-                .GroupBy(x => new { x.IssueCode, Data = x.Serialize() })
-                .Select(x => x.First())
-                .ToList();
-
-            // If the package failed validation but we could not find an issue that explains why, use a generic error message.
-            if (groupedIssues == null || !groupedIssues.Any())
+            if (validationSet != null)
             {
-                groupedIssues = new[] { ValidationIssue.Unknown };
+                // Get the failed validation set's validation issues. The issues are ordered by their
+                // key so that it appears that issues are appended as more validations fail.
+                issues = validationSet
+                    .PackageValidations
+                    .SelectMany(v => v.PackageValidationIssues)
+                    .OrderBy(i => i.Key)
+                    .Select(i => ValidationIssue.Deserialize(i.IssueCode, i.Data))
+                    .ToList();
+
+                // Filter out unknown issues and deduplicate the issues by code and data. This also deduplicates cases
+                // where there is extraneous data in the serialized data field or if the issue code is unknown.
+                issues = issues
+                    .GroupBy(x => new { x.IssueCode, Data = x.Serialize() })
+                    .Select(x => x.First())
+                    .ToList();
             }
 
-            return groupedIssues;
+            // If the package failed validation but we could not find an issue that explains why, use a generic error message.
+            if (issues == null || !issues.Any())
+            {
+                issues = new[] { ValidationIssue.Unknown };
+            }
+
+            return issues;
         }
     }
 }
--- a/src/NuGetGallery.Core/Extensions/PackageValidationSetExtensions.cs
+++ b/src/NuGetGallery.Core/Extensions/PackageValidationSetExtensions.cs
@@ -20,10 +20,18 @@ namespace NuGetGallery
 
             // Filter out unknown issues and deduplicate the issues by code and data. This also deduplicates cases
             // where there is extraneous data in the serialized data field or if the issue code is unknown.
-            return issues
+            IReadOnlyList<ValidationIssue> groupedIssues = issues
                 .GroupBy(x => new { x.IssueCode, Data = x.Serialize() })
                 .Select(x => x.First())
                 .ToList();
+
+            // If the package failed validation but we could not find an issue that explains why, use a generic error message.
+            if (groupedIssues == null || !groupedIssues.Any())
+            {
+                groupedIssues = new[] { ValidationIssue.Unknown };
+            }
+
+            return groupedIssues;
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -50,53 +52,35 @@ public void SendPackageAddedNotice(Package package, string packageUrl, string pa
...
-                body,
-                CoreConfiguration.GalleryOwner.DisplayName,
-                package.PackageRegistration.Id,
-                package.Version,
-                packageUrl,
-                packageSupportUrl);
-
-            subject = string.Format(CultureInfo.CurrentCulture, subject, CoreConfiguration.GalleryOwner.DisplayName, package.PackageRegistration.Id, package.Version);
+            var subject = $"[{CoreConfiguration.GalleryOwner.DisplayName}] Package validation failed - {package.PackageRegistration.Id} {package.Version}";
+            var body = $@"The package [{package.PackageRegistration.Id} {package.Version}]({packageUrl}) failed validation because of the following reason(s):
```
agr: `body` should be made a `StringBuilder`. Repeated string concatenation is slow.

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -113,6 +97,35 @@ public void SendSignedPackageNotAllowedNotice(Package package, string packageUrl
...
+                    return "**NU3007:** Package signatures must have format version 1.";
+                case ValidationIssueCode.AuthorCounterSignaturesNotSupported:
+                    return "Author countersignatures are not supported.";
+                case ValidationIssueCode.PackageIsNotSigned:
+                    return "This package must be signed with a registered certificate. [Read more...](https://aka.ms/nuget-signed-ref)";
+                case ValidationIssueCode.PackageIsSignedWithUnauthorizedCertificate:
+                    var certIssue = (UnauthorizedCertificateFailure)validationIssue;
+                    return $"The package was signed, but the signing certificate (SHA-1 thumbprint {certIssue.Sha1Thumbprint}) is not associated with your account. You must register this certificate to publish signed packages. [Read more...](https://aka.ms/nuget-signed-ref)";
+                default:
+                    return "There was an unknown failure when validating your package.";
```
agr: Do we need better default message? This sounds like "something went wrong, but we don't know what"

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -50,53 +52,35 @@ public void SendPackageAddedNotice(Package package, string packageUrl, string pa
...
-                CoreConfiguration.GalleryOwner.DisplayName,
-                package.PackageRegistration.Id,
-                package.Version,
-                packageUrl,
-                packageSupportUrl);
-
-            subject = string.Format(CultureInfo.CurrentCulture, subject, CoreConfiguration.GalleryOwner.DisplayName, package.PackageRegistration.Id, package.Version);
+            var subject = $"[{CoreConfiguration.GalleryOwner.DisplayName}] Package validation failed - {package.PackageRegistration.Id} {package.Version}";
+            var body = $@"The package [{package.PackageRegistration.Id} {package.Version}]({packageUrl}) failed validation because of the following reason(s):
+";
```
loic-sharma: This should also handle the case that the email has no validation issues.

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -50,53 +52,35 @@ public void SendPackageAddedNotice(Package package, string packageUrl, string pa
...
 
-            subject = string.Format(CultureInfo.CurrentCulture, subject, CoreConfiguration.GalleryOwner.DisplayName, package.PackageRegistration.Id, package.Version);
+            if (validationIssues.Any(i => i.IssueCode == ValidationIssueCode.Unknown))
+            {
+                body += $"Please [contact support]({packageSupportUrl}) to help fix your package.";
+            }
+            else
+            {
+                var issuePluralString = validationIssues.Count() > 1 ? "all the issues" : "the issue";
+                body += $"Once you've fixed {issuePluralString} with your package, you can reupload it with the same ID and version.";
```
loic-sharma: >it with the same ID and version."; [](start = 100, length = 34)

I don't think we need to state you are allowed to reuse the same ID and version, as that's the expected behavior. How about:

> You can reupload your package once you've fixed {issuePluralString}.

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -113,6 +97,35 @@ public void SendSignedPackageNotAllowedNotice(Package package, string packageUrl
...
+                case ValidationIssueCode.AuthorCounterSignaturesNotSupported:
+                    return "Author countersignatures are not supported.";
+                case ValidationIssueCode.PackageIsNotSigned:
+                    return "This package must be signed with a registered certificate. [Read more...](https://aka.ms/nuget-signed-ref)";
+                case ValidationIssueCode.PackageIsSignedWithUnauthorizedCertificate:
+                    var certIssue = (UnauthorizedCertificateFailure)validationIssue;
+                    return $"The package was signed, but the signing certificate (SHA-1 thumbprint {certIssue.Sha1Thumbprint}) is not associated with your account. You must register this certificate to publish signed packages. [Read more...](https://aka.ms/nuget-signed-ref)";
+                default:
+                    return "There was an unknown failure when validating your package.";
+            }
```
loic-sharma: We should know when this happens. I would log a warning that we have an unknown validation issue

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -113,6 +97,35 @@ public void SendSignedPackageNotAllowedNotice(Package package, string packageUrl
...
         }
 
+        private static string ParseValidationIssue(ValidationIssue validationIssue, string announcementsUrl, string twitterUrl)
+        {
+            switch (validationIssue.IssueCode)
+            {
+                case ValidationIssueCode.PackageIsSigned:
+                    return $"This package could not be published since it is signed. We do not accept signed packages at this moment. To be notified about package signing and more, watch our [Announcements]({announcementsUrl}) page or follow us on [Twitter]({twitterUrl}).";
+                case ValidationIssueCode.ClientSigningVerificationFailure:
+                    var clientIssue = (ClientSigningVerificationFailure)validationIssue;
```
loic-sharma: >ClientSigningVerificationFailure [](start = 39, length = 32)

This can throw if the code and the type don't match. I would advise testing the type before casting.

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -113,6 +97,35 @@ public void SendSignedPackageNotAllowedNotice(Package package, string packageUrl
             }
         }
 
+        private static string ParseValidationIssue(ValidationIssue validationIssue, string announcementsUrl, string twitterUrl)
```
loic-sharma: >ParseValidationIssue [](start = 30, length = 20)

With this change,  you will need to update two different places to update a validation issue's message (here and in the `ValidationIssue` view). We either need a way to ensure that the messages don't diverge, or, we should merge the messages together somehow. I would prefer the latter.

src/NuGetGallery.Core/Services/CoreMessageService.cs
scottbommarito: I couldn't think of a better message.

This is for both:
- something went wrong but we don't know what
- something went wrong but we can't tell you

src/NuGetGallery.Core/Services/CoreMessageService.cs
scottbommarito: `validationSet.GetValidationIssues()` returns a `ValidationIssue.Unknown` when no validation issues are present

src/NuGetGallery.Core/Services/CoreMessageService.cs
scottbommarito: A couple thoughts:
- ValidationIssue alerting should be handled by the validation pipeline
- `ValidationIssue.Unknown` is returned for failures that we don't expose to users, so this is not always unexpected

src/NuGetGallery.Core/Services/CoreMessageService.cs
scottbommarito: I would like to merge the messages together, but unfortunately `CoreMessageService.ParseValidationIssue` emits a markdown string and `_ValidationIssues.cshtml` emits HTML. I think the work to share the logic by converting one form to the other is too much to warrant the work, especially considering we may actually want to diverge the messaging between the package details page and the failed validation email in the future.

src/NuGetGallery.Core/Services/CoreMessageService.cs
scottbommarito: added a `clientIssue != null` check

src/NuGetGallery.Core/Services/CoreMessageService.cs
scottbommarito: done

src/NuGetGallery.Core/Services/CoreMessageService.cs
scottbommarito: done

src/NuGetGallery.Core/Extensions/PackageValidationSetExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+                issues = issues
+                    .GroupBy(x => new { x.IssueCode, Data = x.Serialize() })
+                    .Select(x => x.First())
+                    .ToList();
+            }
+
+            // If the package failed validation but we could not find an issue that explains why, use a generic error message.
+            if (issues == null || !issues.Any())
+            {
+                issues = new[] { ValidationIssue.Unknown };
```
loic-sharma: A `ValidationIssue.Unknown` means that deserialization of the issue failed. I think we should return an empty array instead and handle the empty list case in the views/emails.

loic-sharma: Also, we should log/track any `ValidationIssue.Unknown`s. This means we need to update our `NuGet.Services.Validation.Issues` dependency in the Gallery. I can easily see it getting out-of-date without us realizing this for a long time.

scottbommarito: For users I believe we should show the same message for a `ValidationIssue` that failed to serialize and for a missing `ValidationIssue`. That is why we expose a `ValidationIssue.Unknown` here.

I agree we should log/track any `ValidationIssue.Unknown`s, but it seems that this is potentially difficult to do given that
- gallery uses a legacy logging system
- there are few logs in the gallery
- this code is shared by both gallery and orchestrator, so the logging system may not work correctly

Given that this is an existing issue and has a nontrivial fix, I don't believe it should block this PR. However, I think it's definitely worthwhile to file an issue and fix this when we make a larger investment into logging in the gallery.

PR #6129, opened 2018-07-02
Account Delete - add telemetry and remove signature column

cristinamanum:
Account Delete - add telemetry and remove signature column


src/NuGetGallery/Services/ITelemetryService.cs
```diff
@@ -122,5 +122,19 @@ public interface ITelemetryService
...
+        /// <param name="deletedUser">The <see cref="User"/> that was deleted.</param>
+        /// <param name="deletedBy">The <see cref="User"/> that performed the delete.</param>
+        /// <param name="success">The success of the operation.</param>
+        void TrackAccountDeletedCompleted(User deletedUser, User deletedBy, bool success);
+
+        /// <summary>
+        /// A telemetry event emitted when a request to have an account deleted is submitted.
+        /// </summary>
+        /// <param name="user">The <see cref="User"/> requesting the delete.</param>
+        void TrackRequestForAccountDeleted(User user);
```
loic-sharma: >TrackRequestForAccountDeleted [](start = 13, length = 29)

What do you think of `TrackRequestForAccountDeletion`?

src/NuGetGallery/Services/ITelemetryService.cs
```diff
@@ -122,5 +122,19 @@ public interface ITelemetryService
...
         /// </summary>
         void TrackException(Exception exception, Action<Dictionary<string, string>> addProperties);
+
+        /// <summary>
+        /// A telemetry event emitted when an account is deleted.
+        /// </summary>
+        /// <param name="deletedUser">The <see cref="User"/> that was deleted.</param>
+        /// <param name="deletedBy">The <see cref="User"/> that performed the delete.</param>
+        /// <param name="success">The success of the operation.</param>
+        void TrackAccountDeletedCompleted(User deletedUser, User deletedBy, bool success);
```
loic-sharma: >TrackAccountDeletedCompleted [](start = 13, length = 28)

What do you think of `TrackAccountDeletionCompleted`?

src/NuGetGallery/Services/ITelemetryService.cs
```diff
@@ -122,5 +122,19 @@ public interface ITelemetryService
...
+
+        /// <summary>
+        /// A telemetry event emitted when a request to have an account deleted is submitted.
```
loic-sharma: >when a request to have an account deleted is submitted [](start = 38, length = 54)

How about "... when an account deletion is requested"?

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -550,6 +559,37 @@ public void TrackOrganizationAdded(Organization organization)
...
+            {
+                throw new ArgumentNullException(nameof(deletedUser));
+            }
+            if (deletedBy == null)
+            {
+                throw new ArgumentNullException(nameof(deletedBy));
+            }
+
+            TrackMetric(Events.AccountDeleteCompleted, 1, properties => {
+                properties.Add(AccountDeletedByRole, string.Join(",", deletedBy.Roles?.Select((role) => role.Name) ?? new List<string>()));
```
loic-sharma: >role) => role.Name [](start = 95, length = 18)

`(role) => ...` can be `role => ...`

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -121,19 +127,17 @@ IAuditingService auditingService
                 await RemoveUserDataInUserTable(userToBeDeleted);
                 await InsertDeleteAccount(
                     userToBeDeleted, 
-                    userToExecuteTheDelete, 
-                    signature);
+                    userToExecuteTheDelete);
```
loic-sharma: >userToExecuteTheDelete [](start = 20, length = 22)

Shouldn't there be a UI change to remove the Signature field?

loic-sharma:
>             string signature = null)

This is no longer necessary right?

---
Refers to: src/NuGetGallery/Services/DeleteAccountService.cs:64 in 62f3e96. [](commit_id = 62f3e9682fc0171ac363d49b37017792b752a697, deletion_comment = False)

loic-sharma:
>         private async Task RemoveMemberships(User user, User requestingUser, AccountDeletionOrphanPackagePolicy orphanPackagePolicy, string signature)

Is this needed now?

---
Refers to: src/NuGetGallery/Services/DeleteAccountService.cs:225 in 62f3e96. [](commit_id = 62f3e9682fc0171ac363d49b37017792b752a697, deletion_comment = False)

loic-sharma:
>                     await DeleteAccountImplAsync(organization, requestingUser, orphanPackagePolicy, signature);

Is this needed now?

---
Refers to: src/NuGetGallery/Services/DeleteAccountService.cs:239 in 62f3e96. [](commit_id = 62f3e9682fc0171ac363d49b37017792b752a697, deletion_comment = False)

src/NuGetGallery/Migrations/201807011920472_RemoveDeleteAccountSignature.cs
```diff
@@ -0,0 +1,18 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class RemoveDeleteAccountSignature : DbMigration
+    {
+        public override void Up()
+        {
```
skofman1: Our process is to deploy schema changes before code, and if we deploy this change, it will break the existing service. What we did in such cases is two deployments: 1. Remove all usage of the column and deploy. 2. Remove column and deploy. Also consider the rollback scenario: we should always be able to rollback to previous version without breaking scenarios.

src/NuGetGallery/Services/ITelemetryService.cs
cristinamanum: It sounds better - updated.

src/NuGetGallery/Services/ITelemetryService.cs
cristinamanum: Updated

src/NuGetGallery/Services/ITelemetryService.cs
cristinamanum: Updated

src/NuGetGallery/Services/TelemetryService.cs
cristinamanum: Updated

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: I like the signature on the UI, The user will think twice before deleting it.

src/NuGetGallery/Migrations/201807031816101_AccountDeleteSignatureNotRequired.cs
```diff
@@ -0,0 +1,18 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AccountDeleteSignatureNotRequired : DbMigration
+    {
+        public override void Up()
+        {
+            AlterColumn("dbo.AccountDeletes", "Signature", c => c.String());
```
skofman1: is this migration needed?

src/NuGetGallery.Core/Entities/AccountDelete.cs
```diff
@@ -37,10 +37,9 @@ public class AccountDelete
...
 
-        /// <summary>
-        /// The signature of the admin.
-        /// </summary>
-        [Required]
-        public string Signature { get; set; }
+         /// <summary>
+         /// The signature of the admin.
+         /// </summary>
+         public string Signature { get; set; }
```
skofman1: can we remove this property completely from the code?

src/NuGetGallery/Migrations/201807031816101_AccountDeleteSignatureNotRequired.cs
cristinamanum: This migration is needed because otherwise an insert with NULL data will fail.

src/NuGetGallery.Core/Entities/AccountDelete.cs
cristinamanum: It will be removed in a subsequent PR to not break the backwards compatibility.

src/NuGetGallery/Migrations/201807011920472_RemoveDeleteAccountSignature.cs
cristinamanum: The new change reverts the "column drop migration" but removing the usage without dropping the column needs to change the column constraints to accept NULL.

src/NuGetGallery/Services/DeleteAccountService.cs
cristinamanum: Updated the source code for the below comments.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -293,13 +293,13 @@ private ActionResult TransformToOrganizationFailed(string errorMessage)
             {
                 return HttpNotFound();
             }
+            TelemetryService.TrackRequestForAccountDeletion(user);
```
shishirx34: nit: on newline

src/NuGetGallery/Migrations/201807031816101_AccountDeleteSignatureNotRequired.cs
shishirx34: we can make nullable true I guess, but I agree better to remove if not using.

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -550,6 +559,37 @@ public void TrackOrganizationAdded(Organization organization)
...
+            {
+                throw new ArgumentNullException(nameof(deletedUser));
+            }
+            if (deletedBy == null)
+            {
+                throw new ArgumentNullException(nameof(deletedBy));
+            }
+
+            TrackMetric(Events.AccountDeleteCompleted, 1, properties => {
+                properties.Add(AccountDeletedByRole, string.Join(",", deletedBy.Roles?.Select( role => role.Name) ?? new List<string>()));
```
shishirx34: nit: extra space in `Select`

cristinamanum: Yes - the column will be removed in a subsequent PR  after this will be deployed.

Issue #5877, opened 2018-05-02
Account/Packages page is slow for users with lots of packages


skofman1:
For Microsoft account takes ~25 seconds to load.

skofman1:
@scottbommarito improved perf significantly by server side perf improvements. However, client side is still slow.

scottbommarito:
Looked at client side. It seems like the majority of the client-side performance issues are related to the fact that the page renders EVERY row of data instead of just what is visible, rather than the Javascript itself being slow.

(I may be wrong here, though, because I am not an expert with performance tooling, and the fact that Knockout was obfuscated/minified made the call stack much more difficult to parse.)

[To mitigate this, through Javascript, I added rendering for ONLY rows that are in view](https://github.com/NuGet/NuGetGallery/compare/dev...sb-manpackjsperf), but due to the poor performance of checking the relative locations of thousands of rectangles to the viewport, this performed worse than without the rendering. I think, however, if the code for staggering the rendering of rows was less complex, e.g. by just rendering row by row with a delay between them, we may see an improvement.

Unfortunately, however, I don't have more time to work on this.

joelverhagen:
My performance fix is verified. The page is still slow (expected) just not as bad as before. I'll keep this issue open since it's still slow.

ErikEJ:
@joelverhagen 13 seconds after my fix - is that still slow?

joelverhagen:
I think 13 seconds is slow, yes. This could be mitigated with caching, perhaps.

ErikEJ:
@joelverhagen Let's see after my last PR, informal testing indicates around 50 % improvement

ErikEJ:
@joelverhagen Could this be closed now?

joelverhagen:
Yes!

Issue #5659, opened 2018-03-22
Performance: Profiles page should page data in the DB


chenriksson:
PR #5658 improved query performance by only selecting latest versions from the DB.

We can further optimize the query by selecting only the latest page from the DB. This requires 2 queries, but should be more performant:
- Count query, for determining # of packages (to calculate # of pages)
- Skip/Take applied to IQueryable, instead of in-memory after ToList is called

chenriksson:
Still relevant, but can boost priority if profiles page is bothersome

skofman1:
We can't easily get a consistent experience (details in the PR). Will return this issue to the backlog since the perf gain is not sufficient to justify the experience impact.

ErikEJ:
My recent fix improved the profile page performance!

ErikEJ:
@joelverhagen Close now?

joelverhagen:
Yes!

Issue #5909, opened 2018-05-10
Signing owner selection box on Manage Packages page can cause table to overflow


scottbommarito:
![image.png](https://images.zenhubusercontent.com/58054e1495e42b6c5baa9ebb/10f07f47-fbbe-4097-ab30-4bdf22d712dc)

xavierdecoster:
@scottbommarito haven't been able to repro this one yet. Got more info on how to reproduce this one? Also, in what browser is this?

scottbommarito:
@xavierdecoster - can you show me what it looks like for you? should show very easily if you have enough owners for the "signing owner" column to appear. You may also need to upload some packages with long names, long version strings, many owners, etc. This was done on Chrome.

joelverhagen:
I verified this on DEV environment.

PR #6139, opened 2018-07-04
Paginate packages for profile page in database

zivkan:
For #5877 and #5659 

Deploying the current dev code into an Azure App Service in the production slot, and putting my changes in a staging slot, then hitting page 2 on both sites 120 times each, App Insights tells me these results:

Version | 50th percentile | 95th percentile
---- | ----- | -----
dev code | 5,027ms | 6,150ms
these changes | 3,674ms | 4,570ms

The contents appear to be the same, except that the total package download count is different. It might be due to bad data in my database (I manually set the TotalDownloads column in the PackageRegistrations table, but didn't change the Packages table). However, I think that the value after my changes appear to be more correct than before my changes, so I'm not sure if this fixes a total downloads count bug.

The SQL that the various queries execute are horrific, but I guess that's part of using an ORM, and the performance is still better than paginating in C#. I have not tried any optimisation, I just re-used the existing query.


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -614,15 +614,15 @@ public virtual ActionResult Profiles(string username, int page = 1)
...
-                .Where(p => p.PackageStatusKey == PackageStatus.Available)
-                .OrderByDescending(p => p.PackageRegistration.DownloadCount)
+            var packages = PackageService.GetPagedPackagesByOwner(user, page, Constants.DefaultPackageListPageSize, includeUnlisted: false)
                 .Select(p => new ListPackageItemViewModel(p, currentUser)
                 {
                     DownloadCount = p.PackageRegistration.DownloadCount
                 }).ToList();
 
-            var model = new UserProfileModel(user, currentUser, packages, page - 1, Constants.DefaultPackageListPageSize, Url);
+            var (totalPackages, totalPackageDownloadCount) = PackageService.GetTotalPackagesStatisticsForOwner(user, includeUnlisted: false);
```
loic-sharma: The tuple could become annoying if we decide to add more statistics. What do you think of adding creating a new class to hold the statistics, and then passing that class to the `UserProfileModel`?

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -226,9 +226,27 @@ public virtual Package FindAbsoluteLatestPackageById(string id, int? semVerLevel
...
         }
 
-        public IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted, bool includeVersions = false)
+        public IEnumerable<Package> GetPagedPackagesByOwner(User user, int page, int pageSize, bool includeUnlisted, bool includeVersions = false)
         {
-            return GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions);
+            return GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions)
+                .Where(p=>p.PackageStatusKey == PackageStatus.Available)
+                .OrderByDescending(p=>p.PackageRegistration.DownloadCount)
+                .Skip((page-1)*pageSize)
```
loic-sharma: >page-1 [](start = 23, length = 6)

I know that the profile page is 1-based indexed, but I think that people would expect this to be 0-based indexed. I would either:

1. Make this method be 0-based indexed, and update the UsersController
2. Document that this method is 1-based indexed in the `IPackageService`

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -226,9 +226,27 @@ public virtual Package FindAbsoluteLatestPackageById(string id, int? semVerLevel
             return package;
         }
 
-        public IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted, bool includeVersions = false)
+        public IEnumerable<Package> GetPagedPackagesByOwner(User user, int page, int pageSize, bool includeUnlisted, bool includeVersions = false)
         {
-            return GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions);
+            return GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions)
+                .Where(p=>p.PackageStatusKey == PackageStatus.Available)
```
loic-sharma: >p=>p [](start = 23, length = 4)

Missing spaces on this line and the next: `p=>p` should be `p => p`

src/NuGetGallery/Services/IPackageService.cs
```diff
@@ -29,7 +29,9 @@ public interface IPackageService : ICorePackageService
 
         Package FindAbsoluteLatestPackageById(string id, int? semVerLevelKey);
 
-        IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted, bool includeVersions = false);
+        IEnumerable<Package> GetPagedPackagesByOwner(User user, int page, int pageSize, bool includeUnlisted, bool includeVersions = false);
```
loic-sharma: >bool includeVersions = false [](start = 110, length = 28)

What uses `includeVersions`? It seems that this parameter can be removed from both `GetPagedPackagesByOwner` and `GetTotalPackagesStatisticsForOwner`

src/NuGetGallery/Services/PackageService.cs
agr: I would suggest passing the skip value into this method instead. I don't think calculating the skip belongs here.

src/NuGetGallery/ViewModels/UserProfileModel.cs
```diff
@@ -8,43 +8,33 @@ namespace NuGetGallery
 {
     public class UserProfileModel
     {
-        public UserProfileModel(User user, User currentUser, List<ListPackageItemViewModel> allPackages, int pageIndex, int pageSize, UrlHelper url)
+        public UserProfileModel(User user, User currentUser, List<ListPackageItemViewModel> pagedPackages, int totalPackages, long totalPackageDownloadCount, int pageIndex, int pageSize, UrlHelper url)
```
agr: Time to split argument list into individual lines (one line per argument), I think.

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -987,8 +987,11 @@ public void ReturnsTheMostRecentVersionWhenSemVerLevel2()
             }
         }
 
-        public class TheFindPackagesByOwnerMethod : TheFindPackagesByOwnersMethodsBase
+        public class TheGetPagedPackagesByOwnerMethod : TheFindPackagesByOwnersMethodsBase
         {
+            private int page = 1;
```
agr: `const`?

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -226,9 +226,27 @@ public virtual Package FindAbsoluteLatestPackageById(string id, int? semVerLevel
...
+
+        public (int totalPackages, long totalPackageDownloadCount) GetTotalPackagesStatisticsForOwner(User user, bool includeUnlisted, bool includeVersions = false)
+        {
+            var packages = GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions);
+            var result = packages.Select(p => new
+            {
+                TotalPackages = packages.Count(),
+                TotalPackageDownloadCount = packages.Sum(p2 => (long)p2.PackageRegistration.DownloadCount)
+            })
+                .First();
```
agr: Inconsistent indentation.

agr: I'd prefer something like

```
            var result = packages
                .Select(p => new
                    {
                        TotalPackages = packages.Count(),
                        TotalPackageDownloadCount = packages.Sum(p2 => (long)p2.PackageRegistration.DownloadCount)
                    })
                .First();
```

(not sure about indenting the curly block, but this looks right to me: it's inner block for `Select`, so should be indented further)

skofman1:
@zivkan , can you provide a link to your webapp so we can try things out?

src/NuGetGallery/Services/PackageService.cs
zivkan: I also prefer your suggested style, but I don't have resharper installed, and I couldn't find a .editorconfig/VS formatting option to make the auto-formatting do that. I prefer using auto-formatting styles, so if someone reformats the entire document, it doesn't cause changes to code they didn't intend to change.

zivkan:
My web app URLS are:
https://anzivk-nuget.azurewebsites.net/profiles/microsoft
https://anzivk-nuget-staging.azurewebsites.net/profiles/microsoft

the times I listed in the PR message are for the microsoft account with 3,760 packages, just like in prod (although my database doesn't have other package registrations, and only has 1 package version per registration).

Sorting is currently broken too, so I have some more work to do.

loic-sharma:
Adding @xavierdecoster since he will know the most about this.

That interceptor gets the latest downloads within 3 hours (it uses the `downloads.v1.json` auxiliary file, which is refreshed by the [CreateAzureCdnWarehouseReports job](https://github.com/NuGet/NuGet.Jobs/blob/master/src/Stats.CreateAzureCdnWarehouseReports/Job.cs) every 3 hours. So, it seems that the interceptor's purpose is to get fresher download counts than what's in the Gallery DB. If that's true, I think we can skip the interceptor entirely. What do you think @xavierdecoster

xavierdecoster:
The main reason for this interception is to be consistent across the gallery in the numbers we show (on the web site, in v3 and v2 protocols). Due to the time it takes to generate the blobs and aggregate the download counts into the galllery database, there may be a time window during which results would be inconsistent.

If newer data is available in the blobs, v2 should also return these and discard the not-yet-updated counts in the db. In other words: download counts from blob storage overrule those found in the gallery database.

xavierdecoster:
I remember some people had strong opinions on data (in)consistency, which is why this interceptor was built in the first place.

What if we decided to no longer show the download counts on this particular view? 
This way, stats have nothing to do with rendering this view, and it could also help solve another issue: https://github.com/NuGet/NuGetGallery/issues/5909

skofman1:
Sounds like we can't easily get a consistent experience. Will return this issue to the backlog since the perf gain is not sufficient to justify the experience impact.

Issue #5923, opened 2018-05-14
Deletion of organizations that require an AAD tenant fails


scottbommarito:
Exception thrown:

```
System.NotSupportedException: Subscription 'RequireOrganizationTenantPolicy' not found.
   at NuGetGallery.Security.SecurityPolicyService.UnsubscribeAsync(User user, String subscriptionName) in C:\vsts-agent\_work\30\s\src\Gallery\submodules\Gallery\src\NuGetGallery\Security\SecurityPolicyService.cs:line 306
   at NuGetGallery.DeleteAccountService.<RemoveSecurityPolicies>d__15.MoveNext() in C:\vsts-agent\_work\30\s\src\Gallery\submodules\Gallery\src\NuGetGallery\Services\DeleteAccountService.cs:line 153
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGetGallery.DeleteAccountService.<DeleteAccountImplAsync>d__12.MoveNext() in C:\vsts-agent\_work\30\s\src\Gallery\submodules\Gallery\src\NuGetGallery\Services\DeleteAccountService.cs:line 98
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGetGallery.DeleteAccountService.<RunAccountDeletionTask>d__26.MoveNext() in C:\vsts-agent\_work\30\s\src\Gallery\submodules\Gallery\src\NuGetGallery\Services\DeleteAccountService.cs:line 303
```

PR #6140, opened 2018-07-05
Delete org with policies fix

chenriksson:
Fixes https://github.com/NuGet/NuGetGallery/issues/5923
- Allow search of organization accounts from Delete Account admin view
- Allow unsubscribe of organization tenant policy (used by delete account)

### DIFF
--- a/src/NuGetGallery/Security/RequireOrganizationTenantPolicy.cs
+++ b/src/NuGetGallery/Security/RequireOrganizationTenantPolicy.cs
@@ -12,7 +12,7 @@ namespace NuGetGallery.Security
 {
     public class RequireOrganizationTenantPolicy : UserSecurityPolicyHandler, IUserSecurityPolicySubscription
     {
-        public const string _SubscriptionName = "AzureActiveDirectoryTenant";
+        public const string _SubscriptionName = nameof(RequireOrganizationTenantPolicy);
         public const string PolicyName = nameof(RequireOrganizationTenantPolicy);
 
         public string SubscriptionName => _SubscriptionName;
@@ -25,18 +25,13 @@ namespace NuGetGallery.Security
             public string Tenant { get; set; }
         }
 
-        public RequireOrganizationTenantPolicy()
-            : base(PolicyName, SecurityPolicyAction.JoinOrganization)
-        {
-        }
-
         private RequireOrganizationTenantPolicy(IEnumerable<UserSecurityPolicy> policies)
             : base(PolicyName, SecurityPolicyAction.JoinOrganization)
         {
             Policies = policies;
         }
 
-        public static IUserSecurityPolicySubscription Create(string tenantId)
+        public static RequireOrganizationTenantPolicy Create(string tenantId = "")
         {
             var value = JsonConvert.SerializeObject(new State()
             {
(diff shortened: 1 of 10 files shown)


src/NuGetGallery/Security/RequireOrganizationTenantPolicy.cs
```diff
@@ -12,7 +12,7 @@ namespace NuGetGallery.Security
 {
     public class RequireOrganizationTenantPolicy : UserSecurityPolicyHandler, IUserSecurityPolicySubscription
     {
-        public const string _SubscriptionName = "AzureActiveDirectoryTenant";
+        public const string _SubscriptionName = nameof(RequireOrganizationTenantPolicy);
```
cristinamanum: Does it matter that this value is the same as for the Policy Name below?

src/NuGetGallery/Security/RequireOrganizationTenantPolicy.cs
```diff
@@ -25,18 +25,13 @@ public class State
...
-        }
-
         private RequireOrganizationTenantPolicy(IEnumerable<UserSecurityPolicy> policies)
             : base(PolicyName, SecurityPolicyAction.JoinOrganization)
         {
             Policies = policies;
         }
 
-        public static IUserSecurityPolicySubscription Create(string tenantId)
+        public static RequireOrganizationTenantPolicy Create(string tenantId = "")
```
cristinamanum: Why not returning the interface as previously? Empty tenantId would not be an invalid case?

cristinamanum: null check?

src/NuGetGallery/Security/RequireOrganizationTenantPolicy.cs
cristinamanum: Why are both needed?

chenriksson: The original code persisted PolicyName in the Subscription column, so I need to keep this consistent with what's currently in the DB.

see: https://github.com/NuGet/NuGetGallery/pull/6140/files#diff-5a32f572c33602bd9d4d9c5312c9396aL48

Note that a subscription is a set of policies, which is why there's a different column for it. There's only a single policy in this case... but the original "Secure Push" policy had 2 policies associated with the Subscription, which is why it may differ.

src/NuGetGallery/Security/RequireOrganizationTenantPolicy.cs
chenriksson: I replaced the default ctor above with this factory method, to ensure that the policies were always getting set on the subscription object. I think that was mostly used in tests, but I needed the strong type for validations.

scottbommarito: To be honest, this seems a little hacky to me. We shouldn't be creating empty policies. Having to do this indicates to me a larger problem in the structure of this code.

As far as I can see, the real problem here is that you have a single class for both
- storing information about a user's subscription to a policy (e.g. the tenant ID the user must be)
- handling a user's subscription/unsubscription for a policy (e.g. `OnSubscribeAsync` and `OnUnsubscribeAsync`)

This hasn't been as much of a problem in the past because the cardinality of our policies has been extremely limited, but if we continue to use security policies for more and more things, this will likely be a problem again for future policies (if it hasn't already).

If you can do this refactoring quickly I'd highly recommend it. Otherwise, I think this fix is sufficient for now.

Issue #6143, opened 2018-07-05
Large package size should be rejected with 413 - Payload too large


shishirx34:
I believe IIS rejects with 500 right now.

singku:
It's happens when using nuget.exe, so here is the original issue. https://github.com/NuGet/Home/issues/7091.

The questions is, will this size limit be updated to higher value?

skofman1:
@singku , 250MB is currently a hard limit on package size on nuget.org and is necessary to deliever a good e2e experience. We will consider raising it in the future as the requirements of customers change.
For now, we recommend spliting your package into two packages. We will respond with detailed instructions on NuGet/Home#7091.

singku:
@skofman1 Thank you. Will follow the instructions to split if it's applicable in our case.

PR #6147, opened 2018-07-06
Improve experience when users hit package size limit

chenriksson:
Fixes https://github.com/NuGet/NuGetGallery/issues/6143

**API push**
Hits `maxRequestLength` (ASP.NET) limit which throws HttpException.
We now handle this to return status 413 with better error message, and client no longer retries.

```
NuGet Version: 4.6.2.5055
Pushing chenriks252.nupkg to 'https://localhost/api/v2/package/'...
  PUT https://localhost/api/v2/package/
  RequestEntityTooLarge https://localhost/api/v2/package/ 8448ms
Response status code does not indicate success: 413 (The package file exceeds the size limit of 250MB.).
System.Net.Http.HttpRequestException: Response status code does not indicate success: 413 (The package file exceeds the
size limit of 250MB.).
   at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
   at NuGet.Protocol.Core.Types.PackageUpdateResource.<>c.<PushPackageToServer>b__18_0(HttpResponseMessage response)
   at NuGet.Protocol.HttpSource.<ProcessResponseAsync>d__14`1.MoveNext()
```

Currently, we get 500/InternalServerError with retry:
```
NuGet Version: 4.6.2.5055
Pushing chenriks252.nupkg to 'https://dev.nugettest.org/api/v2/package/'...
  PUT https://dev.nugettest.org/api/v2/package/
  InternalServerError https://dev.nugettest.org/api/v2/package/ 71503ms
  PUT https://dev.nugettest.org/api/v2/package/
```

**UI upload**
Hits `maxAllowedContentLength` (IIS) limit which returns 404.13 before the request reaches ASP.NET.
We now handle this at the client to display an immediate error message, instead of silently failing.

image

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/async-file-upload.js
+++ b/src/NuGetGallery/Scripts/gallery/async-file-upload.js
@@ -177,7 +177,7 @@
                 case "error":
                     // IIS returns 404.13 (NotFound) when maxAllowedContentLength limit is exceeded.
                     if (fullResponse === "Not Found") {
-                        displayErrors(["The package file exceeds the size limit of 250MB. Please try again."]);
+                        displayErrors(["The package file exceeds the size limit. Please try again."]);
                     }
                     break;
                 default:
--- a/src/NuGetGallery/Scripts/gallery/async-file-upload.js
+++ b/src/NuGetGallery/Scripts/gallery/async-file-upload.js
@@ -174,6 +174,12 @@
                 case "abort":
                     displayErrors(["The operation was aborted. Please try again."]);
                     break;
+                case "error":
+                    // IIS returns 404.13 (NotFound) when maxAllowedContentLength limit is exceeded.
+                    if (fullResponse === "Not Found") {
+                        displayErrors(["The package file exceeds the size limit of 250MB. Please try again."]);
+                    }
+                    break;
                 default:
                     displayErrors(model.responseJSON);
                     break;
--- a/src/NuGetGallery/Services/TelemetryService.cs
+++ b/src/NuGetGallery/Services/TelemetryService.cs
@@ -113,6 +113,8 @@ namespace NuGetGallery
         public const string CreatedDateForAccountToBeDeleted = "CreatedDateForAccountToBeDeleted";
         public const string AccountDeleteSucceeded = "AccountDeleteSucceeded";
 
+        public const string ValueUnknown = "Unknown";
+
         public TelemetryService(IDiagnosticsService diagnosticsService, ITelemetryClient telemetryClient = null)
         {
             if (diagnosticsService == null)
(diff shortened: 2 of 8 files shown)


src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -184,11 +186,13 @@ public void TrackPackagePushEvent(Package package, User user, IIdentity identity
 
         public void TrackPackagePushFailureEvent(string id, NuGetVersion version)
```
chenriksson: ApiController originally handled the HttpException in the default catch, which logged the push failure. This resulted in a NullReferenceException because CreatePackageInternal never reached the code that set the package version.

I don't think we need to continue logging push failures as a result of package size limit... but I think we should guard against NullReferenceException here, since version isn't populated until later in the CreatePackageInternal method.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -174,6 +174,12 @@
                 case "abort":
                     displayErrors(["The operation was aborted. Please try again."]);
                     break;
+                case "error":
+                    // IIS returns 404.13 (NotFound) when maxAllowedContentLength limit is exceeded.
```
cristinamanum: If the file is not found will be the same error message?

loic-sharma: >404.13 [](start = 35, length = 6)

`404.13` should be `413`, right?

src/NuGetGallery/Strings.resx
```diff
@@ -911,4 +911,7 @@ If you would like to update the linked Microsoft account you can do so from the
   <data name="FailedValidationHardDeleteReason" xml:space="preserve">
     <value>Automatic hard-delete for reupload of package that failed validation</value>
   </data>
+  <data name="PackageFileTooLarge" xml:space="preserve">
+    <value>The package file exceeds the size limit of 250MB.</value>
```
shishirx34: 250MB should be a constant configured value, so that we can change the limit without code change.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
agr: 404.13 is an IIS response which means "Content too large"

chenriksson: What @agr said :-)

chenriksson: No - it won't reach this point if the user provides a path to a non-existent file.

src/NuGetGallery/Strings.resx
chenriksson: I'll make the string less specific, so only the original limits (below) need updating.
- ASP.NET config: `httpRuntime.maxRequestLength="256000KB"`
- IIS config: `requestLimits.maxAllowedContentLength="262144000"`

PR #6148, opened 2018-07-06
Refactor CorePackageFileService 

cristinamanum:
Refactor CorePackageFileService to allow the file metadata to be passed at construction time instead to use directly the CoreConstants values. This will allow the same logic for file manipulation to be used for packages and symbols.

### DIFF
--- a/src/NuGetGallery.Core/Services/CorePackageFileService.cs
+++ b/src/NuGetGallery.Core/Services/CorePackageFileService.cs
@@ -13,10 +13,12 @@ namespace NuGetGallery
     public class CorePackageFileService : ICorePackageFileService
     {
         private readonly ICoreFileStorageService _fileStorageService;
+        private readonly IFileServiceMetadata _fileServiceMetadata;
 
-        public CorePackageFileService(ICoreFileStorageService fileStorageService)
+        public CorePackageFileService(ICoreFileStorageService fileStorageService, IFileServiceMetadata fileServiceMetadata)
         {
-            _fileStorageService = fileStorageService;
+            _fileStorageService = fileStorageService ?? throw new ArgumentNullException(nameof(fileStorageService));
+            _fileServiceMetadata = fileServiceMetadata ?? throw new ArgumentNullException(nameof(fileServiceMetadata));
         }
 
         public Task SavePackageFileAsync(Package package, Stream packageFile)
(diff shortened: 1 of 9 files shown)


src/NuGetGallery.Core/Services/IMetadataFileService.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    public interface IFileServiceMetadata
```
loic-sharma: >IFileServiceMetadata [](start = 21, length = 20)

This type name doesn't match the file name

src/NuGetGallery.Core/Services/CorePackageFileService.cs
```diff
@@ -13,10 +13,12 @@ namespace NuGetGallery
     public class CorePackageFileService : ICorePackageFileService
     {
         private readonly ICoreFileStorageService _fileStorageService;
+        private readonly IFileServiceMetadata _fileServiceMetadata;
```
loic-sharma: >_fileServiceMetadata [](start = 46, length = 20)

This seems a little long. What do you think of naming this field `_metadata`?

src/NuGetGallery.Core/Services/IMetadataFileService.cs
shishirx34: I think the better name would be `IFileMetadataService`, rename everywhere as well with field names.

src/NuGetGallery.Core/Services/PackageFileServiceMetadata.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    public class PackageFileServiceMetadata : IFileServiceMetadata
```
shishirx34: `PackageFileMetadataService`

agr: Is it even a service?

src/NuGetGallery.Core/Services/IMetadataFileService.cs
cristinamanum: Updated to be IFileMetadataService

src/NuGetGallery.Core/Services/CorePackageFileService.cs
cristinamanum: Updated

src/NuGetGallery.Core/Services/PackageFileServiceMetadata.cs
cristinamanum: It is metadata for service. Updated to be PackageFileMetadataService

PR #6150, opened 2018-07-09
Auto-add Microsoft a co-owner for packages pushed by blessed Organizations

### LINKED ISSUE #1419: Add galops command to audit package data
[anurse] 
We need a galops command to scan Blob storage and the Database and create the following lists:

```
Things which are in DB but not packages containing in Blob storage
Things which are in packages container but not DB
Things which are in packages but not packagebackups
```

Ported from https://github.com/NuGet/NuGetOperations/issues/35

xavierdecoster:
Fixes https://github.com/NuGet/Engineering/issues/1419

Opening this PR already for early review and feedback.
Note: this focuses on the package push API side of things. Not covering the web upload scenario, yet.


src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -32,6 +33,36 @@ public void SendPackageAddedNotice(Package package, string packageUrl, string pa
...
+                AddOwnersSubscribedToPackagePushedNotification(package.PackageRegistration, mailMessage);
+
+                if (mailMessage.To.Any())
+                {
+                    SendMessage(mailMessage, copySender: false);
+                }
+            }
+        }
+
+        public void SendPackageAddedWithWarningsNotice(Package package, string packageUrl, string packageSupportUrl, string emailSettingsUrl, IEnumerable<string> warningMessages)
```
chenriksson: Why copy, instead of adding optional `warningMessages` parameter to existing `SendPackageAddedNotice` API?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -520,11 +532,23 @@ public virtual Task<ActionResult> CreatePackagePost()
...
                             if (!(ConfigurationService.Current.AsynchronousPackageValidationEnabled && ConfigurationService.Current.BlockingAsynchronousPackageValidationEnabled))
                             {
-                                // Notify user of push unless async validation in blocking mode is used
-                                MessageService.SendPackageAddedNotice(package,
-                                    Url.Package(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
-                                    Url.ReportPackage(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
-                                    Url.AccountSettings(relativeUrl: false));
+                                if (!packagePolicyResult.HasWarnings)
+                                {
+                                    // Notify user of push with warnings unless async validation in blocking mode is used
```
chenriksson: nit: comment says this is push with warnings, but that's the `else` block

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// </summary>
+        public override SecurityPolicyResult Evaluate(PackageSecurityPolicyEvaluationContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // This particular package policy assumes the existence of a 'Microsoft' user.
+            // Succeed silentely (effectively ignoring this policy when enabled) when that user does not exist.
```
chenriksson: nit: silently

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+                return SecurityPolicyResult.SuccessResult;
+            }
+
+            // If the package being evaluated has not been registered yet (new package ID),
+            // this policy requires the prefix to be reserved.
+            var isWarning = false;
+            if (context.ExistingPackageRegistration == null)
+            {
+                // We are evaluating a newly pushed package with a new ID.
+                var packageId = context.Package.PackageRegistration.Id;
```
chenriksson: nit: `packageRegistrationId`?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+
+                // If the generated PackageRegistration is not marked as verified (by `PackageUploadService.GeneratePackageAsync`),
+                // the account pushing the package has not registered the prefix yet.
+                if (!context.Package.PackageRegistration.IsVerified)
+                {
+                    // The owner has not reserved the prefix. Check whether the Microsoft user has.
+                    var prefixIsReservedByMicrosoft = IsPrefixReservedByAccount(context.EntitiesContext, microsoftUser, packageId);
```
chenriksson: If current owner doesn't own prefix, but MS does -- would the push have already failed before we got here?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+
+                    // If the prefix has not been reserved by the 'Microsoft' user either,
+                    // then generate a warning which will result in an alternate email being sent to the package owners when validation of the metadata succeeds, and the package is pushed.
+                    isWarning = !prefixIsReservedByMicrosoft;
+
+                    // Mark the package as verified if the prefix has been reserved by Microsoft.
+                    context.Package.PackageRegistration.IsVerified = prefixIsReservedByMicrosoft;
```
chenriksson: Do we need to monitor for these warnings, and possibly register these prefixes for these unverified packages to the team accounts we notice? I assume PMs would need to notify us, since they manage the MS account?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+            // All good!
+            return SecurityPolicyResult.SuccessResult;
+        }
+
+        private bool IsPrefixReservedByAccount(IEntitiesContext entitiesContext, User account, string id)
+        {
+            var reservedNamespacesForId = (from request in entitiesContext.Set<ReservedNamespace>()
+                                           where (request.IsPrefix && id.StartsWith(request.Value))
+                                               || (!request.IsPrefix && id.Equals(request.Value))
+                                           select request).ToList();
```
chenriksson: Should this logic be in the `ReservedNamespaceService`, like the other checks?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+                    .ToList()
+                    .AsReadOnly();
+
+            return ownedMatchingReservedNamespaces.Any();
+        }
+
+        private bool IsPackageMetadataCompliant(Package package)
+        {
+            // Author validation
+            if (!package.FlattenedAuthors.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries).Contains(MicrosoftUsername))
```
chenriksson: `Contains` should probably use `IEqualityComparer` w/o ignore case?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+            if (microsoftUser == null)
+            {
+                // This may happen on gallery deployments that don't have a 'Microsoft' user.
+                return SecurityPolicyResult.SuccessResult;
+            }
+
+            // If the package being evaluated has not been registered yet (new package ID),
+            // this policy requires the prefix to be reserved.
+            var isWarning = false;
+            if (context.ExistingPackageRegistration == null)
```
skofman1: Can we split this policy into two policies:
1. Policy that ensure reserved prefix
2. Policy that validates metadata. The metadata policy will be general, i.e. the required values will be loaded from the DB config. This will also be useful to configure it differently for teams that use different copyrights, like Xamarin.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+            if (!package.FlattenedAuthors.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries).Contains(MicrosoftUsername))
+            {
+                return false;
+            }
+
+            // Copyright validation
+            if (!string.Equals(package.Copyright, "(c) Microsoft Corporation. All rights reserved.")
+                && !string.Equals(package.Copyright, "© Microsoft Corporation. All rights reserved."))
+            {
+                return false;
```
skofman1: We should give the user a good error message about why the package is  not compliant.

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -118,6 +138,75 @@ public virtual IEnumerable<IUserSecurityPolicySubscription> UserSubscriptions
...
 
+        /// <summary>
+        /// Evaluate package security policies for the specified action.
+        /// </summary>
+        /// <param name="action">Gallery action to evaluate.</param>
+        /// <param name="httpContext">Current http context.</param>
+        /// <param name="package">The package to evaluate.</param>
+        /// <param name="packageRegistration">The package registration. Will be <code>null</code> if the <paramref name="package"/> has a new package ID.</param>
+        /// <returns></returns>
+        public Task<SecurityPolicyResult> EvaluatePackagePoliciesAsync(
```
skofman1: The name implies that the security policy is on a package, and not on a user. But it's actually on a user... perhaps we should use the existing method name (EvaluateUserPoliciesAsync)? Same thought about other parts of the code. What do you think?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+namespace NuGetGallery.Security
+{
+    /// <summary>
+    /// User security policy that requires packages pushed by this user to be compliant with Microsoft package policy.
+    /// </summary>
+    public class RequireMicrosoftPackageCompliancePolicy : PackageSecurityPolicyHandler, IUserSecurityPolicySubscription
+    {
+        public const string PolicyName = nameof(RequireMicrosoftPackageCompliancePolicy);
+
+        internal const string MicrosoftUsername = "Microsoft";
```
skofman1: can Microsoft be a parameter to the policy (from DB)?

src/NuGetGallery.Core/Services/CoreMessageService.cs
xavierdecoster: Ah, good one! Fixing :)

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Are usernames case-sensitive? I thought they weren't, so we should ignore casing?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: The tricky part is that I don't have access to the `ReservedNamespaceService` and I can't inject it without causing a circular dependency loop (which is why the `IEntitiesContext` is passed along instead, which is the most generic way of providing access to the db).

src/NuGetGallery/Security/SecurityPolicyService.cs
xavierdecoster: They evaluate a package, not a user. The user is the subscriber to the policy, but the object being evaluated by the policy is the package. 

The API is also quite different, as a `Package`, `PackageRegistration` and `IEntitiesContext` are provided to the handlers to do their thing. They are also evaluated at a different time in the flow. To me, they are similar, yet different.

src/NuGetGallery/Controllers/ApiController.cs
xavierdecoster: Fixed.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Thx! Fixed.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Fixed.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Yes... good catch!

The following error is shown when this happens today:

> Response status code does not indicate success: 409 (This package ID has been reserved. Please request access to upload to this reserved namespace from the owner of the reserved prefix, or re-upload the package with a different ID.).

This happens [in `ApiController` on line 424](https://github.com/NuGet/NuGetGallery/pull/6150/files#diff-5f932537400045c63d3058184adb0647R424), which happens before evaluating the package policies (as we only read the Stream and construct the package later in that call...).

Does that mean that, for users that have this `RequireMicrosoftPackageCompliancePolicy` set, we should somehow skip that check?

According to the requirements set forth in NuGet/Engineering#1419, we generate a warning (to be sent in the publish successful email), and continue validation, which results in successful package push if the package is deemed compliant.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: As per the requirements set forth in NuGet/Engineering#1419, the owners will receive an email containing warning messages, and instructions to reserve the prefix.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: I'm assuming you refer to the JSON `State` stored with the policy? Sure thing.

There's likely more to it than just changing this username to make this a reusable policy (e.g. copyright notice, maybe additional or different validations).
I thought we discussed that for v1, this was not critical. 

To be able to use this in a flexible way, we'll need to work on admin UI, right?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
chenriksson: They're not case sensitive, but I believe code analysis usually suggests that equality comparer be used in these APIs. It may be good to be explicit anyways - using InvariantCulture, instead of InvariantCultureIgnoreCase.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+                {
+                    // The owner has not reserved the prefix. Check whether the Microsoft user has.
+                    var prefixIsReservedByMicrosoft = IsPrefixReservedByAccount(context.EntitiesContext, microsoftUser, packageRegistrationId);
+
+                    // If the prefix has not been reserved by the 'Microsoft' user either,
+                    // then generate a warning which will result in an alternate email being sent to the package owners when validation of the metadata succeeds, and the package is pushed.
+                    isWarning = !prefixIsReservedByMicrosoft;
+
+                    // Mark the package as verified if the prefix has been reserved by Microsoft.
+                    context.Package.PackageRegistration.IsVerified = prefixIsReservedByMicrosoft;
```
scottbommarito: This feels hacky. By setting the flag here directly we are bypassing the normal code path that's used to set this flag. Furthermore, this could lead to many issues because most of the code does not expect a package to be marked as verified when it doesn't actually have any owners that own its reserved namespace.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+            if (!packageIsCompliant)
+            {
+                // Microsoft package policy not met.
+                return SecurityPolicyResult.CreateErrorResult(Strings.SecurityPolicy_RequireMicrosoftPackageMetadataComplianceForPush);
+            }
+
+            // Automatically add 'Microsoft' as co-owner when metadata is compliant.
+            if (!context.Package.PackageRegistration.Owners.Select(o => o.Username).Contains(MicrosoftUsername, StringComparer.OrdinalIgnoreCase))
+            {
+                context.Package.PackageRegistration.Owners.Add(microsoftUser);
```
scottbommarito: `PackageService.AddPackageOwnerAsync` is the intended way to add owners to a `PackageRegistration`. We are bypassing the normal code path here which WILL result in the required signer of the package not being set as expected. I know there is also a method that will not only add the owner but also mark the package as verified if the owner owns the reserved namespace it's in. I would suggest doing that instead of making these changes to the package manually.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
scottbommarito: Can you explain the circular dependency loop? I don't see it here, and I think it would be easy for `ApiController` to pass a `ReservedNamespaceService` to the context.

src/NuGetGallery/Strings.resx
```diff
@@ -911,4 +911,10 @@ If you would like to update the linked Microsoft account you can do so from the
   <data name="FailedValidationHardDeleteReason" xml:space="preserve">
     <value>Automatic hard-delete for reupload of package that failed validation</value>
   </data>
+  <data name="SecurityPolicy_RequireMicrosoftPackageMetadataComplianceForPush" xml:space="preserve">
+    <value>Package is not compliant with metadata requirements for Microsoft packages on NuGet.org. Go to &lt;insert URL here&gt; for more information.</value>
```
scottbommarito: reminder that `insert URL` needs to be updated when the URL is available

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+                    .ToList()
+                    .AsReadOnly();
+
+            return ownedMatchingReservedNamespaces.Any();
+        }
+
+        private bool IsPackageMetadataCompliant(Package package)
+        {
+            // Author validation
+            if (!package.FlattenedAuthors.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries).Contains(MicrosoftUsername))
```
scottbommarito: Is there anything else that parses the list of `FlattenedAuthors`? I feel like there should be a shared API for parsing the `FlattenedAuthors` into a string list.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,158 @@
...
+                return false;
+            }
+
+            // If we made it this far, the package metadata is compliant.
+            return true;
+        }
+
+        public Task OnSubscribeAsync(UserSecurityPolicySubscriptionContext context)
+        {
+            return Task.CompletedTask;
```
scottbommarito: Maybe we should enumerate through the user's packages and add `Microsoft` as a package owner if the package passes the metadata requirements when a user is onboarded to this policy.

src/NuGetGallery/Security/SecurityPolicyResult.cs
```diff
@@ -42,5 +62,9 @@ public static SecurityPolicyResult CreateErrorResult(string errorMessage)
         /// Error message, if the security policy criteria was not met.
         /// </summary>
         public string ErrorMessage { get; }
+
+        public bool HasWarnings { get { return _warningMessages.Any(); } }
```
scottbommarito: nit: `public bool HasWarnings => _warningMessages.Any();`

src/NuGetGallery/Security/SecurityPolicyResult.cs
```diff
@@ -42,5 +62,9 @@ public static SecurityPolicyResult CreateErrorResult(string errorMessage)
...
+
+        public IReadOnlyCollection<string> WarningMessages { get { return _warningMessages; } }
```
scottbommarito: nit: `public IReadOnlyCollection<string> WarningMessages => _warningMessages;`

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
shishirx34: I agree, I haven't yet reviewed the code, but, we shouldn't mark the package as verified during upload, rather we should allow the upload(the restriction on the non-owner of namespace) and when Microsoft is added, it should auto add the IsVerified since it owns the namespace it belongs to.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
skofman1: I think we have two options here:
1. Keep the policies general and configurable, and add the "Microsoft" specific settings in the Subscription level. i.e, there will be a MicrosoftTeam subscription that will create policies with the specific configs for user name, copyright, etc. The admin UI can remain simple. In edge cases like Xamarin packages needing different config, we can go to DB.
2. Keep the subscription general as well, and have the user provide the config in the admin UI. It's more UI work but helps avoid harding "Microsoft" in our code.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
karann-msft: can we make such emails also go to account@nuget.org? That's the alias we monitor for prefix reservation.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
karann-msft: Per https://github.com/NuGet/Engineering/issues/1419
The package is not compliant with metadata requirements for Microsoft packages on NuGet.org. Go to http://aka.ms/Microsoft-NuGet-Compliance for more information.

src/NuGetGallery/Strings.resx
karann-msft: The package is not compliant with metadata requirements for Microsoft packages on NuGet.org. Go to http://aka.ms/Microsoft-NuGet-Compliance for more information.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
karann-msft: We should not! There are packages that have been abandoned and will not be updated. Such packages have been locked and Microsoft should not be added as co-owner. (Although the likelihood that an existing non-Microsoft co-owned package is compliant with the metadata is slim to none.)

karann-msft: That sounds like a good idea. We should also unlock the package if it is locked as part of adding Microsoft a co-owner.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: I believe you meant `InvariantCultureIgnoreCase` to be used, as the comparison is not case-sensitive. Fixed.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: We can, but it's not the policy who is in charge of sending the emails. It's the `ApiController` that sends them.

Do we want to include `account@nuget.org` in all package added notices that generated warnings during package policy validation? (currently, this is the only package policy)

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: I did exactly that before changing it again to the current approach, and had a circular dependency error when injecting both `ReservedNamespaceService` and `UserService` instead of `IEntitiesContext`, as `ReservedNamespaceService` also depends on `UserService`.

src/NuGetGallery/Strings.resx
xavierdecoster: Thanks! :)

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Thanks for the pointers! Those calls all commit to the db, right?
This "manual" stuff may happen before the registration even exists in the db... (e.g. when pushing a new ID). Committing to the db should remain a single atomic operation handled by the `IUploadService`. I may need to tweak those existing APIs with a `bool commitChanges = true` argument.

xavierdecoster: Adding such overload for `SetRequiredSignerAsync` is maybe not what we're after...

scottbommarito: Yep, we regularly add `bool commitChanges = true` to our existing APIs if we need to call them but not commit. I would do that and then call the API here.

After looking through the code, it seems like `PackageOwnershipManagementService.AddPackageOwnerAsync` is the API you should be calling. I would add the `commitChanges` parameter there and to any APIs that it calls.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
karann-msft: only in case of prefix not reserved warning, we should send to account@nuget.org. If this is easily doable, great! If not, that's ok too, Daniel and I can just create a new filter based on the subject line :)

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Adding the `commitChanges` parameter to `IPackageOwnershipManagementService.AddPackageOwnerAsync` (which creates a db transaction) affects many other APIs called by this method.

Including the `PackageService.SetRequiredSignerAsync` method. How would you add this parameter in there, which also saves audit records?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: It's non-trivial to make that distinction currently, so I guess the filter is the way to go for now :)

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: This all makes me wonder if it would make sense for these new package policies to have a Pre-Upload and Post-Upload evaluation moment... 

Check prefix ownership and metadata validation pre-upload, and if the policy generates no errors, commit the package (and registration in case of new id), then continue with post-upload actions as defined by the policy and based on pre-upload evaluation result...

However, the co-owner must be added before signing validation.

Thoughts? Could this potentially mess up async validation? (what triggers the async validation?)

src/NuGetGallery/Strings.resx
```diff
@@ -914,4 +914,10 @@ If you would like to update the linked Microsoft account you can do so from the
   <data name="PackageFileTooLarge" xml:space="preserve">
     <value>The package file exceeds the size limit. Please try again.</value>
   </data>
+  <data name="SecurityPolicy_RequireMicrosoftPackageMetadataComplianceForPush" xml:space="preserve">
+    <value>The package is not compliant with metadata requirements for Microsoft packages on NuGet.org. Go to http://aka.ms/Microsoft-NuGet-Compliance for more information.</value>
+  </data>
+  <data name="SecurityPolicy_RequirePackagePrefixReserved" xml:space="preserve">
+    <value>You have not published a package with this prefix in the past. This means other users may be able to push packages starting with the same prefix. Contact account@nuget.org to reserve the prefix. Learn more about Package ID prefix reservation.</value>
```
loic-sharma: >You have not published a package with this prefix in the past. This means other users may be able to push packages starting with the same prefix. Contact account@nuget.org to reserve the prefix. Learn more about Package ID prefix reservation. [](start = 11, length = 242)

The "learn more" bit is missing a link. What do you think of this wording?

> This is the first time you've published a package with this prefix. Other users may be able to push packages with similar IDs. You can contact account@nuget.org to reserve the prefix. Go to <insert URL> for more information.

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -27,10 +28,24 @@ public CoreMessageService(IMailSender mailSender, ICoreMessageServiceConfigurati
...
+        public void SendPackageAddedNotice(Package package, string packageUrl, string packageSupportUrl, string emailSettingsUrl, IEnumerable<string> warningMessages = null)
         {
-            string subject = $"[{CoreConfiguration.GalleryOwner.DisplayName}] Package published - {package.PackageRegistration.Id} {package.Version}";
+            bool hasWarnings = warningMessages != null && warningMessages.Any();
+
+            string subject;
+            var warningMessagesPlaceholder = string.Empty;
+            if (hasWarnings)
+            {
+                subject = $"[{CoreConfiguration.GalleryOwner.DisplayName}] Package published with Warnings - {package.PackageRegistration.Id} {package.Version}";
```
loic-sharma: >Warnings [](start = 98, length = 8)

Should this be lower cased?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
scottbommarito: So, to clarify

- `UserService` depends on `SecurityPolicyService`
- `ReservedNamespaceService` depends on `UserService`
- `SecurityPolicyService` creates the `PackageSecurityPolicyEvaluationContext` to pass to the `RequireMicrosoftPackageCompliancePolicy` using the services it's been injected with

Regardless of this loop, I think it's really important for you to be able to get these services for your DI. After discussing this a bit with @loic-sharma, he suggested making `SecurityPolicyService` dependent on the DI container (`IContainer`?) and then using it to get the services. There might be a better way, but I think it's important you can get access to whatever services you need in the policy handler.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
scottbommarito: Any additional commits in the upload flow WILL break async validation. We had a live site that was caused by a similar bug. You may be able to get away with additional commits if they happen AFTER the existing commit but it sounds risky, especially if the first commit succeeds and the second fails.

Regarding adding `commitChanges`, I would check around, but I think in most places we just change
```
await _entitiesContext.SaveChangesAsync();
```
to
```
if (commitChanges)
{
    await _entitiesContext.SaveChangesAsync();
}
```
while ignoring anything else going on in the method, including audit records. I think we already accept that it's possible that an audit record may be emitted for an operation that failed afterwards.

src/NuGetGallery/Strings.resx
xavierdecoster: @karann-msft what is the link to be used here? (to learn more about prefix reservation)

src/NuGetGallery.Core/Services/CoreMessageService.cs
xavierdecoster: Fixed.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: I agree it is important to gain access to those services... 

Do we have an example where we inject the `Autofac.IContainer` anywhere? Or should we use `DependencyResolver.Current`?

loic-sharma: To clarify @scottbommarito's comment: I suggested that you use the service locator pattern. From [this StackOverflow post](https://stackoverflow.com/questions/7190795/using-autofac-as-a-service-locator), it seems that the proper pattern in Autofac is to accept an `IComponentContext`.

 I don't know off-hand if the Gallery has used this pattern before. However, I do know that the Service Bus jobs use the service locator pattern using ASP.NET's dependency injection libraries [here](https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.ServiceBus/ScopedMessageHandler.cs#L49).

agr: Circular dependencies are typically resolved by requesting `Lazy<X>` instead of `X` in one of the constructors to break the loop. Is it possible to do in this case?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: @skofman1 @chenriksson any thoughts on this question?

> Does that mean that, for users that have this `RequireMicrosoftPackageCompliancePolicy` set, we should somehow skip that check?

scottbommarito: I wouldn't skip the check--the more this Microsoft upload flow differs from the normal upload flow, the more likely we are to have bugs.

@skofman1 mentioned that she thinks this policy should be split into 2, one handling the reserved namespace ownership and the other verifying the metadata of the package. I think the reserved namespace policy could be evaluated after getting the ID and version (before the ownership checks are) and the metadata policy could trigger where the combined policy does.

This would enable you to check the reserved namespaces that the package applies to and add Microsoft as an owner of the package before the existing ownership checks are so you won't have to skip them.

xavierdecoster: @scottbommarito in the event of a newly uploaded package id for which the prefix is not reserved by either the account pushing or Microsoft, we should still continue to validate metadata (see flowchart on NuGet/Engineering#1419), so there's an ordered dependency between such two policies (which is why I considered them one).

Furthermore, [the current api scope validation would prevent the code to reach the metadata validation part](https://github.com/NuGet/NuGetGallery/blob/166d4c21134c11f0800aa7e0f0026a53cc583d48/src/NuGetGallery/Controllers/ApiController.cs#L311).

In fact, the only thing that the *prefix policy* does according to the flowchart, is deciding whether an alternate email is to be sent upon successful validation of the *metadata policy*...

scottbommarito: I think the flowchart is not important. The goal of the feature is that Microsoft is automatically added as a co-owner of packages owned by certain users, those packages must abide by certain metadata rules, and that an email must be sent when those packages are not in a reserved namespace. As long as those goals are accomplished, I don't think the ordering is important, especially given that the current ordering isn't working with our existing flow.

I think to resolve your issue, you should add Microsoft as a co-owner of the package (before the ownership check), make sure the ownership check includes Microsoft, and then do all the validations that you currently do where the evaluation currently is.

skofman1: Agree with Scott on this. I would keep the code as simple as possible to achieve the goals:
1. Have a separate policy for metadata - if policy exists for account fail package push. In this case we don't need to worry about reserved namespaces at all.
2.  Have a separate policy for "auto add c-owner" - make sure to run this verification after the metadata one in the code (this might be as simple as putting it in the next block) if you see this on the account add co-owner and send mail if needed.


Regarding:
> Does that mean that, for users that have this RequireMicrosoftPackageCompliancePolicy set, we should somehow skip that check?

Lets keep it simple and continue with the check. When a new MSFT team onboards we will manually assign to them all prefixes and security policies they need.

xavierdecoster: So we can close up on this, and conclude that **we do want an upload to fail if Microsoft owns the prefix but the pushing user does not**.

This simplifies things a lot, and all validations can be part of a single package policy evaluated where it currently is (after api scope evaluation and before committing the package).

skofman1:
>                             if (!(ConfigurationService.Current.AsynchronousPackageValidationEnabled && ConfigurationService.Current.BlockingAsynchronousPackageValidationEnabled))

on nuget.org async validation is enabled, so this code block will never be called

---
Refers to: src/NuGetGallery/Controllers/ApiController.cs:533 in 6228911. [](commit_id = 6228911d9196b59402fffc2c12f6350378d1eb02, deletion_comment = False)

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,148 @@
...
+                        out var ownedMatchingReservedNamespaces);
+
+                    // If the prefix has not been reserved by the 'Microsoft' user either,
+                    // then generate a warning which will result in an alternate email being sent to the package owners when validation of the metadata succeeds, and the package is pushed.
+                    isWarning = !prefixIsReservedByMicrosoft;
+                }
+            }
+
+            // Evaluate Microsoft metadata validations
+            var packageIsCompliant = IsPackageMetadataCompliant(context.Package);
```
skofman1: you should start with validating the metadata. If this fails the prefix check is not needed.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,148 @@
...
+                // Microsoft package policy not met.
+                return SecurityPolicyResult.CreateErrorResult(Strings.SecurityPolicy_RequireMicrosoftPackageMetadataComplianceForPush);
+            }
+
+            // Automatically add 'Microsoft' as co-owner when metadata is compliant.
+            if (!context.Package.PackageRegistration.Owners.Select(o => o.Username).Contains(MicrosoftUsername, StringComparer.OrdinalIgnoreCase))
+            {
+                // This will also mark the package as verified if the prefix has been reserved by Microsoft.
+                // The entities context is committed later as a single atomic transaction (see PackageUploadService).
+                await context.PackageOwnershipManagementService.AddPackageOwnerAsync(context.Package.PackageRegistration, microsoftUser, commitChanges: false);
```
skofman1: How about this order of validations:
1. Check metadata - exit with failure if not compliant.
2. Add MSFT as co-owner - this will set the "isVerified" if needed.
3. Check if "IsVerified" is true - return warning if not.

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -27,10 +28,24 @@ public CoreMessageService(IMailSender mailSender, ICoreMessageServiceConfigurati
         public IMailSender MailSender { get; protected set; }
         public ICoreMessageServiceConfiguration CoreConfiguration { get; protected set; }
 
-        public void SendPackageAddedNotice(Package package, string packageUrl, string packageSupportUrl, string emailSettingsUrl)
+        public void SendPackageAddedNotice(Package package, string packageUrl, string packageSupportUrl, string emailSettingsUrl, IEnumerable<string> warningMessages = null)
```
skofman1: This mail will never be sent by the Gallery on nuget.org (but by the orchestrator). You should create a separate mail for the warning.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        private bool IsPackageMetadataCompliant(Package package)
+        {
+            // Author validation
+            if (!package.FlattenedAuthors
+                .Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries)
+                .Contains(MicrosoftUsername, StringComparer.InvariantCultureIgnoreCase))
+            {
+                return false;
```
joelverhagen: >false [](start = 23, length = 5)

How does the caller know what the specific problem is?

Shouldn't we be returning an error message, e.g. "The package copyright does not match the policy blah blah".

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -489,6 +490,17 @@ public virtual Task<ActionResult> CreatePackagePost()
...
 
+                            var packagePolicyResult = await SecurityPolicyService.EvaluatePackagePoliciesAsync(
+                                securityPolicyAction, 
+                                HttpContext, 
+                                package,
+                                packageRegistration);
+
+                            if (!packagePolicyResult.Success)
+                            {
+                                return new HttpStatusCodeWithBodyResult(HttpStatusCode.BadRequest, packagePolicyResult.ErrorMessage);
```
joelverhagen: >HttpStatusCodeWithBodyResult [](start = 43, length = 28)

We have to now performed validations AFTER adding the entity to the entity context.

This is only okay if we are very rigorous about not committing (SaveChangesAsync on entity context) until we are very sure all validations are done.

There was a time in the past when uploading a readme Markdown during UI upload actually did a commit early. This was a bug that went undetected for months.

Would it be possible to run this validation before adding entities to the entity context?

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
```diff
@@ -0,0 +1,148 @@
...
+            if (microsoftUser == null)
+            {
+                // This may happen on gallery deployments that don't have a 'Microsoft' user.
+                return SecurityPolicyResult.SuccessResult;
+            }
+
+            // If the package being evaluated has not been registered yet (new package ID),
+            // this policy requires the prefix to be reserved.
+            var isWarning = false;
+            if (context.ExistingPackageRegistration == null)
```
joelverhagen: >ExistingPackageRegistration [](start = 24, length = 27)

Why isn't this a boolean, `PackageRegistrationAlreadyExists`? We don't actually need the existing instance since it is on Package.PackageRegistration, right?

joelverhagen:
>                             if (!(ConfigurationService.Current.AsynchronousPackageValidationEnabled && ConfigurationService.Current.BlockingAsynchronousPackageValidationEnabled))

Great point! Plumbing the warnings down to orchestrator would be a mess.

Couldn't we do the warning emails asynchronously for "blessed" owners? This would also mean we could check for existing IDs that don't fall under a namespace.

---
In reply to: [407812683](https://github.com/NuGet/NuGetGallery/pull/6150#issuecomment-407812683) [](ancestors = 407812683)

---
Refers to: src/NuGetGallery/Controllers/ApiController.cs:533 in 6228911. [](commit_id = 6228911d9196b59402fffc2c12f6350378d1eb02, deletion_comment = False)

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -489,6 +490,17 @@ public virtual Task<ActionResult> CreatePackagePost()
                                 owner,
                                 currentUser);
 
+                            var packagePolicyResult = await SecurityPolicyService.EvaluatePackagePoliciesAsync(
```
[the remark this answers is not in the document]
xavierdecoster: I can move it up, so this is called before creating the `Package` and BEFORE adding the entity to the entity context, but that would likely mean we make the call to check for IsVerified for the pushing user twice (once when evaluating the policy, once when creating the package).

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: True! Fixing.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Updated.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Updated.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Sounds good. In that case, I don't even need the `ReservedNamespaceService` in this policy, as I can simply check on `context.Package.PackageRegistration.IsVerified` at the end of the policy evaluation.

src/NuGetGallery/Controllers/ApiController.cs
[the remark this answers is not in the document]
joelverhagen: I see, there is a sneaky dependency on `IsVerified`. Makes sense.

src/NuGetGallery.Core/Services/CoreMessageService.cs
joelverhagen: Great point! Plumbing the warnings down to orchestrator would be a mess.

Couldn't we do the warning emails asynchronously for "blessed" owners? This would also mean we could check for existing IDs that don't fall under a namespace.

xavierdecoster: Where do you suggest I add this warning email in code?

I could add the following in `ApiController` after the [`if` block](https://github.com/NuGet/NuGetGallery/pull/6150/files#diff-5f932537400045c63d3058184adb0647L521) that checks for blocking validation mode:

```
// Emit warning messages if any
else if (packagePolicyResult.HasWarnings)
{
  // Notify user of push unless async validation in blocking mode is used
  MessageService.SendPackageAddedNotice(package,
  Url.Package(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
  Url.ReportPackage(package.PackageRegistration.Id, package.NormalizedVersion, relativeUrl: false),
  Url.AccountSettings(relativeUrl: false),
  packagePolicyResult.WarningMessages);
}
```

src/NuGetGallery/Controllers/ApiController.cs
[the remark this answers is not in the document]
xavierdecoster: I updated `PackageSecurityPolicyContext` to no longer depend on `IEntitiesContext` directly, but instead leverage `IUserService` (and `IPackageOwnershipManagementService`).

I added the following test that asserts we call their APIs with `commitChanges: false`.
https://github.com/NuGet/NuGetGallery/pull/6150/files#diff-3e9af76f191871062da1ac6b81a382ecR48

xavierdecoster:
In today's update:
* new `MicrosoftTeamSubscription`
image
* the new `RequirePackageMetadataCompliancePolicy` is now generic, and validation values are passed in via policy `State` (which is configurable in the db)
* no more using `IEntitiesContext` directly in `RequirePackageMetadataCompliancePolicy` (instead using `IUserService` and `IPackageOwnershipManagementService`
* a new unit test asserts that the new package security policy calls service APIs using `commitChanges:false`

Open questions:
* if package policy results in warnings, a sync message is currently sent when gallery is configured to use async validation
* there's still a todo in the `OnSubscribeAsync` of `MicrosoftTeamSubscription`: should that be part of v1?

src/NuGetGallery.Core/Services/CoreMessageService.cs
skofman1: I like @joelverhagen 's idea to send warning e-mails about packages that don't fall under a namespace from a separate job. However, it will be much more work, so I vote to keep it in the gallery.
One thing I would do, is have a separate e-mail with this warning and not the "package added notice" because people will get the same e-mail twice - one from Gallery and one from Orchestrator and will find it confusing.

src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
```diff
@@ -0,0 +1,159 @@
...
+            }
+
+            // Evaluate package metadata validations
+            if (!IsPackageMetadataCompliant(context.Package, state, out var complianceFailures))
+            {
+                // Package policy not met.
+                return SecurityPolicyResult.CreateErrorResult(
+                    string.Format(
+                        CultureInfo.CurrentCulture,
+                        Strings.SecurityPolicy_RequireMicrosoftPackageMetadataComplianceForPush,
```
skofman1: you removed "Microsoft" from this code, but it's still here. I guess there's no easy way to make the string generic given the URL is hardcoded.

src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
```diff
@@ -0,0 +1,159 @@
...
+                .Where(p => !string.IsNullOrEmpty(p.Value))
+                .Select(p => JsonConvert.DeserializeObject<State>(p.Value));
+
+            // TODO: what if there are multiple?
+            return policyStates.First();
+        }
+
+        public class State
+        {
+            [JsonProperty("requiredCoOwnerUsername")]
```
skofman1: perhaps shorter json property names?

src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
```diff
@@ -0,0 +1,159 @@
...
+            {
+                complianceFailures.Add(Strings.SecurityPolicy_RequiredProjectUrlMissing);
+            }
+
+            return !complianceFailures.Any();
+        }
+
+
+        /// <summary>
+        /// In case of multiple, select the max of the minimum required protocol versions.
```
skofman1: copy-pasted comment

src/NuGetGallery/Security/SecurityPolicyService.cs
```diff
@@ -37,11 +42,19 @@ private static Lazy<IEnumerable<UserSecurityPolicyHandler>> _userHandlers
...
         }
 
-        public SecurityPolicyService(IEntitiesContext entitiesContext, IAuditingService auditing, IDiagnosticsService diagnostics, IAppConfiguration configuration)
+        public SecurityPolicyService(
+            IEntitiesContext entitiesContext,
+            IAuditingService auditing,
+            IDiagnosticsService diagnostics,
+            IAppConfiguration configuration,
+            IComponentContext componentContext,
+            MicrosoftTeamSubscription microsoftTeamSubscription = null)
```
skofman1: do we still need DI here?

src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
xavierdecoster: Oversight :) The error message could be part of the policy state.

src/NuGetGallery/Security/SecurityPolicyService.cs
xavierdecoster: I was anticipating on the TODO in the OnSubscribeAsync method of the subscription, which would need DI.

src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
xavierdecoster: Oops. Fixed :)

src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
xavierdecoster: Better?

xavierdecoster:
@skofman1 oops, good catch :) I've pushed an update that addresses that scenario. Notice the subtle difference in email subject (`published` vs `pushed`) and email body (no link to `email settings` when `push` warning message is sent).

Merging is blocked by @scottbommarito's request for changes, but if all good, feel free to merge and deploy to DEV later today.

src/NuGetGallery/Strings.resx
karann-msft: @xavierdecoster yes that looks right. sorry for the late reply.

xavierdecoster:
@scottbommarito your review is still requesting changes. Can you please check if changes are good, or let me know what needs changed still?

@karann-msft @skofman1 had an offline discussion with @joelverhagen regarding the email with warnings message, and he had an interesting suggestion:

> **Is it necessary the email is sent immediately?**
> 
> Why don’t we just create a dedicated job to scan Microsoft owner IDs every day and send an email once on IDs that don’t have reserved namespaces. This way we get the benefit for existing IDs too. Or we can make it a report that PMs see and reach out to teams manually. Since this is sent from the same sender as normal package push and to the same recipient, don’t you think it’s likely this message will go unnoticed? I imagine the package push notification is noise for a lot of users.

I guess it all depends on the answer to the above question, and the amount of effort we want to put into it.

skofman1:
Xavier, as I commented above regrading Joel's suggestion: it's good, but let's not invest the work now, and leave the code as is.
Feel free to merge if you think Scott's comments were addressed.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: We're no longer setting `IsVerified` manually. This is now done by calling the `IPackageOwnershipManagementService.AddPackageOwnerAsync` API.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Handled.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: This seems to be the only place in the gallery code where we need to do this.

src/NuGetGallery/Security/RequireMicrosoftPackageCompliancePolicy.cs
xavierdecoster: Created [tracking issue](https://github.com/NuGet/Engineering/issues/1598) for this suggestion.

Issue #4941, opened 2017-11-01
Users should be able to see package repository metadata


joelverhagen:
## Overview

Since August 2016, the NuGet package manifest (.nuspec) has supported the `<repository>` XML element. This is an optional metadata element with two optional attributes: `type` and `url`. Presumably, the purpose of this element is to communicate to package consumers the version control type and location of the source code used to produce the package.

Currently, I see no NuGet documentation about this property and no support for it in the various NuGet UIs (Visual Studio and the nuget.org web gallery).

This metadata adds additional valuable information about a package that augments the existing `<projectUrl>` package metadata field. Project URL seems to me like a "product" or "company" specific website and repository URL is where the source code (and maybe even bug tracker) lives. If a project has project site that is different than their GitHub repository, this differentiation is helpful to package consumers.

This property is already widely adopted on nuget.org but is not surfaced in any UI or used for any scenario. Let's give an easy win to our package consumers!

## History

As far as I can tell, the repository metadata has the following history:

- 2015-07-15: Repository `type` is added to the project.json schema ([PR](https://github.com/SchemaStore/schemastore/pull/32))
- 2015-07-25: the `repository` node is added to project.json for `dnu sources` ([announcement](https://github.com/aspnet/Announcements/issues/46))
- 2015-09-14: `dnu sources` and the `repository` node are removed ([announcement](https://github.com/aspnet/Announcements/issues/66))
- 2016-01-17: Repository `url` is added to the project.json schema ([PR](https://github.com/SchemaStore/schemastore/pull/88))
- 2016-04-27: Repository URL and type are added to .NET CLI under `packOptions` ([PR](https://github.com/dotnet/cli/pull/2493))
- 2016-05-02: Repository fields' move to `packOptions` is announced ([accouncement](https://github.com/aspnet/Announcements/issues/175))
- 2016-08-12: NuGet adds `<repository>` to the .nuspec and to the .csproj ([commit](https://github.com/NuGet/NuGet.Client/commit/9b899f5542d51507520eb358bdc1e5c325b57576))
- 2016-09-22: .NET CLI migrate (.xproj to .csproj) adds support for `packOptions` ([PR](https://github.com/dotnet/cli/pull/4233))
- 2016-10-03: NuGet merged MSBuild new support to `dev` branch ([PR](https://github.com/NuGet/NuGet.Client/pull/907))
[...]

PR #6151, opened 2018-07-09
Add repository url to package page

natemcmaster:
Initial work towards https://github.com/NuGet/NuGetGallery/issues/4941.

I wanted to get your take on the initial work to show repository url from the `<repository>` metadata in a package. I imagine there is follow up work. Open questions;

 * Is the RepositoryUrl column in nuget.org's Packages table currently empty? It doesn't look like this was beening populated on upload.
 * What should we do with the `commit`, `branch`, and `type` attributes? We could introduce some some inference. For example, if `repository.url == https://github.com/natemcmaster/CommandLineUtils` and `repository.commit == 311673414d6530d97c70871a60c1a26f924f0d17`, should the packages page could show this link: https://github.com/natemcmaster/CommandLineUtils/tree/311673414d6530d97c70871a60c1a26f924f0d17

@diverdan92 @anangaur


natemcmaster:
To make it easier to review:

## Upload page
Verifying metadata:
image

## Package page

image

loic-sharma:
I verified, all packages on nuget.org have a null `RepositoryUrl`.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -564,6 +564,15 @@
                         </a>
                     </li>
                 }
+                @if (!Model.Deleted && PackageHelper.ShouldRenderUrl(Model.RepositoryUrl))
+                {
+                    <li>
+                        <i class="ms-Icon ms-Icon--Code" aria-hidden="true"></i>
+                        <a href="@Model.RepositoryUrl" data-track="outbound-repository-url" title="View the source code for this package." rel="nofollow">
```
loic-sharma: >. [](start = 136, length = 1)

We usually don't put periods in our title.

src/NuGetGallery.Core/Packaging/PackageMetadata.cs
```diff
@@ -228,13 +236,14 @@ public static PackageMetadata FromNuspecReader(NuspecReader nuspecReader, bool s
                 }
             }
 
+
             return new PackageMetadata(
```
loic-sharma: >return [](start = 12, length = 6)

There's an extra line above here

MaximRouiller:
Can this be fixed retroactively? There's tons of Nuspecs that already have that field defined.

natemcmaster:
@skofman1 what's the path to getting this approved to merge?

MaximRouiller:
Pinging this to keep this alive. Any timeframe to get this merged?

anangaur:
@natemcmaster , if you could do the following, it would make it look better:

1.	If the URL is a github URL, use the GitHub icon instead. 
2.	Else, if the type is git, use the git icon 
3.	Else fall back to the default icon (that’s being used now).

natemcmaster:
It looks like there isn't an icon for these in the current icon set. The gallery is currently only imports Office UI Fabric 7.0, and there is nothing in that set that looks like a good fit for 'git' or GitHub repo types.

If we upgraded to [Office UI Fabric 9](https://developer.microsoft.com/en-us/fabric#/styles/icons), you could use one of these icons, but still no GitHub in this set.

image

Also, it looks like you have font-awesome 3.2 in your source code. That has the [github icon](https://fontawesome.com/v3.2.1/icon/github) in it, but it doesn't look like the gallery pages import font-awesome, only the admin pages. 

So, how would you prefer I go about getting the right icons?

joelverhagen:
Looking at the release notes of Office UI Fabric, it seems like there is a lot of churn on icons. It seems reasonable to update to 9 but I am concerned about unexpected breaking changes. If we can assess that the update is safe, let's do it. Searching for all pages that use the Office UI Fabric CSS rules and manually testing them seems like the best approach.

For the GitHub icon, I think we should used their prescribed branding: https://github.com/logos. The "mark" zip has vector graphics. We should convert it to SVG and use that icon. Note that this is what we have for our Facebook and Twitter share icons.

The non-GitHub icons should be pulled from Fabric 9.

natemcmaster:
I think I'd rather skip the Fabric 9 update for now, but I've added the github icon.

screen shot 2018-07-19 at 8 26 50 pm

skofman1:
Looks good Nate! You can add a git logo in the same way: https://git-scm.com/downloads/logos
Also, please add UTs.

natemcmaster:
> Also, please add UTs.

Noobie question: what's that? My googling didn't help.

diverdan92:
My guess - unit tests?

natemcmaster:
Added the git icon and a few unit tests. Not sure what kind of other coverage you would expect from automated testing.

Here's what the packages page looks like now.

Example: `<repository url="https://github.com/aspnet/Mvc" />`
image

Example: `<repository url="https://bitbucket.com/aspnet/Mvc" type="git" />`

image

Example: `<repository url="https://bitbucket.com/aspnet/Mvc" type="tfs" />`

image

anangaur:
LGTM 👍🏽

MaximRouiller:
@anangaur I could do the "review required" but I'm a partial party to this PR. I'll let someone else do the final review before we can merge this. 😂

MaximRouiller:
What's left to do before we can merge this?

I see a review required, I wonder who should do this. @anangaur?
I see a conflicting file. @natemcmaster?

natemcmaster:
@MaximRouiller I chatted a bit with @joelverhagen, and from what I'm told, this may take a little bit longer to merge, but it's still on the radar.

src/NuGetGallery/ViewModels/PackageViewModel.cs
```diff
@@ -111,25 +115,52 @@ public PackageStatusSummary PackageStatusSummary
...
         }
+
+        private RepositoryKind GetRepositoryKind(string repositoryUrl, string repositoryType)
+        {
+            if (string.IsNullOrEmpty(repositoryUrl))
+            {
+                return RepositoryKind.Unknown;
+            }
+
+            if (repositoryUrl.IndexOf("github.com", StringComparison.OrdinalIgnoreCase) >= 0)
```
skofman1: A better check would be to parse the URL and make sure the domain is github.com

skofman1:
@natemcmaster , please apply the migrations again. We had other schema changes that were already merged and I'm getting this error: 

> Unable to update database to match the current model because there are pending changes and automatic migration is disabled. Either write the pending model changes to a code-based migration or enable automatic migration. Set DbMigrationsConfiguration.AutomaticMigrationsEnabled to true to enable automatic migration.
> You can use the Add-Migration command to write the pending model changes to a code-based migration.

natemcmaster:
Thanks @skofman1! Are there any other changes you want me to make?

skofman1:
@natemcmaster , this is the spec @anangaur wrote: https://github.com/NuGet/Home/wiki/Show-source-repository-information-for-packages-on-nuget.org

The current PR takes care of most of the requirements. There are warnings we would like to show when a user uploads an invalid URL, but we will take care of this in a different PR. 
For This PR: please improve the github URL check, and I will merge. If you can do this before tomorrow EOD we can release next week.

natemcmaster:
Thanks @anangaur and @skofman1!

I pushed a fix for the GitHub URL check. It does a proper System.Uri check now to ensure the domain is github.com or www.github.com.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -565,6 +565,31 @@
...
                     </li>
                 }
+                @if (!Model.Deleted && PackageHelper.ShouldRenderUrl(Model.RepositoryUrl, secureOnly: true))
+                {
+                    <li>
+                        @switch (Model.RepositoryType)
+                        {
+                            case PackageViewModel.RepositoryKind.GitHub:
+                                <img class="icon" alt="GitHub logo"
+                                     src="@Url.Absolute("~/Content/gallery/img/github.svg")"
```
loic-sharma: These `img` icons should also have the `aria-hidden="true"` attribute for accessibility.

src/NuGetGallery/ViewModels/PackageViewModel.cs
```diff
@@ -111,25 +115,56 @@ public PackageStatusSummary PackageStatusSummary
...
+
+        private RepositoryKind GetRepositoryKind(string repositoryUrl, string repositoryType)
+        {
+            if (string.IsNullOrEmpty(repositoryUrl))
+            {
+                return RepositoryKind.Unknown;
+            }
+
+            if (Uri.TryCreate(repositoryUrl, UriKind.Absolute, out var repoUri))
+            {
```
loic-sharma: This will return `RepositoryKind.Git` if the `repositoryUrl` is invalid but the `repositoryType` is `git`. Should `GetRepositoryKind` return `RepositoryKind.Unknown` instead? I see that `DisplayPackage.cshtml` doesn't use the `RepositoryType` if the `repositoryUrl` is invalid, but I think this could lead to bugs down the road.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
skofman1: fixed

natemcmaster:
Thanks @skofman1 and @loic-sharma !

PR #6154, opened 2018-07-10
Improve email sending

zivkan:
* Retry failed sends, to improve reliability
* Send App Insights telemetry, so we have visibility into the extent of the problem (plus the quantity of messages sent)
* NuGetGallery to send emails in the background, so slow/retried emails don't slow HTTP responses


src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -192,31 +196,52 @@ protected static void AddOwnersSubscribedToPackagePushedNotification(PackageRegi
             }
         }
 
-        protected void SendMessage(MailMessage mailMessage)
+        protected virtual void SendMessage(MailMessage mailMessage)
         {
-            SendMessage(mailMessage, copySender: false);
+            int attempt = 0;
+            for (;;)
```
shishirx34: we don't really use much infinite loops, but I think a couple of places where we do use `while (true)`, may be use the same for consistency?

scottbommarito: Not sure exactly what this means. I assume it means "keep looping infinitely".
I'd rather see `while (true)` than this. Even better, I'd prefer to see this loop rewritten with a clearer condition, e.g. `while (attempt < RetryDelays.Count)`

src/NuGetGallery/Services/MessageService.cs
```diff
@@ -1014,21 +1021,20 @@ private bool AddAddressesForAccountManagementToEmail(MailMessage mailMessage, Us
...
+                success = true;
             }
-            catch (SmtpException ex)
+            finally
             {
-                // Log but swallow the exception
-                QuietLog.LogHandledException(ex);
+                sw.Stop();
+                telemetryClient.TrackDependency("SMTP", smtpUri, "SendMessage", null, now, sw.Elapsed, null, success);
             }
```
loic-sharma: The pattern we use for this kind of thing is a `TrackMetric`, through some sort of descriptive wrapper method. For example: [https://github.com/NuGet/NuGet.Jobs/blob/master/src/Validation.PackageSigning.ProcessSignature/Telemetry/TelemetryService.cs#L61](https://github.com/NuGet/NuGet.Jobs/blob/master/src/Validation.PackageSigning.ProcessSignature/Telemetry/TelemetryService.cs#L61)

Also, this out of scope for this PR, but we really should merge the Gallery's and ServerCommon's telemetry APIs...

loic-sharma: Also, it may be worth tracking whether the operation was a success or not so that we measure how good/bad our mail system is.

---
In reply to: [201454393](https://github.com/NuGet/NuGetGallery/pull/6154#discussion_r201454393) [](ancestors = 201454393)

src/NuGetGallery.Core/Services/CoreMessageService.cs
zivkan: Yes, it's an infinite loop. You're right, `while (true)` is easier to read. I guess some habits die hard.

I'll change the loop to `while (!success)`. The issue with looping until the retry count is exceeded, is that if we want to rethrow the exception and keep the stack trace, we need to keep the exception in a variable, then outside the exception handler do `ExceptionDispatchInfo.Capture(exception).Throw()`. I think it's much cleaner to just `throw;` in the exception handler.

src/NuGetGallery/Services/MessageService.cs
zivkan: I wanted to use TrackDependency for a few reasons.
* I wanted to have an event which told us if it was successful or failed. Keeping the duration is just a side-effect of using the dependency event type
* AI's Smart Monitoring can automatically alert us when failed dependencies increase. Using metrics we would have to create custom alerts, assuming we use App Insights for alerting. Perhaps if we export App Insights logs to something else and do alerting there, using metrics doesn't lose this advantage.
* Dependencies are shown in AI's Application Map tab. If all dependencies are recorded in AI, then the application map is an auto-generated, self-updating, service architecture diagram, which is great for newbies like me 😄 (side note, we must be using an old App Insights SDK, because newer ones automatically capture outgoing HTTP calls, including Blob Storage, but ours does not)
* Sending emails to a SMTP server really is a service dependency, and not an application specific metric. Therefore, using the dependency event type, rather than metric, seems more semantically correct.

I'm happy to change it to a metric, but those are the reasons I added `TrackDependency` to `ITelemetryClient`.  Let me know what you think.

loic-sharma: Those all sound like very good points! It sounds like `TrackDependency` is indeed the right pattern here. I would still add a helper method in [`TelemetryService`](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Services/TelemetryService.cs)

src/NuGetGallery/Services/BackgroundMessageService.cs
```diff
@@ -0,0 +1,37 @@
...
+        public BackgroundMessageService(IMailSender mailSender, IAppConfiguration config, ITelemetryClient telemetryClient)
+            :base(mailSender, config, telemetryClient)
+        {
+        }
+
+        protected override void SendMessage(MailMessage mailMessage)
+        {
+            // Send email as background task, as we don't want to delay the HTTP response.
+            // Particularly when sending email fails and needs to be retried with a delay.
+            Task.Run(() =>
```
agr: I have a concerth that we might run into thread pool depletion if we try to send a lot of email while SMTP server is unresponsive and causes retries. `base.SendMessage` uses `Thread.Sleep` which blocks thread preventing other tasks from running on it (including processing the web requests? not sure if asp.net uses the same thread pool as `Task.Run` for request processing).

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -192,31 +196,53 @@ protected static void AddOwnersSubscribedToPackagePushedNotification(PackageRegi
...
+                {
+                    AttemptSendMessage(mailMessage);
+                    success = true;
+                }
+                catch (SmtpException)
+                {
+                    if (attempt < RetryDelays.Count)
+                    {
+                        Thread.Sleep(RetryDelays[attempt]);
+                        attempt++;
```
agr: We probably need to track the amount of retries we do.

loic-sharma:
Hey @zivkan, it looks like we should make the message service asynchronous to avoid blocking threads during retries. I believe that converting and then testing the code may take 1-2 days. After discussing it with @skofman1, we think it'd be best if I hand off this work back to you for when you return.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -57,13 +57,13 @@ protected override void SendNewAccountEmail(User account)
         {
             var confirmationUrl = Url.ConfirmOrganizationEmail(account.Username, account.EmailConfirmationToken, relativeUrl: false);
 
-            MessageService.SendNewAccountEmail(account, confirmationUrl);
+            MessageService.SendNewAccountEmailAsync(account, confirmationUrl);
         }
```
loic-sharma: Shouldn't you await this? Same with line 66.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -892,7 +892,7 @@ private void NotifyReportMyPackageSupportRequest(ReportMyPackageViewModel report
                 CopySender = reportForm.CopySender
             };
 
-            _messageService.ReportMyPackage(request);
+            _messageService.ReportMyPackageAsync(request);
```
loic-sharma: Is this missing an await?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -989,7 +989,7 @@ public virtual ActionResult ContactOwners(string id, string version, ContactOwne
 
             var user = GetCurrentUser();
             var fromAddress = new MailAddress(user.EmailAddress, user.Username);
-            _messageService.SendContactOwnersMessage(
+            _messageService.SendContactOwnersMessageAsync(
                 fromAddress,
```
loic-sharma: Is this missing an await?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -1065,7 +1065,7 @@ private ActionResult SendPasswordResetEmail(User user, bool forgotPassword)
                 user.PasswordResetToken,
                 forgotPassword,
                 relativeUrl: false);
-            MessageService.SendPasswordResetInstructions(user, resetPasswordUrl, forgotPassword);
+            MessageService.SendPasswordResetInstructionsAsync(user, resetPasswordUrl, forgotPassword);
```
loic-sharma: Should this be awaited?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -74,13 +74,13 @@ protected override void SendNewAccountEmail(User account)
         {
             var confirmationUrl = Url.ConfirmEmail(account.Username, account.EmailConfirmationToken, relativeUrl: false);
 
-            MessageService.SendNewAccountEmail(account, confirmationUrl);
+            MessageService.SendNewAccountEmailAsync(account, confirmationUrl);
```
loic-sharma: >SendNewAccountEmailAsync [](start = 27, length = 24)

Should this line and line 83 be awaited?

loic-sharma:
Did another pass, looks great! 👍

PR #6159, opened 2018-07-11
Use (short) version string in package version history

### LINKED ISSUE #5651: semver2 version's metadata should be displayed in a separate column
displaying the metadata with the version in the same column as version gets ugly when the metadata gets long. My suggestion is to have a separate column for version metadata if it is a semver2 version.

zivkan:
fixes #5651

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -1,3 +1,4 @@
+
 @using NuGet.Services.Validation;
 
 @model DisplayPackageViewModel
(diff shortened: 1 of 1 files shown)


agr:
How about some screenshots?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -469,9 +469,9 @@
                                         }
                                     </td>
                                     <td>
-                                        <a href="@Url.Package(packageVersion)" title="@packageVersion.FullVersion">
+                                        <a href="@Url.Package(packageVersion)" title="@packageVersion.Version">
```
joelverhagen: `title` is a mouse-over so we can be more verbose right?

zivkan: yes, you're right. I'll change it back.

zivkan:
Screenshot after changes. See #5651 for a similar screenshot without these changes.

image

Issue #5058, opened 2017-11-22
Disable the ID-title validation on package upload


joelverhagen:
Package title is much less prominent in the NuGet.org experience. It is not even visible in Visual Studio NuGet experience.

skofman1:
This gets in the way of customers pushing. Let's get this done.

PR #6163, opened 2018-07-13
Disable ID-Title validation for upload

shishirx34:
Fixes https://github.com/NuGet/NuGetGallery/issues/5058 

No more title conflicts with the ID during package uploads


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -456,10 +449,6 @@ private PackageRegistration CreateOrGetPackageRegistration(User owner, User curr
 
             if (packageRegistration == null)
             {
-                if (_packageNamingConflictValidator.IdConflictsWithExistingPackageTitle(packageMetadata.Id))
-                {
-                    throw new EntityException(Strings.NewRegistrationIdMatchesExistingPackageTitle, packageMetadata.Id);
```
loic-sharma: Did you remove this string?

shishirx34: Ah, good catch. removed it.

PR #6172, opened 2018-07-17
[Revalidation] Add Admin Panel

loic-sharma:
This change adds two features:

* A revalidation state service - This service allows parallel updates to the state. This is used by the Gallery, and later, the revalidation job for things like the killswitch and the desired revalidation rate.
* An admin UI to stop/start the revalidation job.

# Screenshots

## Default view

activate

# View if Revalidation job is killswitched

enable-revalidations

Addresses https://github.com/NuGet/Engineering/issues/1437


src/NuGetGallery/Filters/UiAuthorizeAttribute.cs
```diff
@@ -27,6 +27,7 @@ public override void OnAuthorization(AuthorizationContext filterContext)
                     new RouteValueDictionary(
                         new
                         {
+                            area = "",
```
loic-sharma: This is a bug fix as the `UiAuthorize` was broken on admin routes.

src/NuGetGallery.Core/Services/IRevalidationSettingsService.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// Attempt to update the latest settings.
+        /// </summary>
+        /// <param name="updateAction">The action used to update the settings.</param>
+        /// <returns>A task that completes once the settings have been updated.</returns>
+        Task UpdateSettingsAsync(Action<RevalidationSettings> updateAction);
+
+        /// <summary>
+        /// Attempt to update the latest settings.
+        /// </summary>
+        /// <param name="updateAction">The callback that updates the settings. Changes are only persisted if the callback returns ture</param>
```
chenriksson: nit: ture => true

src/NuGetGallery.Core/Services/IRevalidationSettingsService.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <returns>The updated settings.</returns>
+        Task<RevalidationSettings> MaybeUpdateSettingsAsync(Func<RevalidationSettings, bool> updateAction);
```
chenriksson: `Maybe` sounds weird. Feels like a strange variant of the try pattern... what about?
```
Task<bool> TryUpdateSettingsAsync(Func<RevalidateSettings, bool> tryUpdateAction, out RevalidationSettings);
```

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.WindowsAzure.Storage;
+using Newtonsoft.Json;
+
+namespace NuGetGallery
+{
+    public class RevalidationSettingsService : IRevalidationSettingsService
+    {
+        private const string SettingsFileName = "settings.json";
```
chenriksson: We're doing per job killswitch until the need for it grows? This validator uses `/revalidation/settings.json`?

chenriksson: Revalidation already has a dependency on Gallery.Core? Just wondering why we define this here, instead of ServerCommon?

src/NuGetGallery/Areas/Admin/Views/Revalidation/Index.cshtml
```diff
@@ -0,0 +1,75 @@
...
+                <strong>Started Revalidations In Last Hour:</strong>
+                @Model.Statistics.StartedRevalidationsInLastHour.ToString("n0")
+            </p>
+            <p>
+                <strong>Remaining Revalidations:</strong>
+                @Model.Statistics.PendingRevalidations.ToString("n0")
+            </p>
+            <p>
+                <strong>Desired Package Event Rate:</strong>
+                @Model.Settings.DesiredPackageEventRate events per hour
```
chenriksson: Setting to adjust rate at which we revalidate? Is there more guidance we could show on admin page for on-call engineer, as to how to adjust this?

Is there a monitor to see the actual package event rate, that on-call can refer to?

src/NuGetGallery/Areas/Admin/Views/Revalidation/Index.cshtml
```diff
@@ -0,0 +1,75 @@
...
+                            }
+                        </div>
+                    }
+                </text>
+            </div>
+
+            <h2>Statistics</h2>
+            <p>
+                <strong>Initialized:</strong>
+                @if (Model.Settings.Initialized)
```
chenriksson: What does `Initialized` mean? Is it indication for when the killswitch change has propagated?

Should admin page specify, for on-call engineer?

tests/NuGetGallery.Core.Facts/Services/RevalidationSettingsServiceFacts.cs
```diff
@@ -0,0 +1,311 @@
...
+            protected RevalidationSettings Deserialize(string input)
+            {
+                using (var reader = new StreamReader(input))
+                using (var jsonReader = new JsonTextReader(reader))
+                {
+                    return new JsonSerializer().Deserialize<RevalidationSettings>(jsonReader);
+                }
+            }
+
+            private class TestableFileReference : IFileReference, IDisposable
```
chenriksson: nit: could be useful helper class to extract, but could do when we need it

src/NuGetGallery/Areas/Admin/Views/Revalidation/Index.cshtml
```diff
@@ -0,0 +1,75 @@
...
+
+            <h2>Killswitch</h2>
+            <div class="form-group danger-zone">
+                <text>
+                    @using (Html.BeginForm("ActivateKillswitch", "Revalidation", FormMethod.Post))
+                    {
+                        @Html.AntiForgeryToken()
+
+                        <p>
+                            Use the killswitch to disable the revalidation job if it causes issues ingestion pipeline.
```
dtivel: ...if it causes ~issues~ ingestion pipeline **issues**.

src/NuGetGallery.Core/Services/RevalidationSettings.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// The settings for the revalidation job.
+    /// </summary>
+    public class RevalidationSettings
+    {
+        /// <summary>
+        /// Whether the revalidation job has been disactivated.
+        /// </summary>
+        public bool Killswitch { get; set; } = false;
```
dtivel: Would prefer `Is-` for boolean properties.  Perhaps `IsKillswitchActive`?

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
loic-sharma: Yes, @skofman1 felt that an HTTP killswitch would complicate the system too much. As each NuGet repository has its own storage abstractions today, we can't have a universal blob storage killswitch system. Given these restrictions, each job will need to create its own killswitch system.

ServerCommon and Gallery have different storage abstractions. As the settings blob needs to be updated by the Gallery, it made the most sense to use the Gallery's storage APIs. Also, the Orchestrator uses Gallery.Core's storage APIs instead of ServerCommon's.

The state of our storage abstractions across our projects is a complete mess. Consolidating these abstractions will be a non-trivial effort, so I'm not doing that as part of this change.

src/NuGetGallery.Core/Services/RevalidationSettings.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        /// <summary>
+        /// Whether the revalidation job's state has been initialized.
+        /// </summary>
+        public bool Initialized { get; set; } = false;
```
dtivel: It seems like an odd place to store job state:  in a settings object.  Plus, it feels like this should be publicly read-only, maybe mutable only by the job itself.

That said, how about `IsInitialized`?

src/NuGetGallery.Core/Services/RevalidationSettings.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// The settings for the revalidation job.
+    /// </summary>
+    public class RevalidationSettings
+    {
+        /// <summary>
+        /// Whether the revalidation job has been disactivated.
```
dtivel: deactivated

src/NuGetGallery.Core/Extensions/StorageExceptionExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Net;
+using Microsoft.WindowsAzure.Storage;
+
+namespace NuGetGallery
+{
+    public static class StorageExceptionExtensions
+    {
+        public static bool IsFileAlreadyExistsException(this StorageException e)
+        {
+            return e.RequestInformation?.HttpStatusCode == (int?)HttpStatusCode.Conflict;
```
dtivel: `e` can be `null`.  Maybe `e?`?

src/NuGetGallery/Areas/Admin/Views/Revalidation/Index.cshtml
loic-sharma: > Setting to adjust rate at which we revalidate? Is there more guidance we could show on admin page for on-call engineer, as to how to adjust this?

This is not part of the spec. The job gets to choose its revalidation rate based off its configurations (see [`MinPackageEventRate`](https://github.com/NuGet/NuGet.Jobs/pull/481/files#diff-0d89ded38e53506639b67d3d2c49ea90R15) and [`MaxPackageEventRate`](https://github.com/NuGet/NuGet.Jobs/pull/481/files#diff-0d89ded38e53506639b67d3d2c49ea90R21)). If the job is causing issues in the ingestion pipeline, the on-call engineer should killswitch the job. If the job is too slow or too fast, the job should be redeployed with new configurations.

> Is there a monitor to see the actual package event rate, that on-call can refer to?

Yes, that already exists, see the "Package Events" dashboard on Geneva: https://jarvis-west.dc.ad.msft.net/dashboard/share/3B2996DB

src/NuGetGallery/Areas/Admin/Controllers/RevalidationController.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            var settings = await _settings.GetSettingsAsync();
+            var statistics = _revalidationAdminService.GetStatistics();
+
+            return View(nameof(Index), new RevalidationPageViewModel(settings, statistics));
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<ActionResult> ActivateKillswitch(bool killswitch)
```
dtivel: `SetKillswitch`?  If `killswitch` is `false` then `ActivateKillswitch(...)` would be deactivating the killswitch.

src/NuGetGallery/Areas/Admin/Views/Revalidation/Index.cshtml
loic-sharma: See [the spec on OneNote](https://microsoft.sharepoint.com/teams/NuGet/_layouts/OneNote.aspx?id=%2Fteams%2FNuGet%2FTeam%2FNugetServer%2FNugetServerTeamNote&wd=target%28Specs.one%7C910181C5-8695-4A1C-9792-A56DBB86F1C3%2FInitialization%7CEFB8826E-0AEC-403F-8483-E4103FE59DC7%2F%29onenote:https://microsoft.sharepoint.com/teams/NuGet/Team/NugetServer/NugetServerTeamNote/Specs.one#Initialization&section-id={910181C5-8695-4A1C-9792-A56DBB86F1C3}&page-id={EFB8826E-0AEC-403F-8483-E4103FE59DC7}&end).

The job has two phases:

1. It initializes itself
2. It enqueues packages revalidations

It cannot do step 2 until step 1 is done. The `Initialized` setting is how the job determines that step 1 is complete.

src/NuGetGallery.Core/Services/IRevalidationSettingsService.cs
loic-sharma: This isn't the try pattern; `MaybeUpdateSettingsAsync` only updates the settings if the inputted `tryUpdateAction` returns true. If the `tryUpdateAction` returns false, the settings are not updated. Does that make sense? Let me know if you have suggestions for a better name.

tests/NuGetGallery.Core.Facts/Services/RevalidationSettingsServiceFacts.cs
loic-sharma: Agreed

src/NuGetGallery.Core/Services/RevalidationSettings.cs
loic-sharma: The revalidation job's settings can be mutated by the Gallery and the job (Gallery mutates the killswitch, job mutates `IsInitialized` and `DesiredPackageEventRate`). As this type is reused by both services, all of these properties have to be publicly settable.

src/NuGetGallery.Core/Services/IRevalidationSettingsService.cs
loic-sharma: Fixed

src/NuGetGallery/Areas/Admin/Views/Revalidation/Index.cshtml
loic-sharma: Fixed

src/NuGetGallery.Core/Services/RevalidationSettings.cs
loic-sharma: Fixed

src/NuGetGallery.Core/Services/RevalidationSettings.cs
loic-sharma: Fixed

src/NuGetGallery.Core/Extensions/StorageExceptionExtensions.cs
loic-sharma: Fixed

src/NuGetGallery/Areas/Admin/Controllers/RevalidationController.cs
loic-sharma: Good idea. Fixed

src/NuGetGallery.Core/Services/CloudBlobWrapper.cs
```diff
@@ -71,22 +71,27 @@ public Task DownloadToStreamAsync(Stream target)
...
+                await UploadFromStreamAsync(source, AccessCondition.GenerateIfNoneMatchCondition("*"));
             }
         }
 
+        public async Task UploadFromStreamAsync(Stream source, AccessCondition accessCondition)
+        {
+            await _blob.UploadFromStreamAsync(
+                source,
+                accessCondition,
+                new BlobRequestOptions(),
```
joelverhagen: can't you pass null instead of this default `new`? This gives more control to the Azure SDK since we don't care

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
```diff
@@ -0,0 +1,98 @@
...
+
+                return settings;
+            }
+            catch (StorageException e) when (e.IsPreconditionFailedException())
+            {
+                throw new InvalidOperationException("Failed to update the settings blob as the access condition failed", e);
+            }
+        }
+
+        private async Task<Tuple<IFileReference, RevalidationSettings>> GetSettingsInternalAsync()
```
joelverhagen: please no tuple, a private class is more readable

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
```diff
@@ -0,0 +1,98 @@
...
+        public async Task UpdateSettingsAsync(Action<RevalidationSettings> updateAction)
+        {
+            await MaybeUpdateSettingsAsync(s =>
+            {
+                updateAction(s);
+                return true;
+            });
+        }
+
+        public async Task<RevalidationSettings> MaybeUpdateSettingsAsync(Func<RevalidationSettings, bool> updateAction)
```
joelverhagen: Why is `MaybeUpdateSettingsAsync` public?

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
```diff
@@ -0,0 +1,98 @@
...
+                    stream.Position = 0;
+
+                    await _storage.SaveFileAsync(CoreConstants.RevalidationFolderName, SettingsFileName, stream, accessCondition);
+                }
+
+                return settings;
+            }
+            catch (StorageException e) when (e.IsPreconditionFailedException())
+            {
+                throw new InvalidOperationException("Failed to update the settings blob as the access condition failed", e);
```
joelverhagen: `since` reads clears than `as`

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        public async Task<RevalidationSettings> MaybeUpdateSettingsAsync(Func<RevalidationSettings, bool> updateAction)
+        {
+            var getResult = await GetSettingsInternalAsync();
+            var fileReference = getResult.Item1;
+            var settings = getResult.Item2;
+
+            // Only update the settings if the update action returns true.
+            if (!updateAction(settings))
```
joelverhagen: This is a bit weird, since the modified instance can be returned to the caller even if it does not get persisted... this can lead to UI showing one thing when coming back from a POST but when you refresh the page you see something else (for example). Not terrible, just could be confusing dependong on how the callback is implemented.

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
joelverhagen: Okay, sounds good.

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
```diff
@@ -0,0 +1,98 @@
...
+                    _serializer.Serialize(jsonWriter, settings);
+                    jsonWriter.Flush();
+                    stream.Position = 0;
+
+                    await _storage.SaveFileAsync(CoreConstants.RevalidationFolderName, SettingsFileName, stream, accessCondition);
+                }
+
+                return settings;
+            }
+            catch (StorageException e) when (e.IsPreconditionFailedException())
```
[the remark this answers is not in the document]
loic-sharma: Nothing catches this and you'll get the normal "Internal Error" page. I think/hope that the likelihood of this happening is low enough that that is okay considering this isn't internal facing.

I agree completely that we should handle the exceptions should this ever become customer facing.

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
loic-sharma: Yeah... I wish we had named tuples, then it wouldn't be so bad. I'll use a private class.

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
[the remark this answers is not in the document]
joelverhagen: I don't think we should knowingly add HTTP 500s since it is noise in our AI. We could easily catch this exception type in the controller and pass the exception message to the page (or something).

src/NuGetGallery.Core/Services/CloudBlobWrapper.cs
loic-sharma: I believe so. I'll test this.

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
```diff
@@ -0,0 +1,98 @@
...
+                return settings;
+            }
+
+            try
+            {
+                var accessCondition = AccessConditionWrapper.GenerateIfMatchCondition(fileReference.ContentId);
+
+                using (var stream = new MemoryStream())
+                using (var writer = new StreamWriter(stream))
+                using (var jsonWriter = new JsonTextWriter(writer))
```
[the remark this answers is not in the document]
loic-sharma: Talked to @joelverhagen offline. The reader path is necessary as `IFileReference` returns a `Stream`, not a `MemoryStream`. @joelverhagen recommended keeping the reading/writing the same.

loic-sharma:
@joelverhagen @dtivel I've addressed all of y'all's feedback, please take another look 😄

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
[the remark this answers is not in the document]
loic-sharma: Sure I can do that

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
loic-sharma: Good catch. I think it'd be easy to fall into this trap and do something really weird. I've fixed this behavior and updated tests.

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
loic-sharma: Fixed

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
loic-sharma: Fixed

src/NuGetGallery.Core/Services/CloudBlobWrapper.cs
loic-sharma: Fixed

src/NuGetGallery.Core/Services/RevalidationSettingsService.cs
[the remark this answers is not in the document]
loic-sharma: Fixed

src/NuGetGallery.Core/Services/IRevalidationSettingsService.cs
chenriksson: I get that it's not the try pattern. My reaction is because `Maybe` sounds noncommittal which feels like it weakens the API contract. Feel free to ignore. :-)

src/NuGetGallery/Areas/Admin/Views/Revalidation/Index.cshtml
chenriksson: So `DesiredPackageEventRate` is configured guidance for when we should killswitch / redeploy? Just making sure the data / settings on the page are easily actionable.

src/NuGetGallery.Core/Services/IRevalidationSettingsService.cs
loic-sharma: Gotcha. I don't think this method fits the try pattern as it throws if there are any issues. Normally, the try pattern reports errors through the returned value. Really, the method name should be something along the lines of `UpdateStateIfUpdateActionReturnsTrue(Func<RevalidationState, bool> updateAction)`... but that's hideous so I went with `MaybeUpdateState` instead.

src/NuGetGallery/Areas/Admin/Views/Revalidation/Index.cshtml
chenriksson: Maybe the admin page could include a brief description, for on-call guidance? If so, maybe it would be useful for the settings to include the min/max values for guidance text?
```
Desired Package Event Rate:
500 package events / hour
* Package ingestion rate which should be achievable without pipeline degradation, as determined by the revalidation job. The current range is 100-500 package events.
```

loic-sharma: Good suggestion, done!

Issue #6173, opened 2018-07-17
Fix the text indentation on Upload package page


anangaur:
![image](https://user-images.githubusercontent.com/14800916/42835842-b0153ed4-89ae-11e8-94a6-aa3440dc87ca.png)

PR #6174, opened 2018-07-17
Update UploadPackage.cshtml

anangaur:
Fix for #6173


loic-sharma:
Could you add a picture of what this looks like?

Issue #6175, opened 2018-07-17
Package uploads broken when using file storage


JustinMoss:
It appears that the commit 8b4c3eab moved the upload package deletion inside the using for the uploaded file, which causes an error.  The line in question is: https://github.com/NuGet/NuGetGallery/blob/941861ab5b0adc7f6034e0672cce3b296605b968/src/NuGetGallery/Controllers/PackagesController.cs#L1695

On the upload page, the error is generic: "There was an unexpected error. Please try uploading the package again."

But digging into the admin errors, I find: "The process cannot access the file 'C:\Code\GitHub\Nuget\NuGetGallery\src\NuGetGallery\App_Data\Files\uploads\1.nupkg' because it is being used by another process."

This makes sense since the file is still open by that using statement.

The package still actually uploads, but the failure causes a weird state where the system thinks the user has an upload in progress, but it's already uploaded.  This can be worked around by simply canceling the upload after the error.

PR #6176, opened 2018-07-17
Move delete outside of using so FileStorage can delete the file.

ryuyu:
There was a bug introduced by #6038 that caused FileStorage uploads to behave incorrectly.
The behaviour is detailed by #6175 

This PR moves the delete outside of the Using statement and swallows if it fails.
Behaviour can change from swallow if we decide this is not how we want to handle this.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1691,18 +1691,10 @@ namespace NuGetGallery
                                 Url.AccountSettings(relativeUrl: false));
                         }
 
-                        // delete the uploaded binary in the Uploads container
-                        await _uploadFileService.DeleteUploadFileAsync(currentUser.Key);
-
                         _telemetryService.TrackPackagePushEvent(package, currentUser, User.Identity);
 
                         TempData["Message"] = String.Format(
                             CultureInfo.CurrentCulture, Strings.SuccessfullyUploadedPackage, package.PackageRegistration.Id, package.Version);
-
-                        return Json(new
-                        {
-                            location = Url.Package(package.PackageRegistration.Id, package.NormalizedVersion)
-                        });
                     }
                     catch (Exception e)
                     {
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1691,25 +1691,36 @@ private void SendAddPackageOwnerNotification(PackageRegistration package, User n
...
+                try
+                {
+                    // delete the uploaded binary in the Uploads container
+                    await _uploadFileService.DeleteUploadFileAsync(currentUser.Key);
+                }
+                catch(Exception e)
+                {
+                    // Log the exception here and swallow it for now.
+                    // We want to know the delete has failed, but the user shouldn't get a failed request here since everything has actually gone through
+                    // Note that this will still lead to the strange behavioru where the next time a user comes to the upload page, an upload will be "in progress"
```
loic-sharma: >behavioru [](start = 69, length = 9)

Behavior

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1691,25 +1691,36 @@ private void SendAddPackageOwnerNotification(PackageRegistration package, User n
...
                         return Json(HttpStatusCode.BadRequest, new[] { Strings.VerifyPackage_UnexpectedError });
                     }
                 }
+
+                try
+                {
+                    // delete the uploaded binary in the Uploads container
+                    await _uploadFileService.DeleteUploadFileAsync(currentUser.Key);
+                }
+                catch(Exception e)
```
loic-sharma: >( [](start = 21, length = 1)

Space is missing after the `catch`

loic-sharma:
This bit still needs to be addressed:

> We should change the upload page such that it doesn't show an upload as "in progress" if the package is already available. I would either fix that as part of this change, or, create an issue on GitHub and leave a TODO in this comment that links to the issue.

PR #6177, opened 2018-07-17
[Symbol Server]Add DB Schema changes

shishirx34:


src/NuGetGallery.Core/Entities/Symbol.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using System;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace NuGetGallery
+{
+    public class Symbol
+        : IEntity
+    {
+        public Symbol()
```
chenriksson: Can remove?

src/NuGetGallery.Core/Entities/Symbol.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+        }
+
+        public int Key { get; set; }
+
+        public Package Package { get; set; }
+
+        public int PackageKey { get; set; }
```
chenriksson: Package row has to exist before the symbols can be pushed?

src/NuGetGallery.Core/Entities/Symbol.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// Availability status of the symbol
+        /// </summary>
+        public PackageStatus StatusKey { get; set; }
+
+        /// <summary>
+        /// Used for optimistic concurrency when updating the symbols.
+        /// </summary>
+        public byte[] RowVersion { get; set; }
+    }
```
chenriksson: Do we need a column for FileSize, like we have on the Packages table?

src/NuGetGallery.Core/Entities/Symbol.cs
shishirx34: Yes.

src/NuGetGallery.Core/Entities/Symbol.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace NuGetGallery
+{
+    public class Symbol
```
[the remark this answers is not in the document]
shishirx34: Yes this is the snupkg table, I feel like `SymbolPackage` is a bit redundant. `Symbols` table seems right to me. If no strong objections, I would leave it as is.

src/NuGetGallery.Core/Entities/Symbol.cs
shishirx34: sure.

src/NuGetGallery.Core/Entities/Symbol.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Availability status of the symbol
+        /// </summary>
+        public PackageStatus StatusKey { get; set; }
+
+        /// <summary>
+        /// Used for optimistic concurrency when updating the symbols.
+        /// </summary>
+        public byte[] RowVersion { get; set; }
```
joelverhagen: Could you explain how this will be used? I like the idea but I am concerned that it would result in 500s in the gallery. Will the gallery be modifying this row? What actors will be modifying this row in parallel?

Contrast this to package which does not have a row version. Again, I'm not against it I just want to understand more about the expected concurrency here and how optimistic concurrency exceptions will be handled in the front-end.

src/NuGetGallery.Core/Entities/Symbol.cs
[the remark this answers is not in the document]
joelverhagen: Last time I heard, there could be multiple symbol files in a single package so symbol to symbol package is really many to one. This record is for the symbol package concept not each individual symbol.

src/NuGetGallery.Core/Entities/Symbol.cs
joelverhagen: If there is a possibility for an optimistic concurrency exception in the gallery, we should have a plan around it. A 500 in this case is a bad experience.

src/NuGetGallery.Core/Entities/Symbol.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <summary>
+        /// Timestamp when this symbol was created.
+        /// </summary>
+        [DatabaseGenerated(DatabaseGeneratedOption.Computed)]
+        public DateTime Created { get; set; }
+
+        /// <summary>
+        /// Time when this symbol package is available for consumption. It will be updated after validations are complete.
+        /// </summary>
+        public DateTime? Published { get; set; }
```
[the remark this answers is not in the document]
joelverhagen: This is closer to `LastEdited` then. That is the column that is updated on `Package` when it moves to the available state.

src/NuGetGallery.Core/Entities/Symbol.cs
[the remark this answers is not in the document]
cristinamanum: SymbolPackage would be probably indicate better the grouping of multiple symbols in one package.

src/NuGetGallery.Core/Entities/Symbol.cs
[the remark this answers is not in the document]
cristinamanum: It would be good to stay as close as possible to the packages names.

src/NuGetGallery.Core/Entities/Symbol.cs
chenriksson: ok, wasn't sure if symbols pushed before source package was still a thing :-)

src/NuGetGallery.Core/Entities/Symbol.cs
chenriksson: For optimistic concurrency to work, it should be used *always* when creating/updating the Symbol row... but EF should do that for us. It doesn't work well on the Package table because of our `IsLatest` calculations... e.g., updating one package row may require updates to all rows of the same package. It should work for the Symbol table though, since requests should only touch a single row.

src/NuGetGallery.Core/Entities/Symbol.cs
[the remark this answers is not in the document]
shishirx34: Yes it is closer to `LastEdited` in the behavior perspective, but, I do not think `LastEdited` is the right name here. In packages we had `Edit` scenario. Here its the time when the package is actually "published", i.e. becomes available for downloading. I do not think staying close as possible to the packages names is a good reason to continue the misnomer.

src/NuGetGallery.Core/Entities/Symbol.cs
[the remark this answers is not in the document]
shishirx34: Ok, will make this `SymbolPackage`

shishirx34:
@cristinamanum @chenriksson @joelverhagen - addressed feedback.

PR #6190, opened 2018-07-19
[NuGet Symbol Server] Add core symbol package service

shishirx34:
Adds the entity and definitions required by gallery, jobs and orchestrator. Also added tests.

### DIFF
--- a/src/NuGetGallery.Core/Services/CoreSymbolPackageService.cs
+++ b/src/NuGetGallery.Core/Services/CoreSymbolPackageService.cs
@@ -1,0 +1,66 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using NuGetGallery.Packaging;
+
+namespace NuGetGallery
+{
+    public class CoreSymbolPackageService : ICoreSymbolPackageService
+    {
+        protected readonly IEntityRepository<SymbolPackage> _symbolPackageRepository;
+        protected readonly ICorePackageService _corePackageService;
+
+        public CoreSymbolPackageService(
+            IEntityRepository<SymbolPackage> symbolPackageRepository,
+            ICorePackageService corePackageService)
+        {
+            _symbolPackageRepository = symbolPackageRepository ?? throw new ArgumentNullException(nameof(symbolPackageRepository));
+            _corePackageService = corePackageService ?? throw new ArgumentNullException(nameof(corePackageService));
+        }
+
+        public IEnumerable<SymbolPackage> FindSymbolPackageByIdAndVersion(string id, string version)
+        {
+            var package = _corePackageService.FindPackageByIdAndVersionStrict(id, version);
+
+            return package?.SymbolPackages;
+        }
+
+        public virtual async Task UpdateStatusAsync(SymbolPackage symbolPackage, PackageStatus newPackageStatus, bool commitChanges = true)
+        {
+            if (symbolPackage == null)
+            {
+                throw new ArgumentNullException(nameof(symbolPackage));
+            }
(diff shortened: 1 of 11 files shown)


src/NuGetGallery.Core/Services/ICoreSymbolPackageService.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// Symbol Package business logic that is needed in multiple components (not just the Gallery process).
+    /// </summary>
+    public interface ICoreSymbolPackageService
+    {
+        /// <summary>
+        /// Gets all the symbol packages associated with the Package ID and version
+        /// </summary>
+        /// <param name="id">The package ID.</param>
+        /// <param name="version">The package version.</param>
+        /// <returns></returns>
```
cristinamanum: Returns description.

src/NuGetGallery.Core/Services/CoreSymbolPackageService.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            _symbolPackageRepository = symbolPackageRepository ?? throw new ArgumentNullException(nameof(symbolPackageRepository));
+            _corePackageService = corePackageService ?? throw new ArgumentNullException(nameof(corePackageService));
+        }
+
+        public IEnumerable<SymbolPackage> FindSymbolPackagesByIdAndVersion(string id, string version)
+        {
+            var package = _corePackageService.FindPackageByIdAndVersionStrict(id, version);
+
+            return package?.SymbolPackages;
```
joelverhagen: Do we have an `Include(p => p.SymbolsPackages)` inside `FindPackageByIdAndVersionStrict` so that this relationship is populated? Or is this supposed to be a lazy load?

src/NuGetGallery.Core/Services/CoreSymbolPackageService.cs
```diff
@@ -0,0 +1,56 @@
...
+            // Avoid all of this work if the package status is not changing.
+            if (symbolPackage.StatusKey != newPackageStatus)
+            {
+                symbolPackage.StatusKey = newPackageStatus;
+
+                /// If the package is being made available.
+                if (newPackageStatus == PackageStatus.Available)
+                {
+                    symbolPackage.Published = DateTime.UtcNow;
+                }
```
joelverhagen: If we move to the `FailedValidation` state do you want to "unset" `Published`? What about the other state transitions?

src/NuGetGallery.Core/Services/CoreSymbolPackageService.cs
```diff
@@ -0,0 +1,56 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using NuGetGallery.Packaging;
+
+namespace NuGetGallery
+{
+    public class CoreSymbolPackageService : ICoreSymbolPackageService
+    {
+        protected readonly IEntityRepository<SymbolPackage> _symbolPackageRepository;
```
joelverhagen: Why not private?

src/NuGetGallery.Core/Services/CoreSymbolPackageService.cs
shishirx34: Ah, good catch. Missed adding it.

src/NuGetGallery.Core/Services/CoreSymbolPackageService.cs
shishirx34: No, we will only set the published, when transition is to "Available" state. For all other transitions we do not set this value. `FailedValidation` would mean it is not published.

src/NuGetGallery.Core/Services/CoreSymbolPackageService.cs
shishirx34: can be private, I copied the code over from `CorePackageService`. Will change.

Issue #6181, opened 2018-07-18
Provide faster feedback to package author when uploaded package does not match signature requirements


joelverhagen:
@joelverhagen commented on [Tue Jul 17 2018](https://github.com/NuGet/Engineering/issues/1538)

When a package must be signed, we can quickly reject an unsigned version by checking for the presence of a signature file.

When a package must not be signed, we can quickly reject a signed version by checking for the presence of a signature file.

Both operations are safe to be done in gallery process.

---

@joelverhagen commented on [Tue Jul 17 2018](https://github.com/NuGet/Engineering/issues/1538#issuecomment-405776672)

I need text for both cases and for both the upload flow and the command line push flow.

/cc @anangaur

joelverhagen:
Proposed strings:

When publishing an unsigned package and signing is required:

`
The uploaded package does not have a signature. A signature is required.
`

When publishing a signed package and no certificate is configured.

`
The uploaded package has a signature. No certificate is configured on the owner's account so only unsigned packages can be uploaded.
`

@anangaur?

PR #6199, opened 2018-07-23
Add IsSigningAllowed method

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6181.

The fast check (not in this PR) will reject the package if any of these conditions are true:

- Signing is required on the ID and there is no signature file
   - signing is required if all signers have at least one certificate
- Signing is not allowed on the ID and there is a signature file
   - signing is not allowed if all signers have no certificates

It is possible for signing to be allowed but not required if there are two owners on a package: one with certificates and one without certificates. In this case we can't do a fast check based off of the presence of a signature file.

I chose to name the method `IsSigningAllowed` instead of `IsSigningNotAllowed` so the call sign can negate the behavior without a double negative. The usage I will be adding will be `!IsSigningAllowed()`.

### DIFF
--- a/src/NuGetGallery.Core/Extensions/PackageRegistrationExtensions.cs
+++ b/src/NuGetGallery.Core/Extensions/PackageRegistrationExtensions.cs
@@ -9,6 +9,30 @@ namespace NuGetGallery.Extensions
     public static class PackageRegistrationExtensions
     {
         /// <summary>
+        /// Determines if package signing is allowed for the specified package registration.
+        /// </summary>
+        /// <param name="packageRegistration">A package registration.</param>
+        /// <returns>A flag indicating whether package signing is allowed.</returns>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="packageRegistration" />
+        /// is <c>null</c>.</exception>
+        public static bool IsSigningAllowed(this PackageRegistration packageRegistration)
+        {
+            if (packageRegistration == null)
+            {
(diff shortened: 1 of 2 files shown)


src/NuGetGallery.Core/Extensions/PackageRegistrationExtensions.cs
```diff
@@ -8,6 +8,30 @@ namespace NuGetGallery.Extensions
...
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="packageRegistration" />
+        /// is <c>null</c>.</exception>
+        public static bool IsSigningAllowed(this PackageRegistration packageRegistration)
+        {
+            if (packageRegistration == null)
+            {
+                throw new ArgumentNullException(nameof(packageRegistration));
+            }
+
+            var requiredSigner = packageRegistration.RequiredSigners.FirstOrDefault();
```
loic-sharma: What happens if a registration has two required signers, the first of which doesn't have any certificates registered?

loic-sharma: If this is a possible scenario, I would consider adding a test.

loic-sharma: Talked to @joelverhagen offline. I misunderstood `RequiredSigners`, please disregard these comments.

Issue #6169, opened 2018-07-16
[NuGet Symbol Server] Feature flag for symbol server


cristinamanum:
Add a feature flag for Symbol feature to be enabled only for MVPs.

PR #6207, opened 2018-07-24
Add Symbols Feature flag

shishirx34:
Addresses https://github.com/NuGet/NuGetGallery/issues/6169 

Also, I have reduced the reload duration to 5 minutes for configuration, it was 1 hour. This feature flag will be used in subsequent PRs to allow/prevent symbols upload until we go live with this feature.

### DIFF
--- a/src/NuGetGallery/Services/ContentObjectService.cs
+++ b/src/NuGetGallery/Services/ContentObjectService.cs
@@ -10,7 +10,7 @@ namespace NuGetGallery
 {
     public class ContentObjectService : IContentObjectService
     {
-        private const int RefreshIntervalHours = 1;
+        private const int RefreshIntervalMinutes = 5;
 
         private readonly IContentService _contentService;
 
@@ -20,10 +20,12 @@ namespace NuGetGallery
 
             LoginDiscontinuationConfiguration = new LoginDiscontinuationConfiguration();
             CertificatesConfiguration = new CertificatesConfiguration();
+            SymbolsConfiguration = new SymbolsConfiguration();
         }
 
         public ILoginDiscontinuationConfiguration LoginDiscontinuationConfiguration { get; set; }
         public ICertificatesConfiguration CertificatesConfiguration { get; set; }
+        public ISymbolsConfiguration SymbolsConfiguration { get; set; }
 
         public async Task Refresh()
         {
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/Services/ContentObjectService.cs
```diff
@@ -10,7 +10,7 @@ namespace NuGetGallery
 {
     public class ContentObjectService : IContentObjectService
     {
-        private const int RefreshIntervalHours = 1;
+        private const int RefreshIntervalMinutes = 5;
```
cristinamanum: Why changing this ? What it would be the impact?

src/NuGetGallery/Services/SymbolsConfiguration.cs
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json;
+
+namespace NuGetGallery.Services
+{
+    public sealed class SymbolsConfiguration : ISymbolsConfiguration
```
scottbommarito: This is basically a clone of `CertificatesConfiguration`, right?

IMO we should write a base class that both can use so we don't have to keep rewriting this.
Bonus points if the base class is resusable by `ILoginDiscontinuationConfiguration` as well.
I don't think it would take very long.

src/NuGetGallery/Services/ContentObjectService.cs
scottbommarito: These content configuration objects will refresh more often.
If we change configuration, it will take less time for us to see it change.
We've determined that 1 hour has been too long, and we would like to increase the frequency to every 5 minutes.

shishirx34: An hour was way too long for any config change to take effect. We've been wanting to reduce this for a while. I consulted with @scottbommarito and decided 5 mins seems appropriate. There shouldn't be any perf impact by this change.

src/NuGetGallery/Services/SymbolsConfiguration.cs
shishirx34: Yeah I agree, however, we are going to get rid of `CertificatesConfiguration` and I would rather not spend time on not strictly required refactoring right now. I have opened an [issue](https://github.com/NuGet/NuGetGallery/issues/6211) to track this.

Issue #6229, opened 2018-07-30
Gallery does not log search service exceptions encountered during retry


joelverhagen:
In `RetryHttpClientWrapper`, there is logic to try a sequence of HTTP endpoints until the list is exhausted or an endpoint succeeds. Suppose endpoint A fails then endpoint B succeeds. We do not log the failure encountered on endpoint A which makes debugging hard.

We should log all swallowed exceptions, somewhere around here:

https://github.com/NuGet/NuGetGallery/blob/cc1d16b135fdd553d99225e123f36aac79847e4f/src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper.cs#L94-L103

PR #6232, opened 2018-07-30
Log search service exceptions encountered during retry

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6229

Unfortunately we can't use `QuietLog` directly because `NuGet.Services.Search.Client` doesn't have a dependency on `NuGetGallery`, so I've added an `Action` parameter.

### DIFF
--- a/src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper.cs
+++ b/src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper.cs
@@ -15,6 +15,7 @@ namespace NuGet.Services.Search.Client
     {
         private readonly HttpClient _httpClient;
         private readonly IEndpointHealthIndicatorStore _endpointHealthIndicatorStore;
+        private readonly Action<Exception> _onException;
 
         private static readonly int PeriodToDelayAlternateRequest = 3000;
         private static readonly IComparer<int> HealthComparer;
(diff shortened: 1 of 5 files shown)


src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
```diff
@@ -52,7 +52,14 @@ public ExternalSearchService()
...
             if (_client == null)
             {
-                _client = new SearchClient(ServiceUri, "SearchGalleryQueryService/3.0.0-rc", null, _healthIndicatorStore, new TracingHttpHandler(Trace), new CorrelatingHttpClientHandler());
+                _client = new SearchClient(
+                    ServiceUri, 
+                    "SearchGalleryQueryService/3.0.0-rc", 
+                    null, 
+                    _healthIndicatorStore, 
+                    (exception) => QuietLog.LogHandledException(exception), 
+                    new TracingHttpHandler(Trace),
```
loic-sharma: Instead of `(exception) => QuietLog.LogHandledException(exception)` can you just do `QuietLog.LogHandledException`?

src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper.cs
```diff
@@ -24,16 +25,16 @@ static RetryingHttpClientWrapper()
...
         {
-            _httpClient = httpClient;
         }
 
-        public RetryingHttpClientWrapper(HttpClient httpClient, IEndpointHealthIndicatorStore endpointHealthIndicatorStore)
+        public RetryingHttpClientWrapper(HttpClient httpClient, IEndpointHealthIndicatorStore endpointHealthIndicatorStore, Action<Exception> onException)
         {
             _httpClient = httpClient;
             _endpointHealthIndicatorStore = endpointHealthIndicatorStore;
+            _onException = onException;
```
loic-sharma: >onException [](start = 27, length = 11)

Could we add null checks, or, default to an empty exception action?

skofman1:
@scottbommarito , why didn't you merge this?

scottbommarito:
@skofman1 - I believe I got distracted by something else and forgot to finish it. I think it was an on-call task.

PR #6240, opened 2018-07-31
Assert package ID and version instead of file size

joelverhagen:
This assertion is wrong now given repository signing.


joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6241.

PR #6242, opened 2018-08-01
[NuGet Symbol Server] Symbol package API push implementation

shishirx34:
This includes the gallery side standalone changes for symbol push with API. In subsequent PRs I will add support for validations on snupkgs, telemetry and auditing. 

Please review the code, this is work in progress and I will add tests to the same PR in subsequent commits.

@diverdan92 - Please review the strings in this PR.

I will be deploying this change to DEV to validate everything looks good before merging it in.


src/NuGetGallery/Strings.resx
```diff
@@ -924,4 +924,25 @@ If you would like to update the linked Microsoft account you can do so from the
     <value>The package was signed. The owner '{0}' must register the signing certificate to publish signed packages.</value>
     <comment>{0} is the signer name.</comment>
   </data>
+  <data name="SymbolsPackage_PackageIdAndVersionNotFound" xml:space="preserve">
+    <value>A package with ID '{0}' and version '{1}' does not exist. Please upload the package before uploading it's symbols.</value>
```
loic-sharma: >it's [](start = 112, length = 4)

its

src/NuGetGallery/Strings.resx
```diff
@@ -924,4 +924,25 @@ If you would like to update the linked Microsoft account you can do so from the
...
     <comment>{0} is the signer name.</comment>
   </data>
+  <data name="SymbolsPackage_PackageIdAndVersionNotFound" xml:space="preserve">
+    <value>A package with ID '{0}' and version '{1}' does not exist. Please upload the package before uploading it's symbols.</value>
+  </data>
+  <data name="SymbolsPackage_NotSymbolPackage" xml:space="preserve">
+    <value>The uploaded package is not a Symbols package. It is missing the only required 'SymbolsPackage' type or multiple package types specified.</value>
+  </data>
+  <data name="SymbolsPackage_FailedToReadPackage" xml:space="preserve">
+    <value>Failed to read the symbols package. Ensure it is a valid symbols package(snupkg).</value>
```
loic-sharma: >package [](start = 76, length = 7)

Should there be a space between `package` and `(snupkg)`?

src/NuGetGallery/Strings.resx
```diff
@@ -924,4 +924,25 @@ If you would like to update the linked Microsoft account you can do so from the
     <value>The package was signed. The owner '{0}' must register the signing certificate to publish signed packages.</value>
     <comment>{0} is the signer name.</comment>
   </data>
+  <data name="SymbolsPackage_PackageIdAndVersionNotFound" xml:space="preserve">
+    <value>A package with ID '{0}' and version '{1}' does not exist. Please upload the package before uploading it's symbols.</value>
+  </data>
+  <data name="SymbolsPackage_NotSymbolPackage" xml:space="preserve">
+    <value>The uploaded package is not a Symbols package. It is missing the only required 'SymbolsPackage' type or multiple package types specified.</value>
```
loic-sharma: >It is missing the only required 'SymbolsPackage' type or multiple package types specified. [](start = 58, length = 90)

This sentence seems weird to me: `It is missing the only required 'SymbolsPackage' type or multiple package types specified.`

How about something like: `The uploaded package is not a valid symbols package. The 'SymbolsPackage' package type is required`.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -333,6 +341,161 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+                {
+                    return new HttpStatusCodeWithBodyResult(HttpStatusCode.Unauthorized, Strings.SymbolsPackage_UploadNotAllowed);
+                }
+
+                // Read symbol package
+                using (var symbolPackageStream = ReadPackageFromRequest())
+                {
+                    try
+                    {
+                        if (FoundEntryInFuture(symbolPackageStream, out ZipArchiveEntry entryInTheFuture))
```
loic-sharma: >ZipArchiveEntry [](start = 72, length = 15)

Nit: `ZipArchiveEntry` can be just `var`. Feel free to ignore if you'd rather be explicit on the type.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -333,6 +341,161 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+
+                            // Do not allow to upload snupkg to a package which has symbols package pending validations.
+                            if (package.SymbolPackages.Any(sp => sp.StatusKey == PackageStatus.Validating))
+                            {
+                                return new HttpStatusCodeWithBodyResult(HttpStatusCode.Conflict, Strings.SymbolsPackage_ConflictValidating);
+                            }
+
+                            // Check if this user has the permissions to push the corresponding symbol package
+                            var apiScopeEvaluationResult = EvaluateApiScope(ActionsRequiringPermissions.UploadSymbolPackage, package.PackageRegistration, NuGetScopes.PackagePushVersion, NuGetScopes.PackagePush);
+                            if (!apiScopeEvaluationResult.IsSuccessful())
```
loic-sharma: Should the scope be evaluated before we check for validating symbol packages? If I don't own the package, I shouldn't be able to see that the package has symbols that are undergoing validation.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -333,6 +341,161 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+            {
+                // ASP.NET throws HttpException when maxRequestLength limit is exceeded.
+                return new HttpStatusCodeWithBodyResult(
+                    HttpStatusCode.RequestEntityTooLarge,
+                    Strings.PackageFileTooLarge);
+            }
+            catch (Exception)
+            {
+                //log telemetry
+                throw;
```
loic-sharma: Should there be an `ex.Log()` here?

src/NuGetGallery/Helpers/PackageHelper.cs
```diff
@@ -27,5 +29,92 @@ public static bool ShouldRenderUrl(string url)
...
+                    }
+
+                    // NuGet.Core compatibility - dependency versionspec can not be > 256 characters
+                    if (dependency.VersionRange != null && dependency.VersionRange.ToString().Length > 256)
+                    {
+                        throw new EntityException(Strings.NuGetPackagePropertyTooLong, "Dependency.VersionSpec", "256");
+                    }
+                }
+
+                // NuGet.Core compatibility - flattened dependencies should be < Int16.MaxValue
```
loic-sharma: >< [](start = 79, length = 1)

`<=`?

src/NuGetGallery/Services/IPackageUploadService.cs
```diff
@@ -42,5 +42,16 @@ public interface IPackageUploadService
...
+        /// <summary>
+        /// Create the symbols package entry in database, and upload the package to validation/uploads container for symbols.
+        /// This method commits the shared <see cref="IEntitiesContext"/>. This method can throw exceptions in exceptional
+        /// cases (such as database failures). This method does not dispose the provided stream but does read it.
+        /// </summary>
+        /// <param name="package">The package for which the symbols are to be uploaded.</param>
+        /// <param name="packageStreamMetadata">The metadata object for the uploaded snupkg.</param>
+        /// <param name="symbolPackageFile">The seekable stream containing the symbol package content (.snupkg).</param>
+        /// <returns>Awaitable task with <see cref="PackageCommitResult"/></returns>
+        Task<PackageCommitResult> CreateAndUploadSymbolsPackage(Package package, PackageStreamMetadata packageStreamMetadata, Stream symbolPackageFile);
```
loic-sharma: >CreateAndUploadSymbolsPackage [](start = 34, length = 29)

The other method is called `CommitPackageAsync`. Should we call this `CommitAndUploadSymbolsPackage`? It would make it a little clearer that this method commits to the database.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -155,6 +161,80 @@ public class PackageUploadService : IPackageUploadService
...
+                throw new InvalidOperationException(
+                    $"The symbol package to commit must have either the {PackageStatus.Available} or {PackageStatus.Validating} package status.");
+            }
+
+            try
+            {
+                if (symbolPackage.StatusKey == PackageStatus.Validating)
+                {
+                    await _symbolPackageFileService.SaveValidationPackageFileAsync(symbolPackage.Package, symbolPackageFile);
+                }
```
loic-sharma: It looks like this should handle the same corner case as the one covered by lines 256-298.

src/NuGetGallery/Strings.resx
diverdan92: Agreed it's awkward, but how about: `The uploaded package is not a valid symbols package. The required 'SymbolsPackage' package type is missing.`

src/NuGetGallery/Strings.resx
diverdan92: Also `(.snupkg)`?

src/NuGetGallery/Strings.resx
```diff
@@ -924,4 +924,25 @@ If you would like to update the linked Microsoft account you can do so from the
...
+    <value>Failed to read the symbols package. Ensure it is a valid symbols package(snupkg).</value>
+  </data>
+  <data name="SymbolsPackage_InvalidDataInNuspec" xml:space="preserve">
+    <value>Invalid metadata items specified in nuspec. Please remove '{0}' from the nuspec.</value>
+  </data>
+  <data name="SymbolsPackage_InvalidFiles" xml:space="preserve">
+    <value>Invalid file found in the uploaded package. Symbols package should only contain '{0}' files.</value>
+  </data>
+  <data name="SymbolsPackage_UploadNotAllowed" xml:space="preserve">
+    <value>You do not have the permissions to upload symbol packages.</value>
```
diverdan92: `symbols`

When would this ever get displayed, though?

src/NuGetGallery/Strings.resx
```diff
@@ -924,4 +924,25 @@ If you would like to update the linked Microsoft account you can do so from the
...
+    <value>The uploaded package is not a Symbols package. It is missing the only required 'SymbolsPackage' type or multiple package types specified.</value>
+  </data>
+  <data name="SymbolsPackage_FailedToReadPackage" xml:space="preserve">
+    <value>Failed to read the symbols package. Ensure it is a valid symbols package(snupkg).</value>
+  </data>
+  <data name="SymbolsPackage_InvalidDataInNuspec" xml:space="preserve">
+    <value>Invalid metadata items specified in nuspec. Please remove '{0}' from the nuspec.</value>
+  </data>
+  <data name="SymbolsPackage_InvalidFiles" xml:space="preserve">
+    <value>Invalid file found in the uploaded package. Symbols package should only contain '{0}' files.</value>
```
diverdan92: `Symbols packages`

src/NuGetGallery/Strings.resx
```diff
@@ -924,4 +924,25 @@ If you would like to update the linked Microsoft account you can do so from the
...
+    <value>A package with ID '{0}' and version '{1}' does not exist. Please upload the package before uploading it's symbols.</value>
+  </data>
+  <data name="SymbolsPackage_NotSymbolPackage" xml:space="preserve">
+    <value>The uploaded package is not a Symbols package. It is missing the only required 'SymbolsPackage' type or multiple package types specified.</value>
+  </data>
+  <data name="SymbolsPackage_FailedToReadPackage" xml:space="preserve">
+    <value>Failed to read the symbols package. Ensure it is a valid symbols package(snupkg).</value>
+  </data>
+  <data name="SymbolsPackage_InvalidDataInNuspec" xml:space="preserve">
+    <value>Invalid metadata items specified in nuspec. Please remove '{0}' from the nuspec.</value>
```
diverdan92: When would this get displayed?

src/NuGetGallery/Strings.resx
```diff
@@ -924,4 +924,25 @@ If you would like to update the linked Microsoft account you can do so from the
...
+  </data>
+  <data name="SymbolsPackage_InvalidFiles" xml:space="preserve">
+    <value>Invalid file found in the uploaded package. Symbols package should only contain '{0}' files.</value>
+  </data>
+  <data name="SymbolsPackage_UploadNotAllowed" xml:space="preserve">
+    <value>You do not have the permissions to upload symbol packages.</value>
+  </data>
+  <data name="SymbolsPackage_ConflictValidating" xml:space="preserve">
+    <value>It looks like there is already another copy of snupkg pending validation(s). Please wait for the validation(s) to finish before replacing the symbols package.</value>
```
diverdan92: `It looks like there is another copy of this symbols package pending validation(s). Please wait for the validation(s) to finish before trying to replace the symbols package.`

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -333,6 +341,161 @@ public virtual Task<ActionResult> CreatePackagePost()
             return CreatePackageInternal();
         }
 
+        [HttpPut]
+        [ApiAuthorize]
+        [ApiScopeRequired(NuGetScopes.PackagePush, NuGetScopes.PackagePushVersion)]
+        [ActionName("PushSymbolPackageApi")]
+        public virtual async Task<ActionResult> CreateSymbolPackagePutAsync()
```
joelverhagen: >CreateSymbolPackagePutAsync [](start = 48, length = 27)

Can we put the new symbol APIs into a new controller? This controller is already big.

src/NuGetGallery/Controllers/ApiController.cs
joelverhagen: >FoundEntryInFuture [](start = 28, length = 18)

Can we wrap these validations into `IPackageUploadService` or a new `ISymbolPackageUploadService`? These are duplicates with the packages controller and the package push endpoint in this same controller.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -155,6 +161,80 @@ public class PackageUploadService : IPackageUploadService
...
+                else if (symbolPackage.StatusKey == PackageStatus.Available)
+                {
+                    // Mark any other associated available symbol package for deletion.
+                    var availableSymbolPackages = package
+                        .SymbolPackages
+                        .Where(sp => sp.StatusKey == PackageStatus.Available 
+                            && sp != symbolPackage);
```
joelverhagen: >symbolPackage [](start = 37, length = 13)

If we are using `!=` I don't think we should depend on the default `ReferenceEquals`. Consider implementing `IEquatable<SymbolPackage>` on `SymbolPackage` which uses the `Key` for comparison. We do this on `User`.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -155,6 +161,80 @@ public class PackageUploadService : IPackageUploadService
...
+
+            try
+            {
+                if (symbolPackage.StatusKey == PackageStatus.Validating)
+                {
+                    await _symbolPackageFileService.SaveValidationPackageFileAsync(symbolPackage.Package, symbolPackageFile);
+                }
+                else if (symbolPackage.StatusKey == PackageStatus.Available)
+                {
+                    // Mark any other associated available symbol package for deletion.
```
joelverhagen: >package [](start = 66, length = 7)

packages (plural)

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -155,6 +161,80 @@ public class PackageUploadService : IPackageUploadService
             return package;
         }
 
+        public async Task<PackageCommitResult> CreateAndUploadSymbolsPackage(Package package, PackageStreamMetadata packageStreamMetadata, Stream symbolPackageFile)
+        {
+            var symbolPackage = _symbolPackageService.CreateSymbolPackage(package, packageStreamMetadata);
+
+            // TODO: Add Validations for symbols, for now set the status to Available.
```
joelverhagen: >TODO [](start = 15, length = 4)

TODO should mention a GitHub issue

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -155,6 +161,80 @@ public class PackageUploadService : IPackageUploadService
...
+
+                    if (availableSymbolPackages.Any())
+                    {
+                        // Mark the currently available package for deletion, and remove the file from the container, before uploading a newer one.
+                        availableSymbolPackages.First().StatusKey = PackageStatus.Deleted;
+
+                        await _symbolPackageFileService.DeletePackageFileAsync(symbolPackage.Package.PackageRegistration.Id, symbolPackage.Package.Version);
+                    }
+
+                    await _symbolPackageFileService.SavePackageFileAsync(symbolPackage.Package, symbolPackageFile);
```
joelverhagen: >SavePackageFileAsync [](start = 52, length = 20)

Why delete the save? Can't we just replace in this case? This whole area needs to be thoroughly analyzed for race conditions.

src/NuGetGallery/Services/SymbolPackageFileService.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using System.Web.Mvc;
+
+namespace NuGetGallery
+{
+    public class SymbolPackageFileService : CorePackageFileService, ISymbolPackageFileService
```
joelverhagen: >CorePackageFileService [](start = 44, length = 22)

Does it really need to bring in this whole base class? Can we make this API surface area more minimal?

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -0,0 +1,181 @@
...
+using NuGetGallery.Auditing;
+using NuGetGallery.Packaging;
+
+namespace NuGetGallery
+{
+    public class SymbolPackageService : CoreSymbolPackageService, ISymbolPackageService
+    {
+        private static readonly string SymbolPackageTypeName = "SymbolsPackage";
+        private static readonly string PDBExtension = ".pdb";
+        private static HashSet<string> AllowedExtensions = new HashSet<string>() {
```
joelverhagen: >HashSet [](start = 23, length = 7)

readonly

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -0,0 +1,181 @@
...
+using NuGet.Packaging.Core;
+using NuGetGallery.Auditing;
+using NuGetGallery.Packaging;
+
+namespace NuGetGallery
+{
+    public class SymbolPackageService : CoreSymbolPackageService, ISymbolPackageService
+    {
+        private static readonly string SymbolPackageTypeName = "SymbolsPackage";
+        private static readonly string PDBExtension = ".pdb";
```
joelverhagen: >PDBExtension [](start = 39, length = 12)

This strings can be `const`

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -0,0 +1,181 @@
...
+            }
+
+            return true;
+        }
+
+        private static bool IsSymbolPackage(PackageMetadata metadata)
+        {
+            var packageTypes = metadata.GetPackageTypes();
+            return packageTypes.Any()
+                && packageTypes.Count() == 1
```
joelverhagen: >packageTypes [](start = 19, length = 12)

Assert package type version as well. It should be `0.0` I think (which is the default).

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -0,0 +1,181 @@
...
+                && string.Equals(packageTypes.First().Name, SymbolPackageTypeName, StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static bool IsPortable(string pdbFile)
```
joelverhagen: >IsPortable [](start = 28, length = 10)

Unused?

src/NuGetGallery/Services/SymbolPackageService.cs
joelverhagen: >AllowedExtensions [](start = 39, length = 17)

Note that this will reject symbols packages that happened to be signed (p7s extension). I think this is right for now but I just wanted to call it out.

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -0,0 +1,181 @@
...
+                    Hash = symbolPackageStreamMetadata.Hash
+                };
+
+                _symbolPackageRepository.InsertOnCommit(symbolPackage);
+
+                return symbolPackage;
+            }
+            catch (Exception ex) when (ex is EntityException || ex is PackagingException)
+            {
+                throw new InvalidPackageException(ex.Message, ex);
```
joelverhagen: >InvalidPackageException [](start = 26, length = 23)

How could these exceptions be thrown by this code?

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -0,0 +1,181 @@
...
+
+                if (!IsSymbolPackage(packageMetadata))
+                {
+                    throw new InvalidPackageException(Strings.SymbolsPackage_NotSymbolPackage);
+                }
+
+                ValidateSymbolPackage(symbolPackage, packageMetadata);
+
+                // This will throw if the package contains an entry which will extract outside of the target extraction directory
+                await symbolPackage.ValidatePackageEntriesAsync(CancellationToken.None);
```
joelverhagen: >ValidatePackageEntriesAsync [](start = 36, length = 27)

Nit pass `CancellationToken` through service layer and force caller to `CancellationToken.None`

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -0,0 +1,181 @@
...
+
+            return true;
+        }
+
+        private static bool IsSymbolPackage(PackageMetadata metadata)
+        {
+            var packageTypes = metadata.GetPackageTypes();
+            return packageTypes.Any()
+                && packageTypes.Count() == 1
+                && string.Equals(packageTypes.First().Name, SymbolPackageTypeName, StringComparison.OrdinalIgnoreCase);
```
joelverhagen: >Equals [](start = 26, length = 6)

Can we use client's `PackageType` type for this comparison?

src/NuGetGallery/Helpers/PackageHelper.cs
```diff
@@ -27,5 +29,92 @@ public static bool ShouldRenderUrl(string url)
 
             return false;
         }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Maintainability", "CA1502:AvoidExcessiveComplexity")]
+        public static void ValidateNuGetPackageMetadata(PackageMetadata packageMetadata)
```
joelverhagen: >ValidateNuGetPackageMetadata [](start = 27, length = 28)

Nice. Glad this is in a shared location. :+1:

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -80,7 +82,9 @@ protected ApiController()
             ISecurityPolicyService securityPolicies,
             IReservedNamespaceService reservedNamespaceService,
             IPackageUploadService packageUploadService,
-            IPackageDeleteService packageDeleteService)
+            IPackageDeleteService packageDeleteService,
+            ISymbolPackageService symbolPackageService = null,
```
skofman1: why are those allowed to be null?

src/NuGetGallery/Controllers/ApiController.cs
skofman1: Agree. This should be as early as possible in the flow.

---
In reply to: [206733980](https://github.com/NuGet/NuGetGallery/pull/6242#discussion_r206733980) [](ancestors = 206733980)

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -333,6 +341,161 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+                            {
+                                HashAlgorithm = CoreConstants.Sha512HashAlgorithmId,
+                                Hash = CryptographyService.GenerateHash(
+                                    symbolPackageStream.AsSeekableStream(),
+                                    CoreConstants.Sha512HashAlgorithmId),
+                                Size = symbolPackageStream.Length
+                            };
+
+                            PackageCommitResult commitResult;
+                            using (Stream uploadStream = symbolPackageStream)
```
skofman1: won't this cause symbolPackageStream to be disposed twice? It's also in the using in line 362.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -333,6 +341,161 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+                var currentUser = GetCurrentUser();
+
+                // Check if symbol package upload is allowed for this user.
+                if (!ContentObjectService.SymbolsConfiguration.IsSymbolsUploadEnabledForUser(currentUser))
+                {
+                    return new HttpStatusCodeWithBodyResult(HttpStatusCode.Unauthorized, Strings.SymbolsPackage_UploadNotAllowed);
+                }
+
+                // Read symbol package
+                using (var symbolPackageStream = ReadPackageFromRequest())
```
joelverhagen: >ReadPackageFromRequest [](start = 49, length = 22)

Are we intentionally supporting both multipart request bodies and bodies that are just the snupkg bytes? Client only sends multipart with a constant file name.

If multipart, the content-type of the multi-part ignored?

These are all status quo with package but do we really need this broad of a protocol for a new endpoint?

src/NuGetGallery/Controllers/ApiController.cs
skofman1: There's a lot of code duplication between this method and the package push one. Explore ways to reduce this.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -333,6 +341,161 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+                            {
+                                // User cannot push a symbol package as the current user's scopes does not allow it to push for the corresponding package.
+                                return GetHttpResultFromFailedApiScopeEvaluationForPush(apiScopeEvaluationResult, id, version);
+                            }
+
+                            var packageStreamMetadata = new PackageStreamMetadata
+                            {
+                                HashAlgorithm = CoreConstants.Sha512HashAlgorithmId,
+                                Hash = CryptographyService.GenerateHash(
+                                    symbolPackageStream.AsSeekableStream(),
```
joelverhagen: >symbolPackageStream [](start = 36, length = 19)

This is buffering the whole thing to memory... is there a low cost fix to make this more reasonable?

src/NuGetGallery/Services/ISymbolPackageFileService.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using System.Web.Mvc;
+
+namespace NuGetGallery
+{
+    public interface ISymbolPackageFileService : ICorePackageFileService
+    {
+        /// <summary>
+        /// Creates an ActionResult that allows a third-party client to download the nupkg for the package.
```
skofman1: snupkg

src/NuGetGallery/Services/ISymbolPackageFileService.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// </summary>
+        Task<ActionResult> CreateDownloadSymbolPackageActionResultAsync(Uri requestUrl, SymbolPackage package);
+
+        /// <summary>
+        /// Creates an ActionResult that allows a third-party client to download the nupkg for the package.
```
skofman1: same

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -0,0 +1,181 @@
...
+
+            if (!CheckForAllowedFiles(symbolPackage))
+            {
+                throw new InvalidDataException(string.Format(Strings.SymbolsPackage_InvalidFiles, PDBExtension));
+            }
+        }
+
+        private static bool CheckForAllowedFiles(PackageArchiveReader symbolPackage)
+        {
+            foreach (var filePath in symbolPackage.GetFiles())
```
skofman1: please check the perf of this code on zips with max file count. We might want to avoid checking all the extensions here and check again async. 
For example: if this is slow, we might check the first 1000 files here and the rest in async validation.

src/NuGetGallery/Services/SymbolPackageService.cs
skofman1: We shouldn't validate portable pdbs here. It's a security and a perf concern.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -155,6 +161,80 @@ public class PackageUploadService : IPackageUploadService
             return package;
         }
 
+        public async Task<PackageCommitResult> CreateAndUploadSymbolsPackage(Package package, PackageStreamMetadata packageStreamMetadata, Stream symbolPackageFile)
```
skofman1: can you add documentation to this method? The flow is not clear

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: for out variable I like to be explicit since it can't be inferred.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: the validations don't really happen here, its the synchronous validation of the snupkg. But I agree, moved the check up.

src/NuGetGallery/Helpers/PackageHelper.cs
shishirx34: I just moved this code from `PackageService` but you are right. Fixed.

src/NuGetGallery/Strings.resx
shishirx34: This is because we want to not allow `Authors` or `Owners` metadata in the nuspec. We will validate these items during upload and return with a synchronous error. Feel free to phrase it better. I put these strings without putting too much thought into it.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: I think we should keep all the API related things together. Symbols needs only a single route for now and I don't think it warrants a need to create a new controller. I will look into refactoring the code between package put and this action which will reduce the redundancy.

src/NuGetGallery/Services/PackageUploadService.cs
shishirx34: In this case the references are going to match because we create the entity above, but I agree IEquatable is much better.

src/NuGetGallery/Services/PackageUploadService.cs
shishirx34: Good idea, I can surface another api from file service to overwrite file.

shishirx34: Lets talk about race conditions offline.

src/NuGetGallery/Services/SymbolPackageService.cs
shishirx34: Is that the version that the pack would add by default? The version attribute is optional in xsd, so someone might not specify it when hand crafting, hence I did not add the check.

src/NuGetGallery/Services/SymbolPackageService.cs
shishirx34: Yeah, we might want to add the support in for the future for "fast fail" scenario. I wanted to document the supposed validation. I would leave it in.

src/NuGetGallery/Services/SymbolPackageService.cs
shishirx34: Is that the extension of the signature file? I can add it in since we do not have any restrictions on signed/unsigned symbols as of now.

src/NuGetGallery/Services/SymbolPackageService.cs
shishirx34: Only might be thrown `EntityException`, removed the other one.

src/NuGetGallery/Services/SymbolPackageFileService.cs
shishirx34: talked offline, not needed.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: Will create the new service for symbol package upload.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -333,6 +341,161 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+                                    package,
+                                    packageStreamMetadata,
+                                    uploadStream.AsSeekableStream());
+                            }
+
+                            switch (commitResult)
+                            {
+                                case PackageCommitResult.Success:
+                                    break;
+                                case PackageCommitResult.Conflict:
```
cristinamanum: In what case we will get conflict? We support the symbol overwrite.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -333,6 +341,161 @@ public virtual Task<ActionResult> CreatePackagePost()
...
+                    Strings.PackageFileTooLarge);
+            }
+            catch (Exception)
+            {
+                //log telemetry
+                throw;
+            }
+        }
+
+        private bool FoundEntryInFuture(Stream stream, out ZipArchiveEntry entry)
```
cristinamanum: I am not sure what the purpose of this method is. Comments would be great.

src/NuGetGallery/Helpers/PackageHelper.cs
cristinamanum: Were any new validations introduced?

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: Its for not breaking the existing tests, I will remove those when I add tests for it.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: talked offline, leaving it as is since we can technically reuse the two parts in future for uploading package and symbols together.

src/NuGetGallery/Services/SymbolPackageService.cs
shishirx34: We aren't validating it, I simply added the code here for posterity, if and when we might need it.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -155,6 +161,80 @@ public class PackageUploadService : IPackageUploadService
...
+        {
+            var symbolPackage = _symbolPackageService.CreateSymbolPackage(package, packageStreamMetadata);
+
+            // TODO: Add Validations for symbols, for now set the status to Available.
+            symbolPackage.StatusKey = PackageStatus.Available;
+
+            if (symbolPackage.StatusKey != PackageStatus.Available
```
cristinamanum: Will it be case when the CreateSymbol package will return the status Available? I was expected to be Validating or FailedValidation

src/NuGetGallery/Services/SymbolPackageService.cs
shishirx34: I'll do an analysis on this.

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// <summary>
+        /// When no exceptions thrown, this method ensures the symbol package metadata is valid.
+        /// </summary>
+        /// <param name="symbolPackage">
+        /// The <see cref="PackageArchiveReader"/> instance providing the package metadata.
+        /// </param>
+        /// <exception cref="InvalidPackageException">
+        /// This exception will be thrown when a package metadata property violates a data validation constraint.
+        /// </exception>
+        public async Task EnsureValid(PackageArchiveReader symbolPackage)
```
cristinamanum: If Async please use the async in the method name.

src/NuGetGallery/Services/SymbolPackageService.cs
cristinamanum: If any validations are needed in the async, please add an issue for tracking. Currently the async validation does not validate the file extensions. 
IN this case also the allowed files extension set needs to be in the gallery core to be shared with the Orchestrator/jobs.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -155,6 +161,80 @@ public class PackageUploadService : IPackageUploadService
             return package;
         }
 
+        public async Task<PackageCommitResult> CreateAndUploadSymbolsPackage(Package package, PackageStreamMetadata packageStreamMetadata, Stream symbolPackageFile)
+        {
```
cristinamanum: Please remember when the message is sent to the Service bus it needs to be with SymbolPackage ValidatingType

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: You are right, looks like this bug exists in nupkg upload as well, but since we do not use it again it doesn't have impact.

src/NuGetGallery/Services/PackageUploadService.cs
shishirx34: Evaluated the concurrency issues here and it seems we are safe from such issues. There is a caveat for non async validations flow, but I think its fine to ignore it for now. I have added relevant comment to address it.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: This will be when you are trying to upload snupkg multiple times, we will have a conflict with existing snupkg in validations container.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: Its an existing nupkg code check for invalid zip entries. I will add comments.

src/NuGetGallery/Services/PackageUploadService.cs
shishirx34: It will be validating, once I add validation support. The available will be the status for non-async validations, like in local scenarios.

src/NuGetGallery/Services/PackageUploadService.cs
shishirx34: noted.

src/NuGetGallery/Services/SymbolPackageService.cs
shishirx34: If needed I will move this method to the corepackage file service, which you can reuse from async validation.

src/NuGetGallery/Services/SymbolPackageService.cs
skofman1: so please comment out/add comment/move to OneNote

src/NuGetGallery/Services/IPackageUploadService.cs
loic-sharma: I disagree. Commit indicates that we're persisting entity changes to the database, whereas create is ambiguous.

src/NuGetGallery/Helpers/PackageHelper.cs
shishirx34: No these are the same validations that we do on a nupkg, I refactored the code to reuse in symbols. Although I do check for `authors` and `owners` metadata to not be present in `EnsureValidAsync` of `SymbolPackageService`

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: Had an offline talk with Joel about this, and I am adding a new `ISymbolPackageUploadService` which would be the de-duplication place for all things symbols between API and Packages controllers. I will do the required refactoring when we need it in `PackagesController` i.e. UI upload support for symbols. I do not see the point in refactoring much here with the nupkg push code right now.

src/NuGetGallery/Services/IPackageUploadService.cs
shishirx34: sure it indicates persist action, but in this context we are creating an entity and uploading everything happens together. The reason having the other one differently was because there one method created the object and you passed it to another one for committing the object.

src/NuGetGallery/Services/SymbolPackageService.cs
shishirx34: I ran the analysis.. this doesn't have any perf impact. It ran the check on 64000 pdbs in 500 milliseconds. Which I believe is the max file entries we have. Took 20 seconds to run through 1.9 Million entries. This is an O(n) algorithm, reading file entries is also O(n) where 'n' is the number of entries in the zip archive.

src/NuGetGallery/Controllers/ApiController.cs
shishirx34: removed nulls

src/NuGetGallery.Core/Entities/SymbolPackage.cs
```diff
@@ -19,7 +19,6 @@ public class SymbolPackage
         /// <summary>
         /// Timestamp when this symbol was created.
         /// </summary>
-        [DatabaseGenerated(DatabaseGeneratedOption.Computed)]
```
cristinamanum: This will need migration right? I did not see any migration files included.

src/NuGetGallery/Services/SymbolPackageUploadService.cs
```diff
@@ -0,0 +1,127 @@
...
+                {
+                    await _symbolPackageFileService.SaveValidationPackageFileAsync(symbolPackage.Package, symbolPackageFile);
+                }
+                else if (symbolPackage.StatusKey == PackageStatus.Available)
+                {
+                    // Mark any other associated available symbol packages for deletion.
+                    var availableSymbolPackages = package
+                        .SymbolPackages
+                        .Where(sp => sp.StatusKey == PackageStatus.Available 
+                            && sp != symbolPackage);
```
cristinamanum: It was a comment before the refactoring. Please use the key or implement the IEquatable that uses the key as comparer.

src/NuGetGallery/Services/SymbolPackageService.cs
skofman1: That's great! Thanks for checking!

src/NuGetGallery/Services/SymbolPackageUploadService.cs
```diff
@@ -0,0 +1,127 @@
...
+                    $"The symbol package to commit must have either the {PackageStatus.Available} or {PackageStatus.Validating} package status.");
+            }
+
+            try
+            {
+                if (symbolPackage.StatusKey == PackageStatus.Validating)
+                {
+                    await _symbolPackageFileService.SaveValidationPackageFileAsync(symbolPackage.Package, symbolPackageFile);
+                }
+                else if (symbolPackage.StatusKey == PackageStatus.Available)
```
cristinamanum: This code path should not happen other than while testing. Can we add logs for this path?

cristinamanum: Also, if this is for testing why to have it at all? Why not add it under #if Debug directive and throw in  release?

src/NuGetGallery/Services/SymbolPackageUploadService.cs
shishirx34: I did add IEquatable to `SymbolPackage`, see [here](https://github.com/NuGet/NuGetGallery/pull/6242/files/57904a0a38af5eafc097fef752a410b235fc74ab#diff-ea910b7e84808753060a9515e369fb86R11)

src/NuGetGallery/Services/SymbolPackageUploadService.cs
shishirx34: This is not just for testing, it is for the standalone on prem gallery solution. If someone wants to host symbol server along with gallery, they can simply use this route to push the snupkg and it will become available since it won't have Async validations. We do same thing for nupkg upload as well. We cannot hide it and I don't think we should throw.

src/NuGetGallery.Core/Entities/SymbolPackage.cs
shishirx34: Acutally it doesn't need migration. Not super familiar with EF and migrations, but in my testing it looks like adding this value is only on the model and EF throws when inserting an entity for `Created` column cannot be null. Simply removing this attribute works fine.

shishirx34:
I have verified these changes in DEV env. I can now upload multiple snupkg files against existing nupkg and everything seems to be working as expected. Waiting on @joelverhagen to sign off and then merge this change into dev branch, followed by adding validations support for symbols.

src/NuGetGallery/Services/SymbolPackageUploadService.cs
```diff
@@ -0,0 +1,124 @@
...
+                        {
+                            availableSymbolPackage.StatusKey = PackageStatus.Deleted;
+                        }
+
+                        overwrite = true;
+                    }
+
+                    // Caveat: This doesn't really affect our prod flow since the package is validating, however, when the async validation
+                    // is disabled there is a chance that there could be concurrency issues when pushing multiple symbols simultaneously. 
+                    // This could result in an inconsistent data or multiple symbol entities marked as available. This could be sovled using etag
```
joelverhagen: >sovled [](start = 128, length = 6)

solved

tests/NuGetGallery.Facts/Services/SymbolPackageServiceFacts.cs
```diff
@@ -0,0 +1,242 @@
...
+            [Fact]
+            public async Task WillThrowForNullPackageArchiveReader()
+            {
+                var service = CreateService();
+
+                await Assert.ThrowsAsync<ArgumentNullException>(async () => await service.EnsureValidAsync(null));
+            }
+
+            [Fact]
+            public async Task WillThrowForMissingSymbolsPackageType()
```
joelverhagen: >WillThrowForMissingSymbolsPackageType [](start = 30, length = 37)

Can we also test correct package type name but incorrect version? And for missing package type?

Issue #6258, opened 2018-08-02
Align order of owners in gallery experience


joelverhagen:
The order of owners on NuGet.org seems to be random.

Take, for example, NLog. On the [package details page](https://www.nuget.org/packages/NLog/), we have this order:

![image](https://user-images.githubusercontent.com/94054/43613238-928264ac-9663-11e8-858c-54c838d8075f.png)

On [search results on the gallery](https://www.nuget.org/packages?q=nlog), we have this order:

![image](https://user-images.githubusercontent.com/94054/43613270-affbdc02-9663-11e8-9d26-aeee938c702f.png)

Oh, and the “Manage Owners” page has another different ordering:

![image](https://user-images.githubusercontent.com/94054/43613313-cc51bd40-9663-11e8-98a8-21b50169eb97.png)

We should align this.

joelverhagen:
These strings should be sorted using some variant of [`StringComparison`](https://msdn.microsoft.com/en-us/library/system.stringcomparison(v=vs.110).aspx)

I am thinking `Ordinal` or `OrdinalIgnoreCase`.

joelverhagen:
Decided on `InvariantCultureIgnoreCase`. Given the string `é, e, E, d, D, f, F`, our SQL Server collation and `InvariantCultureIgnoreCase` behave consistently and put the `é` before `f` which is probably a bit more user friendly.

joelverhagen:
Note that only the Scan and Sign job needs immediate attention. The other experiences can be fixed later.

joelverhagen:
As mentioned on #6614, manage packages page needs the fix up too.

PR #6259, opened 2018-08-02
[WIP] Order owners alphabetically

loic-sharma:
Part of https://github.com/NuGet/NuGetGallery/issues/6258

⚠️ TODO: 
* The manage owners page
* Verify on DEV that [this package's](https://dev.nugettest.org/packages/jQuery/) owners are ordered.

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -693,24 +693,24 @@
             <h2>Owners</h2>
             <ul class="list-unstyled owner-list">
                 @if (!Model.Owners.Any())
-    {
-            @ViewHelpers.AlertWarning(@<text>This package has no owners and is not being actively maintained.</text>)
-}
-else
-{
-foreach (var owner in Model.Owners)
-{
-            <li>
-                @if (!String.IsNullOrEmpty(owner.EmailAddress))
-    {
-            <a href="@Url.User(owner)" title="@owner.Username">@ViewHelpers.GravatarImage(owner.EmailAddress, owner.Username, Constants.GravatarElementSize)</a>
-}
-                <a href="@Url.User(owner)" title="@owner.Username">
-                    @owner.Username
-                </a>
-            </li>
-}
-}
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -693,24 +693,24 @@
...
-                </a>
-            </li>
-}
-}
+                {
+                    @ViewHelpers.AlertWarning(@<text>This package has no owners and is not being actively maintained.</text>)
+                }
+                else
+                {
+                    foreach (var owner in Model.Owners.OrderBy(o => o.Username, StringComparer.InvariantCultureIgnoreCase))
```
chenriksson: Should we OrderBy from the view model, so that any views that work with owners get the same ordering?

https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs#L29

loic-sharma: I considered that, but the view model does queries off of the owners field (for example, see [`HasSingleOwner`](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs#L75). This would affect all these queries as well.

scottbommarito: I think as long as you do a `.ToList()` when you `OrderBy` in the view model there shouldn't be any impact on the other queries. Also then we'd get the added bonus that all of those queries will then be ordered as well if we show them in the UI.

shishirx34: `HasSingleOwner` does a `Union` operation and that would kill of any `orderby` done previously. We will need to make sure we do `orderby` at all places where you are returning owners list. I would leave it as is if there are whole lot of places where you might have to order usernames.

loic-sharma:
Closing this PR as it has gotten stale and we have no plans of finishing this work in the short term

PR #6260, opened 2018-08-03
Signing owner selection box on Manage Packages page can cause table to overflow

### LINKED ISSUE #5909: Signing owner selection box on Manage Packages page can cause table to overflow

xavierdecoster:
Fixes #5909 

Explanation of fix: https://stackoverflow.com/questions/10672586/how-to-make-select-elements-shrink-to-max-width-percent-style-within-fieldset

Screenshot:
image

### DIFF
--- a/src/Bootstrap/dist/css/bootstrap-theme.css
+++ b/src/Bootstrap/dist/css/bootstrap-theme.css
@@ -1205,6 +1205,10 @@ img.reserved-indicator-icon {
 .page-manage-packages .inner-table {
   margin-bottom: 0;
 }
+.page-manage-packages .requiredSigner {
+  width: 100%;
+  max-width: 90%;
+}
 .page-delete-package h1 {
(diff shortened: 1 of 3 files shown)


joelverhagen:
Could we put the CSS in the page's CSS file rather than inline?

xavierdecoster:
@joelverhagen yes we can! :)

src/Bootstrap/dist/css/bootstrap-theme.css
```diff
@@ -1205,6 +1205,10 @@ img.reserved-indicator-icon {
 .page-manage-packages .inner-table {
   margin-bottom: 0;
 }
+.page-manage-packages .requiredSigner {
```
joelverhagen: nit: `required-signer`

xavierdecoster: Fixed!

Issue #6253, opened 2018-08-02
Support warnings during UI upload


joelverhagen:
Today, there can be errors returned by the UI upload flow. There is no warning except for the "warning" indicating the upload was resumed from the `uploads` container.

There can be warning during API push. In fact we have one when the pushed package is SemVer 2.0.0.

During UI upload, it would be ideal to show warnings before the user even submits the package. So the user would upload a .nupkg and, along with the "Review metadata" area, warnings could be shown.

This is required for https://github.com/NuGet/NuGetGallery/issues/6184.

Issue #6184, opened 2018-07-18
Warn package authors that they uploaded an unsigned package version after a signed version


joelverhagen:
@joelverhagen commented on [Tue Jul 17 2018](https://github.com/NuGet/Engineering/issues/1541)

"after" as defined by SemVer 2.0.0, not chronologically.

PR #6261, opened 2018-08-03
Warn when an unsigned package is pushed after a signed one

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6184.
Address https://github.com/NuGet/NuGetGallery/issues/6253.

## UI upload

image

## Command line

```
> nuget.exe push "system.rido.1.0.3 (unsigned).nupkg" -source https://localhost/api/v2/package -apikey myapikey
Pushing system.rido.1.0.3 (unsigned).nupkg to 'https://localhost/api/v2/package'...
  PUT https://localhost/api/v2/package/
WARNING: The previous package version '1.0.1' is signed but the uploaded package is unsigned. To avoid this warning, sign the package before uploading.
  Created https://localhost/api/v2/package/ 38504ms
Your package was pushed.

### DIFF
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Data;
 using System.Data.SqlClient;
 using System.Globalization;
@@ -631,6 +632,14 @@ namespace NuGetGallery
                                 }
                             }
 
+                            // Perform all the validations we can before adding the package to the entity context.
+                            var beforeValidationResult = await PackageUploadService.ValidateBeforeGeneratePackageAsync(packageToPush);
+                            var beforeValidationActionResult = GetActionResultOrNull(beforeValidationResult);
+                            if (beforeValidationActionResult != null)
+                            {
+                                return beforeValidationActionResult;
+                            }
+
                             var packageStreamMetadata = new PackageStreamMetadata
                             {
                                 HashAlgorithm = CoreConstants.Sha512HashAlgorithmId,
@@ -657,21 +666,16 @@ namespace NuGetGallery
                                 return new HttpStatusCodeWithBodyResult(HttpStatusCode.BadRequest, packagePolicyResult.ErrorMessage);
                             }
 
-                            var validationResult = await PackageUploadService.ValidatePackageAsync(
+                            // Perform validations that require the package already being in the entity context.
+                            var afterValidationResult = await PackageUploadService.ValidateAfterGeneratePackageAsync(
                                 package,
                                 packageToPush,
                                 owner,
                                 currentUser);
-                            switch (validationResult.Type)
+                            var afterValidationActionResult = GetActionResultOrNull(afterValidationResult);
+                            if (afterValidationActionResult != null)
                             {
-                                case PackageValidationResultType.Accepted:
-                                    break;
-                                case PackageValidationResultType.Invalid:
-                                case PackageValidationResultType.PackageShouldNotBeSigned:
-                                case PackageValidationResultType.PackageShouldNotBeSignedButCanManageCertificates:
-                                    return new HttpStatusCodeWithBodyResult(HttpStatusCode.BadRequest, validationResult.Message);
-                                default:
-                                    throw new NotImplementedException($"The package validation result type {validationResult.Type} is not supported.");
+                                return afterValidationActionResult;
                             }
 
                             await AutoCuratePackage.ExecuteAsync(package, packageToPush, commitChanges: false);
@@ -724,12 +728,15 @@ namespace NuGetGallery
 
                             TelemetryService.TrackPackagePushEvent(package, currentUser, User.Identity);
 
+                            var warnings = new List<string>();
+                            warnings.AddRange(beforeValidationResult.Warnings);
+                            warnings.AddRange(afterValidationResult.Warnings);
                             if (package.SemVerLevelKey == SemVerLevelKey.SemVer2)
                             {
-                                return new HttpStatusCodeWithServerWarningResult(HttpStatusCode.Created, Strings.WarningSemVer2PackagePushed);
+                                warnings.Add(Strings.WarningSemVer2PackagePushed);
                             }
 
-                            return new HttpStatusCodeResult(HttpStatusCode.Created);
+                            return new HttpStatusCodeWithServerWarningResult(HttpStatusCode.Created, warnings);
                         }
                     }
                     catch (InvalidPackageException ex)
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: src/NuGetGallery/Services/IPackageUploadService.cs [hunk 1]
```diff
@@ -28,7 +38,7 @@ public interface IPackageUploadService
         /// <param name="owner">The owner of the package.</param>
         /// <param name="currentUser">The current user.</param>
         /// <returns>The package validation result.</returns>
-        Task<PackageValidationResult> ValidatePackageAsync(
+        Task<PackageValidationResult> ValidateAfterGenerateAsync(
```
chenriksson: nit: should this be `ValidateAfterGeneratePackageAsync` to match the API above?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -38,7 +40,63 @@ public class PackageUploadService : IPackageUploadService
...
+                warnings);
+            if (result != null)
+            {
+                return result;
+            }
+
+            return PackageValidationResult.AcceptedWithWarnings(warnings);
+        }
+
+        private async Task<PackageValidationResult> ValidateSignedUnsignedTransitionAsync(
```
chenriksson: nit: name is a bit confusing. Would `ValidatePackageSigningState` be less complex? Or could we add comment to what the name means?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -38,7 +40,63 @@ public class PackageUploadService : IPackageUploadService
...
+
+            // If the package registration does not exist yet, there's no problem.
+            if (packageRegistration == null)
+            {
+                return null;
+            }
+
+            // Find the highest package version less than the new package that is Available. Deleted packages should
+            // be ignored and Validating or FailedValidation packages will not necessarily have certificate information.
+            var previousPackage = packageRegistration
```
chenriksson: Why only look at one version less? Should we continue to warn even if >1 unsigned packages are pushed after a signed package?

joelverhagen: Spoke to @rido-min about this and we decided against that. If the user wants to legitimately stop author signing, they shouldn't be stuck with the warning forever.

src/NuGetGallery/Services/PackageUploadService.cs
joelverhagen: Sure, I'll try to improve this.

src/NuGetGallery/Services/IPackageUploadService.cs
joelverhagen: Yeah, good point.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -38,7 +40,63 @@ public class PackageUploadService : IPackageUploadService
...
+                .Packages
+                .Where(x => x.PackageStatusKey == PackageStatus.Available)
```
loic-sharma: Nit: you may want to do a `.Select` with the fields you care about to reduce the size of the query results.

src/NuGetGallery/Strings.resx
```diff
@@ -943,4 +943,8 @@ Policy violations: {0}</value>
   <data name="SecurityPolicy_RequiredProjectUrlMissing" xml:space="preserve">
     <value>The package metadata is missing required ProjectUrl.</value>
   </data>
+  <data name="UploadPackage_SignedToUnsignedTransition" xml:space="preserve">
+    <value>The previous package version '{0}' is signed but the uploaded package is unsigned. To avoid this warning, sign the package before uploading.</value>
```
loic-sharma: Is it worth mentioning that the previous package was author signed? Technically, the previous package should be signed with a repository signature. Feel free to veto if you think this would confuse customers.

joelverhagen: Sure, adding the word "author" signed could help.

src/NuGetGallery/Services/PackageUploadService.cs
joelverhagen: I'd prefer to go the simple route until we see a perf need.

loic-sharma: Sure. I would test uploads on a large registration on DEV and make sure nothing weird happens then.

Now that I think of it, would it be worth limiting this query to the last 100 packages ordered by their key? Some registrations have more than a thousand packages.

joelverhagen: We already fetch the full package history for package details page. This PR is a much less common scenario. I can do the projection since it is simple. I would want to avoid a heuristic like last 100 packages by key since this is not conclusive.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -607,6 +614,21 @@ private async Task<ActionResult> CreatePackageInternal()
             }
         }
 
+        private static ActionResult GetActionResultOrNull(PackageValidationResult validationResult)
+        {
```
loic-sharma: The `PackagesController` has a test that ensures that the action doesn't throw for any `PackageValidationResultType` (see [this](https://github.com/NuGet/NuGetGallery/blob/945ffef30e20933942b664cadd04e2305462e79b/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs#L4211)). Does the `ApiController` have such a test too? I couldn't find one.

src/NuGetGallery/Services/IPackageUploadService.cs
```diff
@@ -10,6 +10,16 @@ namespace NuGetGallery
 {
     public interface IPackageUploadService
     {
+        /// <summary>
+        /// Validate the provided package archive reader before
+        /// <see cref="GeneratePackageAsync(string, PackageArchiveReader, PackageStreamMetadata, User, User)"/> is
+        /// called. This is useful for finding errors or warnings that can be should before the user verifies their UI
```
loic-sharma: >This is useful for finding errors or warnings that can be should before the user verifies their UI [](start = 20, length = 98)

`This is useful for finding errors or warnings that can be should before the user verifies their UI package upload` -> `This is useful for finding errors or warnings that should be caught before the user verifies their package upload`?

src/NuGetGallery/Controllers/ApiController.cs
joelverhagen: See `DoesNotThrowForAnyPackageValidationResultType` in `ApiControllerFacts`.

src/NuGetGallery/Services/IPackageUploadService.cs
joelverhagen: Nice catch.

joelverhagen: Fixed.

src/NuGetGallery/Strings.resx
joelverhagen: Fixed.

src/NuGetGallery/Services/PackageUploadService.cs
joelverhagen: Fixed.

src/NuGetGallery/Services/PackageUploadService.cs
joelverhagen: Fixed.

src/NuGetGallery/Services/IPackageUploadService.cs
joelverhagen: Fixed.

PR #6264, opened 2018-08-06
Support MicrosoftTeamSubscription on organization accounts

xavierdecoster:


xavierdecoster:
@scottbommarito allowing us to subscribe _organizations_ to the policy

PR #6273, opened 2018-08-07
Fix logging exceptions from background thread

zivkan:
QuietLog uses Elmah, which uses DI resolving at log time, rather than being contructed at HTTP request time. Autofac MVC's lifetime manager requires a HttpContext, which is null in a background thread, so there was no good/easy way to resolve the Elmah logger in a background thread. Logging only to app insights instead.

### DIFF
--- a/src/NuGetGallery/Services/BackgroundMessageService.cs
+++ b/src/NuGetGallery/Services/BackgroundMessageService.cs
@@ -5,16 +5,20 @@ using System;
 using System.Net.Mail;
 using System.Threading.Tasks;
 using AnglicanGeek.MarkdownMailer;
+using Elmah;
 using NuGetGallery.Configuration;
 
 namespace NuGetGallery.Services
 {
     public class BackgroundMessageService : MessageService
     {
-        public BackgroundMessageService(IMailSender mailSender, IAppConfiguration config, ITelemetryService telemetryService)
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Services/BackgroundMessageService.cs
```diff
@@ -34,7 +37,7 @@ protected override Task SendMessageAsync(MailMessage mailMessage)
                 catch (Exception ex)
                 {
                     // Log but swallow the exception.
-                    QuietLog.LogHandledException(ex);
+                    telemetryService.TrackException(ex, _ => { });
```
joelverhagen: why not use the overload that doesn't require an `Action`?

src/NuGetGallery/Services/BackgroundMessageService.cs
```diff
@@ -5,17 +5,21 @@
...
     public class BackgroundMessageService : MessageService
     {
-        public BackgroundMessageService(IMailSender mailSender, IAppConfiguration config, ITelemetryService telemetryService)
+        public BackgroundMessageService(IMailSender mailSender, IAppConfiguration config, ITelemetryService telemetryService, ErrorLog errorLog)
             :base(mailSender, config, telemetryService)
         {
+            this.errorLog = errorLog;
         }
 
+        private ErrorLog errorLog;
```
joelverhagen: nit: `readonly`

Issue #6049, opened 2018-06-12
[NuGet Symbol Server] Symbols revalidate APIs Implementation


cristinamanum:
Symbols need to have a way of revalidating. Task to capture the implementation work related with the revalidate scenario in Gallery.

PR #6276, opened 2018-08-07
[NuGet Symbol Server] Add Symbol push validation support

shishirx34:
Addresses https://github.com/NuGet/NuGetGallery/issues/6235 

Added test coverage, I have verified this locally, will check in DEV. Sending out a separate PR for the new configs for symbol validation service bus and topic.


src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -254,10 +254,10 @@
       <Version>4.8.0-preview4.5287</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Validation">
-      <Version>2.26.0-master-33196</Version>
+      <Version>2.28.0-master-37018</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Validation.Issues">
-      <Version>2.26.0-master-33196</Version>
+      <Version>2.28.0-master-37018</Version>
```
chenriksson: Is there a stable version to update to?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -468,29 +468,56 @@ private static DbConnection CreateDbConnection(ISqlConnectionFactory connectionF
...
             {
                 ConfigureValidationEntitiesContext(builder, diagnostics, configuration, secretInjector);
 
                 builder
-                    .RegisterType<AsynchronousPackageValidationInitiator>()
-                    .As<IPackageValidationInitiator>();
+                    .Register(c => {
+                        return new AsynchronousPackageValidationInitiator(
+                            c.ResolveKeyed<IPackageValidationEnqueuer>("PackageValidationEnqueuerBindingKey"),
+                            c.ResolveKeyed<IPackageValidationEnqueuer>("SymbolsPackageValidationEnqueuerBindingKey"),
```
chenriksson: Should these be constants - to keep this in sync w/ code above?

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -2220,21 +2208,9 @@
...
     </PackageReference>
-    <PackageReference Include="NuGet.Services.Validation">
-      <Version>2.26.0-master-33196</Version>
-    </PackageReference>
-    <PackageReference Include="NuGet.Services.Validation.Issues">
-      <Version>2.26.0-master-33196</Version>
-    </PackageReference>
-    <PackageReference Include="NuGet.Versioning">
-      <Version>4.8.0-preview4.5287</Version>
-    </PackageReference>
```
chenriksson: Nice! I assume these dependencies are pulled in from Gallery.Core?

src/NuGetGallery/Services/IPackageValidationInitiator.cs
```diff
@@ -21,5 +21,17 @@ public interface IPackageValidationInitiator
...
+        /// Starts the validation for the specified symbol package. The validation can be done asynchronously with respect
+        /// to the gallery and therefore may not be complete when the returned <see cref="Task"/> completes. This
+        /// pending validation state is indicated by the symbol package having the <see cref="PackageStatus.Validating"/>
+        /// status.
+        /// </summary>
+        /// <param name="package">The symbol package to initiate validation for.</param>
+        /// <returns>
+        /// The task which signals the completion of the validation initiation. The result of the <see cref="Task"/>
+        /// is the <see cref="PackageStatus"/> that should be applied to the package.</returns>
+        Task<PackageStatus> StartSymbolsPackageValidationAsync(SymbolPackage symbolPackage);
```
chenriksson: It's a little strange that the initiator has APIs for both nupkg and snupkg packages, but are only configured for one or the other.

It it possible to have a `IPackageValidationInitiator<TPackage>` so that the initiator can only enqueue based on the configured package type?

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -95,5 +98,13 @@ public IReadOnlyList<ValidationIssue> GetLatestValidationIssues(Package package)
...
             return issues;
         }
+
+        public async Task StartSymbolsPackageValidationAsync(SymbolPackage symbolPackage)
+        {
+            var symbolPackageStatus = await _initiator.StartSymbolsPackageValidationAsync(symbolPackage);
+            await _symbolPackageService.UpdateStatusAsync(symbolPackage,
+                symbolPackageStatus,
+                commitChanges: false);
+        }
```
chenriksson: Would having an `IPackageEntity` interface for use by `Package` and `SymbolsPackage` allow more code reuse?

src/NuGetGallery/NuGetGallery.csproj
shishirx34: Yes.

src/NuGetGallery.Core/NuGetGallery.Core.csproj
shishirx34: I don't see one in manage nuget solution.

shishirx34:
@chenriksson - I have managed to refactor the code and setup DI appropriately(thanks @joelverhagen), I tested this in synchronous and blocked async mode, looks good.  Please take a look at the refactoring.

src/NuGetGallery/Services/IPackageValidationInitiator.cs
```diff
@@ -8,18 +8,18 @@ namespace NuGetGallery
     /// <summary>
     /// Initiates validation for a specific package.
     /// </summary>
-    public interface IPackageValidationInitiator
+    public interface IPackageValidationInitiator<IPackageEntity>
```
agr: >IPackageEntity [](start = 49, length = 14)

Generic type arguments should start with `T`. It also could be useful to have `where` constraint to restrict it to the classes implementing the `IPackageEntity`

src/NuGetGallery/Services/SymbolPackageUploadService.cs
```diff
@@ -61,6 +58,11 @@ public async Task<PackageCommitResult> CreateAndUploadSymbolsPackage(Package pac
                 }
                 else if (symbolPackage.StatusKey == PackageStatus.Available)
                 {
+                    if (symbolPackage.Published == null)
+                    {
```
agr: nit: `!symbolPackage.Published.HasValue`?

src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
```diff
@@ -13,9 +13,9 @@ namespace NuGetGallery
     /// Initiates asynchronous validation on a package by enqueuing a message containing the package identity and a new
     /// <see cref="Guid"/>. The <see cref="Guid"/> represents a unique validation request.
     /// </summary>
-    public class AsynchronousPackageValidationInitiator : IPackageValidationInitiator
+    public class AsynchronousPackageValidationInitiator<TPackage> : IPackageValidationInitiator<TPackage> where TPackage: IPackageEntity
     {
```
agr: Why does it have to be generic?

chenriksson: My suggestion... see my comment [here](https://github.com/NuGet/NuGetGallery/pull/6276/files/60e5433d37f87f02c5459f96f89aaab59b690fe1#r208372485)

My concern was that you could create a `snupkg` initiator with an (invalid) API to initiate `nupkg` validation.

src/NuGetGallery/Services/SymbolPackageUploadService.cs
```diff
@@ -61,6 +58,11 @@ public async Task<PackageCommitResult> CreateAndUploadSymbolsPackage(Package pac
                 }
                 else if (symbolPackage.StatusKey == PackageStatus.Available)
                 {
+                    if (!symbolPackage.Published.HasValue)
+                    {
+                        symbolPackage.Published = DateTime.UtcNow;
+                    }
```
chenriksson: I was comparing this code with `nupkg` validation initiation... and they look very different.

https://github.com/NuGet/NuGetGallery/blob/760a81496fbdfcbfec932f3ad1b64cea097d80b8/src/NuGetGallery/Services/PackageUploadService.cs#L225

Just wondering, since the design reuses much of the validation infrastructure (even a copy)... why do the code paths look different? Again, feel free to ignore since I don't have as much context here.

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -95,5 +101,13 @@ public IReadOnlyList<ValidationIssue> GetLatestValidationIssues(Package package)
 
             return issues;
         }
+
+        public async Task StartSymbolsPackageValidationAsync(SymbolPackage symbolPackage)
```
chenriksson: nit: naming is a little inconsistent with corresponding `nupkg` API.

Maybe just `StartValidationAsync(SymbolPackage)` is enough?

src/NuGetGallery/Services/SymbolPackageUploadService.cs
shishirx34: I created this file for symbol specific stuff, the validation here isn't a copy though. The Package upload doesn't allow an overwrite, symbols does. It is meant to be different.

src/NuGetGallery.Core/Entities/IPackageEntity.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace NuGetGallery
+{
+    public interface IPackageEntity
+    {
+        string Id { get; }
+
+        string Version { get; }
+
+        ValidatingType Type { get; }
```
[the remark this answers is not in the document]
shishirx34: Yes it is an enum. This completely removes the dependency of checking the type in initiator. Also, it is going to be specific to each `IPackageEntity` so it makes sense to put it in here.

src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
shishirx34: @agr - code reusability for pushing to Package validation topic and symbol validation topic. The DI initializes these generic types to be used in Validation service.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -39,6 +39,14 @@
 
 namespace NuGetGallery
 {
+    public static class BindingKeys
```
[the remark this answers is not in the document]
shishirx34: I think we should keep it as a general class for binding constants, going forward use it for any DI setup that requires binding. This isn't specific for PackageValidation.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -468,36 +476,71 @@ private static DbConnection CreateDbConnection(ISqlConnectionFactory connectionF
...
+                            c.Resolve<IDiagnosticsService>());
+                    }).As<IPackageValidationInitiator<Package>>();
+
+                builder
+                    .Register(c => {
+                        return new AsynchronousPackageValidationInitiator<SymbolPackage>(
+                            c.ResolveKeyed<IPackageValidationEnqueuer>(BindingKeys.SymbolsPackageValidationEnqueuer),
+                            c.Resolve<IAppConfiguration>(),
+                            c.Resolve<IDiagnosticsService>());
+                    }).As<IPackageValidationInitiator<SymbolPackage>>();
```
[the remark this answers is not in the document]
shishirx34: I missed instantiating the `AsynchronousPackageValidationInitiator` and `ImmediateValidationInitiator` to be `Generic` which was causing the problem. See the code below of `RegisterGenerics` which sets up the DI appropriately.

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -95,5 +101,13 @@ public IReadOnlyList<ValidationIssue> GetLatestValidationIssues(Package package)
 
             return issues;
         }
+
+        public async Task StartValidationAsync(SymbolPackage symbolPackage)
```
skofman1: We will need the same validation functionality for symbols and for packages: 1. revalidation, browsing state, monitoring. It's not part of the MVP but is it tracked with issues?

src/NuGetGallery.Core/Entities/IPackageEntity.cs
skofman1: Consider renaming this to "PackageType". Validation type is not an attribute of the package, it's a behavior derived from the type of the package.

src/NuGetGallery/Services/ValidationService.cs
shishirx34: Yes, https://github.com/NuGet/NuGetGallery/issues/6049

src/NuGetGallery.Core/Entities/IPackageEntity.cs
shishirx34: I am reusing this from the Validations side for consistency, we use this to identify the package. While I agree with you, I do not think we should diverge for this enum. I'll alias it here I guess for ease.

src/NuGetGallery.Core/Entities/Package.cs
```diff
@@ -243,5 +244,9 @@ public Package()
         public virtual Certificate Certificate { get; set; }
 
         public virtual ICollection<SymbolPackage> SymbolPackages { get; set; }
+
+        public string Id => PackageRegistration.Id;
+
+        public ValidatingType Type => ValidatingType.Package;
```
chenriksson: `Type` could be confused with `GetType()`. Consider renaming to `ValidatingType`.

Is this used in this PR? Would remove if not.

src/NuGetGallery.Core/Entities/IPackageEntity.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    public interface IPackageEntity
+    {
+        string Id { get; }
+
+        string Version { get; }
```
chenriksson: Are there any shared code paths as part of this PR that would require `NormalizedVersion`?

src/NuGetGallery.Core/Entities/SymbolPackage.cs
```diff
@@ -3,12 +3,12 @@
...
 using System;
 using System.ComponentModel.DataAnnotations;
-using System.ComponentModel.DataAnnotations.Schema;
+using NuGet.Services.Validation;
 
 namespace NuGetGallery
 {
     public class SymbolPackage
-        : IEntity, IEquatable<SymbolPackage>
+        : IEntity, IPackageEntity, IEquatable<SymbolPackage>
```
chenriksson: Outside the scope of this PR... but why does `SymbolPackage` implement `IEquatable`, and `Package` does not? Is there functionality only specific to symbols that requires this?

src/NuGetGallery.Core/Entities/SymbolPackage.cs
```diff
@@ -51,6 +51,12 @@ public class SymbolPackage
         /// </summary>
         public byte[] RowVersion { get; set; }
 
+        public string Id => Package?.Id;
+
+        public string Version => Package?.Version;
+
+        public ValidatingType Type => ValidatingType.SymbolPackage;
```
chenriksson: See comment above (Package.cs).

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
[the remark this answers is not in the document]
chenriksson: Common coding standard I've seen is to keep 1:1 relation between classes and files, with names matching.

Given this, would consider moving `BindingKeys` into separate file or nesting under `DefaultDependenciesModule` class.

src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
```diff
@@ -24,36 +25,51 @@ public class AsynchronousPackageValidationInitiator : IPackageValidationInitiato
...
+                validatingType = ValidatingType.Package;
+            }
+            else if (package is SymbolPackage)
+            {
+                validatingType = ValidatingType.SymbolPackage;
+            }
+            else
+            {
+                throw new ArgumentException($"Unknown IPackageEntity type: {nameof(package)}");
+            }
```
chenriksson: Why add `ValidatingType` to `Package` and `SymbolPackage`, but not `IPackageEntity`? Seems like it complicates the code here.

src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
agr: `TPackageEntity` is not used for anything in this class. `IPackageValidationInitiator` can be made non-generic with `Task<PackageStatus> StartValidationAsync(IPackageEntity package)` method instead and then the code would still work unless I am missing something.

src/NuGetGallery/Services/SymbolPackageUploadService.cs
chenriksson: By "copy", I meant to say `snupkg` files share much of the same validation pipeline as `nupkg` files from what I remember... but Gallery just sends the initiate messages to a different service bus / pipeline instance.

From Gallery's perspective... wouldn't initiating a validation, and receiving validation status look about the same between `nupkg` and `snupkg` packages?

e.g., shouldn't Gallery set `package.Published = DateTime.UtcNow` in roughly the same place for both `nupkg` and `snupkg`?

Or are the code paths between the two that different in the Gallery?

src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
shishirx34: Talked and resolved this query offline.

agr: Talked offline: this makes DI setup quite complicated, so disregard.

src/NuGetGallery.Core/Entities/SymbolPackage.cs
shishirx34: This will be deleted. missed it.

src/NuGetGallery.Core/Entities/IPackageEntity.cs
shishirx34: not the code I am touching. I wouldn't refactor much for anything else right now.

src/NuGetGallery.Core/Entities/Package.cs
shishirx34: removing this.

src/NuGetGallery/Services/AsynchronousPackageValidationInitiator.cs
shishirx34: see @skofman1's comment. I think it doesn't make sense to put the `ValidatingType` on the entity.

src/NuGetGallery.Core/Entities/SymbolPackage.cs
shishirx34: See: https://github.com/NuGet/NuGetGallery/pull/6276/files#diff-c93f5b23258b3e89d85636deb43d0236R70

src/NuGetGallery/Services/SymbolPackageUploadService.cs
shishirx34: talked offline to clarify the differences here.

src/NuGetGallery.Core/NuGetGallery.Core.csproj
chenriksson: I believe our (undocumented) process for ServerCommon is to kick off a stable build when making changes there. I had similiar feedback in another PR, which is why I mention it.

shishirx34:
I've verified these changes in DEV with the new configs. Looks good.

PR #6288, opened 2018-08-09
Backfill repo metadata tool

skofman1:
/// This tool collect repository metadata for all packages in the DB from nuspec files in V3 flat container and updates DB with this data.
 /// Usage:
 /// 1. To collect repository metadata:
 ///     a. Configure app.config with DB information and service index url
   ///     b. Run this tool with: GalleryTools.exe -c
    /// This will create a file repositoryMetadata.txt with all collected data. You can stop the job anytime and restart. cursor.txt contains current position.    
    /// 
    /// 2. To update DB:
    ///     a. Run GalleryTools.exe -u  
    /// This will update DB from file repositoryMetadata.txt. You can stop the job anytime and restart.

### DIFF
--- a/src/GalleryTools/App.config
+++ b/src/GalleryTools/App.config
@@ -13,8 +13,10 @@
     <add key="Gallery.AzureStorage.Auditing.ConnectionString" value=""/>
     <add key="Gallery.AzureStorage.Errors.ConnectionString" value=""/>
     <add key="Gallery.AzureStorage.Packages.ConnectionString" value=""/>
-    <add key="Gallery.SqlServer" value="Data Source=(localdb)\mssqllocaldb; Initial Catalog=NuGetGallery; Integrated Security=True; MultipleActiveResultSets=True"/>
 
+    <add key="Gallery.SqlServer" value="Data Source=(localdb)\mssqllocaldb; Initial Catalog=NuGetGallery; Integrated Security=True; MultipleActiveResultSets=True"/>
+    <add key="Gallery.SupportRequestSqlServer" value="Data Source=(localdb)\mssqllocaldb; Initial Catalog=NuGetGallery; Integrated Security=True; MultipleActiveResultSets=True"/>
+ 
     <!-- KeyVault configuration to fetch secrets. -->
     <add key="KeyVault.VaultName" value=""/>
     <add key="KeyVault.ClientId" value=""/>
@@ -25,6 +27,8 @@
     <add key="Gallery.FileStorageDirectory" value="..\..\..\NuGetGallery\App_Data\Files"/>
     <add key="Gallery.LuceneIndexLocation" value="Temp"/>
     <add key="Gallery.IsHosted" value="false"/>
+
+    <add key="Gallery.ServiceDiscoveryUri" value="https://api.nuget.org/v3/index.json" />
   </appSettings>
   <entityFramework>
     <defaultConnectionFactory type="System.Data.Entity.Infrastructure.LocalDbConnectionFactory, EntityFramework">
--- a/src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
+++ b/src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
@@ -311,15 +311,15 @@ namespace GalleryTools.Commands
 
                             if (package != null)
                             {
-                                package.RepositoryUrl = metadata.Url;
+                                package.RepositoryUrl = metadata.RepositoryMetadata.Url;
 
-                                if (metadata.Type.Length >= 100)
+                                if (metadata.RepositoryMetadata.Type.Length >= 100)
                                 {
-                                    await _log.LogError(metadata.PackageId, metadata.PackageVersion, $"Repository type too long: {metadata.Type}");
+                                    await _log.LogError(metadata.PackageId, metadata.PackageVersion, $"Repository type too long: {metadata.RepositoryMetadata.Type}");
                                 }
                                 else
                                 {
-                                    package.RepositoryType = metadata.Type;
+                                    package.RepositoryType = metadata.RepositoryMetadata.Type;
                                 }
 
                                 counter++;
@@ -410,22 +410,23 @@ namespace GalleryTools.Commands
             }
         }
 
-        private class RepositoryMetadataLog : RepositoryMetadata
+        private class RepositoryMetadataLog
         {
             public DateTime CreationDate { get; set; }
             public string PackageId { get; set; }
             public string PackageVersion { get; set; }
+            public RepositoryMetadata RepositoryMetadata { get; set; }
 
             public RepositoryMetadataLog()
             {
             }
 
-            public RepositoryMetadataLog(RepositoryMetadata repositoryMetadata, DateTime creationDate, string packageId, string packageVersion) : 
-                base(repositoryMetadata.Type, repositoryMetadata.Url, repositoryMetadata.Branch, repositoryMetadata.Commit)
+            public RepositoryMetadataLog(RepositoryMetadata repositoryMetadata, DateTime creationDate, string packageId, string packageVersion)
             {
                 CreationDate = creationDate;
                 PackageId = packageId;
                 PackageVersion = packageVersion;
+                RepositoryMetadata = repositoryMetadata;
             }
         }
 
@@ -434,6 +435,12 @@ namespace GalleryTools.Commands
             public RepositoryMetadataLogMap()
             {
                 Map(x => x.CreationDate).TypeConverter<DateTimeConverter>();
+                Map(x => x.PackageId).Index(1);
+                Map(x => x.PackageVersion).Index(2);
+                Map(x => x.RepositoryMetadata.Type).Index(3);
+                Map(x => x.RepositoryMetadata.Url).Index(4);
+                Map(x => x.RepositoryMetadata.Branch).Index(5);
+                Map(x => x.RepositoryMetadata.Commit).Index(6);
             }
         }
 
--- a/src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
+++ b/src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
@@ -1,0 +1,501 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Data.Entity;
+using System.IO;
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Xml;
+using System.Xml.Linq;
+using Autofac;
+using Microsoft.Extensions.CommandLineUtils;
+using NuGet.Packaging;
+using NuGet.Packaging.Core;
+using NuGet.Services.Search.Client;
+using NuGetGallery;
+using NuGetGallery.Configuration;
+
+namespace GalleryTools.Commands
+{
+    /// <summary>
+    /// This tool collect repository metadata for all packages in the DB from nuspec files in V3 flat container and updates DB with this data.
+    /// Usage:
+    /// 1. To collect repository metadata:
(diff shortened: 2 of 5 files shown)


src/GalleryTools/App.config
```diff
@@ -13,8 +13,10 @@
     <add key="Gallery.AzureStorage.Auditing.ConnectionString" value=""/>
     <add key="Gallery.AzureStorage.Errors.ConnectionString" value=""/>
     <add key="Gallery.AzureStorage.Packages.ConnectionString" value=""/>
-    <add key="Gallery.SqlServer" value="Data Source=(localdb)\mssqllocaldb; Initial Catalog=NuGetGallery; Integrated Security=True; MultipleActiveResultSets=True"/>
 
+    <add key="Gallery.SqlServer" value="Data Source=(localdb)\mssqllocaldb; Initial Catalog=NuGetGallery; Integrated Security=True; MultipleActiveResultSets=True"/>
+    <add key="Gallery.SupportRequestSqlServer" value="Data Source=(localdb)\mssqllocaldb; Initial Catalog=NuGetGallery; Integrated Security=True; MultipleActiveResultSets=True"/>
```
joelverhagen: Why is `Gallery.SupportRequestSqlServer` added?

joelverhagen:
Spoke offline about this, but we should have a README.md that is next to the csproj that describes how this command works. Granted, the other commands don't have docs but we can improve this as we go on.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
```diff
@@ -0,0 +1,504 @@
...
+        {
+            config.Description = "Backfill repository information for packages in the Gallery";
+
+            CommandOption lastCreateTimeOption = config.Option("-l | --lastcreatetime", "The latest creation time of package we should check", CommandOptionType.SingleValue);
+            CommandOption collectData = config.Option("-c | --collect", "Collect Repository metadata and save in file", CommandOptionType.NoValue);
+            CommandOption updateDB = config.Option("-u | --update", "Update DB with Repository metadata", CommandOptionType.NoValue);
+            CommandOption fileName = config.Option("-f | --file", "File to use", CommandOptionType.SingleValue);
+
+            config.HelpOption("-? | -h | --help");
+            config.OnExecute(() =>
```
joelverhagen: `OnExecute` can accept a `Func<Task<int>>`. This is means the whole method can be `async` and you can avoid `.GetAwaiter().GetResult()`.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
```diff
@@ -0,0 +1,504 @@
...
+                }
+
+                Console.Read();
+            }
+
+            private static void Initialize(string connectionString, Uri serviceDiscoveryUri)
+            {
+                Log.LogMessage("Initializing");
+
+                _context = new EntitiesContext(connectionString, readOnly: true);
```
joelverhagen: 👍, readonly. nice.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
```diff
@@ -0,0 +1,504 @@
...
+                try
+                {
+                    Initialize(connectionString, serviceDiscoveryUri);
+
+                    // Get start time from cursor.
+                    var startTime = _cursor.GetCursorTime();
+
+                    Log.LogMessage($"Start time: {startTime.ToString("u")}");
+
+                    var packagesRepository = new EntityRepository<Package>(_context);
```
joelverhagen: You have DI above, why no use it? Makes this more flexible for the future (future backfills...)

joelverhagen: nit:

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
```diff
@@ -0,0 +1,504 @@
...
+
+                    var allPackages = packagesRepository.GetAll().Where(p => p.Created < lastCreateTime && p.Created > startTime && p.PackageStatusKey == PackageStatus.Available);
```
joelverhagen: So you can miss validating packages?

joelverhagen: Or less importantly failed validation packages that we reflow?

Paging over `Edited` will mean you never miss a package.

Also, consider adding a 5-30 minute subtract when first starting.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
```diff
@@ -0,0 +1,504 @@
...
+                if (_httpClient != null)
+                {
+                    _httpClient.Dispose();
+                    _httpClient = null;
+                }
+            }
+
+            private static async Task<RepositoryMetadata> GetRepositoryMetadata(string id, string normalizedVersion)
+            {
+                string nuspecUri = $"{_flatContainerUri}{id.ToLowerInvariant()}/{normalizedVersion.ToLowerInvariant()}/{id.ToLowerInvariant()}.nuspec";
```
joelverhagen: nit: trim trailing '/' and include '/' in the string format

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
```diff
@@ -0,0 +1,504 @@
...
+                    var xml = LoadXml(nuspecStream);
+                    var reader = new NuspecReader(xml);
+                    return reader.GetRepositoryMetadata();
+                }
+            }
+
+            private static string GetFlatContainerUri(Uri serviceDiscoveryUri)
+            {
+                var serviceDiscoveryClient = new ServiceDiscoveryClient(serviceDiscoveryUri);
+                return serviceDiscoveryClient.GetEndpointsForResourceType("PackageBaseAddress/3.0.0").GetAwaiter().GetResult().First().AbsoluteUri;
```
joelverhagen: These should be `async` methods. `.GetAwaiter().GetResult()` should only be in one place per app. (In most cases)

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
```diff
@@ -0,0 +1,504 @@
...
+            }
+
+            private static async Task WriteMetadata(DateTime creationDate, string packageId, string packageVersion, RepositoryMetadata repositoryMetadata)
+            {
+                await _metadataFileStreamWriter.WriteLineAsync(
```
joelverhagen: Consider one JSON object per line to avoid escaping problems. Or use a CSV library. Repository metadata could have a comma (type in particular).

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
```diff
@@ -0,0 +1,504 @@
...
+                {
+
+                    Dispose();
+                }
+
+                Console.Read();
+            }
+
+            private static async Task CommitBatch(DateTime cursorTime)
+            {
```
joelverhagen: You might run into memory problems since the EF context will hold onto all of the old entities.

joelverhagen: You can recreate the EF context each batch to mitigate this.

src/GalleryTools/App.config
skofman1: DefaultDependenciesModule can't be created without it.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
```diff
@@ -0,0 +1,504 @@
...
+    /// </summary>
+    public class BackfillRepositoryMetadataCommand
+    {
+        public static void Configure(CommandLineApplication config)
+        {
+            config.Description = "Backfill repository information for packages in the Gallery";
+
+            CommandOption lastCreateTimeOption = config.Option("-l | --lastcreatetime", "The latest creation time of package we should check", CommandOptionType.SingleValue);
+            CommandOption collectData = config.Option("-c | --collect", "Collect Repository metadata and save in file", CommandOptionType.NoValue);
+            CommandOption updateDB = config.Option("-u | --update", "Update DB with Repository metadata", CommandOptionType.NoValue);
```
joelverhagen: Consider a less aggressive starting point for how quickly `updateDB` updates batches. This could be done with  `batchSize` and `sleepSeconds` parameters. Then you can do some batches slowly on PROD before kicking it into high gear.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
```diff
@@ -0,0 +1,504 @@
...
+                            }
+                        }
+
+                        if (counter >= BatchSize)
+                        {
+                            await CommitBatch(metadata.createdDate);
+                            counter = 0;
+                        }
+
+                        metadata = await TryReadNextMetadata();
```
joelverhagen: nit: `do ... while` should be able to eliminate the duplicated `metadata = await TryReadNextMetadata();` line.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
skofman1: LastEdited is null for ~1M packages. Will stay with Created. Added support for validating packages. Don't see how clock skew is a factor if we compare always to Created that's DB generated.

joelverhagen: Those can come in out of order. There is no guarantee that a sequence of Created timestamps are committed in the same order that they are generated.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
joelverhagen: `_flatContainerUri` might have trailing `/` -- or might not. It's an easy config mistake to make.

just a nit

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
joelverhagen: You can order by LastEdited, PackageKey then. But it's fine if you include validating.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
skofman1: Using CsvHelper

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
skofman1: this is indeed an issue for the "collect" part. However, introducing batching there is too much work. Will leave as is. Thanks for the cursor the job can continue after crash.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
skofman1: I'm using the same batching config as I did for "hash" Gallery tool. It saved a 100 and had no sleep. There was not DB impact.

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
skofman1: will leave as is...

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
skofman1: done

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
skofman1: explain?

src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs
skofman1: done

PR #6289, opened 2018-08-09
Restore auto-add co-owner feature

xavierdecoster:
And select correct target account for package security policies.

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -396,7 +396,8 @@ namespace NuGetGallery
             RegisterCookieComplianceService(builder, configuration, diagnosticsService);
 
             builder.RegisterType<MicrosoftTeamSubscription>()
-                .SingleInstance();
+                .AsSelf()
+                .InstancePerLifetimeScope();
 
             // todo: bind all package curators by convention
             builder.RegisterType<WebMatrixPackageCurator>()
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -395,6 +395,9 @@ namespace NuGetGallery
 
             RegisterCookieComplianceService(builder, configuration, diagnosticsService);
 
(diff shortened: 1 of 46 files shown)


xavierdecoster:
@joelverhagen @skofman1 this is the commit you may want to zoom into: https://github.com/NuGet/NuGetGallery/pull/6289/commits/8ff0fcb27f6e771c6edd2cf9b40929c8e94a2b16

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -395,6 +395,9 @@ protected override void Load(ContainerBuilder builder)
 
             RegisterCookieComplianceService(builder, configuration, diagnosticsService);
 
+            builder.RegisterType<MicrosoftTeamSubscription>()
+                .SingleInstance();
```
joelverhagen: Does this really need to be a singleton? Would prefer if it was a new instance each time unless we are concerned and have measure performance impact.

src/NuGetGallery/Security/MicrosoftTeamSubscription.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Threading.Tasks;
+
+namespace NuGetGallery.Security
+{
+    public class MicrosoftTeamSubscription : IUserSecurityPolicySubscription
+    {
+        private Lazy<List<UserSecurityPolicy>> _policies = new Lazy<List<UserSecurityPolicy>>(InitializePoliciesList, isThreadSafe: true);
+
+        internal const string MicrosoftUsername = "Microsoft";
+        internal const string Name = nameof(MicrosoftTeamSubscription);
```
joelverhagen: This should be a const. Refactoring can break behavior here.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
xavierdecoster: Changed to instance per lifetime-scope

src/NuGetGallery/Security/MicrosoftTeamSubscription.cs
xavierdecoster: True. Updated.

Issue #6285, opened 2018-08-08
Add Cache-Control to "packages" container's blobs


loic-sharma:
The flat container has a Cache-Control of 2 minutes on its uploaded blobs. The "packages" container doesn't set a Cache-Control, thereby causing the CDN to cache it for 7 days instead. This causes issues with repository signing as a repository signed package will appeared as unsigned for up to 7  days.

joelverhagen:
This doesn't seems to be working:
```
PS > (invoke-webrequest https://api.nuget.org/packages/nanoframework.tools.debugger.uwp.1.0.0-rc001.nupkg -method head).RawContent
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: x-ms-request-id,Server,x-ms-version,x-ms-meta-da7b2905_0f3c_4262_921c_b1593d1336f1_ESRP_RequestId,x-ms-meta-SHA512,Content-Type,Last-Modified,ETag,Content-MD5,x-ms-lease-status,x-ms-blob-type,Content-Length,Date,Transfer-Encoding
Content-MD5: KHhf0+mU++N1nU2pEzatiQ==
Strict-Transport-Security: max-age=31536000
X-CDN-Rewrite: Root path in dist
X-Content-Type-Options: nosniff
x-ms-blob-type: BlockBlob
x-ms-lease-status: unlocked
x-ms-meta-da7b2905_0f3c_4262_921c_b1593d1336f1_ESRP_RequestId: 00634208-8203-4a44-b9ab-c5ac73d4c43e
x-ms-meta-SHA512: YzpxugyIVdu2/XF7Dk274NN4WvvOruY2unwsDoQYqPsV2SgxBsCVp1H8BUvuNJKqY3kWjk7tzM+tPECv4u1KZQ==
x-ms-request-id: a8fc2966-601e-007c-5cb7-64ad60000000
x-ms-version: 2009-09-19
Content-Length: 251996
Content-Type: application/octet-stream
Date: Mon, 15 Oct 2018 18:48:11 GMT
ETag: 0x8D632CDD9F0E109
Last-Modified: Mon, 15 Oct 2018 18:41:42 GMT
Server: Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
```

I think it's because the server-side copy is originating from the validation container which has no `Cache-Control` header.

loic-sharma:
* [ ] Add end-to-end test
[...]

PR #6297, opened 2018-08-10
Set the CacheControl property on stored blobs

loic-sharma:
This sets the `CacheControl` property on all blobs to 2 minutes. This affects all uploaded blobs for all blob containers.

Fixes https://github.com/NuGet/NuGetGallery/issues/6285

### DIFF
--- a/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
@@ -318,7 +318,7 @@ namespace NuGetGallery
             }
 
             blob.Properties.ContentType = GetContentType(folderName);
-            blob.Properties.CacheControl = CoreConstants.DefaultCacheControl;
+            blob.Properties.CacheControl = GetCacheControl(folderName);
             await blob.SetPropertiesAsync();
         }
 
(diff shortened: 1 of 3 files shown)


src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -318,6 +318,7 @@ public async Task SaveFileAsync(string folderName, string fileName, Stream file,
             }
 
             blob.Properties.ContentType = GetContentType(folderName);
+            blob.Properties.CacheControl = CoreConstants.DefaultCacheControl;
```
joelverhagen: Is this adding cache control to all files saved to blob storage in the gallery? Or just packages?

loic-sharma: This affects all uploaded blobs for all blob containers.

loic-sharma: I changed this to only affect the packages container.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -549,6 +550,7 @@ private static string GetContentType(string folderName)
                 case CoreConstants.PackageReadMesFolderName:
                     return CoreConstants.TextContentType;
 
+                case CoreConstants.ContentFolderName:
                 case CoreConstants.RevalidationFolderName:
```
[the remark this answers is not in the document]
loic-sharma: The the content type tests skipped `ContentFolderName`, which caused it to fail my new test. I'll fix the content type test.

Issue #5509, opened 2018-02-20
ConfirmEmailAddresses configuration doesn't behave as expected


scottbommarito:
We have a configuration named [`ConfirmEmailAddresses`](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Configuration/IAppConfiguration.cs#L114) that is intended to control whether or not email addresses must be confirmed.

It seems, however, after looking through the code, that this configuration only controls whether or not a person registering an account must confirm their email. Changing your email still requires confirmation regardless of what the configuration is set to. Ideally, this configuration should prevent changed email addresses from being confirmed as well.

PR #6310, opened 2018-08-14
Cleanup

skofman1:
1. Removed broken AppVeyor integration
2. Removed unused PagerDuty settings
3. Cleaned up + updated deployment instructions


src/NuGetGallery/Web.config
```diff
@@ -135,7 +131,7 @@
     <add key="Gallery.Brand" value="NuGet Gallery"/>
     <add key="Gallery.GalleryOwner" value="NuGet Gallery &lt;support@nuget.org&gt;"/>
     <add key="Gallery.GalleryNoReplyAddress" value="NuGet Gallery &lt;noreply@nuget.org&gt;"/>
-    <add key="Gallery.ConfirmEmailAddresses" value="true"/>
+    <add key="Gallery.ConfirmEmailAddresses" value="false"/>
```
skofman1: Having 'true' as default creates a bad experience for local deployments/web app deployments. Doesn't make sense to require confirmation if SMTP is not configured..

scottbommarito: This may make https://github.com/NuGet/NuGetGallery/issues/5509 more obvious.

PR #6312, opened 2018-08-14
Retry failed emails

zivkan:
The first commit is just a revert of what was previously reviewed and merged in dev. So, you can save yourself some time by just looking at the second commit's changes.

The problem with the previous PR that needed to be reverted was that previously emails are sent synchronously, so only one email being sent at a time. In order to prevent users waiting a long time for HTTP responses, emails now are sent in the background. However, when a single MVC action sends two or more messages, only the first would be sent and the others would throw an exception because the SmtpClient couldn't start sending a new message while an async operation was already in progress.

This PR fixes that issue by creating a BackgroundMessageServiceFactory. When an instance of BackgroundMessageService detects that it is being reused, it calls the factory to create a new instance, and sends the new email using the new instance.

### DIFF
--- a/src/NuGetGallery/Services/BackgroundMessageService.cs
+++ b/src/NuGetGallery/Services/BackgroundMessageService.cs
@@ -1,0 +1,95 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Mail;
+using System.Threading.Tasks;
+using AnglicanGeek.MarkdownMailer;
+using Elmah;
+using NuGetGallery.Configuration;
+
+namespace NuGetGallery.Services
+{
+    public class BackgroundMessageService : MessageService
+    {
+        public BackgroundMessageService(IMailSender mailSender, IAppConfiguration config, ITelemetryService telemetryService, ErrorLog errorLog)
+            :base(mailSender, config, telemetryService)
+        {
+            this.errorLog = errorLog;
+        }
+
+        private ErrorLog errorLog;
+
+        protected override Task SendMessageAsync(MailMessage mailMessage)
+        {
+            // Send email as background task, as we don't want to delay the HTTP response.
+            // Particularly when sending email fails and needs to be retried with a delay.
+            // MailMessage is IDisposable, so first clone the  message, to ensure if the
+            // caller disposes it, the message is available until the async task is complete.
+
+            var messageCopy = CloneMessage(mailMessage);
+
+            Task.Run(async () =>
+            {
+                try
+                {
+                    await base.SendMessageAsync(messageCopy);
+                }
+                catch (Exception ex)
+                {
+                    // Log but swallow the exception.
+                    QuietLog.LogHandledException(ex, errorLog);
+                }
+                finally
+                {
+                    messageCopy.Dispose();
+                }
+            });
+
+            return Task.CompletedTask;
+        }
+
+        private MailMessage CloneMessage(MailMessage mailMessage)
+        {
+            string from = mailMessage.From.ToString();
+            string to = mailMessage.To.ToString();
+
+            MailMessage copy = new MailMessage(from, to, mailMessage.Subject, mailMessage.Body);
+
+            copy.IsBodyHtml = mailMessage.IsBodyHtml;
+            copy.BodyTransferEncoding = mailMessage.BodyTransferEncoding;
+            copy.BodyEncoding = mailMessage.BodyEncoding;
+            copy.HeadersEncoding = mailMessage.HeadersEncoding;
+            foreach (System.Collections.Specialized.NameValueCollection header in mailMessage.Headers)
+            {
+                copy.Headers.Add(header);
+            }
+            copy.SubjectEncoding = mailMessage.SubjectEncoding;
+            copy.DeliveryNotificationOptions = mailMessage.DeliveryNotificationOptions;
+            foreach (var cc  in mailMessage.CC)
+            {
+                copy.CC.Add(cc);
+            }
+            foreach(var attachment in mailMessage.Attachments)
+            {
+                copy.Attachments.Add(attachment);
+            }
+            foreach (var bcc in mailMessage.Bcc)
+            {
+                copy.Bcc.Add(bcc);
+            }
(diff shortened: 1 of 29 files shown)


skofman1:
> namespace NuGetGallery.Services

missing file header


---
Refers to: src/NuGetGallery/Services/IBackgroundMessageServiceFactory.cs:1 in a1f6a8a. [](commit_id = a1f6a8ad9297264eaca562bddb63eed2670ff883, deletion_comment = False)

skofman1:
> using Autofac;

file header

---
Refers to: src/NuGetGallery/Services/BackgroundMessageServiceFactory.cs:1 in a1f6a8a. [](commit_id = a1f6a8ad9297264eaca562bddb63eed2670ff883, deletion_comment = False)

src/NuGetGallery/Services/BackgroundMessageService.cs
```diff
@@ -12,41 +13,58 @@ namespace NuGetGallery.Services
...
             :base(mailSender, config, telemetryService)
         {
             this.errorLog = errorLog;
+            this.messageServiceFactory = messageServiceFactory;
+            sentMessage = false;
         }
 
         private ErrorLog errorLog;
+        private BackgroundMessageServiceFactory messageServiceFactory;
+        private bool sentMessage;
```
scottbommarito: `private` fields should be prefixed with `_`

e.g. `_sentMessage`

src/NuGetGallery/Services/BackgroundMessageService.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        public BackgroundMessageService(IMailSender mailSender, IAppConfiguration config, ITelemetryService telemetryService, ErrorLog errorLog, BackgroundMessageServiceFactory messageServiceFactory)
+            :base(mailSender, config, telemetryService)
+        {
+            this.errorLog = errorLog;
+            this.messageServiceFactory = messageServiceFactory;
+            sentMessage = false;
+        }
+
+        private ErrorLog errorLog;
```
skofman1: coding conventions: fields are first in the class

src/NuGetGallery/Services/BackgroundMessageService.cs
```diff
@@ -0,0 +1,113 @@
...
+
+namespace NuGetGallery.Services
+{
+    public class BackgroundMessageService : MessageService
+    {
+        public BackgroundMessageService(IMailSender mailSender, IAppConfiguration config, ITelemetryService telemetryService, ErrorLog errorLog, BackgroundMessageServiceFactory messageServiceFactory)
+            :base(mailSender, config, telemetryService)
+        {
+            this.errorLog = errorLog;
+            this.messageServiceFactory = messageServiceFactory;
```
skofman1: null checks

src/NuGetGallery/Services/BackgroundMessageService.cs
```diff
@@ -12,41 +13,58 @@ namespace NuGetGallery.Services
 {
     public class BackgroundMessageService : MessageService
     {
-        public BackgroundMessageService(IMailSender mailSender, IAppConfiguration config, ITelemetryService telemetryService, ErrorLog errorLog)
+        public BackgroundMessageService(IMailSender mailSender, IAppConfiguration config, ITelemetryService telemetryService, ErrorLog errorLog, BackgroundMessageServiceFactory messageServiceFactory)
             :base(mailSender, config, telemetryService)
         {
             this.errorLog = errorLog;
+            this.messageServiceFactory = messageServiceFactory;
```
scottbommarito: if you prefix the private fields with `_` you don't need the `this.`

src/NuGetGallery/Services/BackgroundMessageServiceFactory.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Autofac;
+
+namespace NuGetGallery.Services
+{
+    public class BackgroundMessageServiceFactory : IBackgroundMessageServiceFactory
```
scottbommarito: I think `Autofac` is smart enough that you can replace this class with `Func<IMessageService>` and it will return a function that does the same thing as this.

skofman1:
How did you test this?

src/NuGetGallery/Services/BackgroundMessageService.cs
```diff
@@ -12,41 +13,58 @@ namespace NuGetGallery.Services
...
-            // MailMessage is IDisposable, so first clone the  message, to ensure if the
-            // caller disposes it, the message is available until the async task is complete.
+            // Some MVC controller actions send more than one message. Since this method sends
+            // the message async, we need a new IMessageService per email, to avoid calling
+            // SmtpClient.SendAsync on an instance already with an Async operation in progress.
 
-            var messageCopy = CloneMessage(mailMessage);
-
-            Task.Run(async () =>
+            if (sentMessage)
```
scottbommarito: It seems really odd to have an object that makes clones of itself to perform its primary purpose.
I think it's pretty confusing that the "BackgroundMessageService" is really just the "BackgroundSingleMessageService".
We also lose any reference to the additional `IMessageService`s created to send mails, so there's no visibility on how many mails are sent or how many instances of this class there are created by a single request.

I would rather have
- a class that sends emails (the section of this method inside the `else`)
- a class that has a method that sends emails by creating an instance of the class that sends emails

The big issue here is that we have one class (`IMessageService`) that creates and sends emails instead of one class for each of those purposes. If the class was split, it would improve the structure of this code, but I think that's outside the scope of this change.

Given that caveat, it may be better if additional emails sent during a single request waited for previous emails to finish, but I'm with how you have it here as well.

zivkan:
Testing was done manually by changing the SmtpUri to smtps://user:passwrd@localhost:25/, with nothing listening on the port so all attempts will fail with a network error. Putting a breakpoint on BackgroundMessageService's background task's exception handler, without the last commit, I could see many exceptions were about an async operation is already underway. Checking App Insights telemetry shows the smtp dependency failed. With the changes, the exception breakpoint only sees network related exceptions. This morning I tried setting SmtpUri to a valid SMTP server, and using the contact owner functionality with , I got two emails.

---
In reply to: [413279980](https://github.com/NuGet/NuGetGallery/pull/6312#issuecomment-413279980) [](ancestors = 413279980)

PR #6315, opened 2018-08-15
Typosquatting: add typosquatting check service codes 

zhhyu:
1. Typosquatting check service with distance calculation and comparison; 
2. Normalize the string before checking;
3. Call `public bool IsDistanceLessThanThreshold(string str1, string str2, int threshold)` to compare two strings; (changed to private)
4. Call `public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId)` to check typosquatting in the checlist.

Fixes: https://github.com/NuGet/Engineering/issues/1593


src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,314 @@
...
+            {'v', "VƔƲνѴѵѶѷ"},
+            {'w', "WŴŵƜẀẁẂẃẄẅωώШЩшщѡѿὠὡὢὣὤὥὦὧὼώᾠᾡᾢᾣᾤᾥᾦᾧῲῳῴῶῷԜԝ"},
+            {'x', "X×ΧχХхҲҳӼӽӾӿჯ"},
+            {'y', "YÝýÿŶŷŸƳƴȲȳɎɏỲỳΎΥΫγϒϓϔЎУЧуўҮүҶҷҸҹӋӌӮӯӰӱӲӳӴӵὙὛὝὟῨῩῪΎႯႸ𐒋𐒦"},
+            {'z', "ZŹźŻżŽžƵƶȤȥΖჍ"},
+            {'3', "ƷǮǯȜȝʒЗзэӞӟӠӡჳ"},
+            {'8', "Ȣȣ"},
+            {'_', ".-" }
+        };
+        private static Dictionary<char, char> _NormalizedMappingDictionary = new Dictionary<char, char>();
```
shishirx34: I think the convention we use is for static variables do not prefix with `_`

zhhyu: The format of variables has been changed.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    public class TyposquattingCheckService
+    {
+        private static readonly char[] SpecialCharacters = {'.', '_', '-'}; 
+        private static readonly Dictionary<char, string> SimilarCharacterDictionary = new Dictionary<char, string>()
```
joelverhagen: `private static readonly IReadOnlyDictionary<char, string>`

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+            }
+
+            for (int i = 1; i <= str1.Length; i++)
+            {
+                for (int j = 1; j <= str2.Length; j++)
+                {
+                    if (str1[i - 1] == str2[j - 1])
+                    {
+                        distances[i, j] = distances[i - 1, j - 1];
+                        path[i, j] = 'M';
```
joelverhagen: Looks like 'M', 'S', 'D', and 'I' are special codes. Can we use descriptive enums here?

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+            {
+                return false;
+            }
+
+            return GetDistance(str1, str2, threshold) <= threshold;
+        }
+
+        private int GetDistance(string str1, string str2, int threshold)
+        {
+            BasicEditDistanceInfo basicEditDistanceInfo = GetBasicEditDistanceWithPath(str1, str2);
```
joelverhagen: nit: `var`

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    public class TyposquattingCheckService
```
joelverhagen: A lot of these methods can be `static`

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+    {
+        private static readonly char[] SpecialCharacters = {'.', '_', '-'};
```
joelverhagen: `IReadOnlyList<char>`

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+
+namespace NuGetGallery
+{
+    public class TyposquattingCheckService
+    {
+        private static readonly char[] SpecialCharacters = {'.', '_', '-'}; 
+        private static readonly Dictionary<char, string> SimilarCharacterDictionary = new Dictionary<char, string>()
+        {
+            {'a', "AÀÁÂÃÄÅàáâãäåĀāĂăĄąǍǎǞǟǠǡǺǻȀȁȂȃȦȧȺΆΑάαἀἁἂἃἄἅἆἇἈἉἊἋἌἍἎἏӐӑӒӓὰάᾀᾁᾂᾃᾄᾅᾆᾇᾈᾊᾋᾌᾍᾎᾏᾰᾱᾲᾳᾴᾶᾷᾸᾹᾺΆᾼАДад"},
+            {'b', "BƀƁƂƃƄƅɃḂḃΒϦЂБВЪЬвъьѢѣҌҍႦႪხҔҕӃӄ" },
```
joelverhagen: Where did we get this data? Perhaps an XML doc should contain the source or methodology.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+            {'x', "X×ΧχХхҲҳӼӽӾӿჯ"},
+            {'y', "YÝýÿŶŷŸƳƴȲȳɎɏỲỳΎΥΫγϒϓϔЎУЧуўҮүҶҷҸҹӋӌӮӯӰӱӲӳӴӵὙὛὝὟῨῩῪΎႯႸ𐒋𐒦"},
+            {'z', "ZŹźŻżŽžƵƶȤȥΖჍ"},
+            {'3', "ƷǮǯȜȝʒЗзэӞӟӠӡჳ"},
+            {'8', "Ȣȣ"},
+            {'_', ".-" }
+        };
+        private static readonly Dictionary<char, char> NormalizedMappingDictionary = GetNormalizedMappingDictionary(SimilarCharacterDictionary);
+
+        // TODO: Threshold parameters will be saved in the configuration file.
```
joelverhagen: TODO should have a GitHub issue associated.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+            {
+                throw new ArgumentException("The lengths of two aligned strings are not same!");
+            }
+
+            int sameSubstitution = 0;
+            for (int i = 0; i < alginedStr2.Length; i++)
+            {
+                if (alignedStr1[i] != alginedStr2[i])
+                {
+                    if (alignedStr1[i] == '*' && SpecialCharacters.Contains(alginedStr2[i]))
```
joelverhagen: Would a hash set have better perf? This is a three item list so it should be measured....

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+                    {
+                        sameSubstitution += 1;
+                    }
+                    else if (SpecialCharacters.Contains(alignedStr1[i]) && alginedStr2[i] == '*')
```
joelverhagen: Named constant for '*'. I believe this character _has_ to be one that is not a valid ID character right? This should be commented on.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+            {'3', "ƷǮǯȜȝʒЗзэӞӟӠӡჳ"},
+            {'8', "Ȣȣ"},
+            {'_', ".-" }
+        };
+        private static readonly Dictionary<char, char> NormalizedMappingDictionary = GetNormalizedMappingDictionary(SimilarCharacterDictionary);
+
+        // TODO: Threshold parameters will be saved in the configuration file. 
+        private const int TyposquattingThreshold1 = 0;
+        private const int TyposquattingThreshold2 = 1;
+        private const int TyposquattingThreshold3 = 2;
```
joelverhagen: Suffixing a number to these `TyposquattingThreshold` is a little confusing. Are these three "tiers"? Could this be an array of three items instead? Why is the code hard coded to three?

Same comment for below with `TyposquattingThresholdInterval*`

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+            if (str1 == null)
+            {
+                throw new ArgumentNullException(nameof(str1));
+            }
+            if (str2 == null)
+            {
+                throw new ArgumentNullException(nameof(str2));
+            }
+
+            string newStr1 = Regex.Replace(str1, "[" + new string(SpecialCharacters) + "]", "");
```
joelverhagen: We don't need to recreate this string each time. It is also a `static readonly` (or perhaps a `const` and you generate the char[] from that).

joelverhagen: nit: `""` -> `string.Empty`.

src/NuGetGallery/Services/TyposquattingCheckService.cs
shishirx34: May be consider making this as a Interval Class array/list with the appropriate threshold for each interval.
```
public class Interval
{
   public int MinLength;
   public int MaxLength;
   public int Threshold;
}
```

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+        public void SetPackageIdCheckList(List<string> packageIdCheckList)
+        {
+            _packageIdCheckList = packageIdCheckList;
+            return;
+        }      
+        
+        private static Dictionary<char, char> GetNormalizedMappingDictionary(Dictionary<char, string> similarCharacterDictionary)
+        {
+            Dictionary<char, char> normalizedMappingDictionary = new Dictionary<char, char>();
+            foreach(var item in similarCharacterDictionary)
```
loic-sharma: >foreach( [](start = 12, length = 8)

We usually put spaces between `foreach` and the `(`: `foreach (...)`

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+            return normalizedMappingDictionary;
+        }
+
+        private int GetThreshold(string packageId)
+        {
+            if (packageId.Length < TyposquattingThresholdInterval1)
+            {
+                return TyposquattingThreshold1;
+            }
+            else if (packageId.Length >= TyposquattingThresholdInterval1 && packageId.Length < TyposquattingThresholdInterval2)
```
loic-sharma: >packageId.Length >= TyposquattingThresholdInterval1 [](start = 21, length = 51)

Can you remove the `packageId.Length >= TyposquattingThresholdInterval1` check as this is covered by the first `if` and this is an `else if`?

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+            {
+                return basicEditDistanceInfo.Distance;  
+            }
+            var alignedStrings = TraceBackAndAlignStrings(basicEditDistanceInfo.Path, str1, str2);
+            int refreshedEditDistance = RefreshEditDistance(alignedStrings[0], alignedStrings[1], basicEditDistanceInfo.Distance);
+
+            return refreshedEditDistance;
+        }
+
+        private BasicEditDistanceInfo GetBasicEditDistanceWithPath(string str1, string str2)
```
loic-sharma: >GetBasicEditDistanceWithPath [](start = 38, length = 28)

Could we add comments detailing what `GetBasicEditDistanceWithPath`, `TraceBackAndAlignStrings`, and `RefreshEditDistance` do? Examples of inputs to their resulting output would be great, too.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,314 @@
...
+
+        // TODO: Threshold parameters will be saved in the configuration file. 
+        private const int _TyposquattingThreshold1 = 0;
+        private const int _TyposquattingThreshold2 = 1;
+        private const int _TyposquattingThreshold3 = 2;
+        private const int _TyposquattingThresholdInterval1 = 30;
+        private const int _TyposquattingThresholdInterval2 = 50;
+
+        // TODO: popular packages checklist will be implemented
+        private List<string> _PackageIdCheckList = new List<string>();
```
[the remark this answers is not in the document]
zhhyu: Now it's a list of objects which contain Ids and owners

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,314 @@
...
+        public TyposquattingCheckService(List<string> packageIdCheckList) : this()
+        {
+            SetPackageIdCheckList(packageIdCheckList);
+        }
+
+        public void SetPackageIdCheckList(List<string> packageIdCheckList)
+        {
+            _PackageIdCheckList = packageIdCheckList;
+            return;
+        }
```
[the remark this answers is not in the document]
zhhyu: We may need to keep this because checklist will change along with the time.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: The special characters have been changed to descriptive enums for operations.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Methods have been changed to static.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Now we use Hashset for these special characters.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: XML comment has been added.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: It has been changed to Hashset.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: const placeholder is defined and used.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Thresholds are defined in another way.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+                throw new ArgumentNullException(nameof(str1));
+            }
+            if (str2 == null)
+            {
+                throw new ArgumentNullException(nameof(str2));
+            }
+
+            string newStr1 = Regex.Replace(str1, "[" + new string(SpecialCharacters) + "]", ""); 
+            string newStr2 = Regex.Replace(str2, "[" + new string(SpecialCharacters) + "]", "");
+            if (Math.Abs(newStr1.Length - newStr2.Length) > threshold)
```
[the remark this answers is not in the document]
zhhyu: I will think about that! Thanks!

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,313 @@
...
+                newStr1.Insert(k, '*');
+            }
+
+            alignedStrs[0] = newStr1.ToString();
+            alignedStrs[1] = newStr2.ToString();
+
+            return alignedStrs;
+        }
+
+        private int RefreshEditDistance(string alignedStr1, string alginedStr2, int basicEditDistance)
```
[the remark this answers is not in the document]
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,367 @@
...
+                    return thresholdInfo.Threshold;
+                }
+            }
+
+            throw new ArgumentException("There is no predefined typo-squatting threshold for this package Id!");
+        }
+
+        private static string NormalizeString(string str)
+        {
+            StringBuilder normalizedStr = new StringBuilder(str);
```
loic-sharma: You can use `var` here: `var normalizedStr`

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,367 @@
...
+            for (int i = 0; i < normalizedStr.Length; i++)
+            {
+                if (NormalizedMappingDictionary.ContainsKey(normalizedStr[i]))
```
loic-sharma: You could access the key just once:

```
if (NormalizedMappingDictionary.TryGetValue(normalizedStr[i], out var normalizedCharacter))
{
  normalizedStr[i] = normalizedCharacter;
}
```

zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,379 @@
...
+        private class BasicEditDistanceInfo
+        {
+            public int Distance { get; set; }
+            public PathInfo[,] Path { get; set; }
+        }  
+        
+        private enum PathInfo
+        {
+            [Description("Match")]
+            MATCH,
```
joelverhagen: >MATCH [](start = 12, length = 5)

Nit: No need to uppercase `MATCH`. And why the `Description`?

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,379 @@
...
+            [Description("Delete")]
+            DELETE,
+            [Description("Substitute")]
+            SUBSTITUTE,
+            [Description("Insert")]
+            INSERT,
+        }
+
+        public TyposquattingCheckService()
```
joelverhagen: Why the default ctor?

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,379 @@
...
+            if (str1 == null)
+            {
+                throw new ArgumentNullException(nameof(str1));
+            }
+            if (str2 == null)
+            {
+                throw new ArgumentNullException(nameof(str2));
+            }
+
+            string newStr1 = Regex.Replace(str1, SpecialCharactersToString, string.Empty);
```
joelverhagen: >Regex [](start = 29, length = 5)

Did we test the perf of this vs. hash set? need to test this on Release build btw

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,379 @@
...
+        }             
+    }
+
+    public class PackageInfo
+    {
+        public string Id { get; set; }
+        public HashSet<string> Owners { get; set; }
+    }
+
+    public class ThresholdInfo
```
joelverhagen: >ThresholdInfo [](start = 17, length = 13)

Can this be a read only type? This means properties only have { get; } and they are set in the ctor.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,379 @@
...
+                        continue;
+                    }
+                }
+            }
+
+            return basicEditDistance - sameSubstitution;
+        }             
+    }
+
+    public class PackageInfo
```
joelverhagen: >PackageInfo [](start = 17, length = 11)

Same, can this be readonly?

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: I am not very sure whether it could be a read only type. The `_thresholdsList` is a list of  `ThresholdInfo`, and each `ThresholdInfo` defines an interval. May we initialize each of them in the constructor as an element in the list? Besides, do we need a UI or configuration file to change the threshold in the future ?

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: I think `PackageInfo` will be changed because we will periodically update the checklist and the information of each package will be changed (like the owners of that package).

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Ok, I will test it later.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Chang to lower case and delete the description.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Delete the default constructor.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,371 @@
...
+
+        private class BasicEditDistanceInfo
+        {
+            public int Distance { get; set; }
+            public PathInfo[,] Path { get; set; }
+        }  
+        
+        private enum PathInfo
+        {
+            match,
```
loic-sharma: >match [](start = 12, length = 5)

.NET style is to capitalize enum values

src/NuGetGallery/Services/TyposquattingCheckService.cs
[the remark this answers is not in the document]
loic-sharma: If this is meant to be set externally, you should use properties. Example:

```csharp
public List<PackageInfo> PackageCheckList { get; set; }
```

---
In reply to: [210452093](https://github.com/NuGet/NuGetGallery/pull/6315#discussion_r210452093) [](ancestors = 210452093)

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,371 @@
...
+
+        public static void SetThresholdsList(List<ThresholdInfo> thresholdsList)
+        {
+            _thresholdsList = thresholdsList;
+            return;
+        }
+        
+        private static Dictionary<char, char> GetNormalizedMappingDictionary(IReadOnlyDictionary<char, string> similarCharacterDictionary)
+        {
+            Dictionary<char, char> normalizedMappingDictionary = new Dictionary<char, char>();
```
loic-sharma: >Dictionary<char, char> [](start = 12, length = 22)

Please use `var` especially if the resulting type is clear. On this line, the exact type is described by the `new` expression so annotating the variable's type adds no value.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,371 @@
...
+        {
+            match,
+            delete,
+            substitute,
+            insert,
+        }
+   
+        public TyposquattingCheckService(List<PackageInfo> packagesCheckList, List<ThresholdInfo> thresholdsList, IPackageService packageService) 
+        {
+            PackageService = packageService;
```
skofman1: please add null checks. 
Example:

PackageService = packageService ?? throw ArgumentNullException(nameof(packageService ));

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,371 @@
...
+            SetPackageIdCheckList(packagesCheckList);
+            SetThresholdsList(thresholdsList);
+        }
+
+        public static void SetPackageIdCheckList(List<PackageInfo> packagesCheckList)
+        {
+            _packagesCheckList = packagesCheckList;
+            return;
```
skofman1: no need for a return; if the return value of the method is void

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,371 @@
...
+                Distance = distances[str1.Length, str2.Length],
+                Path = path
+            };
+        }
+
+        /// <summary>
+        /// The following function is used to traceback based on the construction path and align two strings.
+        /// Example:  For two strings: "asp.net" "aspnet". After traceback and alignment, we will have aligned strings as "asp.net" "asp*net" ('*' is the placeholder).
+        /// </summary>
+        private static string[] TraceBackAndAlignStrings(PathInfo[,] path, string str1, string str2)
```
loic-sharma: Let's leave a comment about what the returned `string[]` contains: the two inputted strings after alignment.

skofman1:
> using System;

file header

---
Refers to: tests/NuGetGallery.Facts/Services/TyposquattingCheckServiceFacts.cs:1 in 25e9bdc. [](commit_id = 25e9bdc85ec55ef70bbdfe51a43328e5017c5c05, deletion_comment = False)

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,371 @@
...
+            int sameSubstitution = 0;
+            for (int i = 0; i < alignedStr2.Length; i++)
+            {
+                if (alignedStr1[i] != alignedStr2[i])
+                {
+                    if (alignedStr1[i] == PlaceholderForAlignment && SpecialCharacters.Contains(alignedStr2[i]))
+                    {
+                        sameSubstitution += 1;
+                    }
+                    else if (SpecialCharacters.Contains(alignedStr1[i]) && alignedStr2[i] == PlaceholderForAlignment)
```
loic-sharma: > alignedStr2[i] == PlaceholderForAlignment [](start = 74, length = 42)

The equality expression is very likely less expensive than the `Contains` method call. The performance may improve if the order of the `Contains` call and the equality expressions are switched.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,371 @@
...
+                if (package.Owners.Contains(uploadedPackageOwner.Username))
+                {
+                    return;
+                }
+                else
+                {
+                    if (IsDistanceLessThanThreshold(uploadedPackageId, package.Id, threshold))
+                    {
+                        var owners = PackageService.FindPackageRegistrationById(package.Id).Owners;
+                        foreach (var owner in owners)
```
loic-sharma: Let's describe the fact that this ensures correctness against the latest DB state.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/ITyposquattingOwnersDoubleCheck.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    public interface ITyposquattingOwnersDoubleCheck
+    {
+        bool IsOwnerAllowedTyposquatting(string packageId, string userName);
```
loic-sharma: >IsOwnerAllowedTyposquatting [](start = 13, length = 27)

Let's add a some documentation explaining what this does. Also, what do you think of the name `CanUserTyposquat` ?

zhhyu: Yes, I agree with you. : ) This looks better!

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,362 @@
...
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    public class TyposquattingCheckService : ITyposquattingCheckService
+    {
+        private readonly ITyposquattingOwnersDoubleCheck _typosquattingOwnersDoubleCheck;
+
```
loic-sharma: The .NET style is to place `const` fields first, then `static readonly`, and lastly `readonly`. Please update this code to match those guidelines.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,362 @@
...
+                if (packageId.Length >= thresholdInfo.LowerBound && packageId.Length < thresholdInfo.UpperBound)
+                {
+                    return thresholdInfo.Threshold;
+                }
+            }
+
+            throw new ArgumentException("There is no predefined typo-squatting threshold for this package Id!");
+        }
+
+        public static string NormalizeString(string str)
```
loic-sharma: >NormalizeString [](start = 29, length = 15)

Does this method need to be public?

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,362 @@
...
+            uploadedPackageId = NormalizeString(uploadedPackageId);
+
+            int countCollision = 0;
+            Parallel.ForEach(_packagesCheckList, (package, loopState) =>
+            {
+                if (package.Owners.Contains(uploadedPackageOwner.Username))
+                {
+                    return;
+                }
+                else
```
loic-sharma: >else [](start = 16, length = 4)

Nit: you could get remove the `else` to unnest the code by a level. Feel free to ignore if you prefer the code as is :)

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,362 @@
...
+                        // Double check the owners list in the latest DB. 
+                        if (_typosquattingOwnersDoubleCheck.IsOwnerAllowedTyposquatting(package.Id, uploadedPackageOwner.Username))
+                        {
+                            return;
+                        }
+
+                        Interlocked.Increment(ref countCollision);
+                        loopState.Stop();
+                    }
+                }
```
loic-sharma: >                 [](start = 17, length = 16)

You have a bunch of whitespace after the `}` here.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,362 @@
...
+                        Interlocked.Increment(ref countCollision);
+                        loopState.Stop();
+                    }
+                }                
+            });
+
+            return countCollision != 0;
+        }
+
+        protected internal static bool IsDistanceLessThanThreshold(string str1, string str2, int threshold)
```
loic-sharma: >protected internal [](start = 8, length = 18)

Why is this `protected internal`? Can it be `private` instead?

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,362 @@
...
+
+            return GetDistance(str1, str2, threshold) <= threshold;
+        }
+
+        private static int GetDistance(string str1, string str2, int threshold)
+        {
+            var basicEditDistanceInfo = GetBasicEditDistanceWithPath(str1, str2);
+            if (basicEditDistanceInfo.Distance <= threshold)
+            {
+                return basicEditDistanceInfo.Distance;
```
loic-sharma: >Distance [](start = 45, length = 8)

You have some whitespace at the end of line 195.

src/NuGetGallery/Services/TyposquattingOwnersDoubleCheck.cs
```diff
@@ -0,0 +1,31 @@
...
+            {
+                if (owner.Username == userName)
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }     
+    }
```
loic-sharma: There's some whitespace at the end of line 29

src/NuGetGallery/Services/TyposquattingOwnersDoubleCheck.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGetGallery.Services
+{
+    public class TyposquattingOwnersDoubleCheck : ITyposquattingOwnersDoubleCheck
+    {
+        private static IPackageService PackageService { get; set; }
+
```
loic-sharma: This should be `private readonly`. Each request will have its own scope and will create its own `IPackageService`. Thus, each request will need its own `TyposquattingOwnersDoubleCheck`

src/NuGetGallery/Services/ITyposquattingOwnersDoubleCheck.cs
zhhyu: I have renamed the type as `TyposquattingUserService` because I think it will better denote that this service belongs to typo-squatting issue.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: changed to private.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: remove it to unnest.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done

src/NuGetGallery/Services/TyposquattingOwnersDoubleCheck.cs
zhhyu: Done

src/NuGetGallery/Services/TyposquattingOwnersDoubleCheck.cs
zhhyu: Done

src/NuGetGallery/Services/TyposquattingOwnersDoubleCheck.cs
```diff
@@ -0,0 +1,31 @@
...
+        public TyposquattingOwnersDoubleCheck(IPackageService packageService)
+        {
+            PackageService = packageService ?? throw new ArgumentNullException(nameof(packageService));
+        }
+
+        public bool IsOwnerAllowedTyposquatting(string packageId, string userName)
+        {
+            var owners = PackageService.FindPackageRegistrationById(packageId).Owners;
```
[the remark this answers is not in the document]
zhhyu: Done

src/NuGetGallery/Services/TyposquattingUserService.cs
```diff
@@ -0,0 +1,37 @@
...
+            if (package == null)
+            {
+                return true;
+            }
+
+            var owners = package.Owners;
+            foreach (var owner in owners)
+            {
+                if (owner.Username == userName)
+                {
```
loic-sharma: I don't think this will work if the package is owned by an organization that the user is a part of. I'd ask Scott or Christie as they know that code best.

src/NuGetGallery/Services/TyposquattingUserService.cs
```diff
@@ -0,0 +1,37 @@
...
+            PackageService = packageService ?? throw new ArgumentNullException(nameof(packageService));
+        }
+
+        public bool CanUserTyposquat(string packageId, string userName)
+        {
+            var package = PackageService.FindPackageRegistrationById(packageId);
+            if (package == null)
+            {
+                return true;
+            }
```
loic-sharma: I misspoke earlier - this should return `false`. If a package no longer exists, the user cannot typosquat it.

src/NuGetGallery/Services/TyposquattingUserService.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGetGallery
+{
+    public class TyposquattingUserService : ITyposquattingUserService
+    {
+        private readonly IPackageService PackageService;
```
loic-sharma: >PackageService [](start = 41, length = 14)

For `private readonly` variables we use lowercased names that are prefixed by an underscore: `_packageService`.

tests/NuGetGallery.Facts/Services/TyposquattingCheckServiceFacts.cs
```diff
@@ -0,0 +1,132 @@
...
+            new PackageInfo { Id = "selenium_webDriver_microsoftdriver", Owners = new HashSet<string> { "owner7" } }
+        };
+
+        private static Mock<ITyposquattingUserService> _typosquattingOwnersDoubleCheck = new Mock<ITyposquattingUserService>();
+
+        [Fact]
+        public void CheckNotTyposquattingByDifferentOwnersTest()
+        {
+            // Arrange            
+            User owner = new User();
```
loic-sharma: >User [](start = 12, length = 4)

Please use `var` instead of repeating the type. There's a bunch these issues in the tests.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,361 @@
...
+        private static List<ThresholdInfo> _thresholdsList = new List<ThresholdInfo>
+        {
+            new ThresholdInfo { LowerBound = 0, UpperBound = 30, Threshold = 0 },
+            new ThresholdInfo { LowerBound = 30, UpperBound = 50, Threshold = 1 },
+            new ThresholdInfo { LowerBound = 50, UpperBound = 120, Threshold = 2 }
+        };
+
+        // TODO: popular packages checklist will be implemented
+        // https://github.com/NuGet/Engineering/issues/1624
+        public static List<PackageInfo> _packagesCheckList { get; set; }
```
loic-sharma: >public [](start = 8, length = 6)

For public properties we use the camel case format: `PackagesCheckList`.

src/NuGetGallery/Services/TyposquattingUserService.cs
zhhyu: Add new issue and TODO for this problem.

src/NuGetGallery/Services/TyposquattingUserService.cs
zhhyu: Changed to false.

src/NuGetGallery/Services/TyposquattingUserService.cs
zhhyu: Done!

tests/NuGetGallery.Facts/Services/TyposquattingCheckServiceFacts.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,363 @@
...
+            {'y', "YÝýÿŶŷŸƳƴȲȳɎɏỲỳΎΥΫγϒϓϔЎУЧуўҮүҶҷҸҹӋӌӮӯӰӱӲӳӴӵὙὛὝὟῨῩῪΎႯႸ𐒋𐒦"},
+            {'z', "ZŹźŻżŽžƵƶȤȥΖჍ"},
+            {'3', "ƷǮǯȜȝʒЗзэӞӟӠӡჳ"},
+            {'8', "Ȣȣ"},
+            {'_', ".-" }
+        };
+        private static readonly IReadOnlyDictionary<char, char> NormalizedMappingDictionary = GetNormalizedMappingDictionary(SimilarCharacterDictionary);
+
+        private readonly ITyposquattingUserService _userTyposquattingService;
+
```
loic-sharma: To follow .NET style, this instance field should be below the static fields.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,363 @@
...
+                }
+            }
+
+            throw new ArgumentException("There is no predefined typo-squatting threshold for this package Id!");
+        }
+
+        private static string NormalizeString(string str)
+        {
+            var normalizedStr = new StringBuilder(str);
+            for (int i = 0; i < normalizedStr.Length; i++)
```
loic-sharma: >int [](start = 17, length = 3)

You can use `var` here instead of `int`. I guess it doesn't matter much for this instance, but usually we use `var` whenever we can.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,363 @@
...
+            Parallel.ForEach(PackagesCheckList, (package, loopState) =>
+            {
+                // TODO: handle the package which is owned by an organization. 
+                // https://github.com/NuGet/Engineering/issues/1656
+                if (package.Owners.Contains(uploadedPackageOwner.Username))
+                {
+                    return;
+                }
+
+                if (IsDistanceLessThanThreshold(uploadedPackageId, package.Id, threshold))
```
loic-sharma: >IsDistanceLessThanThreshold [](start = 20, length = 27)

You could reduce a little nesting by inverting this check. For example:

```csharp
if (package.Owners.Contains(uploadedPackageOwner.Username))
{
    return;
}

if (!IsDistanceLessThanThreshold(uploadedPackageId, package.Id, threshold))
{
  return;
}

// Double check the owners list in the latest DB.
if (_userTyposquattingService.CanUserTyposquat(package.Id, uploadedPackageOwner.Username))
{
  return;
}

// The uploaded package id has collided to this well known package!
Interlocked.Increment(ref countCollision);
loopState.Stop();
```

This is entirely a stylistic recommendation, so feel free to do whatever you prefer :)

tests/NuGetGallery.Facts/Services/TyposquattingCheckServiceFacts.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            // Act
+            var typosquattingCheckResult = newService.IsUploadedPackageIdTyposquatting(uploadedPackageId, owner);
+
+            // Assert
+            Assert.True(typosquattingCheckResult);
+        }
+
+        [Fact]
+        public void CheckTyposquattingNulluploadedPackageId()
```
loic-sharma: >CheckTyposquattingNulluploadedPackageId [](start = 20, length = 39)

`CheckTyposquattingNulluploadedPackageId` -> `CheckTyposquattingNullUploadedPackageId` (I capitalized the u in `Uploaded`)

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,363 @@
...
+                if (NormalizedMappingDictionary.TryGetValue(normalizedStr[i], out var normalizedCharacter))
+                {
+                    normalizedStr[i] = normalizedCharacter;
+                }
+            }
+            
+            return normalizedStr.ToString();
+        }
+
+        public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner)
```
loic-sharma: >uploadedPackageOwner [](start = 84, length = 20)

I would also add a null check for the `uploadedPackageOwner` parameter.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: I have changed to `var` if it can be changed.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: I prefer to keep the original one because it could help better understand the logic there.

tests/NuGetGallery.Facts/Services/TyposquattingCheckServiceFacts.cs
zhhyu: Done

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Null check for owner and test case have been added.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,368 @@
...
+            Substitute,
+            Insert,
+        }   
+        
+        public TyposquattingCheckService(ITyposquattingUserService typosquattingUserService) 
+        {
+            _userTyposquattingService = typosquattingUserService ?? throw new ArgumentNullException(nameof(typosquattingUserService));
+        }
+
+        private static Dictionary<char, char> GetNormalizedMappingDictionary(IReadOnlyDictionary<char, string> similarCharacterDictionary)
```
skofman1: nit: private methods should be located below public method in the class. Helps to make the code more readable.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,368 @@
...
+        {
+            foreach (var thresholdInfo in _thresholdsList)
+            {
+                if (packageId.Length >= thresholdInfo.LowerBound && packageId.Length < thresholdInfo.UpperBound)
+                {
+                    return thresholdInfo.Threshold;
+                }
+            }
+
+            throw new ArgumentException("There is no predefined typo-squatting threshold for this package Id!");
```
skofman1: add the id itself to the error message. It will help you debug

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,368 @@
...
+
+            var threshold = GetThreshold(uploadedPackageId);
+            uploadedPackageId = NormalizeString(uploadedPackageId);
+
+            var countCollision = 0;
+            Parallel.ForEach(PackagesCheckList, (package, loopState) =>
+            {
+                // TODO: handle the package which is owned by an organization. 
+                // https://github.com/NuGet/Engineering/issues/1656
+                if (package.Owners.Contains(uploadedPackageOwner.Username))
```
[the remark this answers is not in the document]
zhhyu: The first check is used to help us avoid distance calculation if the package shares the same user. The second check is used to query the latest owner list from the database and double check it.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -0,0 +1,368 @@
...
+            var newStr2 = Regex.Replace(str2, SpecialCharactersToString, string.Empty);
+            if (Math.Abs(newStr1.Length - newStr2.Length) > threshold)
+            {
+                return false;
+            }
+
+            return GetDistance(str1, str2, threshold) <= threshold;
+        }
+
+        private static int GetDistance(string str1, string str2, int threshold)
```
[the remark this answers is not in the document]
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingDistanceCalculation.cs
```diff
@@ -0,0 +1,213 @@
...
+                    {
+                        continue;
+                    }
+                }
+            }
+
+            return basicEditDistance - sameSubstitution;
+        }
+
+        public static bool IsDistanceLessThanThreshold(string str1, string str2, int threshold)
```
skofman1: nit: public method should be above all private methods

zhhyu: Oops, sorry, I make the opposite ones.

zhhyu: Done!

Issue #5978, opened 2018-05-22
[NuGet Symbol Server] Gallery UI - Download Symbols 


cristinamanum:
UI for the download snupkg support.

shishirx34:
merged into DEV, please verify.

cristinamanum:
Verified in Dev and Int

Issue #5970, opened 2018-05-21
[NuGet Symbol Server] Download Rest APIs  Implementation


cristinamanum:
Task  for implementation of all the APIs needed to download a snupkg

shishirx34:
merged to dev. another issue available for tracking.

PR #6320, opened 2018-08-16
[NuGet Symbol Server] Add download symbols support to Package details page

shishirx34:
This PR adds support to download the snupkg from the package details page.

Addresses #5978  and #5970, also added tests.

The idea is to allow downloading a snupkg if the latest uploaded snupkg is "Available". If there is a newer uploaded snupkg that failed validations or still validating, no download allowed. This is taken care of in the download API and in the package details page view.

image

### DIFF
--- a/src/NuGetGallery/Strings.resx
+++ b/src/NuGetGallery/Strings.resx
@@ -950,6 +950,6 @@ If you would like to update the linked Microsoft account you can do so from the 
     <comment>{0} is the previous package's normalized version.</comment>
   </data>
   <data name="SymbolsPackage_PackageNotAvailable" xml:space="preserve">
-    <value>No availabe symbol package found ID {0} and version {1}.</value>
+    <value>No availabe symbol package found for ID {0} and version {1}.</value>
   </data>
 </root>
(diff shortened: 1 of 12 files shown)


src/NuGetGallery/Strings.resx
```diff
@@ -949,4 +949,7 @@ If you would like to update the linked Microsoft account you can do so from the
     <value>The previous package version '{0}' is author signed but the uploaded package is unsigned. To avoid this warning, sign the package before uploading.</value>
     <comment>{0} is the previous package's normalized version.</comment>
   </data>
+  <data name="SymbolsPackage_PackageNotAvailable" xml:space="preserve">
+    <value>No availabe symbol package found for ID {0} and version {1}.</value>
```
skofman1: please verify text with PM

shishirx34: @diverdan92 - verify.

skofman1:
Probably should get a different icon. Looks great otherwise!

src/NuGetGallery/Strings.resx
diverdan92: Where is this displayed? Otherwise, verified.

shishirx34: This is displayed when you try to download via rest endpoint and its not available, or the latest uploaded symbol package is validating/failed validation/deleted.

Issue #6133, opened 2018-07-03
[Organizations] Add member by email address


loic-sharma:
@loic-sharma commented on [Mon Jul 02 2018](https://github.com/NuGet/Engineering/issues/1510)

If you navigate to your organization and add a member, the input field reads "Add existing NuGet.org user". It's not clear that this field accepts only usernames but NOT email addresses. We should either:

1. Allow adding Organization members by email addresses
2. Change the input field to read something along the lines of "Add existing NuGet.org user by username"
3. Detect when an email address is inputted, and if so, show an error message that you must search by username

I think option 1 would be the best experience all around.

anangaur:
@loic-sharma 

Option 1 was originally proposed but was taken back later as there was privacy concern. We do not divulge email addresses of NuGet users.

loic-sharma:
I see. Do you have a preference between option 2 and 3?

scottbommarito:
After discussion with @anangaur, we are planning to do both 2 and 3.

PR #6329, opened 2018-08-20
Add member/add owner/transform to org fails with specific message if email is input instead of username

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6133

For add member/add owner/transform to org, if an email address is input instead of a username, fail with a specific message telling the user to type in a username instead of "account not found".

### DIFF
--- a/src/NuGetGallery/Constants.cs
+++ b/src/NuGetGallery/Constants.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using NuGet.Versioning;
+using System;
 
 namespace NuGetGallery
 {
@@ -69,6 +70,7 @@ namespace NuGetGallery
         internal const string EmailValidationRegexFirstPart = @"[-A-Za-z0-9!#$%&'*+\/=?^_`{|}~\.]+";
         internal const string EmailValidationRegexSecondPart = @"[A-Za-z0-9]+[\w\.-]*[A-Za-z0-9]*\.[A-Za-z0-9][A-Za-z\.]*[A-Za-z]";
         public const string EmailValidationRegex = "^" + EmailValidationRegexFirstPart + "@" + EmailValidationRegexSecondPart + "$";
+        public static TimeSpan EmailValidationRegexTimeout = TimeSpan.FromSeconds(5);
 
         public const string EmailValidationErrorMessage = "This doesn't appear to be a valid email address.";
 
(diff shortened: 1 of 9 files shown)


src/NuGetGallery/Constants.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using NuGet.Versioning;
+using System;
```
chenriksson: nit: sort System usings first

src/NuGetGallery/Constants.cs
```diff
@@ -69,6 +70,7 @@ public static class Constants
         internal const string EmailValidationRegexFirstPart = @"[-A-Za-z0-9!#$%&'*+\/=?^_`{|}~\.]+";
         internal const string EmailValidationRegexSecondPart = @"[A-Za-z0-9]+[\w\.-]*[A-Za-z0-9]*\.[A-Za-z0-9][A-Za-z\.]*[A-Za-z]";
         public const string EmailValidationRegex = "^" + EmailValidationRegexFirstPart + "@" + EmailValidationRegexSecondPart + "$";
+        public static TimeSpan EmailValidationRegexTimeout = TimeSpan.FromSeconds(5);
```
chenriksson: If Gallery has a lot of regex, maybe we should have a single, shared timeout for all? If there are expensive regex, they could use a separate constant - w/ comment as to why.

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -108,8 +109,12 @@ public virtual ActionResult GetPackageOwners(string id, string version)
         [ValidateAntiForgeryToken]
         public async Task<JsonResult> AddPackageOwner(string id, string username, string message)
         {
-            ManagePackageOwnerModel model;
-            if (TryGetManagePackageOwnerModel(id, username, isAddOwner: true, model: out model))
+            if (Regex.IsMatch(username, Constants.EmailValidationRegex, RegexOptions.None, Constants.EmailValidationRegexTimeout))
```
chenriksson: Maybe it would also be useful to add our own `Regex` helper, which ensures that timeout is always set?

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -108,8 +109,12 @@ public virtual ActionResult GetPackageOwners(string id, string version)
...
+            {
```
loic-sharma: We have legacy users with weird usernames. Are there any usernames in the database that match this regex? If so, we should move this check to after the database query.

scottbommarito: I discussed this with @anangaur, he said no.

Issue #3849, opened 2017-04-28
Remove dependency on NuGet.Services.Platform.Client


scottbommarito:
`NuGet.Services.Platform.Client` is a project from an old repository (`NuGet.Services.Platform`) that we haven't touched in years. We should remove the dependency (which should be trivial) and remove the repository.

skofman1:
@scottbommarito , which project is it referenced in?

scottbommarito:
`NuGetGallery` at least. It's used by the OData hijacking code.

skofman1:
@scottbommarito , and what should we be using instead?

scottbommarito:
We should be able to just use HttpClient directly and remove all dependencies.
________________________________

PR #6337, opened 2018-08-22
Use strong name version of third party dependencies

joelverhagen:
I've created a strong name version of all dependencies that are not already strong named. These packages are on `nuget-build`. The same SNK was used for all of these packages and is stored in our KeyVault.

This is the same process taken over a year ago to enable strong naming on NuGetGallery.Core, with:

- elmah.corelibrary.strongname
- elmah.sqlserver.strongname
- elmah.strongname

Progress on https://github.com/NuGet/Engineering/issues/1577.

### DIFF
--- a/src/NuGet.Services.Search.Client/Client/ServiceResponse.cs
+++ b/src/NuGet.Services.Search.Client/Client/ServiceResponse.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Search.Client
+{
+    public class ServiceResponse<T>
+    {
+        private readonly Func<Task<T>> _reader;
+
+        public ServiceResponse(HttpResponseMessage httpResponse)
+            : this(httpResponse, () => httpResponse.Content.ReadAsAsync<T>())
+        {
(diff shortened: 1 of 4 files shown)


src/NuGet.Services.Search.Client/Client/ServiceResponse.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
joelverhagen: Copied and fixed up from here:
https://github.com/NuGet/NuGet.Services.Platform/blob/master/src/NuGet.Services.Platform.Client/Client/ServiceResponse.cs

src/NuGet.Services.Search.Client/Client/ServiceResponse.cs
```diff
@@ -0,0 +1,36 @@
...
+        private readonly Func<Task<T>> _reader;
+
+        public ServiceResponse(HttpResponseMessage httpResponse)
+            : this(httpResponse, () => httpResponse.Content.ReadAsAsync<T>())
+        {
+        }
+
+        public ServiceResponse(HttpResponseMessage httpResponse, Func<Task<T>> reader)
+        {
+            HttpResponse = httpResponse;
```
skofman1: null checks

joelverhagen: Fixed.

joelverhagen:
Also addresses https://github.com/NuGet/NuGetGallery/issues/3849.

PR #6345, opened 2018-08-28
Accessibility - Fix nav roles

scottbommarito:
JAWS has a bug where they do not read our nav bar's tabs correctly.

After more analysis, it appears that we are using roles incorrectly, and we can avoid the bug by fixing our usage of roles.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/643637

### DIFF
--- a/src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
+++ b/src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
@@ -73,7 +73,7 @@
                              @ViewHelpers.ImageFallback(Url.Absolute("~/Content/gallery/img/logo-header-94x29.png")) />
                     </a>
                 </div>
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -73,7 +73,7 @@
                              @ViewHelpers.ImageFallback(Url.Absolute("~/Content/gallery/img/logo-header-94x29.png")) />
                     </a>
                 </div>
-                <div id="navbar" class="navbar-collapse collapse" role="menubar">
+                <div id="navbar" class="navbar-collapse collapse">
                     <ul class="nav navbar-nav" role="tablist">
```
chenriksson: `role="tablist"` is still ok?

Issue #2854, opened 2016-01-10
Navigation from "Manage My Packages" to Edit/Delete


robertmuehsig:
Current behavior:
Under "Manage my package" you have two options per package: Edit and Delete.

![image](https://cloud.githubusercontent.com/assets/756703/12224118/dd84ba4c-b7e8-11e5-9899-5c65fdfd1392.png)

Both options have a Save and Cancel option:

![image](https://cloud.githubusercontent.com/assets/756703/12224106/b27aae1a-b7e8-11e5-8607-f8b2336092c7.png)

I doesn't matter what you click you will be redirected to the package detail page.

I was a bit suprised that - at least - the cancel button didn't go back to the previous page (Manage My Package).

So, as an idea:
Option a) When you navigate to Edit/Delete from the Manage My Package page you will be redirected back to it when the user clicks on "Cancel"
Option b) ... "Cancel" and "Save" will redirect the user to the "Manage My Package" page (if this was the "previous page).

Enhancement to b) If the user clicks on save and he will be redirected to the "Manage My Package" page a small hint will be displayed like "Package XYZ now has a pending edit..."

Result of this change:
Hopefully managing existing packages on NuGet.org will be more intuitive.

What do you think?

csharpfritz:
There are a number of usability edits like this that we should consider as
we work on the look and feel of the public gallery.

You can also consider the inconsistency of the buttons at the bottom of the
Edit package page, Edit and Cancel should either both be buttons or both be
action links.  My preference: two buttons.





Jeff

robertmuehsig:
Would you accept a PR that changes the redirect behavior on Save/Cancel like i mentioned in Option b) ? If you navigate to Edit/Delete from "Manage My Packages" and hit save or cancel you will be redirected to the "Manage My Package" page?

Next question: Would you accept a PR where I change the "Cancel"-links to "Cancel" buttons?

maartenba:
Yes to option B, if the navigation can be done reliably. E.g. if I switch between editing various versions, cancel still should go back to where you'd expect it to go.

robertmuehsig:
All right - I will see what I can do.

AlexBHarley:
Hi I have started working on this. However my check at the moment is something like the below, as we only want to redirect if it's from the Manage Packages page.

`if (returnUrl == "http://nuget.org/account/Packages")` 

Is this the right way to go about it? returnUrl is just a `@Request.Url.AbsoluteUri` passed through.

[...]

PR #6350, opened 2018-08-28
Change button redirect

rasodu:
#2854 

- Edit page "Cancel" buttons redirect to previous page. "Submit" on the page still redirects to package detail page.(We can change submit button to redirect back to previous page. But do we want this? With current flow user has opportunity to see final result.)
- Delete page "Save" button redirect to previous page.


rasodu:
Will look into possibility to implement this bases on parameter passes in the URL instead of referral header. Will report back soon.

PR #6372, opened 2018-08-31
Telemetry metrics for symbol push.

cristinamanum:
Add 3 telemetry metrics for symbol push flow.
TrackSymbolPackagePushEvent
TrackSymbolPackagePushFailureEvent
TrackSymbolPackageFailedGalleryValidationEvent

### DIFF
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -480,8 +480,6 @@ namespace NuGetGallery
                                 Size = symbolPackageStream.Length
                             };
 
-                            TelemetryService.TrackSymbolPackagePushEvent(id, normalizedVersion);
-
                             PackageCommitResult commitResult = await SymbolPackageUploadService.CreateAndUploadSymbolsPackage(
                                 package,
                                 packageStreamMetadata,
@@ -498,6 +496,8 @@ namespace NuGetGallery
                                 default:
                                     throw new NotImplementedException($"The symbol package commit result {commitResult} is not supported.");
                             }
+
+                            TelemetryService.TrackSymbolPackagePushEvent(id, normalizedVersion);
 
                             return new HttpStatusCodeResult(HttpStatusCode.Created);
                         }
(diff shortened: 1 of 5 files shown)


src/NuGetGallery/Services/ITelemetryService.cs
```diff
@@ -146,5 +146,26 @@ public interface ITelemetryService
         /// <param name="success">Whether sending the email was successful.</param>
         /// <param name="attemptNumber">The number of attempts the message has tried to be sent.</param>
         void TrackSendEmail(string smtpUri, DateTimeOffset startTime, TimeSpan duration, bool success, int attemptNumber);
+
+        /// <summary>
+        /// A telemetry event emitted when a symbol package is pushed.
+        /// </summary>
+        /// <param name="packageId">The id of the package that has the symbols uploaded.</param>
+        /// <param name="packageVersion">The version of the package that has the symbols uploaded.</param>
+        void TrackSymbolPackagePushEvent(string packageId, string packageVersion);
```
loic-sharma: Could these be `NuGetVersion` instead of `string`?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -476,6 +480,8 @@ public virtual async Task<ActionResult> CreateSymbolPackagePutAsync()
                                 Size = symbolPackageStream.Length
                             };
 
+                            TelemetryService.TrackSymbolPackagePushEvent(id, normalizedVersion);
+
```
loic-sharma: Shouldn't this be tracked only if the `commitResult` is `PackageCommitResult.Success`?

loic-sharma: Could we move this telemetry into the `CreateAndUploadSymbolsPackage`?

---
In reply to: [215082841](https://github.com/NuGet/NuGetGallery/pull/6372#discussion_r215082841) [](ancestors = 215082841)

shishirx34: Yes, should be on successful push. may be move this to the end of this scope, before the return of `HttpStatusCode.Created`.

src/NuGetGallery/Services/ITelemetryService.cs
cristinamanum: It can. The majority of the APIs in the ITelemetry service uses the version as string, trying to be consistent.

src/NuGetGallery/Controllers/ApiController.cs
cristinamanum: Updated.

PR #6383, opened 2018-09-04
Add tool to verify the clear-text of an API key

joelverhagen:
I had a support request where I had to verify the identity of a nuget.org account with a lost email. Part of this verification process was comparing an expired API key against what we have in the database.

I wrote this tool to support this.

### DIFF
--- a/src/GalleryTools/Commands/VerifyApiKeyCommand.cs
+++ b/src/GalleryTools/Commands/VerifyApiKeyCommand.cs
@@ -1,0 +1,95 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.CommandLineUtils;
+using NuGetGallery;
+using NuGetGallery.Infrastructure.Authentication;
+
+namespace GalleryTools.Commands
+{
+    public static class VerifyApiKeyCommand
+    {
+        public static void Configure(CommandLineApplication config)
+        {
+            config.Description = "Verify a clear text API key matches a hashed API key.";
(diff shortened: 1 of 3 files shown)


joelverhagen:
Example output:

```
Testing 3 API key(s).

API key 1:
  Type:    apikey.v3
  Value:   **********************************************************************************************
  Matches: FALSE

API key 2:
  Type:    apikey.v3
  Value:   **********************************************************************************************
  Matches: TRUE

API key 3:
  Type:    apikey.v4
  Value:   ****************************************************************************************************************************
  Matches: FALSE
```

Real API keys instead of stars, of course.

src/GalleryTools/Commands/VerifyApiKeyCommand.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        private static int Execute(
+            CommandOption clearTextOption,
+            CommandOption credentialTypesOption,
+            CommandOption credentialValuesOption)
+        {
+            if (!clearTextOption.HasValue())
+            {
+                Console.Error.WriteLine($"The {clearTextOption.Template} option is required.");
```
loic-sharma: Does Microsoft's CommandLineUtils not have `IsRequired` extensions? See Nate McMaster's CommandLineUtils: https://github.com/natemcmaster/CommandLineUtils/blob/master/src/CommandLineUtils/Validation/ValidationExtensions.cs#L23

joelverhagen: It does not (as far as I can tell). Yeah, I think using it would be fine in the future.

Issue #6318, opened 2018-08-15
Inconsistent messaging around expired legacy key


joelverhagen:
From tweet:
https://twitter.com/kent_boogaart/status/1029263904080244737

![image](https://user-images.githubusercontent.com/94054/44176258-fd9c7900-a09d-11e8-94b9-9aa5550372b8.png)

PR #6384, opened 2018-09-05
Fix inconsistent messaging around expired legacy API key

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6318. This problem happens when the legacy API key was used more than 365 days ago but has not been used for an API call. This leads to the `HasExpired` boolean to be `true` but the `Expires` nullable `DateTime` to be null. The product code sets `HasExpired` but leaves `Expires` as null since this is set to `DateTime.UtcNow` at API call time.

The server side code is here:
https://github.com/NuGet/NuGetGallery/blob/1b9875706fb5fda0313de0e7c235603946ef3856/src/NuGetGallery/Authentication/AuthenticationService.cs#L591-L593

Also, fix some cases where we say "url" instead of "URL", because it bothered me.

# Expired

image

# Not yet expired

image

# Related

The aka.ms link in the PR goes here:
https://blog.nuget.org/20170202/introducing-scoped-api-keys.html#faqs

Eventually, this should go to a docs page. This work is non-blocking and is tracked here:
https://github.com/NuGet/docs.microsoft.com-nuget/issues/1050

### DIFF
--- a/src/NuGetGallery/Views/Packages/_ImportReadMe.cshtml
+++ b/src/NuGetGallery/Views/Packages/_ImportReadMe.cshtml
@@ -59,7 +59,7 @@
                 @ReadMePanel(
                     @<div class="form-group editable">
                         <input type="url" data-bind="value: Edit.ReadMe.SourceUrl" name="Edit.ReadMe.SourceUrl" class="form-control" id="ReadMeUrlInput"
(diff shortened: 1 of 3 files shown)


joelverhagen:
@anangaur prescribed the new text for not-yet-expired legacy API keys.

Issue #6349, opened 2018-08-28
[Gallery] Stop turning http urls into link automatically


loic-sharma:
The display package page automatically transforms links in the package's description and release notes to HTML anchor tags. This should only happen for HTTPS links; HTTP links should be kept as plaintext.

NickCraver:
I'm curious here: why? I'm all for `https://` everywhere too, but: what does this help?

anangaur:
We do not want people to click on links that are non https as they can be insecure. We will be showing the http  URLs but not converting them into links automatically.

skofman1:
@NickCraver , we are looking into ways to encourage package publishers to use `https` everywhere. We will punt this work item for the time being and start with warnings on upload.

PR #6391, opened 2018-09-06
Don't convert http urls to links in 'release notes' and 'description'

skofman1:
Tracking issue: https://github.com/NuGet/NuGetGallery/issues/6349


NickCraver:
Question added to root issue: https://github.com/NuGet/NuGetGallery/issues/6349 - this change doesn't seem like a net win for anyone, can someone add info on what the goal is there please?

PR #6398, opened 2018-09-07
Transform Project Url scheme to HTTPs for known domains

skofman1:


src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -183,14 +185,51 @@ private void InitializeRepositoryMetadata(string repositoryUrl, string repositor
...
+
+        private static bool IsGitHubUri(Uri uri)
         {
             return string.Equals(uri.Authority, "www.github.com", StringComparison.OrdinalIgnoreCase) ||
                    string.Equals(uri.Authority, "github.com", StringComparison.OrdinalIgnoreCase) ||
                    string.Equals(uri.Authority, "www.github.com:443", StringComparison.OrdinalIgnoreCase) ||
                    string.Equals(uri.Authority, "github.com:443", StringComparison.OrdinalIgnoreCase);
         }
 
+        private static bool IsCodeplexUri(Uri uri)
```
joelverhagen: This should be configuration driven.

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -183,14 +185,51 @@ private void InitializeRepositoryMetadata(string repositoryUrl, string repositor
...
+        {
+            return uri.Authority.EndsWith("codeplex.com", StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static bool IsMicrosoftUri(Uri uri)
+        {
+            return uri.Authority.EndsWith("microsoft.com", StringComparison.OrdinalIgnoreCase) ||
```
joelverhagen: EndsWith here is wrong. iworkatmicrosoft.com
Maybe ".microsoft.com"

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -183,14 +185,51 @@ private void InitializeRepositoryMetadata(string repositoryUrl, string repositor
...
         {
             return string.Equals(uri.Authority, "www.github.com", StringComparison.OrdinalIgnoreCase) ||
                    string.Equals(uri.Authority, "github.com", StringComparison.OrdinalIgnoreCase) ||
                    string.Equals(uri.Authority, "www.github.com:443", StringComparison.OrdinalIgnoreCase) ||
                    string.Equals(uri.Authority, "github.com:443", StringComparison.OrdinalIgnoreCase);
         }
 
+        private static bool IsCodeplexUri(Uri uri)
+        {
+            return uri.Authority.EndsWith("codeplex.com", StringComparison.OrdinalIgnoreCase);
```
joelverhagen: You can convert Microsoft.com:80

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
skofman1: what do you mean?

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
skofman1: done (for all domains)

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
skofman1: It actually just works: https://github.com/NuGet/NuGetGallery/pull/6398/files#diff-90a8ff4311e1d64cff7032bf34eab504R78

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
joelverhagen: In configuration, a list of domain names that are known to support HTTPS.

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
joelverhagen: 👍

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
skofman1: In which cases do you think this list will be different between deployments?

joelverhagen: I was thinking that it would be a lot easier to add more domains to the list if it's configuration. Then you don't have to write a new method (which is consistent with your approach above). Also, this would allow other people running their own gallery to add their own corporate/interesting domains.

Not a huge deal.

PR #6408, opened 2018-09-07
Covert to https URLs in known domains in release notes, description

skofman1:


src/NuGetGallery/Helpers/UriHelpers.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGetGallery
+{
+    public static class UriHelpers
```
chenriksson: nit: I think `UriExtensions` would be more consistent with our existing naming patterns

src/NuGetGallery/Helpers/UriHelpers.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            return IsGitHubUri(uri) || IsCodeplexUri(uri) || IsMicrosoftUri(uri);
+        }
+
+        public static bool IsGitHubUri(this Uri uri)
+        {
+            return string.Equals(uri.Authority, "www.github.com", StringComparison.OrdinalIgnoreCase) ||
+                   string.Equals(uri.Authority, "github.com", StringComparison.OrdinalIgnoreCase) ||
+                   string.Equals(uri.Authority, "www.github.com:443", StringComparison.OrdinalIgnoreCase) ||
+                   string.Equals(uri.Authority, "github.com:443", StringComparison.OrdinalIgnoreCase);
```
chenriksson: Why is port specified only for github's `uri.Authority`? Couldn't someone explicity specify port for Microsoft or Codeplex as well? Should we be comparing `uri.Host` instead?

src/NuGetGallery/Helpers/UriHelpers.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        public static bool IsGitProtocol(this Uri uri)
+        {
+            return uri.Scheme == Constants.GitRepository;
+        }
+
+        public static bool IsDomainWithHttpsSupport(this Uri uri)
+        {
+            return IsGitHubUri(uri) || IsCodeplexUri(uri) || IsMicrosoftUri(uri);
```
chenriksson: nit: If this is only caller for `IsCodeplexUri` and `IsMicrosoftUri`, consider making them private.

src/NuGetGallery/Helpers/HtmlExtensions.cs
```diff
@@ -65,17 +65,27 @@ public static IHtmlString PreFormattedText(this HtmlHelper self, string text)
...
                 {
                     if (match.Value.EndsWith(trimmedEntity))
                     {
                         // Remove trailing html entity from anchor URL
-                        var trimmedAnchorValue = match.Value.Substring(0, match.Value.Length - trimmedEntity.Length);
-                        return $"<a href=\"{trimmedAnchorValue}\" rel=\"nofollow\">{trimmedAnchorValue}</a>" + trimmedEntity;
+                        trimmedAnchorValue = match.Value.Substring(0, match.Value.Length - trimmedEntity.Length);
+                        trimmedEntityValue = trimmedEntity;
+
+                        break;
```
loic-sharma: Why do you `break;` here? What if the URL ends with something weird like `'-`?

src/NuGetGallery/Helpers/PackageHelper.cs
```diff
@@ -34,19 +34,30 @@ public static bool ShouldRenderUrl(string url, bool secureOnly = false)
...
-            return uri.Scheme == Uri.UriSchemeHttps;
-        }
+            Uri returnUri = null;
+            readyUriString = null;
 
-        public static bool IsHttpProtocol(this Uri uri)
-        {
-            return uri.Scheme == Uri.UriSchemeHttp;
-        }
+            if (Uri.TryCreate(uriString, UriKind.Absolute, out var uri))
```
loic-sharma: Can we invert the `Uri.TryCreate` conditional and `return false;` early to reduce the nesting here?

src/NuGetGallery/Helpers/PackageHelper.cs
```diff
@@ -34,19 +34,30 @@ public static bool ShouldRenderUrl(string url, bool secureOnly = false)
...
+            {
+                if (uri.IsHttpProtocol() && uri.IsDomainWithHttpsSupport())
+                {
+                    returnUri = uri.ToHttps();
+                }
+                else if (uri.IsHttpsProtocol() || uri.IsHttpProtocol())
+                {
+                    returnUri = uri;
```
loic-sharma: Could you add a comment as to why this case is necessary? :)

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -77,7 +77,7 @@ protected override void Load(ContainerBuilder builder)
 
             configuration.SecretInjector = secretInjector;
 
-            UrlExtensions.SetConfigurationService(configuration);
+             UrlHelperExtensions.SetConfigurationService(configuration);
```
loic-sharma: It looks like there's an extra space here

src/NuGetGallery/Helpers/UriHelpers.cs
skofman1: done

src/NuGetGallery/Helpers/PackageHelper.cs
skofman1: I return false if the scheme is not http/https as well.

src/NuGetGallery/Helpers/PackageHelper.cs
skofman1: added a comment

src/NuGetGallery/Helpers/HtmlExtensions.cs
skofman1: You are right, if the url ends with '- the method won't work. However, this is the previous behavior of the method and out of scope for this change.

src/NuGetGallery/Helpers/UriHelpers.cs
skofman1: done

PR #6414, opened 2018-09-11
Typo-squatting Integration

zhhyu:
Finish the integration of typo-squatting check service. 

Fix: https://github.com/NuGet/Engineering/issues/1596


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -323,6 +323,16 @@ protected override void Load(ContainerBuilder builder)
                 .As<ICertificateService>()
                 .InstancePerLifetimeScope();
 
+            builder.RegisterType<TyposquattingCheckService>()
+                .AsSelf()
+                .As<ITyposquattingCheckService>()
+                .SingleInstance();
```
joelverhagen: Why not `InstancePerLifetimeScope`?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -833,6 +835,8 @@ private static ActionResult GetActionResultOrNull(PackageValidationResult valida
                 case PackageValidationResultType.PackageShouldNotBeSigned:
                 case PackageValidationResultType.PackageShouldNotBeSignedButCanManageCertificates:
                     return new HttpStatusCodeWithBodyResult(HttpStatusCode.BadRequest, validationResult.Message);
+                case PackageValidationResultType.TyposquattingCheckFails:
```
joelverhagen: Can't you just use `PackageValidationResultType.Invalid` with message `Strings.TyposquattingCheckFails + validationResult.Message`?

src/NuGetGallery/Services/IPackageUploadService.cs
```diff
@@ -37,12 +37,14 @@ public interface IPackageUploadService
...
         /// <param name="currentUser">The current user.</param>
+        /// <param name="packageRegistration">The package registration info if it exists for uploaded package</param>
         /// <returns>The package validation result.</returns>
         Task<PackageValidationResult> ValidateAfterGeneratePackageAsync(
             Package package,
             PackageArchiveReader nuGetPackage,
             User owner,
-            User currentUser);
+            User currentUser,
+            PackageRegistration packageRegistration);
```
joelverhagen: Maybe call it `existingPackageRegistration`?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -195,6 +202,19 @@ private PackageValidationResult CheckRepositoryMetadata(PackageMetadata packageM
                 return result;
             }
 
+            if (_contentObjectService.TyposquattingConfiguration.IsCheckEnabled && packageRegistration == null)
+            {
+                if (_typosquattingCheckService.TryIsUploadedPackageIdTyposquatting(package.Id, owner, out string typosquattingCheckCollisionIds))
```
joelverhagen: Just curious, why isn't `_contentObjectService.TyposquattingConfiguration.IsCheckEnabled` part of `_typosquattingCheckService.TryIsUploadedPackageIdTyposquatting` implementation? This would minimize the dependencies of `PackageUploadService`.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -195,6 +202,19 @@ private PackageValidationResult CheckRepositoryMetadata(PackageMetadata packageM
...
+                {
+                    // TODO: save in the log metric for typosquatting collision pairs (typosquattingCheckResult). 
+                    // https://github.com/NuGet/Engineering/issues/1537
```
joelverhagen: We will leave `IsCheckEnabled` as `false` until this work is done, right?

src/NuGetGallery/Controllers/ApiController.cs
joelverhagen: Going a bit deeper, I added special enum values for `PackageShouldNotBeSigned` and `PackageShouldNotBeSignedButCanManageCertificates` because the messaging was going to be different in the web UI vs. command line. I don't think this panned out due to some other problems but in general we should not need a new enum if the message carries all the differentiation that you need.

src/NuGetGallery/Services/ITyposquattingCheckService.cs
```diff
@@ -13,6 +13,7 @@ public interface ITyposquattingCheckService
         /// </summary>
         /// <param name="uploadedPackageId"> The package ID of the uploaded package. We check the pacakge ID with the packages in the gallery for typo-squatting issue</param>
         /// <param name="uploadedPackageOwner"> The package owner of the uploaded package.</param>
-        bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner);
+        /// <param name="typosquattingCheckCollisionIds"> The return collision package Id if it exists</param>
+        bool TryIsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner, out string typosquattingCheckCollisionIds);
```
skofman1: The previous name was fine. I know this looks similar to the "Try" pattern, but this is not really the case.

src/NuGetGallery/Services/PackageUploadService.cs
zhhyu: The "IsCheckEnabled" is used for feature flag, which is read from Typosquatting-Configuration.json in blob storage.

src/NuGetGallery/Services/PackageUploadService.cs
zhhyu: Yes, both "IsCheckEnabled" and "IsBlockUsersEnabled" are set to false before we decide to turn on the button, and default values for both of them are set to false.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
zhhyu: Done

src/NuGetGallery/Controllers/ApiController.cs
zhhyu: Delete that and use Invalid message.

src/NuGetGallery/Services/IPackageUploadService.cs
zhhyu: Done

src/NuGetGallery/Services/PackageUploadService.cs
zhhyu: Move to Typo-squatting check service.

src/NuGetGallery/Services/ITyposquattingCheckService.cs
zhhyu: Change back to the previous name.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -33,8 +33,14 @@ public TyposquattingCheckService(ITyposquattingUserService typosquattingUserServ
             TyposquattingCheckListLength = _contentObjectService.TyposquattingConfiguration.PackageIdChecklistLength;
         }
               
-        public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner)
+        public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner, out string typosquattingCheckCollisionIds)
```
skofman1: this should be an array of strings. You don't know how the calling method will use this value, so you should leave the formatting to the caller.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -195,6 +199,11 @@ private PackageValidationResult CheckRepositoryMetadata(PackageMetadata packageM
                 return result;
             }
 
+            if (existingPackageRegistration == null && _typosquattingCheckService.IsUploadedPackageIdTyposquatting(package.Id, owner, out string typosquattingCheckCollisionIds))
+            {
+                return PackageValidationResult.Invalid(Strings.TyposquattingCheckFails + typosquattingCheckCollisionIds);
```
skofman1: The pattern we use for string formatting is 
string.Format(Strings.TyposquattingCheckFails, typosquattingCheckCollisionIds)

Please use the same patterns as the rest of the code.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -55,21 +61,35 @@ public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uplo
...
                 string normalizedPackageId = TyposquattingStringNormalization.NormalizeString(packageId);
                 if (TyposquattingDistanceCalculation.IsDistanceLessThanThreshold(uploadedPackageId, normalizedPackageId, threshold))
                 {
-                    collisionPackageIds.Add(packageId);
+                    collisionIds.Add(packageId);
                 }
             });
 
-            foreach (var packageId in collisionPackageIds)
+            List<string> typosquattingUserDoubleCheckIds = new List<string>();
```
skofman1: > List<string> [](start = 11, length = 13)

var

src/NuGetGallery/Services/ITyposquattingCheckService.cs
```diff
@@ -13,6 +13,7 @@ public interface ITyposquattingCheckService
         /// </summary>
         /// <param name="uploadedPackageId"> The package ID of the uploaded package. We check the pacakge ID with the packages in the gallery for typo-squatting issue</param>
         /// <param name="uploadedPackageOwner"> The package owner of the uploaded package.</param>
-        bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner);
+        /// <param name="typosquattingCheckCollisionIds"> The return collision package Id if it exists</param>
+        bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner, out string typosquattingCheckCollisionIds);
```
shishirx34: Change the `out` variable to be `List<string>` instead of `string`, the service should simply return the list of collisions. The controller should format it as it pleases.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -183,7 +186,8 @@ private PackageValidationResult CheckRepositoryMetadata(PackageMetadata packageM
             Package package,
             PackageArchiveReader nuGetPackage,
             User owner,
-            User currentUser)
+            User currentUser,
+            PackageRegistration existingPackageRegistration)
```
shishirx34: you do not need a `PackageRegistration` instead pass, `bool isNewPackageRegistration` as the parameter here.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -55,21 +61,35 @@ public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uplo
...
                 if (!_userTyposquattingService.CanUserTyposquat(packageId, uploadedPackageOwner.Username))
+                {
+                    typosquattingUserDoubleCheckIds.Add(packageId);
+                }
+            }
+
+            if (typosquattingUserDoubleCheckIds.Any())
+            {
+                // TODO: save in the log metric for typosquatting collision Ids (typosquattingCheckCollisionIds). 
+                // https://github.com/NuGet/Engineering/issues/1537
```
shishirx34: don't know your priorities, but, we should do this task immediately after this PR.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -55,21 +61,35 @@ public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uplo
...
+                    collisionIds.Add(packageId);
                 }
             });
 
-            foreach (var packageId in collisionPackageIds)
+            List<string> typosquattingUserDoubleCheckIds = new List<string>();
+            foreach (var packageId in collisionIds)
             {
+                // TODO: refactor user check services into one query
+                // https://github.com/NuGet/Engineering/issues/1684
```
shishirx34: its simple to do this task with this PR, I can help you.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -55,21 +61,35 @@ public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uplo
...
                 if (!_userTyposquattingService.CanUserTyposquat(packageId, uploadedPackageOwner.Username))
```
shishirx34: Why did you create a new service for user check? You should have simply merged it into this service.

src/NuGetGallery/Services/ITyposquattingCheckService.cs
shishirx34: I'd argue it is a `Try` pattern and better name could be `TryGetTyposquattingPackageIds`

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Changed to list.

src/NuGetGallery/Services/PackageUploadService.cs
zhhyu: Changed to string.Format.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done

src/NuGetGallery/Services/ITyposquattingCheckService.cs
zhhyu: Done

src/NuGetGallery/Services/PackageUploadService.cs
zhhyu: Changed to bool.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Ok

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Delete the user service for typo-squatting and integrate it into check service.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -55,27 +60,35 @@ public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uplo
...
+                .Where(pr => collisionIds.Contains(pr.Id))
+                .Select(pr => new { Id = pr.Id, Owners = pr.Owners.Select(x => x.Key).ToList() })
+                .ToList();
+
+            typosquattingCheckCollisionIds = collisionPackagesIdAndOwners
+                .Where(pio => !pio.Owners.Any(k => k == uploadedPackageOwner.Key))
+                .Select(pio => pio.Id)
+                .ToList();
+
+            var isUserAllowedTyposquatting = collisionPackagesIdAndOwners
```
skofman1: please add a really detailed comment around this scenario

zhhyu: Done

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -20,21 +20,27 @@ public class TyposquattingCheckService : ITyposquattingCheckService
...
+        private readonly IReservedNamespaceService _reservedNamespaceService;
 
-        public TyposquattingCheckService(ITyposquattingUserService typosquattingUserService, IEntityRepository<PackageRegistration> packageRegistrationRepository, IContentObjectService contentObjectService)
+        public TyposquattingCheckService(IContentObjectService contentObjectService, IPackageService packageService, IReservedNamespaceService reservedNamespaceService)
         {
-            _userTyposquattingService = typosquattingUserService ?? throw new ArgumentNullException(nameof(typosquattingUserService));
-            _packageRegistrationRepository = packageRegistrationRepository ?? throw new ArgumentNullException(nameof(packageRegistrationRepository));
             _contentObjectService = contentObjectService ?? throw new ArgumentNullException(nameof(contentObjectService));
+            _packageService = packageService ?? throw new ArgumentNullException(nameof(packageService));
+            _reservedNamespaceService = reservedNamespaceService ?? throw new ArgumentNullException(nameof(reservedNamespaceService));
```
shishirx34: why do you have the reserved namespace service in here? I think we do the reserved namespace check earlier in the upload, you do not need to repeat the check here. Also, if the ID matches a reserved namespace, you should also check if the owner is uploading this ID, if yes, then the user should be allowed to typosquat.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -20,21 +20,27 @@ public class TyposquattingCheckService : ITyposquattingCheckService
...
-              
-        public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner)
+
+        public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner, out List<string> typosquattingCheckCollisionIds)
         {
+            typosquattingCheckCollisionIds = new List<string>();
+            if (!_contentObjectService.TyposquattingConfiguration.IsCheckEnabled || _reservedNamespaceService.GetReservedNamespacesForId(uploadedPackageId).Count != 0)
+            {
+                return false;
+            }
```
shishirx34: remove the namespace check, see my comment above.

src/NuGetGallery/Strings.resx
```diff
@@ -989,4 +989,8 @@ Policy violations: {0}</value>
   <data name="UploadPackage_PackageContainsTooManyEntries" xml:space="preserve">
     <value>The package contains too many files and/or folders.</value>
   </data>
+  <data name="TyposquattingCheckFails" xml:space="preserve">
+    <value>Typosquatting check fails with {0} </value>
+    <comment>{0} is the collision Ids compared with the uploaded package Id.</comment>
```
loic-sharma: How about an error message like `The uploaded package's id is too similar to the already existing package 'XYZ'`?

src/NuGetGallery/Services/TyposquattingCheckService.cs
shishirx34: change to `.Any()`, ignore my previous comment.

loic-sharma:
>     public interface ITyposquattingCheckService

What do you think of the name `ITyposquattingService`?

---
Refers to: src/NuGetGallery/Services/ITyposquattingCheckService.cs:11 in b66cdaf. [](commit_id = b66cdaf3ad51ee0462d426d9970389e91ace8eb2, deletion_comment = False)

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -55,27 +62,44 @@ public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uplo
...
+            /// Example:
+            /// User "a" is uploading a package named "xyz", while "xyz" collides with existing packages "xyzz" (owned by "a", "b", "c"), "xyyz" (owned by "b"), "xxyz" (owned by "b", "c").
+            /// We will pass this package because "a" has the ownership of package "xyzz" even though this package Id collides with "xyyz" and "xxyz".
+            /// The "typosquattingCheckCollisionIds" will be saved as "xyyz" and "xxyz" because this package collides with these two packages which are not owned by "a", while "xyzz" will not be saved as "a" owns it.
+            /// </summary>
+            var isUserAllowedTyposquatting = collisionPackagesIdAndOwners
+                .Any(pio => pio.Owners.Any(k => k == uploadedPackageOwner.Key));
+
+            return _contentObjectService.TyposquattingConfiguration.IsBlockUsersEnabled && !isUserAllowedTyposquatting;
         }
```
loic-sharma: Should we add logs about decisions made by the `TyposquattingCheckService`?

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -55,27 +62,44 @@ public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uplo
...
             }
 
-            return false;
+            var collisionPackagesIdAndOwners = packageRegistrations
+                .Where(pr => collisionIds.Contains(pr.Id))
+                .Select(pr => new { Id = pr.Id, Owners = pr.Owners.Select(x => x.Key).ToList() })
+                .ToList();
+
+            typosquattingCheckCollisionIds = collisionPackagesIdAndOwners
+                .Where(pio => !pio.Owners.Any(k => k == uploadedPackageOwner.Key))
```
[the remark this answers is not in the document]
shishirx34: I dont think so.. we do not have any special allowances for organization members, neither with package ownership or namespace ownerships. This should stay consistent with that behavior.

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: The log metrics for decisions and performance will be committed very soon after this PR.

src/NuGetGallery/Strings.resx
zhhyu: We will discuss the messages for customers with PM.

src/NuGetGallery/Services/TyposquattingCheckService.cs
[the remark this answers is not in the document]
zhhyu: We will discuss how we should show the messages for the customers to change the user name to organization's name with PM.
Currently we keep the same check logic for organization's name as organization names and user names are in the same domain.

src/NuGetGallery/Services/TyposquattingCheckService.cs
```diff
@@ -55,27 +62,44 @@ public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uplo
...
+                .Select(pio => pio.Id)
+                .ToList();
+
+            /// <summary>
+            /// The following statement is used to double check whether the collision Id belongs to the same user who is uploading the package.
+            /// The current policy is that if the user has the ownership of any of the collision packages, we will pass the package.
+            /// The reason is that maybe this user is trying to update an existing package who is owned by himself.
```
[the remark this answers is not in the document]
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: Done!

src/NuGetGallery/Services/TyposquattingCheckService.cs
zhhyu: This one will be considered in the future.

src/NuGetGallery/Services/ITyposquattingCheckService.cs
```diff
@@ -13,6 +15,7 @@ public interface ITyposquattingCheckService
         /// </summary>
         /// <param name="uploadedPackageId"> The package ID of the uploaded package. We check the pacakge ID with the packages in the gallery for typo-squatting issue</param>
         /// <param name="uploadedPackageOwner"> The package owner of the uploaded package.</param>
-        bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner);
+        /// <param name="typosquattingCheckCollisionIds"> The return collision package Id if it exists</param>
```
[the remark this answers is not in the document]
zhhyu: Done!

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1636,7 +1635,9 @@ public virtual async Task<JsonResult> VerifyPackage(VerifyPackageRequest formDat
                         package,
                         nugetPackage,
                         owner,
-                        currentUser);
+                        currentUser,
+                        existingPackageRegistration == null);
```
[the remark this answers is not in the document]
zhhyu: Done!

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -723,7 +723,9 @@ private async Task<ActionResult> CreatePackageInternal()
                                 package,
                                 packageToPush,
                                 owner,
-                                currentUser);
+                                currentUser,
+                                packageRegistration == null);
```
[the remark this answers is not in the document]
zhhyu: Done!

src/NuGetGallery/Strings.resx
zhhyu: Done!

PR #6415, opened 2018-09-11
Add the Send email for symbol apis.

cristinamanum:
Add the Send email for symbol apis.

### DIFF
--- a/src/NuGetGallery.Core/Services/CoreMessageService.cs
+++ b/src/NuGetGallery.Core/Services/CoreMessageService.cs
@@ -80,15 +80,15 @@ namespace NuGetGallery.Services
             var warningMessagesPlaceholder = string.Empty;
             if (hasWarnings)
             {
-                subject = $"[{CoreConfiguration.GalleryOwner.DisplayName}] Symbol package published with warnings - {symbolPackage.Package.PackageRegistration.Id} {symbolPackage.Version}";
+                subject = $"[{CoreConfiguration.GalleryOwner.DisplayName}] Symbol package published with warnings - {symbolPackage.Id} {symbolPackage.Version}";
                 warningMessagesPlaceholder = Environment.NewLine + string.Join(Environment.NewLine, warningMessages);
             }
             else
             {
-                subject = $"[{CoreConfiguration.GalleryOwner.DisplayName}] Symbol package published - {symbolPackage.Package.PackageRegistration.Id} {symbolPackage.Version}";
+                subject = $"[{CoreConfiguration.GalleryOwner.DisplayName}] Symbol package published - {symbolPackage.Id} {symbolPackage.Version}";
             }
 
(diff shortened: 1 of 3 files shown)


src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -72,6 +72,46 @@ public async Task SendPackageAddedNoticeAsync(Package package, string packageUrl
...
+                mailMessage.From = CoreConfiguration.GalleryNoReplyAddress;
+
+                AddOwnersSubscribedToPackagePushedNotification(symbolPackage.Package.PackageRegistration, mailMessage);
+
+                if (mailMessage.To.Any())
+                {
+                    await SendMessageAsync(mailMessage);
+                }
+            }
+        }
```
chenriksson: Could the same template be used for source and symbols packages, to avoid duplicated code? It looks like everything is the same, except we prepend "Symbols " to "package" in some places.

e.g.:
```
public async Task SendSymbolPackageAddedNoticeAsync(...) {
    return SendPackageAddedNoticeAsync(..., isSymbols=true);
}

public async Task SendPackageAddedNoticeAsync(Package p, ..., bool isSymbols=false)
{
    var packageNoun = isSymbols ? "Symbols package" : "Package";
...
}
```

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -142,6 +182,52 @@ public async Task SendPackageValidationFailedNoticeAsync(Package package, Packag
...
+                mailMessage.From = CoreConfiguration.GalleryNoReplyAddress;
+
+                AddAllOwnersToMailMessage(symbolPackage.Package.PackageRegistration, mailMessage);
+
+                if (mailMessage.To.Any())
+                {
+                    await SendMessageAsync(mailMessage);
+                }
+            }
+        }
```
chenriksson: Same comment as above - can we reuse the existing code?

src/NuGetGallery.Core/Services/CoreMessageService.cs
```diff
@@ -210,6 +300,42 @@ public async Task SendValidationTakingTooLongNoticeAsync(Package package, string
...
+                mailMessage.From = CoreConfiguration.GalleryNoReplyAddress;
+
+                AddOwnersSubscribedToPackagePushedNotification(symbolPackage.Package.PackageRegistration, mailMessage);
+
+                if (mailMessage.To.Any())
+                {
+                    await SendMessageAsync(mailMessage);
+                }
+            }
+        }
```
chenriksson: Same comment as above - can we reuse the existing code?

cristinamanum: I would prefer to let the messages independent.

Issue #6256, opened 2018-08-02
[SymbolServer] Delete snupkg when package is deleted.


cristinamanum:
After delete the snupkg and the db will be cleaned

PR #6423, opened 2018-09-12
[NuGet symbol server]Delete snupkg with Package Hard/soft delete

shishirx34:
This PR adds support to backup and delete the snupkgs when a package is hard/soft deleted. 

Addresses #6256


shishirx34:
@cristinamanum - addressed your feedback, and also fixed a bug I discovered where you could upload a symbol package for a soft deleted package. Added tests.

PR #6427, opened 2018-09-17
[NuGet Symbol Server] Auditing - Gallery core changes

shishirx34:
Adding the needed enums with this PR, so that I can modify `Shims` which I can use in `NuGetGallery` project. 

I decided to keep auditing simple and simply reuse existing logic for auditing packages and authentication options, given that there isn't anything new needed for symbols package.

### DIFF
--- a/src/NuGetGallery.Core/Auditing/AuditedAuthenticatedOperationAction.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditedAuthenticatedOperationAction.cs
@@ -28,6 +28,6 @@ namespace NuGetGallery.Auditing
         /// <summary>
         /// Symbol package push was attempted by a non-owner of the package
         /// </summary>
-        SymbolPackagePushAttemptByNonOwner
+        SymbolsPackagePushAttemptByNonOwner
     }
 }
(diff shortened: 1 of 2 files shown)


src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
```diff
@@ -15,6 +15,8 @@ public enum AuditedPackageAction
         Edit,
         [Obsolete("Undo package edit functionality is being retired.")]
         UndoEdit,
-        Verify
+        Verify,
+        SymbolsCreate,
+        SymbolsDelete
```
chenriksson: There's `Delete` and `SoftDelete` for source packages... should we do the same for symbols?

src/NuGetGallery.Core/Auditing/AuditedAuthenticatedOperationAction.cs
```diff
@@ -23,6 +23,11 @@ public enum AuditedAuthenticatedOperationAction
         /// <summary>
         /// Login failed, user is an organization and should not have credentials.
         /// </summary>
-        FailedLoginUserIsOrganization
+        FailedLoginUserIsOrganization,
+
+        /// <summary>
+        /// Symbol package push was attempted by a non-owner of the package
+        /// </summary>
+        SymbolsPackagePushAttemptByNonOwner
```
chenriksson: nit: another approach could be to reuse the existing actions, but add a flag to AuditedPackageIdentifier that indicates whether it's the source or symbols package. That might be less work.

src/NuGetGallery.Core/Auditing/AuditedPackageAction.cs
shishirx34: nope.. only delete.

PR #6429, opened 2018-09-18
More cleanup of Elmah

cristinamanum:
More cleanup of Elmah

### DIFF
--- a/src/NuGetGallery.Core/Infrastructure/TableErrorLog.cs
+++ b/src/NuGetGallery.Core/Infrastructure/TableErrorLog.cs
@@ -191,7 +191,10 @@ namespace NuGetGallery.Infrastructure
             private void Obfuscate(Error error)
             {
                 error.User = string.Empty;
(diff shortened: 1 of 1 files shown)


src/NuGetGallery.Core/Infrastructure/TableErrorLog.cs
```diff
@@ -190,6 +190,12 @@ public override string Log(Error error)
 
             private void Obfuscate(Error error)
             {
+                error.User = string.Empty;
+                if (error.Form != null)
+                {
+                    error.Form.Clear();
```
chenriksson: Would help to see the elmah error... maybe if there were screenshot in https://github.com/NuGet/Engineering/issues/1713.

Elmah logs for postbacks with exceptions include user and all form data? Are User and Form the only properties that contain PII?

cristinamanum: Discussed offline

PR #6433, opened 2018-09-19
Allow tracing of specific uploads on back end.

ryuyu:
Adds a new header to allow us to track which client is making the upload.
This solves the strange behaviour on the upload progress bar when a single account is making multiple UI uploads.

@joelverhagen @loic-sharma


src/NuGetGallery/Authentication/AsyncFileUpload/AsyncFileUploadModule.cs
```diff
@@ -50,8 +51,21 @@ private void PostAuthorizeRequest(object sender, EventArgs e)
             string boundary = "--" + contentType.Substring(boundaryIndex + 9);
             var requestParser = new AsyncFileUploadRequestParser(boundary, request.ContentEncoding);
 
+            var headers = request.Headers;
+            string uploadTracingKey;
+            try
+            {
+                uploadTracingKey = headers[CoreConstants.UploadTracingKeyHeaderName];
+                Guid.Parse(uploadTracingKey);
```
shishirx34: don't you need to reassign this value to `uploadTracingKey` after parsing? I am assuming you want to parse here to remove trailing/leading spaces?

src/NuGetGallery/Authentication/AsyncFileUpload/AsyncFileUploadModule.cs
```diff
@@ -50,8 +51,21 @@ private void PostAuthorizeRequest(object sender, EventArgs e)
...
+            }
+            catch (Exception ex) when (ex is FormatException || ex is KeyNotFoundException)
+            {
+                uploadTracingKey = Guid.Empty.ToString();
+            }
+
             var progress = new AsyncFileUploadProgress(request.ContentLength);
-            _cacheService.SetProgress(username, progress);
+            var uploadKey = username + uploadTracingKey;
```
shishirx34: nit: may be use a `-` separator for easy reading?

src/NuGetGallery/Authentication/AsyncFileUpload/AsyncFileUploadModule.cs
shishirx34: Ah, I see you are checking for formatting, still a good idea to reassign after parsing. Upto you.

PR #6478, opened 2018-09-24
[NuGet Symbol Server] Gallery upload - Refactoring

shishirx34:
This is a preparation PR for adding support to gallery for symbols package upload.

I have refactored code out from the `ApiController` for the symbols upload, so that we can reuse the code in `PackagesController` and keep common code in single place. Also, as part of this I had to refactor/re-write some tests for the coverage. I do not touch the Package upload flow(which should also be refactored) as part of this PR, since its a lot more refactor and out of scope for the symbol upload flow. 

Also, refactored some part of `PackagesController` to reuse upload page in gallery. In subsequent PRs I will be doing additive changes to enhance this controller for symbols upload.

Apologies for the big PR(refactoring is a pain)

### DIFF
--- a/src/NuGetGallery.Core/PackageMetadataExtensions.cs
+++ b/src/NuGetGallery.Core/PackageMetadataExtensions.cs
@@ -15,6 +15,12 @@ namespace NuGetGallery
         private const string SymbolPackageTypeName = "SymbolsPackage";
         private static readonly ClientPackageType SymbolPackageType = new ClientPackageType(SymbolPackageTypeName, ClientPackageType.EmptyVersion);
 
+        /// <summary>
+        /// The package is a symbol package, if and only if it has metadata 
+        /// element of type <see cref="SymbolPackageTypeName"/> and only that element in package types.
+        /// </summary>
+        /// <param name="metadata"><see cref="PackageMetadata"/> for package</param>
+        /// <returns>True if the package is a symbols package</returns>
         public static bool IsSymbolPackage(this PackageMetadata metadata)
         {
             var packageTypes = metadata.GetPackageTypes();
--- a/src/NuGetGallery.Core/PackageMetadataExtensions.cs
+++ b/src/NuGetGallery.Core/PackageMetadataExtensions.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGetGallery.Packaging;
+using System.Linq;
+using ClientPackageType = NuGet.Packaging.Core.PackageType;
+
+namespace NuGetGallery
+{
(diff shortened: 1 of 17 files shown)


src/NuGetGallery.Core/PackageMetadataExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+        private const string SymbolPackageTypeName = "SymbolsPackage";
+        private static readonly ClientPackageType SymbolPackageType = new ClientPackageType(SymbolPackageTypeName, ClientPackageType.EmptyVersion);
+
+        public static bool IsSymbolPackage(this PackageMetadata metadata)
+        {
+            var packageTypes = metadata.GetPackageTypes();
+            return packageTypes.Any()
+                && packageTypes.Count() == 1
+                && packageTypes.First() == SymbolPackageType;
+        }
```
cristinamanum: Why not packageTypes.Where.(..).Any()?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -244,155 +250,161 @@ public virtual async Task<JsonResult> UploadPackage(HttpPostedFileBase uploadFil
...
                     }
-                }
 
-                await _uploadFileService.SaveUploadFileAsync(currentUser.Key, uploadStream);
+                    return Json(HttpStatusCode.Conflict, new[] { message });
+                }
             }
 
+            await _uploadFileService.SaveUploadFileAsync(currentUser.Key, uploadStream);
+            uploadStream.Dispose();
```
skofman1: uploadStream  should be disposed by the caller of this method (UploadPackage(HttpPostedFileBase uploadFile)) as part of finally block

src/NuGetGallery/Helpers/ZipArchiveHelpers.cs
```diff
@@ -0,0 +1,34 @@
...
+                if (entryInTheFuture != null)
+                {
+                    entry = entryInTheFuture;
+                    return true;
+                }
+            }
+
+            return false;
+        }
+    }
```
cristinamanum: I think that this is an old method, but can we add few comments about its purpose? What is are found ore than one entries in the future?  It does not matter which is returned ?

src/NuGetGallery/Services/SymbolPackageValidationResult.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        /// <summary>
+        /// There is no corresponding nuget package available for the uploaded symbols package.
+        /// </summary>
+        MissingPackage,
+
+        /// <summary>
+        /// There is a symbols package already present, pending validations.
+        /// </summary>
+        SymbolsPackageExists
```
cristinamanum: Why not SymbolPackagePendingValidation ?

PR #6482, opened 2018-09-25
log the original uploaded package Id rather than normalized one

zhhyu:
In log metrics, we should log the original uploaded package Id rather than normalized one.


joelverhagen:
This should have corresponding unit test changes.

PR #6489, opened 2018-09-27
Add configuration value to redirect specific curated feeds to main feed

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6472. Depends on https://github.com/NuGet/NuGetGallery/pull/6486.

If a curated feed name is in this new list, it will not even be checked for existence in the DB. It will just use all packages. I considered an HTTP redirect or calling the `ODataV2FeedController` but chose to take this approach since it has the best change of not breaking existing clients.

### DIFF
--- a/tests/NuGetGallery.Facts/Controllers/ODataFeedControllerFactsBase.cs
+++ b/tests/NuGetGallery.Facts/Controllers/ODataFeedControllerFactsBase.cs
@@ -24,7 +24,7 @@ namespace NuGetGallery.Controllers
     public abstract class ODataFeedControllerFactsBase<TController>
         where TController : NuGetODataController
     {
-        private const string _siteRoot = "https://nuget.localtest.me";
+        protected const string _siteRoot = "https://nuget.localtest.me";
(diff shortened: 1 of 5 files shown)


tests/NuGetGallery.Facts/Controllers/ODataFeedControllerFactsBase.cs
```diff
@@ -217,7 +222,7 @@ private static ODataQueryContext CreateODataQueryContext<TFeedPackage>()
             return new ODataQueryContext(oDataModelBuilder.GetEdmModel(), typeof(TFeedPackage));
         }
 
-        private static async Task<dynamic> GetValueFromQueryResult<TEntity>(QueryResult<TEntity> queryResult)
+        protected static async Task<dynamic> GetValueFromQueryResult<TEntity>(QueryResult<TEntity> queryResult)
```
chenriksson: Is this change necessary?

joelverhagen: Nope, nice catch.

PR #6495, opened 2018-09-28
Emails should not be sent synchronously

xavierdecoster:
Work-in-progress PR, but in *working* state.

This PR introduces async email messaging to the NuGetGallery, leveraging the new ServerCommon `NuGet.Services.Messaging` library (for which an official release needs to be created still, after which I can upgrade the nupkg dependencies here).

To enable it, configure the following settings in `web.config`:

```xml
    <add key="Gallery.AsynchronousEmailServiceEnabled" value="true"/>
    <add key="AzureServiceBus.EmailPublisher.ConnectionString" value=""/>
    <add key="AzureServiceBus.EmailPublisher.TopicName" value=""/>
```

By default, `Gallery.AsynchronousEmailServiceEnabled` will be `false` or even missing, returning the NuGetGallery email functionality into its current SMTP-mode (still using the Markdown infrastructure we already had).

The new `AsynchronousEmailService` will construct both *plain-text* and *html* email message bodies (plain-text is a fallback for email clients).

For this purpose, I had to do some refactoring in the way email message bodies are constructed today, by introducing an `IEmailMessageBodyBuilder` interface (3 concrete types: `PlainTextEmailMessageBodyBuilder`, `MarkdownEmailMessageBodyBuilder` and `HtmlEmailMessageBodyBuilder`), and a central `EmailSubjectBuilder`.

Some jobs are consuming the `ICoreMessageService` type from `NuGetGallery.Core` to reuse some of the email sending functionality, so I replicated this by introducing a `CoreAsynchronousEmailMessageService` base type. This way, the consuming code can simply change DI and config to onboard to async messaging.


src/NuGetGallery.Core/Infrastructure/mail/MarkdownEmailBodyBuilder.cs
```diff
@@ -0,0 +1,414 @@
...
+using System.Net.Mail;
+using System.Text;
+using NuGet.Services.Validation;
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    /// <summary>
+    /// Builds email-body messages using Markdown syntax.
+    /// </summary>
+    public class MarkdownEmailBodyBuilder : PlainTextEmailBodyBuilder
```
xavierdecoster: I let `MarkdownEmailBodyBuilder` inherit from the plain-text one, and override as needed. Turns out we don't have *that* many places that rely on Markdown (mostly the emails containing hyperlinks).

joelverhagen:
Especially considering this PR is in progress, could we split this up into two or more PRs? 4,000 added LOC is a lot 😓

scottbommarito:
We didn't have a plaintext email fallback before--why do we need to add one now? Can we proceed with this PR and the asynchronous email service without implementing plaintext fallback? It seems really redundant that we have up to 3 copies of every email text. I think somebody could change one implementation without changing the others and we could very easily end up with mails that differ between the two, and it doesn't look like you added tests to prevent this from happening.

Also, why do we need Markdown at all if we have HTML? I don't think it's necessary for us to have separate creation processes for the asynchronous mailing and the synchronous mailing. Can't we just reuse the existing mail message building service, but have the `Send` method do different things based on whether or not the user has enabled asynchronous mails?

xavierdecoster:
@joelverhagen I don't see how to split this easily... :( The bulk of the changed LOC is moving around the message body construction, which doesn't make sense to split up one by one.

@scottbommarito we keep the Markdown because that's the SMTP fallback. Didn't want to change that as part of this PR... but we could decide to kill that and convert the SMTP code path to HTML moving forward. It would make sense to do so... but don't be mistaken about the plain-text fallback. The MarkdownMailer handles this for us behind the scenes... So we'd still need both plain-text and html.

src/NuGetGallery/Security/RequirePrefixReservationPolicy.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+
+namespace NuGetGallery.Security
+{
+    /// <summary>
+    /// User security policy that requires packages pushed by this user to have a prefix reservation.
+    /// </summary>
+    public class RequirePrefixReservationPolicy : SecurityPolicyHandler<UserSecurityPolicyEvaluationContext>
```
skofman1: why is this part of the PR?

src/NuGetGallery.Core/Infrastructure/mail/IEmailBodyBuilder.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Net.Mail;
+using NuGet.Services.Validation;
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    public interface IEmailBodyBuilder
```
skofman1: the construction of a single e-mail is separated between multiple classes and interfaces: IEmailBodyBuilder, EmailSubjectBuilder, etc. A more readable approach would be to have a class per e-mail that creates all the artifacts this e-mail needs. This way, you can also avoid bugs happening when you modify one class and not the others.

skofman1:
I recommend splitting this PR into two pieces: 
1. Refactoring of existing code and introducing new interfaces. This can be merged immediately with no dependencies on async work and processing job.
2. Introducing service bus integration and async queuing. 

What do you think?

scottbommarito:
@xavierdecoster - if markdown mailer handles the plaintext fallback itself, can we write something that converts the HTML emails into plaintext?

Or, better yet because it requires no mail changes, could we keep the emails in markdown, use a markdown transformer to change the emails into HTML, and then keep the untransformed markdown as the plaintext fallback?

src/NuGetGallery.Core/Infrastructure/mail/IEmailBodyBuilder.cs
scottbommarito: I love this idea.

```
public interface IEmailBuilder
{
    string GetSubject();
    string GetBody();
    string GetTo();
    ...etc...
}
```

Then the mail service just needs to accept an instance of `IEmailBuilder`.

src/NuGetGallery/Security/RequirePrefixReservationPolicy.cs
xavierdecoster: Old file left-behind on disk that accidentally made it into this PR. Removed!

src/NuGetGallery.Core/Infrastructure/mail/IEmailBodyBuilder.cs
xavierdecoster: The tricky part is that each e-mail builder needs a different set of parameters...
Great idea though! I'll see what I can do :)

src/NuGetGallery.Core/Infrastructure/mail/EmailSubjectBuilder.cs
```diff
@@ -0,0 +1,225 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    public static class EmailSubjectBuilder
```
agr: Please, add documentation for new classes/methods.

src/NuGetGallery.Core/Infrastructure/mail/ValidationIssueExtensions.cs
```diff
@@ -0,0 +1,116 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.Validation;
+using NuGet.Services.Validation.Issues;
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    internal static class ValidationIssueExtensions
```
agr: This class methods are called `Parse*`, but they does not seem to parse anything. Some other names might be better.
`ToPlainTextString`, `ToMarkdownString` and `ToHtmlString` maybe?

xavierdecoster:
Created a separate PR that handles refactoring of the code as suggested: https://github.com/NuGet/NuGetGallery/pull/6505

src/NuGetGallery.Core/Infrastructure/mail/ValidationIssueExtensions.cs
xavierdecoster: Fixed.

xavierdecoster:
Closing this one in favor of https://github.com/NuGet/NuGetGallery/pull/6555

PR #6498, opened 2018-09-28
Fix delete registration checkbox on delete page

joelverhagen:
This should align with the admin panel page:

https://github.com/NuGet/NuGetGallery/blob/3364ff692b86149c275f45da3572b5b44b9c9717/src/NuGetGallery/Areas/Admin/Views/Delete/Index.cshtml#L128-L134


src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -152,13 +152,14 @@
                 window.location.href = this.value;
             });
 
-            $('#DeletePackagesRequest_SoftDelete').change(function(e) {
+            $('#SoftDelete').change(function(e) {
```
scottbommarito: Could we move these functions to a shared place (a function in the `common.js` file maybe) so the likelihood of these functions being inconsistent in the future is much less?

joelverhagen: Sure!

joelverhagen:
Addresses https://github.com/NuGet/NuGetGallery/issues/6501.

PR #6503, opened 2018-09-29
[NuGet Symbol Server] Gallery upload for Symbols package

shishirx34:
This PR adds support for the uploading symbols package to the gallery. 

I refactored a bit more code within the `PackagesController`, for convenience. Also added tests.

I will verify this branch on `Dev` env before merging it in. Also, added test cases for the new things. I have not refactored too much with regards to the `package` upload workflow itself, so as to not break the primary flow upload flow. @diverdan92 - was cool with the UI changes. If you have any UI related feedback, please feel free to tag @diverdan92 as well into the conversation.

I will add screenshots subsequently, when I validate on DEV. Apologies for the size of the PR, again its mostly refactoring work than anything new.

NOTE: **Do not forget to expand `PackagesController.cs`,** github seems to hide the large file diffs.


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -185,41 +188,91 @@ public virtual async Task<ActionResult> UploadPackage()
...
-                    verifyRequest.Warnings.AddRange(validationResult.Warnings);
+        private async Task<ActionResult> UploadSymbolsPackageInternal(SubmitPackageRequest model,
+            Stream uploadStream,
+            PackageArchiveReader packageArchiveReader,
+            PackageMetadata packageMetadata,
+            User currentUser)
+        {
+            var symbolsPackageValidationResult = await _symbolPackageUploadService.ValidateUploadedSymbolsPackage(uploadStream, currentUser);
+            var uploadResult = GetJsonResultOrNull(symbolsPackageValidationResult);
+            if (uploadResult != null)
```
cristinamanum: Does this mean that it was an error during validation? If yes maybe add a comment?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -185,41 +188,91 @@ public virtual async Task<ActionResult> UploadPackage()
...
+            {
+                TempData["Message"] = symbolsPackageValidationResult.Message;
+                return View(model);
+            }
 
-                    model.InProgressUpload = verifyRequest;
-                }
+            var packageForUploadingSymbols = symbolsPackageValidationResult.Package;
```
cristinamanum: Can the symbolsPackageValidationResult be null?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -185,41 +188,91 @@ public virtual async Task<ActionResult> UploadPackage()
...
+            var existingPackageRegistration = packageForUploadingSymbols.PackageRegistration;
+            var id = existingPackageRegistration.Id;
+            var version = packageForUploadingSymbols.NormalizedVersion;
+
+            IEnumerable<User> accountsAllowedOnBehalfOf = Enumerable.Empty<User>();
+            bool isAllowed = ActionsRequiringPermissions.UploadNewPackageVersion.CheckPermissionsOnBehalfOfAnyAccount(currentUser, existingPackageRegistration, out accountsAllowedOnBehalfOf) == PermissionsCheckResult.Allowed;
```
cristinamanum: nit: shorter line

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -185,41 +188,91 @@ public virtual async Task<ActionResult> UploadPackage()
...
+            if (!isAllowed)
+            {
```
cristinamanum: If not allowed why set it as allowed?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -185,41 +188,91 @@ public virtual async Task<ActionResult> UploadPackage()
...
+                TempData["Message"] = symbolsPackageValidationResult.Message;
+                return View(model);
+            }
 
-                    model.InProgressUpload = verifyRequest;
-                }
+            var packageForUploadingSymbols = symbolsPackageValidationResult.Package;
+            var existingPackageRegistration = packageForUploadingSymbols.PackageRegistration;
+            var id = existingPackageRegistration.Id;
+            var version = packageForUploadingSymbols.NormalizedVersion;
```
cristinamanum: Is the version used anywhere?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -285,6 +345,43 @@ public virtual async Task<JsonResult> UploadPackage(HttpPostedFileBase uploadFil
             }
         }
 
+        private async Task<JsonResult> UploadSymbolsPackageInternal(PackageArchiveReader packageArchiveReader, Stream uploadStream, NuspecReader nuspec, PackageMetadata packageMetadata)
+        {
+            var currentUser = GetCurrentUser();
+
```
cristinamanum: Can the user be null?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -285,6 +345,43 @@ public virtual async Task<JsonResult> UploadPackage(HttpPostedFileBase uploadFil
...
+            var packageForUploadingSymbols = symbolsPackageValidationResult.Package;
+            var existingPackageRegistration = packageForUploadingSymbols.PackageRegistration;
+            var id = existingPackageRegistration.Id;
+            var version = packageForUploadingSymbols.NormalizedVersion;
+
+            // Evaluate the permissions for user on behalf of any account possible, since the user 
+            // could change the ownership before submitting the package.
+            if (ActionsRequiringPermissions.UploadNewPackageVersion.CheckPermissionsOnBehalfOfAnyAccount(
+                currentUser, existingPackageRegistration, out accountsAllowedOnBehalfOf) != PermissionsCheckResult.Allowed)
```
cristinamanum: If the user is not allowed to submit why we return the Id not available? Is this the single way for the user not being allowed to submit?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -285,6 +345,43 @@ public virtual async Task<JsonResult> UploadPackage(HttpPostedFileBase uploadFil
...
+            var symbolsPackageValidationResult = await _symbolPackageUploadService.ValidateUploadedSymbolsPackage(uploadStream, currentUser);
+            var uploadResult = GetJsonResultOrNull(symbolsPackageValidationResult);
+            if (uploadResult != null)
+            {
+                return uploadResult;
+            }
+
+            var packageForUploadingSymbols = symbolsPackageValidationResult.Package;
+            var existingPackageRegistration = packageForUploadingSymbols.PackageRegistration;
+            var id = existingPackageRegistration.Id;
```
cristinamanum: Are id and version used anywhere?

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -290,7 +290,9 @@
                 window.nuget.configureExpanderHeading("submit-package-form");
             }
 
-            bindReadMeData(model);
+            if (model === null || !model.IsSymbolsPackage) {
+                bindReadMeData(model);
```
cristinamanum: Will this call be ok on null model?

src/NuGetGallery/Controllers/PackagesController.cs
shishirx34: This is the same as that of Package synchronous validations, I think it is self explanatory. If you can suggest better method or variable names to make it readable, I am open.

src/NuGetGallery/Controllers/PackagesController.cs
shishirx34: No. Also, the `.Package` will not be null if code reaches here.

src/NuGetGallery/Controllers/PackagesController.cs
shishirx34: The above checks if `currentUser` is allowed to push as anyone else(org?). `CheckPermissionsOnBehalfOfAnyAccount`. If the user is not allowed to upload as anyone else, then the only allowed list of owners is the `currentUser`.

src/NuGetGallery/Controllers/PackagesController.cs
shishirx34: Yes. https://github.com/NuGet/NuGetGallery/blob/symbols-gallery-upload/src/NuGetGallery/Services/SymbolPackageUploadService.cs#L81

shishirx34: id/version not used, removed it.

src/NuGetGallery/Controllers/PackagesController.cs
shishirx34: nope.. we have `UIAuthorize` attribute on the entry points to ensure.

src/NuGetGallery/Controllers/PackagesController.cs
shishirx34: This checks if the user is allowed to push to an existing package for a new version. If this returns as not allowed it means that the package is owned by someone else and this user is trying to push to a package which already exists. Hence Id not available.

src/NuGetGallery/Scripts/gallery/async-file-upload.js
shishirx34: Yes

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1497,253 +1604,381 @@ public virtual async Task<JsonResult> VerifyPackage(VerifyPackageRequest formDat
...
+                            packageArchiveReader,
+                            packageMetadata,
+                            currentUser,
+                            owner);
                     }
+                }
+            }
+            catch (Exception ex)
+            {
+                _telemetryService.TrackPackagePushFailureEvent(id: null, version: null);
```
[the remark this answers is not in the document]
shishirx34: Yes nulls are handled correctly, this is an exception case. Shouldn't happen ideally, while unlikely it is possible to happen. I added this check to satisfy a test case which was verifying if unknown exception logs an event with null Id/version. And yes, for symbols and packages there will be additional null events for packages, but I think its very less likely so fine to ignore.

shishirx34:
validated the new changes in DEV env. The Functional and E2E tests passed just fine. new changes look good, except for one minor issue that I have fixed in recent commit.

PR #2358, opened 2015-02-09
Escape underscores in confirmation urls for new accounts or changes emai...

Niels-V:
...l addresses #2357

Markdown sees the underscore as italics indicator, so underscores are
stripped in the message. This prevents cut and pasting of the address or
the use of text only email readers.

PR #6505, opened 2018-10-02
Refactoring of email messaging infrastructure

xavierdecoster:
This PR only refactors the code so that it is easier to extend the email infrastructure.


src/NuGetGallery.Core/Infrastructure/mail/EmailRecipients.cs
```diff
@@ -0,0 +1,32 @@
...
+            IReadOnlyList<MailAddress> bcc = null,
+            IReadOnlyList<MailAddress> replyTo = null)
+        {
+            To = to ?? throw new ArgumentNullException(nameof(to));
+            CC = cc ?? new List<MailAddress>();
+            Bcc = bcc ?? new List<MailAddress>();
+            ReplyTo = replyTo ?? new List<MailAddress>();
+        }
+
+        public IReadOnlyList<MailAddress> To { get; private set; }
```
agr: Do you need `private set` here and below?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -421,6 +376,56 @@ protected override void Load(ContainerBuilder builder)
             ConfigureAutocomplete(builder, configuration);
         }
 
+        private static void RegisterMessagingService(ContainerBuilder builder, ConfigurationService configuration)
+        {
+            MailSender mailSenderFactory()
+            {
+                var settings = configuration;
+                if (settings.Current.SmtpUri != null && settings.Current.SmtpUri.IsAbsoluteUri)
```
agr: Are we OK with extending the lifetime of the `configuration` object here by referencing it from the local function? Can we get it from the `IComponentContext` (which is not currently passed into this function but can be)?

src/NuGetGallery.Core/Infrastructure/mail/ValidationIssueExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+            switch (validationIssue.IssueCode)
+            {
+                case ValidationIssueCode.PackageIsSigned:
+                    return $"This package could not be published since it is signed. We do not accept signed packages at this moment. To be notified about package signing and more, watch our Announcements ({announcementsUrl}) page or follow us on Twitter ({twitterUrl}).";
+                case ValidationIssueCode.ClientSigningVerificationFailure:
+                    var clientIssue = (ClientSigningVerificationFailure)validationIssue;
+                    return clientIssue != null
+                        ? $"{clientIssue.ClientCode}: {clientIssue.ClientMessage}"
+                        : "This package's signature was unable to be verified.";
+                case ValidationIssueCode.PackageIsZip64:
```
agr: That does not sound right to me.

src/NuGetGallery/Infrastructure/Mail/EmailRecipientsWithPermission.cs
```diff
@@ -0,0 +1,79 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Mail;
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    public class EmailRecipientsWithPermission
```
scottbommarito: This seems like an improper use of classes.

I think you should only have a single `EmailRecipients` class, no interface, that just contains a constructor and `To`, `CC`, `Bcc`, and `ReplyTo` fields.

You can then have another `static` class with methods like `GetAddressesWithPermission` and `GetOwnersSubscribedToPackagePushedNotification` which can be used to create an instance of `EmailRecipients`.

src/NuGetGallery/Infrastructure/Mail/Messages/PackageOwnershipRequestInitiatedMessage.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        protected override string GetPlainTextBody()
+        {
+            return string.Format(
+                CultureInfo.CurrentCulture,
+                $@"The user '{_requestingOwner.Username}' has requested that user '{_newOwner.Username}' be added as an owner of the package '{_packageRegistration.Id}'.
+
+To cancel this request:
+{_rawCancellationUrl}
```
scottbommarito: Every plaintext body I've seen is just the markdown with links reformatted. Can we do this automatically and not store the message body twice?

src/NuGetGallery/Infrastructure/Mail/Requests/OrganizationMembershipRequest.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                if (ConfirmationUrl == null)
+                {
+                    return null;
+                }
+
+                return HttpUtility.UrlDecode(ConfirmationUrl).Replace("_", "\\_");
+            }
+        }
+        public string RejectionUrl
```
scottbommarito: nit: add line above

src/NuGetGallery/Infrastructure/Mail/Requests/OrganizationMembershipRequest.cs
```diff
@@ -0,0 +1,43 @@
...
+        public string ConfirmationUrl
+        {
+            get
+            {
+                if (ConfirmationUrl == null)
+                {
+                    return null;
+                }
+
+                return HttpUtility.UrlDecode(ConfirmationUrl).Replace("_", "\\_");
```
scottbommarito: Can you explain what the purpose of this decode call is? Might be useful to add a comment as well.

src/NuGetGallery/Services/MarkdownMessageService.cs
```diff
@@ -0,0 +1,417 @@
...
+            IMailSender mailSender,
+            IAppConfiguration config,
+            ITelemetryService telemetryService)
+            : base(mailSender, config)
+        {
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+            _smtpUri = config.SmtpUri?.Host;
+        }
+
+        public async Task ReportAbuseAsync(ReportPackageRequest request)
```
scottbommarito: It seems really unnecessary that we have a clone of this method for every single `IEmailBuilder` implementation. I think if we used some sort of generics `IEmailBuilder<TContext>` we could make this method and generic and cut out a TON of duplicate code.

src/NuGetGallery/Strings.resx
```diff
@@ -990,4 +964,56 @@ Policy violations: {0}</value>
     <value>The uploaded package's id is too similar to the already existing packages: {0} </value>
     <comment>{0} is the existing collision Ids compared with the uploaded package Id.</comment>
   </data>
+  <data name="Emails_ForgotPassword_MarkdownBody" xml:space="preserve">
```
scottbommarito: IMO it's confusing that we have some email text in `Strings` (but not all). Would it be much work to move all of the emails to strings right now? Should we move this email text OUT of strings?

src/NuGetGallery.Core/Infrastructure/mail/EmailRecipients.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6505/commits/9c8f6d0f46368433a2956f456538842d6a3c4a74

xavierdecoster:
Please note that I didn't change any wording in the messages, or where they are stored (in resource file, inline, etc.). Same for Url decoding. It's good that this surfaces now as part of this PR, so we can consolidate. 

Also, please bare in mind that there's an additional abstraction layer which may seem overkill at this point, until I add the HTML and async messaging infrastructure, which will require it...

src/NuGetGallery/Strings.resx
xavierdecoster: Agreed, this is not consistent. But it is the way it is today, even without this PR :)
This email mess really needs a clean-up and I'd prefer all of the strings to end up in resource files, but we have a timeline here... Should this be part of this effort? or should we create a separate issue and postpone this consolidation?

src/NuGetGallery/Infrastructure/Mail/Requests/OrganizationMembershipRequest.cs
xavierdecoster: I didn't invent or add this, it was there before this PR, without comments. So my best guess is that it is a way to make the Url display in a user-friendly way (e.g. replace `%20` with an actual whitespace character when rendering the Url in email).

src/NuGetGallery/Infrastructure/Mail/Requests/OrganizationMembershipRequest.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6505/commits/a3fbdf702053898e413e0a857dd4c09564fe0189

src/NuGetGallery.Core/Infrastructure/mail/ValidationIssueExtensions.cs
xavierdecoster: Didn't change any wording, but fixed as part of https://github.com/NuGet/NuGetGallery/pull/6505/commits/bedd72247d16639836bf76139e2f5c900f519519

xavierdecoster:
Please correct me if I'm wrong, but didn't we have the intention of getting rid of the whole AnglicanGeek.MarkdownMailer dependency at some point? If so, I'd like to rely as little as possible on any of the Markdown magic, and prefer explicit over implicit in the code for now, and revisit when we have the async HTML implementation.

src/NuGetGallery/Services/MarkdownMessageService.cs
xavierdecoster: This class transforms the input (context required for the particular message, which is different for each) into output (the actual email message), and sends it (which may behave different for each). 

And even though it may seem like a lot of duplicate code at first, there are differences in behavior for some (e.g. cc sender vs sending separate email to sender to not disclose email address, etc.).

In addition, for async messaging, we won't be constructing `System.Net.Mail.EmailMessage` objects nor will we use `AnglicanGeek.MarkdownMailer.IMailSender` to send it. That's why I can't expose any actual *message* or *sender* types in the interface of this message service.

I'll give it another iteration and see if I can further simplify and reduce this particular code.

src/NuGetGallery/Infrastructure/Mail/EmailRecipientsWithPermission.cs
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    if (!member.EmailAllowed)
+                    {
+                        continue;
+                    }
+
+                    recipients.Add(member.ToMailAddress());
+                }
+            }
+            else
```
chenriksson: could simplify this block:
```
else if (user.EmailAllowed)
{
  recipients.Add(user.ToMailAddress());
}
return recipients;
```

src/NuGetGallery/Infrastructure/Mail/EmailRecipientsWithPermission.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (action == null)
+            {
+                throw new ArgumentNullException(nameof(action));
+            }
+
+            To = AddAddressesWithPermission(user, action);
+
+            CC = cc ?? new List<MailAddress>();
+            Bcc = bcc ?? new List<MailAddress>();
+            ReplyTo = replyTo ?? new List<MailAddress>();
```
chenriksson: Are these empty allocations necessary?

src/NuGetGallery/Infrastructure/Mail/Messages/ApiKeyAddedMessage.cs
```diff
@@ -0,0 +1,56 @@
...
+        protected override string GetMarkdownBody()
+        {
+            return GetPlainTextBody();
+        }
+
+        protected override string GetPlainTextBody()
+        {
+            return string.Format(
+                CultureInfo.CurrentCulture,
+                Strings.Emails_ApiKeyAdded_Body,
```
chenriksson: Why do we use resources in some cases, and inline for others? Is there a chance to clean up this inconsistency?

src/NuGetGallery.Core/Infrastructure/mail/ConfirmationEmailBuilder.cs
```diff
@@ -0,0 +1,32 @@
...
+        protected ConfirmationEmailBuilder(
+            ICoreMessageServiceConfiguration configuration,
+            User user,
+            string confirmationUrl)
+        {
+            Configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+            User = user ?? throw new ArgumentNullException(nameof(user));
+            IsOrganization = user is Organization;
+            RawConfirmationUrl = confirmationUrl ?? throw new ArgumentNullException(nameof(confirmationUrl));
+            ConfirmationUrl = HttpUtility.UrlDecode(confirmationUrl).Replace("_", "\\_");
```
chenriksson: This `UrlDecode` logic is in multiple classes. Maybe it's worth adding a helper to the base class? e.g., `string FormatDisplayUrl(string encodedUrl);`?

src/NuGetGallery.Core/Infrastructure/mail/EmailBuilder.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Mail;
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    public abstract class EmailBuilder : IEmailBuilder
```
chenriksson: The individual email classes definitely looks nicer than having it all crammed into the single service. Nice!

src/NuGetGallery/Services/MarkdownMessageService.cs
scottbommarito: IMO the differences in behavior should be entirely represented by the individual `IEmailBuilder` implementations. It's confusing for people to have to go both the `IEmailBuilder` implementation and this class for the full "description" of an email.

src/NuGetGallery/Infrastructure/Mail/Messages/PackageOwnershipRequestInitiatedMessage.cs
scottbommarito: Regardless of what is done, [`MarkdownMailer` does it automatically](https://github.com/half-ogre/MarkdownMailer/blob/1b37a110ae5106bce9d3830d2fcfbb5a60839107/Lib/MailSender.cs#L96), and the implementation is pretty simple and only requires a dependency on `System.Net.Mail`. We don't need to hard-code a plaintext conversion for every mail when we can just write 3 or 4 lines of code to do it automatically.

src/NuGetGallery/Infrastructure/Mail/Requests/OrganizationMembershipRequest.cs
scottbommarito: I found the reason for it!

https://github.com/NuGet/NuGetGallery/pull/2358

I think we should move this UrlDecode to some utility class with a name like `EscapeLinkForMarkdown` and then give it a more descriptive explanation of why we're doing it.

src/NuGetGallery/Infrastructure/Mail/EmailRecipientsWithPermission.cs
xavierdecoster: Based on an earlier suggestion by @loic-sharma to avoid null checks everywhere (e.g. when constructing the message). I can leave them `null` and add the `null check`, all fine by me :)

src/NuGetGallery/Infrastructure/Mail/Messages/ApiKeyAddedMessage.cs
xavierdecoster: To answer the *why*: I call it *legacy* :)

See https://github.com/NuGet/NuGetGallery/pull/6505#discussion_r222250671

xavierdecoster:
Working on another iteration. Please hold off on further reviews until then, processing some good feedback here :)

src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -168,6 +168,22 @@
...
+    <Compile Include="Infrastructure\Mail\Messages\ContactSupportMessage.cs" />
+    <Compile Include="Infrastructure\Mail\Messages\PackageAddedMessage.cs" />
+    <Compile Include="Infrastructure\Mail\Messages\PackageAddedWithWarningsMessage.cs" />
+    <Compile Include="Infrastructure\Mail\Messages\PackageValidationFailedMessage.cs" />
+    <Compile Include="Infrastructure\Mail\Messages\PackageValidationTakingTooLongMessage.cs" />
+    <Compile Include="Infrastructure\Mail\Messages\SymbolPackageAddedMessage.cs" />
+    <Compile Include="Infrastructure\Mail\Messages\SymbolPackageValidationFailedMessage.cs" />
+    <Compile Include="Infrastructure\Mail\Messages\SymbolPackageValidationTakingTooLongMessage.cs" />
+    <Compile Include="Infrastructure\Mail\Requests\ContactSupportRequest.cs" />
+    <Compile Include="Infrastructure\Mail\ValidationIssueExtensions.cs" />
```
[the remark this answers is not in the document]
xavierdecoster: We could definitely look into that. Need to first check if this mail stuff has dependencies on anything defined within Gallery.Core. If this refactoring makes it in, that might be the next thing to work on before distributing this to other clients for consumption.

src/NuGetGallery.Core/Infrastructure/mail/EmailBuilder.cs
```diff
@@ -0,0 +1,33 @@
...
+        public string GetBody(EmailFormat format)
+        {
+            switch (format)
+            {
+                case EmailFormat.PlainText:
+                    return GetPlainTextBody();
+                case EmailFormat.Markdown:
+                    return GetMarkdownBody();
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(format));
```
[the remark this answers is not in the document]
xavierdecoster: I was wondering the same thing. Should it? We'll be adding `EmailFormat.Html` later on.

src/NuGetGallery.Core/Services/CoreMarkdownMessageService.cs
```diff
@@ -0,0 +1,208 @@
...
+                    }
+                    else
+                    {
+                        throw;
+                    }
+                }
+            }
+        }
+
+        protected virtual Task AttemptSendMessageAsync(MailMessage mailMessage, int attemptNumber)
```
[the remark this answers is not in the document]
xavierdecoster: It is used in an override in `MarkdownMessageService`. (I didn't change this code)

src/NuGetGallery.Core/Infrastructure/mail/ConfirmationEmailBuilder.cs
xavierdecoster: Addressed in https://github.com/NuGet/NuGetGallery/pull/6505/commits/288d5714ec677e86d35b49cd3a038bda8f7a740c

src/NuGetGallery.Core/Services/CoreMarkdownMessageService.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        protected virtual async Task SendMessageAsync(MailMessage mailMessage)
+        {
+            if (!mailMessage.To.Any())
+            {
+                return;
+            }
+
+            int attempt = 0;
+            bool success = false;
```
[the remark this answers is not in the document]
xavierdecoster: Fixed.

src/NuGetGallery/Services/MarkdownMessageService.cs
xavierdecoster: You might like this new iteration: https://github.com/NuGet/NuGetGallery/pull/6505/commits/288d5714ec677e86d35b49cd3a038bda8f7a740c

src/NuGetGallery/Infrastructure/Mail/Requests/OrganizationMembershipRequest.cs
xavierdecoster: Addressed in https://github.com/NuGet/NuGetGallery/pull/6505/commits/288d5714ec677e86d35b49cd3a038bda8f7a740c

src/NuGetGallery.Core/Infrastructure/mail/Messages/PackageAddedMessage.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            return $@"The package {_package.PackageRegistration.Id} {_package.Version} ({_packageUrl}) was recently published on {_configuration.GalleryOwner.DisplayName} by {_package.User.Username}. If this was not intended, please contact support: {_packageSupportUrl}.
+{warningMessagesPlaceholder}
+
+-----------------------------------------------
+    To stop receiving emails as an owner of this package, sign in to the {_configuration.GalleryOwner.DisplayName} and
+    change your email notification settings: {_emailSettingsUrl}";
+        }
+
+        private IReadOnlyList<MailAddress> AddOwnersSubscribedToPackagePushedNotification()
```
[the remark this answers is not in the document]
xavierdecoster: Addressed in https://github.com/NuGet/NuGetGallery/pull/6505/commits/288d5714ec677e86d35b49cd3a038bda8f7a740c

src/NuGetGallery.Core/Infrastructure/mail/EmailFormat.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    public enum EmailFormat
+    {
+        PlainText,
```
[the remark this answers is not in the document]
xavierdecoster: Addressed in https://github.com/NuGet/NuGetGallery/pull/6505/commits/288d5714ec677e86d35b49cd3a038bda8f7a740c

src/NuGetGallery.Core/Infrastructure/mail/EmailBuilder.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using System.Net.Mail;
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    public abstract class EmailBuilder : IEmailBuilder
+    {
+        public abstract MailAddress Sender { get; }
+
+        public string GetBody(EmailFormat format)
```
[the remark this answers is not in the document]
xavierdecoster: There's more shared logic, and the implementation for the abstract methods will be entirely different for the async messaging feature.

src/NuGetGallery/Infrastructure/Mail/EmailRecipientsWithPermission.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6505/commits/f6362f13ed3b2d1e78434b49c50da4d2035ecb8d

src/NuGetGallery.Core/Infrastructure/mail/EmailRecipients.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    public class EmailRecipients : IEmailRecipients
+    {
+        public EmailRecipients(
+            IReadOnlyList<MailAddress> to,
+            IReadOnlyList<MailAddress> cc = null,
+            IReadOnlyList<MailAddress> bcc = null,
+            IReadOnlyList<MailAddress> replyTo = null)
```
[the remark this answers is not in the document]
xavierdecoster: Not sure I'm 100% agreeing with that. The reply-to addresses are also recipients... for replies to this email message.

If this code is to be extracted into ServerCommon, it should not be limited to support only NuGetGallery scenarios.

The class makes it easier to test recipient-related logic, and share some common utility-methods (as introduced in the latest refactoring iteration).

xavierdecoster:
Did another iteration.

Next up:
looking into generating plain-text from markdown to avoid duplicate strings, possibly by consuming some OSS code from MarkdownMailer

Open questions:
* https://github.com/NuGet/NuGetGallery/pull/6505#discussion_r222426338
* https://github.com/NuGet/NuGetGallery/pull/6505#discussion_r222426073
* https://github.com/NuGet/NuGetGallery/pull/6505#discussion_r222431559
* https://github.com/NuGet/NuGetGallery/pull/6505#discussion_r222432005

Feel free to give it another pair of :eyes:

src/NuGetGallery.Core/Infrastructure/mail/IMessageServiceConfiguration.cs
```diff
@@ -3,9 +3,9 @@
 
 using System.Net.Mail;
 
-namespace NuGetGallery.Services
+namespace NuGetGallery.Infrastructure.Mail
 {
-    public interface ICoreMessageServiceConfiguration
+    public interface IMessageServiceConfiguration
```
xavierdecoster: No more need for separate `ICoreMessageService` and `IMessageService`, so this can just be `IMessageServiceConfiguration`.

src/NuGetGallery.Core/Services/ICoreMessageService.cs
```diff
@@ -1,20 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System.Collections.Generic;
-using NuGet.Services.Validation;
-using System.Threading.Tasks;
-
-namespace NuGetGallery.Services
-{
-    public interface ICoreMessageService
```
xavierdecoster: No longer needed. `IMessageService` has the one method we need to send messages.

src/NuGetGallery/Infrastructure/Mail/EmailRecipientsWithPermission.cs
xavierdecoster: The problem here is that this particular class depends on the `ActionRequiringAccountPermissions` type defined in `NuGetGallery`. This type is not available in `NuGetGallery.Core`.

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -23,7 +23,8 @@
     <RunCodeAnalysis Condition=" '$(CodeAnalysis)' == 'true' ">true</RunCodeAnalysis>
     <OldToolsVersion>12.0</OldToolsVersion>
     <TargetFrameworkProfile />
-    <IISExpressSSLPort>443</IISExpressSSLPort>
+    <IISExpressSSLPort>
```
loic-sharma: >IISExpressSSLPort [](start = 5, length = 17)

Why was this changed?

xavierdecoster: Unintended change, good catch! Fixed in https://github.com/NuGet/NuGetGallery/pull/6505/commits/bdcbcd605c7732baf7d5adb2dc436b1a2b7025a2

xavierdecoster:
It took a bit longer than anticipated, but I managed to add auto-conversion from markdown to plaintext (and it supports to html too), including test coverage for all email message content in every format.

Please give it another pair of 👀 when you can, so we can finish up on this refactoring and move forward with adding the async messaging stuff (which uses HTML and plain text).

xavierdecoster:
To facilitate reviewing this massive PR: changes were approved up to commit https://github.com/NuGet/NuGetGallery/pull/6505/commits/c4bf87f96ddec7dade23a1d1db0b776b4c5f97b4.

Since then I pushed two additional commits:
- Use Markdig to auto-convert markdown messages to plain text aef14cd
- Test coverage b15450a

PR #6510, opened 2018-10-04
Clean Bootstrap dependencies

scottbommarito:
https://github.com/NuGet/Engineering/issues/1729

I determined what tasks in the `Gruntfile.js` we actually use and removed all others and the dependencies pulled in by them. I then deleted the `node-modules` folder and `package-lock.json` file and reinstalled everything to "refresh" the lock file, which pulled in the latest versions of all the dependencies that are actually used.

```
npm install
grunt
```
still runs and emits the exact same output as before


scottbommarito:
CI failed due to lack of disk space...investigating now

Issue #6422, opened 2018-09-12
Deleted API key notifications are composed using the wrong API key name


agr:
Currently if user deletes an API key they receive the email notification with the following text:

> API key 'Full access API key' was removed from your account and can no longer be used. If you did not request this change, please reply to this email to contact support.

That `Full access API key` portion is always there regardless of the actual API key name.

This seem to happen because email is composed *after* the key is deleted:
https://github.com/NuGet/NuGetGallery/blob/619bede5c0580cd553d4b5332f35454a12ebebbf/src/NuGetGallery/Controllers/UsersController.cs#L995-L998

PR #6512, opened 2018-10-04
Fix bugs - issue 6422

zhhyu:
Fix bugs - issue 6422:  https://github.com/NuGet/NuGetGallery/issues/6422

The previous code will remove the credential firstly, which lead to empty  scopes for any kinds of credentials, so the message will always be "NonScopedApiKeyDescription". In order to fix that, we get the credential description firstly before removing operation, and use that description to determine the email content.

### DIFF
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -990,10 +990,12 @@ namespace NuGetGallery
                 return Json(Strings.CredentialNotFound);
             }
 
+            var isNonScopedAPIkey = !cred.IsScopedApiKey();
+
             await AuthenticationService.RemoveCredential(user, cred);
 
             // Notify the user of the change
-            await MessageService.SendCredentialRemovedNoticeAsync(user, AuthenticationService.DescribeCredential(cred));
+            await MessageService.SendCredentialRemovedNoticeAsync(user, AuthenticationService.DescribeCredential(cred, isNonScopedAPIkey));
 
             return Json(Strings.CredentialRemoved);
         }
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -990,10 +990,12 @@ private async Task<JsonResult> RemoveApiKeyCredential(User user, Credential cred
                 return Json(Strings.CredentialNotFound);
             }
 
+            var isNonScopedAPIkey = !cred.IsScopedApiKey();
+
             await AuthenticationService.RemoveCredential(user, cred);
 
             // Notify the user of the change
-            await MessageService.SendCredentialRemovedNoticeAsync(user, AuthenticationService.DescribeCredential(cred));
+            await MessageService.SendCredentialRemovedNoticeAsync(user, AuthenticationService.DescribeCredential(cred, isNonScopedAPIkey));
```
scottbommarito: Any reason we can't just do
```
var credDescription = AuthenticationService.DescribeCredential(cred);
await AuthenticationService.RemoveCredential(user, cred);
await MessageService.SendCredentialRemovedNoticeAsync(user, credDescription);
```

zhhyu: Great idea, thanks!

tests/NuGetGallery.Facts/Controllers/UsersControllerFacts.cs
```diff
@@ -1642,6 +1642,59 @@ public async Task GivenValidRequest_ItRemovesCredAndSendsNotificationToUser()
...
+                        {
+                            new []
+                            {
+                                new Scope("*", NuGetScopes.All)
+                            },
+                            "theApiKey",
+                        },
+                        new object[]
+                        {
+                            new Scope[]{ },
```
joelverhagen: nit `new Scope[0],`

PR #6513, opened 2018-10-05
[NuGet Symbol Server] Add symbols delete support to gallery

shishirx34:
This PR adds delete support based on the spec: https://github.com/NuGet/Home/wiki/Symbols-Package-Upload-and-Delete-Workflow 

Also, added tests. 

Screenshot:
Signed in Owner:
image

Everyone else;
image

Delete page;

image

Confirm:

image

Symbols deleted! Yay!! no more option to delete the symbols package/or download symbols

image

### DIFF
--- a/src/NuGetGallery.Core/Auditing/PackageCreatedVia.cs
+++ b/src/NuGetGallery.Core/Auditing/PackageCreatedVia.cs
@@ -15,4 +15,17 @@ namespace NuGetGallery.Auditing
         /// </summary>
         public const string Web = "Created via web.";
     }
+
+    public static class PackageDeletedVia
+    {
+        /// <summary>
+        /// Package has been deleted via NuGet API (nuget.exe push)
+        /// </summary>
+        public const string Api = "Deleted via API.";
+
+        /// <summary>
+        /// Package has been deleted via NuGet web interface (browser)
+        /// </summary>
+        public const string Web = "Deleted via web.";
+    }
 }
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -21,6 +21,7 @@ using NuGetGallery.Areas.Admin.Models;
 using NuGetGallery.AsyncFileUpload;
 using NuGetGallery.Auditing;
 using NuGetGallery.Configuration;
+using NuGetGallery.Diagnostics;
 using NuGetGallery.Filters;
 using NuGetGallery.Helpers;
 using NuGetGallery.Infrastructure.Lucene;
@@ -88,6 +89,7 @@ namespace NuGetGallery
         private readonly IPackageOwnershipManagementService _packageOwnershipManagementService;
         private readonly IContentObjectService _contentObjectService;
         private readonly ISymbolPackageUploadService _symbolPackageUploadService;
+        private readonly IDiagnosticsSource _trace;
 
         public PackagesController(
             IPackageService packageService,
@@ -111,7 +113,8 @@ namespace NuGetGallery
             IValidationService validationService,
             IPackageOwnershipManagementService packageOwnershipManagementService,
             IContentObjectService contentObjectService,
-            ISymbolPackageUploadService symbolPackageUploadService)
+            ISymbolPackageUploadService symbolPackageUploadService,
+            IDiagnosticsService diagnosticsService)
         {
             _packageService = packageService;
             _uploadFileService = uploadFileService;
@@ -135,6 +138,7 @@ namespace NuGetGallery
             _packageOwnershipManagementService = packageOwnershipManagementService;
             _contentObjectService = contentObjectService;
             _symbolPackageUploadService = symbolPackageUploadService;
+            _trace = diagnosticsService?.SafeGetSource(nameof(PackagesController)) ?? throw new ArgumentNullException(nameof(diagnosticsService));
         }
 
         [HttpGet]
@@ -1172,12 +1176,13 @@ namespace NuGetGallery
                 return HttpNotFound();
             }
 
-            if (ActionsRequiringPermissions.DeleteSymbolPackage.CheckPermissionsOnBehalfOfAnyAccount(GetCurrentUser(), package) != PermissionsCheckResult.Allowed)
+            var currentUser = GetCurrentUser();
+            if (ActionsRequiringPermissions.DeleteSymbolPackage.CheckPermissionsOnBehalfOfAnyAccount(currentUser, package) != PermissionsCheckResult.Allowed)
             {
                 return HttpForbidden();
             }
 
-            var model = new DeletePackageViewModel(package, GetCurrentUser(), DeleteReasons);
+            var model = new DeletePackageViewModel(package, currentUser, DeleteReasons);
 
             model.VersionSelectList = new SelectList(
                 model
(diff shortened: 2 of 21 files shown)


shishirx34:
@diverdan92 - please verify the UI and strings look good.

shishirx34:
@cristinamanum @chenriksson @joelverhagen @scottbommarito - please review if you can.

src/NuGetGallery.Core/Auditing/PackageCreatedVia.cs
```diff
@@ -15,4 +15,17 @@ public static class PackageCreatedVia
         /// </summary>
         public const string Web = "Created via web.";
     }
+
+    public static class PackageDeletedVia
+    {
+        /// <summary>
+        /// Package has been deleted via NuGet API (nuget.exe push)
```
cristinamanum: I guess is nuget.exe delete instead of nuget.exe push

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1278,6 +1311,61 @@ public virtual async Task<ActionResult> Delete(DeletePackagesRequest deletePacka
...
+            }
+
+            // Get all available symbol packages for a given package, ideally this should
+            // always return one symbol package. For thoroughness we can cleanup the data
+            // for any inconsistencies.
+            var availableSymbolPackages = package
+                .SymbolPackages
+                .Where(sp => sp.StatusKey == PackageStatus.Available);
+            if (availableSymbolPackages.Any())
+            {
```
cristinamanum: Can we log an warning if more than one packages are found?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1278,6 +1311,61 @@ public virtual async Task<ActionResult> Delete(DeletePackagesRequest deletePacka
             return Delete(firstPackage.PackageRegistration.Id, firstPackage.Version);
         }
 
+        [UIAuthorize]
+        [HttpPost]
+        [RequiresAccountConfirmation("delete a symbols package")]
+        [ValidateAntiForgeryToken]
+        public virtual async Task<ActionResult> DeleteSymbolsPackage(string id, string version)
+        {
+            var package = _packageService.FindPackageByIdAndVersionStrict(id, version);
```
cristinamanum: Can we add a telemetryMetric for deleted Symbols action?

cristinamanum: I see that you already have it below. Please ignore.

src/NuGetGallery/Services/SymbolPackageUploadService.cs
```diff
@@ -230,5 +230,20 @@ public async Task<PackageCommitResult> CreateAndUploadSymbolsPackage(Package pac
 
             return PackageCommitResult.Success;
         }
+
+        public async Task DeleteSymbolsPackageAsync(SymbolPackage symbolPackage)
```
cristinamanum: The class name is SymbolPackageUploadService and this a delete operation. Maybe rename to SymbolPackageService ?

src/NuGetGallery/Views/Packages/DeleteSymbols.cshtml
```diff
@@ -0,0 +1,137 @@
...
+                            {
+                                @class = "form-control",
+                                title = "Select a version",
+                                id = "input-select-version"
+                            })
+                    </div>
+                </div>
+            }
+
+            @if (User.IsAdministrator())
```
cristinamanum: Why is the delete flow only for the Admin?

diverdan92: What's the difference between admin vs. user from the NuGet user perspective? Does this mean only NuGet.org admins (i.e. us) can delete symbols?

shishirx34: My bad, you are right, this shouldn't be here. Unfortunately while testing I was an administrator and missed it. I will fix this and test with non-admin delete flow.

src/NuGetGallery/Services/SymbolPackageUploadService.cs
shishirx34: We already have a symbol package service, which updates the status of the symbols package, this method calls it. I agree the naming is a bit inconsistent, but this is the common service that we want to use for any symbols package related flow between different controllers(`Packages` or `Api`). I would leave it as is for now to avoid unnecessary refactoring and possibly consider renaming this file to something better(`SymbolPackageManagementService`, maybe?) in the future.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1161,6 +1161,39 @@ public virtual ActionResult Delete(string id, string version)
...
+        [RequiresAccountConfirmation("delete a symbols package")]
+        public virtual ActionResult DeleteSymbols(string id, string version)
+        {
+            var package = _packageService.FindPackageByIdAndVersion(id, version);
+            if (package == null)
+            {
+                return HttpNotFound();
+            }
+
+            if (ActionsRequiringPermissions.DeleteSymbolPackage.CheckPermissionsOnBehalfOfAnyAccount(GetCurrentUser(), package) != PermissionsCheckResult.Allowed)
```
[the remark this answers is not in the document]
shishirx34: only if the user doesn't exist in the `OwinContext` which is not the case for any of the routes in this controller. still will cache it for minor perf gain.

src/NuGetGallery/Services/SymbolPackageUploadService.cs
```diff
@@ -230,5 +230,20 @@ public async Task<PackageCommitResult> CreateAndUploadSymbolsPackage(Package pac
...
+        {
+            if (symbolPackage == null)
+            {
+                throw new ArgumentNullException(nameof(symbolPackage));
+            }
+
+            if (await _symbolPackageFileService.DoesPackageFileExistAsync(symbolPackage.Package))
```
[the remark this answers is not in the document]
shishirx34: possibly, but this operation is repeatable and next time if the file doesn't exist, the DB entry will be cleaned up. Same is the scenario with package deletes.

src/NuGetGallery/Views/Packages/DeleteSymbols.cshtml
```diff
@@ -0,0 +1,137 @@
+﻿@model DeletePackageViewModel
+@{
+    ViewBag.Title = "Manage Symbols Package " + Model.Id;
+    ViewBag.Tab = "Symbol Packages";
+    ViewBag.MdPageColumns = Constants.ColumnsFormMd;
+}
+
+<section role="main" class="container main-container page-delete-package">
```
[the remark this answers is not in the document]
shishirx34: I thought about this, but then decided to keep it separate due to the fact that package delete is very site admin specific, will need to add a lot `if..else` for accommodating symbols delete and don't want to give a chance to accidentally miss anything and make package delete page public. better to duplicate and be safe.

src/NuGetGallery.Core/Auditing/PackageCreatedVia.cs
```diff
@@ -15,4 +15,17 @@ public static class PackageCreatedVia
         /// </summary>
         public const string Web = "Created via web.";
     }
+
+    public static class PackageDeletedVia
```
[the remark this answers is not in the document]
shishirx34: renamed the file for now.

PR #6544, opened 2018-10-09
Latest fxcop fixes

chenriksson:


joelverhagen:
Let's see if `GlobalSuppressions.cs` is still necessary.

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -32,6 +32,7 @@
     <UseGlobalApplicationHostFile />
     <MvcBuildViews Condition=" '$(MvcBuildViews)' == '' ">false</MvcBuildViews>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <NoWarn>$(NoWarn);CA3103;CA3119</NoWarn>
```
joelverhagen: nit: perhaps we should define this internally since, AFAIK, these are internal codes.

chenriksson:
@joelverhagen updates since our discussion:
- `NoWarn` didn't work for `CA3103` for some reason, so I left `GlobalSuppressions.cs` with module-level suppressions (generated by FxCop.exe) that actually worked. Changed to UTF-8 encoding, since GitHub thinks this file is binary.
- Didn't need all the csproj changes I had, so reverted
- Removed a lot of the old inline suppressions and CodeAnalysis references in csproj, which don't seem to be relevant anymore
- This PR fixes errors, but we have lots of TSA issues for warnings to fix (~700 across all repos)

/cc @agr @scottbommarito Checking in, to finish off this task... but feel free to 'r' me with feedback for changes since you approved.

PR #6547, opened 2018-10-09
[NuGet Symbol Server] Revalidate Symbols

shishirx34:
This PR adds support for symbols package revalidation. I will be validating everything on DEV before I merge this PR. Added tests.

### DIFF
--- a/src/NuGetGallery.Core/Extensions/PackageExtensions.cs
+++ b/src/NuGetGallery.Core/Extensions/PackageExtensions.cs
@@ -1,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+
+namespace NuGetGallery.Extensions
+{
+    public static class PackageExtensions
+    {
+        /// <summary>
+        /// Get the latest(last) uploaded symbols package for the given package.
+        /// </summary>
+        /// <param name="package"><see cref="Package"/> for which latest symbol package is to be retrieved.</param>
+        /// <returns>The latest symbol package if present, null otherwise</returns>
+        public static SymbolPackage LatestSymbolPackage(this Package package)
+        {
+            return package
+                .SymbolPackages
+                .OrderByDescending(sp => sp.Created)
+                .FirstOrDefault();
+        }
+
+        /// <summary>
+        /// Get the latest(last) uploaded symbols package for the given package, if it is <see cref="PackageStatus.Available"/>.
+        /// </summary>
+        /// <param name="package"><see cref="Package"/> for which latest available symbol package is to be retrieved.</param>
+        /// <returns>The latest symbol package if its status is <see cref="PackageStatus.Available"/>, null otherwise</returns>
+        public static SymbolPackage LatestAvailableSymbolPackage(this Package package)
+        {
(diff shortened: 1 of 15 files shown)


src/NuGetGallery.Core/Extensions/PackageExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        /// <summary>
+        /// Get the latest(last) uploaded symbols package for the given package.
+        /// </summary>
+        /// <param name="package"><see cref="Package"/> for which latest symbol package is to be retrieved.</param>
+        /// <returns>The latest symbol package if present, null otherwise</returns>
+        public static SymbolPackage LatestSymbolPackage(this Package package)
+        {
+            return package
+                .SymbolPackages
```
agr: This assumes that `SymbolPackages` navigation property was loaded from DB (we have lazy load disabled, do we?) Is this always the case? Do we need check it for `null` here and throw some meaningful exception?

src/NuGetGallery.Core/Extensions/PackageExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+                .FirstOrDefault();
+        }
+
+        /// <summary>
+        /// Get the latest(last) uploaded symbols package for the given package, if it is <see cref="PackageStatus.Available"/>.
+        /// </summary>
+        /// <param name="package"><see cref="Package"/> for which latest available symbol package is to be retrieved.</param>
+        /// <returns>The latest symbol package if its status is <see cref="PackageStatus.Available"/>, null otherwise</returns>
+        public static SymbolPackage LatestAvailableSymbolPackage(this Package package)
+        {
```
cristinamanum: If it is not used please delete.

src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -296,10 +297,10 @@
       <Version>4.8.0-preview4.5287</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Validation">
-      <Version>2.28.0-master-37018</Version>
+      <Version>2.28.0-master-42295</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Validation.Issues">
-      <Version>2.28.0-master-37018</Version>
+      <Version>2.28.0-master-42295</Version>
```
cristinamanum: There is a 29.0 version available.

src/NuGetGallery.Core/Extensions/PackageExtensions.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+
+namespace NuGetGallery.Extensions
```
chenriksson: Don't our other extensions classes use the `NuGetGallery` namespace?

shishirx34: `PackageRegistrationExtensions` uses `NuGetGallery.Extensions` kept in sync with it.

src/NuGetGallery.Core/NuGetGallery.Core.csproj
shishirx34: updated.

Issue #6389, opened 2018-09-05
Don't ask for e-mail verification for new accounts


skofman1:
NuGet.org shouldn't ask for e-mail verification for new accounts since only MSA/AAD are used to create those accounts, meaning the e-mail is verified.
However, we should still verify e-mail in cases the user changes the e-mail later on.

PR #6548, opened 2018-10-10
Fix email verification issue for AAD/MSA registration

zhhyu:
Fix issue https://github.com/NuGet/NuGetGallery/issues/6389

If the user registers using the MSA/AAD account, we will confirm the account and set up the email address automatically without sending any confirmation email. 

When the user changes email address, we will send the confirmation email as before.

### DIFF
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -299,7 +299,7 @@ namespace NuGetGallery.Authentication
             return claims.ToArray();
         }
 
-        public virtual async Task<AuthenticatedUser> Register(string username, string emailAddress, Credential credential)
+        public virtual async Task<AuthenticatedUser> Register(string username, string emailAddress, Credential credential, bool autoConfirm = false)
         {
             if (_config.FeedOnlyMode)
             {
@@ -332,7 +332,7 @@ namespace NuGetGallery.Authentication
             // Add a credential for the password
             newUser.Credentials.Add(credential);
 
-            if (!_config.ConfirmEmailAddresses || credential.IsExternal())
+            if (!_config.ConfirmEmailAddresses || (credential.IsExternal() && autoConfirm))
             {
                 newUser.ConfirmEmailAddress();
             }
@@ -651,7 +651,8 @@ namespace NuGetGallery.Authentication
                     ExternalIdentity = externalIdentity,
                     Authenticator = authenticator,
                     Credential = _credentialBuilder.CreateExternalCredential(userInfo.AuthenticationType, userInfo.Identifier, identity, userInfo.TenantId),
-                    LoginDetails = new ExternalLoginSessionDetails(userInfo.Email, userInfo.UsedMultiFactorAuthentication)
+                    LoginDetails = new ExternalLoginSessionDetails(userInfo.Email, userInfo.UsedMultiFactorAuthentication),
+                    UserInfo = userInfo
                 };
             }
             catch (Exception ex)
--- a/src/NuGetGallery/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery/Authentication/AuthenticationService.cs
@@ -332,7 +332,7 @@ namespace NuGetGallery.Authentication
             // Add a credential for the password
             newUser.Credentials.Add(credential);
 
-            if (!_config.ConfirmEmailAddresses || (credential.IsExternal() && autoConfirm))
+            if (!_config.ConfirmEmailAddresses || autoConfirm)
             {
                 newUser.ConfirmEmailAddress();
             }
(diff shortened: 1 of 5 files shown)


src/NuGetGallery/Authentication/AuthenticationService.cs
```diff
@@ -332,7 +332,7 @@ public virtual async Task<AuthenticatedUser> Register(string username, string em
             // Add a credential for the password
             newUser.Credentials.Add(credential);
 
-            if (!_config.ConfirmEmailAddresses)
+            if (!_config.ConfirmEmailAddresses || (credential.IsExternal() && autoConfirm))
```
joelverhagen: >credential [](start = 51, length = 10)

Why do we need to check `IsExternal()` here? Isn't it up to the caller to check that now?

tests/NuGetGallery.Facts/Authentication/AuthenticationServiceFacts.cs
```diff
@@ -844,6 +845,56 @@ public async Task WillSaveTheNewUserAsConfirmedWhenConfigured()
...
+                // Act
+                var authUser = await auth.Register(
+                    "newUser",
+                    "theEmailAddress",
+                    new CredentialBuilder().CreateExternalCredential(credType, "blorg", "Bloog"),
+                    true);
+
+                // Assert
+                Assert.True(auth.Entities.Users.Contains(authUser.User));
+                Assert.True(authUser.User.Confirmed);
```
joelverhagen: Maybe assert that the EmailAddress field equals "theEmailAddress"?

tests/NuGetGallery.Facts/Authentication/AuthenticationServiceFacts.cs
```diff
@@ -844,6 +845,56 @@ public async Task WillSaveTheNewUserAsConfirmedWhenConfigured()
...
+                    "theEmailAddress",
+                    new CredentialBuilder().CreateExternalCredential(credType, "blorg", "Bloog"),
+                    false);
+
+                // Assert
+                Assert.True(auth.Entities.Users.Contains(authUser.User));
+                auth.Entities.VerifyCommitChanges();
+
+                Assert.NotNull(authUser.User.EmailConfirmationToken);
+                Assert.False(authUser.User.Confirmed);
```
joelverhagen: Maybe assert that the UnconfirmedEmailAddress field equals "theEmailAddress"?

joelverhagen:
>                 ResultAssert.IsSafeRedirectTo(result, "/theReturnUrl");

We should verify that "autoConfirm: true" is passed down

---
Refers to: tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs:896 in 878b937. [](commit_id = 878b93795177930180abbcc5815167c3bc73dcdf, deletion_comment = False)

joelverhagen:
>                 GetMock<IMessageService>()

We should verify that "autoConfirm: false" is passed down (I _think_ that's what's happening here)

---
Refers to: tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs:800 in 878b937. [](commit_id = 878b93795177930180abbcc5815167c3bc73dcdf, deletion_comment = False)

src/NuGetGallery/Authentication/AuthenticationService.cs
shishirx34: @joelverhagen is right, remove the `IsExternal` check.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -267,7 +267,9 @@ public virtual async Task<ActionResult> Register(LogOnViewModel model, string re
                     user = await _authService.Register(
                         model.Register.Username,
                         model.Register.EmailAddress,
-                        result.Credential);
+                        result.Credential,
+                        string.Equals(result.UserInfo.Email, model.Register.EmailAddress, StringComparison.OrdinalIgnoreCase)
```
shishirx34: should ensure the `UserInfo` object is not null. 
```
string.Equals(result.UserInfo?.Email, model.Register.EmailAddress, StringComparison.OrdinalIgnoreCase);
```

src/NuGetGallery/Authentication/AuthenticationService.cs
zhhyu: That's true. Move to the caller.

tests/NuGetGallery.Facts/Authentication/AuthenticationServiceFacts.cs
zhhyu: Done

tests/NuGetGallery.Facts/Authentication/AuthenticationServiceFacts.cs
zhhyu: Done

src/NuGetGallery/Controllers/AuthenticationController.cs
zhhyu: Done

joelverhagen:
Spoke offline about one more test case. Once you add that, we're good to go.

PR #6555, opened 2018-10-11
Emails should not be sent synchronously

xavierdecoster:
Now that the [refactoring PR](https://github.com/NuGet/NuGetGallery/issues/6505) has been completed, this PR supersedes https://github.com/NuGet/NuGetGallery/pull/6495.

This PR introduces async email messaging to the NuGetGallery, leveraging the new ServerCommon `NuGet.Services.Messaging` library (for which an official release needs to be created still, after which I can upgrade the nupkg dependencies here).

To enable it, configure the following settings in `web.config`:

```xml
    <add key="Gallery.AsynchronousEmailServiceEnabled" value="true"/>
    <add key="AzureServiceBus.EmailPublisher.ConnectionString" value=""/>
    <add key="AzureServiceBus.EmailPublisher.TopicName" value=""/>
```

By default, `Gallery.AsynchronousEmailServiceEnabled` will be `false` (or even missing), returning the NuGetGallery email functionality into its current SMTP-mode (still using the Markdown infrastructure we already had).

### DIFF
--- a/src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
+++ b/src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
@@ -1,0 +1,95 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using NuGet.Services.Messaging;
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    public class AsynchronousEmailMessageService : IMessageService
+    {
+        private readonly IMessageServiceConfiguration _configuration;
+        private readonly IEmailMessageEnqueuer _emailMessageEnqueuer;
+
+        public AsynchronousEmailMessageService(
+            IMessageServiceConfiguration configuration,
+            IEmailMessageEnqueuer emailMessageEnqueuer)
+        {
+            _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+            _emailMessageEnqueuer = emailMessageEnqueuer ?? throw new ArgumentNullException(nameof(emailMessageEnqueuer));
+        }
+
+        public Task SendMessageAsync(
+            IEmailBuilder emailBuilder,
+            bool copySender = false,
+            bool discloseSenderAddress = false)
+        {
+            var message = CreateMessage(
+                emailBuilder,
+                copySender,
+                discloseSenderAddress);
+
+            return EnqueueMessageAsync(message);
+        }
+
+        private static EmailMessageData CreateMessage(
+            IEmailBuilder emailBuilder,
+            bool copySender = false,
+            bool discloseSenderAddress = false)
+        {
+            var recipients = emailBuilder.GetRecipients();
+
+            return new EmailMessageData(
+                emailBuilder.GetSubject(),
+                emailBuilder.GetBody(EmailFormat.PlainText),
+                emailBuilder.GetBody(EmailFormat.Html),
+                emailBuilder.Sender.Address,
+                to: recipients.To.Select(e => e.Address).ToList(),
+                cc: GenerateCC(
+                    emailBuilder.Sender.Address,
+                    recipients.CC.Select(e => e.Address).ToList(),
+                    copySender,
+                    discloseSenderAddress),
+                bcc: recipients.Bcc.Select(e => e.Address).ToList(),
+                replyTo: recipients.ReplyTo.Select(e => e.Address).ToList(),
+                messageTrackingId: Guid.NewGuid());
+        }
+
+        private static IReadOnlyList<string> GenerateCC(
+            string fromAddress,
+            IReadOnlyList<string> cc, bool copySender,
+            bool discloseSenderAddress)
+        {
+            var ccList = new List<string>();
+            if (cc != null)
+            {
+                ccList.AddRange(cc);
+            }
+
+            if (copySender && discloseSenderAddress && !ccList.Contains(fromAddress))
+            {
+                ccList.Add(fromAddress);
+            }
+
+            return ccList;
+        }
+
+        private Task EnqueueMessageAsync(EmailMessageData message)
+        {
+            if (message == null)
+            {
+                throw new ArgumentNullException(nameof(message));
+            }
+
+            if (!message.To.Any())
+            {
+                return Task.CompletedTask;
+            }
+
+            return _emailMessageEnqueuer.SendEmailMessageAsync(message);
+        }
+    }
+}
--- a/tests/NuGetGallery.Core.Facts/Infrastructure/Mail/Messages/ContactSupportMessageFacts.cs
+++ b/tests/NuGetGallery.Core.Facts/Infrastructure/Mail/Messages/ContactSupportMessageFacts.cs
@@ -101,6 +101,7 @@ namespace NuGetGallery.Infrastructure.Mail.Messages
             [Theory]
             [InlineData(EmailFormat.Markdown, _expectedMarkdownBody)]
             [InlineData(EmailFormat.PlainText, _expectedPlainTextBody)]
+            [InlineData(EmailFormat.Html, _expectedHtmlBody)]
             public void ReturnsExpectedBody(EmailFormat format, string expectedString)
             {
                 var message = CreateMessage();
@@ -146,5 +147,12 @@ reason
 
 Message:
 message";
+
+        private const string _expectedHtmlBody =
+            "<p><strong>Email:</strong> requestingUser (requestUser@gallery.org)</p>\n" +
+            "<p><strong>Reason:</strong>\n" +
+            "reason</p>\n" +
+            "<p><strong>Message:</strong>\n" +
+            "message</p>\n";
     }
 }
--- a/src/NuGetGallery.Core/Infrastructure/mail/EmailBuilder.cs
+++ b/src/NuGetGallery.Core/Infrastructure/mail/EmailBuilder.cs
@@ -22,6 +22,8 @@ namespace NuGetGallery.Infrastructure.Mail
                     return GetPlainTextBody();
                 case EmailFormat.Markdown:
                     return GetMarkdownBody();
+                case EmailFormat.Html:
+                    return GetHtmlBody();
                 default:
                     throw new ArgumentOutOfRangeException(nameof(format));
             }
@@ -33,6 +35,7 @@ namespace NuGetGallery.Infrastructure.Mail
 
         protected abstract string GetPlainTextBody();
         protected abstract string GetMarkdownBody();
+        protected abstract string GetHtmlBody();
 
         /// <summary>
         /// Markdown sees the underscore as italics indicator, so underscores are stripped in the message.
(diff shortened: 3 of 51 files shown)


src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                throw new ArgumentNullException(nameof(message));
+            }
+
+            if (!message.To.Any())
+            {
+                return Task.CompletedTask;
+            }
+
+            if (string.IsNullOrEmpty(message.HtmlBody)
```
joelverhagen: Shouldn't both be set, i.e. `&&` to `||`?

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
```diff
@@ -0,0 +1,103 @@
...
+        private static EmailMessageData CreateMessage(
+            IEmailBuilder emailBuilder,
+            bool copySender = false,
+            bool discloseSenderAddress = false)
+        {
+            var recipients = emailBuilder.GetRecipients();
+
+            return new EmailMessageData(
+                emailBuilder.GetSubject(),
+                emailBuilder.GetBody(EmailFormat.PlainText),
```
joelverhagen: Why did we choose to do the Markdown to HTML/plain text conversion before enqueue? If all of the enqueuers just enqueued markdown, then wouldn't there only need to be one place with a markdown dependency? Additionally, the `EmailMessageData` API allows for the HTML and plain text to diverge which seems like something that could be avoided if there was only one notion of the message body until the very end (when we send it to the external mail service).

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
```diff
@@ -0,0 +1,103 @@
...
+                emailBuilder.Sender.Address,
+                to: recipients.To.Select(e => e.Address).ToList(),
+                cc: GenerateCC(
+                    emailBuilder.Sender.Address,
+                    recipients.CC.Select(e => e.Address).ToList(),
+                    copySender,
+                    discloseSenderAddress),
+                bcc: recipients.Bcc.Select(e => e.Address).ToList(),
+                replyTo: recipients.ReplyTo.Select(e => e.Address).ToList(),
+                messageTrackingId: Guid.NewGuid());
```
joelverhagen: >messageTrackingId [](start = 16, length = 17)

Can we log this GUID? so we can correlate logs on both sides?

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
```diff
@@ -0,0 +1,103 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using NuGet.Services.Messaging;
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    public class AsynchronousEmailMessageService : IMessageService
```
joelverhagen: unit tests for this

joelverhagen:
All of those test changes seem a bit mechanical, and suggest we are testing a component that isn't ours...

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            if (string.IsNullOrEmpty(message.HtmlBody)
+                && string.IsNullOrEmpty(message.PlainTextBody))
+            {
+                throw new ArgumentException(
+                    "No message body defined. Both plain-text and html bodies are empty.",
+                    nameof(message));
+            }
+
+            return _emailMessageEnqueuer.SendEmailMessageAsync(message);
```
joelverhagen: Where are the guarantees that this queue message only lives for up to 2 days (per GDPR export threshold)?

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
```diff
@@ -0,0 +1,103 @@
...
+            IReadOnlyList<string> cc, bool copySender,
+            bool discloseSenderAddress)
+        {
+            var ccList = new List<string>();
+            if (cc != null)
+            {
+                ccList.AddRange(cc);
+            }
+
+            if (copySender && discloseSenderAddress && !ccList.Contains(fromAddress))
```
joelverhagen: When would `copySender` be true and `discloseSenderAddress` be false? Or vice versa?

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
xavierdecoster: Should they? Perhaps from Gallery point-of-view you're right, but I want to know how ICP behaves. I'm assuming plain-text is a fallback for when the email client is unable to render the HTML body, but not sure whether it is mandatory from their point-of-view.

I'm reaching out to ICP folks to learn about expected behavior here.

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
xavierdecoster: Is that controller by `BrokeredMessage.ExpiresAtUtc`? Where can I set it? 
I believe that should be the responsibility of `IEmailMessageEnqueuer`.

xavierdecoster:
> All of those test changes seem a bit mechanical, and suggest we are testing a component that isn't ours...

@joelverhagen you are partially right. We customized plain text rendering by overriding components in Markdig to work around bugs. See https://github.com/NuGet/NuGetGallery/tree/dev-issue2565-2/src/NuGetGallery.Core/Infrastructure/mail/Internal

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
xavierdecoster: This happens. 

When `copySender=true` and `discloseSenderAddress=true`, we simply CC the sender.
When `copySender=true` and `discloseSenderAddress=false`, we send a separate email copy to the sender (so 2 emails are sent).

Of course, when `copySender=false`, the `discloseSenderAddress` value is ignored.

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
xavierdecoster: Centralizing email message construction was a non-goal when starting this work. It could be a convenient improvement though, but given the way this whole refactoring already side-tracked the work at hand (move off SMTP).

I'm also not sure whether we should make it impossible to be able to diverge plain text and html bodies... Isn't that a caller decision?

I'm not feeling to keen on doing this right now. Perhaps we can log an issue for this as a potential improvement?

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
xavierdecoster: Added in https://github.com/NuGet/NuGetGallery/pull/6555/commits/ddbbdc0e143503e7ae1008a5c04a5ee9cf8ee0dc

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
joelverhagen: Yeah sounds good. Assuming it's for backwards compat, it seems like we should always care about that from a NuGet perspective and enforce it.

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
joelverhagen: Wasn't it centralized before with markdown mailer?

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
joelverhagen: yeah, that would be the easiest way to handle it i think. alternatively you can have monitoring on the dequeue side based off of some timespan calculated.

joelverhagen: The "expiration" approach means that we could drop messages without realizing it i think.

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
joelverhagen: Sounds good.

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
joelverhagen: If this is affecting the delivery of the main goal here, tracking an issue seems fine. I guess it just surprised me because i thought we originally had just markdown handed to markdown mailer. I was expecting to just change that to handing markdown to an async flow (like a queue message).

xavierdecoster: There's a few cases where we can't originate from Markdown to create the Plain-Text and HTML formats. E.g. `SymbolPackageAddedMessage` will return a different footer depending on target format, because of limitations in the Markdown conversion. Without the flexibility in the servicebus message, it would be really hard to stick to markdown for most of our messages, and we'd end up duplicating message bodies in many places.

I like the idea of extracting and centralizing message construction, especially if we can get rid of some of these dependencies in the gallery, but it's non-trivial to do so right now.
Also bear in mind that people running their own gallery should still be able to use SMTP and send email messages as they do today, which are markdown-based, so not sure we can get rid of the dependency at all.

src/NuGetGallery.Core/Infrastructure/mail/AsynchronousEmailMessageService.cs
xavierdecoster: @joelverhagen @cristinamanum what's the approach we take in other jobs? let's make it consistent.

joelverhagen: Up until now, we don't have queue messages containing PII.

tests/NuGetGallery.Core.Facts/Infrastructure/Mail/Messages/ContactSupportMessageFacts.cs
```diff
@@ -146,5 +147,12 @@ private static ContactSupportMessage CreateMessage(bool copySender = false)
 
 Message:
 message";
+
+        private const string _expectedHtmlBody =
+            "<p><strong>Email:</strong> requestingUser (requestUser@gallery.org)</p>\n" +
+            "<p><strong>Reason:</strong>\n" +
```
agr: What wraps that partial HTML with the rest of required tags (html, body, etc.)?

xavierdecoster: That's up to the ICP service AFAIK.

xavierdecoster: AFAIK that's up to the ICP service.

xavierdecoster: AFAIK that's up to the ICP service.

Issue #6558, opened 2018-10-12
[Licenses] Feature flag for the Gallery to accept/reject packages with license specified


agr:
Until Gallery and ingestion pipeline is ready to accept packages with license specified in it, we should reject  such packages.

PR #6562, opened 2018-10-13
Configurable rejection of packages with <license>

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/6558

Added configuration option to reject the packages that have <license> metadata node in them in preparation for proper implementation of license metadata processing, so we don't create a backlog of packages to reprocess.

### DIFF
--- a/src/NuGetGallery/Services/PackageUploadService.cs
+++ b/src/NuGetGallery/Services/PackageUploadService.cs
@@ -17,6 +17,7 @@ namespace NuGetGallery
 {
     public class PackageUploadService : IPackageUploadService
     {
+        private const string LicenseNodeName = "license";
         private readonly IPackageService _packageService;
         private readonly IPackageFileService _packageFileService;
         private readonly IEntitiesContext _entitiesContext;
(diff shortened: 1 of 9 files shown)


tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
```diff
@@ -441,6 +440,47 @@ public async Task WithNotTooManyPackageEntries_WhenRejectPackagesWithTooManyPack
                 Assert.Empty(result.Warnings);
             }
 
+            private static string[] LicenseNodeVariants => new string[]
+            {
+                "<license></license>",
```
joelverhagen: maybe also:
`<license />`
and 
`<license>    </license>`

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -70,8 +71,52 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+            }
+
+            public bool HasLicenseMetadata()
+                => MetadataNode
+                    .Elements()
+                    .Any(e => LicenseNodeName == e.Name.LocalName);
+        }
+
+        private PackageValidationResult CheckLicenseMetadata(PackageArchiveReader nuGetPackage, List<string> warnings)
+        {
```
cristinamanum: Is warnings argument used ?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -70,8 +71,52 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+            }
+
+            LicenseCheckingNuspecReader nuspecReader = null;
+
+            using (var nuspec = nuGetPackage.GetNuspec())
+            {
+                nuspecReader = new LicenseCheckingNuspecReader(nuspec);
+            }
+
+            if (nuspecReader.HasLicenseMetadata())
```
cristinamanum: The nuspecReader will be used over a disposed stream(nuspec). Is this invocation ok, or it should be in the using?

agr: `NuspecReader` reads the content of the stream in the constructor ([code](https://github.com/NuGet/NuGet.Client/blob/08954c34ef54bef39d4026a07dde0e0628ca9206/src/NuGet.Core/NuGet.Packaging.Core/NuspecCoreReaderBase.cs#L212)), produces the `XDocument` object from its content and does not use stream anymore (otherwise it would have to implement `IDisposable` itself, which it does not), getting the data from `XDocument`.

tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
agr: Added

PR #6577, opened 2018-10-18
GetEtagAsync

cristinamanum:
Add an api for reading the etag for an azure blob.

### DIFF
--- a/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
@@ -426,6 +426,24 @@ namespace NuGetGallery
             }
         }
 
+        public async Task<string> GetETagAsync(
+            string folderName,
+            string fileName)
+        {
+            folderName = folderName ?? throw new ArgumentNullException(nameof(folderName));
+            fileName = fileName ?? throw new ArgumentNullException(nameof(fileName));
+
+            var container = await GetContainerAsync(folderName);
+            var blob = container.GetBlobReference(fileName);
+
+            if(blob == null)
+            {
+                return null;
+            }
+            await blob.FetchAttributesAsync();
+            return blob.ETag;
+        }
+
         private static SharedAccessBlobPermissions MapFileUriPermissions(FileUriPermissions permissions)
         {
             return (SharedAccessBlobPermissions)permissions;
--- a/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
@@ -426,7 +426,7 @@ namespace NuGetGallery
             }
         }
 
-        public async Task<string> GetETagAsync(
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Services/FileSystemFileStorageService.cs
```diff
@@ -273,6 +273,13 @@ public static string ResolvePath(string fileStorageDirectory)
             return fileStorageDirectory;
         }
 
+        public async Task<string> GetETagAsync(
+           string folderName,
+           string fileName)
+        {
+            return await Task.FromResult<string>(null);
```
joelverhagen: It would be clearer to throw NotImplementedException.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -426,6 +426,24 @@ public async Task<Uri> GetFileReadUriAsync(string folderName, string fileName, D
...
 
+        public async Task<string> GetETagAsync(
+            string folderName,
+            string fileName)
+        {
+            folderName = folderName ?? throw new ArgumentNullException(nameof(folderName));
+            fileName = fileName ?? throw new ArgumentNullException(nameof(fileName));
+
+            var container = await GetContainerAsync(folderName);
+            var blob = container.GetBlobReference(fileName);
```
joelverhagen: When would this return null?

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -426,6 +426,24 @@ public async Task<Uri> GetFileReadUriAsync(string folderName, string fileName, D
             }
         }
 
+        public async Task<string> GetETagAsync(
```
joelverhagen: Don't we need a method that calls this in the package file service?

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
loic-sharma: If the blob doesn't exist, `FetchAttributesAsync` will throw. If the intended behavior is to return null if the blob doesn't exist, I'd rename the method. Maybe something like `GetETagOrNullAsync`

src/NuGetGallery/Services/FileSystemFileStorageService.cs
cristinamanum: Updated.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
cristinamanum: Updated the name and also to catch the case of FetchAttributesAsync throwing.

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
cristinamanum: I would not add it until needed.

Issue #6545, opened 2018-10-09
[Licenses] Gallery package ingestion for packages with embedded license


agr:
Gallery should not reject .nupkg files with embedded licenses.

Issue #6532, opened 2018-10-05
[Licenses] Implement license storage for the standalone Gallery setup


agr:
Standalone third party Gallery setups have to be able to process packages with embedded licenses. Deletions included.

Issue #6529, opened 2018-10-05
[Licenses] Gallery embedded license file validation


agr:
Package verification when .nuspec specifies the presence of embedded license file should:
* Check for file presence;
* Check that the whole file contents is a valid UTF-8 string;
* licenseUrl must point to deprecation URL;
* no license expression is specified at the same time.

Issue #6516, opened 2018-10-05
[Licenses] Gallery DB schema update for embedded license files


agr:
To support embedded license files in NuGet packages, a column has to be added to the Gallery DB:
`EmbeddedLicenseFilename NVARCHAR(max)`

PR #6580, opened 2018-10-18
Embedded license file ingestion for Gallery

agr:
Addresses #6545, #6529, #6516, #6532 

New column was added to GalleryDB to contain the type of the embedded license file (plain text or markdown). Storage is implemented for the case when no async validation pipeline is present.

I ended up diverging from the spec which prescribed storing a presence of an embedded file as a bool flag because this way for the local storage case we lose the document type information. So an enum was introduced that persists that information in DB.

Warning when package does not contain &lt;license&gt; but specifies &lt;licenseUrl&gt;:
image

Warning when package does not contain any license information:
image

License URL rejection message (when enabled):
image

License file rejection message:
image

Warning when license URL is not aligned with license expression:
image


src/NuGetGallery.Core/EmbeddedLicenseFileType.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        /// <summary>
+        /// Indicates that embedded license file is plain text.
+        /// </summary>
+        PlainText = 1,
+
+        /// <summary>
+        /// Indicates that embedded license file is markdown.
+        /// </summary>
+        MarkDown = 2,
```
joelverhagen: `Markdown`

src/NuGetGallery.Core/Services/CorePackageFileService.cs
```diff
@@ -12,6 +12,10 @@ namespace NuGetGallery
 {
     public class CorePackageFileService : ICorePackageFileService
     {
+        private const string ContentTypePlainText = CoreConstants.TextContentType;
```
joelverhagen: nit: why copy a const? Can't we just refer to the original?

src/NuGetGallery.Core/Services/SymbolPackageFileServiceMetadata.cs
```diff
@@ -7,6 +7,9 @@ public class SymbolPackageFileMetadataService : IFileMetadataService
     {
         public string FileFolderName => CoreConstants.SymbolPackagesFolderName;
 
+        public string PackageContentFolderName => throw new System.NotImplementedException();
```
joelverhagen: Nit `using System;` instead of fully qualified

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,15 +80,130 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                {
+                    return PackageValidationResult.Invalid(
+                        string.Format(
+                            Strings.UploadPackage_LicenseFileDoesNotExist,
+                            licenseMetadata.License));
+                }
+
+                // check if specified file has allowed extension
+                var licenseFileExtension = Path.GetExtension(licenseMetadata.License);
+                if (!AllowedLicenseFileExtensions.Contains(licenseFileExtension))
```
joelverhagen: can we have a tracking issue and todo mentioning the issue regarding extensionless license files?

joelverhagen: Should `AllowedLicenseFileExtensions` use case insensitive comparer?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,15 +80,130 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+
+            return byteValue >= TextRangeStart || byteValue == LineFeed || byteValue == CarriageReturn;
+        }
+
+        private bool IsTextStream(Stream stream)
+        {
+            int valueRead = -1;
+            do
+            {
+                valueRead = stream.ReadByte();
```
joelverhagen: How does this perform? Where are the buffers larger than 1 byte, if any?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,15 +80,130 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                }
+            }
+
+            return null;
+        }
+
+        private bool IsTextByte(int byteValue)
+        {
+            const int TextRangeStart = 32;
+            const int LineFeed = 10;
```
joelverhagen: nit: pretty sure you can cast `(int) '\n'` and `(int) '\r'` instead.

Also, looks like we are missing horizontal tab (decimal 9).

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,15 +80,130 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                    }
+                }
+            }
+
+            return null;
+        }
+
+        private bool IsTextByte(int byteValue)
+        {
+            const int TextRangeStart = 32;
```
joelverhagen: AFAIK this accepts UTF-8 but would reject UTF-16. I see the error message calls this out, but the implementation should say this too.

Also, this allows non-ASCII unicode control characters. Expected?

src/NuGetGallery/Services/PackageUploadService.cs
joelverhagen: We are also allowing the DELETE character (127).

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,15 +80,130 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+            const int LineFeed = 10;
+            const int CarriageReturn = 13;
+
+            return byteValue >= TextRangeStart || byteValue == LineFeed || byteValue == CarriageReturn;
+        }
+
+        private bool IsTextStream(Stream stream)
```
joelverhagen: This seems like nice general purpose code. Can we move it out and test the hell out of it on its own?

src/NuGetGallery/Services/PackageUploadService.cs
agr: I am assuming that one of the underlying streams handles the buffering. My only concern here is that naive implementation of the `ReadByte` may run in this case which allocates `new byte[1]` on each call and passes it to the `Read` method.

Regarding the underlying stream: there are 3 cases:
1. UI upload when blob storage is used. In this case our implementation reads blob into the `MemoryStream` which itself is one large buffer;
2. UI upload when file system storage is used. `FileStream` has buffering of its own.
3. API push. `HttpInputStream` is used, which is also buffered.

src/NuGetGallery/Strings.resx
```diff
@@ -1002,4 +1002,37 @@ The {1} Team</value>
   <data name="UploadPackage_NotAcceptingPackagesWithLicense" xml:space="preserve">
     <value>This package contains a &lt;license&gt; metadata which is currently not supported.</value>
   </data>
+  <data name="UploadPackage_DeprecatingLicenseUrl" xml:space="preserve">
+    <value>Specifying external license URLs is being deprecated, please consider switching to specifying the license in the package.</value>
```
loic-sharma: >Specifying external license URLs is being deprecated [](start = 11, length = 52)

I recommend: `Specifying extern license URLs will be deprecated...`. Also, could we link to documentation here?

src/NuGetGallery/Strings.resx
```diff
@@ -1002,4 +1002,37 @@ The {1} Team</value>
...
     <value>This package contains a &lt;license&gt; metadata which is currently not supported.</value>
   </data>
+  <data name="UploadPackage_DeprecatingLicenseUrl" xml:space="preserve">
+    <value>Specifying external license URLs is being deprecated, please consider switching to specifying the license in the package.</value>
+  </data>
+  <data name="UploadPackage_DeprecationUrlRequired" xml:space="preserve">
+    <value>For backwards compatibility when a license is specified in the package, its &lt;licenseUrl&gt; node must point to https://aka.ms/deprecateLicenseUrl</value>
+  </data>
+  <data name="UploadPackage_DeprecationUrlUsage" xml:space="preserve">
+    <value>The license deprecation URL must be used in conjunction with specifying the license in the package.</value>
```
loic-sharma: >The license deprecation URL must be used in conjunction with specifying the license in the package. [](start = 11, length = 99)

This needs a link to documentation. I doubt people will know what "the license deprecation URL" means.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -441,12 +581,37 @@ public async Task<PackageCommitResult> CommitPackageAsync(Package package, Strea
...
         }
+
+        private static async Task SavePackageLicenseFile(Stream packageFile, Func<Stream, Task> saveLicenseAsync)
+        {
+            packageFile.Seek(0, SeekOrigin.Begin);
+            using (var packageArchiveReader = new PackageArchiveReader(packageFile, leaveStreamOpen: true))
+            {
+                var packageMetadata = PackageMetadata.FromNuspecReader(packageArchiveReader.GetNuspecReader(), strict: true);
+                if (packageMetadata.LicenseMetadata != null && packageMetadata.LicenseMetadata.Type == LicenseType.File && !string.IsNullOrWhiteSpace(packageMetadata.LicenseMetadata.License))
+                {
```
loic-sharma: Nit: what do you think of inversing this conditional so that the save flow is less nested?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,15 +80,130 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                    return PackageValidationResult.Invalid(Strings.UploadPackage_LegacyLicenseUrlNotAllowed);
+                }
+                else
+                {
+                    warnings.Add(Strings.UploadPackage_DeprecatingLicenseUrl);
+                }
+            }
+
+            if (licenseMetadata != null && licenseMetadata.Type == LicenseType.Expression)
```
loic-sharma: >licenseMetadata != null && licenseMetadata.Type == LicenseType.Expression [](start = 16, length = 73)

What do you think of `if (licenseMetadata?.Type == LicenseType.Expression)` instead?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -413,7 +537,23 @@ public async Task<PackageCommitResult> CommitPackageAsync(Package package, Strea
                 }
                 else
                 {
-                    await _packageFileService.SavePackageFileAsync(package, packageFile);
+                    if (package.EmbeddedLicenseType != EmbeddedLicenseFileType.Absent)
+                    {
+                        // if the package is immediately made avaialble, it means there is a high chance we don't have
```
loic-sharma: >avaialble [](start = 62, length = 9)

available

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -554,6 +565,9 @@ private static string GetContentType(string folderName)
                 case CoreConstants.UserCertificatesFolderName:
                     return CoreConstants.CertificateContentType;
 
+                case CoreConstants.PackagesContentFolderName:
```
loic-sharma: >PackagesContentFolderName [](start = 35, length = 25)

Don't you need to add a case for `CoreConstants.PackagesContentFolderName` in `GetCacheControl` too?

src/NuGetGallery.Core/Services/CorePackageFileService.cs
```diff
@@ -198,6 +202,74 @@ public async Task StorePackageFileInBackupLocationAsync(Package package, Stream
...
+                throw new ArgumentException("Package must have embedded license", nameof(package));
+            }
+
+            var fileName = BuildLicenseFileName(package);
+
+            // Gallery will generally ignore the content type on license files and will use value from the DB,
+            // but we'll be nice and try to specify correct content type for them.
+            var contentType = package.EmbeddedLicenseType == EmbeddedLicenseFileType.MarkDown ? ContentTypeMarkdown : ContentTypePlainText;
+
+            return _fileStorageService.SaveFileAsync(_metadata.PackageContentFolderName, fileName, contentType, licenseFile, overwrite: false);
```
loic-sharma: >SaveFileAsync [](start = 39, length = 13)

As far as I can tell, `SaveFileAsync` doesn't reset the position of the stream. Should we reset the position of `licenseFile`?

src/NuGetGallery/Services/PackageUploadService.cs
agr: It appears `'\r'` can be implicitly cast to `int`... Updated.

Regarding the delete character: I don't think it has any special meaning, at least in Windows world.

src/NuGetGallery.Core/Services/ICorePackageFileService.cs
```diff
@@ -19,11 +19,21 @@ public interface ICorePackageFileService
         /// </summary>
         Task SavePackageFileAsync(Package package, Stream packageFile, bool overwrite);
 
+        /// <summary>
+        /// Extracts the license file from the package file and saves it to the public container for package content.
```
loic-sharma: >Extracts [](start = 12, length = 8)

This documentation is a little confusing. The license file has already been extracted as the `licenseFile` parameter, right?

agr: Technically, reading the compressed data happens inside the method, so quite likely extraction/decompression happens also at the same time. But yes, you are right, since this is abstracted away by the `Stream` documentation should not be talking about extraction.

src/NuGetGallery/Services/PackageUploadService.cs
agr: I added support straight away.

agr: Yes, it should.

src/NuGetGallery/Services/PackageUploadService.cs
agr: Replaced the `ReadByte` call with `Read`. Will extract the code a bit later.

src/NuGetGallery/Services/PackageUploadService.cs
agr: Done.

src/NuGetGallery.Core/Services/CorePackageFileService.cs
agr: Not sure why would we. Method's purpose is to save what was passed to it, not seek the stream here and there.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,15 +80,130 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+            result = CheckLicenseMetadata(nuGetPackage, warnings);
             if (result != null)
             {
                 return result;
             }
 
             return PackageValidationResult.AcceptedWithWarnings(warnings);
         }
+
+        private PackageValidationResult CheckLicenseMetadata(IPackageCoreReader nuGetPackage, List<string> warnings)
```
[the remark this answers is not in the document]
agr: Extracting license validations only sounds weird. Otherwise, validation code is ~70% of the file, moving all that away sounds like weird thing to do, too.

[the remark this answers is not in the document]
loic-sharma: It seems reasonable to compose the upload service out of many validation services. This has the added benefit that each validation service can be tested individually.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,52 +87,161 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                {
+                    return PackageValidationResult.Invalid(Strings.UploadPackage_LicenseNodeValueTooLong);
+                }
+
+                if (nuspecReader.LicenseHasChildElements())
+                {
+                    return PackageValidationResult.Invalid(Strings.UploadPackage_LicenseNodeContainsChildren);
+                }
+
+                var versionText = nuspecReader.GetLicenseVersion();
```
joelverhagen: Why doesn't this return `System.Version`?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,52 +87,161 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                {
+                    return PackageValidationResult.Invalid(
+                        string.Format(
+                            Strings.UploadPackage_InvalidLicenseFileExtension,
+                            licenseFileExtension,
+                            string.Join(", ", AllowedLicenseFileExtensions.Select(extension => $"'{extension}'"))));
+                }
+
+                var licenseFileEntry = nuGetPackage.GetEntry(licenseMetadata.License);
+                if (licenseFileEntry.Length > MaxAllowedLicenseLength)
```
joelverhagen: Is it possible for the zip entry length to be a lie? For example, central directory says 100 bytes but it actually decompresses to 1100 megabytes?

src/NuGetGallery/Services/PackageUploadService.cs
joelverhagen: Oh, nevermind. We wrote this method. String is fine.

src/NuGetGallery/Strings.resx
```diff
@@ -1002,4 +1002,51 @@ The {1} Team</value>
   <data name="UploadPackage_NotAcceptingPackagesWithLicense" xml:space="preserve">
     <value>This package contains a &lt;license&gt; metadata which is currently not supported.</value>
   </data>
+  <data name="UploadPackage_DeprecatingLicenseUrl" xml:space="preserve">
+    <value>Specifying external license URLs will be deprecated, please consider switching to specifying the license in the package.</value>
```
loic-sharma: >package [](start = 123, length = 7)

It looks like my previous comment was dropped by new changes. Readding it here:

>... could we link to the documentation in all of these error messages?

src/NuGetGallery/Helpers/TextHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                throw new ArgumentException($"{nameof(stream)} argument must be a readable stream", nameof(stream));
+            }
+
+            if (bufferSize <= 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(bufferSize), $"{nameof(bufferSize)} must be greater than 0");
+            }
+
+            byte[] buffer = new byte[bufferSize];
```
loic-sharma: >byte[] [](start = 12, length = 6)

`var`

src/NuGetGallery/Helpers/TextHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// Checks if the stream contains only bytes that are OK to have in a UTF-8 encoded plain text file.
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <returns></returns>
+        /// <remarks>
+        /// This method does NOT check if file contains valid UTF-8 sequences. The purpose is to be able to detect obviously binary files.
+        /// Method will read from the stream until end of stream is encountered. It is caller's responsibility to resolve all potential
+        /// concerns related to the size of the stream.
+        /// </remarks>
+        public static bool IsUtf8TextStream(Stream stream, int bufferSize = 1024)
```
loic-sharma: >IsUtf8TextStream [](start = 27, length = 16)

Considering this isn't a perfect test, what do you think of a name like `LooksLIkeUtf8TextStream`?

Also, should this method rewind the stream to the beginning? If not, maybe call this out in the `remarks`?

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -575,6 +578,48 @@ public virtual IEnumerable<NuGetFramework> GetSupportedFrameworks(PackageArchive
...
+            switch (packageMetadata.LicenseMetadata.Type)
+            {
+                case LicenseType.File:
+                    return GetEmbeddedLicenseType(packageMetadata.LicenseMetadata.License);
+
+                case LicenseType.Expression:
+                    // execution shouldn't be able to reach here: validation should catch it first
+                    // so if we get here, it's a bug
+                    throw new NotImplementedException("License expression support is not implemented yet");
+
```
loic-sharma: Is this correct? The PackageUploadService accepts license expressions

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -413,7 +538,23 @@ public async Task<PackageCommitResult> CommitPackageAsync(Package package, Strea
                 }
                 else
                 {
-                    await _packageFileService.SavePackageFileAsync(package, packageFile);
+                    if (package.EmbeddedLicenseType != EmbeddedLicenseFileType.Absent)
+                    {
+                        // if the package is immediately made available, it means there is a high chance we don't have
+                        // validation pipeline that would normally store the license file, so we'll do it ourselves here.
+                        await SavePackageLicenseFile(packageFile, licenseStream => _packageFileService.SaveLicenseFileAsync(package, licenseStream));
```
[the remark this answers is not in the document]
agr: We need to get the license file stream before we can call the `SaveLicenseFileAsync`, that's what `SavePackageLicenseFile` method does: get's the file stream from the zip archive. Then we have a choice: we can call `SaveLicenseFileAsync` from the `SavePackageLicenseFile`, but to do that we need to pass a `Package` object into `SavePackageLicenseFile` which will not use it in any way, or pass a delegate that just accepts a stream and does whatever necessary to save it (including incapsulating all the extra knowledge it needs). I chose latter, as a bonus, it allowed me to make the method `static`.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,52 +87,161 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                {
+                    return PackageValidationResult.Invalid(Strings.UploadPackage_LegacyLicenseUrlNotAllowed);
+                }
+                else
+                {
+                    warnings.Add(Strings.UploadPackage_DeprecatingLicenseUrl);
+                }
+            }
+
+            if (licenseMetadata != null && IsLicenseDeprecationUrl(licenseUrl))
```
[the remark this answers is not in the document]
agr: Incomplete refactoring. Essentially, the situation with deprecation URLs is a bit unclear at this moment and in the future we might need a bit more complex logic for determining if it is a deprecation URL or not. I wanted to have the check in a single location, so in the future, I would not need to hunt down all uses to change the check.

src/NuGetGallery/Services/PackageService.cs
agr: Fixed

src/NuGetGallery/Helpers/TextHelper.cs
agr: Are there any precedents for methods that read from stream to rewind it after? The assumption that stream can be rewound is pretty strong one.

Renamed the method.

src/NuGetGallery/Services/PackageUploadService.cs
agr: I ran some tests, and it seems .net's `ZipArchive` is not protected against it. I managed to create an archive that reports one file size, but when read produces much more. All archivers I have on my computer fail one way or another, but `ZipArchive` reads it just fine. I'll add some validations around it.

agr: https://github.com/dotnet/corefx/issues/33058

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,52 +87,190 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+            // so we'll try reading the stream ourselves and check if reported length matches the actual.
+
+            var buffer = new byte[4096];
+            long totalBytesRead = 0;
+            int read = 0;
+            do
+            {
+                read = licenseFileStream.Read(buffer, 0, buffer.Length);
+                totalBytesRead += read;
+            } while (read > 0 && totalBytesRead < reportedLength + 1); // we want to try to read past the reported length
```
loic-sharma: 👍

src/NuGetGallery/Strings.resx
agr: Done.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -71,52 +87,190 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+            // Unfortunately, .Net's ZipArchive is not handling this case properly and allows to read full
+            // data without throwing any exceptions.
+            // so we'll try reading the stream ourselves and check if reported length matches the actual.
+
+            var buffer = new byte[4096];
+            long totalBytesRead = 0;
+            int read = 0;
+            do
+            {
+                read = licenseFileStream.Read(buffer, 0, buffer.Length);
```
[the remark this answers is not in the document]
agr: Yeah, good point.

src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -203,7 +203,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="NuGet.Services.Entities">
-      <Version>2.31.0</Version>
+      <Version>2.32.0-agr-license-2170862</Version>
```
agr: Will update it to release version once this PR gets better approval.

src/NuGetGallery/Helpers/TextHelper.cs
loic-sharma: I meant rewind *before* you read the stream. I agree it's a strong assumption that you can rewind the stream.

src/NuGetGallery/GalleryConstants.cs
```diff
@@ -106,6 +106,8 @@ public static class GalleryConstants
 
         public const string GitRepository = "git";
 
+        public const string LicenseDeprecationUrl = "https://aka.ms/deprecateLicenseUrl";
```
loic-sharma: >https://aka.ms/deprecateLicenseUrl [](start = 53, length = 34)

FYI, this link redirects to nuget's documentation home page. Is that intended?

src/NuGetGallery/Strings.resx
```diff
@@ -1005,4 +1005,69 @@ The {1} Team</value>
...
+  </data>
+  <data name="UploadPackage_DeprecationUrlUsage" xml:space="preserve">
+    <value>The license deprecation URL must be used in conjunction with specifying the license in the package.</value>
+  </data>
+  <data name="UploadPackage_InvalidLicenseMetadata" xml:space="preserve">
+    <value>Invalid license metadata: {0}</value>
+    <comment>{0} is the list of the issues found with license expression.</comment>
+  </data>
+  <data name="UploadPackage_LegacyLicenseUrlNotAllowed" xml:space="preserve">
+    <value>Specifying &lt;licenseUrl&gt; in package metadata is not allowed anymore, please specify the license in the package.</value>
```
loic-sharma: > is not allowed anymore [](start = 60, length = 23)

What do you think of `is no longer allowed` instead of `is not allowed anymore`?

src/NuGetGallery/Strings.resx
```diff
@@ -1005,4 +1005,69 @@ The {1} Team</value>
...
+  </data>
+  <data name="UploadPackage_LicenseFileDoesNotExist" xml:space="preserve">
+    <value>License file '{0}' does not exist in the package.</value>
+    <comment>{0} is the license file name specified in the .nuspec</comment>
+  </data>
+  <data name="UploadPackage_LicenseMustBePlainText" xml:space="preserve">
+    <value>License file must be plain text using UTF-8 encoding.</value>
+  </data>
+  <data name="UploadPackage_LicenseFileTooLong" xml:space="preserve">
+    <value>License file is too long. The limit is {0} bytes.</value>
```
loic-sharma: >bytes [](start = 54, length = 5)

Is bytes the best unit to use here? What about megabytes?

src/NuGetGallery/Strings.resx
```diff
@@ -1005,4 +1005,69 @@ The {1} Team</value>
...
+    <comment>{0} is the max allowed number of bytes in the license.</comment>
+  </data>
+  <data name="UploadPackage_LicenseNodeContainsChildren" xml:space="preserve">
+    <value>License node cannot contain children nodes.</value>
+  </data>
+  <data name="UploadPackage_LicenseNodeValueTooLong" xml:space="preserve">
+    <value>License node value is too long. Please make sure it is under 500 characters long.</value>
```
loic-sharma: >License node value is too long. Please make sure it is under 500 characters long.</ [](start = 11, length = 83)

How about `The license node must be under 500 characters`? By the way, I think may of these messages should have a `The ` at the beginning.

src/NuGetGallery/Configuration/IAppConfiguration.cs
```diff
@@ -370,5 +370,16 @@ public interface IAppConfiguration : IMessageServiceConfiguration
         /// Flag that indicates whether packages with `license` node in them should be rejected.
         /// </summary>
         bool RejectPackagesWithLicense { get; set; }
+
+        /// <summary>
+        /// Indicates whether packages that specify the licene the "old" way (with a "licenseUrl" node only) should be rejected.
```
loic-sharma: >licene [](start = 56, length = 6)

license

src/NuGetGallery/Services/LicenseUrlDeprecationValidationMessage.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Web;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// A generic license url deprecation message that appends a "Learn more" documentation link to a given base message.
+    /// </summary>
+    public class LicenseUrlDeprecationValidationMessage : IValidationMessage
+    {
+        private readonly string DeprecationLink = $"<a href=\"{GalleryConstants.LicenseDeprecationUrl}\">{Strings.UploadPackage_LearnMore}.</a>";
```
loic-sharma: >DeprecationLink [](start = 32, length = 15)

Should this be static?

src/NuGetGallery/Services/LicenseUrlDeprecationValidationMessage.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        private readonly string _baseMessage;
+        
+        public LicenseUrlDeprecationValidationMessage(string basePlainTextMessage)
+        {
+            if (string.IsNullOrWhiteSpace(basePlainTextMessage))
+            {
+                throw new ArgumentException($"{nameof(basePlainTextMessage)} must be non-empty string", nameof(basePlainTextMessage));
```
loic-sharma: >$"{nameof(basePlainTextMessage)} must be non-empty string", nameof(basePlainTextMessage)); [](start = 44, length = 90)

I think you can use `Strings.ParameterCannotBeNullOrEmpty` here.

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -461,6 +461,7 @@ private async Task BackupPackageBinaries(IEnumerable<Package> packages)
                             : package.NormalizedVersion;
 
                 await _packageFileService.DeletePackageFileAsync(id, version);
+                await _packageFileService.DeleteLicenseFileAsync(id, version);
```
loic-sharma: >DeleteLicenseFileAsync [](start = 42, length = 22)

Why don't we backup the license file? Please add a comment if this is by design

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -83,42 +105,231 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                if (versionText != null && AllowedLicenseVersion != versionText)
+                {
+                    return PackageValidationResult.Invalid(
+                        string.Format(
+                            Strings.UploadPackage_UnsupportedLicenseVersion,
+                            versionText));
+                }
 
+                // TODO: remove when all pipeline changes are done
+                if (LicenseType.File.ToString().Equals(typeText, StringComparison.OrdinalIgnoreCase))
```
loic-sharma: Does this have a tracking issue?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -83,42 +105,231 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                    {
+                        return PackageValidationResult.Invalid(Strings.UploadPackage_LicenseMustBePlainText);
+                    }
+                }
             }
 
             return null;
         }
 
+        private static async Task<bool> IsStreamLengthMatchesReportedAsync(Stream licenseFileStream, long reportedLength)
```
loic-sharma: >IsStreamLengthMatchesReportedAsync [](start = 40, length = 34)

This could be a nice `Stream` extension method. Maybe you could make a `Stream` extension class that also has `LooksLikeUtf8TextStreamAsync`

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -83,42 +105,231 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+            } while (read > 0 && totalBytesRead < reportedLength + 1); // we want to try to read past the reported length
+
+            return totalBytesRead == reportedLength;
+        }
+
+        private static string GetExpectedLicenseUrl(LicenseMetadata licenseMetadata)
+        {
+
+            if (licenseMetadata == null || LicenseType.File == licenseMetadata.Type)
+            {
```
loic-sharma: There's an extra line above here

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -83,42 +105,231 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                return GalleryConstants.LicenseDeprecationUrl;
+            }
+
+            if (LicenseType.Expression == licenseMetadata.Type)
+            {
+                var expectedUrl = new Uri(string.Format(LicenseExpressionDeprecationUrlFormat, licenseMetadata.License)).AbsoluteUri;
+                return expectedUrl;
```
loic-sharma: >expectedUrl [](start = 23, length = 11)

Nit: return the `expectedUrl` directly instead of storing it in a variable?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -83,42 +105,231 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+            }
+
+            throw new InvalidOperationException($"Unsupported license metadata type: {licenseMetadata.Type}");
+        }
+
+        private static bool HasChildElements(XElement xElement)
```
loic-sharma: >HasChildElements [](start = 28, length = 16)

What do you think of mocing `HasChildElements`, `GetLicenseVersion` and `GetLicenseType` into `LicenseCheckingNuspecReader`?

agr: I specifically moved them out not to call `LicenseCheckingNuspecReader.LicenseElement` repeatedly.

src/NuGetGallery/Strings.resx
agr: Mebibytes? Bytes are precise. And there is no agreement on how many in one megabyte.

src/NuGetGallery/Strings.resx
agr: Value is too long, not the node, and it is value *length* that should be under 500 (501 actually), not the value. How about `The license node value must be shorter than 500 characters`?

agr: Also sprinkled `The` all around new messages.

src/NuGetGallery/Services/LicenseUrlDeprecationValidationMessage.cs
agr: Why would I use `Strings` in a argument validation code?

src/NuGetGallery/Services/PackageUploadService.cs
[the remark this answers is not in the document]
agr: Yes, full support for license files is not in yet.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -83,42 +105,231 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
             return PackageValidationResult.AcceptedWithWarnings(warnings);
         }
 
-        private class LicenseCheckingNuspecReader : NuspecReader
+        private async Task<PackageValidationResult> CheckLicenseMetadataAsync(PackageArchiveReader nuGetPackage, List<IValidationMessage> warnings)
```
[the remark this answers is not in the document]
agr: Introducing new service was discussed and it is out of scope. We probably should move all validation code out of `PackageUploadService` and I am not doing that.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -5,21 +5,44 @@
...
+        private static readonly IReadOnlyCollection<string> AllowedLicenseTypes = new HashSet<string>
+        {
+            LicenseType.File.ToString(),
+            LicenseType.Expression.ToString()
+        };
+
         private const string LicenseNodeName = "license";
+        private const string AllowedLicenseVersion = "1.0.0";
+        private const string Unlicensed = "UNLICENSED";
+        private const int MaxAllowedLicenseLength = 1024 * 1024;
```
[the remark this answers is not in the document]
agr: It's an arbitrarily large number of bytes so that it is not too large to be able to process it fast, and larger than any sensible license.

src/NuGetGallery/Helpers/TextHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                throw new ArgumentException($"{nameof(stream)} argument must be a readable stream", nameof(stream));
+            }
+
+            if (bufferSize <= 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(bufferSize), $"{nameof(bufferSize)} must be greater than 0");
+            }
+
+            var buffer = new byte[bufferSize];
```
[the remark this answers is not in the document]
agr: Added more remarks to both methods in that file.

src/NuGetGallery/Services/PackageUploadService.cs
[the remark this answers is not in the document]
scottbommarito: You should mention this in a comment. It should be clear that this number was generated arbitrarily.

src/NuGetGallery/Services/PackageUploadService.cs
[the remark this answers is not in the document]
scottbommarito: I would argue that it IS in scope, because you are doubling the size of the file, and a lot of package validation (but not all) is already in its own service.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -83,42 +105,231 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+                    {
+                        return PackageValidationResult.Invalid(new LicenseUrlDeprecationValidationMessage(Strings.UploadPackage_MissingLicenseInformation));
+                    }
+                    else
+                    {
+                        warnings.Add(new LicenseUrlDeprecationValidationMessage(Strings.UploadPackage_LicenseShouldBeSpecified));
+                    }
+                }
+
+                if (licenseDeprecationUrl == licenseUrl)
```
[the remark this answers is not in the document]
agr: URLs in general *are* case sensitive. The fact that `aka.ms` treats them specially should not have any effect on us (they might stop to do it tomorrow).
Operator `==` calls `String.Equals(a, b)`. What issues do you expect?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -83,42 +105,231 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+
+        private static string GetLicenseVersion(XElement licenseElement)
+            => licenseElement
+                .GetOptionalAttributeValue("version");
+
+        private static string GetLicenseType(XElement licenseElement)
+            => licenseElement
+                .GetOptionalAttributeValue("type");
+
+        private class LicenseCheckingNuspecReader : NuspecReader
```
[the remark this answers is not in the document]
agr: Why? It is only useful in the context of validation, I don't want to make it visible to all the code.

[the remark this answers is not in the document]
scottbommarito: This file is huge. Anything you can do to split it up into more logical parts would improve the readability of it.

src/NuGetGallery/Services/PackageUploadService.cs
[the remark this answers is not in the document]
scottbommarito: Ok, that's fine then

src/NuGetGallery/Services/PackageUploadService.cs
scottbommarito: Ideally, you could move all the functions into `LicenseCheckingNuspecReader` such that you can make `LicenseElement` private.

```
private class LicenseCheckingNuspecReader : NuspecReader
{
    private XElement _licenseElement => MetadataNode.Element(MetadataNode.Name.Namespace + LicenseNodeName);

    public string GetLicenseVersion() => _licenseElement.GetOptionalAttributeValue("version");

    /* other functions here */
}
```

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -83,42 +105,231 @@ public async Task<PackageValidationResult> ValidateBeforeGeneratePackageAsync(Pa
...
+            // Unfortunately, .Net's ZipArchive is not handling this case properly and allows to read full
+            // data without throwing any exceptions.
+            // so we'll try reading the stream ourselves and check if reported length matches the actual.
+
+            var buffer = new byte[4096];
+            long totalBytesRead = 0;
+            int read = 0;
+            do
+            {
+                read = await licenseFileStream.ReadAsync(buffer, 0, buffer.Length);
```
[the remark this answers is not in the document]
agr: I fail to imagine a function that would suit both use cases and make the code more readable at the same time.

src/NuGetGallery/Services/LicenseUrlDeprecationValidationMessage.cs
agr: I didn't want it to sit in a memory when class is not used. I'll make it a property.

src/NuGetGallery/Strings.resx
loic-sharma: What? Maybe "Mb" vs "MB" has no agreement, but there are clearly 1024 bytes in a megabyte. If the limit is greater than 1024 bytes, we should display this as megabytes.

src/NuGetGallery/Services/PackageDeleteService.cs
loic-sharma: Fair enough. I'd add a comment as it's not readily apparent that one may need extract the license.

src/NuGetGallery/Services/PackageUploadService.cs
agr: I don't know what is the cost of calling `MetadataNode.Element(MetadataNode.Name.Namespace + LicenseNodeName)` multiple times and I don't want to care about it: I call it once and reuse the result as much as possible. If I make them members of `LicenseCheckingNuspecReader` then I'd have to make that `Element` call in each method or do something atrocious to keep the value between calls.

src/NuGetGallery/Services/LicenseUrlDeprecationValidationMessage.cs
agr: I see it is already used here and there, so I added it. But we should figure out our exception messaging story better.

src/NuGetGallery/Services/PackageUploadService.cs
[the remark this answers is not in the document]
agr: The file is under 1000 lines. I've see methods longer than that :)

src/NuGetGallery.Core/Services/CorePackageFileService.cs
```diff
@@ -13,6 +13,8 @@ namespace NuGetGallery
 {
     public class CorePackageFileService : ICorePackageFileService
```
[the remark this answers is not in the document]
agr: Created an engineering issue for that.

src/NuGetGallery/Services/PackageUploadService.cs
agr: I tried to do that and rolled back the change. It looks ugly.

It makes `LicenseCheckingNuspecReader` to provide methods that check the content of the license element in various ways  (a methods to check if it exists, a method to check its length, a method to check if there are child elements, etc.). Class methods cannot assume licenseElement is not null, so this check creeps into every method whereas in the original code its a single check.

src/NuGetGallery/Services/PackageUploadService.cs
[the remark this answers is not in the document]
agr: I created work item to refactor the `PackageUploadService`, we'll address these issues when we'd be working on it.

src/NuGetGallery/Strings.resx
agr: [Mebibyte](https://en.wikipedia.org/wiki/Mebibyte). It's a thing.

agr: Changed to use `ToUserFriendlyBytesLabel` extension.

src/NuGetGallery/Services/PackageUploadService.cs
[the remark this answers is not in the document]
agr: Added comment.

Issue #6100, opened 2018-06-26
[MSA/AAD] Add more logging for exceptions


shishirx34:
AI shows `/authenticate` route returning 500, however there aren't any correlated exceptions. The authentication middleware could possibly be throwing an exception(authentication issue) which doesn't get logged and doesn't redirect to app code resulting in 500. 

We also have customers reporting seeing the 500 during login with AAD. Adding more logging for exceptions could help us investigate this issue.

PR #6581, opened 2018-10-19
[MSA] Pass the error messages from AADv2 to gallery

shishirx34:
Addresses #6100 - It is likely that the users were seeing 500 due to unhandled exceptions in AADv2 OIDC middleware. With this change, we will handle all the exceptions/failures that might occur during authentication and surface the received error message to the user for easier analysis and debugging.

### DIFF
--- a/src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
+++ b/src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
@@ -48,6 +48,12 @@ namespace NuGetGallery.Authentication.Providers.AzureActiveDirectoryV2
         private static HashSet<string> _alternateSiteRootList;
         private const string SELECT_ACCOUNT = "select_account";
 
+        private static class ErrorQueryKeys
+        {
+            public const string Error = "error";
+            public const string ErrorDescription = "errorDescription";
+        }
+
         /// <summary>
         /// The possible values returned by <see cref="V2Claims.ACR"/> claim, and also the possible token values to be sent
         /// for authentication to the common endpoint.
@@ -199,7 +205,7 @@ namespace NuGetGallery.Authentication.Providers.AzureActiveDirectoryV2
         // error handling is done.
         private Task AuthenticationFailed(AuthenticationFailedNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions> notification)
         {
-            // For every 'Challenge' sent to the external providers, we store the 'State'
+            // For every 'Challenge' sent to the external providers, we pass the 'State'
             // with the redirect uri where we intend to return after successful authentication.
             // Extract this "RedirectUri" property from this "State" object for redirecting on failed authentication as well.
             var authenticationProperties = GetAuthenticationPropertiesFromProtocolMessage(notification.ProtocolMessage, notification.Options);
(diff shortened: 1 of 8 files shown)


shishirx34:
@agr - this will possibly address the issues you were seeing with customers for AAD/MSA login.

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -200,16 +199,23 @@ public override bool SupportsMultiFactorAuthentication()
         // error handling is done.
         private Task AuthenticationFailed(AuthenticationFailedNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions> notification)
         {
-            if (_errorMessageList.Contains(notification.Exception.Message))
+            // For every 'Challenge' sent to the external providers, we store the 'State'
```
agr: Should it say "pass" instead of "store"?
>  With every 'Challenge' sent to the external providers, we **pass** the 'State'

I.e. at least in my mind "storing" implies saving something on our side.

shishirx34: old comment but sure..

src/NuGetGallery/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -200,16 +199,23 @@ public override bool SupportsMultiFactorAuthentication()
...
+            // Pass the errors as the query string to show appropriate message to the user.
+            var queryString = new Dictionary<string, string>()
             {
-                // For every 'Challenge' sent to the external providers, we store the 'State'
-                // with the redirect uri where we intend to return after successful authentication.
-                // Extract this "RedirectUri" property from this "State" object for redirecting on failed authentication as well.
-                var authenticationProperties = GetAuthenticationPropertiesFromProtocolMessage(notification.ProtocolMessage, notification.Options);
+                { "error", notification.ProtocolMessage.Error },
+                { "errorDescription", notification.ProtocolMessage.ErrorDescription }
+            };
```
agr: Constants for dictionary keys?

src/NuGetGallery/Helpers/UriExtensions.cs
```diff
@@ -57,5 +59,19 @@ public static Uri ToHttps(this Uri uri)
 
             return uriBuilder.Uri;
         }
+
+        public static string AppendQueryStringToRelativeUri(string relativeUrl, IDictionary<string, string> queryStringCollection)
```
agr: >IDictionary<string, string> queryStringCollection [](start = 80, length = 49)

`queryStringCollection` should probably be declared as `IReadOnlyCollection<KeyValuePair<string>>` instead of dictionary. That will also reduce the complexity of the foreach loop over it, since it would remove lookups.

src/NuGetGallery/Helpers/UriExtensions.cs
```diff
@@ -57,5 +59,19 @@ public static Uri ToHttps(this Uri uri)
...
+        {
```
agr: Need tests and documentation.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -749,11 +756,11 @@ private string GetEmailAddressFromExternalLoginResult(AuthenticateExternalLoginR
...
         }
 
-        private ActionResult ExternalLinkExpired()
+        private ActionResult AuthenticationFailureOrExternalLinkExpired(string errorMessage = null)
         {
             // User got here without an external login cookie (or an expired one)
             // Send them to the logon action with a message
-            TempData["Message"] = Strings.ExternalAccountLinkExpired;
+            TempData["Message"] = string.IsNullOrEmpty(errorMessage) ? Strings.ExternalAccountLinkExpired : errorMessage;
             return Redirect(Url.LogOn(null, relativeUrl: false));
```
agr: Do we need to log this message or passing it to user is enough?

shishirx34: I think for most cases its going to be the same for the users. Simply passing it to users should be enough to give them an idea of what to do. If not they can reach out to us with appropriate messages. I don't see the need for logging here.

PR #6583, opened 2018-10-19
 Add custom query telemetry and response header to OData endpoint

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/1798.

This PR does two things:

1. Add custom metric `ODataCustomQuery` to OData queries.
   - This has a property `IsCustomQuery = True | False | Unknown`
1. Add a response header `X-NuGet-CustomQuery: true | false`
   - This can be used by APIM

The telemetry will allow us to verify that official clients are not doing custom queries. Note that "custom query" does not necessarily mean executed without the database. I chose to make this distinction for the `GetUpdates` OData function. This can be called by official client (2.x for many cases, 3.x+ for `nuget.exe update -self`) so it should not be throttled like custom OData queries.

~Another case: when search hijack throws an exception, some (maybe all?) codepaths fall back to DB. This should probably not be throttled...~

### DIFF
--- a/src/NuGetGallery/Controllers/ODataV1FeedController.cs
+++ b/src/NuGetGallery/Controllers/ODataV1FeedController.cs
@@ -30,8 +30,9 @@ namespace NuGetGallery.Controllers
         public ODataV1FeedController(
             IEntityRepository<Package> packagesRepository,
             IGalleryConfigurationService configurationService,
-            ISearchService searchService)
-            : base(configurationService)
+            ISearchService searchService,
+            ITelemetryService telemetryService)
+            : base(configurationService, telemetryService)
         {
             _packagesRepository = packagesRepository;
             _configurationService = configurationService;
@@ -56,7 +57,7 @@ namespace NuGetGallery.Controllers
                                 .WithoutSortOnColumn(Id, ShouldIgnoreOrderById(options))
                                 .ToV1FeedPackageQuery(_configurationService.GetSiteRoot(UseHttps()));
 
-            return QueryResult(options, queryable, MaxPageSize);
+            return TrackedQueryResult(options, queryable, MaxPageSize, customQuery: true);
         }
 
         // /api/v1/Packages/$count
@@ -108,6 +109,8 @@ namespace NuGetGallery.Controllers
                 packages = packages.Where(p => p.Version == version);
             }
 
+            bool? customQuery = null;
+
             // try the search service
             try
             {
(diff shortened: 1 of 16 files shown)


joelverhagen:
> Another case: when search hijack throws an exception, some (maybe all?) codepaths fall back to DB. This should probably not be throttled...

I've solved this case with an "Unknown" state. This is registered as "Unknown" in telemetry and no `X-NuGet-CustomQuery` header in the response body.

src/NuGetGallery/Constants.cs
```diff
@@ -87,6 +87,7 @@ public static class Constants
         internal const string NuGetProtocolHeaderName = "X-NuGet-Protocol-Version";
         internal const string WarningHeaderName = "X-NuGet-Warning";
         internal const string UserAgentHeaderName = "User-Agent";
+        internal const string CustomQueryHeaderName = "X-NuGet-CustomQuery";
```
loic-sharma: >CustomQueryHeaderName [](start = 30, length = 21)

This name is opaque without any internal NuGet knowledge. A comment describing this constant would go a long way :)

src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -130,15 +135,25 @@ private async Task<IHttpActionResult> GetCore(ODataQueryOptions<V1FeedPackage> o
...
                         return NotFound();
                     }
 
                     var pagedQueryable = packages
                         .Take(options.Top != null ? Math.Min(options.Top.Value, MaxPageSize) : MaxPageSize)
                         .ToV1FeedPackageQuery(GetSiteRoot());
 
-                    return QueryResult(options, pagedQueryable, MaxPageSize, totalHits, (o, s, resultCount) =>
-                       SearchAdaptor.GetNextLink(Request.RequestUri, resultCount, new { id }, o, s));
+                    return QueryResult(
```
loic-sharma: >QueryResult [](start = 27, length = 11)

I was a little confused as it looks like the telemetry service isn't called in this path. What do you think of renaming `QueryResult` to `TrackedQueryResult`?

src/NuGetGallery/Constants.cs
joelverhagen: Done.

src/NuGetGallery/Controllers/ODataV1FeedController.cs
joelverhagen: Done.

joelverhagen:
Tried this out on DEV environment. 

```xml
<rate-limit-by-key
    calls="10"
    renewal-period="60"
    counter-key="@(context.Request.IpAddress)"
    increment-condition="@(context.Response.Headers.GetValueOrDefault("X-NuGet-CustomQuery", "false") == "true")" />
```

Works great!

joelverhagen:
I would like to ship this to PROD _without_ the throttle rule and observe the telemetry. Any objections?

PR #6588, opened 2018-10-23
Consume NuGet.Services.Entities and NuGet.Services.Messaging.Email

xavierdecoster:
This PR:
* Moves out the NuGetGallery entities and consumes _NuGet.Services.Entities_ instead
* Moves out any tests that have been moved to `ServerCommon.sln`
* Moves out any email messaging logic from `NuGetGallery.Core` and consumes `NuGet.Services.Messaging.Email` instead

Hopefully, this will also help us limit the dependency on `NuGetGallery.Core` elsewhere :)


src/NuGetGallery.Core/CoreConstants.cs
```diff
@@ -7,8 +7,6 @@ public static class CoreConstants
     {
         public const string AdminRoleName = "Admins";
 
-        public const int MaxPackageIdLength = 128;
```
xavierdecoster: This is now a constant in `NuGet.Services.Entities`.

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -1442,7 +1444,6 @@
     <Compile Include="Services\IPackageService.cs" />
     <Compile Include="Services\PackageService.cs" />
     <Compile Include="Configuration\AppConfiguration.cs" />
-    <Compile Include="Constants.cs" />
```
xavierdecoster: Renamed `NuGetGallery.Constants` to `GalleryConstants` to reduce/avoid type collisions in views, as both `NuGetGalllery` and `NuGet.Services.Entities` are known namespaces in each razor view.

PR #6597, opened 2018-10-25
Typosquatting optimization: add caching

zhhyu:
Add the cache of checklist to speed up.
Fix: https://github.com/NuGet/Engineering/issues/1809

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -331,7 +331,7 @@ namespace NuGetGallery
             builder.RegisterType<TyposquattingService>()
                 .AsSelf()
                 .As<ITyposquattingService>()
-                .InstancePerLifetimeScope();
+                .SingleInstance();
 
             RegisterMessagingService(builder, configuration);
 
--- a/src/NuGetGallery/Services/TyposquattingService.cs
+++ b/src/NuGetGallery/Services/TyposquattingService.cs
@@ -19,8 +19,12 @@ namespace NuGetGallery
             new ThresholdInfo (lowerBound: 30, upperBound: 50, threshold: 1),
             new ThresholdInfo (lowerBound: 50, upperBound: 129, threshold: 2)
         };
+        private static readonly object Locker = new object();
 
         private static int TyposquattingCheckListLength;
+        private List<string> _checkListCache = null;
+        private DateTime _checkListCacheLastRefreshTime = DateTime.UtcNow;
+        private TimeSpan _checkListCacheDefaultExpireTime = TimeSpan.FromDays(1);
 
         private readonly IContentObjectService _contentObjectService;
         private readonly IPackageService _packageService;
@@ -41,37 +45,47 @@ namespace NuGetGallery
         {
             typosquattingCheckCollisionIds = new List<string>();
             var wasUploadBlocked = false;
+
             if (!_contentObjectService.TyposquattingConfiguration.IsCheckEnabled || _reservedNamespaceService.GetReservedNamespacesForId(uploadedPackageId).Any())
             {
                 return wasUploadBlocked;
             }
-
             if (uploadedPackageId == null)
             {
                 throw new ArgumentNullException(nameof(uploadedPackageId));
             }
-
             if (uploadedPackageOwner == null)
             {
                 throw new ArgumentNullException(nameof(uploadedPackageOwner));
             }
 
-            var checklistRetrievalStopwatch = Stopwatch.StartNew();
-            var packageRegistrations = _packageService.GetAllPackageRegistrations();
-            var packagesCheckList = packageRegistrations
-                .OrderByDescending(pr => pr.IsVerified)
-                .ThenByDescending(pr => pr.DownloadCount)
-                .Select(pr => pr.Id)
-                .Take(TyposquattingCheckListLength)
-                .ToList();
-            checklistRetrievalStopwatch.Stop();
+            Stopwatch checklistRetrievalStopwatch = null;
+            if (_checkListCache == null || IsCheckListCacheExpired())
+            {
+                lock(Locker)
+                {
+                    if (_checkListCache == null || IsCheckListCacheExpired())
+                    {
+                        checklistRetrievalStopwatch = Stopwatch.StartNew();
+                        _checkListCache = _packageService.GetAllPackageRegistrations()
+                            .OrderByDescending(pr => pr.IsVerified)
+                            .ThenByDescending(pr => pr.DownloadCount)
+                            .Select(pr => pr.Id)
+                            .Take(TyposquattingCheckListLength)
+                            .ToList();
+                        checklistRetrievalStopwatch.Stop();
+
+                        _checkListCacheLastRefreshTime = DateTime.UtcNow;
+                    }
+                }
+            }
+            var packageIdsCheckList = _checkListCache;
 
             var algorithmProcessingStopwatch = Stopwatch.StartNew();
             var threshold = GetThreshold(uploadedPackageId);
             var normalizedUploadedPackageId = TyposquattingStringNormalization.NormalizeString(uploadedPackageId);
-
             var collisionIds = new ConcurrentBag<string>();
-            Parallel.ForEach(packagesCheckList, (packageId, loopState) =>
+            Parallel.ForEach(packageIdsCheckList, (packageId, loopState) =>
             {
                 string normalizedPackageId = TyposquattingStringNormalization.NormalizeString(packageId);
                 if (TyposquattingDistanceCalculation.IsDistanceLessThanThreshold(normalizedUploadedPackageId, normalizedPackageId, threshold))
@@ -79,12 +93,16 @@ namespace NuGetGallery
                     collisionIds.Add(packageId);
                 }
             });
-
             algorithmProcessingStopwatch.Stop();
 
-            var totalTime = checklistRetrievalStopwatch.Elapsed.Add(algorithmProcessingStopwatch.Elapsed);
-            _telemetryService.TrackMetricForTyposquattingChecklistRetrievalTime(uploadedPackageId, checklistRetrievalStopwatch.Elapsed);
             _telemetryService.TrackMetricForTyposquattingAlgorithmProcessingTime(uploadedPackageId, algorithmProcessingStopwatch.Elapsed);
+            var totalTime = algorithmProcessingStopwatch.Elapsed;
+
+            if (checklistRetrievalStopwatch != null)
+            {
+                _telemetryService.TrackMetricForTyposquattingChecklistRetrievalTime(uploadedPackageId, checklistRetrievalStopwatch.Elapsed);
+                totalTime = totalTime.Add(checklistRetrievalStopwatch.Elapsed);
+            }
 
             if (collisionIds.Count == 0)
             {
@@ -99,7 +117,7 @@ namespace NuGetGallery
             }
 
             var ownersCheckStopwatch = Stopwatch.StartNew();
-            var collisionPackagesIdAndOwners = packageRegistrations
+            var collisionPackagesIdAndOwners = _packageService.GetAllPackageRegistrations()
                 .Where(pr => collisionIds.Contains(pr.Id))
                 .Select(pr => new { Id = pr.Id, Owners = pr.Owners.Select(x => x.Key).ToList() })
                 .ToList();
@@ -122,11 +140,11 @@ namespace NuGetGallery
                 .Any(pio => pio.Owners.Any(k => k == uploadedPackageOwner.Key));
 
             wasUploadBlocked = _contentObjectService.TyposquattingConfiguration.IsBlockUsersEnabled && !isUserAllowedTyposquatting;
-
             ownersCheckStopwatch.Stop();
 
-            totalTime = totalTime.Add(ownersCheckStopwatch.Elapsed);
             _telemetryService.TrackMetricForTyposquattingOwnersCheckTime(uploadedPackageId, ownersCheckStopwatch.Elapsed);
+            totalTime = totalTime.Add(ownersCheckStopwatch.Elapsed);
+
             _telemetryService.TrackMetricForTyposquattingCheckResultAndTotalTime(
                     uploadedPackageId,
                     totalTime,
@@ -135,6 +153,11 @@ namespace NuGetGallery
                     TyposquattingCheckListLength);
 
             return wasUploadBlocked;
+        }
+
+        private bool IsCheckListCacheExpired()
+        {
+            return DateTime.UtcNow >= _checkListCacheLastRefreshTime.Add(_checkListCacheDefaultExpireTime);
         }
 
         private static int GetThreshold(string packageId)
--- a/tests/NuGetGallery.Facts/Services/TyposquattingServiceFacts.cs
+++ b/tests/NuGetGallery.Facts/Services/TyposquattingServiceFacts.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using Moq;
 using NuGet.Services.Entities;
 using Xunit;
@@ -12,11 +13,6 @@ namespace NuGetGallery
 {
     public class TyposquattingServiceFacts
     {
-        private static Mock<IPackageService> _packageService = new Mock<IPackageService>();
-        private static Mock<IContentObjectService> _contentObjectService = new Mock<IContentObjectService>();
-        private static Mock<IReservedNamespaceService> _reservedNamespaceService = new Mock<IReservedNamespaceService>();
-        private static Mock<ITelemetryService> _telemetryService = new Mock<ITelemetryService>();
-
         private static List<string> _packageIds = new List<string>
         {
             "microsoft_netframework_v1",
@@ -28,7 +24,7 @@ namespace NuGetGallery
             "System.Spatial"
         };
 
-        private IQueryable<PackageRegistration> _pacakgeRegistrationsList = Enumerable.Range(0, _packageIds.Count()).Select(i =>
+        private static IQueryable<PackageRegistration> PacakgeRegistrationsList = Enumerable.Range(0, _packageIds.Count()).Select(i =>
                 new PackageRegistration()
                 {
                     Id = _packageIds[i],
@@ -39,27 +35,48 @@ namespace NuGetGallery
 
         private User _uploadedPackageOwner = new User() { Username = string.Format("owner{0}", _packageIds.Count() + 1), Key = _packageIds.Count() + 1 };
 
-        public TyposquattingServiceFacts()
+        private static ITyposquattingService CreateService(
+            Mock<IPackageService> packageService = null,
+            Mock<IContentObjectService> contentObjectService = null,
+            Mock<IReservedNamespaceService> reservedNamespaceService = null,
+            Mock<ITelemetryService> telemetryService = null)
         {
-            _packageService
+            if (packageService == null)
+            {
+                packageService = new Mock<IPackageService>();
+                packageService
                 .Setup(x => x.GetAllPackageRegistrations())
-                .Returns(_pacakgeRegistrationsList);
+                .Returns(PacakgeRegistrationsList);
+            }
 
-            _contentObjectService
-                .Setup(x => x.TyposquattingConfiguration.PackageIdChecklistLength)
-                .Returns(20000);
-
-            _contentObjectService
+            if (contentObjectService == null)
+            {
+                contentObjectService = new Mock<IContentObjectService>();
+                contentObjectService
+                 .Setup(x => x.TyposquattingConfiguration.PackageIdChecklistLength)
+                 .Returns(20000);
+                contentObjectService
                 .Setup(x => x.TyposquattingConfiguration.IsCheckEnabled)
                 .Returns(true);
+                contentObjectService
+                    .Setup(x => x.TyposquattingConfiguration.IsBlockUsersEnabled)
+                    .Returns(true);
+            }
 
-            _contentObjectService
-                .Setup(x => x.TyposquattingConfiguration.IsBlockUsersEnabled)
-                .Returns(true);
-
-            _reservedNamespaceService
+            if (reservedNamespaceService == null)
+            {
+                reservedNamespaceService = new Mock<IReservedNamespaceService>();
+                reservedNamespaceService
                 .Setup(x => x.GetReservedNamespacesForId(It.IsAny<string>()))
                 .Returns(new List<ReservedNamespace>());
+            }
+
+            if (telemetryService == null)
+            {
+                telemetryService = new Mock<ITelemetryService>();
+            }
+
+            return new TyposquattingService(contentObjectService.Object, packageService.Object, reservedNamespaceService.Object, telemetryService.Object); ;
         }
 
         [Fact]
@@ -67,8 +84,9 @@ namespace NuGetGallery
         {
             // Arrange            
             var uploadedPackageId = "new_package_for_testing";
-            var newService = new TyposquattingService(_contentObjectService.Object, _packageService.Object, _reservedNamespaceService.Object, _telemetryService.Object);
 
+            var newService = CreateService();
+            
             // Act
             var typosquattingCheckResult = newService.IsUploadedPackageIdTyposquatting(uploadedPackageId, _uploadedPackageOwner, out List<string> typosquattingCheckCollisionIds);
 
@@ -85,7 +103,7 @@ namespace NuGetGallery
             _uploadedPackageOwner.Key = 1;
             var uploadedPackageId = "microsoft_netframework.v1";
 
-            var newService = new TyposquattingService(_contentObjectService.Object, _packageService.Object, _reservedNamespaceService.Object, _telemetryService.Object);
+            var newService = CreateService();
 
             // Act
             var typosquattingCheckResult = newService.IsUploadedPackageIdTyposquatting(uploadedPackageId, _uploadedPackageOwner, out List<string> typosquattingCheckCollisionIds);
@@ -100,8 +118,8 @@ namespace NuGetGallery
         {
             // Arrange            
             var uploadedPackageId = "Mícrosoft.NetFramew0rk.v1";
-            var newService = new TyposquattingService(_contentObjectService.Object, _packageService.Object, _reservedNamespaceService.Object, _telemetryService.Object);
-
+            var newService = CreateService();
+            
             // Act
             var typosquattingCheckResult = newService.IsUploadedPackageIdTyposquatting(uploadedPackageId, _uploadedPackageOwner, out List<string> typosquattingCheckCollisionIds);
 
@@ -111,13 +129,12 @@ namespace NuGetGallery
             Assert.Equal("microsoft_netframework_v1", typosquattingCheckCollisionIds[0]);
         }
 
-
         [Fact]
         public void CheckIsTyposquattingMultiCollisionsWithoutSameUser()
         {
             // Arrange
             var uploadedPackageId = "microsoft_netframework.v1";
-            _pacakgeRegistrationsList = _pacakgeRegistrationsList.Concat(new PackageRegistration[]
+            var pacakgeRegistrationsList = PacakgeRegistrationsList.Concat(new PackageRegistration[]
             {
                 new PackageRegistration {
                     Id = "microsoft-netframework-v1",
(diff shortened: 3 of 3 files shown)


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -331,7 +331,7 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<TyposquattingService>()
                 .AsSelf()
                 .As<ITyposquattingService>()
-                .InstancePerLifetimeScope();
+                .SingleInstance();
```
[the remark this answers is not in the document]
zhhyu: Change back to lifetime scope and add another singleton instance for caching.

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -19,8 +19,12 @@ public class TyposquattingService : ITyposquattingService
             new ThresholdInfo (lowerBound: 30, upperBound: 50, threshold: 1),
             new ThresholdInfo (lowerBound: 50, upperBound: 129, threshold: 2)
         };
+        private static readonly object Locker = new object();
```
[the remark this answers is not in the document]
zhhyu: If it's not static, multiple requests will have different locks because it's life time scope rather than singleton. Is that right?

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -19,72 +19,94 @@ public class TyposquattingService : ITyposquattingService
             new ThresholdInfo (lowerBound: 30, upperBound: 50, threshold: 1),
             new ThresholdInfo (lowerBound: 50, upperBound: 129, threshold: 2)
         };
-
+        private static readonly object Locker = new object();
```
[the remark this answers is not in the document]
zhhyu: Move to cache class now.

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -19,72 +19,93 @@ public class TyposquattingService : ITyposquattingService
...
+                                    TyposquattingCheckListCache typosquattingCheckListCache)
         {
             _contentObjectService = contentObjectService ?? throw new ArgumentNullException(nameof(contentObjectService));
             _packageService = packageService ?? throw new ArgumentNullException(nameof(packageService));
             _reservedNamespaceService = reservedNamespaceService ?? throw new ArgumentNullException(nameof(reservedNamespaceService));
             _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+            _typosquattingCheckListCache = typosquattingCheckListCache ?? throw new ArgumentNullException(nameof(typosquattingCheckListCache));
 
             TyposquattingCheckListLength = _contentObjectService.TyposquattingConfiguration.PackageIdChecklistLength;
+            _typosquattingCheckListCache.DefaultExpireTime = TimeSpan.FromDays(1);
```
joelverhagen: These should be initialized on the singleton, not per `TyposquattingService`. With this code, every time we create a new `TyposquattingService`, the last refresh time will be moved forward. Not sure, but I think this will make the cache never expire.

joelverhagen: I think we should have a unit test that initializes two `TyposquattingService` and verifies that one initializes the cache and the next uses the cached value without hitting the DB again.

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -19,72 +19,93 @@ public class TyposquattingService : ITyposquattingService
...
+            }
+            var packageIdsCheckList = _typosquattingCheckListCache.Cache;
 
             var algorithmProcessingStopwatch = Stopwatch.StartNew();
             var threshold = GetThreshold(uploadedPackageId);
             var normalizedUploadedPackageId = TyposquattingStringNormalization.NormalizeString(uploadedPackageId);
-
             var collisionIds = new ConcurrentBag<string>();
-            Parallel.ForEach(packagesCheckList, (packageId, loopState) =>
+            Parallel.ForEach(packageIdsCheckList, (packageId, loopState) =>
```
agr: What happens if `packageIdsCheckList` content changes (cache update occurs in concurrent thread)?

src/NuGetGallery/Services/TyposquattingCheckListCache.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.Generic;
+
+namespace NuGetGallery
+{
+    public class TyposquattingCheckListCache
+    {
+        public object Locker { get; } = new object();
+        public List<string> Cache { get; set; }
+        public DateTime LastRefreshTime { get; set; }
+        public TimeSpan DefaultExpireTime { get; set; }
```
agr: I'd prefer to see the cache object a bit more self-contained and protected from accidental changes. Right now it exposes all it's guts to anyone who has a reference to it allowing to change the cached content in an uncontrollable fashion, or change its expiration status, or even worse acquire update lock and keep it forever, preventing actual updates.

I'd suggest moving the refresh code into the cache object, make `Locker` private and make `Cache` `IReadOnlyCollection<string>` for an outside observer, so it cannot be modified from outside of the class.
Similarly, `LastRefreshTime` and `DefaultExpireTime` should not have public setters. Probably they should not be public at all, since for the user of that class it only matters if it is expired or not.

Additionally, I would extract interface and put (and consume) this object as an interface, not as a concrete implementation. This will reduce coupling and make testing easier.

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -19,72 +19,93 @@ public class TyposquattingService : ITyposquattingService
...
+                {
+                    if (_typosquattingCheckListCache.Cache == null || IsCheckListCacheExpired())
+                    {
+                        checklistRetrievalStopwatch = Stopwatch.StartNew();
+                        _typosquattingCheckListCache.Cache = _packageService.GetAllPackageRegistrations()
+                            .OrderByDescending(pr => pr.IsVerified)
+                            .ThenByDescending(pr => pr.DownloadCount)
+                            .Select(pr => pr.Id)
+                            .Take(TyposquattingCheckListLength)
+                            .ToList();
```
agr: `ToListAsync()`?

src/NuGetGallery/Services/TyposquattingCheckListCache.cs
agr: You might consider having the following interface:

```
public interface ISomethingSomethingCache
{
   string IReadOnlyCollection<string> Cache;
}
```
(or getter method if you want it `async`)
and the implementation would internally check for the expiration and refresh if needed. This would completely seal the implementation from the outside observer.

src/NuGetGallery/Services/TyposquattingService.cs
zhhyu: We keep the typo-squatting check as a sync method.

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -19,72 +19,93 @@ public class TyposquattingService : ITyposquattingService
...
             algorithmProcessingStopwatch.Stop();
 
-            var totalTime = checklistRetrievalStopwatch.Elapsed.Add(algorithmProcessingStopwatch.Elapsed);
-            _telemetryService.TrackMetricForTyposquattingChecklistRetrievalTime(uploadedPackageId, checklistRetrievalStopwatch.Elapsed);
             _telemetryService.TrackMetricForTyposquattingAlgorithmProcessingTime(uploadedPackageId, algorithmProcessingStopwatch.Elapsed);
+            var totalTime = algorithmProcessingStopwatch.Elapsed;
+
+            if (checklistRetrievalStopwatch != null)
+            {
+                _telemetryService.TrackMetricForTyposquattingChecklistRetrievalTime(uploadedPackageId, checklistRetrievalStopwatch.Elapsed);
```
agr: Why not report the metric as soon as we know its value? If anything throws between the cache refresh code and this line, we'll lose the data point.

tests/NuGetGallery.Facts/Services/TyposquattingServiceFacts.cs
```diff
@@ -39,36 +35,64 @@ public class TyposquattingServiceFacts
...
+            {
+                telemetryService = new Mock<ITelemetryService>();
+            }
+
+            if (typosquattingCheckListCache == null)
+            {
+                typosquattingCheckListCache = new TyposquattingCheckListCache();
+            }
+
+            return new TyposquattingService(contentObjectService.Object, packageService.Object, reservedNamespaceService.Object, telemetryService.Object, typosquattingCheckListCache); ;
```
agr: Extra semicolon at the end of the line.

src/NuGetGallery/Services/TyposquattingService.cs
zhhyu: Decouple to a new interface.

src/NuGetGallery/Services/TyposquattingService.cs
zhhyu: I think the lock will handle this.

src/NuGetGallery/Services/TyposquattingCheckListCache.cs
zhhyu: Decouple to a new interface.

src/NuGetGallery/Services/TyposquattingService.cs
zhhyu: Add to the telemetry immediately after getting.

tests/NuGetGallery.Facts/Services/TyposquattingServiceFacts.cs
zhhyu: Delete it now.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,61 @@
...
+        private static DateTime LastRefreshTime;
+        private static TimeSpan DefaultExpireTime;
+
+        private static int TyposquattingCheckListLength;
+
+        public TyposquattingCheckListCacheService()
+        {
+            TyposquattingCheckListLength = -1;
+            DefaultExpireTime = TimeSpan.FromDays(1);
+            LastRefreshTime = DateTime.UtcNow;
```
skofman1: this is not a good initialization value since it gives the impression the cache is up to date. Use DateTime.MinValue

src/NuGetGallery/Services/TyposquattingService.cs
agr: This `ForEach` happens outside of the lock as far as I can see.

agr: I think it's fine though, since cache update does not mutate existing list, but generate a new one. So it should be fine.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        public IReadOnlyCollection<string> GetTyposquattingCheckList(int checkListLength, IPackageService packageService)
+        {
+            if (checkListLength < 0)
+            {
+                throw new ArgumentNullException(nameof(checkListLength));
```
agr: `ArgumentOutOfRangeException`

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
```
agr: Need to check `packageService` for `null`

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    public class TyposquattingCheckListCacheService : ITyposquattingCheckListCacheService
+    {
+        private static readonly object Locker = new object();
+
+        private static List<string> Cache;
+        private static DateTime LastRefreshTime;
+        private static TimeSpan DefaultExpireTime;
+
+        private static int TyposquattingCheckListLength;
```
agr: Isn't it always the same as `Cache.Length`? Can we remove it?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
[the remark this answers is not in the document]
agr: Alternative solution would be instead of holding on a reference to an instance of `IPackageService`, request a factory producing instances of that service from DI: change `IPackageService packageService` to `Func<IPackageService> packageServiceFactory` in constructor. Then in `GetTyposquattingCheckList` method, you'd need not to pass an instance as an argument, but generate one as needed. Also it would not require `static` members in the cache class, which in my opinion is a cleaner approach.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
shishirx34: This value makes no sense in this context. You are always passing the `checklistLength` to the `GetTyposquattingCheckList` method. Why set this as a field on this service?

zhhyu: Even though we always pass the checklistLength, we want to refresh the cache when length changes like from 5000 to 10000. If we do not keep this value, how should we know whether this length is changed or not?

shishirx34: you can remove this field.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
zhhyu: Change to MinValue now.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
zhhyu: Done

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
zhhyu: Add that check now.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
zhhyu: We decide to keep the current approach now, because other solutions may lead to more complexity.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,66 @@
...
+        public TyposquattingCheckListCacheService()
+        {
+            TyposquattingCheckListLength = -1;
+            DefaultExpireTime = TimeSpan.FromDays(1);
+            LastRefreshTime = DateTime.MinValue;
+        }
+
+        public IReadOnlyCollection<string> GetTyposquattingCheckList(int checkListLength, IPackageService packageService)
+        {
+            if (checkListLength < 0)
```
agr: Should 0 be invalid, too?

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,66 @@
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(checkListLength));
```
agr: Please, provide an exception message stating the allowed range.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
zhhyu: I think we may allow to set check list length as 0.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,66 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery
+{
+    public class TyposquattingCheckListCacheService : ITyposquattingCheckListCacheService
+    {
+        private static readonly object Locker = new object();
```
joelverhagen: These should not be `static`.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,66 @@
...
+        private static List<string> Cache;
+        private static DateTime LastRefreshTime;
+        private static TimeSpan DefaultExpireTime;
+
+        private static int TyposquattingCheckListLength;
+
+        public TyposquattingCheckListCacheService()
+        {
+            TyposquattingCheckListLength = -1;
+            DefaultExpireTime = TimeSpan.FromDays(1);
```
joelverhagen: This should probably be configurable.

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -19,20 +19,25 @@ public class TyposquattingService : ITyposquattingService
...
+                                    ITelemetryService telemetryService,
+                                    ITyposquattingCheckListCacheService typosquattingCheckListCacheService)
         {
             _contentObjectService = contentObjectService ?? throw new ArgumentNullException(nameof(contentObjectService));
             _packageService = packageService ?? throw new ArgumentNullException(nameof(packageService));
             _reservedNamespaceService = reservedNamespaceService ?? throw new ArgumentNullException(nameof(reservedNamespaceService));
             _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+            _typosquattingCheckListCacheService = typosquattingCheckListCacheService ?? throw new ArgumentNullException(nameof(typosquattingCheckListCacheService));
 
             TyposquattingCheckListLength = _contentObjectService.TyposquattingConfiguration.PackageIdChecklistLength;
```
joelverhagen: nit: this does not need to be a field.

tests/NuGetGallery.Facts/Services/TyposquattingCheckListCacheServiceFacts.cs
```diff
@@ -0,0 +1,85 @@
...
+            var newService = new TyposquattingCheckListCacheService();
+
+            // Act
+            int tasksNum = 3;
+            Task[] tasks = new Task[tasksNum];
+            for (int i = 0; i < tasksNum; i++)
+            {
+                tasks[i] = Task.Factory.StartNew(() =>
+                {
+                    newService.GetTyposquattingCheckList(_packageIds.Count, mockPackageService.Object);
```
joelverhagen: per task, maybe execute this twice, so we have parallel _and_ sequential

tests/NuGetGallery.Facts/Services/TyposquattingCheckListCacheServiceFacts.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Moq;
+using NuGet.Services.Entities;
+using Xunit;
+
+namespace NuGetGallery
+{
+    public class TyposquattingCheckListCacheServiceFacts
```
joelverhagen: Make tests that check the various cache invalidation scenarios
 
- cache is too old
- cache has wrong length

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
zhhyu: Add the message now.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
zhhyu: Delete them.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
zhhyu: Add the expire time parameter in configuration.

src/NuGetGallery/Services/TyposquattingService.cs
zhhyu: Delete the field.

tests/NuGetGallery.Facts/Services/TyposquattingCheckListCacheServiceFacts.cs
zhhyu: Execute twice now with sequential test.

tests/NuGetGallery.Facts/Services/TyposquattingCheckListCacheServiceFacts.cs
zhhyu: Add these two tests for coverage.

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -20,71 +20,68 @@ public class TyposquattingService : ITyposquattingService
...
-                .OrderByDescending(pr => pr.IsVerified)
-                .ThenByDescending(pr => pr.DownloadCount)
-                .Select(pr => pr.Id)
-                .Take(TyposquattingCheckListLength)
-                .ToList();
+            var packageIdsCheckList = _typosquattingCheckListCacheService.GetTyposquattingCheckList(checkListLength, checkListExpireTimeInHours, _packageService);
             checklistRetrievalStopwatch.Stop();
 
+            _telemetryService.TrackMetricForTyposquattingChecklistRetrievalTime(uploadedPackageId, checklistRetrievalStopwatch.Elapsed);
+            var totalTime = checklistRetrievalStopwatch.Elapsed;
```
joelverhagen: >totalTime [](start = 16, length = 9)

Why not measure total time directly? It seems error prone to add two timestamps together. Imagine if we add more business logic between the time that stopwatches are running... then there would be unaccounted for time.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,64 @@
...
+            if (checkListLength < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(checkListLength), "Negative values are not supported.");
+            }
+            if (packageService == null)
+            {
+                throw new ArgumentNullException(nameof(packageService));
+            }
+
+            if (Cache == null || checkListLength != TyposquattingCheckListLength || IsCheckListCacheExpired(checkListExpireTimeInHours))
```
joelverhagen: >Cache == null || checkListLength != TyposquattingCheckListLength || IsCheckListCacheExpired(checkListExpireTimeInHours) [](start = 16, length = 119)

These three conditions could be grouped into a method called `ShouldCacheBeUpdated` then you can call it in two places instead of duplicating the three conditions.

src/NuGetGallery/Services/TyposquattingService.cs
zhhyu: Yes, that's true. Change it now.

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
zhhyu: Move to a function now.

src/NuGetGallery/Services/ITyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// /// </summary>
+    public interface ITyposquattingCheckListCacheService
+    {
+        /// <summary>
+        /// The function is used to get the checklist from the cache for typosquatting
+        /// </summary>
+        /// <param name="checkListLength">The length of the checklist for typosquatting</param>
+        /// <param name="checkListExpireTimeInHours">The expire time for checklist caching</param>
+        /// <param name="packageService">The package service for checklist retrieval from database</param>
+        IReadOnlyCollection<string> GetTyposquattingCheckList(int checkListLength, double checkListExpireTimeInHours, IPackageService packageService);
```
shishirx34: the expiry time shouldn't be in unit of measurement, it should be `TimeSpan`

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        private int TyposquattingCheckListConfiguredLength;
+
+        public TyposquattingCheckListCacheService()
+        {
+            TyposquattingCheckListConfiguredLength = -1;
+            LastRefreshTime = DateTime.MinValue;
+        }
+
+        public IReadOnlyCollection<string> GetTyposquattingCheckList(int checkListConfiguredLength, double checkListExpireTimeInHours, IPackageService packageService)
```
shishirx34: don't pass in hours here, pass `TimeSpan`, the caller should determine if it should be hours or mins

src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        private bool ShouldCacheBeUpdated(int checkListConfiguredLength, double checkListExpireTimeInHours)
+        {
+            return Cache == null || checkListConfiguredLength != TyposquattingCheckListConfiguredLength || IsCheckListCacheExpired(checkListExpireTimeInHours);
+        }
+
+        private bool IsCheckListCacheExpired(double checkListExpireTimeInHours)
+        {
+            return DateTime.UtcNow >= LastRefreshTime.Add(TimeSpan.FromHours(checkListExpireTimeInHours));
```
shishirx34: no `FromHours` here.

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -20,86 +20,84 @@ public class TyposquattingService : ITyposquattingService
...
             _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
-
-            TyposquattingCheckListLength = _contentObjectService.TyposquattingConfiguration.PackageIdChecklistLength;
+            _typosquattingCheckListCacheService = typosquattingCheckListCacheService ?? throw new ArgumentNullException(nameof(typosquattingCheckListCacheService));
         }
 
         public bool IsUploadedPackageIdTyposquatting(string uploadedPackageId, User uploadedPackageOwner, out List<string> typosquattingCheckCollisionIds)
         {
+            var checkListConfiguredLength = _contentObjectService.TyposquattingConfiguration.PackageIdChecklistLength;
+            var checkListExpireTimeInHours = _contentObjectService.TyposquattingConfiguration.PackageIdChecklistCacheExpireTimeInHours;
```
shishirx34: convert to timespan here

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -20,86 +20,84 @@ public class TyposquattingService : ITyposquattingService
...
+            var totalTimeStopwatch = Stopwatch.StartNew();
             var checklistRetrievalStopwatch = Stopwatch.StartNew();
-            var packageRegistrations = _packageService.GetAllPackageRegistrations();
-            var packagesCheckList = packageRegistrations
-                .OrderByDescending(pr => pr.IsVerified)
-                .ThenByDescending(pr => pr.DownloadCount)
-                .Select(pr => pr.Id)
-                .Take(TyposquattingCheckListLength)
-                .ToList();
+            var packageIdsCheckList = _typosquattingCheckListCacheService.GetTyposquattingCheckList(checkListConfiguredLength, checkListExpireTimeInHours, _packageService);
```
shishirx34: pass the above timespan

src/NuGetGallery/Services/TyposquattingService.cs
```diff
@@ -20,86 +20,84 @@ public class TyposquattingService : ITyposquattingService
...
+                totalTimeStopwatch.Stop();
                 _telemetryService.TrackMetricForTyposquattingCheckResultAndTotalTime(
                     uploadedPackageId,
-                    totalTime,
+                    totalTimeStopwatch.Elapsed,
                     wasUploadBlocked,
                     typosquattingCheckCollisionIds,
-                    TyposquattingCheckListLength);
+                    packageIdsCheckList.Count,
+                    checkListExpireTimeInHours);
```
shishirx34: what do you get from passing the configured hours here?

zhhyu: Double check the cache expiration time we set in the configuration file.

tests/NuGetGallery.Facts/Services/TyposquattingCheckListCacheServiceFacts.cs
```diff
@@ -0,0 +1,130 @@
...
+                new PackageRegistration()
+                {
+                    Id = _packageIds[i],
+                    DownloadCount = new Random().Next(0, 10000),
+                    IsVerified = true,
+                    Owners = new List<User> { new User() { Username = string.Format("owner{0}", i + 1), Key = i + 1 } }
+                }).AsQueryable();
+
+        [Fact]
+        public void CheckTyposquattingCheckListCache()
```
skofman1: When reading UT code, you should be able to understand what the test checks. Consider using the following name: GetTyposquattingCheckListIsThreadSafe or
WhenGettingCheckListFromMultipleThreadsItIsInitializedOnce

tests/NuGetGallery.Facts/Services/TyposquattingCheckListCacheServiceFacts.cs
```diff
@@ -0,0 +1,130 @@
...
+            // Assert
+            mockPackageService.Verify(
+               x => x.GetAllPackageRegistrations(),
+               Times.Exactly(2));
+        }
+
+        [Fact]
+        public void CheckRefreshedCheckListLengthNotAllowed()
+        {
+            // Arrange
```
skofman1: add tests for input verification (null parameters, etc)

zhhyu: Add some tests for cache.

tests/NuGetGallery.Facts/Services/TyposquattingCheckListCacheServiceFacts.cs
zhhyu: Change the names for cache tests.

PR #6600, opened 2018-10-26
Fix concurrency/consistency issues in Gallery API push

shishirx34:
Fixes https://github.com/NuGet/Engineering/issues/1817. Mostly affects the functional test failures, kind of hard to test but I will deploy this change to dev and run functional tests couple of times to verify.

### DIFF
--- a/src/NuGetGallery.Core/NuGetGallery.Core.csproj
+++ b/src/NuGetGallery.Core/NuGetGallery.Core.csproj
@@ -161,6 +161,7 @@
     <Compile Include="Services\CoreSymbolPackageService.cs" />
     <Compile Include="Services\CorePackageService.cs" />
     <Compile Include="Services\CryptographyService.cs" />
+    <Compile Include="Services\PackageAlreadyExistsException.cs" />
     <Compile Include="Services\FileAlreadyExistsException.cs" />
     <Compile Include="Services\FileUriPermissions.cs" />
     <Compile Include="Services\IAccessCondition.cs" />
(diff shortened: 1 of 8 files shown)


joelverhagen:
>             var package = packageRegistration.Packages.SingleOrDefault(pv => pv.Version == packageMetadata.Version.OriginalVersion);

I don't understand why we even need this line. Isn't this checked at the beginning of the control and again (by the DB) on commit?

Said another way, we already check it in memory here:
[https://github.com/NuGet/NuGetGallery/blob/e582941f6b4cfce14ef1d032f95c1d107dd0e130/src/NuGetGallery/Controllers/ApiController.cs#L599-L602](https://github.com/NuGet/NuGetGallery/blob/e582941f6b4cfce14ef1d032f95c1d107dd0e130/src/NuGetGallery/Controllers/ApiController.cs#L599-L602)

And the DB check it with a unique constraint on (PackageRegistrationKey, NormalizedVersion) on the packages table. What does this one add?

---
Refers to: src/NuGetGallery/Services/PackageService.cs:478 in bee9def. [](commit_id = bee9defcb75cff3150f735263b987c59d769b6a5, deletion_comment = False)

shishirx34:
@joelverhagen - technically we do not check for the conflict on DB commit, rather in the `CommitPackageAsync` we check if the file exists in validation container in which case we return a `Conflict`. But even before that we queue up the validation into the pipeline(which eventually will deadletter or thrown away). If `StartValidationAsync` were to mark package as available then on DB commit unique constraint will still throw an `EntityException` which we could possibly handle(not handled in the code today) to return conflict(that could be another approach).

[https://github.com/NuGet/NuGetGallery/blob/e582941f6b4cfce14ef1d032f95c1d107dd0e130/src/NuGetGallery/Services/PackageUploadService.cs#L352](https://github.com/NuGet/NuGetGallery/blob/e582941f6b4cfce14ef1d032f95c1d107dd0e130/src/NuGetGallery/Services/PackageUploadService.cs#L352)

I do not see any harm in handling it the current way in PR, given that it could avoid unnecessary queuing of validations and the check doesn't result in any performance issues(fail fast).

shishirx34:
I added the concurrency and sql constraint violations check for completion.

PR #6602, opened 2018-10-26
Hotfix: Do not CC sender for Contact Owners emails

scottbommarito:
https://github.com/NuGet/Engineering/issues/1825


xavierdecoster:
@scottbommarito ⌚️ The changes to the `NuGetGallery.Core` types in this PR also need to be ported to `ServerCommon`, where the latest version of these types now live.

Issue #6601, opened 2018-10-26
Certain strange requests result in HTTP 500


joelverhagen:
This can trigger "failed request" alert that we have set up in Application Insights.

Examples:
```
https://www.nuget.org/api/v2/lazy/>=1.0.5 <1.1
https://www.nuget.org/api/v2/mysql/*
https://www.nuget.org/api/v2/mysql-pool/*
https://www.nuget.org/api/v2/comb/*
https://www.nuget.org/api/v2/express/*
https://www.nuget.org/api/v2/hive-cache/*
https://www.nuget.org/api/v2/mysql/*
https://www.nuget.org/api/v2/comb/>=0.0.2
https://www.nuget.org/api/v2/hive-cache/*
https://www.nuget.org/api/v2/mysql/*
```

All of these result in HTTP 500 by redirecting to the 500 error page, e.g.
```
https://www.nuget.org/Errors/500?aspxerrorpath=/api/v2/lazy/%3E=1.0.5%20%3C1.1
```

These should probably be 404s.

joelverhagen:
The fix caused https://github.com/NuGet/NuGetGallery/issues/6655. Re-opening.

Next time this is fixed, we should add a functional test that verifies reason phrases come through for API push.

agr:
Some additional investigation as we were getting similar 500s (not from OData).
Example of the request that causes 500:
https://www.nuget.org/.nuGetV3/feed.json:properties

when `customErrors` are off we get a proper 400 for it:

> A potentially dangerous Request.Path value was detected from the client (:).

We have a bit overzealous `customErrors` settings:

https://github.com/NuGet/NuGetGallery/blob/7a0ed71f4c1f608b01d0947132123197b9bea839/src/NuGetGallery/Web.config#L323-L328

Which grab all errors and transform them into 500s.

agr:
Some details on what exactly happens in different cases:

(ran against localhost with `customErrors mode="On"`)

| Method returns/throws | Class | Result observed from outside |
|-|-|-|
| `return new HttpStatusCodeWithBodyResult(400);` | `ApiController` | 400 |
| `throw new HttpException(400)` | `ApiController` | 500 |
| `throw new HttpResponseException(400)` | `ApiController` | 500 |
| `return new HttpStatusCodeWithBodyResult(500);` | `ApiController` | 500 |
| `throw new HttpException(500)` | `ApiController` | 500 |
[...]

PR #6604, opened 2018-10-29
Redirect HTTP 400 to a specific error page instead of the default for HTTP 500

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6601.

Currently, HTTP 400s produced by IIS (unsafe character validation for URL path) get redirected to the HTTP 500 error page.

### DIFF
--- a/src/NuGetGallery/Web.config
+++ b/src/NuGetGallery/Web.config
@@ -322,6 +322,7 @@
     <customErrors mode="RemoteOnly" defaultRedirect="~/Errors/500" redirectMode="ResponseRedirect">
       <!-- Adding ? at the end of the redirect URL prevents the illegal request to be passed
            as a query parameter to the redirect URL and causing additional failures -->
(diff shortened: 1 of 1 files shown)


joelverhagen:
I was able to verify this locally by setting `<customErrors>` `mode="On"` and `<httpErrors>` `errorMode="Custom"`. Will verify on DEV environment before merging.

skofman1:
I think you need to modify this section as well: 
https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Web.config#L363

but I'm not sure.

joelverhagen:
> I think you need to modify this section as well: 
> https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Web.config#L363
> 
> but I'm not sure.

Yeah, good point. This is not customer impacting but it will help local development.

PR #6608, opened 2018-10-30
Validation message enhancements

agr:
Progress on #6545.

I needed a way to enhance the errors and warning produced on a package upload and verify pages to allow passing raw HTML as an error, so I could include links in an error or warning message.

The gist of the change is that validation code now instead of returning errors and warnings as strings returns objects (implementing the `IValidationMessage` interface). Those object provide a way to render the message in either plain text or optional HTML. If object chooses not to provide HTML, its plain text representation is used instead. For API purposes plain text representation is always used. Errors passed through `TempData["Message"]` are also always plain text (that's a bit more global change that I don't want to touch right now).

Errors and warnings were passed as an array of strings in a JSON object to be rendered on client side (either through API, or baked into the page (search for `InProgressPackage`)). This is replaced with passing an object instead that should either have `PlainTextMessage` or `RawHtmlMessage` properties set and will render one of those preferring plain text if both provided. On C# side it is covered by the `JsonValidationMessage` object.

Given the inherent insecurity of pasting raw HTML into the output, some precautions were taken to not allow mindless HTML emitting: the default generic implementation of `IValidationMessage` (`PlainTextOnlyValidationMessage` a simple wrapper to allow fast transition to passing objects instead of strings) only accepts plain text as an input and comments in the `IValidationMessage` strongly discourage the implementation that would allow any HTML in. The `JsonValidationMessage` constructors either accept a string, treating it as a plain text (and JS code will encode it) or accept a `IValidationMessage` from which raw HTML will be taken if provided.

The only non-generic implementation of `IValidationMessage` that is currently provided is `PackageShouldNotBeSignedUserFixableValidationMessage` which does not actually produce any raw HTML, but is used to provide different message when error is shown on a web page than if the error is passed in API response. previously that difference was hardcoded in `PackagesController` and required its own value in `PackageValidationResultType` which was eliminated.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -253,7 +253,7 @@ namespace NuGetGallery
             var validationErrorMessage = GetErrorMessageOrNull(validationResult);
             if (validationErrorMessage != null)
             {
-                TempData["Message"] = validationErrorMessage;
+                TempData["Message"] = validationErrorMessage.PlainTextMessage;
                 return View(model);
             }
 
@@ -280,7 +280,7 @@ namespace NuGetGallery
             }
 
             var verifyRequest = new VerifyPackageRequest(packageMetadata, accountsAllowedOnBehalfOf, existingPackageRegistration);
-            verifyRequest.Warnings.AddRange(validationResult.Warnings);
+            verifyRequest.Warnings.AddRange(validationResult.Warnings.Select(w => new JsonValidationMessage(w)));
             verifyRequest.IsSymbolsPackage = false;
 
             model.InProgressUpload = verifyRequest;
@@ -301,18 +301,18 @@ namespace NuGetGallery
             {
                 if (existingUploadFile != null)
                 {
-                    return Json(HttpStatusCode.Conflict, new[] { Strings.UploadPackage_UploadInProgress });
+                    return Json(HttpStatusCode.Conflict, new[] { new JsonValidationMessage(Strings.UploadPackage_UploadInProgress) });
                 }
             }
 
             if (uploadFile == null)
             {
-                return Json(HttpStatusCode.BadRequest, new[] { Strings.UploadFileIsRequired });
+                return Json(HttpStatusCode.BadRequest, new[] { new JsonValidationMessage(Strings.UploadFileIsRequired) });
             }
 
             if (!AllowedPackageExtentions.Contains(Path.GetExtension(uploadFile.FileName)))
             {
-                return Json(HttpStatusCode.BadRequest, new[] { Strings.UploadFileMustBeNuGetPackage });
+                return Json(HttpStatusCode.BadRequest, new[] { new JsonValidationMessage(Strings.UploadFileMustBeNuGetPackage) });
             }
 
             using (var uploadStream = uploadFile.InputStream)
@@ -325,10 +325,10 @@ namespace NuGetGallery
                     var errors = ManifestValidator.Validate(packageArchiveReader.GetNuspec(), out nuspec, out packageMetadata).ToArray();
                     if (errors.Length > 0)
                     {
-                        var errorStrings = new List<string>();
+                        var errorStrings = new List<JsonValidationMessage>();
                         foreach (var error in errors)
                         {
-                            errorStrings.Add(error.ErrorMessage);
+                            errorStrings.Add(new JsonValidationMessage(error.ErrorMessage));
                         }
 
                         return Json(HttpStatusCode.BadRequest, errorStrings.ToArray());
@@ -377,12 +377,14 @@ namespace NuGetGallery
             if (ActionsRequiringPermissions.UploadSymbolPackage.CheckPermissionsOnBehalfOfAnyAccount(
                 currentUser, existingPackageRegistration, out accountsAllowedOnBehalfOf) != PermissionsCheckResult.Allowed)
             {
-                return Json(HttpStatusCode.Conflict, new[] { string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, existingPackageRegistration.Id) });
+                return Json(HttpStatusCode.Conflict, new[] {
+                    new JsonValidationMessage(string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, existingPackageRegistration.Id)) });
             }
 
             if (existingPackageRegistration.IsLocked)
             {
-                return Json(HttpStatusCode.Forbidden, new[] { string.Format(CultureInfo.CurrentCulture, Strings.PackageIsLocked, existingPackageRegistration.Id) });
+                return Json(HttpStatusCode.Forbidden, new[] {
+                    new JsonValidationMessage(string.Format(CultureInfo.CurrentCulture, Strings.PackageIsLocked, existingPackageRegistration.Id)) });
             }
 
             // Save the uploaded file
@@ -409,7 +411,7 @@ namespace NuGetGallery
             if (foundEntryInFuture)
             {
                 return Json(HttpStatusCode.BadRequest, new[] {
-                    string.Format(CultureInfo.CurrentCulture, Strings.PackageEntryFromTheFuture, entryInTheFuture.Name) });
+                    new JsonValidationMessage(string.Format(CultureInfo.CurrentCulture, Strings.PackageEntryFromTheFuture, entryInTheFuture.Name)) });
             }
 
             try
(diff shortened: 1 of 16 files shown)


src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml
```diff
@@ -46,14 +59,24 @@
...
-                                We recommend that you fix these issues and upload a new package.
-                                <a href="https://docs.microsoft.com/en-us/nuget/policies/nuget-faq#managing-packages-on-nugetorg" alt="Read more">Read more</a><br />
-                            </span>)
+                @<span>
+                    We found the following issue(s):
+                    <ul data-bind="foreach: $data.Warnings">
+                        <!-- ko if: typeof($data) === "string" -->
+                        <li><span data-bind="text: $data"></span></li>
+                        <!-- /ko -->
+                        <!-- ko ifnot: typeof($data) === "string" -->
```
agr: Need to make it consistent with the code change above.

src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml
```diff
@@ -1,6 +1,19 @@
 ﻿<script type="text/html" id="validation-errors">
     <div data-bind="foreach:$data" class="validation-error-message-list">
-        @ViewHelpers.AlertDanger(@<text><span data-bind="text:$data"></span></text>)
+        @ViewHelpers.AlertDanger(
+            @<text>
+                <!-- ko if: typeof($data) === "string" -->
```
agr: I am a bit conflicted on this check. It would allows us to provide good user experience if I messed up somewhere. On the other hand if I actually did mess up, it will hide it.
I though about making the `string` - `object` to be the distinguishing factor for determining whether we produce text or raw HTML, but that would complicate things on C# side when we return an array of mixed responses.

src/NuGetGallery/Services/IValidationMessage.cs
```diff
@@ -0,0 +1,32 @@
...
+        string PlainTextMessage { get; }
+
+        /// <summary>
+        /// An indicator that raw HTML representation is available.
+        /// </summary>
+        bool HasRawHtmlRepresentation { get; }
+
+        /// <summary>
+        /// HANDLE WITH EXTREME CARE. Raw HTML representation of the message.
+        /// Under no conditions it may contain unvalidated user data.
```
loic-sharma: >it may [](start = 32, length = 6)

`it may` -> `may it`

src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml
```diff
@@ -1,6 +1,19 @@
...
+                <span data-bind="text:$data"></span>
```
loic-sharma: ><span data-bind="text:$data"></span> [](start = 16, length = 36)

This line is missing an indentation

joelverhagen: nit indent

src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml
```diff
@@ -1,6 +1,19 @@
...
+                <!-- /ko -->
+                <!-- ko if: typeof($data) === "object" -->
+                    <!-- ko if: $data.PlainTextMessage -->
+                    <span data-bind="text:$data.PlainTextMessage"></span>
+                    <!-- /ko -->
+                    <!-- ko ifnot: $data.PlainTextMessage -->
+                    <span data-bind="html:$data.RawHtmlMessage"></span>
```
joelverhagen: nit: space between `html:`/`text:` and the value

src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml
```diff
@@ -46,14 +59,24 @@
...
-                                    <li><span data-bind="text: $data"></span></li>
-                                </ul>
-                                We recommend that you fix these issues and upload a new package.
-                                <a href="https://docs.microsoft.com/en-us/nuget/policies/nuget-faq#managing-packages-on-nugetorg" alt="Read more">Read more</a><br />
-                            </span>)
+                @<span>
+                    We found the following issue(s):
+                    <ul data-bind="foreach: $data.Warnings">
+                        <!-- ko if: typeof($data) === "string" -->
+                        <li><span data-bind="text: $data"></span></li>
```
joelverhagen: nit: indent

src/NuGetGallery/Services/PackageShouldNotBeSignedUserFixableValidationMessage.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// presented on a web page, so this class abuses the messaging a little, no actual HTML is generated.
+    /// </summary>
+    public class PackageShouldNotBeSignedUserFixableValidationMessage : IValidationMessage
+    {
+        public string PlainTextMessage => Strings.UploadPackage_PackageIsSignedButMissingCertificate_CurrentUserCanManageCertificates;
+
+        public bool HasRawHtmlRepresentation => true;
+
+        public string RawHtmlMessage 
+            => HttpUtility.HtmlEncode(
```
joelverhagen: >HtmlEncode [](start = 27, length = 10)

Neither of these include HTML today.

You can linkify Account Settings if you want this, otherwise just make it plain text.

src/NuGetGallery/Services/JsonValidationMessage.cs
```diff
@@ -0,0 +1,34 @@
...
+    public class JsonValidationMessage
+    {
+        public JsonValidationMessage(string message)
+        {
+            PlainTextMessage = message;
+            RawHtmlMessage = null;
+        }
+
+        public JsonValidationMessage(IValidationMessage message)
+        {
```
joelverhagen: check nulls

src/NuGetGallery/Services/JsonValidationMessage.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Validation message representation for returning in Json responses for Gallery web pages.
+    /// </summary>
+    public class JsonValidationMessage
+    {
+        public JsonValidationMessage(string message)
+        {
```
joelverhagen: check nulls

src/NuGetGallery/Infrastructure/HttpStatusCodeWithServerWarningResult.cs
```diff
@@ -18,6 +18,12 @@ public HttpStatusCodeWithServerWarningResult(HttpStatusCode statusCode, IReadOnl
             Warnings = warnings ?? new string[0];
         }
 
+        public HttpStatusCodeWithServerWarningResult(HttpStatusCode statusCode, IReadOnlyList<IValidationMessage> warnings)
+            : this(statusCode, warnings.Select(w => w.PlainTextMessage).ToList())
+        {
+
+        }
```
joelverhagen: extra space

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -377,12 +377,14 @@ private async Task<JsonResult> UploadSymbolsPackageInternal(PackageArchiveReader
...
             {
-                return Json(HttpStatusCode.Conflict, new[] { string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, existingPackageRegistration.Id) });
+                return Json(HttpStatusCode.Conflict, new[] {
+                    new JsonValidationMessage(string.Format(CultureInfo.CurrentCulture, Strings.PackageIdNotAvailable, existingPackageRegistration.Id)) });
             }
 
             if (existingPackageRegistration.IsLocked)
             {
-                return Json(HttpStatusCode.Forbidden, new[] { string.Format(CultureInfo.CurrentCulture, Strings.PackageIsLocked, existingPackageRegistration.Id) });
+                return Json(HttpStatusCode.Forbidden, new[] {
```
joelverhagen: >Json [](start = 23, length = 4)

Thoughts on an overload here that accepts a string array still and maps to `JsonValidationMessage`?

src/NuGetGallery/Services/PackageShouldNotBeSignedUserFixableValidationMessage.cs
agr: Previous implementation contained a hack in `PackagesController` that made sure different message is displayed on Web UI vs the CLI. I removed that hack and created this class instead. It exists not to provide raw HTML of that message, but to contain that hack and the code here actively resists raw HTML in the strings. I wanted to emphasize that the strings are not expected to contain HTML code.

src/NuGetGallery/Controllers/PackagesController.cs
agr: I thought about that, but that would be confusing. It would only exist in `PackagesController` and then similar code in other places would behave differently.

PR #6637, opened 2018-11-07
Don't attempt to send emails if they have no recipients

scottbommarito:
Aligns Gallery with https://github.com/NuGet/ServerCommon/pull/222

Also made some coding style improvements from https://github.com/NuGet/NuGetGallery/pull/6638 that I didn't want to make as part of a hotfix.


src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -202,11 +202,8 @@
     <EmbeddedResource Include="Infrastructure\MigrateUserToOrganization.sql" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="NuGet.Services.Entities">
-      <Version>2.31.0</Version>
-    </PackageReference>
-    <PackageReference Include="NuGet.Services.Messaging">
-      <Version>2.31.0</Version>
+    <PackageReference Include="NuGet.Services.Messaging.Email">
+      <Version>2.36.0-sb-emptyto-2190469</Version>
```
scottbommarito: Will align with master build once other PR is out.

src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -202,11 +202,8 @@
     <EmbeddedResource Include="Infrastructure\MigrateUserToOrganization.sql" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="NuGet.Services.Entities">
```
scottbommarito: I removed these dependencies because they are transitively pulled in by `NuGet.Services.Messaging.Email`. I thought having this single dependency to rev will make this easier to manage in the future.

I have no problem with reverting this back if you think this isn't a good idea.

src/NuGetGallery/Infrastructure/Mail/GalleryEmailRecipientsUtility.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Mail;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery.Infrastructure.Mail
+{
+    public static class GalleryEmailRecipientsUtility
+    {
+        public static IReadOnlyList<MailAddress> GetAddressesWithPermission(User user, ActionRequiringAccountPermissions action)
```
scottbommarito: This is `EmailRecipientsWithPermission.AddAddressesWithPermission`, but moved into a static method.

`EmailRecipientsWithPermission` was an unnecessary extra layer of abstraction so I removed it as part of this PR.

scottbommarito: After discussion with @joelverhagen, I'll be moving these changes to a separate PR.

src/NuGetGallery/Infrastructure/Mail/Messages/OrganizationMemberRemovedMessage.cs
```diff
@@ -31,13 +31,10 @@ public class OrganizationMemberRemovedMessage : MarkdownEmailBuilder
 
         public override IEmailRecipients GetRecipients()
         {
-            if (!Organization.EmailAllowed)
-            {
-                return EmailRecipients.None;
```
scottbommarito: I removed the usage of `EmailRecipients.None` because I think it's more misleading than it's worth. We're using `EmailRecipients.None` as a signal for "this email shouldn't be sent" when the real signal for when an email shouldn't be sent is whether or not it has any recipients to send to. Because "an email without recipients to send to" and "an email with no recipients at all" are not logically equivalent, this seems odd.

For example, in this function, we return `EmailRecipients.None` when the intended `To` address doesn't allow emails. Does whether or not the `To` address allows emails affect the `replyTo` field of the email? Technically no. So then why are we returning a completely empty list of recipients when there are people whom replies to the email should go to?

xavierdecoster: Makes sense, good call! 👍

PR #6638, opened 2018-11-08
Hotfix: Don't attempt to send emails if they have no recipients

scottbommarito:
Aligns Gallery with NuGet/ServerCommon#222


src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -202,11 +202,8 @@
     <EmbeddedResource Include="Infrastructure\MigrateUserToOrganization.sql" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="NuGet.Services.Entities">
```
scottbommarito: I removed these dependencies because they are transitively pulled in by NuGet.Services.Messaging.Email. I thought having this single dependency to rev will make this easier to manage in the future.

I have no problem with reverting this back if you think this isn't a good idea.

PR #6640, opened 2018-11-08
Add update blob properties support to cloud storage

shishirx34:


src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -280,10 +280,18 @@ public async Task<IFileReference> GetFileReferenceAsync(string folderName, strin
             if (!string.IsNullOrEmpty(cacheControl))
             {
                 await destBlob.FetchAttributesAsync();
+
                 if (string.IsNullOrEmpty(destBlob.Properties.CacheControl))
```
joelverhagen: We should do `destBlob.Properties.CacheControl != cacheControl`. 

This would support the case where we change the cache control from one value to another _and_ the null to non-null transition.

shishirx34:
based on feedback from @joelverhagen and @loic-sharma I will enhance the `SetMetadataAsync` method to update properties

src/NuGetGallery.Core/Services/CloudBlobCoreFileStorageService.cs
```diff
@@ -280,10 +280,18 @@ public async Task<IFileReference> GetFileReferenceAsync(string folderName, strin
             if (!string.IsNullOrEmpty(cacheControl))
```
joelverhagen: I think this logic should move to `SetMetadataAsync` and make that method allow the setting of cache control in addition to custom metadata. That allows the caller to update metadata even in the ContentMD5 no-op case.

joelverhagen: Additionally, these update cache control part operates on a set of etags completely independent of the other copy logic. This is a bit confusing.

`SetMetadataAsync` also has this property that it operates on whatever the latest version of the blob is, not on the copied version of the blob. In other words, this is another reason `SetMetadataAsync` is a better place for this.

src/NuGetGallery.Core/Services/ICoreFileStorageService.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.WindowsAzure.Storage.Blob;
```
joelverhagen: `System` first

PR #6644, opened 2018-11-09
Aligned NuGet.GalleryCore and NuGetGallery references to NuGet.Services.Entities

agr:
Forgot to update NuGet.GalleryCore package reference after updating the NuGetGallery one.


agr:
@scottbommarito, yes, your PR supercedes this.

Issue #6650, opened 2018-11-10
Display package page does a query per version -- for latest symbols package


joelverhagen:
`DisplayPackageViewModel` has a line where it calls `package.LatestSymbolPackage()`. `DisplayPackageViewModel` is initialized per package version when viewing the display package page. `package.LatestSymbolPackage()` does a SQL query.

This is the average number of SQL queries per display package page view:

![image](https://user-images.githubusercontent.com/94054/48296230-f1c81500-e448-11e8-950f-2df2ae8ad5cf.png)

```
dependencies
| where timestamp > ago(30d)
| where type == "SQL"
| where operation_Name == "GET packages/DisplayPackage" 
| summarize sum(itemCount) / dcount(operation_Id) by bin(timestamp, 6h)
| render timechart 
```

Clearly around October 24, we deployed this bug.

**We should have a unit test that the number of queries to render a page with 3 versions does not grow when there are 4 versions.**

/cc @ryuyu @loic-sharma

joelverhagen:
Request duration is also impacted:

![image](https://user-images.githubusercontent.com/94054/48296266-79ae1f00-e449-11e8-922e-6880880763a7.png)

```
requests
| where timestamp > ago(30d)
| where operation_Name == "GET packages/DisplayPackage" 
| summarize avg(duration) by bin(timestamp, 6h)
| render timechart 
```

shishirx34:
![image](https://user-images.githubusercontent.com/1646506/48584728-6a165680-e8df-11e8-9ebf-0e3c2b0c0920.png)

![image](https://user-images.githubusercontent.com/1646506/48584771-8914e880-e8df-11e8-817a-2aaafd3ffd28.png)

PR #6667, opened 2018-11-14
Fix DisplayPackage Perf regression

shishirx34:
Addresses #6650 

The fix is basically to avoid setting the `LatestSymbolPackage` on all the package versions, and only set on the main package for the display view. The perf gain would be reduced dependency calls. I verified this locally and was able to see the reduced dependency calls. I will analyze the performance over the week after merging and deploying this change in DEV.

The problem is we are reusing the same view model for two different views(which happen to be on the same page), details view and package versions history. Ideally we should have a completely different view model for the package versions list(history) which should only have the required data populated rather than using same `DisplayPackageViewModel`. However, that is overall a bigger change that I do not want to address in this PR.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1221,17 +1221,20 @@ namespace NuGetGallery
             }
 
             var model = new DeletePackageViewModel(package, currentUser, DeleteReasons);
+            var packageViewModelsForAllAvailableSymbolsPackage = package
+                .PackageRegistration
+                .Packages
+                .Where(p => p.PackageStatusKey != PackageStatus.Deleted)
+                .Select(p => p.LatestSymbolPackage())
+                .Where(sp => sp != null && sp.StatusKey == PackageStatus.Available)
+                .Select(sp => new PackageViewModel(sp.Package));
 
             model.VersionSelectList = new SelectList(
-                model
-                .PackageVersions
-                .Where(p => !p.Deleted
-                    && p.LatestSymbolsPackage != null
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1221,17 +1221,20 @@ public virtual ActionResult DeleteSymbols(string id, string version)
             }
 
             var model = new DeletePackageViewModel(package, currentUser, DeleteReasons);
+            var packageViewModelsForAllAvailableSymbolsPackage = package
+                .PackageRegistration
+                .Packages
+                .Where(p => p.PackageStatusKey != PackageStatus.Deleted)
+                .Select(p => p.LatestSymbolPackage())
```
joelverhagen: Let's track an issue to fix this unbounded number of queries. It's certainly not as big of a deal as the package details page.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1221,17 +1221,20 @@ public virtual ActionResult DeleteSymbols(string id, string version)
...
+                .Where(sp => sp != null && sp.StatusKey == PackageStatus.Available)
+                .Select(sp => new PackageViewModel(sp.Package));
```
chenriksson: The code is getting even more complex... would suggest adding comments so we don't accidentally regress. e.g., "Query the VersionSelectList directly off the EF objects and not view models which could trigger unintended. deferred queries."

src/NuGetGallery/Controllers/PackagesController.cs
shishirx34: https://github.com/NuGet/NuGetGallery/issues/6668

PR #6665, opened 2018-11-14
Move email message implementations from ServerCommon to NuGetGallery

xavierdecoster:
Follow-up to https://github.com/NuGet/Engineering/issues/1897

This moves the email message implementations into either `NuGetGallery.Core` or `NuGetGallery`.
The dependency to `NuGet.Services.Entities` is now explicit again for `NuGetGallery.Core`, as `NuGet.Services.Messaging.Email` no longer depends on it.

Note to reviewers: the newly added files are unchanged and originate from ServerCommon. Except for the newly introduced `EmailRecipientsHelper`. The static methods on `EmailRecipientsHelper` used to be in `EmailRecipients`, but required a dependency to `NuGet.Services.Entities`. Therefor, I moved it into `NuGetGallery.Core`, where it can be shared along with the shared email messages.

TODO: need to consume new to-be-created master-release of ServerCommon packages.

@skofman1 can this ServerCommon 2.37.0 release be created?

PR #6672, opened 2018-11-15
Use CachingSecretReader from ServerCommon

chenriksson:
To unblock https://github.com/NuGet/NuGetGallery/pull/6665, I propose we use ServerCommon's CachingSecretReader in Gallery. This includes the following changes:
- Adds 24-hour refresh interval
- Thread safety changes by @loic-sharma 
- Support for secret expiration, needed for auto-rotation of storage credentials (from PS gallery team)

Local testing against DEV looks good, but let me know if there are any concerns with this.

/cc @xavierdecoster


chenriksson:
@scottbommarito I don't remember the full details... but looks like I bumped `NuGet.Services.KeyVault` from `1.0.0` to a more recent version as part of the SQL AAD integration. (https://github.com/NuGet/NuGetGallery/commit/22f8bf6f779feb559513047f50c5f44334a867a8#diff-8110fe0d3615d286a6788c2f19f51e54)

Issue #6518, opened 2018-10-05
[Licenses 12/7] Gallery license endpoint implementation for embedded licenses


agr:
`/packages/<packageid>/<packageversion>/license` should display package embedded license or redirect to licenseUrl.

Initial implementation may just redirect to flat container.

PR #6675, opened 2018-11-16
Display License file from blob storage/local files

zhhyu:
With given package, we show the license content from blob storage.
Fix: https://github.com/NuGet/NuGetGallery/issues/6518


src/NuGetGallery/Services/LicenseFileService.cs
```diff
@@ -0,0 +1,37 @@
...
+using NuGetGallery.Configuration;
+
+namespace NuGetGallery
+{
+    public class LicenseFileService : ILicenseFileService
+    {
+        private readonly IAppConfiguration _config;
+        public LicenseFileService(IAppConfiguration configuration)
+        {
+            _config = configuration;
```
skofman1: null check

src/NuGetGallery.Core/CoreConstants.cs
```diff
@@ -46,6 +46,7 @@ public static class Folders
 
         public const string UploadTracingKeyHeaderName = "upload-id";
 
+        public const string PackageBaseAddress = "PackageBaseAddress/3.0.0";
```
skofman1: this const is used in Gallery and not Gallery.Core, so it shouldn't be here

src/NuGetGallery/Services/LicenseFileService.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+        public async Task<string> GetLicenseFileBlobStoragePath(string packageId, string packageVersion)
```
skofman1: nit: line between methods

src/NuGetGallery/Services/LicenseFileService.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using NuGet.Services.Search.Client;
+using NuGetGallery.Configuration;
+
+namespace NuGetGallery
+{
+    public class LicenseFileService : ILicenseFileService
```
skofman1: We will also need to support a scenario where the Gallery is running in stand alone mode. I recommend renaming the class to BlobLicenseFileService (or something like this)

src/NuGetGallery.Core/CoreConstants.cs
agr: We'll need it in catalog2registration.

src/NuGetGallery/Services/LicenseFileService.cs
```diff
@@ -0,0 +1,37 @@
...
+            if (packageId == null)
+            {
+                throw new ArgumentNullException(nameof(packageId));
+            }
+            if (packageVersion == null)
+            {
+                throw new ArgumentNullException(nameof(packageVersion));
+            }
+
+            var relativePath = String.Join("/", new string[] { packageId.ToLowerInvariant(), NuGetVersionFormatter.Normalize(packageVersion).ToLowerInvariant(), CoreConstants.LicenseFileName});
```
agr: Please, use lowercase `string` everywhere.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -663,6 +666,17 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
             return View(model);
         }
 
+        public virtual async Task<ActionResult> License(string id, string version)
+        {
+            var package = _packageService.FindPackageByIdAndVersionStrict(id, version);
+            if (package == null)
```
agr: Package might not have an embedded license file, we should 404 early in this case, too.

src/NuGetGallery/Services/LicenseFileService.cs
loic-sharma: >String.Join("/", [](start = 31, length = 16)

Instead of `string.Join("/", ...)` you can use `Directory.Combine`: [https://docs.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=netframework-4.7.2#System_IO_Path_Combine_System_String](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=netframework-4.7.2#System_IO_Path_Combine_System_String)___

src/NuGetGallery/Services/LicenseFileService.cs
```diff
@@ -0,0 +1,37 @@
...
+            var serviceDiscoveryClient = new ServiceDiscoveryClient(_config.ServiceDiscoveryUri);
+            var packageBaseAddress = await serviceDiscoveryClient.GetEndpointsForResourceType(CoreConstants.PackageBaseAddress);
+
+            return String.Concat(packageBaseAddress.First().AbsoluteUri, relativePath);
```
loic-sharma: >AbsoluteUri [](start = 60, length = 11)

`AbsoluteUri` can return a string that doesn't end with a `/`. I would recommend ensuring that there's a `/` between `packageBaseAddress.First().AbsoluteUri` and `relativePath`.

src/NuGetGallery/Services/ILicenseFileService.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// This interface is used to get the license file storage path.
+    /// </summary>
+    public interface ILicenseFileService
+    {
+        /// <summary>
+        /// The function is used to get the license file storage path.
+        /// </summary>
+        /// <param name="packageId"> The package ID</param>
+        /// <param name="packageVersion"> The package version</param>
+        Task<string> GetLicenseFileBlobStoragePath(string packageId, string packageVersion);
```
loic-sharma: >packageVersion [](start = 76, length = 14)

It's not clear this needs to be the normalized version string. Could you accept a `NuGetVersion` instead?

src/NuGetGallery/Services/LicenseFileService.cs
loic-sharma: Out of curiosity, how will we remember to update `CoreConstants.PackageBaseAddress` if we bump up the package content's resource version?

---
In reply to: [234839970](https://github.com/NuGet/NuGetGallery/pull/6675#discussion_r234839970) [](ancestors = 234839970)

src/NuGetGallery/Services/LicenseFileService.cs
zhhyu: Add it.

src/NuGetGallery/Services/LicenseFileService.cs
zhhyu: Add it.

src/NuGetGallery/Services/LicenseFileBlobStorageService.cs
```diff
@@ -0,0 +1,42 @@
...
+            }
+            if (packageVersion == null)
+            {
+                throw new ArgumentNullException(nameof(packageVersion));
+            }
+
+            var relativePath = string.Join("/", new string[] { packageId.ToLowerInvariant(), NuGetVersionFormatter.Normalize(packageVersion).ToLowerInvariant(), CoreConstants.LicenseFileName});
+            var packageBaseAddress = await _serviceDiscoveryClient.GetEndpointsForResourceType(CoreConstants.PackageBaseAddress);
+
+            return string.Concat(packageBaseAddress.First().AbsoluteUri, "/", relativePath);
```
loic-sharma: This could return a result like `https://api.nuget.org/v3-flatcontainer//x/1.0.0/...`. How about `packageBaseAddress.First().AbsoluteUri.TrimEnd("/")`?

src/NuGetGallery.Core/CoreConstants.cs
zhhyu: Move to Gallery const now.

src/NuGetGallery/Services/LicenseFileService.cs
zhhyu: Support the local license files now.

src/NuGetGallery/Services/LicenseFileService.cs
zhhyu: Done.

src/NuGetGallery/Controllers/PackagesController.cs
zhhyu: Return 404 for this case.

src/NuGetGallery/Services/LicenseFileService.cs
zhhyu: Add it now.

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -227,6 +227,11 @@ public static void RegisterUIRoutes(RouteCollection routes)
                 "account/ConfirmationRequired",
                 new { controller = "Users", action = "ConfirmationRequired" });
 
+            routes.MapRoute(
+                RouteName.License,
+                "packages/{id}/{version}",
```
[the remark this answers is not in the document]
zhhyu: Change it to the "packages/{id}/{version}/license" now.
It looks that even if with "package/{id}/{version}", it can redirects to the corresponding action when we use the Url like "package/{id}/{version}/License", where "License" is the function name in package controller.

src/NuGetGallery/Services/LicenseFileBlobStorageService.cs
zhhyu: Change it now, and I have found similar examples in Gallery. Thanks!

src/NuGetGallery/Services/ILicenseFileService.cs
zhhyu: We actually pass the "package.NormalizedVersion" at the caller in the controller. So the reason here we use the "NuGetVersionFormatter.Normalize" is to double check the version because it's the key of constructing the path.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -7009,5 +7020,148 @@ public async Task WhenCurrentUserIsAuthenticatedOwner_ReturnsOK(bool multiFactor
...
+
+                // Act
+                await controller.License(_packageId, _packageVersion);
+
+                // Assert
+                _packageFileService
+                    .Verify(p => p.DownloadLicenseFileAsync(package),
+                        Times.Once);
+            }
+        }
```
agr: We probably need to check if returned value is `FileStreamResult` in addition to existing assertion.

tests/NuGetGallery.Facts/Services/LicenseFileFlatContainerServiceFacts.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace NuGetGallery
+{
+    public class LicenseFileFlatContainerServiceFacts
+    {
+        private IAppConfiguration GetConfiguration()
+        {
+            var mockConfiguration = new Mock<IAppConfiguration>();
+            mockConfiguration.SetupGet(c => c.ServiceDiscoveryUri).Returns(new Uri("https://api.nuget.org/v3/index.json"));
+            return mockConfiguration.Object;
```
agr: Tests shouldn't use actual URLs and must not send requests to running services. Services that provide data should be mocked.

src/NuGetGallery/Services/LicenseFileFlatContainerService.cs
```diff
@@ -0,0 +1,42 @@
...
+        private readonly ServiceDiscoveryClient _serviceDiscoveryClient;
+
+        public LicenseFileFlatContainerService(IAppConfiguration configuration)
+        {
+            if (configuration == null)
+            {
+                throw new ArgumentNullException(nameof(configuration));
+            }
+            _serviceDiscoveryClient = new ServiceDiscoveryClient(configuration.ServiceDiscoveryUri);
+        }
```
agr: This should be mockable, otherwise testing `LicenseFileFlatContainerService` would involve sending actual HTTP requests to some service that might not be configured the expected way or not respond properly, which might introduce flakiness to tests.

zhhyu: Done

tests/NuGetGallery.Facts/Services/LicenseFileFlatContainerServiceFacts.cs
zhhyu: Done

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
zhhyu: Done

src/NuGetGallery/Services/LicenseFileFlatContainerService.cs
```diff
@@ -0,0 +1,39 @@
...
+            if (packageVersion == null)
+            {
+                throw new ArgumentNullException(nameof(packageVersion));
+            }
+
+            var packageBaseAddress = await _serviceDiscoveryClient.GetEndpointsForResourceType(GalleryConstants.PackageBaseAddress);
+
+            var licenseUriBuilder = new UriBuilder(packageBaseAddress.First().AbsoluteUri);
+            licenseUriBuilder.Path = string.Join("/", new string[] { packageId.ToLowerInvariant(), NuGetVersionFormatter.Normalize(packageVersion).ToLowerInvariant(), CoreConstants.LicenseFileName });
+
```
agr: `PackageBaseAddress` typically looks like this: `[https://api.nuget.org/v3-flatcontainer/`](https://api.nuget.org/v3-flatcontainer/`). This line will discard the `v3-flatcontainer` portion of the URL and will append the `id/version/license` portion directly after the host name resulting in invalid URL.

tests/NuGetGallery.Facts/Services/LicenseFileFlatContainerServiceFacts.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            if (serviceDiscoveryClient == null)
+            {
+                var uriList = new List<Uri>()
+                {
+                    _packageBaseAddressUri
+                };
+                serviceDiscoveryClient = new Mock<IServiceDiscoveryClient>();
+                serviceDiscoveryClient.Setup(c => c.GetEndpointsForResourceType(It.IsAny<string>())).Returns(Task.FromResult<IEnumerable<Uri>>(uriList));
+            }
```
agr: Please break long lines. You could also use `ReturnsAsync` to save on typing `Task.FromResult`

tests/NuGetGallery.Facts/Services/LicenseFileFlatContainerServiceFacts.cs
```diff
@@ -0,0 +1,78 @@
...
+using Moq;
+using Xunit;
+using System.Collections.Generic;
+
+namespace NuGetGallery
+{
+    public class LicenseFileFlatContainerServiceFacts
+    {
+        private Uri _packageBaseAddressUri = new Uri("https://test.org");
+        private ILicenseFileFlatContainerService createService(Mock<IServiceDiscoveryClient> serviceDiscoveryClient = null)
```
agr: In its current form tests seem to miss the bug that would happen if deployed to actual environment: the package base address looks like `[https://api.nuget.org/v3-flatcontainer/`](https://api.nuget.org/v3-flatcontainer/`) and call to `GetLicenseFileFlatContainerPathAsync` would discard the `v3-flatcontainer` portion of it. It is better to be able to test various base URLs to cover cases like this. Using `[Theory]` instead of `[Fact]` would allow you to pass package base address as an input to a case, so you'd be able to test various cases.

tests/NuGetGallery.Facts/Services/LicenseFileFlatContainerServiceFacts.cs
```diff
@@ -0,0 +1,78 @@
...
+            var packageVersion = "1.0.0";
+            var licenseFileFlatContainerService = createService();
+
+            // Act
+            var licenseFileFlatContainerPath = await licenseFileFlatContainerService.GetLicenseFileFlatContainerPathAsync(packageId, packageVersion);
+
+            // Assert
+            var expectedLicenseFileUriBuilder = new UriBuilder(_packageBaseAddressUri);
+            expectedLicenseFileUriBuilder.Path = string.Join("/", new string[] { packageId.ToLowerInvariant(), NuGetVersionFormatter.Normalize(packageVersion).ToLowerInvariant(), CoreConstants.LicenseFileName });
+            var expectedLicenseFileFlatContainerPath = expectedLicenseFileUriBuilder.Uri.ToString();
```
agr: Please don't copy implementation into test to produce the expected value. It hides bugs. We need as verbatim expected values as possible. In this case it should be something like

`var expectedUri = $"{packageBaseUri}/packageId/1.0.0/license";`

tests/NuGetGallery.Facts/Services/LicenseFileFlatContainerServiceFacts.cs
agr: We also need to verify that it works correctly regardless of the presence of the trailing `/` in the url.

---
In reply to: [238561582](https://github.com/NuGet/NuGetGallery/pull/6675#discussion_r238561582) [](ancestors = 238561582)

agr: >createService [](start = 49, length = 13)

Method names should start with upper-case letter.

tests/NuGetGallery.Facts/Services/LicenseFileFlatContainerServiceFacts.cs
```diff
@@ -0,0 +1,78 @@
...
+using Xunit;
+using System.Collections.Generic;
+
+namespace NuGetGallery
+{
+    public class LicenseFileFlatContainerServiceFacts
+    {
+        private Uri _packageBaseAddressUri = new Uri("https://test.org");
+        private ILicenseFileFlatContainerService createService(Mock<IServiceDiscoveryClient> serviceDiscoveryClient = null)
+        {
```
agr: Please make sure methods have empty lines around them.

src/NuGetGallery/Services/ILicenseFileFlatContainerService.cs
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// This interface is used to get the license file path.
+    /// </summary>
+    public interface ILicenseFileFlatContainerService
```
skofman1: >ILicenseFileFlatContainerService [](start = 21, length = 32)

I would keep the interface more general: IFlatContainerService
Because you will be accessing more files from flat container for icons and readme and will need to add more methods here.

src/NuGetGallery/RouteName.cs
```diff
@@ -98,5 +98,6 @@ public static class RouteName
         public const string GetOrganizationCertificate = "GetOrganizationCertificate";
         public const string GetOrganizationCertificates = "GetOrganizationCertificates";
         public const string SetRequiredSigner = "SetRequiredSigner";
+        public const string License = "License";
     }
```
agr: What is it referenced by?

src/NuGetGallery/Services/IFlatContainerService.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// This interface is used to get the license file path.
+    /// </summary>
+    public interface IFlatContainerService
+    {
+        /// <summary>
+        /// The function is used to get the license file path.
+        /// </summary>
+        /// <param name="packageId"> The package ID</param>
+        /// <param name="packageVersion"> The package version</param>
+        Task<string> GetLicenseFileFlatContainerPathAsync(string packageId, string packageVersion);
```
loic-sharma: >GetLicenseFileFlatContainerPathAsync [](start = 21, length = 36)

Path makes it sound like we're getting a filesystem path. What do you think of renaming this to something like `GetLicenseFileFlatContainerUrlAsync`?

tests/NuGetGallery.Facts/Services/FlatContainerServiceFacts.cs
```diff
@@ -0,0 +1,82 @@
...
+        [InlineData("https://test.org", "https://test.org/packageid/1.0.0/license")]
+        [InlineData("https://test.org/", "https://test.org/packageid/1.0.0/license")]
+        [InlineData("https://test.org/flat-container", "https://test.org/flat-container/packageid/1.0.0/license")]
+        [InlineData("https://test.org/flat-container/", "https://test.org/flat-container/packageid/1.0.0/license")]
+        [InlineData("https://test.org/flat-container//", "https://test.org/flat-container/packageid/1.0.0/license")]
+        public async Task GivenPackageIdAndVersionReturnFlatContainerUrl(string packageBaseUri, string expectedLicenseFileFlatContainerPath)
+        {
+            // Arrange
+            var packageId = "packageId";
+            var packageVersion = "1.0.0";
```
loic-sharma: >packageVersion [](start = 16, length = 14)

Can we also test that this lowercases and normalizes the package version?

src/NuGetGallery/Services/IFlatContainerService.cs
zhhyu: Change it now.

tests/NuGetGallery.Facts/Services/FlatContainerServiceFacts.cs
zhhyu: Done.

PR #6680, opened 2018-11-16
Fixing the readme preview.

agr:
Undoing a bit too overzealous messaging change application.
Tested that readme preview is fixed, tested that errors are displayed properly for preview, too.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -2319,18 +2319,18 @@ namespace NuGetGallery
         {
             if (formData == null || !_readMeService.HasReadMeSource(formData))
             {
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -2319,18 +2319,18 @@ public virtual async Task<JsonResult> PreviewReadMe(ReadMeRequest formData)
         {
             if (formData == null || !_readMeService.HasReadMeSource(formData))
             {
-                return Json(HttpStatusCode.BadRequest, new[] { new JsonValidationMessage(Strings.PreviewReadMe_ReadMeMissing) });
+                return Json(HttpStatusCode.BadRequest, new[] { Strings.PreviewReadMe_ReadMeMissing });
```
skofman1: Please add UTs in a follow up PR

agr: #6682

PR #6695, opened 2018-11-26
add plus sign between version and build metadata

adamralph:
This is unnerving:

### DIFF
--- a/src/NuGetGallery/Strings.resx
+++ b/src/NuGetGallery/Strings.resx
@@ -160,7 +160,7 @@
     <value>The file must be a .nupkg or .snupkg file.</value>
   </data>
   <data name="SuccessfullyUploadedPackage" xml:space="preserve">
-    <value>You successfully uploaded {0} {1}.</value>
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Strings.resx
```diff
@@ -160,7 +160,7 @@
     <value>The file must be a .nupkg or .snupkg file.</value>
   </data>
   <data name="SuccessfullyUploadedPackage" xml:space="preserve">
-    <value>You successfully uploaded {0} {1}.</value>
+    <value>You successfully uploaded {0}+{1}.</value>
```
loic-sharma: I don't think this is correct, as `{0}` is the package id and `{1}` is the package version (see [this code](https://github.com/NuGet/NuGetGallery/blob/ac41391913bb309a3749df81fceb1a5c58689195/src/NuGetGallery/Controllers/PackagesController.cs#L2182)). This will show messages like `You successfully uploaded Newtonsoft.Json+12.0.0`.

adamralph: image

You're right.

PR #6698, opened 2018-11-27
Fix casing and rewording of left nav items

anangaur:
The second word should start with lowercase. See other texts on a package's page. `Source Code` renamed to `Source repository`.


anangaur:
@skofman1 Please merge if you agree with the change and is approved.

Issue #6678, opened 2018-11-16
Add Open Graph support for social media


skofman1:
Instructions: https://css-tricks.com/essential-meta-tags-social-media/#article-header-id-1

Let's make sure our front page looks nice.

PR #6708, opened 2018-11-28
 Add Open Graph metadata for the front page

joelverhagen:
image

https://cards-dev.twitter.com/validator

Address https://github.com/NuGet/NuGetGallery/issues/6678


joelverhagen:
I am in the process of:

1. [ ] Changing the image to 600x600 so there is a bit of space around the logo.
1. [ ] Changing the text to:
    - The NuGet Gallery is the central package repository for NuGet, the package manager for .NET.

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -1299,7 +1300,52 @@
...
+    <Content Include="Content\gallery\img\purple-circle.svg" />
+    <Content Include="Content\gallery\img\reserved-indicator-20x20.png" />
+    <Content Include="Content\gallery\img\reserved-indicator-256x256.png" />
+    <Content Include="Content\gallery\img\reserved-indicator-25x25.png" />
+    <Content Include="Content\gallery\img\reserved-indicator.svg" />
+    <Content Include="Content\gallery\img\rss-36x36.png" />
+    <Content Include="Content\gallery\img\rss.svg" />
+    <Content Include="Content\gallery\img\twitter-24x24.png" />
+    <Content Include="Content\gallery\img\twitter-36x36.png" />
+    <Content Include="Content\gallery\img\twitter.svg" />
```
agr: That's a lot of unreferenced resources it seems. Can you comment on why do we need all those?

joelverhagen: Looks like VS expands `<Content Include="Content\gallery\**\*" />` when I add a content file.

joelverhagen: I'd rather not fight the tooling -- if VS wants it this way, I am fine with it.

src/NuGetGallery/Views/Pages/Home.cshtml
```diff
@@ -6,6 +6,16 @@
     ViewBag.HasJumbotron = true;
 }
 
+@section SocialMeta {
```
chenriksson: Should we consider support from `Downloads` page, or from `Blog`? Are those likely to be shared?

src/NuGetGallery/Views/Pages/Home.cshtml
```diff
@@ -6,6 +6,16 @@
...
+    <meta name="twitter:card" content="summary">
+    <meta name="twitter:site" content="@("@nuget")">
+
+    <meta property="og:title" content="@(Config.Current.Brand) @(String.IsNullOrWhiteSpace(ViewBag.Title) ? "" : "| " + ViewBag.Title)" />
+    <meta property="og:url" content="@(Url.Home(relativeUrl: false))" />
+    <meta property="og:description" content="The NuGet Gallery is the central package repository for NuGet, the package manager for .NET." />
+    <meta property="og:image" content="@(Url.Absolute("~/Content/gallery/img/logo-og-600x600.png"))" />
```
chenriksson: Comparing against `view-source:https://github.com/`, looks like only the `og:` metadata is necessary? Or is `twitter:*` actually used and recommended?

Should we include image type, width or height?

anangaur:
Can somebody share the new card look for different pages: NuGet.org, a package page, blog, downloads page?

src/NuGetGallery/NuGetGallery.csproj
scottbommarito: I **strongly** prefer the unexpanded form. Is there any setting you could find to turn this off? This massive list of references is unmanageable.

joelverhagen: What do you mean by "unmanageable"? This is typically edited through VS. Even with wildcards, it's a long list in the UI. We could take a route like [this](https://stackoverflow.com/a/21680084) but we already have this expansion for C# files. I don't see why we should treat content different, considering the default behavior in tooling is fighting us.

src/NuGetGallery/Views/Pages/Home.cshtml
joelverhagen: Tracked here: https://github.com/NuGet/NuGetGallery/issues/6709

src/NuGetGallery/Views/Pages/Home.cshtml
joelverhagen: This provides content attribution. See https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started.html.

joelverhagen:
> Can somebody share the new card look for different pages: NuGet.org, a package page, blog, downloads page?

@anangaur, I have only updated the card for the front page. This was what was originally time estimated and requested. I have tracked https://github.com/NuGet/NuGetGallery/issues/6709 to cover the other pages.

src/NuGetGallery/NuGetGallery.csproj
scottbommarito: Just because you edit it with VS doesn't make it easy to manage.

When you add a `.cs` file, you're usually creating it with VS. With a content file, you're probably copying it from somewhere else and pasting it into the folder manually. It's a little obnoxious to have to then manually include it in VS when your intentions were very clear to begin with.

I personally don't like how VS handles `.cs` files either--it's just so confusing to have code in the project's folders that isn't included in the project--but I do think it's worse with `content` files or things like our jobs' `PowerShell` scripts.

joelverhagen: > It's a little obnoxious to have to then manually include it in VS when your intentions were very clear to begin with. 

"Show All Files" in solution explorer makes this easier.

scottbommarito: Interesting. I'll try it out.

anangaur:
> @anangaur, I have only updated the card for the front page. This was what was originally time estimated and requested. I have tracked #6709 to cover the other pages.

Works for me. I typically share the blog items and downloads page on twitter apart from GitHub issues pages.

Issue #6684, opened 2018-11-17
Deleting organization with one co-owned package fails


joelverhagen:
![image](https://user-images.githubusercontent.com/94054/48654412-64913d00-e9c1-11e8-9dde-f86ca5db16ed.png)

```
System.InvalidOperationException: Deleting user 'jverorg' will make package 'System.Rido' an orphan, but no orphans were expected.
   at NuGetGallery.DeleteAccountService.<RemovePackageOwnership>d__19.MoveNext() in E:\A\_work\172\s\src\Gallery\submodules\Gallery\src\NuGetGallery\Services\DeleteAccountService.cs:line 192
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGetGallery.DeleteAccountService.<DeleteAccountImplAsync>d__14.MoveNext() in E:\A\_work\172\s\src\Gallery\submodules\Gallery\src\NuGetGallery\Services\DeleteAccountService.cs:line 104
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGetGallery.DeleteAccountService.<RunAccountDeletionTask>d__28.MoveNext() in E:\A\_work\172\s\src\Gallery\submodules\Gallery\src\NuGetGallery\Services\DeleteAccountService.cs:line 319
```

This happened on DEV environment.

scottbommarito:
I think this is specifically because the only member of the org is the same as the only remaining owner of the package.

See https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Services/DeleteAccountService.cs#L186 and https://github.com/NuGet/NuGetGallery/blob/73a5c54629056b25b3a59960373e8fef88abff36/src/NuGetGallery/Services/PackageService.cs#L302

This logic should be made smarter.

PR #6714, opened 2018-11-30
Improve the logic for how we determine whether or not packages will be orphaned by an account deletion.

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6684

Previously, we determined that a package would be orphaned by deleting a user if that package was owned by only a single distinct user account.

@joelverhagen demonstrated that this logic was faulty by trying to delete an organization with a single member that owned a package that the single member also only owned. The logic saw that the package was owned by only a single user account and thought it would be orphaned when it actually wouldn't because it would still be owned by the member.

We now determine a package will be orphaned if it has no other user account owners **other than the user that is being deleted**. In other words, the logic is now filtering out the account that is being deleted and checking if there are **any** user account owners. Furthermore, the delete account and delete organization pages used different logic (but similar in implementation) to determine whether or not a package will be orphaned. I have moved the logic to `IPackageService` and shared it between the two.

This PR improves the logic and improves unit testing so that we test against this scenario. I noticed that we didn't have any tests on the different orphan policies, which is where this bug surfaced, so I have fixed that as well.


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -377,6 +357,20 @@ public async Task RemovePackageOwnerAsync(PackageRegistration package, User user
                 await _packageRepository.CommitChangesAsync();
             }
         }
+        
+        public bool WillPackageBeOrphanedIfOwnerRemoved(PackageRegistration package, User owner)
+        {
+            var remainingOwners = package.Owners
```
loic-sharma: This code is getting a little gnarly for linq. Maybe use a foreach instead?

tests/NuGetGallery.Facts/Services/DeleteAccountServiceFacts.cs
```diff
@@ -79,71 +85,116 @@ public async Task DeleteDeletedUser()
...
+                    Assert.Null(testUser.OrganizationMigrationRequest);
+                    Assert.Empty(testUser.OrganizationMigrationRequests);
+                    Assert.Empty(testUser.OrganizationRequests);
+
+                    Assert.Empty(testUser.Organizations);
+                    foreach (var testUserOrganization in testUserOrganizations)
                     {
-                        Assert.Contains(notDeletedMembers, m => m.IsAdmin);
+                        var notDeletedMembers = testUserOrganization.Organization.Members.Where(m => m.Member != testUser);
+                        if (!notDeletedMembers.Any())
```
dtivel: Invert the logic to make it more easily readable.  Instead of:

    if (!notDeletedMembers.Any()) { } else { }

...do:

    if (notDeletedMembers.Any()) { } else { }

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -2076,6 +2076,164 @@ public async Task WontRemoveLastOwner()
...
+
+            public enum AccountToDelete
+            {
+                User1,
+                User2,
+                Organization1,
+                Organization2
+            }
+
+            public static IEnumerable<object[]> WillBeOrphaned_Input
```
dtivel: Nice!

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -2076,6 +2076,164 @@ public async Task WontRemoveLastOwner()
...
+                        }
+                    }
+                }
+            }
+
+            [Theory]
+            [MemberData(nameof(WillBeOrphaned_Input))]
+            public void WillBeOrphaned(OwnershipState state, AccountToDelete accountToDelete)
+            {
+                var user1 = new User("testUser") { Key = 0 };
```
dtivel: Postpend 1 to `testUser` for symmetry.  Same with `testOrganization` below.

src/NuGetGallery/ViewModels/DeleteAccountViewModel.cs
```diff
@@ -25,37 +24,43 @@ public class DeleteAccountViewModel<TAccount> : DeleteAccountViewModel where TAc
...
             User = userToDelete;
 
             Packages = packageService
                  .FindPackagesByAnyMatchingOwner(User, includeUnlisted: true)
-                 .Select(p => new ListPackageItemViewModel(p, currentUser))
+                 .Select(p => new DeleteAccountListPackageItemViewModel(p, userToDelete, currentUser, packageService))
                  .ToList();
 
-            _hasOrphanPackages = new Lazy<bool>(() => Packages.Any(packageIsOrphaned));
+            HasOrphanPackages = Packages.Any(p => p.WillBeOrphaned);
```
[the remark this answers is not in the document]
scottbommarito: `HasPackagesThatWillBeOrphaned`

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -2076,6 +2076,164 @@ public async Task WontRemoveLastOwner()
...
+                }
+
+                if (state.HasFlag(OwnershipState.OwnedByOrganization2))
+                {
+                    package.Owners.Add(organization2);
+                }
+
+                var expectedResult = true;
+                User userToDelete;
+                if (accountToDelete == AccountToDelete.User1)
```
chenriksson: Consider extracting this block into `GetWillBeOrphanedExpectedResult(User accountToDelete)`

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -2076,6 +2076,164 @@ public async Task WontRemoveLastOwner()
...
+            [Theory]
+            [MemberData(nameof(WillBeOrphaned_Input))]
+            public void WillBeOrphaned(OwnershipState state, AccountToDelete accountToDelete)
+            {
+                var user1 = new User("testUser1") { Key = 0 };
+                var user2 = new User("testUser2") { Key = 1 };
+                var organization1 = new Organization("testOrganization1") { Key = 2 };
+                var organization2 = new Organization("testOrganization2") { Key = 3 };
+
+                if (state.HasFlag(OwnershipState.User1InOrganization1))
```
chenriksson: Consider extracting setup into helper `SetWillBeOrphanedMemberships(User user1, User user2, Organization organization1, Organization organization2, OwnershipState state)`?

src/NuGetGallery/Services/PackageService.cs
scottbommarito: Done!

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
scottbommarito: I'll just add comments to increase the readability. I think extracting into a method would require me to do some obnoxious parameter things like `out` variables.

PR #6726, opened 2018-12-05
Controller should redirect to licenses.nuget.org with license expression

zhhyu:
Controller should redirect to licenses.nuget.org with license expression.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -677,6 +677,11 @@ namespace NuGetGallery
                 return HttpNotFound();
             }
 
+            if (!string.IsNullOrWhiteSpace(package.LicenseExpression))
+            {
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -677,6 +677,11 @@ public virtual async Task<ActionResult> License(string id, string version)
                 return HttpNotFound();
             }
 
+            if (!string.IsNullOrWhiteSpace(package.LicenseExpression))
+            {
+                return Redirect(LicenseExpressionRedirectUrlHelper.GetLicenseExpressionRedirectUrl(package.LicenseExpression));
```
chenriksson: Do we have monitoring for licenses.nuget.org?

zhhyu: Yeah, I think we have some kinds of monitoring for "license.nuget.org".

PR #6728, opened 2018-12-06
RequirePackageMetadataCompliancePolicy: allow exceptions for authors

xavierdecoster:
Fixes NuGet/Engineering#1955, NuGet/Engineering#1954

### DIFF
--- a/src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
+++ b/src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
@@ -118,29 +118,7 @@ namespace NuGetGallery.Security
             complianceFailures = new List<string>();
 
             // Author validation
-            var packageAuthors = package.FlattenedAuthors
-                .Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
-                .Select(s => s.Trim());
-
-            if (state.AllowedAuthors?.Length > 0)
-            {
-                foreach (var packageAuthor in packageAuthors)
-                {
-                    if (!state.AllowedAuthors.Contains(packageAuthor))
-                    {
-                        complianceFailures.Add(string.Format(CultureInfo.CurrentCulture, Strings.SecurityPolicy_PackageAuthorNotAllowed, packageAuthor));
-                    }
-                }
-            }
-            else
-            {
-                // No list of allowed authors is defined for this policy.
-                // We require the required co-owner to be defined as the only package author.
-                if (packageAuthors.Count() > 1 || packageAuthors.Single() != state.RequiredCoOwnerUsername)
-                {
-                    complianceFailures.Add(string.Format(CultureInfo.CurrentCulture, Strings.SecurityPolicy_RequiredAuthorMissing, state.RequiredCoOwnerUsername));
-                }
-            }
+            ValidatePackageAuthors(package, state, complianceFailures);
 
             // Copyright validation
             if (!state.AllowedCopyrightNotices.Contains(package.Copyright))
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
```diff
@@ -116,12 +118,28 @@ private bool IsPackageMetadataCompliant(Package package, State state, out IList<
...
                 .Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
-                .Select(s => s.Trim())
-                .Contains(state.RequiredCoOwnerUsername, StringComparer.InvariantCultureIgnoreCase))
+                .Select(s => s.Trim());
+
+            if (state.AllowedAuthors?.Length > 0)
+            {
+                foreach (var packageAuthor in packageAuthors)
+                {
+                    if (!state.AllowedAuthors.Contains(packageAuthor))
```
chenriksson: Per https://github.com/NuGet/Engineering/issues/1954, do we still need to restrict to 1 author? Microsoft or the other?
/cc @karann-msft

src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
```diff
@@ -116,12 +118,28 @@ private bool IsPackageMetadataCompliant(Package package, State state, out IList<
...
+                        complianceFailures.Add(string.Format(CultureInfo.CurrentCulture, Strings.SecurityPolicy_PackageAuthorNotAllowed, packageAuthor));
+                    }
+                }
+            }
+            else
             {
-                complianceFailures.Add(string.Format(CultureInfo.CurrentCulture, Strings.SecurityPolicy_RequiredAuthorMissing, state.RequiredCoOwnerUsername));
+                // No list of allowed authors is defined for this policy.
+                // We require the required co-owner to be defined as the only package author.
+                if (packageAuthors.Count() > 1 || packageAuthors.Single() != state.RequiredCoOwnerUsername)
```
chenriksson: Can we guard against data containing both `AllowedAuthors` and `RequiredCoOwnerUsername`?

src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
```diff
@@ -116,12 +118,28 @@ private bool IsPackageMetadataCompliant(Package package, State state, out IList<
             complianceFailures = new List<string>();
 
             // Author validation
-            if (!package.FlattenedAuthors
+            var packageAuthors = package.FlattenedAuthors
```
skofman1: This doesn't satisfy Karan's requirement. Please see comment here: [https://github.com/NuGet/Engineering/issues/1954#issuecomment-444951761](https://github.com/NuGet/Engineering/issues/1954#issuecomment-444951761)

tests/NuGetGallery.Facts/Security/RequirePackageMetadataCompliancePolicyFacts.cs
```diff
@@ -206,6 +205,83 @@ public async Task Evaluate_NonCompliantPackage_CreatesErrorResult(Package nonCom
...
+        }
+
+        [Fact]
+        public async Task Evaluate_CompliantPackageAuthors_CreatesSuccessResult()
+        {
+            // Arrange
+            var nugetUser = new User("NuGet");
+            var newPackageRegistration = new PackageRegistration { Id = "NewPackageId", Owners = new List<User> { nugetUser } };
+            var packageAuthors = new[] { MicrosoftTeamSubscription.MicrosoftUsername, "The Most-Awesome Package Authors" };
+            var nonCompliantPackage = Fakes.CreateCompliantPackage("1.0.0", newPackageRegistration, packageAuthors);
```
skofman1: >nonCompliantPackage [](start = 16, length = 19)

compliant

src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
karann-msft: I believe it is ok for it to be a comma-separated list as long as each element is one from the allowed list. I am assuming hat the list would not allow repeated values. For example - `Microsoft, Microsoft` fails validation.

xavierdecoster: Never assume... :) This code doesn't check for duplicates. Will fix.

tests/NuGetGallery.Facts/Security/RequirePackageMetadataCompliancePolicyFacts.cs
xavierdecoster: fixed :)

xavierdecoster:
Now also checking for duplicate package authors in metadata and generating a compliance error when duplicates are found: https://github.com/NuGet/NuGetGallery/pull/6728/commits/5d79c7585cc887a519d56d3d1064e9955ebdb7bc

src/NuGetGallery/Security/RequirePackageMetadataCompliancePolicy.cs
chenriksson: It looks like `RequiredCoOwnerUsername` is only used if `AllowedAuthors` is not specified. Since the behavior is either/or, I think it may still be good to enforce only one option in the data... otherwise it can create confusion.

xavierdecoster: `RequiredCoOwnerUsername` is used elsewhere, to auto-add the specific ownership. It is also used here to verify the *author* information, as a fallback, if no `AllowedAuthors` is specified.

All new policy subscriptions will have `AllowedAuthors` specified, but the old ones (the ones already subscribed today) do not, and need the fallback.

Issue #6737, opened 2018-12-07
[Licenses] Gallery is sensitive to path separator type used in license file paths


agr:
`<license type="file">license\license.txt</license>` vs `<license type="file">license/license.txt</license>`. Both cases pack fine, bit the first one is rejected by Gallery. We need to support both.

PR #6739, opened 2018-12-07
Fix for the Gallery unable to find license file if paths with backslashes are used

agr:
Fixes https://github.com/NuGet/NuGetGallery/issues/6737.

Per zip file format [specification](http://www.pkware.com/documents/casestudies/APPNOTE.TXT) (4.4.17.1), only forward slashes can be used in paths inside zip file, but paths in nuspec can contain both forward and back slashes. So added code that fixes the paths before they are passed into `PackageArchiveReader`.

On the other hand, `ZipArchive` implementation in .net does not enforce forward slashes and allows to create zip files with backslashes in paths, so alternative pack implementations might exist that produce technically invalid zip files, not sure we want to support those though.

### DIFF
--- a/tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
+++ b/tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
@@ -1046,6 +1046,26 @@ namespace NuGetGallery
                 Assert.Empty(result.Warnings);
             }
 
+            [Theory]
+            [InlineData("license/something.txt")]
+            [InlineData("license\\something.txt")]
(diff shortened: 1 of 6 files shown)


tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
```diff
@@ -1046,6 +1046,26 @@ public async Task RejectsNupkgsWithUnknownLicenseTypes(string licenseNode)
                 Assert.Empty(result.Warnings);
             }
 
+            [Theory]
+            [InlineData("license/something.txt")]
```
nkolev92: Might be useful to test things like `.\license\something.txt` as well.

Issue #6741, opened 2018-12-08
Latest package count includes unlisted packages


SteveWilkes:
The latest package download count on the Display Package view does not filter out unlisted packages. This can cause the latest package download count to differ from that of the latest listed package. It can be seen on [the AgileMapper](https://www.nuget.org/packages/AgileObjects.AgileMapper) page at the time of writing.

I will create a PR to fix this shortly.

SeanKilleen:
Looking at this via the `up-for-grabs` tag and it seems as though it is fixed, verified & merged. Looks like this can be closed?

SteveWilkes:
Yep! Closing :)

chenriksson:
Reopening for tracking. Will close once this reaches production environment. /cc @SteveWilkes

samedder:
I believe this is back, you can test this with https://www.nuget.org/stats/packages/Microsoft.ServiceFabric?groupby=Version

PR #6742, opened 2018-12-08
Fix latest download count, re: issue #6741

SteveWilkes:
Fixes issue #6741.


skofman1:
@SteveWilkes , this will break if there are no listed versions at all..

SteveWilkes:
I've closed this PR and opened #6745 to address the same issue.

PR #6745, opened 2018-12-10
Fix latest download count, re: issue #6741

SteveWilkes:
Updated fix for issue #6741 from PR #6742.

Displays the latest listed package download count, or no package download count if there are no listed packages.

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -748,11 +748,18 @@
                     <i class="ms-Icon ms-Icon--Download" aria-hidden="true"></i>
                     @Model.TotalDownloadCount.ToNuGetNumberString() total @(Model.TotalDownloadCount == 1 ? "download" : "downloads")
                 </li>
-                <li>
-                    <i class="ms-Icon ms-Icon--Giftbox" aria-hidden="true"></i>
-                    @Model.PackageVersions.First().DownloadCount.ToNuGetNumberString() @(Model.PackageVersions.First().DownloadCount == 1 ? "download" : "downloads")
-                    of latest version
-                </li>
+                @{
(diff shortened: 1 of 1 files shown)


chenriksson:
@SteveWilkes Thanks for your contribution!

I think a more consistent experience would be to always show this version-specific download count, but default to the current version instead of the latest. By default, `DisplayPackage.cshtml` displays the latest listed version but could show an unlisted version if no listed are available.

This would simplify the logic to just:
```
@Model.DownloadCount.ToNuGetNumberString() @(Model.DownloadCount == 1 ? "download" : "downloads")	
                    of current version
```

scottbommarito:
I think we need to ask the question "what do users want to see here" and then answer that before we go forward with this.

I think we may also want to consider showing the download count of
- the latest stable version
- the most popular version (the one with the most downloads)

I tend to agree with @chenriksson that "current version" is probably the most useful thing to show there, however.

SteveWilkes:
Thanks for the feedback! I've updated the PR to use `Model.DownloadCount` as requested.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -750,7 +750,7 @@
                 </li>
                 <li>
                     <i class="ms-Icon ms-Icon--Giftbox" aria-hidden="true"></i>
-                    @Model.PackageVersions.First().DownloadCount.ToNuGetNumberString() @(Model.PackageVersions.First().DownloadCount == 1 ? "download" : "downloads")
+                    @Model.DownloadCount.ToNuGetNumberString() @(Model.DownloadCount == 1 ? "download" : "downloads")
                     of latest version
```
scottbommarito: This needs to be updated to say "current version".

SteveWilkes: Done! :)

chenriksson:
Verified and merged... thanks @SteveWilkes!

Issue #2992, opened 2016-04-17
RSS/Atom Feed for Packages


robertmuehsig:
This idea is pretty old, but I would like to see such a feature on NuGet.org: A simple feed for packages updates. The other issue about the badge API did remind me about this obvious feature.

Two years ago I tried it with this old PR https://github.com/NuGet/NuGetGallery/pull/2128 - at this time the NuGet API v3 was in the making and NuGet.org was fighting against OData/SQL issues, so the PR got rejected.

My idea would be that the feed should contain only the same information as the version history shows:
![image](https://cloud.githubusercontent.com/assets/756703/14590194/2a898b20-04f5-11e6-80ec-15086b69c33b.png)

I'm not sure how this would impact the server - even with a very long output cache duration (e.g. output cache duration for a whole day). 

The use case for such a feed would be to do such things like use ifttt to send an email when an package got updated.

What do you think? This "feed" should be a human readable and accessible feed, whichout knowing to much about the NuGetV3 api.

maartenba:
Is RSS an implementation detail or a feature?

Reason I'm asking is we're planning on creating a "replication API" which will give you a nice library to fetch things like "latest since" from our catalog store. Of course, that's not an RSS feed as such, but if the goal is fetching latest items then it'll do fine.

robertmuehsig:
The goal would be to provide a really simple RSS feed, which points to the details page of each version - without using a API. 
Currently I could use the NuGet API to build a NuGet2RSS web app myself - I thought it would be nice to have such functionality builtin.

RubyGems has something like this implemented: 
![image](https://cloud.githubusercontent.com/assets/756703/14596732/e00a4aa4-0548-11e6-8e72-dd80f353f060.png)

https://rubygems.org/gems/mysql & the feed: https://rubygems.org/gems/mysql/versions.atom

maartenba:
If it's per package, the RSS feed could easily be based off the registrations. E.g. `https://api.nuget.org/v3/registration0/newtonsoft.json/index.json` or even the flat container for DNX (`https://api.nuget.org/v3/flatcontainer/newtonsoft.json/index.json`).

Let's check first if we'd take it. @yishaigalatzer @harikm86 is an RSS feed per package something we'd like to have short term?

robertmuehsig:
The registration json seems good - it contains version and publishing date, which should fit in the feed model.

[...]

PR #6747, opened 2018-12-10
[WIP] Adding a RSS Feed for packages

robertmuehsig:
Based on this very old issue https://github.com/NuGet/NuGetGallery/issues/2992 I will try it again.

The goal is to add a package feed to the NuGet Gallery like [RubyGems](https://rubygems.org/gems/mysql/versions.atom).

In [this comment](https://github.com/NuGet/NuGetGallery/issues/2992#issuecomment-222423263) I summarized the dicussion that we had in the past:

> Maybe something like this:
>Introduce a new config entry like "Gallery.ExposePackageAsSyndicationFeedMode"
>
>With following possible values:
>
>Off (so no RSS/Atom Feed is available - like today)
>UseLocalDatabase => for On-Prem
>UseV3Api => for NuGet.org
>If value="UseV3Api" the "ServiceDiscoveryUri" must also be set.

The current implementation just uses the OnPrem way:
image

With this feed:

<details>

```
<rss version="2.0">
    <channel>
        <title>NuGet Package 'NUnit' Feed</title>
        <link>https://localhost/packages/NUnit/</link>
        <description>NUnit features a fluent assert syntax, parameterized, generic and theory tests and is user-extensible.

This package includes the NUnit 3 framework assembly, which is referenced by your tests. You will need to install version 3 of the nunit3-console program or a third-party runner that supports NUnit 3 in order to execute tests. Runners intended for use with NUnit 2.x will not run NUnit 3 tests correctly.

Supported platforms:
- .NET 2.0+
- .NET Standard 1.3 and 1.6
- .NET Core</description>
        <lastBuildDate>Mon, 10 Dec 2018 18:36:46 GMT</lastBuildDate>
        <language>en-us</language>
            <item>
                <title>3.9.0;</title>
                <link>
                https://localhost/packages/NUnit/3.9.0
                </link>
                <guid>https://localhost/packages/NUnit/3.9.0</guid>
                <pubDate>Mon, 10 Dec 2018 18:36:46 GMT</pubDate>
                <description>This package includes the NUnit 3 framework assembly, which is referenced by your tests. You will need to install version 3 of the nunit3-console program or a third-party runner that supports NUnit 3 in order to execute tests. Runners intended for use with NUnit 2.x will not run NUnit 3 tests correctly.</description>
            </item>
            <item>
                <title>3.8.1;</title>
                <link>
                https://localhost/packages/NUnit/3.8.1
                </link>
                <guid>https://localhost/packages/NUnit/3.8.1</guid>
                <pubDate>Mon, 10 Dec 2018 18:38:51 GMT</pubDate>
                <description>This package includes the NUnit 3 framework assembly, which is referenced by your tests. You will need to install version 3 of the nunit3-console program or a third-party runner that supports NUnit 3 in order to execute tests. Runners intended for use with NUnit 2.x will not run NUnit 3 tests correctly.</description>
            </item>
            <item>
                <title>3.5.0;</title>
                <link>
                https://localhost/packages/NUnit/3.5.0
                </link>
                <guid>https://localhost/packages/NUnit/3.5.0</guid>
                <pubDate>Fri, 18 Aug 2017 18:48:49 GMT</pubDate>
                <description>This package includes the NUnit 3.0 framework assembly, which is referenced by your tests. You will need to install version 3.0 of the nunit3-console program or a third-party runner that supports NUnit 3.0 in order to execute tests. Runners intended for use with NUnit 2.x will not run 3.0 tests correctly.</description>
            </item>
    </channel>
</rss>


```
</details>

The implementation side it's pretty simple.

Question now is: Is the NuGet team still interested in this feature and is the above plan still valid?

If yes: 
Would this be a good API endpoint to fetch the needed package metadata? https://api.nuget.org/v3/registration3-gz-semver2/newtonsoft.json/index.json 

As far as I understand the NuGet API only the registration would hold the needed package information (I'm not sure if I understand the structure right now, but this seems to be manageable ;) )

What do you think?


skofman1:
Thanks for the PR @robertmuehsig ! We would like to work with you on some polishing and accept it. @joelverhagen will provide details 😄

joelverhagen:
Hey @robertmuehsig. Thanks for moving forward with this! Here's an initial round of feedback:

## Response body

Have we considered Atom over RSS? Why was RSS picked? I don't have a strong feeling yet but our blog exposes the posts as Atom so this is a little inconsistent.

If we go with RSS, can we use `Rss20FeedFormatter` or some other built-in type to do the serialization?

What is the content-type returned here? I think it should be `application/rss+xml`.

The [example RSS](http://static.userland.com/gems/backend/rssTwoExample2.xml) feed has `<?xml version="1.0"?>` so let's have that too.

## Filtering

I think this should hide unlisted packages and hide anything but `PackageStatusKey.Available`.

What do you think about showing or hiding prerelease? I think the default should show both prerelease and stable, but if you have a sense that people would only want stable maybe we can add a `prerel=true | false` query parameter like the search page but not provide any UI around it.

## Paging, sorting

AFAIK paging is not supported in RSS, so we are just returning all versions, right? Seems fine to me since this is exactly what the package details page is doing.

It looks like the sort order you have is by SemVer order. This is not necessarily the same as the publish order. Wouldn't it be more in the spirit of RSS to sort descending by published or last edited date? I don't feel strongly but perhaps we should try to align with what RSS users expect (whatever that ordering might be).

## What API to call?

I think going to the database is appropriate here. It is basically the same work (if not less) as the package details page. If the RSS feed begins to have scalability issues we can throttle/cache using APIM (which is sitting in front of nuget.org) or switch to another solution like what the V2 API has, i.e. redirecting queries to the search service.

Theoretically, we could call the same APIs as `/api/v2/FindPackagesById()` if we are okay with sorting by published date.
https://www.nuget.org/api/v2/FindPackagesById()?id=%27NLog%27&$orderby=Published%20desc

I don't think we should implement querying the V3 protocol at all here. That's a lot of complexity.

## UI on package details page

I talked to @anangaur and he would like to put the RSS button under the "Share" section. The alt text should be a descriptive phrase like "Use RSS to subscribe to new versions of {ID}". The icon should be the same size as the Twitter and Facebook icons. 

## URL

The URL should be `https://www.nuget.org/packages/{ID}/rss.xml`. This is okay since `rss.xml` will never collide with a package version.

## RSS fields

w.r.t. `<channel>`

Field Name | Feedback
--- | ---
title | Isn't the word "feed" at the end a little redundant? Or is this idiomatic?
lastBuildDate | From reading this should be the max of last edited across _all_ packages (including unlisted or non-`Available`). This should change whenever a version is removed due to delete/unlist or added.
pubDate | This wasn't in your PR but maybe this should be set to the max Created or LastEdited of the visible packages.
image | Can we make this the latest version's package icon?
item/description | This should be the package description not release notes, right?

## Unit tests

Could we have some unit tests calling the MVC action testing things like:

1. Happy path
1. ID does not exist
1. ID has no versions
1. ID has only unlisted versions
1. Filters out non-`PackageStatusKey.Available`
1. Filters out unlisted

## End

Overall, I love the idea and think this is a great extension point for nuget.org.

robertmuehsig:
Great to hear from you! I picked RSS without any deep feelings. Both formats should be "fine" for this usecase, but I will give Atom a try.

The filtering: Yes - good idea to keep the "main" feed clean of pre-releases. 

Sorting: Yeah - it would make sense to use the published date.

I hope to push an updated version in the upcoming week.

robertmuehsig:
I found the time to give it a try with Atom feeds. I switched from RSS to ATOM by using the "SyndicationItem" stuff, which is part of the .NET Framework. 
I also did the UI work and the ATOM feed should at least be "valid".

Some screenshots:

image

image

A example feed could look like this:
```
<?xml version="1.0" encoding="utf-8"?>
<feed xmlns="http://www.w3.org/2005/Atom">
  <title type="text">NuGet.org: NUnit</title>
  <subtitle type="text">NUnit is a unit-testing framework for all .NET languages with a strong TDD focus.</subtitle>
  <id>https://localhost/packages/NUnit/</id>
  <updated>2019-02-06T22:46:05+01:00</updated>
  <logo>https://cdn.rawgit.com/nunit/resources/master/images/icon/nunit_256.png</logo>
  <link rel="self" type="application/atom+xml" href="https://localhost/packages/NUnit/atom.xml" />
  <link rel="alternate" type="text/html" href="https://localhost/packages/NUnit/" />
  <entry>
    <id>https://localhost/packages/NUnit/3.9.0</id>
    <title type="text">NUnit: 3.9.0</title>
    <updated>2018-12-10T19:36:46+01:00</updated>
    <author>
      <name>Charlie Poole, Rob Prouse</name>
    </author>
    <link rel="alternate" href="https://localhost/packages/NUnit/3.9.0" />
    <content type="text">NUnit features a fluent assert syntax, parameterized, generic and theory tests and is user-extensible.

This package includes the NUnit 3 framework assembly, which is referenced by your tests. You will need to install version 3 of the nunit3-console program or a third-party runner that supports NUnit 3 in order to execute tests. Runners intended for use with NUnit 2.x will not run NUnit 3 tests correctly.

Supported platforms:
- .NET 2.0+
- .NET Standard 1.3 and 1.6
- .NET Core</content>
  </entry>
  <entry>
    <id>https://localhost/packages/NUnit/3.8.1</id>
    <title type="text">NUnit: 3.9.0</title>
    <updated>2018-12-10T19:38:51+01:00</updated>
    <author>
      <name>Charlie Poole, Rob Prouse</name>
    </author>
    <link rel="alternate" href="https://localhost/packages/NUnit/3.8.1" />
    <content type="text">NUnit features a fluent assert syntax, parameterized, generic and theory tests and is user-extensible.

This package includes the NUnit 3 framework assembly, which is referenced by your tests. You will need to install version 3 of the nunit3-console program or a third-party runner that supports NUnit 3 in order to execute tests. Runners intended for use with NUnit 2.x will not run NUnit 3 tests correctly.

Supported platforms:
- .NET 2.0+
- .NET Standard 1.3 and 1.6
- .NET Core</content>
  </entry>
  <entry>
    <id>https://localhost/packages/NUnit/3.5.0</id>
    <title type="text">NUnit: 3.9.0</title>
    <updated>2017-08-18T20:48:49+02:00</updated>
    <author>
      <name>Charlie Poole</name>
    </author>
    <link rel="alternate" href="https://localhost/packages/NUnit/3.5.0" />
    <content type="text">NUnit features a fluent assert syntax, parameterized, generic and theory tests and is user-extensible.

This package includes the NUnit 3 framework assembly, which is referenced by your tests. You will need to install version 3 of the nunit3-console program or a third-party runner that supports NUnit 3 in order to execute tests. Runners intended for use with NUnit 2.x will not run NUnit 3 tests correctly.

Supported platforms:
- .NET 2.0+
- .NET Standard 1.3 and 1.6
- .NET Core</content>
  </entry>
</feed>
```

The feed itself is valid according to the w3c validator: https://validator.w3.org/feed/#validate_by_input

Still open (beside the conflict with the current master): 

* Tests

Before I continue just one question: Is the implementation OK-ish from your perspective? I'm using a different approach now with the SyndicationFeed that's why I'm asking. The AtomFeedActionResult is based from this [blogpost](https://rehansaeed.com/building-rssatom-feeds-for-asp-net-mvc/)

The good thing is, that this could be tested quite good I think.

joelverhagen:
> Before I continue just one question: Is the implementation OK-ish from your perspective?

More than OK-ish 😄. Great!

I'll leave some specific code comments and wait on those tests, but this really looking good.

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -194,6 +194,15 @@ public static void RegisterUIRoutes(RouteCollection routes)
                 },
                 new { version = new VersionRouteConstraint() });
 
+            routes.MapRoute(
+                RouteName.DisplayPackageFeed,
+                "packages/{id}/atom.xml",
+                new
+                {
+                    controller = "packages",
+                    action = "AtomFeed"
```
joelverhagen: you can use `nameof` here.

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -194,6 +194,15 @@ public static void RegisterUIRoutes(RouteCollection routes)
                 },
                 new { version = new VersionRouteConstraint() });
 
+            routes.MapRoute(
+                RouteName.DisplayPackageFeed,
+                "packages/{id}/atom.xml",
+                new
+                {
+                    controller = "packages",
```
joelverhagen: nit: `Packages`

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1174,6 +1176,58 @@ public virtual ActionResult ManagePackageOwners(string id)
             return View(model);
         }
 
+        [HttpGet]
+        public virtual ActionResult AtomFeed(string id, bool allowPrerelease = false)
+        {
+            var package = _packageService.FindPackageByIdAndVersion(id, string.Empty, null, allowPrerelease);
```
joelverhagen: Use `FindPackageRegistrationById`

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1174,6 +1176,58 @@ public virtual ActionResult ManagePackageOwners(string id)
...
+            if (package == null)
+            {
+                return HttpNotFound();
+            }
+
+            SyndicationFeed feed = new SyndicationFeed()
+            {
+                Id = Url.Package(package.PackageRegistration.Id, version: null, relativeUrl: false),
+                Title = SyndicationContent.CreatePlaintextContent($"NuGet.org: {package.Id}"),
+                Description = SyndicationContent.CreatePlaintextContent(package.Summary),
```
joelverhagen: Any reason we chose `Summary` over `Description` here? I think `Description` is more common among packages.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1174,6 +1176,58 @@ public virtual ActionResult ManagePackageOwners(string id)
...
+                LastUpdatedTime = DateTimeOffset.Now,
+            };
+
+            if(!string.IsNullOrWhiteSpace(package.IconUrl))
+            {
+                feed.ImageUrl = new Uri(package.IconUrl);
+            }
+
+            var packageVersions  = package.PackageRegistration.Packages
+                                   .OrderByDescending(p => new NuGetVersion(p.Version), Comparer<NuGetVersion>.Create((a, b) => a.CompareTo(b)))
```
joelverhagen: Make sure to filter out unlisted packages.

Make sure to filter out packages where `PackageStatusKey != PackageStatus.Available`. These are packages that are either deleted, still validating (e.g. virus scanning), or packages that failed validation.

joelverhagen: Personally I would order by `Created` instead of SemVer 2.0.0. This will mean the most recently published package is always at the top. It seems like this is more idiomatic in the Atom/RSS space, but is just my opinion. If you think SemVer order is better, feel free to keep it.

joelverhagen: FWIW, the Atom spec says that order of the feed is not significant.
https://validator.w3.org/feed/docs/rfc4287.html#element.feed

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1174,6 +1176,58 @@ public virtual ActionResult ManagePackageOwners(string id)
...
+
+            List<SyndicationItem> feedItems = new List<SyndicationItem>();
+
+            foreach(var packageVersion in packageVersions)
+            {
+                SyndicationItem syndicationItem = new SyndicationItem($"{packageVersion.Title}: {package.Version}", 
+                                                                      package.Description,
+                                                                      new Uri(Url.Package(package.PackageRegistration.Id, version: packageVersion.Version, relativeUrl: false)));
+                syndicationItem.Id = Url.Package(package.PackageRegistration.Id, version: packageVersion.Version, relativeUrl: false);
+                syndicationItem.LastUpdatedTime = packageVersion.LastUpdated;
```
joelverhagen: Can we make `PublishDate` be the `Created` property?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1174,6 +1176,58 @@ public virtual ActionResult ManagePackageOwners(string id)
...
+                syndicationItem.Authors.Add(new SyndicationPerson() { Name = packageVersion.FlattenedAuthors });
```
joelverhagen: We could make `SyndicationPerson` one per owner where the `Name` is username and the URL is the URL to the profile page, e.g. https://www.nuget.org/profiles/charliepoole

joelverhagen: (there should be a helper for this URL)

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: And use the latest (by SemVer) version to opulate the description and image URL.

src/NuGetGallery/UrlHelperExtensions.cs
```diff
@@ -1052,11 +1052,45 @@ public static RouteUrlTemplate<IPackageVersionModel> ManagePackageOwnersTemplate
             return new RouteUrlTemplate<IPackageVersionModel>(linkGenerator, routesGenerator);
         }
 
+        public static RouteUrlTemplate<IPackageVersionModel> PackageAtomFeedTemplate(this UrlHelper url, bool relativeUrl = true)
```
joelverhagen: What is this used for?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -828,8 +828,14 @@ foreach (var owner in Model.Owners)
                              src="@Url.Absolute("~/Content/gallery/img/twitter.svg")"
                              @ViewHelpers.ImageFallback(Url.Absolute("~/Content/gallery/img/twitter-24x24.png")) />
                     </a>
+                    <a href="@Url.PackageAtomFeed(Model.Id)" target="_blank">
+                        <img width="24" height="24" alt="Use the Atom feed to subscribe to new versions of @Model.Id"
```
joelverhagen: This image says 24x24 but the PNG is 36x36. Can we make a 24x24 version so the browser doesn't have to resize?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -828,8 +828,14 @@ foreach (var owner in Model.Owners)
                              src="@Url.Absolute("~/Content/gallery/img/twitter.svg")"
                              @ViewHelpers.ImageFallback(Url.Absolute("~/Content/gallery/img/twitter-24x24.png")) />
                     </a>
+                    <a href="@Url.PackageAtomFeed(Model.Id)" target="_blank">
```
joelverhagen: Let's not do `target="_blank"` for this one since since it is an internal URL.

Also, can we add `data-track="atom-feed"`? This will make a Google Analytics event get emitted when someone clicks this so we can understand how many people are discovering the Atom feed through this UI.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -767,24 +767,24 @@
             <h2>Owners</h2>
```
joelverhagen: IIRC there is a way in the HTML `<head>` to advertise that there is an Atom feed for this page... maybe a `<link rel="...">` of some sort. Can you investigate whether this is appropriate/a common practice?

robertmuehsig: https://stackoverflow.com/questions/440391/how-to-tell-web-browsers-about-atom-feeds

src/NuGetGallery/UrlHelperExtensions.cs
robertmuehsig: Not used - "copy in action" ;)

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
joelverhagen: Great! What do you think about a title of "NuGet.org: {ID}"?

robertmuehsig: At least the Nuget.org shouldn't be hardcoded - maybe it should point to the Gallery.Brand config entry? So it would be "NuGet Gallery: {ID}" - maybe it would be wise to include something like "NuGet Gallery: {ID} Feed"?

src/NuGetGallery/Controllers/PackagesController.cs
robertmuehsig: I used the "owners" from the packageRegistration - this should be OK I think.

src/NuGetGallery/Controllers/PackagesController.cs
robertmuehsig: Done with the next update

src/NuGetGallery/Controllers/PackagesController.cs
robertmuehsig: Done with the next update

robertmuehsig:
Not sure if I use the GitHub review tool as it meant to be - short summary:
I think I addressed all requested changes. Still missings are the tests and the hardcoded title for the feed title. 
My suggestion for the feed title would be a combination of the Gallery.Brand + the package id + feed like "{Gallery.Brand} Feed for package {package.Id}".

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1174,6 +1176,71 @@ public virtual ActionResult ManagePackageOwners(string id)
...
+            var packageRegistration = _packageService.FindPackageRegistrationById(id);
+            if (packageRegistration == null)
+            {
+                return HttpNotFound();
+            }
+
+            // ToDo: Nuget.org text hardcoded
+            // ToDo: Tests!
+            var packageVersions  = packageRegistration.Packages
+                                   .Where(x => x.PackageStatusKey == PackageStatus.Available)
```
joelverhagen: `.Where(x => x.Listed)`

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1174,6 +1176,71 @@ public virtual ActionResult ManagePackageOwners(string id)
...
+                ownersAsAuthors.Add(new SyndicationPerson() { Name = packageOwner.Username, Uri = Url.User(packageOwner, relativeUrl: false) });
+            }
+
+            foreach(var packageVersion in packageVersions)
+            {
+                SyndicationItem syndicationItem = new SyndicationItem($"{packageVersion.Title}: {packageVersion.Version}",
+                                                                      packageVersion.Description,
+                                                                      new Uri(Url.Package(packageVersion.Id, version: packageVersion.Version, relativeUrl: false)));
+                syndicationItem.Id = Url.Package(packageVersion.PackageRegistration.Id, version: packageVersion.Version, relativeUrl: false);
+                syndicationItem.LastUpdatedTime = packageVersion.LastUpdated;
```
joelverhagen: `LastEdited` is better than `LastUpdated` -- since this indicates a significant metadata change. `LastUpdated` is more granular and less interesting.

src/NuGetGallery/Infrastructure/SyndicationAtomFeedResult.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// Executes the call to the ActionResult method and returns the created feed to the output response.
+        /// </summary>
+        /// <param name="context">The context in which the result is executed. The context information includes the 
+        /// controller, HTTP content, request context, and route data.</param>
+        public override void ExecuteResult(ControllerContext context)
+        {
+            context.HttpContext.Response.ContentType = "application/atom+xml";
+            Atom10FeedFormatter feedFormatter = new Atom10FeedFormatter(this.syndicationFeed);
+            XmlWriterSettings xmlWriterSettings = new XmlWriterSettings();
+            xmlWriterSettings.Encoding = Encoding.UTF8;
```
joelverhagen: Does this send down a BOM? Let's avoid that.

src/NuGetGallery/Infrastructure/SyndicationAtomFeedResult.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            if (HttpContext.Current.IsDebuggingEnabled)
+            {
+                // Indent the XML for easier viewing but only in Debug mode. In Release mode, everything is output on 
+                // one line for best performance.
+                xmlWriterSettings.Indent = true;
```
joelverhagen: 👍

joelverhagen:
> My suggestion for the feed title would be a combination of the Gallery.Brand + the package id + feed like "{Gallery.Brand} Feed for package {package.Id}".

Sounds good.

robertmuehsig:
Ok - unit tests are added

image

The happy path is "quite" simple. Just tell me if I should add some more tests or change anything else.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -17,44 +17,44 @@
...
-            new PackageManagerViewModel()
-            {
-                Id = "dotnet-cli",
-                Name = ".NET CLI",
-                CommandPrefix = "> ",
-                InstallPackageCommand = string.Format("dotnet tool install --global {0} --version {1}", Model.Id, Model.Version),
-                AlertLevel = AlertLevel.Info,
-                AlertMessage = string.Format("This package contains a <a href=\"{0}\">.NET Core Global Tool</a> you can call from the shell/command line.", "https://aka.ms/global-tools"),
-            }
+    new PackageManagerViewModel()
```
joelverhagen: Seems like there is a lot of unintended whitespace changes here -- perhaps our editor configurations are different?

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -827,6 +828,196 @@ private class TestIssue : ValidationIssue
...
+            : TestContainer
+        {
+            [Fact]
+            public async Task GivenANonNormalizedVersionIt302sToTheNormalizedVersion()
+            {
+                // Arrange
+                var controller = CreateController(GetConfigurationService());
+
+                // Act
+                var result = await controller.DisplayPackage("Foo", "01.01.01");
```
joelverhagen: Shouldn't this call `AtomFeed`? And do we even do this normalized redirect?

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -827,6 +828,196 @@ private class TestIssue : ValidationIssue
...
+                Assert.Equal(dateTimeNow, syndicationFeedItems[0].LastUpdatedTime);
+
+                Assert.Equal("https://localhost/?id=Foo&version=1.0.0", syndicationFeedItems[1].Id);
+                Assert.Equal("Foo: 1.0.0", syndicationFeedItems[1].Title.Text);
+                Assert.Equal("Test Package", (syndicationFeedItems[1].Content as System.ServiceModel.Syndication.TextSyndicationContent).Text);
+                Assert.Equal(dateTimeYesterDay, syndicationFeedItems[1].PublishDate);
+                Assert.Equal(dateTimeYesterDay, syndicationFeedItems[1].LastUpdatedTime);
+
+            }
+
```
joelverhagen: nit: extra whitespace

joelverhagen:
There are some merge conflict. Can you resolve them? If they start getting hairy, let me know and I can take care of them as part of merging your PR.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
robertmuehsig: Ah - sorry - not needed.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
robertmuehsig: I will try to resolve the conflicts and check why the whitespaces are changed.

robertmuehsig:
urgh... mhh... this was not supposed to happen. Wait...

joelverhagen:
@robertmuehsig i think that was because the PR was based off of `master` branch. We typically do all feature work in the `dev` branch. I changed the base branch of your PR to `dev`. Looks better aside from some merge commits. I would recommend rebased your changes off of latest `dev`.

joelverhagen:
Also, could you update the screenshots? I would like to show @anangaur to make sure he likes it.

robertmuehsig:
ok - I was in the middle of creating a new (clean) branch and I guess this is now the better option. Hope this is ok, but I wasn't sure if I made some dumb git moves or not - sorry.

The new PR (with screenshots and a sample feed) is here https://github.com/NuGet/NuGetGallery/pull/6891

joelverhagen:
Closed in lieu of https://github.com/NuGet/NuGetGallery/pull/6891.

PR #6749, opened 2018-12-10
Add paging to the Manage Packages page

scottbommarito:
With this PR, the packages on the Manage Packages page are now paged. Upon initial load, the page only loads the user's first twenty packages. Additional packages can be accessed by clicking on a dropdown.

Before loaded:
image

After loaded:
image

With no results (same as PROD):
image

We also, in the background, load pages before the user requests them.
The current logic is:
- on load, start loading all pages for the current owner filter in order, one at a time
- on load, start loading the first page for each possible owner filter
- when a new owner is selected, stop loading the pages for old owner filter and start loading the pages for the new owner filter

Unfortunately, this was not simple to implement. Additionally, I think we'd still also gain a lot, performance-wise, from paging the reserved namespaces and ownership requests as well.

A couple things to call out:
1. Don't forget to review `page-manage-packages.js`! This is where the majority of the work went in, and GitHub decided to hide it because there were so many changes.
2. The `ManagePackagesSerializable...ViewModel`s consist of code that was existing in `Packages.cshtml`. I moved it so that this code could be reused across endpoint, such as in the paged packages.
3.  I added a couple `.Include` calls to `PackageService` and `UserService` because I noticed they greatly increased performance locally with around 1100 packages.


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+                .Select(p => new ManagePackagesSerializablePackageViewModel(
+                    p,
+                    packageUrlTemplate,
+                    editUrlTemplate,
+                    manageOwnersUrlTemplate,
+                    deleteUrlTemplate,
+                    setRequiredSignerUrlTemplate,
+                    profileUrlTemplate));
+
+            return new
```
chenriksson: Consider avoiding anonymous projections / dynamic types in public APIs.

If it were only used for `JsonResult`, it's probably ok... but if we're using this for Packages() then it starts bleeding into the ViewModel.

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -409,6 +409,12 @@ public static void RegisterUIRoutes(RouteCollection routes)
                 "account/{action}",
                 new { controller = "Users", action = "Account" });
 
+            routes.MapRoute(
+                RouteName.ManagePackagesPackages,
+                "account/Packages/packages",
```
chenriksson: The "account/Packages/packages" is confusing. Could it be something like "account/PackagesPage"?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+            return View(model);
+        }
+
+        [HttpGet]
+        [UIAuthorize]
+        public virtual JsonResult PackagesPaged(bool listed, int page = 0, string username = null)
+        {
+            return Json(
+                GetPagedPackagesForManagePackages(listed, page, username),
+                JsonRequestBehavior.AllowGet);
```
chenriksson: Could inline, if you handle page 0 same as rest.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+        }
+
+        private dynamic GetPagedPackagesForManagePackages(bool listed, int page = 0, string username = null)
+        {
+            var currentUser = GetCurrentUser();
+            IEnumerable<Package> packages;
+            if (username == null)
+            {
+                packages = PackageService.FindPackagesByAnyMatchingOwner(
+                    currentUser, includeUnlisted: true, includeVersions: true);
```
chenriksson: Should `includeUnlisted` be `!listed`? Same for below.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+            }
+            else
+            {
+                var user = UserService.FindByUsername(username);
+                if (user != null)
+                {
+                    packages = PackageService.FindPackagesByOwner(
+                        user, includeUnlisted: true, includeVersions: true);
```
chenriksson: Should we confirm that `currentUser` has permissions to manage on behalf of `user`?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+                }
+                else
                 {
-                    Username = "All packages"
-                },
-                new ListPackageOwnerViewModel(currentUser)
-            }.Concat(currentUser.Organizations.Select(o => new ListPackageOwnerViewModel(o.Organization)));
+                    packages = Enumerable.Empty<Package>();
```
chenriksson: This could indicate a bug, or at least stale data... should we throw instead? Probably should match behavior of whatever we do if current user doesn't have permissions.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+            var downloadCount = filteredPackages.Sum(p => p.DownloadCount);
+
+            var packageUrlTemplate = Url.PackageRegistrationTemplate();
+            var editUrlTemplate = Url.EditPackageTemplate();
+            var manageOwnersUrlTemplate = Url.ManagePackageOwnersTemplate();
+            var deleteUrlTemplate = Url.DeletePackageTemplate();
+            var setRequiredSignerUrlTemplate = Url.SetRequiredSignerTemplate();
+            var profileUrlTemplate = Url.UserTemplate();
+            var pagedPackages = filteredPackages
+                .Skip(GalleryConstants.ManagePackagesPagingSize * page)
```
chenriksson: I'm assuming `filteredPackages` is no longer an `IQueryable`. Can we do the paging DB-side, for the best performance? I think that's what I was originally considering... the only blocker I remember was that we had custom logic to handle scenarios where `IsLatest` wasn't set correctly.

src/NuGetGallery/ViewModels/ManagePackagesSerializableViewModel.cs
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using NuGet.Services.Entities;
+using NuGetGallery.Helpers;
+
+namespace NuGetGallery
+{
+    public class ManagePackagesSerializableViewModel
```
chenriksson: Looks like you're introducing a naming pattern to differentiate "light-weight" view models that we use in JS, from the heavier ones that we only use from razor?

Should we formalize a pattern across all pages, or could we just move towards light-weight view models altogether?

src/NuGetGallery/ViewModels/ManagePackagesViewModel.cs
```diff
@@ -10,11 +10,11 @@ public class ManagePackagesViewModel
...
         public virtual User User { get; set; }
 
-        public IEnumerable<ListPackageOwnerViewModel> Owners { get; set; }
+        public IEnumerable<string> Owners { get; set; }
 
-        public IEnumerable<ListPackageItemRequiredSignerViewModel> ListedPackages { get; set; }
+        public dynamic ListedPackages { get; set; }
 
-        public IEnumerable<ListPackageItemRequiredSignerViewModel> UnlistedPackages { get; set; }
+        public dynamic UnlistedPackages { get; set; }
```
chenriksson: Per earlier comment, would avoid letting `dynamic` types bleed into the view models.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+            };
 
-            var owners = new List<ListPackageOwnerViewModel> {
-                new ListPackageOwnerViewModel
+            return View(model);
+        }
+
+        [HttpGet]
+        [UIAuthorize]
+        public virtual JsonResult PackagesPaged(bool listed, int page = 0, string username = null)
```
chenriksson: Tests for the new API?

src/NuGetGallery/Controllers/UsersController.cs
chenriksson: I may be thinking of a different perf issue... I guess we still need to order by version from client?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
         public virtual ActionResult Packages()
         {
             var currentUser = GetCurrentUser();
+            
+            var model = new ManagePackagesViewModel
+            {
+                User = currentUser,
+                Owners = GetOwnersForUser(currentUser),
+                ListedPackages = GetPagedPackagesForManagePackages(true),
+                UnlistedPackages = GetPagedPackagesForManagePackages(false),
```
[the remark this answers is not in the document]
scottbommarito: I did this initially but experience was really bad. The first set of results wouldn't load for a couple seconds, and the page looks awkward without any data.

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: Yeah I can make a class to contain this structure.

src/NuGetGallery/App_Start/Routes.cs
scottbommarito: I wasn't sure what to put for this. `account/packages` is the manage packages page, and ideally I'd like to make this URL under that one. My other thought was `account/packages/page` but I decided against that in case we want to add paging to owner requests and reserved namespaces. Maybe `account/packages/page/packages`?

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: We have to do the filtering below regardless of what we do up here. I'm concerned that filtering here would cause us to do an extra `Where`, which is less performant.

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: Good catch. I think this page could leak `RequiredSigner` information otherwise.

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: In general, I decided not to implement any stale data handling for the paging because it would complicate things substantially and provide little benefit. Not only is a users' list of packages and owners unlikely to change, but there isn't any sort of "e-tag" or other identifier for the client to determine that its cache is out of date.

Technically, I could make this return a `400` or `404`, but then I would have to add handling for this case on the client-side. By instead returning an empty list, I don't have to do any additional work--the client side already handles empty lists. I think, with this logic, it may be a problem if a user loads the first page of results for a user, but then that user is deleted..but again, this is extraordinarily unlikely.

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: Looking at the code for `FindPackagesByOwner` and `FindPackagesByAnyMatchingOwner`,, it looks like we call `ToList` at the end of a query that otherwise appears to be on the SQL side. When I wrote this I intended it to be on the SQL side. Do you think I should add paging to this function? Locally, this SQL call didn't seem to be the server-side bottleneck--the bigger issue was that we have to make requests for every owner request and reserved namespace.

src/NuGetGallery/ViewModels/ManagePackagesSerializableViewModel.cs
scottbommarito: Well, I think the situation on this page is unlike ones that we have elsewhere. We have a `ViewModel` that's passed to the view **in addition** to a model that's passed to the `js`. Normally, I would have just named this class `ManagePackagesViewModel`, but we already have one of those. I think we can think about formalizing a pattern when we have this situation on multiple pages.

src/NuGetGallery/Controllers/UsersController.cs
chenriksson: We should understand what's being optimized here. So far, I believe this just reduces the amount of data serialized on initial page load, and the amount of items managed with ko events. It likely adds more work on the server and DB, because we're now doing these queries per page with ajax requests... which could be problematic.

I remember the packages query being challenging because the semantic versioning couldn't be ordered in the database. I *think* we can still optimize this, by limiting the first `SELECT` to just the `[Key]` and [`Version`] columns for the sole purpose of sorting and paging. Then, we could do the full EF query for only the current page... which should save on I/O from the DB, and maybe the join operations.

The owner and namespace queries could potentially be optimized as well... but we really should profile first to understand where the biggest bottlenecks are. We can discuss more when you get in, and I'm happy to help.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
         public virtual ActionResult Packages()
         {
             var currentUser = GetCurrentUser();
+            
+            var model = new ManagePackagesViewModel
+            {
+                User = currentUser,
+                Owners = GetOwnersForUser(currentUser),
+                ListedPackages = GetPagedPackagesForManagePackages(true),
```
joelverhagen: >GetPagedPackagesForManagePackages [](start = 33, length = 33)

Can we explicitly state page 0 here? that's an imporant detail for understanding the flow here

src/NuGetGallery/Controllers/UsersController.cs
joelverhagen: >packages [](start = 20, length = 8)

When would this case occur? Why not return an error code?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+
+        private IEnumerable<string> GetOwnersForUser(User user)
+        {
+            return new List<string> {
+                GalleryConstants.ManagePackagesAllPackagesFilter,
+                user.Username,
+            }.Concat(user.Organizations.Select(o => o.Organization.Username));
+        }
+
+        private IEnumerable<ListPackageItemRequiredSignerViewModel> FilterPackagesForManagePackages(IEnumerable<Package> packages, User currentUser, bool listed)
```
joelverhagen: >FilterPackagesForManagePackages [](start = 68, length = 31)

What's the expected behavior for deleted/validating/failed validation packages?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+                .Where(p => (p.Listed && p.PackageStatusKey == PackageStatus.Available) == listed)
                 .Select(p => new ListPackageItemRequiredSignerViewModel(p, currentUser, SecurityPolicyService, wasAADLoginOrMultiFactorAuthenticated))
-                .OrderBy(p => p.Id)
-                .ToList();
-            var unlistedPackages = packages
-                .Where(p => !p.Listed || p.PackageStatusKey != PackageStatus.Available)
-                .Select(p => new ListPackageItemRequiredSignerViewModel(p, currentUser, SecurityPolicyService, wasAADLoginOrMultiFactorAuthenticated))
-                .OrderBy(p => p.Id)
-                .ToList();
+                .OrderBy(p => p.Id);
```
joelverhagen: >OrderBy [](start = 17, length = 7)

Is this order-by done on server side?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+        [HttpGet]
+        [UIAuthorize]
+        public virtual JsonResult PackagesPaged(bool listed, int page = 0, string username = null)
+        {
+            return Json(
+                GetPagedPackagesForManagePackages(listed, page, username),
+                JsonRequestBehavior.AllowGet);
+        }
+
+        private dynamic GetPagedPackagesForManagePackages(bool listed, int page = 0, string username = null)
```
joelverhagen: >GetPagedPackagesForManagePackages [](start = 24, length = 33)

Please double check the number of SQL queries performed by this method. Because `FilterPackagesForManagePackages` returns an `IEnumerable` it seems like 3 queries are done each invocation of this method.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -462,62 +463,128 @@ public virtual ActionResult Thanks()
...
+
+            return new
+            {
+                totalCount = packageCount,
+                totalDownloadCount = downloadCount,
+                packages = pagedPackages
+            };
+        }
+
+        private IEnumerable<string> GetOwnersForUser(User user)
```
joelverhagen: The fact that a non username is in here suggests to me this shouldn't be a helper method. This method pretty much only makes sense in one context (that drop-down).

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -1,13 +1,11 @@
...
     'use strict';
 
-    function showInitialPackagesData(dataSelector, packagesList) {
-        var downloadsCount = 0;
-        $.each(packagesList, function () { downloadsCount += this.TotalDownloadCount; });
-        $(dataSelector).text(formatPackagesData(packagesList.length, downloadsCount));
-    }
-
     function formatPackagesData(packagesCount, downloadsCount) {
+        if (packagesCount === null || downloadsCount === null) {
```
joelverhagen: When would these values be null?

src/NuGetGallery/ViewModels/ManagePackagesSerializableViewModel.cs
joelverhagen: >ManagePackagesSerializableViewModel [](start = 17, length = 35)

What is this "serializable" naming pattern mean?

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -170,42 +154,222 @@
...
+
+            this.CreatePackagePageIdentity = function (ownerFilter, page) {
+                return {
+                    ownerFilter: ownerFilter,
+                    page: page
+                };
+            };
+
+            this.DefaultPage = 0;
+            this.PackagePageRange = 10;
```
joelverhagen: >PackagePageRange [](start = 17, length = 16)

unused?

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -170,42 +154,222 @@
...
+            
+            this.PackagesCache = {};
+            this.PackagesCache[self.GetPackagePageIdentityCacheKey(self.PackagePageIdentity())] = ko.observable(initialPackages);
+            this.IsCachingPage = {};
+            this.GetPackagePage = function (identity, callback) {
+                var packageCacheKey = self.GetPackagePageIdentityCacheKey(identity);
+                var cachedPage = self.PackagesCache[packageCacheKey];
+                if (cachedPage && cachedPage()) {
+                    // The page has already been loaded.
+                    callback && callback();
```
joelverhagen: >callback [](start = 20, length = 8)

When will the callback be null?

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -170,42 +154,222 @@
...
+                    return;
+                }
+                
+                var ownerFilter = identity.ownerFilter;
+                var page = identity.page;
+                var isCaching = self.IsCachingPage[packageCacheKey];
+                if (isCaching) {
+                    // This page is already being loaded.
+                    // Don't try to load it again while it's already being loaded.
+                    callback && callback();
```
joelverhagen: >callback [](start = 32, length = 8)

What is the caller's expectation when the callback is called? It seems like sometimes the page is loaded and sometimes it's not

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -170,42 +154,222 @@
...
+                $.ajax({
+                    url: getPagedPackagesUrl + '?page=' + page + '&listed=' + listed + (ownerFilter === allPackagesFilter ? '' : '&username=' + ownerFilter),
+                    dataType: 'json',
+                    success: function (data) {
+                        self.PackagesCache[packageCacheKey](data);
+                        self.IsCachingPage[packageCacheKey] = false;
+
+                        callback && callback();
+                    },
+                    error: function () {
```
joelverhagen: >function [](start = 27, length = 8)

So this retries forever? There is not indication to the user that there is a problem?

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -310,172 +322,16 @@
...
-        };
-    }
-
-    dynamic GetSerializableSigners(IEnumerable<SignerViewModel> signers)
-    {
-        return signers.Select(signer => GetSerializableSigner(signer));
-    }
+@section BottomScripts {
+    <script type="text/javascript">
+        var allPackagesFilter = "@Html.Raw(GalleryConstants.ManagePackagesAllPackagesFilter)";
```
joelverhagen: >Raw [](start = 39, length = 3)

Couldn't you just call `Json.Encode` on the string?

joelverhagen:
How do we know that doing paged AJAX is better than just getting all the data on initial page load but only showing a portion? How did we pick 20 as the query page size (20 for the "visible" page size seems reasonable to me).

For owners with many packages, this will likely be a experience degradation because they have to guess which page their desired page is on. Previously, all they had to do is go to the manage packages page then <kbd>Ctrl + F</kbd> for the ID that they want to go to.

Imagine having 5 pages of packages and not knowing which page the ID lands on... what does the user do? Try page 1, then 2, then ... up to page 5? Or do they just go up to the URL bar on the package details page and type their ID in? Or do they use search?

To me, it seems like if we introduce pages, we also need filtering.

Let's talk offline.

joelverhagen:
Oh, and if we really want the paging, we should make sure browser Back button and ideally Refresh retains the currently selected page and owner.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
scottbommarito: When the page loads, before the data is loaded into the model, I set these to null.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
scottbommarito: Oh yeah, this was used before until I reworked the UI, will change.

src/NuGetGallery/Views/Users/Packages.cshtml
scottbommarito: Makes sense. Will make sure it works.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
scottbommarito: I wasn't sure what to do for retries, so this is more of a placeholder. We can discuss what to do offline.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
scottbommarito: Well, to clarify, the purpose of the callback is to prevent making too many requests at once. For each function call (`PreloadPagesForOwnerFilter` and `PreloadPagesForAllOwners`), we only do one at a time--page 0, then page 1, then page 2--and the way this is done is by creating a callback that triggers loading the next page. There are 4 function calls doing this loading at once, so we make four requests at once, and I thought this amount was ideal. I suppose I could explore a different strategy for this (perhaps kicking them off 1 second apart instead of using a callback) but this was my initial idea.

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: Makes sense.

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: I answered this question above for @chenriksson, but to reiterate

This can occur when:
1. there is a bug in the page that causes it to request an invalid user
2. someone hacks the page to request a user that doesn't exist
3. an organization that the user was a member of was deleted after the page loaded

After thinking about it again, I think returning an error code might be useful so that the user can tell the page is having issues.

scottbommarito: Though about it again--in cases that the page has a bug, I think we should return an error code so we should display to users that the page is having issues. I think the stale data problems are less of a concern though.

scottbommarito:
Closing because I determined that adding an `Include` to `PackageService` provided a massive performance improvement. I don't think paging is necessary.

PR #6752, opened 2018-12-12
Server-side performance fixes for Manage Packages page

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6755

These fixes drastically improve the performance of the Manage Packages page.

1) Include `RequiredSigners` in the initial request for packages. Previously, the page was making a SQL request for every package in `ListPackageItemRequiredSignerViewModel`.

2) Exit the `ChecksPermissionsOnBehalfOfAnyAccount` call early. `ChecksPermissionsOnBehalfOfAnyAccount` has two implementations-- both return whether or not the action can be done on behalf of any owner, but one returns the list of owners the action can be done on behalf as well. To get the full list of owners the action can be done on behalf of, we were enumerating the full list of possible owners every time, even if we didn't care about which owners the action can be done on behalf of. By exiting the loop early when we don't care about the list, we drastically reduce the number of permissions checks that are done on accounts with many packages or packages with many co-owners.

Tested in against NuGetTestAccount on dev, **the page won't even load without the fixes**. NuGetTestAccount has around 30,000 packages, which is 6 times the number of packages owned by the `Microsoft` account.

With the fixes, I got page load times of:

image

### DIFF
--- a/src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
+++ b/src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
@@ -55,25 +55,31 @@ namespace NuGetGallery
 
         public PermissionsCheckResult CheckPermissionsOnBehalfOfAnyAccount(User currentUser, TEntity entity)
         {
-            return CheckPermissionsOnBehalfOfAnyAccount(currentUser, entity, out var accountsAllowedOnBehalfOf);
+            return CheckPermissionsOnBehalfOfAnyAccount(currentUser, entity, exitEarly: true, accountsAllowedOnBehalfOf: out var _);
         }
 
         public PermissionsCheckResult CheckPermissionsOnBehalfOfAnyAccount(User currentUser, TEntity entity, out IEnumerable<User> accountsAllowedOnBehalfOf)
         {
+            return CheckPermissionsOnBehalfOfAnyAccount(currentUser, entity, exitEarly: false, accountsAllowedOnBehalfOf: out accountsAllowedOnBehalfOf);
+        }
+
+        private PermissionsCheckResult CheckPermissionsOnBehalfOfAnyAccount(User currentUser, TEntity entity, bool exitEarly, out IEnumerable<User> accountsAllowedOnBehalfOf)
+        {
             accountsAllowedOnBehalfOf = new List<User>();
 
-            var possibleAccountsOnBehalfOf =
-                new[] { currentUser }
-                    .Concat(GetOwners(entity));
+            var possibleAccountsOnBehalfOf = new List<User>
+            {
+                currentUser
+            };
+
+            possibleAccountsOnBehalfOf.AddRange(GetOwners(entity));
 
             if (currentUser != null)
             {
-                possibleAccountsOnBehalfOf =
-                    possibleAccountsOnBehalfOf
-                        .Concat(currentUser.Organizations.Select(o => o.Organization));
+                possibleAccountsOnBehalfOf.AddRange(currentUser.Organizations.Select(o => o.Organization));
             }
 
-            possibleAccountsOnBehalfOf = possibleAccountsOnBehalfOf.Distinct(new UserEqualityComparer());
+            possibleAccountsOnBehalfOf = possibleAccountsOnBehalfOf.Distinct(new UserEqualityComparer()).ToList();
 
             var aggregateResult = PermissionsCheckResult.Unknown;
 
@@ -84,6 +90,10 @@ namespace NuGetGallery
                 if (result == PermissionsCheckResult.Allowed)
                 {
                     (accountsAllowedOnBehalfOf as List<User>).Add(accountOnBehalfOf);
+                    if (exitEarly)
+                    {
+                        break;
+                    }
                 }
             }
 
--- a/src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
+++ b/src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
@@ -79,11 +79,9 @@ namespace NuGetGallery
                 possibleAccountsOnBehalfOf.AddRange(currentUser.Organizations.Select(o => o.Organization));
             }
 
-            possibleAccountsOnBehalfOf = possibleAccountsOnBehalfOf.Distinct(new UserEqualityComparer());
-
             var aggregateResult = PermissionsCheckResult.Unknown;
 
-            foreach (var accountOnBehalfOf in possibleAccountsOnBehalfOf)
+            foreach (var accountOnBehalfOf in possibleAccountsOnBehalfOf.Distinct(new UserEqualityComparer()))
             {
                 var result = CheckPermissions(currentUser, accountOnBehalfOf, entity);
                 aggregateResult = ChoosePermissionsCheckResult(aggregateResult, result);
--- a/src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
+++ b/src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
@@ -79,7 +79,7 @@ namespace NuGetGallery
                 possibleAccountsOnBehalfOf.AddRange(currentUser.Organizations.Select(o => o.Organization));
             }
 
-            possibleAccountsOnBehalfOf = possibleAccountsOnBehalfOf.Distinct(new UserEqualityComparer()).ToList();
+            possibleAccountsOnBehalfOf = possibleAccountsOnBehalfOf.Distinct(new UserEqualityComparer());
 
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -473,16 +473,17 @@ public virtual ActionResult Packages()
 
             var wasAADLoginOrMultiFactorAuthenticated = User.WasMultiFactorAuthenticated() || User.WasAzureActiveDirectoryAccountUsedForSignin();
 
-            var packages = PackageService.FindPackagesByAnyMatchingOwner(currentUser, includeUnlisted: true);
-            var listedPackages = packages
-                .Where(p => p.Listed && p.PackageStatusKey == PackageStatus.Available)
+            var packages = PackageService.FindPackagesByAnyMatchingOwner(currentUser, includeUnlisted: true)
```
joelverhagen: Are these changes necessary? If not, exclude them.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -55,25 +55,31 @@ public PermissionsCheckResult CheckPermissions(IPrincipal currentPrincipal, User
...
+        }
+
+        private PermissionsCheckResult CheckPermissionsOnBehalfOfAnyAccount(User currentUser, TEntity entity, bool exitEarly, out IEnumerable<User> accountsAllowedOnBehalfOf)
         {
             accountsAllowedOnBehalfOf = new List<User>();
 
-            var possibleAccountsOnBehalfOf =
-                new[] { currentUser }
-                    .Concat(GetOwners(entity));
+            var possibleAccountsOnBehalfOf = new List<User>
```
joelverhagen: Does this have unit tests so we are sure this isn't broken?

joelverhagen:
Let's get a graph of page load durations for 10-20 iterations before and after.

Let's run against a real DB, like DEV DB.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -473,16 +473,17 @@ public virtual ActionResult Packages()
...
                 .Select(p => new ListPackageItemRequiredSignerViewModel(p, currentUser, SecurityPolicyService, wasAADLoginOrMultiFactorAuthenticated))
-                .OrderBy(p => p.Id)
```
cristinamanum: I think that this change will move the orderby  execution from the sql server side to the client side. Isn't it slower?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -473,16 +473,17 @@ public virtual ActionResult Packages()
...
+
+            var listedPackages = packages
+                .Where(p => p.Listed && p.Available)
                 .ToList();
+
             var unlistedPackages = packages
-                .Where(p => !p.Listed || p.PackageStatusKey != PackageStatus.Available)
-                .Select(p => new ListPackageItemRequiredSignerViewModel(p, currentUser, SecurityPolicyService, wasAADLoginOrMultiFactorAuthenticated))
-                .OrderBy(p => p.Id)
+                .Where(p => !p.Listed || !p.Available)
```
cristinamanum: Order is not necessary anymore?

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: Will exclude.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: There's a lot of tests that test this, because it's a static method and can't be mocked.

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: Reverting because it isn't a part of the perf change.

The answer is no, though, because the `PackageService` method already calls `ToList` and makes the SQL query. Everything in this controller is client-side.

src/NuGetGallery/Controllers/UsersController.cs
scottbommarito: Reverting because it isn't a part of the perf change.

My thoughts were that the `OrderBy` on the `packages` variable above would keep the list in order after this `Where`. I'm not sure if it's guaranteed though.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -55,25 +55,31 @@ public PermissionsCheckResult CheckPermissions(IPrincipal currentPrincipal, User
...
+            {
+                currentUser
+            };
+
+            possibleAccountsOnBehalfOf.AddRange(GetOwners(entity));
 
             if (currentUser != null)
```
chenriksson: I realized this didn't change... but if `currentUser` is null, should it even be in the enumerable above?

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
```diff
@@ -55,25 +55,31 @@ public PermissionsCheckResult CheckPermissions(IPrincipal currentPrincipal, User
...
             {
-                possibleAccountsOnBehalfOf =
-                    possibleAccountsOnBehalfOf
-                        .Concat(currentUser.Organizations.Select(o => o.Organization));
+                possibleAccountsOnBehalfOf.AddRange(currentUser.Organizations.Select(o => o.Organization));
             }
 
-            possibleAccountsOnBehalfOf = possibleAccountsOnBehalfOf.Distinct(new UserEqualityComparer());
+            possibleAccountsOnBehalfOf = possibleAccountsOnBehalfOf.Distinct(new UserEqualityComparer()).ToList();
```
chenriksson: Is the `ToList` necessary? Did `possibleAccountsOnBehalfOf.Distinct` somehow cause a query execution (it looks like it's already in memory).

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -255,30 +255,28 @@ private IEnumerable<Package> GetPackagesForOwners(IEnumerable<int> ownerKeys, bo
...
-                return packages
-                .Include(p => p.PackageRegistration)
-                .Include(p => p.PackageRegistration.Owners)
-                .ToList();
-            }
-
-            // Do a best effort of retrieving the latest version. Note that UpdateIsLatest has had concurrency issues
-            // where sometimes packages no rows with IsLatest set. In this case, we'll just select the last inserted
-            // row (descending [Key]) as opposed to reading all rows into memory and sorting on NuGetVersion.
+            if (!includeVersions)
```
chenriksson: Would keep the `if (included)` block here... it's usually preferred to start with the positive condition. Makes the change smaller too.

scottbommarito: The primary reason I made this change was to guarantee that the list of `Include` calls for both `includeVersions == true` and `includeVersions == false` is the same. I don't want it to be possible that one has `Include(p => p.RequiredSigners)` and one does not.

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: I think we were playing around with the data types of `possibleAccountsOnBehalfOf` (e.g. considering `HashSet`) and needed the `ToList` when it wasn't a `List`. Will remove.

chenriksson:
Agree with @joelverhagen. I'd like to see average/percentile of before and after results, preferably quantified with the %diff (after - before / before * 100) calculation.

I also like to document the investigation by sharing a screenshot of the profile showing the inclusive %CPU before optimization. Of course, you also need to define the scenario that's being tested (e.g., lots of packages, lots of packages with lots of versions, etc).

For instance: https://github.com/NuGet/Engineering/issues/1453

src/NuGetGallery/Services/ActionRequiringEntityPermissions.cs
scottbommarito: Yes. `null` current user means a user that is not logged in. There are actions that users can do without being logged in. We need to make sure the `PermissionsHelper` call is made with the `null` current user so we can test the permissions for a logged out user.

scottbommarito:
Unfortunately, it's not very easy for me to test an account that loads the page before the change locally against a SQL DB (due to AAD login), so I plan to gather data after this is deployed to DEV.

chenriksson:
@scottbommarito You can test with AAD locally. In OneNote "Microsoft Logins (MSA) and NuGet Gallery", search for "Test Accounts for local usage". It would probably be nice if we stored this in the DEV AKV instead.

PR #6758, opened 2018-12-14
Add OpenReadAsync to ISimpleCloudBlob wrapper

joelverhagen:
This will allow me to use `If-None-Match: "old-etag"` in the search service to only download auxiliary files when they have changed.

I am using this storage abstraction because it has etags exposed.

### DIFF
--- a/tests/NuGetGallery.Core.Facts/Services/CloudBlobCoreFileStorageServiceIntegrationTests.cs
+++ b/tests/NuGetGallery.Core.Facts/Services/CloudBlobCoreFileStorageServiceIntegrationTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.IO;
+using System.Net;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


tests/NuGetGallery.Core.Facts/Services/CloudBlobCoreFileStorageServiceIntegrationTests.cs
```diff
@@ -58,6 +59,129 @@ public CloudBlobCoreFileStorageServiceIntegrationTests(BlobStorageFixture fixtur
...
+            await file.FetchAttributesAsync();
+
+            // Act & Assert
+            var ex = await Assert.ThrowsAsync<StorageException>(
+                () => file.OpenReadAsync(accessCondition: AccessCondition.GenerateIfMatchCondition("WON'T MATCH")));
+            Assert.Equal(HttpStatusCode.PreconditionFailed, (HttpStatusCode)ex.RequestInformation.HttpStatusCode);
+        }
+
+        [BlobStorageFact]
+        public async Task OpenReadAsyncThrowsNotModifiedWhenIfNoneMatchFails()
```
joelverhagen: This is the behavior I will be using.

Issue #6721, opened 2018-12-04
[Licenses 1/15] Display License expression on package details page


agr:
Instead of a blind link to a "license info" when package has the license expression specified, package details page should display that license expression with links to individual licenses pointing to licenses.nuget.org (see https://licenses.nuget.org/MIT%20OR%20Apache-2.0 for a basic example of how license expression should be presented).

The code that generates the presentation is in NuGet.Licenses, we just need to figure out a way to share it between two projects.

PR #6784, opened 2019-01-04
License expression display in Gallery

agr:
Addresses NuGet/NuGetGallery#6721

Will update PR once https://github.com/NuGet/ServerCommon/pull/234 is merged and non-private builds of a new library are available.

image

Worse case:

image

Should we put `<nobr>` (or set up css accordingly) around license names, so they are not split between lines?

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -16,6 +16,7 @@ using System.Web.Caching;
 using System.Web.Mvc;
 using NuGet.Packaging;
 using NuGet.Services.Entities;
+using NuGet.Services.Licenses;
 using NuGet.Services.Messaging.Email;
 using NuGet.Versioning;
 using NuGetGallery.Areas.Admin;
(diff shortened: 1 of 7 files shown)


joelverhagen:
Screenshots please

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -146,6 +155,10 @@
     </div>
 }
 
+@* The following two helpers must be on a single line each so no extra whitespace is introduced in the expression when rendered. *@
```
scottbommarito: These are only used in one place each and look pretty simple. Maybe consider just inlining them?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -624,6 +628,21 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
 
             model.ReadMeHtml = await _readMeService.GetReadMeHtmlAsync(package);
 
+            if (!string.IsNullOrWhiteSpace(package.LicenseExpression))
```
scottbommarito: Add tests that verify that the license expression is split if it exists?

agr: Added

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
scottbommarito: Sounds good. Maybe add that comment in there?

Issue #6772, opened 2018-12-19
Consolidate "Manage package" page


xavierdecoster:

scottbommarito:
After consolidating, it seems both the `Edit` (ReadMe) and `Delete` (listing) pages rely on page reload to switch versions. Nontrivial work will need to be done to fix this because otherwise the page will function very poorly.

PR #6805, opened 2019-01-09
Consolidate Edit, Delete, and Manage Owners into a single page

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6772

This PR combines the edit, delete, and manage owners page for packages into a single page now referred to as the "manage" page.

Most of the work was fairly trivial (make the old pages partials and move them into a single page) but I did have to make some nontrivial changes to the `Delete` and `Edit` pages, which previously reloaded the page whenever a different version was selected.

---

**Full page:**
image

---

**Owners section:**
image

---

**Listing section:**
image

---

**Documentation section:**
image


src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -609,7 +604,7 @@ public static void RegisterUIRoutes(RouteCollection routes)
                     new { controller = "Packages", action = "ContactOwners" },
                     // This next bit looks bad, but it's not. It will never change because
                     // it's mapping the legacy routes to the new better routes.
-                    new { action = "ContactOwners|ManagePackageOwners" }),
+                    new { action = "ContactOwners|Manage" }),
```
chenriksson: nit: Are legacy routes still used? Are we supposed to support legacy routes and redirect?
This link may be a little more confusing... unless it's possible to link to page w/ Owners section expanded?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1194,53 +1194,30 @@ public virtual async Task<ActionResult> ContactOwners(string id, string version,
...
-        [RequiresAccountConfirmation("delete a package")]
-        public virtual ActionResult Delete(string id, string version)
-        {
-            var package = _packageService.FindPackageByIdAndVersion(id, version);
-            if (package == null)
-            {
-                return HttpNotFound();
-            }
-            if (ActionsRequiringPermissions.UnlistOrRelistPackage.CheckPermissionsOnBehalfOfAnyAccount(GetCurrentUser(), package) != PermissionsCheckResult.Allowed)
+            if (!model.CanEdit && !model.CanManageOwners && !model.CanUnlistOrRelist)
```
chenriksson: Are these identical? If so, can we consolidate the permissions as well (e.g., `model.CanManage`)?

If not identical, it seems strange to make the whole page forbidden as opposed to disabling the relevant section.

chenriksson: ~~Also, if `CanX` are server-side only permissions checks... why are they on the view model?~~ Oh, just realized it's a shared view model with the display page... but isn't really relevant for manage page anymore.

src/Bootstrap/less/theme/page-delete-package.less
```diff
@@ -1,6 +1,8 @@
 .page-delete-package {
```
chenriksson: If we're consolidating the cshtml, can we also do the less and js files? Especially, since they're based on the page names.

src/NuGetGallery/UrlHelperExtensions.cs
```diff
@@ -618,19 +618,57 @@ public static string GetOrganizationCertificates(this UrlHelper url, string acco
...
+            return GetRouteLink(
+                url,
+                RouteName.PackageAction,
+                relativeUrl,
+                routeValues: new RouteValueDictionary
+                {
+                        { "action", action },
+                        { "id", id }
+                });
+        }
```
chenriksson: With consolidation, is there only one "PackageAction" route (Manage)? If so, looks like some of the code could be simplified.  Assume same would apply to "PackageVersionAction"?

src/NuGetGallery/UrlHelperExtensions.cs
```diff
@@ -644,53 +682,85 @@ public static string GetOrganizationCertificates(this UrlHelper url, string acco
...
         }
 
-        public static string ReflowPackage(
+        /// <summary>
+        /// Initializes a manage package link that can be resolved at a later time.
+        /// 
+        /// Callers should only use this API if they need to generate many links, such as the ManagePackages view
+        /// does. This template reduces the calls to RouteCollection.GetVirtualPath which can be expensive. Callers
+        /// that only need a single link should call Url.ManagePackage instead.
+        public static RouteUrlTemplate<IPackageVersionModel> ManagePackageTemplate(
```
chenriksson: `ManagePackageVersionTemplate`?

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Web.Mvc;
+using NuGet.Services.Entities;
+using NuGet.Versioning;
+
+namespace NuGetGallery
+{
+    public class ManagePackageViewModel : DisplayPackageViewModel
```
chenriksson: I remember there being a lot of complexity due to many layers of inheritance between different view models. Would it make sense to completely separate view models between pages?

src/NuGetGallery/ViewModels/TrivialPackageVersionModel.cs
```diff
@@ -4,8 +4,13 @@ namespace NuGetGallery
 {
     public class TrivialPackageVersionModel : IPackageVersionModel
     {
+        public TrivialPackageVersionModel(string id, string version)
```
chenriksson: No dependencies on the old, default constructor?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -698,25 +698,13 @@
                     </li>
                 }
 
-                @if ((Model.Available || Model.Validating) && Model.CanEdit)
+                @if (((Model.Available || Model.Validating) && Model.CanEdit)
+                    || Model.CanManageOwners
+                    || (!Model.Deleted && Model.CanUnlistOrRelist))
```
chenriksson: Again, can we consolidate permissions - given there's a single link now?

skofman1:
@anangaur , can we remove the "Editing package metadata is not supported.." message. It's been a year..

anangaur:
If we say “Manage” everywhere then yes edit info message may not be required. I just wanted it to be there not as a warning but info message. But I don’t mind it not being there at all.

src/NuGetGallery/App_Start/Routes.cs
scottbommarito: I looked at AI and I don't see any successful requests with `Packages/ContactOwners/{id}` or `Packages/ManagePackageOwners/{id}` in the last week. I think this is safe to remove.

src/NuGetGallery/ViewModels/TrivialPackageVersionModel.cs
scottbommarito: Nope. I'm not sure if the class was even used at all.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: To clarify, this conditional only returns `Forbidden` if the user can't perform ANY of the required permissions.

e.g. "if you can't edit and you can't manage owners and you can't unlist or relist, return forbidden"

When you don't have permission to do any individual thing, that section of the page is hidden for you as you expected.

We shouldn't consolidate the permissions because
- managing owners has different permissions than editing and unlisting/relisting
- it's a large change
- although the manage page is a single link, the forms are all different routes
- it's a reductive change, and it would create more work for us if we want to regain the flexibility that we already have

src/Bootstrap/less/theme/page-delete-package.less
scottbommarito: The JS files are big and complex and I think consolidating them would make the code harder to read because you would have to search for the individual sections of the page inside a giant file.

The less files tho are probably ok to combine, but I also think it's nice to have separate files for these things instead of a massive one. Additionally, I believe `page-delete-package.less` is used by an admin page as well, so I would prefer not to have to worry about breaking things like that.

src/NuGetGallery/UrlHelperExtensions.cs
scottbommarito: Any action that uses `FindPackageByIdAndVersion` and not `FindPackageByIdAndVersionStrict` should be a `PackageAction` route. This includes `Manage`, `DeleteSymbols`, `Reflow`, `Revalidate`, and `RevalidateSymbols`. All of these operate on the latest package version when version is not included. I didn't realize this before, so I will go update the tests to test these as well. IMO only `Manage` and `DeleteSymbols` should actually support this routing.

There are many `PackageVersionAction`s, including the ones I mentioned above but also the `Edit` and `Delete` endpoints hit by the forms on the page.

src/NuGetGallery/UrlHelperExtensions.cs
scottbommarito: I think `ManagePackageVersion` would be confusing because it's called `ManagePackage` everywhere else.

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
scottbommarito: The individual pages use a lot of the same fields (ID, version, etc). I initially made a single model that stored the old models for each page but when I noticed there was so much redundancy, I combined them into a single model with all of the fields used by each. I believe this is less complex.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
scottbommarito: See above--I don't think it's a worthwhile change.

src/NuGetGallery/Controllers/PackagesController.cs
skofman1: @anangaur , right now if a user doesn't have permissions to perform a certain action (for example, an org collaborator can't manage owners of a package) we hide the section. Perhaps we should show the section as grayed out with message "You don't have permissions to do this".
What do you think?

src/NuGetGallery/UrlHelperExtensions.cs
scottbommarito: Did some refactoring of these routes--`PackageAction` wasn't used at all.

src/NuGetGallery/Views/Packages/Manage.cshtml
```diff
@@ -0,0 +1,107 @@
...
+                    this,
+                    "Listing",
+                    "Listing",
+                    @<text>@Html.Partial("_Delete", Model)</text>,
+                    expanded: false)
+            }
+
+            @if (Model.CanEdit)
+            {
+                @ViewHelpers.Section(
```
chenriksson: Per comment above, would recommend removing the "!Available || Validating" check inside here. I'd still display the section, but display a message that editing documentation isn't available until the package is published.

src/NuGetGallery/Views/Packages/Manage.cshtml
```diff
@@ -0,0 +1,107 @@
+﻿@model ManagePackageViewModel
```
chenriksson: I think the DisplayPackage page originally had perf issues for packages with many versions. I'd recommend testing this page on FAKE or similar package before merging.

src/NuGetGallery/Views/Packages/Manage.cshtml
```diff
@@ -0,0 +1,107 @@
...
+
+            EditReadMeManager.init(
+                readMeModel,
+                '@Url.PreviewReadMe()');
+        });
+    </script>
+    @ViewHelpers.SectionsScript(this)
+    @Scripts.Render("~/Scripts/gallery/page-manage-owners.min.js")
+    @Scripts.Render("~/Scripts/gallery/page-delete-package.js")
+    @Scripts.Render("~/Scripts/gallery/page-edit-readme.js")
```
chenriksson: Are we not consistent with JS minimization?

src/NuGetGallery/Views/Packages/Manage.cshtml
```diff
@@ -0,0 +1,107 @@
...
+        var strings_InvalidUsername = "@Html.Raw(Strings.ManagePackageOwners_InvalidUsername)";
+        var strings_AlreadyPending = "@Html.Raw(Strings.ManagePackageOwners_AlreadyPending)";
+        var strings_RemovingOwnership = "@Html.Raw(Strings.ManagePackageOwners_RemovingOwnership)";
+        var strings_RemovingOrganization = "@Html.Raw(Strings.ManagePackageOwners_RemovingOrganization)";
+        var strings_RemovingSelf = "@Html.Raw(Strings.ManagePackageOwners_RemovingSelf)";
+
+        // Set up delete section
+        var versionListedState = @Html.Raw(
+           Json.Encode(
+               Model.VersionSelectList.ToDictionary(
```
chenriksson: Looks like delete and readme sections both serialize the versions list? Are we taking a performance hit by not consolidating the JS files and using shared models?

src/NuGetGallery/Views/Packages/_Delete.cshtml
```diff
@@ -0,0 +1,101 @@
...
+                        {
+                            @class = "form-control",
+                            title = "Select a version",
+                            id = "input-select-version"
+                        })
+
+        if (!Model.FailedValidation)
+        {
+            <b>
+                List/unlist version
```
chenriksson: Why not keep same "Package Version Listing" or at least capatalization?

src/NuGetGallery/Views/Packages/_Delete.cshtml
```diff
@@ -0,0 +1,101 @@
...
+                    </div>
+                }
+            </div>
+        }
+    }
+
+    @if (User.IsAdministrator())
+    {
+        <b>
+            Delete version
```
chenriksson: Why not keep same "Delete Package Version" or at least capatalization?

src/NuGetGallery/Views/Packages/_Delete.cshtml
```diff
@@ -0,0 +1,101 @@
+﻿@model ManagePackageViewModel
```
chenriksson: FYI - was useful to compare against previous view... not much changed.
https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Views/Packages/Delete.cshtml

src/NuGetGallery/Views/Packages/_Owners.cshtml
```diff
@@ -0,0 +1,92 @@
...
+
+    <div style="display: none" id="manage-owners-alert" data-bind="visible: message">
+        @ViewHelpers.AlertDanger(
+            @<text>
+                <span id="manage-owners-message" role="alert" aria-live="assertive" data-bind="text: message"></span>
+            </text>)
+    </div>
+
+    <b>
+        Current owners
```
chenriksson: Can we keep capitalization (Current Owners)? Kind of a subheading, right?

src/NuGetGallery/Views/Packages/_Owners.cshtml
```diff
@@ -0,0 +1,92 @@
+﻿@model ManagePackageViewModel
```
chenriksson: FYI - useful to compare:
https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Views/Packages/ManagePackageOwners.cshtml

src/NuGetGallery/Views/Packages/_Owners.cshtml
```diff
@@ -0,0 +1,92 @@
...
+                        <span>Remove</span>
+                    </a>
+                </div>
+                <!-- /ko -->
+            </div>
+        </div>
+    </div>
+
+    <b>
+        Add owner
```
chenriksson: Keep capitalization (Add Owner)?

src/NuGetGallery/Views/Packages/_ReadMe.cshtml
```diff
@@ -0,0 +1,55 @@
+﻿@model ManagePackageViewModel
```
chenriksson: Useful to compare:
https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Views/Packages/Edit.cshtml

src/NuGetGallery/Views/Shared/_ListPackage.cshtml
```diff
@@ -10,26 +24,31 @@
...
+                        </a>
+                    }
+                </div>
+                <div class="hidden-xs">
+                    @OwnersHeader()
+                </div>
+            </div>
+
+            <div class="hidden-sm hidden-md hidden-lg">
+                @OwnersHeader()
```
chenriksson: Why is there a second `@OwnersHeader()` outside `package-header`?

src/NuGetGallery/Views/Shared/_ListPackage.cshtml
```diff
@@ -1,5 +1,19 @@
 ﻿@model ListPackageItemViewModel
 
+@helper OwnersHeader()
```
chenriksson: Packages route (https://www.nuget.org/packages)? Changes look more extensive than just changing 3 links to 1... can you add a screenshot?

src/NuGetGallery/Views/Users/_UserPackagesList.cshtml
```diff
@@ -1,80 +0,0 @@
-﻿@model ManagePackagesListViewModel
```
chenriksson: Looks like this was already dead code that wasn't in the project anymore?

src/Bootstrap/less/theme/page-delete-package.less
scottbommarito: I like that naming scheme. I will rename the files.

src/NuGetGallery/UrlHelperExtensions.cs
scottbommarito: Nope, I merged them into a single function to align with how the code worked previously.

src/NuGetGallery/Areas/Admin/ViewModels/ValidatedPackageViewModel.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace NuGetGallery.Areas.Admin.ViewModels
 {
-    public class ValidatedPackageViewModel
+    public class ValidatedPackageViewModel : IPackageVersionModel
```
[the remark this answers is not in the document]
scottbommarito: Yup. Since I was touching `UrlHelperExtensions` I felt that I might as well improve the code.

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -1865,15 +1863,16 @@
     <Content Include="Views\Users\Account.cshtml" />
     <Content Include="Views\Packages\ConfirmOwner.cshtml" />
     <Content Include="Views\Errors\NotFound.cshtml" />
-    <Content Include="Views\Packages\Delete.cshtml" />
-    <Content Include="Views\Packages\Edit.cshtml" />
-    <Content Include="Views\Packages\ManagePackageOwners.cshtml" />
+    <Content Include="Views\Packages\_Delete.cshtml" />
+    <Content Include="Views\Packages\_ReadMe.cshtml" />
```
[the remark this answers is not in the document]
scottbommarito: I left "edit' in the code to reduce the number of things I had to rename. I think this file should be `_Documentation.cshtml` and that the other files be named "documentation" as well.

src/NuGetGallery/UrlHelperExtensions.cs
```diff
@@ -772,46 +768,24 @@ public static string GetOrganizationCertificates(this UrlHelper url, string acco
...
 
         /// <summary>
-        /// Initializes a delete package link that can be resolved at a later time.
+        /// Initializes a delete symbols package link that can be resolved at a later time.
         /// 
         /// Callers should only use this API if they need to generate many links, such as the ManagePackages view
         /// does. This template reduces the calls to RouteCollection.GetVirtualPath which can be expensive. Callers
         /// that only need a single link should call Url.DeletePackage instead.
-        public static RouteUrlTemplate<IPackageVersionModel> DeletePackageTemplate(
+        public static RouteUrlTemplate<IPackageVersionModel> DeleteSymbolsPackageTemplate(
```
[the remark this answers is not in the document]
scottbommarito: I thought I deleted this, woop!

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        public bool IsCurrentUserAnAdmin;
```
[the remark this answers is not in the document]
scottbommarito: This was copied over from an old view model. I will fix.

src/NuGetGallery/Views/Packages/Delete.cshtml
```diff
@@ -1,157 +0,0 @@
-﻿@model DeletePackageViewModel
```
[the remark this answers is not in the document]
scottbommarito: I'd prefer not to, because nothing in the view model is specific to symbols.

src/NuGetGallery/Views/Packages/Edit.cshtml
```diff
@@ -1,112 +0,0 @@
-﻿@model EditPackageRequest
```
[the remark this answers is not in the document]
scottbommarito: Doesn't appear to be used. I have removed.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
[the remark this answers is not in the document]
scottbommarito: I agree. Will fix.

src/NuGetGallery/Views/Packages/_Delete.cshtml
```diff
@@ -0,0 +1,101 @@
...
+    {
+        <b>
+            Delete version
+        </b>
+
+        <div class="list-unstyled">
+            @ViewHelpers.AlertDanger(
+                            @<text>
+                                <strong>Deleting this package will make it unavailable for download and package restore.</strong><br />
+                                This package version has been downloaded @Model.DownloadCount times (@Model.TotalDownloadCount downloads in total).
```
scottbommarito: TODO: fix this so that it shows the correct download counts for the package selected

src/NuGetGallery/Views/Packages/Manage.cshtml
scottbommarito: Rethought this a bit. The listing and documentation sections should show as long as any version exists that can be edited or unlisted/relisted.

src/NuGetGallery/Controllers/PackagesController.cs
anangaur: Makes sense to me. That makes it clearer. Have the section which on expansion is grayed out with the warning message.

src/NuGetGallery/Views/Packages/_Delete.cshtml
scottbommarito: done

src/NuGetGallery/Views/Users/_UserPackagesList.cshtml
scottbommarito: Yup, I removed it.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -2384,6 +2309,35 @@ public virtual async Task<JsonResult> PreviewReadMe(ReadMeRequest formData)
             }
         }
 
+        [UIAuthorize]
+        [HttpGet]
+        public virtual async Task<JsonResult> GetReadMe(string id, string version)
```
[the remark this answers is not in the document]
scottbommarito: Yeah, I think so.

src/NuGetGallery/Views/Packages/Manage.cshtml
scottbommarito: It should be okay--the manage packages page is much more complex and runs great. I will check it though.

src/NuGetGallery/Views/Packages/_Delete.cshtml
scottbommarito: Note--I made some more substantial changes after you reviewed this.

src/NuGetGallery/Views/Packages/_Owners.cshtml
scottbommarito: I tried to align with the spec for deprecate packages, which has this capitalization pattern for the subheadings.

E.g. "Select version(s)"

src/NuGetGallery/Views/Packages/_Owners.cshtml
scottbommarito: See above:

I tried to align with the spec for deprecate packages, which has this capitalization pattern for the subheadings.

E.g. "Select version(s)"

src/NuGetGallery/Views/Shared/_ListPackage.cshtml
scottbommarito: To clarify, I wanted the owners list to be on its own line when the page is tiny (xs), but for it to appear on the right of the heading if the page is larger (sm, md, lg). By adding the `hidden-` classes, I have accomplished this effect.

src/NuGetGallery/App_Start/Routes.cs
scottbommarito: I don't think this route even works...

`/packages/ContactOwners/{id}` doesn't work locally.

Will remove.

scottbommarito:
@skofman1 @anangaur I have removed the edit message!

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I currently have the section still openable, with the warning message when it's expanded. This was very easy to implement. I'd need to see a mock-up of how you'd like the warning to display when grayed out because I'm not totally sure what you mean.

src/Bootstrap/less/theme/common-list-packages.less
```diff
@@ -7,12 +7,20 @@
     font-size: 15px;
 
     .package-header {
+      display: flex;
```
joelverhagen: >display [](start = 6, length = 7)

What's the browser support of this?

Let's look at Google Analytics to see what our browser usage is right now and verify this looks okay on our most popular browsers.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: image

src/NuGetGallery/Scripts/gallery/page-delete-package.js
```diff
@@ -20,4 +20,20 @@
             e.preventDefault();
         }
     });
+
+    $('.page-delete-package #input-select-version').change(function () {
```
joelverhagen: >change [](start = 52, length = 6)

Can we do `history.replaceState` here and the readme dropdown so that the URL bar has the currently selected version in case of refresh?

joelverhagen:
Any URLs that we are removing should have redirects. In the future we can consider eliminated the redirect if we see no traffic for a long period, but now is too early.

src/Bootstrap/less/theme/common-list-packages.less
scottbommarito: We have a lot of uses of flex already, I don't think this is necessary.

Note: this was a UI change suggested by @anangaur. I can move it out of the PR if you'd like.

new:
image

old:
image

src/NuGetGallery/Scripts/gallery/page-delete-package.js
scottbommarito: I like that idea. This reminds me that I need to make sure the readmes loaded are cached, so that when it redirects back it doesn't have to make another request (that may fail).

src/NuGetGallery/NuGetGallery.csproj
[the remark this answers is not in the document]
chenriksson: I would be consistent with the naming pattern you choose for JS/CSS sections... which is why I suggested leaving it as `_Edit`.

src/NuGetGallery/Views/Shared/_ListPackage.cshtml
scottbommarito: @anangaur asked me to do a UI change as part of this work.

new:
image

old:
image

I could move it out of this PR if you think it's best.

src/NuGetGallery/Views/Shared/_ListPackage.cshtml
scottbommarito: As mentioned below, this was a UI change that @anangaur wanted me to make as part of this work. I can remove it from this PR if you think it's best.

src/NuGetGallery/Views/Packages/Manage.cshtml
scottbommarito: I fixed this--we didn't have a minified version of `page-delete-package` or `page-edit-readme`

src/Bootstrap/less/theme/page-delete-package.less
scottbommarito: Upon thinking about it some more, I think renaming the `.js` and `.less` files is not a worthwhile change. Both `page-edit-readme` and `page-delete-package` are reused in multiple pages (e.g. the upload flow and the delete symbols page).

src/NuGetGallery/Scripts/gallery/page-delete-package.js
scottbommarito: Actually, I misunderstood your question. Could you give me an example how `history.replaceState` works? I wasn't able to find a good example online.

Also, I think it's important that whatever solution we choose only remembers state if refreshing or going back. If a user goes to the page, navigates away, and then navigates back, it should select the version that the user was looking at and not what they were looking at before.

scottbommarito:
@joelverhagen I have added redirects to the old routes: https://github.com/NuGet/NuGetGallery/pull/6805/commits/8fcd0f873a6ab51c7546fbfaeaddd9b330c8a7be

src/NuGetGallery/NuGetGallery.csproj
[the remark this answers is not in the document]
scottbommarito: I renamed all the partials because they are very specific to the Manage Package page.

They are now `_ManageOwners`, `_ManageListing`, and `_ManageDocumentation`. I think this is a much better pattern.

src/Bootstrap/less/theme/common-list-packages.less
joelverhagen: Alright, sounds good.

joelverhagen:
Make sure to manually test this new page in Chrome, Firefox, Edge, IE and a mobile browser.

Also, we should run that automated accessibility tool as well.

src/NuGetGallery/Views/Shared/_ListPackage.cshtml
chenriksson: Thanks for the clarification - sounds fine to me.

joelverhagen:
Just watched an accessibility training video. Did we do a pass of the following on this new page?

- 200% zoom should be legible
- content wraps nicely with different window sizes
- keyboard interaction works okay, especially tab order
- screen reader
- high contrast

And, as mentioned before, Keros.

scottbommarito:
@joelverhagen - There were some Keros issues (mostly due to the fact that the forms had some duplicate IDs) but I fixed them. No other accessibility issues that I could find though.

PR #6806, opened 2019-01-09
[Feature Flags] Add gallery integration

loic-sharma:
Part of https://github.com/NuGet/Engineering/issues/2016
Depends on https://github.com/NuGet/ServerCommon/pull/273/files
Depends on https://nuget.visualstudio.com/NuGetMicrosoft/_git/NuGetDeployment/pullrequest/740

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2369123


src/NuGetGallery/Configuration/AppConfiguration.cs
```diff
@@ -62,11 +62,18 @@ public class AppConfiguration : IAppConfiguration
...
+        public string AzureStorage_FeatureFlags_ConnectionString { get; set; }
+
         /// <summary>
         /// Gets a setting if Read Access Geo Redundant is enabled in azure storage
         /// </summary>
         public bool AzureStorageReadAccessGeoRedundant { get; set; }
 
+        public TimeSpan FeatureFlagsRefreshInterval { get; set; }
+
+        public TimeSpan FeatureFlagsMaximumStaleness { get; set; }
```
joelverhagen: >TimeSpan [](start = 15, length = 8)

This is non-nullable. The ServerCommon counterpart is nullable. which is right?

src/NuGetGallery/Services/FeatureFlagService.cs
```diff
@@ -0,0 +1,24 @@
...
+        private readonly IFeatureFlagClient _featureFlagClient;
+        private readonly IFlightClient _flightClient;
+
+        public FeatureFlagService(IFeatureFlagClient featureFlagClient, IFlightClient flightClient)
+        {
+            _featureFlagClient = featureFlagClient ?? throw new ArgumentNullException(nameof(featureFlagClient));
+            _flightClient = flightClient ?? throw new ArgumentNullException(nameof(flightClient));
+        }
+
+        // TODO...
```
joelverhagen: >TODO [](start = 11, length = 4)

Unless this will be done in this PR, put a github issue with a TODO

joelverhagen:
> using System;

copyright

---
Refers to: src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs:1 in c8876c7. [](commit_id = c8876c7e6db17513b403a221b441415ee0e7e834, deletion_comment = False)

src/NuGetGallery/App_Start/AppActivator.cs
```diff
@@ -263,6 +265,12 @@ private static void BackgroundJobsPostStart(IAppConfiguration configuration)
                 }
             }
 
+            var featureFlags = DependencyResolver.Current.GetService<IFeatureFlagRefreshService>();
+            if (featureFlags != null)
+            {
+                HostingEnvironment.QueueBackgroundWorkItem(featureFlags.RunAsync);
```
joelverhagen: >RunAsync [](start = 72, length = 8)

Does this need to loop once before any requests are taken?

src/NuGetGallery/App_Data/Files/Feature-Flags/Flags.json
```diff
@@ -0,0 +1,7 @@
+{
+  "Features": {
+  },
+
+  "Flights": {
```
joelverhagen: >Flights [](start = 3, length = 7)

Nit: casing of this directory and file name should match the config (i.e. be all lowercase). It's fine on windows file system but it's must to be consistent when possible IMO

joelverhagen:
If gallery can't load the feature flags file for whatever reason, how will we know?

It think we should have an AI metric for the staleness. This can be Sev 3/4 monitored.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -393,6 +396,38 @@ protected override void Load(ContainerBuilder builder)
...
+
+            builder
+                .RegisterType<FeatureFlagRefreshService>()
+                .As<IFeatureFlagRefreshService>()
+                .SingleInstance();
+
+            builder
+                .RegisterType<FeatureFlagClient>()
+                .As<IFeatureFlagClient>()
+                .InstancePerDependency();
```
skofman1: why InstancePerDependency? FeatureFlagClient depends on a logger and IFeatureFlagCacheService and has no state, meaning it can also be SingleInstance

loic-sharma: This could be SingleInstance, yes. However, we've taken the approach of using the most restrictive lifetime possible for dependencies in the validation pipeline. I was sticking to that pattern.

skofman1: Creating more objects for each request impacts performance, and Gallery is far more sensitive to this since there are many more requests.

src/NuGetGallery/App_Start/AppActivator.cs
scottbommarito: `ContentObjectService` doesn't do this. It has a default state that it returns until it loads successfully. I would imagine that feature flags would function similarly to this (features are off if the configuration fails to load). Is this the plan?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
scottbommarito: Given that
- our approach to dependencies in the gallery has been "make it `InstancePerDependency` unless it must be `SingleInstance`"
- making this dependency, which works perfectly fine as `InstancePerDependency`, a `SingleInstance`, may lead to some cryptic errors if we add state to it
- we don't have evidence that adding this service and making it an `InstancePerDependency` has any substantial performance impact

I would argue we should probably leave this an `InstancePerDependency` and change it when necessary or when we do an analysis of the best practice for dependencies on the gallery.

tests/NuGetGallery.Facts/App_Start/StorageDependentFacts.cs
```diff
@@ -73,14 +75,16 @@ public void StorageDependentsHaveExpectedTypes()
             Assert.Contains(typeof(UploadFileService), implementationToInterface.Keys);
             Assert.Contains(typeof(CoreLicenseFileService), implementationToInterface.Keys);
             Assert.Contains(typeof(RevalidationStateService), implementationToInterface.Keys);
-            Assert.Equal(7, implementationToInterface.Count);
+            Assert.Contains(typeof(FeatureFlagFileStorageService), implementationToInterface.Keys);
```
scottbommarito: We can write this test much less redundantly...

```
var implementationTypes = new []
{
    typeof(CertificateService),
    ...
    typeof(FeatureFlagFileStorageService)
};

Assert.Equal(implementationTypes.Count(), implementationToInterface.Count);

foreach (var implementationType in implementationTypes)
{
    Assert.Contains(implementationType , implementationToInterface.Keys);
    Assert.Equal(implementationType, implementationToInterface[implementationType]);
}
```

src/NuGetGallery/App_Data/Files/Feature-Flags/Flags.json
loic-sharma: I was following the pattern set by already existing files, but I agree with your comment. I'll fix the new files but won't touch the already existing ones.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
skofman1: Still not convinced why we need to create objects when it's not needed.

loic-sharma: I agree with @scottbommarito but I don't feel strongly about this. These classes are now registered as `SingleInstance`

tests/NuGetGallery.Facts/App_Start/StorageDependentFacts.cs
loic-sharma: I think the test is okay-ish right now. If it continues to get bigger, I agree this would useful. Please let me know if you feel strongly that I should change it now

src/NuGetGallery/App_Start/AppActivator.cs
loic-sharma: I went ahead and forced feature flags to be loaded once at startup. This will let us to know at startup if there are any issues with feature flags.

src/NuGetGallery/Configuration/AppConfiguration.cs
loic-sharma: This was removed

src/NuGetGallery/Services/FeatureFlagService.cs
loic-sharma: This was added in this PR

src/NuGetGallery.Core/Features/IFeatureFlagClientExtensions.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using NuGet.Services.Entities;
+using NuGet.Services.FeatureFlags;
+
+namespace NuGetGallery.Features
+{
+    public static class IFeatureFlagClientExtensions
```
agr: `I`FeatureFlagClientExtensions?

loic-sharma: Removed

Issue #6531, opened 2018-10-05
[Licenses] Gallery license preview


agr:
Verify package page should display preview of the embedded license if there is one or display license expression if it was specified in a package.

PR #6807, opened 2019-01-10
Embedded license preview when uploading the package through UI

agr:
Addresses #6531 

`VerifyPackageRequest` view model was extended with two properties: `LicenseExpression` and `LicenseFileContents` which are when set cause "License expression" and "License file" items to appear on verify package page.

Two code paths cause verify package page to be presented to the user: regular upload and picking up the past upload, so you can find two calls to `GetLicenseFileContentsAsync`.

License file contents is presented through `<pre>` to preserve formatting of the original text file as much as possible, yet some styling changes were made to make long lines to properly wrap.

Changes to `bootstrap.js` and `package-lock.json` happened automagically while building the .less files and probably should be ignored (I can revert, but they'll come back next time someone runs `grunt`).

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -603,17 +603,17 @@ namespace NuGetGallery
 
         private static async Task<string> GetLicenseFileContentsAsync(PackageMetadata packageMetadata, PackageArchiveReader packageArchiveReader)
         {
-            if (packageMetadata.LicenseMetadata?.Type == LicenseType.File)
+            if (packageMetadata.LicenseMetadata?.Type != LicenseType.File)
             {
-                var licenseFilename = FileNameHelper.GetZipEntryPath(packageMetadata.LicenseMetadata?.License);
-                using (var licenseFileStream = packageArchiveReader.GetStream(licenseFilename))
-                using (var streamReader = new StreamReader(licenseFileStream, Encoding.UTF8))
-                {
-                    return await streamReader.ReadToEndAsync();
-                }
+                return null;
             }
 
-            return null;
+            var licenseFilename = FileNameHelper.GetZipEntryPath(packageMetadata.LicenseMetadata?.License);
+            using (var licenseFileStream = packageArchiveReader.GetStream(licenseFilename))
+            using (var streamReader = new StreamReader(licenseFileStream, Encoding.UTF8))
+            {
+                return await streamReader.ReadToEndAsync();
+            }
         }
 
         public virtual async Task<ActionResult> DisplayPackage(string id, string version)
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -571,15 +580,42 @@ private async Task<JsonResult> UploadPackageInternal(PackageArchiveReader packag
...
 
             var model = new VerifyPackageRequest(packageMetadata, accountsAllowedOnBehalfOf, existingPackageRegistration);
             model.IsSymbolsPackage = isSymbolsPackageUpload;
             model.HasExistingAvailableSymbols = hasExistingSymbolsPackageAvailable;
             model.Warnings.AddRange(warnings.Select(w => new JsonValidationMessage(w)));
+            model.LicenseFileContents = licenseFileContents;
             return Json(model);
         }
 
+        private static async Task<string> GetLicenseFileContentsAsync(PackageMetadata packageMetadata, PackageArchiveReader packageArchiveReader)
```
joelverhagen: `GetLicenseFileContentsOrNullAsync`

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -571,15 +580,42 @@ private async Task<JsonResult> UploadPackageInternal(PackageArchiveReader packag
...
+        {
+            if (packageMetadata.LicenseMetadata?.Type != LicenseType.File)
+            {
+                return null;
+            }
+
+            var licenseFilename = FileNameHelper.GetZipEntryPath(packageMetadata.LicenseMetadata?.License);
```
joelverhagen: We know `LicenseMetadata` is not null at this point, right?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -288,7 +288,15 @@ public virtual async Task<ActionResult> UploadPackage()
...
             verifyRequest.IsSymbolsPackage = false;
-
+            try
+            {
+                verifyRequest.LicenseFileContents = await GetLicenseFileContentsAsync(packageMetadata, packageArchiveReader);
+            }
+            catch (Exception ex)
+            {
+                _telemetryService.TraceException(ex);
+                return View(model);
```
joelverhagen: We have a `validationErrorMessage` above for some error cases, should we have the same thing here?

joelverhagen:
I think it would be more consistent to linkify the license expression. Could we get a PM's insight on this? @anangaur, @karann-msft

anangaur:
Agree! Hope this can be done easily.

src/NuGetGallery/Controllers/PackagesController.cs
agr: This is the case when someone comes to the upload page, but they had incomplete package upload from earlier. Our code tries to recover that incomplete upload, tries to extract the license file if one exists and fails bad enough so that exception is thrown. The whole scenario is complicated so I am hesitant to bubble up any error to the user at this point, anyway, the only way for them to "recover" is to reupload the file, and that's exactly what's going to be prompted from them.

src/NuGetGallery/Controllers/PackagesController.cs
agr: Yup.

PR #6808, opened 2019-01-10
Not null the createdby on delete account

cristinamanum:
On delete account the issue. createdby field will be obfuscated but not changed to null . The issues will be marked as resolved and still be visible on the support page.


src/NuGetGallery/Areas/Admin/Services/SupportRequestService.cs
```diff
@@ -20,6 +20,8 @@ public class SupportRequestService
         private IAuditingService _auditingService;
         private readonly string _siteRoot;
         private const string _unassignedAdmin = "unassigned";
+        private const string _deletedAccount = "_deletedaccount";
+        private const string _NuGetDSRAccount = "_NuGetDSR";
```
joelverhagen: We should have a unit test in the registration flow asserting that this account name is not a valid account name.

We picked the underscore prefix because today this cannot be an actual username.

cristinamanum: Added unit tests for the username regex verification.

PR #6820, opened 2019-01-14
Introduce package deprecation entities and migration

scottbommarito:
https://github.com/NuGet/Engineering/issues/2000

Based on https://github.com/NuGet/ServerCommon/pull/271 which had to be moved to this repo because NuGet.Services.Entities was moved.


joelverhagen:
What's the SQL generated by this migration?

src/NuGet.Services.Entities/PackageDeprecationStatus.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// The package is deprecated because of some other undefined reason.
+        /// </summary>
+        Other = 1,
+
+        /// <summary>
+        /// The package is deprecated because it is legacy and no longer maintained.
+        /// </summary>
+        IsLegacy = 2,
```
joelverhagen: I think the `Is` prefix is noise. Why not `IsNotDeprecated` and `IsOther` 😉

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -367,6 +367,35 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
...
                 .Property(s => s.RowVersion)
                 .IsRowVersion();
+
+            modelBuilder.Entity<PackageVulnerability>()
+                .HasKey(v => v.Key);
+
+            modelBuilder.Entity<PackageDeprecation>()
+                .HasKey(d => d.Key)
+                .HasRequired(d => d.DeprecatedPackage)
+                .WithRequiredDependent(p => p.Deprecation)
```
joelverhagen: Don't you want `WithOptional` here?

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <summary>
+        /// Gets or sets the primary key for the entity.
+        /// </summary>
+        public int Key { get; set; }
+        
+        /// <summary>
+        /// Gets or sets the deprecated package entity.
+        /// </summary>
+        [Required]
+        public virtual Package DeprecatedPackage { get; set; }
```
joelverhagen: Didn't we need a unique index on the DeprecatedPackageKey?

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,67 @@
...
+        
+        /// <summary>
+        /// Gets or sets the user that executed the package deprecation.
+        /// </summary>
+        public virtual User DeprecatedBy { get; set; }
+
+        /// <summary>
+        /// The date when the package was deprecated.
+        /// </summary>
+        [Required]
```
joelverhagen: see which is more -- required via code first or required via attribute. do the more common approach

joelverhagen: And isn't this required by default since it is a value type?

src/NuGet.Services.Entities/PackageDeprecationStatus.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using System;
+
+namespace NuGet.Services.Entities
+{
+    /// <summary>
+    /// Indicates package deprecation status and reason(s).
+    /// </summary>
+    [Flags]
+    public enum PackageDeprecationStatus
```
joelverhagen: add unit tests that assert these enum values never change. we have something similar for `PackageStatusKey` and `SemVerLevelKey`

src/NuGetGallery.Core/Entities/EntitiesContext.cs
skofman1: I don't see "Deprecation" property in PackageDeprecation

---
In reply to: [247708520](https://github.com/NuGet/NuGetGallery/pull/6820#discussion_r247708520) [](ancestors = 247708520)

scottbommarito: This is the `Deprecation` property on `Package`

src/NuGet.Services.Entities/PackageDeprecation.cs
scottbommarito: Removed `[Required]` due to the nature of value types.

scottbommarito:
@joelverhagen - current SQL:

```
CREATE TABLE [dbo].[PackageDeprecations] (
    [Key] [int] NOT NULL,
    [DeprecatedPackageKey] [int] NOT NULL,
    [Status] [int] NOT NULL,
    [AlternatePackageRegistrationKey] [int],
    [AlternatePackageKey] [int],
    [DeprecatedByKey] [int],
    [DeprecatedOn] [datetime] NOT NULL,
    [CustomMessage] [nvarchar](max),
    [AlternatePackage_Key] [int],
    [AlternatePackageRegistration_Key] [int],
    [DeprecatedBy_Key] [int],
    [PackageVulnerability_Key] [int],
    CONSTRAINT [PK_dbo.PackageDeprecations] PRIMARY KEY ([Key])
)
CREATE INDEX [IX_Key] ON [dbo].[PackageDeprecations]([Key])
CREATE INDEX [IX_AlternatePackage_Key] ON [dbo].[PackageDeprecations]([AlternatePackage_Key])
CREATE INDEX [IX_AlternatePackageRegistration_Key] ON [dbo].[PackageDeprecations]([AlternatePackageRegistration_Key])
CREATE INDEX [IX_DeprecatedBy_Key] ON [dbo].[PackageDeprecations]([DeprecatedBy_Key])
CREATE INDEX [IX_PackageVulnerability_Key] ON [dbo].[PackageDeprecations]([PackageVulnerability_Key])
CREATE TABLE [dbo].[PackageVulnerabilities] (
    [Key] [int] NOT NULL IDENTITY,
    [CVSSRating] [int],
    [CVEIds] [nvarchar](max),
    [CWEIds] [nvarchar](max),
    CONSTRAINT [PK_dbo.PackageVulnerabilities] PRIMARY KEY ([Key])
)
ALTER TABLE [dbo].[PackageDeprecations] ADD CONSTRAINT [FK_dbo.PackageDeprecations_dbo.Packages_AlternatePackage_Key] FOREIGN KEY ([AlternatePackage_Key]) REFERENCES [dbo].[Packages] ([Key])
ALTER TABLE [dbo].[PackageDeprecations] ADD CONSTRAINT [FK_dbo.PackageDeprecations_dbo.PackageRegistrations_AlternatePackageRegistration_Key] FOREIGN KEY ([AlternatePackageRegistration_Key]) REFERENCES [dbo].[PackageRegistrations] ([Key])
ALTER TABLE [dbo].[PackageDeprecations] ADD CONSTRAINT [FK_dbo.PackageDeprecations_dbo.Users_DeprecatedBy_Key] FOREIGN KEY ([DeprecatedBy_Key]) REFERENCES [dbo].[Users] ([Key])
ALTER TABLE [dbo].[PackageDeprecations] ADD CONSTRAINT [FK_dbo.PackageDeprecations_dbo.Packages_Key] FOREIGN KEY ([Key]) REFERENCES [dbo].[Packages] ([Key]) ON DELETE CASCADE
ALTER TABLE [dbo].[PackageDeprecations] ADD CONSTRAINT [FK_dbo.PackageDeprecations_dbo.PackageVulnerabilities_PackageVulnerability_Key] FOREIGN KEY ([PackageVulnerability_Key]) REFERENCES [dbo].[PackageVulnerabilities] ([Key])
VALUES (N'201901142143060_AddDeprecationEntities', N'NuGetGallery.Migrations.MigrationsConfiguration',  0x1F8B0800000000000400ED7D5B731D3972E6FB46F83F30F8643BDAA2A8EE198F15A21D6A8AEA6158B7D0616BF68D513C059165D5A93A5355472267637FD93EEC4FF25F70DD0B970490B8D48DC3E808354F014800890F092091C8FCEFFFF7FF5FFDC7FD2E3EFA4EB23C4A93B3E3D367CF8F8F48B24DC328B93D3B3E145FFFE54FC7FFF1EFFFF0BF5E5D84BBFBA32F5DBE9FAB7C65C9243F3BBE2B8AFDCB93937C7B477641FE6C176DB3344FBF16CFB6E9EE2408D39317CF9FFFDBC9E9E90929491C97B48E8E5E7D3E2445B423F58FF2E7799A6CC9BE3804F1FB342471DE7E2F533635D5A30FC18EE4FB604BCE8E3F1C7E23C56F411C93ECE1F8E8751C05651B3624FE7A7C1424495A0445D9C297BFE76453646972BBD9971F82F8EA614FCA7C5F8338276DCB5F0ED9B19D78FEA2EAC4C950B023B53DE445BA332478FA73CB9513BEB8156F8F7BAE957CBB28F95B3C54BDAE7977767C4EB222FA1A6D83A2EC3F5FDFCBF338ABF2B6DC7DB621D9F7684BF2673599A8FC832AFED39124D34F3D3E9E3FABFFFBE9E8FC1017878C9C25E4506441FCD3D1A7C34D1C6DFF933C5CA5DF4872961CE2986E77D9F2328DF9507EFA94A5FBB2FE87CFE46BDB9B32D3F1D1095BF0842FD997A30B35DDBC4C8A9F5F1C971D89E3E026263D3028966C8A3423BF918464659FC34F415190AC1CD7CB90D4AC15AAE72ADBDC05A7577787DDCD3E8B92A2ABB7C46439B18E8FDE07F7EF48725BDC9D1DFF52CEA4B7D13D09BB0F6D537E4FA2721A96658AEC4080A6AAAB4755FDE20F7F1CA3EECDE1E6BFC8565571F927AA621D8BD3AC98AAB2CB3C3F906C9A3E4D54D7C5FD3ECA5A79D354F5A684FA55299785921F82EFD16D9D95A351CAD98C920DF9F1D16712D719F3BB68DF08665A785C8B05DE66E9EE731AB3324AC877BD490FD9B66CD9558AC87C1564B7A460BBF1EA64108A4A51C951B311971C89279189119914C310F5AA695503604E440A73060E6321BC032D0AE1DD743099A860D3AB04659BC10C4263E15C502B8DE6A1EDE47B9A71981977B10BA2F875186624CF3DEC133453B23C56245FA36C47C269EB2DE19004D5A226ADE88FBF8CB10F6A7A19C7E90F127695FF9A96B325488C695DE66F484C0A7742174995F97D09F9E86DB02DE1F3FA50DC5568D9323B015BF21FD272F23F7C0AB6DF825BF2E990DF796871C5C6F3063A751BEB59A918CD6AD238A3E65390E73FD22CFC4C7252CC54E3B0417B532F3C9A4D1ABF6665A4120ABF175BD392EF82BC785B32BDEC4F7A1B251614A8D2E7E961388438AFC319A9455B590A5E87FBF4EB66F5A0965E36495C6DB974D305F67D29CB4ACCC3EDAA290E39B8C5B54B10F6B86C2AB4A955B5E863761B24D1DFEAA4F7D16D83A4CFE4AF079217F256AA4B712D5765867BA32CE1B3878A81D01433E8A364C8D4455C7A69D639549FF05DF1D20364D3D56D4634D6BC9595E0CDBE93B05736C24D15B25D7FFC516EC09806CBF20852479AD154FC54E414AC6D933996D65F61563649A62CDC90ED212B379E9FD272031D291AC4647C1084B5240B78EA80F2D99C8EB46A0CCB9312C85EBF1A8B61EDB2D2EDF6A59F4E4D9853939D6E8157D09665273F917C09E283AAD69114C257240992E2321CFDA8F786E4DB2CDAD30799D1EA1AD0327245ED3E5EDC825B40FC3CDDED0F3531846A98E4A6DBFEE1C072156DBFE5D40CA9F0AC3F7494330BE8277AF1D9963F61D15D275DD3627290DA7C9AB0C80819BCE9DE0C8E2B82FA5B729CB15A3FEA1EDA2C1D75C1A75503A5DDEE07CC79EDA8B788229565DCCAB50AAFD75B8D14C6D6E64157E02E12F8D9279519E8135135848A76B5E97C93EACF92D634694E72A056B5B5FA061B7140977F920A18A9F06B1AAA76107EA664852278D3AA111964D0DCD96A6FA7123B57A9E72BBF8E69E024A5817E3DE41CA62B984198B8702E5359D274DD45EF085EE6095A492B91A256279A8B1815BD25889C0FE407DD442B0904D170D92FBC0E7751E2E5D46A720DE36B8117B474F558C33722D6B3BD67D1345A6370C2E114CDD81E71209A42E16FDC2B37D1F29EEC6ECA695D76C7419E084496204434124426789ABE58891CAFF2866D8A0BA5CBBC9E98AE9B90472BB67A4E8F711FA49DCED6B2492B98E80CD74D0FA1B62BB209AD57E5F5B6C371DDD178173E47BF063969B1DE41E218394A3CDFC719277E4F8C1953FC9D7C2DDCB14DD7B519D558C7AB9C613D725BCDD6BA8C2D650D9B7901D340DACF1DBA56C04F23D9F5D30E21CB1D85B8700F6F33FB04224B9884CBD781E96E254F5FFC698C1DD965BEB90B3266C85D6D343F65E46B74EF6FB2778625BEAD50F8654C6BAE826D706BF8F999DC46799129EC7DC41AE1A2AACE4025105D038B39ADD900451BF101907912201801A2B42C18497ABC497F24711A8436E6AD82DC28C732FA1AF9B0127F976EBF59D0B19140005C011924CF254C5545564B39846FF85040DDF42E1FAAF17D66734BC4BF1EA27269DA44B746EC17CAA93BC36547F5892F339991A5A7F5416F80A95C565CD6078735E1691DC0AC03C0C879B8F1D83F64D1EDDDF8379873D879E10CE6FCF4AF9CE524C8C987B4180CCB46ABCCE3F27C715FF134885B74FD9EC5A3B7FECF417EF73ABE4DB3A8B8DBA9F63638DF01FACAA6B709BDDCA6C914AC2CB744957DB5B31D434767535485BC5123BB527EBFF0DB384F446B83CC7D38B94CAAEABD0823B05A4DC972694CF26966681492B6BACF649F66CEF87A1724B7877A93209F893E667BBD19A09F654E32A6ADEC7C1B95E88CFE4628815C19061BB2AAFC5999144D31CCD5D8E651D9F787696BD3BC0E387DEE49F07F2641F89E94C350AF700286F5579A9595783B0F5E6F2B1F4E41E2AE48DB1C76BB201BDF2EEE2AB81D7F27721515B172567B7924D09E02267F51F221CD76415C4EE9D05B0B743689D5E2F68E7C27801D356A79A84F7AA30F3BB3384C213CDE95671E1F1AA44F19C99A9DBA2BADB771B54C2424AC5C0CA4D9F83CEF2B7C43F624290F8DCD13C4A96A6D17B9AABAF16B7D1F25E771549E8BF5F3EE171FF3CEC67ED893CB8A96AFE5E6B638E4541B68355893687EBA270A97439A63E1EE868461C9D466A2332B766322CCA4575B9F268FE9C6B0297F71BFAFBCA5A847FA0F76FB02B955643771E5FAC7EB3E8FA06BEC92647AC53EDD5487A8F3C6D4D167F289CDA3FD2BC99A48E7316D262B8514ED64338A0DA5D3A52D6532593435239D0718794BA96CD7DDDFBDDC035AAECC2FA86251854C0D12987558390C4CCE6B79A7C07CB2CEC0994D3BD196FB73B9C4A799064D4DA6074507B81CB2A6F3D92C1BCDDE8A8E751D23E983FAEEC6B027EDC2AE98C86C467122D3E9D289CC643276C070D8574823E1DBACFCFD23CDBE295BDCE752C045C82363B698D194D19B87DD4D1A0F83AF60359F5564369B43CA6E2E9B29C3A56F063AFAFCCB20FABBB451EE4F7A5B4ACD12EB70F3D41078BA7F32B87F72B79F0F94DEEA2C0F103AF1E6BACB9348056117E802E77E93F3E000E981C813ACA784B517331D941A6CB3275B455D7E546E0DDEFB45CF7F7D4E13D6E84C219BBAE0C1C371FE0EE70C9709DC5179CC3358A75CE18E46CE143B1D8AA85DA1186EA968791DD7570C0541D96898D132574B759CFBF5C1BEEC47C06BFA883768E7F564E97D334CB639E0998D550F88E594DA013EBB6C67AC2C63BA775781D2B69F2C0DA33E3387558BFE33E56D3540D5A4305701556590DA9F5F1F307D63F2DBF7C510B4EE3A2DA3AED9E2B62DF7E51057D2E4268AEB4D1CAE8B7059652FA122988E82E57C1C6AB9C65BEF23183A8F7927E1D18BD597CDE67349A95A750C96CDF32F1797E1F89762E77F19A11A5374726641D6E864E83CA173CA93AACA75BAE9A6BA7ABB4CD9C4D8350863463092A166B941DD112F61A40CAF6C0C2E6B86129ADB9A2EA36CF992E4B65CA0C7BB6D42355FB5BDB0146AEEE2EC49906104994603ECEB4D9BE2958EE175A9F90C445D980AF3D505C3ED85A603865B0A4F189E7231B6B13CAA96EABC08767BE7F57B1AF3CDA92CE4A67BCD34E5CBA2A9DE924CF8BEE011DBB84FF80AECC9427DC207600B7AD2D619A478796E62F60ECADFEB1B8F870D739B30C90143663AE6CB5AA6A32F319AA193754DF46642D358ED5AEF19AB4C4F1BC6F59BCF183FB899C34CC7C8005176EB0F5A29BACC20CA46C27A1AF5349EE612662E2DDA38C5CEEA5422EFE5E6A9569065CC41AD42A2D0049EC03A83DA7ECC5D9AE62998970DE7DF81F78731DE727D4E7F708BF2AF51521FD2CED3647BC8B2CAE2F37DDDE6BA434C3F25FA549F063D4E0BBBB1B9BB6C719758C55BC9CA8AB2959FCCB2DC9364F4700FE07B8B599D992C838882BE64D9F8A2561813237CDA204EA4F284BFC9425D6A403CD61273B8F1190CD273A4CD3166AEEF60BB50E05A55505EABF9FD5B10C7247BD890A2312AD2CCED36FB33B6583B99BBC4F312F3C3343E7D9AC61FC87D017A2B9AC944AAF6EAE910B40520F324CFE7F29058051A820352DA85482901E9871E1DF5E5BC44EE24D74DA3C47891478B0466C2F55040382580F9649A1538B30797E0E89EE06CEA754550FDF36143CF0118DD4DA19CBA8B5C7654F7F8323E2E461AF72E4EEFB82A024F921B23B95B5F3AA85747284AD0DB2763811DE493C4F8E5B7F6D16D12543898A06AC52393968DEA971755A66B2A2FF0D282CD227F59C1E51BC3277A5B8942CFC4E5D0B456504759C99BD7DB6DE5C6D75EDE30049EE4CD2CF2A61D030FEF5E7D892F9F62C4506AB4CC00A72203D66BBEC0302155F98469A9CC6CFEBE4E25FAC0AA58D127C9826BB544F4C9A5C96127EE5DBA3B86CBFC6D1CDCE6FDB05AEF651A82FEE3F7FD9E84248B1F4A5CD2506707AA8980D509CBEF4114373EAA6B25D4D9F17361649902BD97BE36FBA93A7B992B0ADB17606D09D11C9729F1B66C1009BB72D586A12DF7B33882CD5829C64FEA56CF7D2425A49730A63779FD040737A09FE2204AAEC87D811DD2F741F62D4C7F24B201450D8CDC55403F328DA8F4E1F66131B3ED435A0CEF79B103F4B1B8AB4E79B8C179476E83EAD20337D7BA27ADD4F4FF053398AFF33CDD46356FBBD5715BCAFCEBF38CD41B8680DFE65C24E15123A4F98C83141F6E76EB3CE5C25A0E42B42FD95ECAE9B3E37F16BAA220DADF2F0D44A9B671944F8FF92DD8C7A4117347AFB7551FCF8ECF837C1B84E2CA5D722764BF94BB369235F59C970B5D899E2829C42D5E946CA37D106BDACF95C36E0EAB86F555F0299D979642332098BA87529256F495716CD371E9D50905310CF21A55881A1F9CFEC3196FAC6E84A2D75C99B0E49E3F7B264E5D07B030954F87138687986A07E5EC2CE8A0700D04331E4694CB07A1442E405450E10923E132BF4C92347C0AA849060353757F053D0BDA2E76E5E6B5F5BA735D1D609498037343C8A3339A600FAE607281A56CC6147852321AD380AECC7CC8AA5BDC762197228AC90521697451236B0B003A33545B210FE4C71488033B8FA9F82A5D00CEE880D9EFA3DBE67EABBB80578EB7AA241E8F7289A6AF04009AB23BA3493D343F26C323866F98C65497BB14A90583552329954597203ED50D7484BA4F198BE2E422806E248D5F87BB28995720D37DB96E94375A702BCA40A8A6B39B0863553500369B1C95EA7F5C40227A3F051211DC419D9F17216981CE18C1CF0277A34855B0694AAC4E07D2B9D1B942588AE1B8DB90F2B2F1971580E029E435397A4B2B429EBE39DA063C814CD3345C911781F8029AD9E139A3A86C2ADEF4F109346D144352095C98486EF10D0238D566E942788C22B5247C994260493880A99A7E303A8B98EADA4E4718D38D34186E4C0B3FC4C4038394D16A6EAA91A31D8E156D99124C009331D5F3710067051513BA4137F6706CB8B9A51A1C5D4268151DCE6554F10671694A5842FC5893A0537AF6D78001E5B25F8F5850549955269F1E4C5C125DBD7ED9C7ECBA1C7A07BE5C30DBE4B9B3187E763127BB991805063D030310D85CB0E16B5918BF2CA6B7221E825B9FCC2BC60B96F92FE64D3A34E1AA851A5987F6CCB37CB59FD81812065083CA2BD0C6C5AA70053A58FB6C8243E2035AD31F9D4B76A117DCD365E30392CE03B5AC42EF6A0AD8E9BB51EBF582D52FB7562C4F951D99508E2A07704DA700DE8DA80648529FA202847ACFECC66095C6B25E0F4C255D9810A092815A233495E67940DE5141398F719EA21133606A5DE69EDD0C68FFDFF851D529BBE8CCCB51C131AD928BC5E6F1D4A8CA37883F532ADF204EAC49B689EE523563AFF09D2A8C3FE522D658CEC95DAFDAA8F32CA629E71551372524BE0F679FAA32D78DD47B16D62BEDA8D315E6D2941316E6C79AA62C6A1FA2DB80D85C05CEBAE5986BAFB1CE4D06A32BC7CA30B090A54DC7B8520D6E297E81F02BD5947C9B10AA4AAE184838C183DB6290DCC6DC092BEF1AA6264A5CD9896C95F85A47375A124DC980669998BC41C527308003ABB5BB7AB467A6CCC1A9AC1B5A6FA7ECA8F3BE87F1FCD2B94B5DFC7B514D07A6909A9AC15AFE5A2FBC5E307FC762FE7CC5E0B5D512CCFBB53D9E0A694ACE601AF181FC6898B620C019206D712FA4E6B1E497F36516201A217011F0AB3AA2815D9D050B37DD2ACB8720A0363C69ECF116B7AEABFA873217D5F493CFBD841926B449B22351D80DFB9C6932164D36D964FC58FEEE82B67646E3525508F4922287C1482855B67036B062F836056631DCC1B4634936EFB0FF72CD215EE3CC5C387FB251078C75066A77E88B3FCD617A31A1224C3978C8BDF6CC9EA8A09E40AA0E135C41FAA969100DD56CA7CB5926D615FD9B0BF68AC15EAB2A5819D7C0048ED22007A34E02599884554A774967E642BB644431CD8122C3CC6B4FCEF9B6D6604CEEE81AB0D86E9CDC1BE358EA247B2DD895756042BCCAC609D304D62FFD02D089BD52964674406213FB88010E08617C116CC00FA5277D597B716EF5875673C129F0F316E5927FF19317D38B29663066D80CA6311DAD6296B92C8BA260842678A5F18B589395C68B1E46D38AD9C0B6F0E5E2A28E8E519629CA12BDBFF88BD6637EF5BD8E182078E72F47B275D0DF79F0E7E151D1DD9042542BE5C747177D4C0E489B28208D2525A834457282EA0F415246475B78F0620CF68CF226AD21547BDE8668B47ECB35C5693FAB1015D60FAB8698DADBA2485CED9050539970A50CD500DC3BA3C9AAE9690909362E103DC08A46431634EF110983BA151C6905392C89CE85929450E78808478EF16222A5497B0041D3ED9EB4AAE90E0F5F7184E937B5CA36734F7A71D49947800AE2DCEB4D23E27AB25882CDEB1D2517FAE751388ACD531A29B5E6190A8E54FFEC40418F7AD6A093C5EC1B024026B3B6F73A0952DF7D0342A3BE8946AC4E8CB9133C02905D9A86321B2F1B22CA07E2C68D05ADE1510C07ABA3C34EF46A87A49CE3CD0656438ED9AC41E4B8FD30478EDA4E710B381D1CE788CA46AFE5D2003ACCC65A1A42A7EFCBB06710B68A0A32DDC69C22C3EC61F82321DB57341FDAA8B5321600CA5FA0D9AC9AD7A6E3ACCA96A2D06EFA9C7BCB871E017AAC8C4EC2B459169F0437564A5223B2008E87013002113883E9833A7406D5136EF3AB608B3A58C608CC61433A004C51C47C605A0E477DD0B5584102E8B4928FB69D571F0E240CC10727107B880A4F60CE385440028AACE610353E73E57033F0A66FC6071760E2DCE04FC86095537680B3681FEE4CD7315EDC8D7889F1D74E11844EDF63700ECB32435EF96392863B3ED82275640DB006E7F49AE994D6ED35D53148A9A1E097D6D1F5082B27640921E79722B7BC5BF24210CF608D8D826B0AF2E3F3AD777F2D6716EC211BEA82E0235B640B8A15826B6B914EDF6C6F8C6074C1726648FD36831D813C37DB3205F2D54C6FF91975B737B6B0CA38395FE4AE87C1CE80CE876D3903FA0C1689B15DF1C521A5835B39C3F07E71E13B7184675C4B76A21CE182DCA5B4AB537097D5765B715A6E6D69CA19D0FAD1C3B260EC33778E91615CE5E24642EE5D57C705D0BFAED1BE4E477419AC3493211AEFBBF8FEBBCD74F37A0CA4951F0EC37731282683458DFA0F5150B040B85932E33658DBD4E896B87A95731CE31B16EAB9C63BACD85FFE664DCF5C8D3F586915A3B1532F24109E63F53DD50B05575E4E2A0E7877A472EE291D9742BD92B92E15FB43DD94EA9925F3563A099B649A7A281BBA27322DBD0B77C656D377DC676FA6E58722269FFE1C4367773E1431C4E47069FBE08B35A2C741397B34DE09A15EC9FD138A3DA32FFBF5FC927B241C758AF18E08156052B92C041120715A680B2889AF41FABE95EB8B372669041052F2E8440E8E0D13C918D8C71B52A5890789DA2B9C6FB5A60A3CFE61A3742E86E4A4D22799B6BF32AF64BEF92AF34336023C911EC8507715604993EB0588C028F7186045965A1ECB4B5CC85399E4E656EBD44CB85655B935E340245A9969AE6C559ECCC632A500FD6661EEB96DAEB73DDF6A4F7E0F0B387EC2B0CA8847BE9833C1FD2BE58848C606C04D91D86CD6519179B759D744B4346BDAE6A7A3E24B09499FD55E62C4E64B1DC4987342EAC98523E5F90A4CE92307B2E043FBD461BA88F2AA83BCE8C3139E8685ED7A083B69916FC1104E5DA03D91DAAD8BB84C7326C8FA4D97DA91CB788701ADA310242BA1A2669D36BB7972633054D7747B2EA845BC5B0A24DB95DE2CB46C90F9B3F0CD6E99078BF1602D3C5155DDC2A85EB3C2F72092F7ACD0BD476BE08FB95D91BC601D9D4B081D80F2F9BEA2378873BF398B263AEB2B5FEA03ACC2BFEC67FA867ADB4FF5907F3AA2601BEA35FF08F092BD13C7324D3909758FCA3DB2CACF24ACBC1A5485FB67CE7DDAAB93CDF68EEC82F6C3AB9332CB96EC8B4310BF4F4312E75DC2FB60BFAF9E480D25DB2F479B4A89506EC8FE65737C74BF8B93FCECF8AE28F62F4F4EF29A74FE6C176DB3344FBF16CFB6E9EE2408D39317CF9FFFDBC9E9E9C9AEA171B265E633FF28BBAFA948B3726671A965D5654BDF46595EBC098AE026A8DEF39D873B215BFD817AD4CD72AF676F571BF06E5B1CBCEECD5457A8FABB75C476F88D14EDEBB1670AC3BA81936FCBCEED4852D4FD24F0A6582C5D96DF6C8338C8BAC7F4F463FEF3343EEC12C5EB7E79F9CD5D707A7577D8DDECB3A8923634293E0D4F5546D18EDAE670F35F64CB37AEFB68D2D7342B60624C0A9EE2659E1FAA3D144DABFB66D82E88149380A77771BF8FBA4D354D8EFE2E527B75C201949F0627C23CE004133FB750330F71CA44CE3EDE1B82F90CD452186B16F2DE4869523A4FA572AABD7B5E9A9CD467EFAC08301BF6CBBCFAFBE3D77F14C6FF9F2C477DBAA1AE5F9EBD0EC38CE439373D991483614EB669F235CA76249413976632835352FE25E2A9F96ACA83384E7F9010E24197622288DB9D142F8BFBCF06AD4B829B98D4DE7BDE06DB122C95A17EF5C2740BC9545D667CBD1FD2729E3FB4E78A4F87FC8EEF0D98C190EBE70D0AEAC65DA5DF08DF1D491E7C2D9F823CFF916661750155003540E92ED48745ED4DFDEC415D179FDB404467A432F9FCBDD872E299FA8EA7F62EC88BB725AB49F82EBD8D12812A948EA74E953C6F0E91346D31D5FF5260B26ED3772316E25B799F23E711568ECFB428D2AFDFEDCF2352F70B98E388A2F0588BA3E98E457AFEA81DAA30270FC0C58A8AC297203E7024DA4F06AD20499014979C141FBEE229BD21F9368BF6E2F2C324182C97F5C8161CA387AFC662119489C66716C2EF88BA8F36679FAB68FB0DA247279A09EC1288A128A69BAF8B111CADC3106B99D138613317179272A39D987AE9241E98D8243CCDC1BD324D4EEE74594EC997C6A2DD0937DE4C596A5CD262F0C77AD6B08621E3C5CF1C8DEAE26381F2D734E408345F0C766EADA316A1254C82011089A0D423F328DFAE52B05FD4E7C540D8D46988C53E57E7E9C570DFAB27273D74921F3421617CA07403F915EECAC30B34EE6C8A81DC579F5F9D8EAE2DF3C4732493B01898028674D6D814BD8E9A03124143BAF2AA20E8803F362425876B45B44039C5CBBC462EAF666A3F3EE1D809C75E00EC84DCF9212BC1EB24609D09008069BB350E44F7C4E67040D0186B07E943ED70996FEE828CF1D0CC8EBF906C42FB5346BE46F73CC9EEEB6230857C97814415C2E810812B1495B190C5ABA10C1550E98F244E8310502773492658FA42B2E86B245EDA0CDF4DA8BD4BB7DF445ADDD7A521D31D8DF6089C0E7500E40592B23C269BA9FD4316DDDE15FC26AAFF3CBD42D3B7CAF633894990930F69C16B49D99479E7F4C57DEDDEA77B40FB7B16F31A58311D4FFDCF417EF73ABE4DB3A8B8DBB184B924339A222923C9B34D13A1A3FD4723095619C01482046BBF9A53DA14D5E5344CAF4BB3A04A76A5807E2121DB25DAB6564D9DCD62A8BEDF87E2CC6612CCE85D841148AEFB6E40AD7123228088FE6E80E828E4032930F016934DFA9DDC1E6A97086CAFBBAF062BC3E1268E441B07EAB3F12AF3362AB111FD4D3002E0120DE86669A5E614C685FE6E22C1F7691E554E3E04825C920D4DF1CA934F3392899F4910BE27974923AE0501D9249B2A25AACBBC167AAFB795FD7190F0A7134536139DF56E1764DC36A3FF68A0B30E6EB9B5B6F96240212A7801DC7E323813922C177611FD4713FBA26C17C4E51408418A40B2C93D432592DF91EF44BC8FE3D38CE5627D520525639B624CB1117C32794BA59A50CEC5D5A0FD6678BECE9A0D1D70C6EE530C6E94E2A0284842C2DEDD2C73AD24A45A50669D9282E4557E4B5175B0AE7EC03A54DE805475BC8F92F3382A8F4BE0B41053A73767016D1B2D2C1B5B0E95BBA8E290CBCE6154EADC06CC17BB1B1286246CE7A5B8C481198CA5C1C5FDBEB28515861E485E9A1641EED7DA4C97D04637B3D628C8CA8FAC579061D8F49A28E0AD9A9B2F4B1B6D8D836AB321A782CF590FBB8AC63A86DE4D2DDAAE099B3DD982DBB326C164C3593D02A4C2B7B17B4F2E7181F01C7CAF7A80671FC3D0059F7222630154F00F2C10847318ECB6EB259ADB65B7DF4C8CCBE4AEB8457B0D5D66FB7AB57559F3BF8D2F0CB35E127C1843F7A3A0D2A5530C764D75B8E13E562DB36962939636D579A7CEAEB35DE98E1A3FDF3564C69AF1E75F369BCF411339921944EABB09B58BCB30E72935DF0CA8FC05A0F217199579D1C4395E760513ABDDB4069386CC3AF6377E9F2D6D2A03224167377C35535C021A182BD5CB79BADB0976AEFDC78582DD1BCC9D013E1DB4D772CE1A1C77BB8E511727DA7A8CA404D6217EBC3DE88A76242F82DD9ED7A5F79FA7D5C983CA540B1DAA4FCB05EF8FC43CDD6BFBBDDD5CC39D9C6F4B8DBFCFBB3377CB8C31EC45C6BA6BF66DA16073BF3EEFAA2B0B3161B6E2D67700D6CB2D5C7A1D6B2D7E77E5EBA6795EBC50C1315C41332855AD91A32031167C56AA2DD60601C1BE54A6E9D8BC5856975FC7A4F765AEEAD31C0B5E1B6D16C5651A7C4AEECFAD2ECE3FA73F40A94B7F5FCCCC95F98CC7BECD298BDB3CC7018BFDDDEA28C00016D6632246B7B0F4A0A62332D678F9D230B8EF9E36871BC9019C4D19E735D84C686C71547EAFAE391C80C812B200A18EC0680283DC170A736B2079318307FA5C77DD4BD3D4ECB7D36A2A23EFCD467EAED4455D009D2D58789BE1BCDE0B64A1741315E5F014BD72C1CC6B2AF954E376AFE59D3AE746DE759E34745C6C5BE0F263CD8DD6DE54B47DE83F1BD302ED33E814132C05B9B0876EBF192CE1D16D526EE233FE8273F8BC1834722EE7ADD1C8D0B140A3A6FC8AD0D8F64406493A794EA42F11A36C48028D7787EB7A15CA8FD12E1CFA025A470DB2308182B0806BE8423B88FCC36158A00B0E6DC5CCBE450E8D6D2346D836567A4692B78F8F3D618C05281C92060DAA2208070B4844C86B71C50476BB68880A4583D78B0BCA5CF79A3702D501445DD6DC3E58CE73554DAECC17E879C48AB2DD8E28376F66797008A3AA2D4797F987431C9F1D7F0D62FED11B861D53808F0D54670F4496CE54A084A2CFF90228437B22B04291FB160E5C1D9BC600316DEA6E025AB6DC5820A56B710525436B241032752C13743A368C0132D0F0DE006C70F9B14007D5E60A3E90E6482004EB5A2618B16CF1054A46C77EDD1B599B3C15A04A619E04A8ADA8E5030AD7E97EF6ACC879C49DA4998EB3A5A533D301A33787E966920E1F400153FB1F398705E29E44914FE923B671D1E2664C09C3ECE15A5BDBB052BB99AA2F84B2483DC6A9A52283AB6F451A0DBEE52B546D88EA3BA0A326CA50B8B856350A9D406D948F50F52BD19B824DF723721D268333C22A545F37A6592A2B9F2E0B64BC039958012C1DC8B88EB8C54CD4B5C871202B22B3A93DB9E8E8988B632A3BD2CBABEA8801C650779C132D55AF4721302CFB32B722F2EEEB712344F2E6B3F4177AED97FE771FC9BB8DA2CD84F7AE395305EB3EEA85221056BBC9727C5432E17B145621B5370F794176CFAA0CCF367F8D1B7F584386F741127D25791361F1ECF8C5F3D317C747AFE328C89BD8EB6DC0F097DBDAA741902469D146664744103FFDB98A204EC2DD095FDC3C0E794525CF4326A01D7553DE19BC104904E057E558F288E890F2997C3D92A2EAD5095FF21584CDAA0D67C77570EC7A4AFF46AAA373ED25A4A8D4897484B60A7D952FD41E81274AFA7C28EFA6AAEF41B6BD0BB2E3A3F7C1FD3B92DC167767C7BF3C3726AE20CC8FD9CBCB2424F767C7FFA72EFAF2E8F27F5FD3C19FAF07523F9553ECF724FAEBA1CC75951DC8D1FF659AF9E20F7F34674217B5A86964D2B6F21F77C1FD3FD1C48A4C34B114194A470CF740B00BF3EDAB6D1EE9D171C31B72D56BB422DA9117951020DB28AF67F3BFAA09D34606CA09A80CC3BDE249C83BC2A3AA42CC133E50FA354BAEDD51FD74F4312B45F2CBA3E7D004326C70BF4B1BB5A5A7C62D3542D2A3810F1BB79B9DD72AD98899E2D2F0E09EABE983852BE8FEF1174BD67461C31BDA3751614C878A176E4F441B12DC9E341801DCA1A59268DF8AC1F9F985E9984301BF47AE808FF2CD2E5AA6F46957396E94A0A8DE6E14C558DE9400F223436571A0572C491DD6356689B558521B17B57EA55FFB26082FAA712DEDE354FB5D03186F247E490F80F04BB77F4FCB4F560930CFD3DDFE50173015C85DC06B37A920C4B9EE96885B4E2E60855613E3DAA45168F102C48D5EB164E1624F9BC917A6B0939419DEF998B5A02B27AB1C75FAF47828E6825C6B28BAAC72F280D12B4663137DDAC33830D1A7CD204515754215193637365B5D9FA8A462569BF1A22F6830B9D108C6C783C6215A1FF9590F70888619CB440A4E82910D3B6DD614BAAC5323B4272E8C6413C932AF26E56BB50BC634219E71C0D206C785C1288D80AB87A123067D0290ED88F15CE8CB3A35A20F016C2F4F1F0386A702EFE340AE356C67C22C1A109A70CF2BDEFFE91503A72FFE643112420069178D6B173ADAFBB86A2D9E563CB21AB58CCDB072315FB17A44684887F8CD2EC0E822378F058C470306990D9A999886A9386E747B37C61E8E5D9369E1141A4AAB86B36E7F3D10F43655A170CC1EDAC7F9E65309AAE736B4BDEBBA7BAFD65E0C39BA80CD4EC28F09D2EC81521733D957A33C10643C1B8F3DA9E9C8CC8EB77894DF721F93458CC7ECC2D22E12B36A86184F3ACA0DE8D8E324B8D096DD5DE0A8515EE1BD4872C627BC578ADA9BC1D3E736C2B28BD40C620AD742A9577917756CEB50DE873AB6F62BEF83501305C2EFDD61EFC5D5EF9D2F10AD195D014EDFCE066D06F71806D2B23E2C7A19253148B317A2B9A3D50F1B9FD99E8E1891D9C7FD0D1C87D92765C689BF0FCA62E06505C07F3106F83836203803086713B396D994C769FB33809B552A5BDA852D602C67FB6E01C19B15F8F983F9DA0A3CF3B03A700F04F0DC33557540E199D7AFF0B0E7B9F32D4E20DAB022849CE9B0C9C22B3F0D9DC33D8291E6D46087D7C47AF6BB79148278B890B7401FEC6D6542F82170E10A083818B3FDC2D30562B6A7A08DB56CBD1907232B5B531362297BA0F4517863E45FC5C0C556F662A706F9A933342F0168B8ECA7F42EC8DCDAC7D373DB12F39EABCCDAC69777698BDC7191D57A25D0196F97A7887EBDE21D031D44DB5AC074F1B37D5C43FDC59294E9702AE24FAF783867DF00E29EF3E096FA3EF0B5FD622FD563395F6AF511B0A7DC3482DED0560C57E8A467686A61CA4230E4F38A5938FB8C9F4FD547C5A7761435A3DC4DF85431FBB5F5F06D82E1F382DFF3FDEF0AAE283D9BAFFCFDDC289A1AABCC6D5CE3F7F25D69DF61BA88C82D10BC2EBEF5B5C3D3CA3BAF9E1C7D8B3EC2A94BE2C775C528F0AA449682CC41453292564411DF79C5A339FB9C969AE18E21D15126087E56ABD519AD7AB9FDA7E34C3754B2F4C7F7EE8B9BE61DFF0247703FBAE2093AF6DDB02ED8F28A5937A35F16AD9EC7D2E1A095AF13FBC75D3E81A60AA8BC629001019947DF54CAA319AF98918B7D76C48451366B0D55D4A90950C865B3968814BCF91C68E2347B52E48CFC5A9B31927B3453870A53EBB88B66A3C89ADE9A0F651DD1DE045AD6400AB7660E716CCDC9A161A5885AFC042B02C63CB6C2D640C0D1466B0128C723D361C38008698C032832A6865850129A418F6EB846D3554FA4B1C0BB3FCF6E096CC7181D7CC4E3282F76D7B7D89146078DC34E6C79A03731B73CDE1A6642F73599CEE2B6A0D380D24D37AB7F2839E1043E7D9AC18F6E069B47D11A7971C6E165A56BF5428069A489B5555DCB03600168814353E94779467D6ADF66434C35C5C6F68F993F9E072DB2FE5321A6BAF34BE569F95AEACEFD22098F2AF6331EDFDBA657519C9ED19FEBE806FB38DA96359F1D8B81293F26CDB1EBA871DD5BD20CF26D108AB0A9A26DC9DAD1B886A69BD07E616BFF6781683966246B9A7A5E0E7629054ADE8B031C25DB681FC462AFB9AC583454DDE9A9F229DD5BBE82E91CA626CEA7B458674F9A63AD8E0D4C1C320C766A4DA41436F529861EADE6033B58CF9F3D3B15C66BAE2117CF5DB30FF6A02E9E659C07AC5D038752C3B11E492C188BA7B5A1C550122962694E8018DA717AEFEC7B5419C1B86AA769B109EBC680DC1FBDA442C645FB2C48A863B1B64DE63758CB901C4FC8812AA4DCD9CF871B03EFF5C658928B1265AD3461754684F85A3666F0DC9734401F2C6029605AA6641A05886B935CCE2864A33BCC823FBA0FADE7732DF058CD846CB47313C126066AA0C902A9A32047A1721909429A0015925A9721BA00E8CC841909581E214A56040F41DDDEA863F2EB4D7AC884A005C3B08AE10DE8D105522759F0603B0765CBA0E5D81304352120468221C2D6435233F2FE68565836CF8E16B9DB5A16F826DB5FD9E36DDE3D15E4700827FBA0EB5A7A98C1F4492028B302D2B46E44186AAFB6474225CA1E4A5237DA7C6266982E58162E0F8893C94317E42D42225E770E3C34830B0DE6B4F06A1DD002EDE852C6946953A20972B52BA9917ECA3B2B8C2827CE72A3002A0F73ED467F37BAC73100A7AF2B7E593FC78504EE969FB06EB8670504E3267EF1C28572930CB4854E7D2C4246E6185A52EB52040DE551587027EA0366388143FB3586F132243F22C0C0BE9C25D5C20E6D978A1D8539AED7B3E0BAF035D749CE156BA2A3E2A5E08E76333CAA89CB529034F551CC143AA2DFE8E541C5E3FA36FA36CA126DFE2C1E56B0342E0A6608CFDFC238B3998191E632AC5272E919B32C54C9BDCECF0935E6E962F7AC0EAB75079DACD028E0324C29EB546F32752DE5B38F895F850FF871F18B7EB42A6986EA95EAFC00D6E8E3D1807802AD02B46B82ABFC99F49C785DCDD6D14DD4AF57AD622E9E97A28A6B830CAC07625D5404A0217DD263811518016225801AFDF9CFDC78985A2762028665DC4AB7FFAF066605B74755FD502B9AEF8F45A4D4BD59933CE9FD95FB5CA270A24570954E93A4121F0B36242EEFD500995F3971CD38785FB2A4613DD133AFDBD994D5234AE1735F0DA7F9D134D5C665427B97A9B72AEBD9A330778C5809B25C33E2093135F3E5F2E3300B6E236085954BD1C768C67E8A311F3E7D9C08D5F8057C2C905D9F49FBACA09CDBA81D87C279D766F12919D099C7F2DE5127D371051EF36B485B39BEE8C79120A271D7910B5CFED784EAB9B605DE70BDA82D8218856AB90EE38088597C4BF8E475EF017421C224D52ECC0D942F873D4BF79232192CEC9CA37C203F9A820B82C6329D38CDE540650600A1AA5B046CAA966BCF0C73AF4ED55F42FDCDC775230676912EA96C01EB4F8316CD7E5C182D68A41E115AD00338295AAA8C33A3A5FA877AB7BBCC55896BA4D00CC503EE354A1BD337DEF30A1DAAB57834593CC57FB4D89AFA49BF0DC496F4B49F8E44DB473B5DA4DC021A0C69BDD8F475CB2FA84BC823DACC7103207469F55CAB50712E078633DF6B1AA37251FA4A08A05C88E52729A884DFE2A52014737BDE77BAD5E80FF167978CAF361032F854B74979149882E23D4B6A64430E2F0047582B35D3A15C9CD5A2F958CD605BD60ECA7CE060229873E1B6172969D890EB7423B894754B1A456479498D40ACF5E520CA7DD9928B962744A81131E7DAC30423A84666205847F9AAB555D757A9DF6B3939562CE31DAC3C168F533804EF28B9A8038D96658AB204C93AED561A92B75196176F8222B80984A0E64DA90D29446DD8F1D1451FBB14D2616EB67764179C1D87376939E44D085446432AE285AD4BD0410AF50939A03A05CD2CA65E4965F21AF464E9108722E7A84490717DBABEA236E8A65047FB1D225F27E929B391D6840AD864A81E3A87BE3A75B824A17A7576A839EA5053BAE6012626429B803C5043846C26B52BABD5D5A7AF0830BA15EA03F240D50AD9F4B583BA43A17E3017D402C88213DB0679BDCABAD0F43B67DAB25ABA74455D9D4B71648DB4775D59AD741E45CD8CD3617CF583233145FD43267503BA7CE80670BED2644DE0B2291A41E7346004E76045D60E2E9BA21D4C4ED366681B80A91A5D69EFEF4156699F41516993C780E38D7F0059954DAAA2BEC63B02B22EEAD1B9AC422A8BA2D63E17620FC03E4D16F7026C3AB827601F686B9789DA84455C19EACFE0625059EBA03663BC7D2FB835E333C9366A4C3E0C627E0BE29864D5874A3B0DD4CE67806A66F3A0A1C3DE0CC8D0C3E6520088CE6820A59B73BC5C4037E94AD95C65D1D7C8690E841AB974A846268BC5E646B2F997E4436D729A3032363B1D4963A439913B1EB306293DD1A1962AF3D5A32967BA9400CF01B1BC3B45338F7B4A698E30DC8E16570C853FB3CDEE608E088A59B93CEF52CDCE75B9E62097EB4F6E6285940A843B165FD3076D2A1B7D44A6F3F07A19CA8E4B38CE9755B19F05AD0EE554843EA3D705DB2FBC4E9EED0ABA9BCD9DB8B487C09539A0F1A35AD77C98A54303479B474A40A7F82C3E3B663CCE165DA4D522B58256D65138A3CFEE422A9EBA249BE0DCE55A033D6882C4AEB21956DB45B5524AD26D9358DA4EACC068D86A4AEA8CE3B34A0E118370F18F8459AA28E50097D041CD55573692DE094B2F4343AA20ADE900A963B006CB93E99821E18287EEEBA24C039C300A4CCD7448AA17AEFB05A42A98A33C86C1F4A0E9EA9365ED33221396412F8F1C65CE7C8C01CE1B7A34E90BC9BBAA3817D51D05D315ACD39CA76534C7679F1C59FA423EB1B50806F55150E5EC8003A5425D819A8E61017BF34297EE52BC7597B9AE967759F14A45F696883E9CD0DFF5DDC7B0CDA1CBECD590BCCFF2B8953EC75ABCF3A229D0A9BE18A00C9C28E7073EDEA267F6F077721C7F86E42918C48A7B2B66A9DE5C8CBFFECCCD5026EC1D8E81F248797E569CE5B0C46C166AC2C23DA66908DF7CA3B80416D57618BC8AA7BBCC6558260BD521B7E4FC3308D5057515B41FA03BCB65D0330F71E923A5CF671F99B9DAEDB4411829052B14DD7E1C0CD5CB42447C239F32D00DD5F66CE1A3EBC819A28CC3E393159C910C5DBC4FF2DD7DD91D04946D844DC2945D66839E288E284CBED1469B364EA2CB36DF7D755A8CD121EFB8269E87CFCE0BD65234012AD1DBD8F3764EF2D15785ACF0C202D06AABB94C6553BC755E33C9479BDDE3AA1CE0000148851B7E7C7D1F1527648AC4873D9245180FF873E8764F7544DB4C93B0D44CCD8BF1D03E8EBE7732A6997908475DBCA808CC7473A5833AAEC054CC36B9E55211986EEA2F95E1A20DB274719565F539D9E576D37D793ED9BF158791F1C628361BD39B2000EE6E31BD1FABDBE3191B880E5A65FD6432F906396DABDA976A3EFAECA25C506A5C8FB2C28B6F2CD4D0E9BB283E91957453EDE5D0C368CA6E34853477B35295F346C8C614EDECD1EB15EDB44C699746D8CFA07C278BF04BE867EF0ABD87A177116CFA28EC00770F38CE4045273F332D8085BCEB3624FB941EDFD68C2FC1698BEAA24DE5DFC5D7F52CFDFC8ABD336B523C771CA1A151BAD1F2D18909F5314AE74F000BF0CEA21CC71F7C7E5717E552C661010C7C9463A3D574DCD2338FCE84D9D2C78F23DB2CCDA2156CAB7C9055447A37357DDAAB93E61958FBA1FC59A459391FDFA72189F3FAEBAB93CFE550453BD2FC7A43F25A97D695286926A476813610EDF25C265FD3CE350FD7A22E4B97DCBBDF2882302882D7D5F62BD81665F296E479FD00F84B101F48F59CE4868497C9C743B13F146597CBC357CC1CB32B2F3FAAFA5F9D086D7EF5715FF3D44717CA66466517C8C7E4D74314877DBBDF06317FCB2C2351B90FFA8D547619F55856DB1072FBD053FA902648422DFB7AAF475764B78FABADF5C764137C27366D2BE1FB8EDC06DB87F2FBF728ACB02C23A21F0896EDAFDE44C16D16ECF296C650BEFC596238DCDDFFFBFF000E7A6B7CA2E20200 , N'6.1.3-40302')
```

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <remarks>
+        /// A <see cref="PackageDeprecation"/> with a <see cref="PackageDeprecationStatus"/> of <see cref="PackageDeprecationStatus.NotDeprecated"/> is meaningless, so <see cref="Status"/> must be at least <c>1</c>.
+        /// </remarks>
+        [Range(1, int.MaxValue)]
+        public PackageDeprecationStatus Status { get; set; }
+        
+        /// <summary>
+        /// Gets or sets the alternate package registration entity.
+        /// </summary>
+        public virtual PackageRegistration AlternatePackageRegistration { get; set; }
```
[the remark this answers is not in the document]
scottbommarito: Added back the key properties.

src/NuGet.Services.Entities/PackageDeprecationStatus.cs
scottbommarito: Done.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
scottbommarito: Went back to `WithOptional`.

src/NuGet.Services.Entities/PackageDeprecation.cs
scottbommarito: I was under the impression that given that I did not set up the EF relationship to be `WithMany`, I do not need to worry about the `DeprecatedPackage` being non-unique. Is this incorrect?

I was also thinking of getting rid of the `Key` property and using the `DeprecatedPackageKey` as the primary key of the table. Do you see any problems with that?

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <summary>
+        /// Gets or sets the alternate package entity key.
+        /// </summary>
+        public int? AlternatePackageKey { get; set; }
+
+        /// <summary>
+        /// Gets or sets the user that executed the package deprecation.
+        /// </summary>
+        public virtual User DeprecatedBy { get; set; }
```
cristinamanum: Why these properties are virtual? (DeprecatedBy, AlternatePackageKey ...)

src/NuGet.Services.Entities/PackageDeprecation.cs
joelverhagen: I think this is fine assuming we never want this to become a (package) 1:N (deprecation) relationship.

src/NuGet.Services.Entities/PackageVulnerability.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the primary key for the entity.
+        /// </summary>
+        public int Key { get; set; }
+
+        /// <summary>
+        /// Gets or sets the CVSS rating for this package vulnerability.
+        /// </summary>
+        public int? CVSSRating { get; set; }
```
joelverhagen: From the documentation this either looks like a float or an enum: None, Low, Medium, High, Critical. How does this int map to that? How did we pick an int?

joelverhagen:
When a package or package registration is hard deleted, how will the "alternate" fields react? Do we have to "null" them out manually? Or will EF handle this?

scottbommarito:
@joelverhagen - added code and tests to null `PackageDeprecation`s that recommend a package when that package is deleted

src/NuGet.Services.Entities/Package.cs
```diff
@@ -22,6 +22,7 @@ public Package()
             PackageTypes = new HashSet<PackageType>();
             SupportedFrameworks = new HashSet<PackageFramework>();
             SymbolPackages = new HashSet<SymbolPackage>();
+            RecommendedByDeprecations = new HashSet<PackageDeprecation>();
```
joelverhagen: This name is confusing. `AlternativeTo`?

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -29,6 +29,10 @@ public class PackageDeleteService
                 DELETE pro FROM PackageRegistrationOwners AS pro
                 WHERE pro.[PackageRegistrationKey] = @key
 
+                UPDATE PackageDeprecations
```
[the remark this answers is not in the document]
scottbommarito: I put this in SQL because the entire `PackageRegistration` deletion process is currently done in SQL (in `DeletePackageRegistrationQuery`), and I didn't want to split it between SQL and EF. We could move this code from SQL to EF, but I didn't feel like it was necessary for this change.

src/NuGet.Services.Entities/Package.cs
joelverhagen: `AlternativesOf`/`AlternativesTo` -- why do we need the `Deprecations` suffix?

src/NuGetGallery/Services/PackageDeleteService.cs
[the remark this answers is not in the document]
joelverhagen: Alright, sounds fine.

src/NuGet.Services.Entities/Package.cs
scottbommarito: I like that...will use!

src/NuGet.Services.Entities/PackageVulnerability.cs
joelverhagen: Let's check with @anangaur on this. Perhaps he knows the domain model better.

src/NuGet.Services.Entities/PackageDeprecation.cs
scottbommarito: [It's required for lazy loading.](https://stackoverflow.com/questions/8542864/why-use-virtual-for-class-properties-in-entity-framework-model-definitions)

src/NuGet.Services.Entities/Package.cs
scottbommarito: done

src/NuGet.Services.Entities/PackageVulnerability.cs
scottbommarito: Sent him an email, will update this based on what he prefers.

scottbommarito: @anangaur agrees that a `float` would be best, and that the UI based around enums is likely to change. I have changed it accordingly.

src/NuGet.Services.Entities/PackageVulnerability.cs
```diff
@@ -0,0 +1,38 @@
...
```
joelverhagen: After much investigation, let's go with `decimal` here and `decimal(3, 1)` in SQL. The 3 means three total digits with one after the point. In other words XX.X.

scottbommarito: Done!

src/NuGet.Services.Entities/Package.cs
```diff
@@ -254,7 +255,33 @@ public bool HasReadMe
...
+        
+        /// <summary>
+        /// Gets and sets the deprecation associated with this package.
+        /// </summary>
+#pragma warning disable 0618
+        public PackageDeprecation Deprecation
+        {
+            get => Deprecations.SingleOrDefault();
+
+            set
```
joelverhagen: The setter should only be in one place and can therefore be done in a service against `Deprecations`. I don't think we should try to make sure all of the complexities of adding a deprecation to a package are handled here.

joelverhagen: In other words, I don't think we should have a setter.

scottbommarito: Makes sense. Will fix.

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <remarks>
+        /// A <see cref="PackageDeprecation"/> with a <see cref="PackageDeprecationStatus"/> of <see cref="PackageDeprecationStatus.NotDeprecated"/> is meaningless, so <see cref="Status"/> must be at least <c>1</c>.
+        /// </remarks>
+        [Range(1, int.MaxValue)]
+        public PackageDeprecationStatus Status { get; set; }
+
+        /// <summary>
+        /// Gets or sets the alternate package registration entity.
+        /// </summary>
+        public virtual PackageRegistration AlternatePackageRegistration { get; set; }
```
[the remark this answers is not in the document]
shishirx34: If that is the case, why not use `RecommendedPackageRegistration`?

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        /// <summary>
+        /// Gets or sets the user-provided custom message for this package deprecation.
+        /// </summary>
+        public string CustomMessage { get; set; }
+
+        /// <summary>
+        /// Gets or sets the CVSS rating for this deprecation.
+        /// </summary>
+        public decimal? CVSSRating { get; set; }
```
shishirx34: I see that CVSS score has the range of `0.0 - 10.0` do you think may it'd be wise to put that range limit constraint here? That way no one can put a CVSS score of 100

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        /// <summary>
+        /// Gets or sets the JSON object containing an array of CVE ID's related to this deprecation.
+        /// The JSON object should have a version property that defines the version of its schema.
```
shishirx34: since its a string may be qualify here as the `serialized JSON object ...`?

src/NuGet.Services.Entities/PackageRegistration.cs
```diff
@@ -37,5 +38,7 @@ public PackageRegistration()
         public virtual ICollection<User> RequiredSigners { get; set; }
 
         public int Key { get; set; }
+
+        public virtual ICollection<PackageDeprecation> AlternativeOf { get; set; }
```
shishirx34: can you clarify what this means with a comment?

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -367,6 +367,37 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
...
+                .HasMany(p => p.AlternativeOf)
+                .WithOptional(d => d.AlternatePackageRegistration)
+                .HasForeignKey(d => d.AlternatePackageRegistrationKey)
+                .WillCascadeOnDelete(false);
+
+            modelBuilder.Entity<PackageDeprecation>()
+                .HasOptional(d => d.DeprecatedBy)
+                .WithMany()
+                .HasForeignKey(d => d.DeprecatedByKey)
+                .WillCascadeOnDelete(false);
```
shishirx34: how would a single `PackageDeprecation` have `DeprecatedBy` many `User`?

src/NuGetGallery/Services/PackageDeleteService.cs
```diff
@@ -510,6 +518,15 @@ private async Task TryDeleteReadMeMdFile(Package package)
...
         }
 
+        private void UnlinkPackageDeprecations(Package package)
+        {
+            foreach (var deprecation in package.AlternativeOf.ToList())
+            {
+                package.AlternativeOf.Remove(deprecation);
+                deprecation.AlternatePackage = null;
+            }
+        }
```
shishirx34: why do you need to do unlink here? You have cascade delete set to true, this might not be needed. Have you tried it out in the DB directly? I would test it locally by creating sample relations in SQL and remove package and see if that removes the relation for `AlternatePackage`.

shishirx34: Oh I see this is the parent package that is being deleted not the one that was deprecated. Ignore my comment.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
scottbommarito: `WithMany` here refers to that a `User` can be the `DeprecatedBy` of many deprecations, not that a `PackageDeprecation` can have many `User`s that deprecated it.

src/NuGet.Services.Entities/PackageDeprecation.cs
scottbommarito: agreed, will fix

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        /// <summary>
+        /// Gets or sets the alternate package entity key.
+        /// </summary>
+        public int? AlternatePackageKey { get; set; }
+
+        /// <summary>
+        /// Gets or sets the user that executed the package deprecation.
+        /// </summary>
+        public virtual User DeprecatedBy { get; set; }
```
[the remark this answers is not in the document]
scottbommarito: Sure.

src/NuGet.Services.Entities/Package.cs
```diff
@@ -251,5 +254,35 @@ public bool HasReadMe
...
+            {
+                Deprecations.Clear();
+                Deprecations.Add(value);
+            }
+        }
+#pragma warning restore 0618
+
+        public virtual ICollection<PackageDeprecation> AlternativeOf { get; set; }
```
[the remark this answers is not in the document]
scottbommarito: done.

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,99 @@
...
+        public int PackageKey { get; set; }
+
+        /// <summary>
+        /// Gets or sets the status of this deprecation.
+        /// </summary>
+        /// <remarks>
+        /// A <see cref="PackageDeprecation"/> with a <see cref="PackageDeprecationStatus"/> of <see cref="PackageDeprecationStatus.NotDeprecated"/> is meaningless, so <see cref="Status"/> must be at least <c>1</c>.
+        /// </remarks>
+        [Range(1, int.MaxValue)]
+        public PackageDeprecationStatus Status { get; set; }
```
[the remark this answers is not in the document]
shishirx34: well the default of int happens to be 0 works out due to the structure of the enum, I would prefer to explicitly state it in the model but don't feel too strongly about it.

src/NuGet.Services.Entities/PackageDeprecation.cs
scottbommarito: done!

xavierdecoster:
👏

Issue #6828, opened 2019-01-18
Gallery search results on DEV have the owners right aligned


joelverhagen:
![image](https://user-images.githubusercontent.com/94054/51405997-b2830800-1b0c-11e9-9891-3d0a0c76e2ce.png)

Should be left aligned.

This has not made it to PROD yet so we should fix it soon.

![image](https://user-images.githubusercontent.com/94054/51406018-c29ae780-1b0c-11e9-96ab-59b254814187.png)

scottbommarito:
This was requested by @anangaur as part of the "manage package" page change. We discussed this offline with him, and we have changed his mind.

We should
- left-align the owners list
- right-align the "edit" button

PR #6830, opened 2019-01-18
Gallery search results should not have owners right aligned

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6828

image

### DIFF
--- a/src/NuGetGallery/Views/Shared/_ListPackage.cshtml
+++ b/src/NuGetGallery/Views/Shared/_ListPackage.cshtml
@@ -1,19 +1,5 @@
 ﻿@model ListPackageItemViewModel
 
-@helper OwnersHeader()
-{
-    if (Model.Owners != null && Model.Owners.Any())
-    {
-        <span class="package-by">
-            by:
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Views/Shared/_ListPackage.cshtml
```diff
@@ -24,31 +10,33 @@
...
-                        </a>
-                    }
-                </div>
-                <div class="package-by-container hidden-xs">
-                    @OwnersHeader()
-                </div>
-            </div>
+                @if (Model.CanEdit || Model.CanManageOwners || Model.CanUnlistOrRelist)
+                {
+                    <a href="@Url.ManagePackage(Model)" class="icon-link edit-button" title="Manage package @Model.Id">
```
agr: How about dropping the word "package" from the `title`? So it becomes "Manage @Model.Id".

scottbommarito: i'll drop "package"

Issue #4003, opened 2017-05-25
Incorrect comparer implementation causes infinite loop


agr:
`RetryingHttpClientWrapper.cs` contains the `WeightedRandomComparer` class that implements `IComparer<int>` in a way that involves rolling random numbers. This implementation does not pass basic sanity check for the comparer: `Compare(a,b) == -Compare(b,a)`. Specifically, for the case when array of zeroes is attempted to be sorted it causes the sort implementation to loop infinitely.
Specifically in case of Gallery, if both instances of search service are considered down, eventually, each request that involves the subsequent request to the search service would create a thread with an infinite loop inside of it. So after few dozens of such requests we'd have Gallery machines completely unresponsive.

agr:
Found the paper offering efficient solution for the problem: http://utopia.duth.gr/~pefraimi/research/data/2007EncOfAlg.pdf

agr:
Even worse, the instance of the `WeightedRandomComparer` is declared static, created once and exists for the life time of the process. It does not contain any synchronization logic inside and uses the object of `Random` class which is [not thread safe](https://msdn.microsoft.com/en-us/library/system.random(v=vs.110).aspx#ThreadSafety).

skofman1:
We will move to traffic manager: https://github.com/NuGet/NuGetGallery/issues/5700

PR #6832, opened 2019-01-19
SearchToTrafficManager_Stage1

cristinamanum:
1. Refactored the SearchClient as such can support the new Gallery Search Client. 
2. Introduced IHttpClientWrapper and RetryingHttpClientWrapper2 that will allow supporting the two search configured by a flag.
3. Added two new configurations for SearchUri and Autocomplete. 
4 Added support in ExternalSearch Service and AutoCompleteServiceQuery for the two search clients respective two httpClientsWrappers. 
5. Added unit tests.


src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
```diff
@@ -99,16 +93,51 @@ public ExternalSearchService(IAppConfiguration config, IDiagnosticsService diagn
...
+                    searchServiceUriCredentials,
+                    QuietLog.LogHandledException,
+                    new TracingHttpHandler(Trace),
+                    new CorrelatingHttpClientHandler());
+            }
         }
 
         private static readonly Task<bool> _exists = Task.FromResult(true);
+
+        public ICredentials ExtractCredentialsFromUri(ref Uri uri)
```
skofman1: I think we can drop the credentials support. I know it's not a part of this PR, but let's do it's as part of the overall work. What do you think?

src/NuGet.Services.Search.Client/Client/GallerySearchClient.cs
```diff
@@ -0,0 +1,46 @@
...
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace NuGet.Services.Search.Client 
+{
+    public class GallerySearchClient : SearchClient
```
skofman1: I recommend replicating SearchClient completely and cleaning it up so GalleryClient has only the code you need. Then you can delete SearchClient after testing of Gallery client is completed.

src/NuGet.Services.Search.Client/Client/HttpRetryMessageHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Search.Client
+{
+    public class HttpRetryMessageHandler : DelegatingHandler
```
skofman1: I assume this is copied code. Can you please add a link to the original location?

src/NuGet.Services.Search.Client/NuGet.Services.Search.Client.csproj
```diff
@@ -46,11 +46,15 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="Client\BaseUrlHealthIndicatorStore.cs" />
```
skofman1: I wonder if one of our goals should be eliminating this project and moving everything into the Gallery project. It's not used by any other component except for Gallery

src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
```diff
@@ -23,11 +23,17 @@ public class ExternalSearchService : ISearchService, IIndexingService, IRawSearc
 
         private static IEndpointHealthIndicatorStore _healthIndicatorStore;
         private static SearchClient _client;
+        private static SearchClient _clientTM;
```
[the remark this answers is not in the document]
cristinamanum: Will update the name to _client2.

src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
cristinamanum: I can do this.

agr:
> }

This class is an abomination of `IComparer` implementations. Can we do something about it while we touch that code? The [bug](https://github.com/NuGet/NuGetGallery/issues/4003) about it was closed more than a year ago, but the code is still there.

---
Refers to: src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper.cs:252 in 423530e. [](commit_id = 423530e8ad2060b7c12efb247944e139ccd4a757, deletion_comment = False)

src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper2.cs
```diff
@@ -0,0 +1,56 @@
...
+        public RetryingHttpClientWrapper2(ICredentials credentials, Action<Exception> onException, params DelegatingHandler[] handlers)
+        {
+            _onException = onException ?? throw new ArgumentNullException(nameof(onException));
+            HttpMessageHandler handler = new HttpRetryMessageHandler(new HttpClientHandler()
+            {
+                Credentials = credentials,
+                AllowAutoRedirect = true,
+                UseDefaultCredentials = credentials == null
+            }, onException, RetryCount);
+
```
agr: Is there a chance a connection will be reused after failure on retry, so we'll end up sending all retries to the same failing instance?

src/NuGet.Services.Search.Client/NuGet.Services.Search.Client.csproj
cristinamanum: Yes, one of the design meeting outputs was to eliminate this project and move to gallery. I propose to do this after the move to the TM search is complete. During the clean-up phase.

src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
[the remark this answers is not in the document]
scottbommarito: The naming of the variables should make their purposes clear. I don't think either `_client` or `_client2` accomplish that goal.

`_deprecatedSearchClient` and `_searchClient` maybe?

src/NuGet.Services.Search.Client/Client/GallerySearchClient.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        public Uri GetDiagnosticsUri()
+        {
+            return AppendPathToUri(_searchUri, "search/diag");
+        }
+
+        public override async Task<IEnumerable<Uri>> GetEndpoints()
+        {
+            await Task.Yield();
```
scottbommarito: Why are we yielding here?

src/NuGet.Services.Search.Client/Client/GallerySearchClient.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        private readonly Uri _searchUri;
+
+        /// <summary>
+        /// Create a search service client from the specified base uri and credentials.
+        /// </summary>
+        /// <param name="baseUri">The URL of the search service.</param>
+        /// <param name="credentials">The credentials to connect to the service with</param>
+        /// <param name="handlers">Handlers to apply to the request in order from first to last</param>
+        public GallerySearchClient(Uri baseUri, ICredentials credentials, Action<Exception> onException, params DelegatingHandler[] handlers)
```
scottbommarito: `Action<Exception> onException` only exists because this class is not in the same project as the rest of the gallery and we can't call `QuietLog`. Please move it to `NuGetGallery` and get rid of it.

src/NuGet.Services.Search.Client/Client/GallerySearchClient.cs
scottbommarito: You can reduce a lot of the code in `SearchClient` (specifically the parts that support multiple endpoints) when you copy it to `NuGetGallery`.

src/NuGetGallery/Configuration/AppConfiguration.cs
```diff
@@ -362,5 +362,15 @@ public string ExternalBrandingMessage
 
         [DefaultValue(true)]
         public bool AllowLicenselessPackages { get; set; }
+
+        /// <summary>
+        /// Gets the Uri for the Search endpoint
+        /// </summary>
+        public Uri SearchServiceUri { get; set; }
```
scottbommarito: I think we should make this a single config to the root of the search service, and then append paths to the URL based on the search endpoint we're hitting. We are already appending to the end of the URL to get the diag endpoint. To have the autocomplete URI supplied through config but not the diag URI is confusing.

src/NuGet.Services.Search.Client/NuGet.Services.Search.Client.csproj
cristinamanum: Moved the majority of the changes to gallery.

src/NuGetGallery/Configuration/AppConfiguration.cs
cristinamanum: Updated.

src/NuGetGallery/Queries/AutoCompleteServiceQuery.cs
```diff
@@ -29,6 +31,9 @@ public AutoCompleteServiceQuery(IAppConfiguration configuration)
             _serviceDiscoveryClient = new ServiceDiscoveryClient(configuration.ServiceDiscoveryUri);
             _autocompleteServiceResourceType = configuration.AutocompleteServiceResourceType;
             _httpClient = new RetryingHttpClientWrapper(new HttpClient(), QuietLog.LogHandledException);
+
+            _autocompleteSearchServiceUri = configuration.AutocompleteSearchServiceUri;
+            _httpClientTM = new RetryingHttpClientWrapper2(credentials: null, onException: QuietLog.LogHandledException);
         }
```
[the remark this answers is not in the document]
cristinamanum: Changed the names, One will be deprecated and the other will use the traffic manager endpoint.

src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
scottbommarito: Can we remove all the credential code now? It supports passing in `null` for no credentials.

src/NuGetGallery/Helpers/UriExtensions.cs
```diff
@@ -79,5 +79,16 @@ public static string AppendQueryStringToRelativeUri(string relativeUrl, IReadOnl
             builder.Query = query.ToString();
             return builder.Uri.PathAndQuery;
         }
+
+        public static Uri AppendPathToUri(this Uri uri, string pathToAppend, string queryString = null)
```
cristinamanum: This is an existing method used in Search. Just copied here.

src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
cristinamanum: I would prefer to not add / remove any of the existing functionality.

src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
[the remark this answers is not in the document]
cristinamanum: Updated.

src/NuGet.Services.Search.Client/Client/GallerySearchClient.cs
cristinamanum: Moved the code in Gallery.

src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
cristinamanum: I would prefer to not add/ remove anything on the current path.

src/NuGet.Services.Search.Client/Client/RetryingHttpClientWrapper2.cs
cristinamanum: Yes - to address I changed the static of the http client to use the HttpClientFactory that should address these issues. 

For the above comment "WeightedRandomComparer... " that code path will be kept until we move to traffic manager.

src/NuGet.Services.Search.Client/Client/GallerySearchClient.cs
cristinamanum: This was changed.

src/NuGet.Services.Search.Client/Client/GallerySearchClient.cs
cristinamanum: Changed.

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -2049,9 +2052,18 @@
...
   <ItemGroup>
+    <PackageReference Include="Autofac.Extensions.DependencyInjection">
+      <Version>4.3.1</Version>
+    </PackageReference>
     <PackageReference Include="Markdig.Signed">
       <Version>0.15.4</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection">
+      <Version>2.2.0</Version>
+    </PackageReference>
```
loic-sharma: We use `1.1.*` every where for `Microsoft.Extensions.*` dependencies. Can we downgrade this to `1.1.*`? For example, see [NuGet.Jobs.Common.csproj](https://github.com/NuGet/NuGet.Jobs/blob/79bdaeac2a3dd633477ea26664b988d01bc3d8cd/src/NuGet.Jobs.Common/NuGet.Jobs.Common.csproj#L97)

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -119,7 +124,9 @@ protected override void Load(ContainerBuilder builder)
                 .As<Lucene.Net.Store.Directory>()
                 .SingleInstance();
 
-            ConfigureSearch(builder, configuration);
+            // it will be used by the Autocomplete as well
+            services.AddTransient((s) => new HttpRetryMessageHandler(ex => QuietLog.LogHandledException(ex)));
+            ConfigureSearch(builder, configuration, services);
```
loic-sharma: Nit: `s => new HttpRetryMessageHandler(QuietLog.LogHandledException)`?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -637,6 +645,21 @@ private static void ConfigureSearch(ContainerBuilder builder, IGalleryConfigurat
             }
             else
             {
+                services.AddTransient<CorrelatingHttpClientHandler>();
+                services.AddTransient((s) => new TracingHttpHandler(DependencyResolver.Current.GetService<IDiagnosticsService>().SafeGetSource("ExternalSearchService")));
+
+                services.AddHttpClient<ISearchClient, GallerySearchClient>(c =>
+                    c.BaseAddress = DependencyResolver.Current.GetService<IAppConfiguration>().SearchServiceUri
```
loic-sharma: >DependencyResolver.Current.GetService<IAppConfiguration>() [](start = 36, length = 58)

Should `` be ``

src/NuGetGallery/Infrastructure/Lucene/GallerySearchClient.cs
```diff
@@ -0,0 +1,122 @@
...
+            bool isLuceneQuery = false,
+            bool countOnly = false,
+            bool explain = false,
+            bool getAllVersions = false,
+            string supportedFramework = null,
+            string semVerLevel = null)
+        {
+            IDictionary<string, string> nameValue = new Dictionary<string, string>();
+            nameValue.Add("q", query);
+            nameValue.Add("skip", skip.ToString());
```
joelverhagen: >nameValue [](start = 12, length = 9)

Can we share the query string building code (or more) with the existing `SearchClient`?

src/NuGetGallery/Infrastructure/Lucene/HttpRetryMessageHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken) =>
+            RetryPolicy<HttpResponseMessage>
+            .HandleExceptionAndResult( (ex) => { _onException?.Invoke(ex); return false; }, r => GetRetryingCodes.Contains(r.StatusCode))
+            .ExecuteWithWaitAndRetryAsync(_retryCount,
+                retryAttempt => TimeSpan.FromSeconds(Math.Pow(2,retryAttempt)),
+                () => base.SendAsync(request, cancellationToken) );
+
+        public List<HttpStatusCode> GetRetryingCodes { get; } = new List<HttpStatusCode>{
```
joelverhagen: This should be a hash set to avoid linear search

joelverhagen: Also, why is this public? Do we really want to support callers modifying this list?

---
In reply to: [251529194](https://github.com/NuGet/NuGetGallery/pull/6832#discussion_r251529194) [](ancestors = 251529194)

src/NuGetGallery/Infrastructure/Lucene/HttpRetryMessageHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            _onException = onException;
+            _retryCount = retryCount;
+            if (additionalHttpStatusCodesWorthRetrying != null)
+            {
+                GetRetryingCodes.AddRange(additionalHttpStatusCodesWorthRetrying);
+            }
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken) =>
```
joelverhagen: This method is complex enough that we should avoid the `=>` syntax for the method body

src/NuGetGallery/Infrastructure/Lucene/HttpRetryMessageHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+            RetryPolicy<HttpResponseMessage>
+            .HandleExceptionAndResult( (ex) => { _onException?.Invoke(ex); return false; }, r => GetRetryingCodes.Contains(r.StatusCode))
+            .ExecuteWithWaitAndRetryAsync(_retryCount,
+                retryAttempt => TimeSpan.FromSeconds(Math.Pow(2,retryAttempt)),
```
joelverhagen: >FromSeconds [](start = 41, length = 11)

IMO, the total duration should not exceed the timeout that client uses, which is 100 seconds.

src/NuGetGallery/Infrastructure/Lucene/AutoCompleteSearchClient.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+using NuGetGallery;
+
+namespace NuGet.Services.Search.Client
+{
+    public class AutoCompleteSearchClient
```
skofman1: why do we need a separate class for this? Can't we add autocomplete functionality to GallerySearchClient?

cristinamanum:
Moved to Circuit Breaker. Close.

PR #6834, opened 2019-01-22
 Added asp.net blob URIs to microsoft URIs list

skofman1:


src/NuGetGallery/Helpers/UriExtensions.cs
```diff
@@ -53,6 +53,7 @@ private static bool IsMicrosoftUri(this Uri uri)
                    string.Equals(uri.Authority, "microsoft.com", StringComparison.OrdinalIgnoreCase) ||
                    string.Equals(uri.Authority, "www.asp.net", StringComparison.OrdinalIgnoreCase) ||
                    string.Equals(uri.Authority, "asp.net", StringComparison.OrdinalIgnoreCase) ||
+                   uri.Authority.EndsWith(".asp.net", StringComparison.OrdinalIgnoreCase) ||
```
agr: Should we add some helpers (for "uri's authority is X" and "uri's authority ends with Y") to reduce that `StringComparison.OrdinalIgnoreCase` repetition in every line?

skofman1: done

src/NuGetGallery/Helpers/UriExtensions.cs
```diff
@@ -53,6 +53,7 @@ private static bool IsMicrosoftUri(this Uri uri)
                    string.Equals(uri.Authority, "microsoft.com", StringComparison.OrdinalIgnoreCase) ||
                    string.Equals(uri.Authority, "www.asp.net", StringComparison.OrdinalIgnoreCase) ||
```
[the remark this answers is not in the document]
skofman1: done

src/NuGetGallery/Helpers/UriExtensions.cs
```diff
@@ -35,26 +35,28 @@ public static bool IsGitHubUri(this Uri uri)
                    string.Equals(uri.Host, "github.com", StringComparison.OrdinalIgnoreCase);
         }
 
-        public static bool IsGitHubPagerUri(this Uri uri)
+        private static bool IsGitHubPagerUri(this Uri uri)
         {
             return uri.Authority.EndsWith(".github.com", StringComparison.OrdinalIgnoreCase) ||
```
scottbommarito: Any reason we can't replace these checks with `uri.IsInDomain("github.com")` and `uri.IsInDomain("github.io")`?

skofman1: There is an intentional separation here between GitHub and GitHub pages.
For example, www.github.com/bla is GitHub, but https://stuff.github.com is GitHub pages. Hence, the IsInDomain method shouldn't be used. 
This is why we have two separate methods: `IsGitHubPagerUri`, `IsGitHubUri`

Issue #6519, opened 2018-10-05
[Licenses] Prettify license endpoint in Gallery


agr:
License endpoint in Gallery should have the decorations from the rest of the web site, it should display both license files and license expressions in the response it produces. The page should explicitly state package details and the license embedded into it. 

 (related to #6518).

PR #6839, opened 2019-01-24
Prettify license endpoint in gallery

zhhyu:
The license endpoint in gallery should display license expression/file/url/no info for different cases. Page in "~/packages/id/version/license" should support this. 
Fix: https://github.com/NuGet/NuGetGallery/issues/6519

image
image
image


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -40,6 +40,8 @@ namespace NuGetGallery
     public partial class PackagesController
         : AppController
     {
+        private const long MaxAllowedLicenseLength = 1024 * 1024;
+
```
zhhyu: This place is obviously not elegant. Need to find some place to set this value globally.

src/NuGetGallery/Strings.resx
```diff
@@ -1088,4 +1088,7 @@ The {1} Team</value>
   <data name="UploadPackage_SymbolsPackageNotAllowed" xml:space="preserve">
     <value>Snupkg upload failed. Please use latest NuGet clients (v 4.9.0 or above) and a V3 endpoint to push Symbol packages. For details, refer https://docs.microsoft.com/nuget/create-packages/symbol-packages-snupkg</value>
   </data>
+  <data name="StreamMaxLengthExceeded" xml:space="preserve">
+    <value>The stream must be less than {0} bytes.</value>
+  </data>
```
zhhyu: This message seems not meaningful for different scenarios if we reuse "streamhelper".

src/NuGetGallery/Views/Packages/License.cshtml
```diff
@@ -0,0 +1,103 @@
...
+                        }
+                    </a>
+                </div>
+            }
+            else
+            {
+                <div>
+                    @ViewHelpers.AlertWarning(
+                        @<text>
+                            This package contains no license info.
```
shishirx34: ...no license information.

tests/NuGetGallery.Facts/Helpers/StreamHelperFacts.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class StreamHelperFacts
+    {
+        [Theory]
+        [InlineData("test content")]
+        [InlineData("测试内容")]
+        public async Task ReadValidStreamToString(string inputContent)
+        {
+            // Arrange
+            var stream = new MemoryStream(Encoding.UTF8.GetBytes(inputContent));
+
```
zhhyu: It's better to use using(stream){} here to dispose it and avoid memory leak.

src/NuGetGallery/Controllers/PackagesController.cs
zhhyu: Move to "CoreLicenseFileService".

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -734,30 +731,41 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
...
             {
                 try
                 {
-                    var licenseFileContent = await _coreLicenseFileService.DownloadLicenseFileAsync(package);
-                    return new FileStreamResult(licenseFileContent, "text/plain");
+                    var licenseFileStream = await _coreLicenseFileService.DownloadLicenseFileAsync(package);
+                    model.LicenseFileContents = await StreamHelper.ReadMaxAsync(licenseFileStream, CoreLicenseFileService.MaxAllowedLicenseSizeInBytes);
+
+                    licenseFileStream.Dispose();
                 }
```
loic-sharma: Why aren't you using a `using` statement on the `licenseFileStream` here?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -734,30 +731,41 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
...
         public virtual async Task<ActionResult> License(string id, string version)
         {
             var package = _packageService.FindPackageByIdAndVersionStrict(id, version);
-            if (package == null)
+            if (package?.PackageRegistration == null)
             {
                 return HttpNotFound();
             }
 
+            DisplayLicenseViewModel model = new DisplayLicenseViewModel(package);
```
loic-sharma: >DisplayLicenseViewModel [](start = 12, length = 23)

`var model = ...`

src/NuGetGallery/Helpers/StreamHelper.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            if (encoding == null)
+            {
+                encoding = Encoding.UTF8;
+            }
+
+            int bytesRead;
+            var offset = 0;
+            var buffer = new byte[maxSize + 1];
+
```
loic-sharma: This is scary. Someone may use `ReadMaxAsync` in the future and pass a `maxSize` of something like 250 megabytes without realizing that 250 megabytes will be allocated on each invocation. Can this be implemented without allocating the maximum allowed size?

loic-sharma:
>         private const string UrlHostRequirement = "raw.githubusercontent.com";

Is this supposed to be 8kb?

---
Refers to: src/NuGetGallery/Services/ReadMeService.cs:30 in c35dd48. [](commit_id = c35dd48217628127ae31fe41e30bef7709bac591, deletion_comment = False)

src/NuGetGallery/Helpers/StreamHelper.cs
zhhyu: I think that the reason why we have maxSize is that we want to restrict the length of string to be read. So we may need to keep this.
The simplest way to solve this is to decide the stream length and use Math.min(stream.length, maxSize) for buffer size. However, the stream length sometimes can not be decided ahead.

src/NuGetGallery.Core/Services/CoreLicenseFileService.cs
```diff
@@ -12,6 +12,7 @@ namespace NuGetGallery
 {
     public class CoreLicenseFileService : ICoreLicenseFileService
     {
+        public const long MaxAllowedLicenseSizeInBytes = 1024 * 1024; // 1Kb
```
agr: That's 1 MB, not KB. And we have similar constant defined in `PackageUploadService`. We must not have two definitions as they might diverge with surprising consequences.

src/NuGetGallery/Helpers/StreamHelper.cs
agr: @loic-sharma, caller asks to read a string up to `n` bytes in size they should expect that O(n) memory is going to be used for the operation. On the other hand `MemoryStream` can be utilized as a temporary storage instead of an array to conserve memory fairly easy.

src/NuGetGallery/Views/Packages/License.cshtml
```diff
@@ -0,0 +1,110 @@
...
+{<span>@segment.Value</span>}
+
+@functions
+{
+    static bool IsLicenseOrException(CompositeLicenseExpressionSegment segment)
+    {
+        return segment.Type == CompositeLicenseExpressionSegmentType.LicenseIdentifier || segment.Type == CompositeLicenseExpressionSegmentType.ExceptionIdentifier;
+    }
+}
+
```
agr: Three methods above are copied from `DisplayPackage.cshtml`. Can we have a single definition used in both places instead of copying?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -734,39 +731,50 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
...
+                try
+                {
+                    model.LicenseExpressionSegments = _licenseExpressionSplitter.SplitExpression(package.LicenseExpression);
+                }
+                catch (Exception ex)
+                {
+                    // Any exception thrown while trying to render license expression beautifully
+                    // is not severe enough to break the client experience, view will fall back to
+                    // display license url.
+                    _telemetryService.TraceException(ex);
```
agr: This comment is copy-pasted from a different location and is wrong in this context.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -734,39 +731,50 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
...
+                    return HttpNotFound();
+                }
```
agr: We shouldn't return `HttpNotFound` on exception in this case either.

src/NuGetGallery/Strings.resx
agr: This message is not aligned with the check in the code: code throws when `bytesRead > maxSize` message says the following must be true: `bytesRead < maxSize`.

tests/NuGetGallery.Facts/Helpers/StreamHelperFacts.cs
```diff
@@ -0,0 +1,40 @@
...
+            var maxSize = 1000;
+            using (var stream = new MemoryStream(new byte[maxSize + 1]))
+            {
+                var exception = await Assert.ThrowsAsync<InvalidOperationException>(() => StreamHelper.ReadMaxAsync(stream, maxSize));
+
+                Assert.Contains(exception.Message, string.Format(CultureInfo.CurrentCulture, Strings.StreamMaxLengthExceeded, maxSize));
+            }
+        }
+    }
+}
```
agr: Need more tests: passing `null` as `stream`, passing zeroes or negative numbers as `maxSize`, trying to read `X` bytes from a stream with `X` bytes in it (to verify that `maxSize` is inclusive).

tests/NuGetGallery.Facts/ViewModels/DisplayLicenseViewModelFacts.cs
```diff
@@ -0,0 +1,80 @@
...
+            Assert.Equal(embeddedLicenseType, model.EmbeddedLicenseType);
+            Assert.Equal(licenseExpression, model.LicenseExpression);
+        }
+
+        [Fact]
+        public void LicenseNamesAreParsedByCommas()
+        {
+            // arrange
+            var licenseUrl = "https://mylicense.com/";
+            var package = new Package
```
agr: just `[https://mylicense/`](https://mylicense/`), please.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -722,6 +714,10 @@ private static void ConfigureForLocalFileSystem(ContainerBuilder builder, IGalle
             builder.RegisterInstance(new SqlErrorLog(configuration.Current.SqlConnectionString))
                 .As<ErrorLog>()
                 .SingleInstance();
+
+            builder.RegisterType<GalleryContentFileMetadataService>()
+                .As<IContentFileMetadataService>()
+                .InstancePerLifetimeScope();
```
skofman1: SingleInstance
No need to create per request since it has no state

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -814,6 +810,10 @@ private static void ConfigureForAzureStorage(ContainerBuilder builder, IGalleryC
             builder.RegisterInstance(new TableErrorLog(configuration.Current.AzureStorage_Errors_ConnectionString, configuration.Current.AzureStorageReadAccessGeoRedundant))
                 .As<ErrorLog>()
                 .SingleInstance();
+
+            builder.RegisterType<FlatContainerContentFileMetadataService>()
+                .As<IContentFileMetadataService>()
+                .InstancePerLifetimeScope();
```
skofman1: SingleInstance

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -734,39 +731,47 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
             return View(model);
         }
 
+        [HttpGet]
+        [UIAuthorize]
         public virtual async Task<ActionResult> License(string id, string version)
         {
             var package = _packageService.FindPackageByIdAndVersionStrict(id, version);
-            if (package == null)
+            if (package?.PackageRegistration == null)
```
skofman1: can we have a package in the DB without PackageRegistration? Why is package == null not sufficient?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -734,39 +731,47 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
...
             {
                 return HttpNotFound();
             }
 
+            var model = new DisplayLicenseViewModel(package);
             if (!string.IsNullOrWhiteSpace(package.LicenseExpression))
```
skofman1: I recommend moving this logic to DisplayLicenseViewModel, will make the code cleaner

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -734,39 +731,47 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
...
-            if (!_config.AsynchronousPackageValidationEnabled)
+            if (package.EmbeddedLicenseType != EmbeddedLicenseFileType.Absent)
             {
                 try
                 {
-                    var licenseFileContent = await _coreLicenseFileService.DownloadLicenseFileAsync(package);
-                    return new FileStreamResult(licenseFileContent, "text/plain");
+                    using (var licenseFileStream = await _coreLicenseFileService.DownloadLicenseFileAsync(package))
+                    {
+                        model.LicenseFileContents = await StreamHelper.ReadMaxAsync(licenseFileStream, PackageUploadService.MaxAllowedLicenseLength);
```
skofman1: consider refactoring this code so that you call the ViewModel constructor only once with all required information. Will make cleaner and more readable code. Also, remove the code duplication in exception handling

src/NuGetGallery/Helpers/StreamHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        private const long BufferSize = 80 * 1024;  // 80 KB
+
+        /// <summary>
+        /// Transform the stream to string given maximum size.
+        /// </summary>
+        /// <param name="stream">stream to be read.</param>
+        /// <param name="maxSize">maximum size.</param>
+        /// <returns>string format of the stream.</returns>
+        public static async Task<string> ReadMaxAsync(Stream stream, long maxSize)
```
skofman1: consider converting to extension methods

src/NuGetGallery/Helpers/StreamHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            if (maxSize <= 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(maxSize), $"{nameof(maxSize)} must be greater than 0");
+            }
+
+            if (encoding == null)
```
skofman1: if you are setting a default value in the method let users skip providing it when calling it: 

ReadMaxAsync(Stream stream, long maxSize, Encoding encoding = null)

src/NuGetGallery/Helpers/StreamHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            using (var memoryStream = new MemoryStream())
+            {
+                int bytesRead;
+                var totalBytesRead = 0;
+                var buffer = new byte[BufferSize];
+                while ((bytesRead = await stream.ReadAsync(buffer, 0, buffer.Length)) > 0)
+                {
+                    totalBytesRead += bytesRead;
+                    if (totalBytesRead > maxSize)
```
skofman1: >if (totalBytesRead > maxSize [](start = 20, length = 28)

if maxSize is smaller than BufferSize, wouldn't this always throw?

zhhyu: Consider the previous logic, the exception will always be thrown once we read larger than maxSize bytes. So if maxSize is smaller than BufferSize, it will depend on the stream size. If the stream size is less than maxSize, it will still get all the bytes from stream. Otherwise, it will throw exception. "BufferSize" used by the stream.readAync function means each time we will put at most BufferSize bytes from the stream into the buffer. 

However, now we change the logic. It will always read "Math.min(stream size, maxSize)" bytes from the stream without any exceptions.

src/NuGetGallery/Controllers/PackagesController.cs
zhhyu: I am not very sure whether it's a good pattern to put async method in the constructor. Need to do some research into it.

src/NuGetGallery/Helpers/StreamHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+using System;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NuGetGallery.Helpers
+{
+    public class StreamHelper
+    {
+        public static int BufferSize { get; } = 80 * 1024;  // 80 KB
```
loic-sharma: I'd make this a const field, and maybe internal instead of public.

zhhyu: The reason to use "public get" is that we need this value for unit test.

src/NuGetGallery/Controllers/PackagesController.cs
zhhyu: Initialize viewmodel after getting license expression and file.

src/NuGetGallery/Helpers/StreamHelper.cs
loic-sharma: You can use public or internal const fields in tests (though for internal you'll need `InternalsVisibleTo` attribute)

src/NuGetGallery/Helpers/StreamHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        /// <summary>
+        /// Read the stream at most maxSize bytes and transform to string.
+        /// </summary>
+        /// <param name="stream">stream to be read.</param>
+        /// <param name="maxSize">maximum size.</param>
+        /// <param name="encoding">encoding format.</param>
+        /// <returns>string format of the stream.</returns>
+        public static async Task<string> ReadMaxAsync(Stream stream, long maxSize, Encoding encoding = null)
```
skofman1: >ReadMaxAsync [](start = 41, length = 12)

GetStringOrThrowIfTooLongAsync

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -23,6 +23,16 @@ namespace NuGetGallery
...
     {
+        /// <summary>
+        /// The upper limit on allowed license file size.
+        /// </summary>
+        /// <remarks>
+        /// This limit is chosen fairly arbitrarily, it has to be large enough to fit any sensible license
+        /// in plain text and small enough to not cause issues with scanning through such file a few times
+        /// during the package validation.
+        /// </remarks>
+        public const long MaxAllowedLicenseLength = 1024 * 1024; // 1 MB
```
joelverhagen: >MaxAllowedLicenseLength [](start = 26, length = 23)

Wait, really?

How does a megabyte of license even perform on your browser?

agr: I don't see much issues [here](https://dev.nugettest.org/packages/Andrei.TestNugetPackage/0.0.3.667). Preview page behaved nicely, too when the package was uploaded.

src/NuGetGallery/Helpers/StreamHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <param name="maxSize">maximum size.</param>
+        /// <returns>truncated memorystream.</returns>
+        public static async Task<TruncatedStream> GetTruncatedStreamWithMaxSizeAsync(this Stream stream, int maxSize)
+        {
+            if (stream == null)
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            if (maxSize <= 0 || maxSize >= int.MaxValue)
```
skofman1: can an int be bigger than int.MaxValue ? :)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -742,31 +750,42 @@ public virtual async Task<ActionResult> License(string id, string version)
...
-                    return new FileStreamResult(licenseFileContent, "text/plain");
+                    licenseExpressionSegments = _licenseExpressionSplitter.SplitExpression(package.LicenseExpression);
                 }
-                catch (Exception ex)
+
+                if (package.EmbeddedLicenseType != EmbeddedLicenseFileType.Absent)
                 {
-                    _telemetryService.TraceException(ex);
-                    return HttpNotFound();
+                    using (var licenseFileStream = await _coreLicenseFileService.DownloadLicenseFileAsync(package))
```
joelverhagen: This returns a fully buffered stream in the blob storage case. Is this expected? In other words, we already have the full file in memory here. You could use `ISimpleCloudBlob.OpenReadAsync` if you want to limit this.

src/NuGetGallery/Helpers/StreamHelper.cs
zhhyu: We don't allow that maxSize equal to int.MaxValue, which will cause overflow, so that's why "equal to" check is a must. The "larger than" is like redundancy.

src/NuGetGallery/Controllers/PackagesController.cs
zhhyu: Talk offline, the "licenseFileStream" is already in memory. I leave a comment so that we will consider refactoring "coreLicenseFileService" and add "GetUnBufferedFileAsync" method in "CloudBlobCoreFileStorageService" in the future.

PR #6842, opened 2019-01-25
[PROPOSAL] Add "simulate error" endpoints to gallery for testing

joelverhagen:
This proposal is in regards to issue https://github.com/NuGet/NuGetGallery/issues/6601. We have seen strange behavior when it comes to how exceptions and non-success status codes are dealt with in the gallery. Right now we have very high Application Insights noise around HTTP 500s since some HTTP 400s get mapped to HTTP 500. This is bad since it is potentially hiding really HTTP 500s that we need to handle.

Also, we have certain experiences built around error cases. For example, in descending order of importance:

1. **Critical:** HTTP 400 reason phrase is used by to pass an error message to client. If these reason phrase or status stops working as expected, `nuget.exe push` experience is significantly degraded.
1. **Important:** Non-HTML response when `api/v2/Packages(Id='...',Version='...')` returns 404. This is contract. Returning our pretty HTML page here has unknown impact on clients.
1. **Nice to have:** pretty 400, 404, and 500 error pages on the website.

My proposal is to introduce three "simulate error" endpoints and add functional tests that assert these intended error behaviors and more, so that we can be sensitive to any unintended change. The reason I added three is because we have three different categories of controllers with regards to error handling:

1. MVC controllers that don't override `OnException`. I added an endpoint to `PagesController` but this behavior should be consistent across all of our non-OData controllers except `ApiController`.

1. MVC controllers that do override `OnException`. Right now this is just `ApiController`.

1. Web API controllers. These are the OData controllers. I added the endpoint to the curated feed controller to minimize visibility.

This PR just adds the endpoints. I will add the functional tests later.

I am comfortable with these endpoints being on all environments and think it's actually important that we run these tests on all environments since other configurations or APIM can be different from environment to environment. Essentially we have a pretty complex sandwich of stuff that we don't own in front of our app code so functional tests are important.

### DIFF
--- a/src/NuGetGallery/App_Start/NuGetODataV2CuratedFeedConfig.cs
+++ b/src/NuGetGallery/App_Start/NuGetODataV2CuratedFeedConfig.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Linq;
+using System.Net.Http;
 using System.Web.Http;
 using System.Web.Http.OData.Batch;
 using System.Web.Http.OData.Builder;
@@ -11,6 +12,7 @@ using System.Web.Http.OData.Routing.Conventions;
 using Microsoft.Data.Edm;
 using Microsoft.Data.Edm.Csdl;
 using Microsoft.Data.OData;
+using NuGetGallery.Controllers;
 using NuGetGallery.OData;
 using NuGetGallery.OData.Conventions;
 using NuGetGallery.OData.Routing;
@@ -67,6 +69,9 @@ namespace NuGetGallery
             var findPackagesAction = builder.Action("FindPackagesById");
             findPackagesAction.Parameter<string>("id");
             findPackagesAction.ReturnsCollectionFromEntitySet<V2FeedPackage>("Packages");
+
+            var simulateErrorAction = builder.Action(nameof(ODataV2CuratedFeedController.SimulateError));
+            simulateErrorAction.Parameter<string>("type");
 
             var model = builder.GetEdmModel();
             model.SetEdmVersion(new Version(1, 0));
--- a/src/NuGetGallery/RequestModels/SimulatedErrorType.cs
+++ b/src/NuGetGallery/RequestModels/SimulatedErrorType.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    public enum SimulatedErrorType
+    {
+        Result400,
+        Result404,
+        Result500,
+        Result503,
+        HttpException400,
+        HttpException404,
+        HttpException500,
+        HttpException503,
+        HttpResponseException400,
+        HttpResponseException404,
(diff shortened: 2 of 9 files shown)


src/NuGetGallery/RequestModels/SimulatedErrorType.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    public enum SimulatedErrorType
```
joelverhagen: I chose to not parameterize the status code or reason phrase because this has unknown security impact.

src/NuGetGallery/App_Start/NuGetODataV2CuratedFeedConfig.cs
```diff
@@ -68,6 +70,9 @@ public static IEdmModel GetEdmModel()
             findPackagesAction.Parameter<string>("id");
             findPackagesAction.ReturnsCollectionFromEntitySet<V2FeedPackage>("Packages");
 
+            var simulateErrorAction = builder.Action(nameof(ODataV2CuratedFeedController.SimulateError));
+            simulateErrorAction.Parameter<string>("type");
+
```
cristinamanum: The fault injection is a great idea. Instead of adding the endpoints to be exposed public maybe different options could be considered. For example : feature flags, having specific builds with faults enabled, or maybe there are fault injection tools that could be used.

joelverhagen:
Spoke offline to @cristinamanum.

- We should document these endpoints internally so we can understand the motivation and how they should be enhanced.
- We should throttle these endpoints via APIM to mitigate an attacker using this endpoint to create a lot of logging/monitoring noise that could drown out real signals.
- Let's add a magic string that must be provided to reach the endpoint. Maybe a header or query string. (This perhaps removes the need for throttling)
- Let's open this discussion up to the rest of the team.

src/NuGetGallery/App_Start/NuGetODataV2CuratedFeedConfig.cs
joelverhagen: We can consider IP range blocking. We need to determine the IP range for our test agents and CORPNET.

joelverhagen: We should consider testing or at least investigating 401/403 errors.

Maybe we can go to AI and look at the whole set of error codes returned in the past 90 days.

One problem with the IP range is that we can't debug the test at home. We can mitigate this by making the test CI-only.

joelverhagen: Consider prototyping the blocking on DEV and documenting it. Then we only apply it elsewhere when necessary.

joelverhagen: I think the IP range blocking is not feasible. APIM supports it but Azure has hundreds ranges per region. For example, uswest2 has 218 IP ranges.

Issue #6845, opened 2019-01-25
Manage package page is slow with many versions


joelverhagen:
Check FAKE on dev

PR #6855, opened 2019-01-26
DisplayPackageViewModel.PackageVersions should not be enumerated multiple times

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6845

This PR adds a `ToList` to `DisplayPackageViewModel.PackageVersions`, which prevents it from be enumerated multiple times. Without this change, it is recalculated every time it is called, which is very expensive with many versions.

I also simplified the constructor by getting rid of the `packageHistory` parameter that was continually set to the same expression by all of our app code and by making the other constructor, which was only intended to be used by the other constructor, private.

### DIFF
--- a/src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
+++ b/src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
@@ -13,7 +13,7 @@ namespace NuGetGallery
 {
     public class DisplayPackageViewModel : ListPackageItemViewModel
     {
-        public DisplayPackageViewModel(Package package, User currentUser, IOrderedEnumerable<Package> packageHistory)
+        public DisplayPackageViewModel(Package package, User currentUser)
             : this(package, currentUser, (string)null)
         {
             HasSemVer2Version = NuGetVersion.IsSemVer2;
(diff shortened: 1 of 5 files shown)


src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -23,7 +23,13 @@ public DisplayPackageViewModel(Package package, User currentUser, IOrderedEnumer
...
 
             Dependencies = new DependencySetsViewModel(package.Dependencies);
-            PackageVersions = packageHistory.Select(p => new DisplayPackageViewModel(p, currentUser, GetPushedBy(p, currentUser)));
+
+            var packageHistory = package
+                .PackageRegistration
+                .Packages
+                .OrderByDescending(p => new NuGetVersion(p.Version))
+                .ToList();
+            PackageVersions = packageHistory.Select(p => new DisplayPackageViewModel(p, currentUser, GetPushedBy(p, currentUser))).ToList();
```
joelverhagen: `PackageVersions` should not be `IEnumerable`. Otherwise, looks great!

PR #6861, opened 2019-01-29
Improve runtime for DuplicatePushesAreRejectedAndNotDeleted

### LINKED ISSUE #6860: Fix DuplicatePushesAreRejectedAndNotDeleted for performance
DuplicatePushesAreRejectedAndNotDeleted waits on 10 package validations synchronously.

shishirx34:
Fixes #6860 and improves the test runtime by 84%. 

Before:

image

Now:

image

After doing some rebalancing of tests we have 20-30mins improvement in test runtime:

Before:
image

Now:
image


tests/NuGetGallery.FunctionalTests.Core/XunitExtensions/InMemoryTestOutputHelper.cs
```diff
@@ -0,0 +1,39 @@
...
+        : ITestOutputHelper
+    {
+        private StringBuilder _output;
+
+        private InMemoryTestOutputHelper()
+        {
+            _output = new StringBuilder();
+        }
+
+        public static InMemoryTestOutputHelper New
```
loic-sharma: What's this pattern for?

tests/NuGetGallery.FunctionalTests/PackageCreation/ApiPushTests.cs
```diff
@@ -30,25 +30,35 @@ public ApiPushTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
         [Description("Pushes many packages of the same ID and version. Verifies exactly one push succeeds and the rest fail with a conflict.")]
         [Priority(2)]
         [Category("P2Tests")]
-        public async Task DuplicatePushesAreRejectedAndNotDeleted()
+        public void DuplicatePushesAreRejectedAndNotDeleted()
```
joelverhagen: How did this become non-async?

`Parallel.For` does not await the a task produced by the delegate. It's parameter is `Action<int>` not `Func<int, Task>`.

These are unawaited tasks as far as I can tell.

You should use something like `Task.WhenAll`.

joelverhagen: See https://gist.github.com/joelverhagen/02e95d10189d42bad123db52a3806f7c for a minimal example of what is happening.

tests/NuGetGallery.FunctionalTests/Commandline/NuGetCommandLineTests.cs
```diff
@@ -61,8 +61,8 @@ public static IEnumerable<object[]> UploadAndUnlistPackages_Data
         [Theory]
         [MemberData(nameof(UploadAndUnlistPackages_Data))]
         [Description("Tests upload and unlist scenarios with API key")]
-        [Priority(0)]
-        [Category("P0Tests")]
+        [Priority(1)]
+        [Category("P1Tests")]
         public async Task UploadAndUnlistPackages(string apiKey)
```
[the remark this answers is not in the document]
shishirx34: This is a hacky way of distributing the tests for improving over all runtime, each category is ran as a single process and the tests under these categories are run sequentially. This test is kind of expensive takes about 30mins, other P0Tests are also time heavy hence moving this to P1 would distribute that time load to other process which doesn't have much execution time. 
Note the categories are meaningless for us since we don't really run them as required per env, all categories are executed all the time.

tests/NuGetGallery.FunctionalTests.Core/XunitExtensions/InMemoryTestOutputHelper.cs
shishirx34: not needed, anymore, I added this to hold the output in memory before writing to TestOutputHelper because of the synchronization issue. But using with `Task.WhenAll` doesn't need it.

shishirx34: deleted.

tests/NuGetGallery.FunctionalTests/PackageCreation/ApiPushTests.cs
shishirx34: Good catch.. I missed it

tests/NuGetGallery.FunctionalTests/PackageCreation/ApiPushTests.cs
```diff
@@ -30,25 +30,35 @@ public ApiPushTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
...
+            List<InMemoryTestOutputHelper> testOutputHelpers = new List<InMemoryTestOutputHelper>();
+            for (var i = 0; i < pushVersionCount; i++)
             {
-                var packageCreationHelper = new PackageCreationHelper(TestOutputHelper);
-                var packageId = $"{nameof(DuplicatePushesAreRejectedAndNotDeleted)}.{DateTime.UtcNow.Ticks}";
+                testOutputHelpers.Add(InMemoryTestOutputHelper.New);
+            }
 
-                for (var i = 0; i < 10; i++)
+            Parallel.For(0, pushVersionCount, async (i) =>
```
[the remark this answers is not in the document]
shishirx34: I used `Parallel.For` simply for readability, however, it runs into the problem Joel mentioned above. changing to Task.

scottbommarito:
Using `?w=1` helped a LOT with this PR.

tests/NuGetGallery.FunctionalTests/PackageCreation/ApiPushTests.cs
```diff
@@ -33,64 +33,72 @@ public ApiPushTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
...
+            var packageId = $"{nameof(DuplicatePushesAreRejectedAndNotDeleted)}.{DateTime.UtcNow.Ticks}";
+
+            int pushVersionCount = 10;
+            var duplicatePushTasks = new List<Task>();
+            for (var duplicateTaskIndex = 0; duplicateTaskIndex < pushVersionCount; duplicateTaskIndex++)
+            {
+                duplicatePushTasks.Add(PushDuplicates(packageId, $"1.0.{duplicateTaskIndex}", duplicateTaskIndex == 0));
+            }
+
+            await Task.WhenAll(duplicatePushTasks);
```
scottbommarito: I have some slight concerns with this. We've been very mindful of not pushing/unlisting a lot of packages at the exact same time to reduce the likelihood of feed2catalog skipping packages. It seems like this test is now pushing 10 packages simultaneously. This sounds like something that may potentially cause issues with feed2catalog. Are we concerned about this?

shishirx34: This test is only going to attempt 160 pushes, only 10 will be successful. I don't think this should cause any issues as such but @joelverhagen would be better able to answer these.

shishirx34:
Reran all functional tests with this change: https://nuget.visualstudio.com/NuGetBuild/_build/results?buildId=49255&view=ms.vss-test-web.test-result-details 

The improvements look good.

tests/NuGetGallery.FunctionalTests/PackageCreation/ApiPushTests.cs
joelverhagen: Frankly, if this causes problems with feed2catalog, this is good. If a test is exposing a bug... then there's a bug that should be fixed. Customers can encounter this on their own and this test doesn't make that worse (AFAIK). If these test packages are skipped there is no customer impact (AFAIK).

This is a case that a customer could simulate so our service should ideally be resilient to it.

PR #6864, opened 2019-01-30
Add logging to the automatic RequireOrganizationTenantPolicy path

scottbommarito:
https://github.com/NuGet/Engineering/issues/2059

- Added logging to the path, so we can now at least detect whether or not the decision not to apply the policy was due to not being able to find the tenant ID of the administrator or if it was done because the domain-tenant-pair check failed.
- Made some minor improvements to the domain-tenant-pair check
    - I replaced the algorithm using VS's `GetHashCode` generation feature.
    - I made the algorithm case-insensitive. Previously, the comparison would fail between "MICROSOFT.COM" and "microsoft.com".
- Added unit tests and improved the existing tests.


src/NuGetGallery/Services/UserService.cs
```diff
@@ -39,6 +40,8 @@ public class UserService : IUserService
 
         public ITelemetryService TelemetryService { get; protected set; }
 
+        public ILogger<UserService> Logger { get; protected set; }
+
```
loic-sharma: Can this be private?

scottbommarito: I don't want to break the existing pattern here. I believe I would have to rewrite some unit tests if I wanted to make this change to every property.

tests/NuGetGallery.Facts/Services/ContentObjectServiceFacts.cs
```diff
@@ -116,7 +116,7 @@ public async Task RefreshRefreshesObject()
                 Assert.True(loginDiscontinuationConfiguration.DiscontinuedForEmailAddresses.SequenceEqual(emails));
                 Assert.True(loginDiscontinuationConfiguration.DiscontinuedForDomains.SequenceEqual(domains));
                 Assert.True(loginDiscontinuationConfiguration.ExceptionsForEmailAddresses.SequenceEqual(exceptions));
-                Assert.True(loginDiscontinuationConfiguration.EnabledOrganizationAadTenants.SequenceEqual(orgTenantPairs, new OrganizationTenantPairComparer()));
+                Assert.True(loginDiscontinuationConfiguration.EnabledOrganizationAadTenants.SequenceEqual(orgTenantPairs));
```
agr: Why not `Assert.Equal`? It works for collections.

PR #6875, opened 2019-02-04
Add support for vulnerability data to entities model

xavierdecoster:


src/NuGet.Services.Entities/CVE.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <summary>
+        /// Gets or sets the primary key for the entity.
+        /// </summary>
+        public int Key { get; set; }
+
+        /// <summary>
+        /// Gets or sets the unique CVE-ID.
+        /// </summary>
+        [Index(IsUnique = true)]
+        public string CVEId { get; set; }
```
scottbommarito: Could we use the `CVE ID` as the key, given that it's unique? Same for CWEs as well?

I can see the following benefits:
- package deprecation entities will store the actual ID that can be used to query an external data source instead of an arbitrary key, potentially reducing the number of queries made if the ID is all that needs to be known
- if CVEs/CWEs are deleted from the database, the database still stores the intended IDs instead of an arbitrary key
- no need for an additional unique index

On the other hand, I was under the impression that SQL is slower when using strings for keys (specifically for inserts?), but I could be wrong. If this is the case, it could be a problem depending on the number of CVEs/CWEs in the table.

Any thoughts?

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -398,6 +404,14 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
             modelBuilder.Entity<PackageDeprecation>()
                 .Property(v => v.CVSSRating)
                 .HasPrecision(3, 1);
+
+            modelBuilder.Entity<PackageDeprecation>()
+                .HasMany(p => p.CVEs)
+                .WithMany(c => c.PackageDeprecations);
```
scottbommarito: Have we verified that when
- a `PackageDeprecation` is deleted, it deletes the corresponding rows in the `PackageDeprecationCVE`/`PackageDeprecationCWE`s tables?
- a `CVE`/`CWE` is deleted, it deletes the corresponding rows in the `PackageDeprecationCVE`/`PackageDeprecationCWE`s tables?

Or is this not the behavior we want to have?

joelverhagen:
To be clear, I am not against this stuff being in the gallery DB, but I wanted to make sure we evaluated other options like an independent database+service or another storage mechanism like Azure Search or in-memory JSON.

scottbommarito:
@joelverhagen - with this PR we are now modeling "deprecation to CVE/CWE" in a proper many-to-many relationship table and no longer a JSON array. I agree that we should make sure we do our due diligence in investigating all options, but I think it's a huge benefit to this approach that it allows our gallery database to store proper relationships instead of JSON.

joelverhagen:
@scottbommarito, oops, sorry I missed that part. I think think we should understand whether it is necessary to couple CVE/CWE data with gallery data. A JSON array would be appropriate if you want to refer to data in another system (potentially backed by another SQL database).

Remember we put validation data in its own database. The idea there was that there is some state about validation that gallery will never care about. In the CVE/CWE case, another database would allow us to scale independently and keep details like ingestion state/etags/timestamps/throttles in tables that have nothing to do with gallery DB.

Again, I don't feel strongly but we should give all of these a approaches an honest consideration properly weighting maintenance costs of another DB with the benefit of decoupling unrelated data.

src/NuGet.Services.Entities/CVE.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Gets or sets the description of the CWE.
+        /// </summary>
+        public string Description { get; set; }
```
joelverhagen: Required?

src/NuGet.Services.Entities/CWE.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Gets or sets the unique CWE-ID.
+        /// </summary>
+        [Index(IsUnique = true)]
+        public string CWEId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the name of the CWE.
+        /// </summary>
+        public string Name { get; set; }
```
joelverhagen: Required?

src/NuGet.Services.Entities/CWE.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Gets or sets the description of the CWE.
+        /// </summary>
+        public string Description { get; set; }
```
joelverhagen: Required?

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -371,6 +371,12 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
             modelBuilder.Entity<PackageDeprecation>()
                 .HasKey(d => d.Key);
 
+            modelBuilder.Entity<CVE>()
+                .HasKey(d => d.Key);
```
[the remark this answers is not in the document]
xavierdecoster: I checked the source data for CVE and CWE:

* CWE max description length = 507 chars
* CWE max name length = 125 chars
* CVE max description length = 3932 chars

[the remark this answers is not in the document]
scottbommarito: Is this the maximum length of the current data, or a maximum length restriction they have?

src/NuGetGallery.Core/Entities/EntitiesContext.cs
joelverhagen: This is a key insight.

This is quite important because we need to understand what to do when a CVE/CWE is not longer present in the data source or if some "delete" signal is presented to us. I think the easiest approach is to mandate that CVE/CWEs are never deleted. If this is already the case in the data source, then we're fine. Otherwise, we need to verify that it's okay for us to keep a deleted CVE/CWE in our database forever.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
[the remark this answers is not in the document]
xavierdecoster: Maximum length of current data. Did not find any documentation regarding limitations they have.

src/NuGet.Services.Entities/CVE.cs
xavierdecoster: We could, although I usually prefer primary keys not to contain any functional data. The PK is a pure technical key, whilst the CVE and CWE IDs are actually meaningful data, which indeed are of type `string` and less optimal than `int` in various query plans.

The main issue with using non-technical keys as primary key is that we better make sure these keys don't change over time. 
Checking syntax history for CVE IDs, the _syntax_ [has been subject to change in the past](https://cve.mitre.org/cve/identifiers/syntaxchange.html), although existing CVE IDs prior to the syntax change where not affected. I couldn't find any changes to CWE ID syntax. I think we can assume that - in order to not break any references - existing IDs will not change over time.
Not sure whether we can make that assumption... even though it appears that way. Which is why I stayed on the safe side here and went for a technical PK.

The benefits you mention could be nice to have though, so not feeling to strong about the above. Happy to hear more opinions :)

src/NuGetGallery.Core/Entities/EntitiesContext.cs
xavierdecoster: I think we can assume these ID's are never hard-deleted, although I couldn't find an official statement on that. 

In general, _[MITRE will maintain CWE as long as it serves the community to do so](https://cwe.mitre.org/about/faq.html#G.7)_. 

As for CVE IDs, these will be marked with a "reject" state and the "description" field will be labeled as such in the data source, according to [this source](https://nvd.nist.gov/general/faq#1da2269b-1c76-4349-be9a-be067a885956).

src/NuGetGallery.Core/Entities/EntitiesContext.cs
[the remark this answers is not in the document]
scottbommarito: Bummer. I would air on the side of caution then--especially for description length.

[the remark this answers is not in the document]
xavierdecoster: Added the following constraints:

* CWE max description length = 600 chars
* CWE max name length = 200 chars
* CVE max description length = 4000 chars

src/NuGet.Services.Entities/CWE.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace NuGet.Services.Entities
+{
+    /// <summary>
+    /// Represents a Common Weakness Enumeration (CWE).
+    /// </summary>
+    public class CWE
```
agr: Improper capitalization? Should be `Cwe`. Like URL -> `Url`.

xavierdecoster: I was wondering about capitalization, too. I can change it as you suggest, but noticed we also use `CVSSRating` on the `PackageDeprecation` entity, which following your suggestion should've been `CvssRating` then?

xavierdecoster: Updated.

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -93,15 +94,13 @@ public class PackageDeprecation
         public decimal? CVSSRating { get; set; }
```
scottbommarito: We should make this `CvssRating`.

xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6875/commits/7a6773f8884cfab03b2fe9008aedfe45873a8d04

src/NuGet.Services.Entities/CVE.cs
xavierdecoster: As discussed (offline) with @skofman1: we'll go with the referential integrity approach in the DB using technical keys for now. 
If one day we need to, we can easily change to JSON arrays containing the actual IDs (instead of keys). The other way around is harder.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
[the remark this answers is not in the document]
xavierdecoster: Looking at the UX design, we can limit the descriptions we store to ca. 2 lines of text, which corresponds to **max 300 chars**. This will also reduce memory pressure and storage size.

The UI will expose hyperlinks to the actual CVE/CWE details pages (hosted by MITRE/NIST), which will provide full details and extended descriptions.

src/NuGet.Services.Entities/CVE.cs
```diff
@@ -0,0 +1,59 @@
...
+        //  * https://cve.mitre.org/cve/identifiers/syntaxchange.html
+        //  * https://cve.mitre.org/about/faqs.html#what_is_cve_id
+        [Index(IsUnique = true)]
+        [Required]
+        [MaxLength(20)]
+        public string CveId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the description of the CVE.
+        /// The description is a plain language field that describes the vulnerability with sufficient detail as to demonstrate that the vulnerability is unique.
```
skofman1: please add to the comment that the description is intentionally truncated

src/NuGet.Services.Entities/CVE.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace NuGet.Services.Entities
+{
+    /// <summary>
+    /// Represents a Common Vulnerability and Exposure (CVE).
+    /// </summary>
+    public class Cve
```
skofman1: >e [](start = 19, length = 1)

nit: file name is still capitalized

src/NuGet.Services.Entities/CWE.cs
```diff
@@ -0,0 +1,50 @@
...
+        [MaxLength(200)]
+        [Required]
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Gets or sets the description of the CWE.
+        /// </summary>
+        [MaxLength(300)]
+        [Required]
+        public string Description { get; set; }
```
skofman1: Is the description here truncated or it can't be more then 300 by definition of CWE

src/NuGet.Services.Entities/CVE.cs
xavierdecoster: Had to run `git mv` to fix this, as the file was properly capitalized on my machine. Thanks for spotting!

src/NuGet.Services.Entities/CVE.cs
xavierdecoster: Added.

src/NuGet.Services.Entities/CWE.cs
xavierdecoster: Intentionally truncated as well (could be up to 600 chars, which is also way too much). Tested and we need approx. 200-250 chars to show two lines of text. Being on the safe side, I went with max 300 chars, which we can reduce further if appropriate.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
xavierdecoster: Verified delete behavior:
* when deleting a `PackageDeprecation`, it properly deletes the corresponding rows in the `PackageDeprecationCwe` and `PackageDeprecationCve` tables, but doesn't delete the `CWE` or `CVE` record.
* when deleting a `CVE` or `CWE`, it properly deletes the corresponding rows in the `PackageDeprecationCwe` and `PackageDeprecationCve` tables, but doesn't delete the `PackageDeprecation` record.

However, we should never delete `CVE` or `CWE` records (adding a `Status` column to those entities for when a CVE/CWE gets rejected after publication, so we can filter them out).

tests/NuGetGallery.Facts/TestUtils/FakeEntitiesContext.cs
```diff
@@ -147,6 +147,29 @@ public IDbSet<SymbolPackage> SymbolPackages
...
+            get
+            {
+                return Set<Cve>();
+            }
+            set
+            {
+                throw new NotSupportedException();
+            }
+        }
+        public IDbSet<Cwe> Cwes
```
joelverhagen: nit: line between

xavierdecoster: Fixed.

xavierdecoster:
For **CVEs**, I added a Boolean column `Rejected` to the entity model, to indicate whether a CVE has been rejected after it was first published (a scenario we may encounter during data refresh, in which case we flag the CVE but don’t delete it - aka "soft-delete").
CVEs available in the NVD database may have [another status](https://nvd.nist.gov/vuln), but except for `REJECT`, I don't think we care, so this is all we need to cover for. (this is in fact a “soft-delete” in the NVD source data feed)

I noticed **CWEs** may have a varying status in the source data as well. 
The ‘Development View’ has three types of CWEs in the latest data set:
•	Incomplete
•	Draft
•	Usable

Looking at the [StatusEnumeration spec](https://cwe.mitre.org/documents/schema/#StatusEnumeration), it’s a bit unclear whether we should reduce the dataset or not…
As the external data source seems to simply list them all, no matter their status, some of these may be used by the public even though they have a status of incomplete or obsolete…
The status enumeration may be subject to change, but it appears that CWE is also taking the same approach of “soft-deleting” CWEs by marking them as `Deprecated`.
Which leads me to think we don’t care about CWE status at all, except for “Deprecated”, which could be an additional column on the CWE entity (similar to the “Rejected” column on CVE).

From usability point of view, we simply need to be able to _hide_ the CVE/CWE items that are _unpublished_, without deleting potentially existing references to `PackageDeprecation`(s). Having a simple Boolean instead of an actual status column that is subject to change seems like the simplest approach here.

Any thoughts? @scottbommarito @skofman1 @anangaur @joelverhagen @loic-sharma

joelverhagen:
> Any thoughts?

Perhaps we can just have a `Listed` boolean on both tables which controls visibility and this value gets populated by any column or combinations of columns on `Cve` and `Cwe`? The reason I think this is beneficial is that it moved the filtering-out logic to the actor that populates the table instead of the temptation of having some complex or domain-specific filtering in the gallery. Gallery could just query for `Listed = true` .

xavierdecoster:
@joelverhagen that's exactly what I'm aiming for! I could name them both `Listed`.

scottbommarito:
@joelverhagen - I think whether or not `Listed` is a good idea depends on the design of the job that inserts into the table. If we ever change the definition of what should show in the table or not, the actor that inserts into the table will have to be updated to reprocess existing entries to change the listed property or a manual query will need to be run. I would steer away from any approach that requires manual queries.

I think regardless of this, we should store all information that would help us determine whether or not the CVE/CWE is relevant, e.g. `StatusEnumeration` for CWE, to prevent needing to fetch the entire set of CVEs/CWEs again just to update the listed status.

xavierdecoster:
@scottbommarito the data update process for CWE requires us to fetch the entire data set every time anyway... (a CSV file download), so if Status changed, it will be processed (and if visibility logic needs changing, the job can be updated, and it will then be processed on the next iteration).

For CVE, there's a `Modified` endpoint we can use to query for changes and update only those records. 
The CVE data feeds do NOT expose a separate `Status` column. Instead, they use some prefix in the `Description` field (e.g. `** REJECT ** Description` or ``** DISPUTE ** Description`). Which means the `Description` will change and be processed accordingly as well. We already store the descriptions for CWE and CVE in the database, so the status is part of the data.

No matter what changes to status or visibility, if the initial import considered the data as _relevant_ at that time, it's going to remain relevant forever (only soft-deletes allowed - aka unlisting). 
Once a CWE or CVE ID has been published in the external data sources, it's out there forever too, no matter the status. They only recommend us to _hide_ Rejected/Deprecated status.

scottbommarito:
@xavierdecoster - perfect. `Listed` sounds great to me!

xavierdecoster:
Almost there. May need to track CVSS scores provided with CVE data. Once confirmed, I'll add the property and update the migration, and merge this PR.

Issue #6786, opened 2019-01-04
[Feature Flags] Admin panel


loic-sharma:
Create a simple admin panel that can update the feature flags.

PR #6877, opened 2019-02-04
[Feature Flags] Add an admin panel

loic-sharma:
image

Addresses https://github.com/NuGet/NuGetGallery/issues/6786


src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -0,0 +1,67 @@
...
+                case FeatureFlagSaveResult.Conflict:
+                    TempData["ErrorMessage"] = "The feature flags were modified by someone else. Please try again.";
+                    break;
+
+                default:
+                    TempData["ErrorMessage"] = $"Unknown save result '{result}'.";
+                    break;
+            }
+
+            return Redirect(Url.Action(actionName: "Index", controllerName: "Features"));
```
joelverhagen: If the POST fails due to invalid JSON, the text box should have the JSON that was POSTed, not the existing JSON. This would allow the admin to correct their mistake.

Also, consider exposing the `JsonException` message.

Since this UI provides basically no help writing a good JSON body, it should at least help debug the problem.

joelverhagen:
I think in an ideal world this would be a fully featured UI where you can add, remove, disable, enable features and flights. Are we trying to minimize cost here?

I personally am comfortable enough with this UI. It's not great for perhaps a new member of the team but we can probably roll with it.

I am curious what other members of the team think about such a minimal approach. My vote is ":shipit:". We have more valuable things to spend our time on that really help our users. This is just my voice though.

I am curious about the accessibility of this. Does this really meet a "B" grade? I presume this is not as good as editing JSON in, say, VS Code.

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
loic-sharma: > If the POST fails due to invalid JSON, the text box should have the JSON that was POSTed, not the existing JSON. This would allow the admin to correct their mistake.

Yes, this is handled by https://github.com/NuGet/NuGetGallery/pull/6877/files#diff-4a08f1899501ce890579ed7700b9cf57R39

> Also, consider exposing the `JsonException` message.

Good idea, I'll add that in

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            var result = await _storage.TrySaveAsync(model.Flags, model.ContentId);
+
+            switch (result)
+            {
+                case FeatureFlagSaveResult.Ok:
+                    TempData["Message"] = "Your feature flags have been saved";
+                    break;
+
+                case FeatureFlagSaveResult.Invalid:
```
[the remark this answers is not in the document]
loic-sharma: It should never be hit, hence the comment on line 51. This request validates the JSON twice:

1. The form's validation checks that the JSON is well formed (line 37). If it isn't, it displays the form with the inputted JSON and an error message (line 39).
2. The `TrySaveAsync` method accepts a raw JSON string but ensures it is well formed. If the contents are incorrect, it returns  `FeatureFlagSaveResult.Invalid`.

loic-sharma:
> I think in an ideal world this would be a fully featured UI where you can add, remove, disable, enable features and flights. Are we trying to minimize cost here?

That's correct. This was called out explicitly in the design review, but I'd love to hear the team's feedback on this.
 
> I am curious about the accessibility of this. Does this really meet a "B" grade? I presume this is not as good as editing JSON in, say, VS Code.

That's a good question. I'll follow up on this.

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -29,5 +30,59 @@ public async Task<FeatureFlags> GetAsync()
...
+            {
+                return FeatureFlagSaveResult.Conflict;
+            }
+        }
+
+        public static bool IsValidFlagsJson(string flags)
+        {
+            try
+            {
+                JsonConvert.DeserializeObject<FeatureFlags>(flags);
```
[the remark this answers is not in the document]
loic-sharma: Self notes: default serialization is to integer, look into StringEnumConvertrer

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -0,0 +1,67 @@
...
+                    // This case shouldn't happen as the ModelState should be invalid.
+                    TempData["ErrorMessage"] = "Could not save feature flags as they were malformed.";
+                    break;
+
+                case FeatureFlagSaveResult.Conflict:
+                    TempData["ErrorMessage"] = "The feature flags were modified by someone else. Please try again.";
```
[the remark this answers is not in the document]
loic-sharma: Agreed. Will look into this as part of validation improvements

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
loic-sharma: Both of these are fixed now by the latest revision

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -29,5 +30,59 @@ public async Task<FeatureFlags> GetAsync()
...
         }
+
+        public async Task<FeatureFlagReference> GetReferenceAsync()
+        {
+            var reference = await _storage.GetFileReferenceAsync(CoreConstants.Folders.ContentFolderName, CoreConstants.FeatureFlagsFileName);
+
+            using (var stream = reference.OpenRead())
+            using (var streamReader = new StreamReader(stream))
+            {
+                return new FeatureFlagReference(streamReader.ReadToEnd(), reference.ContentId);
```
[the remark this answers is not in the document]
loic-sharma: Saving/loading the blob will now format the JSON

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -29,5 +30,59 @@ public async Task<FeatureFlags> GetAsync()
...
+            try
+            {
+                using (var stream = new MemoryStream())
+                using (var writer = new StreamWriter(stream))
+                {
+                    writer.Write(flags);
+                    writer.Flush();
+                    stream.Position = 0;
+
+                    await _storage.SaveFileAsync(CoreConstants.Folders.ContentFolderName, CoreConstants.FeatureFlagsFileName, stream, accessCondition);
```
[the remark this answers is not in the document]
loic-sharma: Fixed

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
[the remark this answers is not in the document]
loic-sharma: Yes it does. The enums are properly serialized into their string representation (see the unit tests).

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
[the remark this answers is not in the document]
loic-sharma: I've changed the experience such that now the form displays your unsaved changes. The form will have an error message that notifies you that need to refresh the page to update the feature flags. The user will need to copy their changes and merge their changes into the latest JSON. Please let me know if you have any additional suggestions to improve the experience here!

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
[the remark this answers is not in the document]
loic-sharma: I've simplified the model validations such that it no longer covers the same validations.

loic-sharma:
TODO: Respect readonly mode config

scottbommarito:
@loic-sharma - if you can't login because it's read-only, you can't even get to this page, so it doesn't need to check the config.

src/NuGetGallery.Core/Features/IMutableFeatureFlagStorageService.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using NuGet.Services.FeatureFlags;
+
+namespace NuGetGallery.Features
+{
+    public interface IMutableFeatureFlagStorageService : IFeatureFlagStorageService
```
scottbommarito: I'm not used to seeing `mutable` on its own, and the "I" in front because it's an interface made me think `immutable` instead. May be best to change this name. `editable` maybe?

src/NuGetGallery.Core/Features/FeatureFlagReference.cs
```diff
@@ -0,0 +1,39 @@
...
+            }
+
+            Flags = flags;
+            ContentId = contentId;
+        }
+
+        /// <summary>
+        /// The feature flag's content, serialized as JSON.
+        /// </summary>
+        public string Flags { get; }
```
scottbommarito: It's not super clear that this is the `JSON` feature flag content.

Maybe `FlagsJson` or just `Json` instead?

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -29,5 +35,74 @@ public async Task<FeatureFlags> GetAsync()
...
+
+        public async Task<FeatureFlagReference> GetReferenceAsync()
+        {
+            var reference = await _storage.GetFileReferenceAsync(CoreConstants.Folders.ContentFolderName, CoreConstants.FeatureFlagsFileName);
+
+            string json;
+            using (var stream = reference.OpenRead())
+            using (var streamReader = new StreamReader(stream))
+            {
+                json = streamReader.ReadToEnd();
```
joelverhagen: `ReadToEndAsync`?

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -0,0 +1,64 @@
...
+                var result = await _storage.TrySaveAsync(model.Flags, model.ContentId);
+
+                switch (result.Type)
+                {
+                    case FeatureFlagSaveResultType.Ok:
+                        TempData["Message"] = "Your feature flags have been saved!";
+                        return Redirect(Url.Action(actionName: "Index", controllerName: "Features"));
+
+                    case FeatureFlagSaveResultType.Conflict:
+                        TempData["ErrorMessage"] = "The feature flags were modified by someone else. Please reload the page and try again.";
```
joelverhagen: Maybe state that "Your changes were not applied."

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -0,0 +1,64 @@
...
+                Flags = reference.Flags,
+                ContentId = reference.ContentId
+            });
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<ActionResult> Index(FeatureFlagsViewModel model)
+        {
+            if (ModelState.IsValid)
```
scottbommarito: No message if the model state is invalid?

Does the view automatically show the model errors?

src/NuGetGallery/Areas/Admin/Views/Features/Index.cshtml
```diff
@@ -0,0 +1,33 @@
+﻿@using NuGetGallery
+@model FeatureFlagsViewModel
+@{
+    ViewBag.Title = "Feature Flags";
+}
+
+<section role="main" class="container main-container page-revalidation">
+    <div class="row">
```
scottbommarito: No need for these two `div`s. They cancel each other out. You only need to use `row` when u intend to do something special with the columns.

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
```diff
@@ -135,12 +135,23 @@
...
+                    <span>Delete Account</span>
                 </a>
             </h2>
             <p>
                 Manage Account Deletion
             </p>
         </li>
+        <li>
+            <h2>
+                <a href="@Url.Action(actionName: "Index", controllerName: "Features")">
```
scottbommarito: You don't HAVE to do this...but maybe we could make a `@helper` here that you pass in the url, header, and description to, and it spits out the list item for the admin page? It could substantially reduce the HTML of this page.

tests/NuGetGallery.Core.Facts/Features/FeatureFlagFileStorageServiceFacts.cs
```diff
@@ -0,0 +1,304 @@
...
+                        It.IsAny<IAccessCondition>()),
+                    Times.Never);
+            }
+
+            public static IEnumerable<object[]> ReturnsInvalidData()
+            {
+                yield return new object[]
+                {
+                    "bad",
+                    "Unexpected character encountered while parsing value: b"
```
scottbommarito: Not sure if it's necessary to test the exact messages here. We're essentially testing `JSON.NET`'s behavior with this, and if we update the version, someone may need to update this tests, which is probably fine.

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
loic-sharma: I'm okay with less abstractions here

src/NuGetGallery.Core/Features/IMutableFeatureFlagStorageService.cs
loic-sharma: Fixed

tests/NuGetGallery.Core.Facts/Features/FeatureFlagFileStorageServiceFacts.cs
loic-sharma: We display these error messages to the user, so I think there's value in ensuring the errors are decent. I would expect that fixing any changes wouldn't too terribly hard, so I'm leaning towards keeping these tests.

That being said, I completely agree that testing Newtonsoft is unnecessary. Let me know if there are specific tests you think are frivolous.

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
loic-sharma: Fixed

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
loic-sharma: Added

src/NuGetGallery.Core/Features/FeatureFlagReference.cs
loic-sharma: Updated

src/NuGetGallery/Areas/Admin/Views/Features/Index.cshtml
loic-sharma: Thanks for the tip, fixed :)

Issue #6849, opened 2019-01-26
Add "alternate package" autocomplete for Deprecation section of Manage Package page


scottbommarito:
When the user types in an alternate package ID, make an AJAX request with the package ID to get the list of versions. Show the list of versions as options as well as "Any Version".

If the package does not exist, show an error in the UI.

Issue #6848, opened 2019-01-26
Implement "multi-select dropdown" or alternative for Deprecation section of Manage Package page


scottbommarito:
Either
1. Custom multi-select dropdown
2. Expandable panel (similar to API keys page)
3. Find open-source solution

scottbommarito:
We are going with 1, as it was fortunately not too complex.

Current keyboard behavior:

clicking header when dropdown is collapsed expands dropdown and focuses on first option
clicking away when dropdown is expanded collapses it

when focused on header and dropdown is collapsed, space or enter expands dropdown and focuses on first option
when focused on header and dropdown is expanded, space or enter collapses dropdown

up/down to toggle between options
typing into search box immediately filters the results

escape collapses dropdown

tabbing while inside dropdown closes it and proceeds to the next form element
shift-tabbing while inside dropdown focuses on the header

PR #6878, opened 2019-02-06
Initial work on the Deprecate section of the Manage Package page

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6773
https://github.com/NuGet/NuGetGallery/issues/6848
https://github.com/NuGet/NuGetGallery/issues/6849

Pending work:
- CVE/CWE autocomplete (probably not in this PR)
- tooltips (probably also in a separate PR)
- how do we handle deprecating multiple versions with different existing states?
- tests for gallery endpoints (need to decide on multiple version behavior first)
- run Keros/do a thorough accessibility scan (potentially also in a separate PR)

Let me know if you'd like me to break this PR up further.

Tons of screenshots below:

image

image

image

image

image

image

image

image

image

image

image


src/NuGet.Services.Entities/Package.cs
```diff
@@ -29,7 +29,7 @@ public Package()
         }
 #pragma warning restore 618
 
-        public PackageRegistration PackageRegistration { get; set; }
+        public virtual PackageRegistration PackageRegistration { get; set; }
```
scottbommarito: This enables lazy loading.

I needed to do this because in `DisplayPackageViewModel`, I get the `AlternatePackage` of a `PackageDeprecation`, and if I don't make this allow lazy loading, `AlternatePackage.PackageRegistration` returns null and causes some strange behavior.

src/NuGet.Services.Entities/Package.cs
```diff
@@ -220,7 +220,7 @@ public bool HasReadMe
         /// The logged in user when this package version was created.
         /// NULL for older packages.
         /// </summary>
-        public User User { get; set; }
+        public virtual User User { get; set; }
```
scottbommarito: Added this for consistency--willing to remove if there's push-back.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1296,6 +1296,115 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+            return Json(HttpStatusCode.OK);
+        }
+
+        private JsonResult DeprecateErrorResponse(HttpStatusCode code, string error)
+        {
+            return Json(HttpStatusCode.BadRequest, new { error });
+        }
+
+        [HttpGet]
+        public virtual JsonResult GetDeprecationAlternatePackageVersions(string id)
```
scottbommarito: Technically I could reuse the existing OData or search APIs, but I thought adding a specific endpoint for the purpose of finding alternate package versions would be okay, especially in case we want special behavior (e.g. you can't choose an alternate version that is deprecated itself).

Also should I move this into `JsonApiController`? I left it here so that it's close to the `Deprecate` endpoint but I could move this or both of these endpoints into the other controller.

src/NuGetGallery/Extensions/PackageDeprecationExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public static class PackageDeprecationExtensions
```
scottbommarito: This class will likely be removed completely due to https://github.com/NuGet/NuGetGallery/pull/6875

src/NuGetGallery/Controllers/PackagesController.cs
xavierdecoster: FYI, the CWE and CVE autocomplete endpoints will likely be added to the `JsonApiController`. Perhaps it makes sense for this one to move there as well, although I don't feel too strongly about it.

src/NuGetGallery/Extensions/PackageDeprecationExtensions.cs
xavierdecoster: Yep, taking the referential integrity approach in the db model.

src/NuGet.Services.Entities/Package.cs
loic-sharma: Lazy loading is scary. I'd lean towards not adding virtual unless absolutely necessary

---
In reply to: [254480895](https://github.com/NuGet/NuGetGallery/pull/6878#discussion_r254480895) [](ancestors = 254480895)

src/NuGet.Services.Entities/Package.cs
loic-sharma: Talked to Scott offline. This should be possible without lazy loading.

---
In reply to: [254480723](https://github.com/NuGet/NuGetGallery/pull/6878#discussion_r254480723) [](ancestors = 254480723)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1312,6 +1312,115 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
 
             return View(model);
         }
+        
+        [HttpPost]
+        [RequiresAccountConfirmation("delete a package")]
+        [ValidateAntiForgeryToken]
+        public virtual async Task<JsonResult> Deprecate(
+            string id, 
+            IEnumerable<string> versions,
```
loic-sharma: >IEnumerable [](start = 12, length = 11)

Can the `IEnumerable` here be `IReadOnlyList`?

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -81,6 +81,23 @@ private DisplayPackageViewModel(Package package, User currentUser, string pushed
...
+                var deprecation = package.Deprecations.Single();
+                CVEIds = deprecation.GetCVEIds();
+                CVSSRating = deprecation.CVSSRating;
+                CWEIds = deprecation.GetCWEIds();
+                AlternatePackageRegistrationId = deprecation.AlternatePackageRegistration?.Id;
+
+                var alternatePackage = deprecation.AlternatePackage;
+                if (alternatePackage != null)
+                {
+                    AlternatePackage = new DisplayPackageViewModel(alternatePackage, currentUser, GetPushedBy(alternatePackage, currentUser));
```
scottbommarito: note to self: verify this can't loop infinitely

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1312,6 +1312,115 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
 
             return View(model);
         }
+        
+        [HttpPost]
+        [RequiresAccountConfirmation("delete a package")]
```
loic-sharma: >delete [](start = 38, length = 6)

`delete` -> `deprecate`

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1312,6 +1312,115 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+                    }
+                }
+            }
+
+            var missingVersions = new List<string>();
+            var foundPackageVersions = new List<Package>();
+            foreach (var version in versions)
+            {
+                var package = registration.Packages.SingleOrDefault(v => v.NormalizedVersion == NuGetVersionFormatter.Normalize(version));
+                if (package == null)
```
loic-sharma: If a package owner wants to deprecate all 1,000 versions of their packages, will this kick off 1,000 queries? Do we want to add limits to protect our database?

scottbommarito: No, `registration.Packages` should be loaded in the query to get `registration`. I will verify this works great with lots of versions though.

src/NuGetGallery/Services/IPackageService.cs
```diff
@@ -104,5 +104,24 @@ public interface IPackageService : ICorePackageService
...
         Task SetRequiredSignerAsync(PackageRegistration registration, User signer, bool commitChanges = true);
+
+        /// <summary>
+        /// Updates the deprecation of many packages.
+        /// If any packages have an existing deprecation, it combines the existing state with the new state.
+        /// If only a single package is provided, the existing deprecation is entirely replaced.
+        /// Commits changes when finished.
+        /// </summary>
+        Task UpdateDeprecation(
+            IEnumerable<Package> packages,
```
loic-sharma: >IEnumerable [](start = 12, length = 11)

Please use `IReadonlyList` wherever possible instead of `IEnumerable`

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -826,5 +830,201 @@ public async Task SetRequiredSignerAsync(PackageRegistration registration, User
...
+            bool isOther, 
+            IEnumerable<string> cveIds, 
+            decimal? cvssRating, 
+            IEnumerable<string> cweIds,
+            PackageRegistration alternatePackageRegistration,
+            Package alternatePackage,
+            string customMessage, 
+            bool shouldUnlist)
+        {
+            Action<Package, bool, bool, bool, IEnumerable<string>, decimal?, IEnumerable<string>, PackageRegistration, Package, string, bool> deprecatePackage;
```
loic-sharma: >Action<Package, bool, bool, bool, IEnumerable<string>, decimal?, IEnumerable<string>, PackageRegistration, Package, string, bool> [](start = 12, length = 129)

This is a little long. Would it be possible to use a delegate instead? [https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/delegates/](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/delegates/)

src/NuGet.Services.Entities/Package.cs
scottbommarito: Fixed without lazy-loading as discussed.

src/NuGet.Services.Entities/Package.cs
scottbommarito: Removed.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I will align with your work then.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -826,5 +830,201 @@ public async Task SetRequiredSignerAsync(PackageRegistration registration, User
...
+                {
+                    deprecation.CustomMessage = customMessage;
+                }
+
+                if (shouldUnlist)
+                {
+                    package.Listed = false;
+                }
+
+            }
```
loic-sharma: >} [](start = 12, length = 1)

There's an extra space above this line

src/NuGetGallery/Services/IPackageService.cs
```diff
@@ -104,5 +104,24 @@ public interface IPackageService : ICorePackageService
...
         /// is <c>null</c>.</exception>
         Task SetRequiredSignerAsync(PackageRegistration registration, User signer, bool commitChanges = true);
+
+        /// <summary>
+        /// Updates the deprecation of many packages.
+        /// If any packages have an existing deprecation, it combines the existing state with the new state.
+        /// If only a single package is provided, the existing deprecation is entirely replaced.
+        /// Commits changes when finished.
+        /// </summary>
+        Task UpdateDeprecation(
```
loic-sharma: >UpdateDeprecation [](start = 13, length = 17)

If `isVulnerable`, `isLegacy`, and `isOther` are all false, this is supposed to delete the deprecations? Could we make that explicit?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1312,6 +1312,115 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+            decimal? cvssRating,
+            IEnumerable<string> cweIds,
+            string alternatePackageId,
+            string alternatePackageVersion,
+            string customMessage,
+            bool shouldUnlist)
+        {
+            if (versions == null || !versions.Any())
+            {
+                return DeprecateErrorResponse(HttpStatusCode.BadRequest, "You must select at least one version to deprecate!");
```
skofman1: >"You must select at least one version to deprecate! [](start = 73, length = 51)

any user visible message should be in Strings

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1312,6 +1312,115 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+            }
+
+            var registration = _packageService.FindPackageRegistrationById(id);
+            if (registration == null)
+            {
+                return DeprecateErrorResponse(HttpStatusCode.NotFound, $"Package '{id}' does not exist.");
```
skofman1: same

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -826,5 +830,201 @@ public async Task SetRequiredSignerAsync(PackageRegistration registration, User
...
+            IEnumerable<string> cveIds, 
+            decimal? cvssRating, 
+            IEnumerable<string> cweIds,
+            PackageRegistration alternatePackageRegistration,
+            Package alternatePackage,
+            string customMessage, 
+            bool shouldUnlist)
+        {
+            Action<Package, bool, bool, bool, IEnumerable<string>, decimal?, IEnumerable<string>, PackageRegistration, Package, string, bool> deprecatePackage;
+            if (packages.Count() == 1)
```
scottbommarito: Current implementation for updating deprecations is:

if one package is selected, replace the deprecation
if multiple packages are selected, combine the new state with the existing state

This is likely not the final implementation and is pending feedback from @anangaur / UI team

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1312,6 +1312,115 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+                {
+                    alternatePackageRegistration = _packageService.FindPackageRegistrationById(alternatePackageId);
+                    if (alternatePackageRegistration == null)
+                    {
+                        return DeprecateErrorResponse(HttpStatusCode.NotFound, $"Alternate package '{alternatePackageId}' does not exist.");
+                    }
+                }
+            }
+
+            var missingVersions = new List<string>();
```
skofman1: why do you need "missingVersions" if it's never used

scottbommarito: Woops, at some point I had code that returns an error if there are any `missingVersions`. I don't know what happened to it but I will add it back.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1312,6 +1312,115 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+            return Json(HttpStatusCode.OK);
+        }
+
+        private JsonResult DeprecateErrorResponse(HttpStatusCode code, string error)
+        {
+            return Json(HttpStatusCode.BadRequest, new { error });
+        }
+
+        [HttpGet]
+        public virtual JsonResult GetDeprecationAlternatePackageVersions(string id)
```
skofman1: Let's add UIAuthorize to be on the safe side

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1312,6 +1312,115 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+        {
+            var registration = _packageService.FindPackageRegistrationById(id);
+            if (registration == null)
+            {
+                return Json(HttpStatusCode.NotFound, null, JsonRequestBehavior.AllowGet);
+            }
+
+            var packageVersions = registration.Packages
```
skofman1: how did you test performance?

src/NuGetGallery/Services/IPackageService.cs
```diff
@@ -104,5 +104,24 @@ public interface IPackageService : ICorePackageService
         /// <exception cref="ArgumentNullException">Thrown if <paramref name="registration" />
         /// is <c>null</c>.</exception>
         Task SetRequiredSignerAsync(PackageRegistration registration, User signer, bool commitChanges = true);
+
+        /// <summary>
+        /// Updates the deprecation of many packages.
+        /// If any packages have an existing deprecation, it combines the existing state with the new state.
```
skofman1: > it combines the existing state with the new state. [](start = 57, length = 51)

are we sure we need to combine state and not override existing state?

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -81,6 +81,23 @@ private DisplayPackageViewModel(Package package, User currentUser, string pushed
                     LicenseNames = licenseNames.Split(',').Select(l => l.Trim());
                 }
             }
+
+            if (DeprecationStatus != PackageDeprecationStatus.NotDeprecated)
```
skofman1: why not add this to ManagePackageViewModel?

src/NuGetGallery/Controllers/PackagesController.cs

src/NuGetGallery/Controllers/PackagesController.cs

src/NuGetGallery/Services/IPackageService.cs
scottbommarito: done

src/NuGetGallery/Services/PackageService.cs
scottbommarito: done

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1312,6 +1312,115 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+                foundPackageVersions, 
+                isVulnerable, 
+                isLegacy, 
+                isOther, 
+                cveIds, 
+                cvssRating, 
+                cweIds, 
+                alternatePackageRegistration, 
+                alternatePackage, 
+                customMessage,
```
skofman1: do we need to do anything with the custom message to prevent SQL injection?

src/NuGetGallery/Services/PackageService.cs
scottbommarito: done

src/NuGetGallery/Services/IPackageService.cs
scottbommarito: Added to xmldoc.
I will confirm with @anangaur that this is the behavior we want.

src/NuGetGallery/Controllers/PackagesController.cs

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: fixed

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I have a package on my localhost with 1000 versions and this endpoint is near instantaneous.

src/NuGetGallery/Controllers/PackagesController.cs
skofman1: ok. For extra safety please compare with other places in the code where we store user input in DB.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: It was not unfortunately. I've made some quick perf improvements, but it is still looking at roughly 35 seconds to deprecate 1000 packages. Will check and see what improvements I can make.

scottbommarito: Fixed the issue--there is a separate method you must use to bulk insert/delete rows in EF.

src/NuGetGallery/Controllers/PackagesController.cs

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Decided to leave `Deprecate` on `PackagesController` because it makes more sense to me there--the other AJAX post endpoints for modifying packages are there.

src/NuGetGallery/Services/IPackageService.cs
loic-sharma: Oh I meant, could we add another parameter for the `isDelete` case? If these parameters are exclusive, consider using an enum.

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -318,5 +319,29 @@ public ManagePackageOwnerModel(PackageRegistration package, User user, User curr
...
+                return Json(HttpStatusCode.NotFound, null, JsonRequestBehavior.AllowGet);
+            }
+
+            var versions = registration.Packages
+                .Where(p => p.PackageStatusKey == PackageStatus.Available)
+                .ToList()
+                .OrderByDescending(p => NuGetVersion.Parse(p.Version))
+                .Select(p => NuGetVersionFormatter.ToFullStringOrFallback(p.Version, p.Version));
+
+            if (!versions.Any())
```
loic-sharma: It seems that doing `Any()` before `ToList()` will cause this query to run twice. I would call `ToList()` immediately after the `Select` on line 337.

src/NuGetGallery/Controllers/PackagesController.cs
loic-sharma: It looks like you missed this method :)

src/NuGetGallery/Controllers/PackagesController.cs
xavierdecoster: Sounds good!

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I tried, but ASP.NET doesn't seem to support it. For some reason, it seems to understands how to create a `IEnumerable<string>` from the JSON data in the request, but throws an exception when creating an `IReadOnlyCollection<string>`. I don't remember if I tried `IReadOnlyList<string>` as well. I thought that it wasn't too big of a deal so I thought it'd be fine to leave it like this.

src/NuGetGallery/Services/IPackageService.cs
scottbommarito: Good call! Will make this method accept an `enum`.

src/NuGetGallery/Controllers/JsonApiController.cs
scottbommarito: There's already a `ToList()` in the versions query, so there's no SQL query being done here.

src/NuGetGallery/Extensions/PackageDeprecationExtensions.cs
scottbommarito: Removed it!

src/NuGetGallery.Core/Services/CorePackageService.cs
```diff
@@ -293,6 +293,10 @@ protected IQueryable<Package> GetPackagesByIdQueryable(string id)
                 .Include(p => p.PackageRegistration)
                 .Include(p => p.User)
                 .Include(p => p.SymbolPackages)
+                .Include(p => p.Deprecations)
```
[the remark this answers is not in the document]
scottbommarito: Will remove. I will make a new `PackageService` method for `DisplayPackage` and `Manage` to use.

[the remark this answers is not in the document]
scottbommarito: I'm having difficulty doing this in a way that doesn't complicate the code. I will do a more thorough look into the perf implications. I will also take a look into ways to improve the perf of the DisplayPackage page, which is affected by this change and will need to query deprecation information.

src/NuGetGallery.Core/Services/CorePackageService.cs
```diff
@@ -285,18 +285,26 @@ public virtual async Task UpdatePackageSigningCertificateAsync(string packageId,
...
-        protected IQueryable<Package> GetPackagesByIdQueryable(string id)
+        protected IQueryable<Package> GetPackagesByIdQueryable(string id, bool withDeprecations = false)
         {
-            return _packageRepository
+            var packages = _packageRepository
                 .GetAll()
                 .Include(p => p.LicenseReports)
                 .Include(p => p.PackageRegistration)
                 .Include(p => p.User)
                 .Include(p => p.SymbolPackages)
```
joelverhagen: Are symbols or license reports needed by deprecations?

scottbommarito: I didn't want to touch existing `.Include`s. The purpose of my change here was to prevent adding the deprecation info to endpoints that don't need it. We can probably revisit what's necessary later.

src/NuGetGallery/Services/IPackageService.cs
```diff
@@ -17,6 +17,8 @@ namespace NuGetGallery
     /// </summary>
     public interface IPackageService : ICorePackageService
     {
+        IReadOnlyCollection<Package> FindPackagesById(string id, bool withDeprecations = false);
```
scottbommarito: A brief summary of how `IPackageService` is changing:

Both `DisplayPackage` and `Manage` require all versions of a package and all deprecations associated with those packages.

Previously, this was done either implicitly or lazily, specifically:
- if a version isn't specified or is missing, `FindPackageByIdAndVersion` must fetch all versions to calculate the latest one
- if a version was specified and found, the remaining versions would be loaded lazily

This is bad because our perf for these endpoints are relying on a hidden behavior of one of our services, and when this hidden behavior doesn't go our way, lazy loading is **very very very very** slow, and adding the deprecation entities associated with this feature makes it *even* worse.

So, to remedy this
- add `FindPackagesById` to get all packages by ID (with a `withDeprecations` override for when deprecation entities are not needed)
- move logic from `FindPackageByIdAndVersion` that fetches the latest version from a list of packages and move it into `GetLatestPackage`
    - `FindPackageByIdAndVersion` now calls `GetLatestPackage` internally
- make DisplayPackage and Manage call `FindPackagesById` and then `GetLatestPackage` if an exact version isn't found
- remove `FindAbsoluteLatestPackageById` because it was only used by `DisplayPackage` (and we replaced it with the above methods)

src/NuGetGallery/Services/UserService.cs
```diff
@@ -383,11 +382,18 @@ public virtual User FindByKey(int key, bool includeDeleted = false)
             {
                 users = users.Where(u => !u.IsDeleted);
             }
-            return users.Include(u => u.Roles)
-                .Include(u => u.Credentials)
+            return IncludeUserFields(users)
                 .SingleOrDefault(u => u.Key == key);
         }
 
+        private IQueryable<User> IncludeUserFields(IQueryable<User> source)
```
scottbommarito: small perf improvement--the majority of times we fetch a user, we need their organizations.
by including it, we reduce the number of queries made and improve our perf on a lot of endpoints

For example, most authenticated endpoints start by getting the current user and then doing a permissions check. This is currently done in two separate queries--by adding `Organizations` to the `Include`, we get rid of the extra query.

scottbommarito:
I have updated the screenshots above with the latest implementation of the multi-select dropdown.

For a more detailed description of how it functions, see https://github.com/NuGet/NuGetGallery/issues/6848

src/NuGetGallery.Core/Services/CorePackageService.cs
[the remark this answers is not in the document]
scottbommarito: After fooling around with for a while, I settled on the solution described by https://github.com/NuGet/NuGetGallery/pull/6878#discussion_r256443329.

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
scottbommarito: no longer using `DisplayPackageViewModel` so this isn't possible.

scottbommarito:
I've decided to work on splitting this PR up into more pieces--stay tuned.

PR #6881, opened 2019-02-08
Need to fetch the license file from Dev0 rather than Devlegacy for DEV. 

zhhyu:
For DEV blob storage, we have separate locations as "nugetdev0" and "nugetdevlegacy". V3 job will put the package with license file at "nugetdev0". However, gallery will go to "nugetdevlegacy" to fetch the license file, which is not there. So there is a need to let the gallery get the license file from "nugetdev0" in DEV.
Fix: https://github.com/NuGet/Engineering/issues/2119

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -795,7 +795,7 @@ namespace NuGetGallery
 
                 var registration = builder.RegisterType(dependent.ImplementationType)
                     .WithParameter(new ResolvedParameter(
-                       (pi, ctx) => pi.ParameterType == typeof(IFileStorageService),
+                       (pi, ctx) => pi.ParameterType.IsAssignableFrom(typeof(IFileStorageService)),
                        (pi, ctx) => ctx.ResolveKeyed<IFileStorageService>(dependent.BindingKey)))
                     .AsSelf()
                     .As(dependent.InterfaceType);
--- a/src/NuGetGallery/App_Start/StorageDependent.cs
+++ b/src/NuGetGallery/App_Start/StorageDependent.cs
@@ -93,7 +93,7 @@ namespace NuGetGallery
                 Create<UploadFileService, IUploadFileService>(configuration.AzureStorage_Uploads_ConnectionString, isSingleInstance: false),
                 Create<RevalidationStateService, IRevalidationStateService>(configuration.AzureStorage_Revalidation_ConnectionString, isSingleInstance: false),
                 Create<FeatureFlagFileStorageService, IFeatureFlagStorageService>(configuration.AzureStorage_Content_ConnectionString, isSingleInstance: true),
-                Create<CoreLicenseFileService, ICoreLicenseFileService>(configuration.AzureStorage_Flatcontainer_ConnectionString, isSingleInstance: false)
+                Create<CoreLicenseFileService, ICoreLicenseFileService>(configuration.AzureStorage_FlatContainer_ConnectionString, isSingleInstance: false)
             };
 
             var connectionStringToBindingKey = dependents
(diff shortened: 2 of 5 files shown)


src/NuGetGallery/Configuration/AppConfiguration.cs
```diff
@@ -53,6 +53,9 @@ public class AppConfiguration : IAppConfiguration
         [DisplayName("AzureStorage.Packages.ConnectionString")]
         public string AzureStorage_Packages_ConnectionString { get; set; }
 
+        [DisplayName("AzureStorage.Flatcontainer.ConnectionString")]
+        public string AzureStorage_Flatcontainer_ConnectionString { get; set; }
```
joelverhagen: `FlatContainer` (capital C)

src/NuGetGallery/App_Start/StorageDependent.cs
```diff
@@ -91,9 +91,9 @@ public static IEnumerable<StorageDependent> GetAll(IAppConfiguration configurati
...
                 Create<SymbolPackageFileService, ISymbolPackageFileService>(configuration.AzureStorage_Packages_ConnectionString, isSingleInstance: false),
                 Create<UploadFileService, IUploadFileService>(configuration.AzureStorage_Uploads_ConnectionString, isSingleInstance: false),
-                Create<CoreLicenseFileService, ICoreLicenseFileService>(configuration.AzureStorage_Packages_ConnectionString, isSingleInstance: false),
                 Create<RevalidationStateService, IRevalidationStateService>(configuration.AzureStorage_Revalidation_ConnectionString, isSingleInstance: false),
-                Create<FeatureFlagFileStorageService, IFeatureFlagStorageService>(configuration.AzureStorage_Content_ConnectionString, isSingleInstance: true)
+                Create<FeatureFlagFileStorageService, IFeatureFlagStorageService>(configuration.AzureStorage_Content_ConnectionString, isSingleInstance: true),
+                Create<CoreLicenseFileService, ICoreLicenseFileService>(configuration.AzureStorage_Flatcontainer_ConnectionString, isSingleInstance: false)
```
joelverhagen: Are there unit tests around this that need to be enhanced?

joelverhagen:
Please add the configuration to our .cscfg files before merging this PR. The .cscfg files are in our internal NuGetDeployment repository.

In general the least risky approach to adding config is to first add it to the .cscfg then add it to gallery. Gallery doesn't care if there are config names that are unused. In other words, adding to the .cscfgs is backwards compatible.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -795,7 +795,7 @@ private static void ConfigureForAzureStorage(ContainerBuilder builder, IGalleryC
 
                 var registration = builder.RegisterType(dependent.ImplementationType)
                     .WithParameter(new ResolvedParameter(
-                       (pi, ctx) => pi.ParameterType == typeof(IFileStorageService),
+                       (pi, ctx) => pi.ParameterType.IsAssignableFrom(typeof(IFileStorageService)),
```
scottbommarito: Is this necessary? I don't see any changes to `IFileStorageService` in this PR.

Furthermore, we have a couple other instances of similar code in this file that use `==` instead of `IsAssignableFrom`.

src/NuGetGallery/App_Start/StorageDependent.cs
```diff
@@ -91,9 +91,9 @@ public static IEnumerable<StorageDependent> GetAll(IAppConfiguration configurati
...
                 Create<SymbolPackageFileService, ISymbolPackageFileService>(configuration.AzureStorage_Packages_ConnectionString, isSingleInstance: false),
                 Create<UploadFileService, IUploadFileService>(configuration.AzureStorage_Uploads_ConnectionString, isSingleInstance: false),
-                Create<CoreLicenseFileService, ICoreLicenseFileService>(configuration.AzureStorage_Packages_ConnectionString, isSingleInstance: false),
                 Create<RevalidationStateService, IRevalidationStateService>(configuration.AzureStorage_Revalidation_ConnectionString, isSingleInstance: false),
-                Create<FeatureFlagFileStorageService, IFeatureFlagStorageService>(configuration.AzureStorage_Content_ConnectionString, isSingleInstance: true)
+                Create<FeatureFlagFileStorageService, IFeatureFlagStorageService>(configuration.AzureStorage_Content_ConnectionString, isSingleInstance: true),
+                Create<CoreLicenseFileService, ICoreLicenseFileService>(configuration.AzureStorage_FlatContainer_ConnectionString, isSingleInstance: false)
```
scottbommarito: Nit: you could have reduced the changelog by keeping this line in the place where it was and only changing the connection string.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
joelverhagen: `ICoreFileStorageService` needs to match this `ResolvedParameter`. `IsAssignableFrom` does this.

Issue #6787, opened 2019-01-04
[Feature Flags] Update the GDPR export tool


loic-sharma:
* [x] The export tool should export the list of flights that refer to the current user. [#761](https://nuget.visualstudio.com/NuGetMicrosoft/_git/NuGet.Internal.Jobs/pullrequest/761?_a=files)
* [X] ~~Investigate if anything is needed for DSR agent / support requests. Do we use the `DeleteAccountService` anywhere?~~ Automated deletions hasn't been implemented yet. See: https://github.com/NuGet/Engineering/issues/2063

Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2414679

loic-sharma:
Tried to build [export tool](https://nuget.visualstudio.com/NuGetMicrosoft/_build/results?buildId=49447), but [the signing](https://nuget.visualstudio.com/NuGetBuild/_build/results?buildId=49448) failed. It seems that the build is unable to skip already signed win32 DLLs and fails on [this step](https://nuget.visualstudio.com/_git/NuGetBuild?path=%2Fsigning%2FCommon.ps1&version=GBmaster&_a=contents&line=131&lineStyle=plain&lineEnd=132&lineStartColumn=1&lineEndColumn=1). Currently playing around with `SignTool verify /v /pa x.dll`

loic-sharma:
Blocked by https://github.com/NuGet/Engineering/issues/2114

PR #6885, opened 2019-02-12
[Feature Flags] On account deletion, remove user from flights

loic-sharma:
Part of https://github.com/NuGet/NuGetGallery/issues/6787

Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2407655

### DIFF
--- a/src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
+++ b/src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
@@ -20,17 +20,14 @@ namespace NuGetGallery.Features
     {
         private const int MaxRemoveUserAttempts = 3;
 
+        private static readonly JsonSerializer Serializer;
+
         private readonly ICoreFileStorageService _storage;
         private readonly ILogger<FeatureFlagFileStorageService> _logger;
-        private readonly JsonSerializer _serializer;
 
-        public FeatureFlagFileStorageService(
-            ICoreFileStorageService storage,
-            ILogger<FeatureFlagFileStorageService> logger)
+        static FeatureFlagFileStorageService()
         {
-            _storage = storage ?? throw new ArgumentNullException(nameof(storage));
-            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
-            _serializer = JsonSerializer.Create(new JsonSerializerSettings
+            Serializer = JsonSerializer.Create(new JsonSerializerSettings
             {
                 Formatting = Formatting.Indented,
                 MissingMemberHandling = MissingMemberHandling.Error,
@@ -41,13 +38,21 @@ namespace NuGetGallery.Features
             });
         }
 
+        public FeatureFlagFileStorageService(
+            ICoreFileStorageService storage,
+            ILogger<FeatureFlagFileStorageService> logger)
+        {
+            _storage = storage ?? throw new ArgumentNullException(nameof(storage));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
         public async Task<FeatureFlags> GetAsync()
         {
             using (var stream = await _storage.GetFileAsync(CoreConstants.Folders.ContentFolderName, CoreConstants.FeatureFlagsFileName))
             using (var streamReader = new StreamReader(stream))
             using (var reader = new JsonTextReader(streamReader))
             {
-                return _serializer.Deserialize<FeatureFlags>(reader);
+                return Serializer.Deserialize<FeatureFlags>(reader);
             }
         }
 
@@ -76,7 +81,7 @@ namespace NuGetGallery.Features
                 using (var reader = new StringReader(flagsJson))
                 using (var jsonReader = new JsonTextReader(reader))
                 {
-                    flags = _serializer.Deserialize<FeatureFlags>(jsonReader);
+                    flags = Serializer.Deserialize<FeatureFlags>(jsonReader);
                 }
             }
             catch (JsonException e)
@@ -98,7 +103,7 @@ namespace NuGetGallery.Features
                 using (var streamReader = new StreamReader(stream))
                 using (var reader = new JsonTextReader(streamReader))
                 {
-                    flags = _serializer.Deserialize<FeatureFlags>(reader);
+                    flags = Serializer.Deserialize<FeatureFlags>(reader);
                 }
 
                 // Don't update the flags if the user isn't listed in any of the flights.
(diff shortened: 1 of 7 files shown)


src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -83,6 +92,89 @@ public async Task<FeatureFlagSaveResult> TrySaveAsync(string flags, string conte
             }
         }
 
+        public async Task<bool> TryRemoveUserAsync(User user)
+        {
+            for (var attempt = 0; attempt < MaxRemoveUserAttempts; attempt++)
```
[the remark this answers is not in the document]
loic-sharma: > Are we sure we need retries here?

Without retries, removing the user from the feature flags could fail if an admin modifies the flags at the same time. This results in a bad user experience.

> Does the admin UI retry when saving feature flags?

The admin UI does not retry as it is difficult to "merge" two concurrent admin updates to the feature flags. Say two admins add two different features, ideally, the result should be a JSON file that contains both new features.

On the other hand, it is very easy to retry removing a user from the feature flags.

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -83,6 +92,89 @@ public async Task<FeatureFlagSaveResult> TrySaveAsync(string flags, string conte
...
+                                .ToDictionary(
+                                    f => f.Key,
+                                    f => RemoveUser(f.Value, user)));
+                    });
+
+                    return true;
+                }
+                catch (StorageException e) when (e.IsPreconditionFailedException())
+                {
+                    _logger.LogWarning(
```
cristinamanum: How are we notified if the user was failed to be deleted? Maybe we can have a custom metric that we can use for alerting?

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -83,6 +92,89 @@ public async Task<FeatureFlagSaveResult> TrySaveAsync(string flags, string conte
...
+                            flags.Flights
+                                .ToDictionary(
+                                    f => f.Key,
+                                    f => RemoveUser(f.Value, user)));
+                    });
+
+                    return true;
+                }
+                catch (StorageException e) when (e.IsPreconditionFailedException())
+                {
```
joelverhagen: Why not throw and show a helpful message to the admin? User delete is an admin-only flow right?

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
cristinamanum: Please ignore, I see below. The delete action initiated by the admin will fail.

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -83,6 +92,89 @@ public async Task<FeatureFlagSaveResult> TrySaveAsync(string flags, string conte
             }
         }
 
+        public async Task<bool> TryRemoveUserAsync(User user)
+        {
+            for (var attempt = 0; attempt < MaxRemoveUserAttempts; attempt++)
+            {
+                try
+                {
+                    await MaybeUpdateFlagsAsync((FeatureFlags flags) =>
```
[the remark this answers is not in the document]
loic-sharma: Removed the func

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -83,6 +92,89 @@ public async Task<FeatureFlagSaveResult> TrySaveAsync(string flags, string conte
...
+            }
+
+            // The action can cancel the update by returning null.
+            var result = updateAction(flags);
+            if (result == null)
+            {
+                return;
+            }
+
+            var accessCondition = AccessConditionWrapper.GenerateIfMatchCondition(reference.ContentId);
```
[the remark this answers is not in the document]
loic-sharma: I tweaked `TrySaveAsync` so that it could be reused

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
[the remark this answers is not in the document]
scottbommarito: I think two people editing the flags at the same time would be very uncommon, however, at least at the current scale of our team. That makes sense though and I have no problem with keeping the retries in.

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
loic-sharma: Talked to @joelverhagen. I removed the boolean result from `RemoveUserAsync` and made it throw an exception for consistency.

src/NuGetGallery/Strings.resx
```diff
@@ -569,7 +569,7 @@ For more information, please contact '{2}'.</value>
     <value>The account:{0} was already deleted. No action was performed.</value>
```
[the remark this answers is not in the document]
loic-sharma: Fixed

tests/NuGetGallery.Core.Facts/Features/FeatureFlagFileStorageServiceFacts.cs
```diff
@@ -267,37 +261,311 @@ public static IEnumerable<object[]> ReturnsInvalidData()
             }
         }
 
+        public class TryRemoveUserAsync : FactsBase
+        {
+            [Fact]
+            public async Task WhenUserNotInFlags_DoesntSave()
```
joelverhagen: We should have a test that makes sure if there is extra data in the JSON, to just hard fail. Otherwise there will be data loss.

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -2,27 +2,42 @@
...
         private readonly JsonSerializer _serializer;
 
-        public FeatureFlagFileStorageService(ICoreFileStorageService storage)
+        public FeatureFlagFileStorageService(
+            ICoreFileStorageService storage,
+            ILogger<FeatureFlagFileStorageService> logger)
         {
             _storage = storage ?? throw new ArgumentNullException(nameof(storage));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
             _serializer = JsonSerializer.Create(new JsonSerializerSettings
```
joelverhagen: nit: you can make the serializer static readonly

tests/NuGetGallery.Core.Facts/Features/FeatureFlagFileStorageServiceFacts.cs
loic-sharma: Good idea, added

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
loic-sharma: Fixed

Issue #6777, opened 2018-12-20
[Support] Add transform to org FAQ link in discontinuation flow.


skofman1:
Problem 1 - 
login discontinuation flow - login errors are not visible enough.

Problem 2 - 
Password accounts that want to transform to organization don't have clear documentation/UI flow. The user needs to know to go to https://nuget.org/account/transform but it's not documented and there is no UI link.

@anangaur

anangaur:
Does this alleviate Problem # 2?
![image](https://user-images.githubusercontent.com/14800916/50317601-511a3b80-0471-11e9-9383-fc1c6008bb7e.png)

shishirx34:
the problem 2 is for accounts that do not have MSA/AAD linked, old password accounts. after they get a blocking dialog to link to microsoft, they could directly go to the route `/account/transform` which will allow these accounts to transform to organization without linking it to MSA/AAD.

shishirx34:
Or we could reintroduce the link for transforming to organization in the blocking dialog page.

anangaur:
The 1st step is to link MSA/AAD. May be - once this gets linked we pop up "Do you want to transform to an Organization?"

shishirx34:
yeah but the linking to MSA/AAD is an unnecessary option. It takes away users' personal MSA/AAD account to link it and then have to create a new/temp MSA account to do all management. Its a pain.

anangaur:
I agree. Let me think 🤔 over it. We did consider to provide both options initially but wanted to make it less confusing. May be we can achieve both now.

skofman1:
@anangaur , @shishirx34 , any conclusions?

shishirx34:
@anangaur - thoughts here?

shishirx34:
had offline conversation with @anangaur - we came up to agreement that we could modify the pop up dialog and link to FAQ for transforming account to an organization. This issue can track that behavior. The wording is TBD.

PR #6886, opened 2019-02-12
Add FAQ links to account login error messages

shishirx34:
Addresses https://github.com/NuGet/NuGetGallery/issues/6777 https://github.com/NuGet/Engineering/issues/2071 https://github.com/NuGet/Engineering/issues/2066, I ended up dropping the idea that @agr used for validation messages, it is not straight forward to integrate it with `TempData`, main issue being `TempData` is stored in cookies and it can only be stored as strings. The full fledged support would include JSON serialization and Deserialization with cookie support which is a much larger change for addressing this issue. 

This PR introduces a new property `RawErrorMessage` in `TempData` which would be rendered as a RawHtml on the view side. We have to be careful when using this property for showing error messages making sure we do not format user data in this property. It is fine to be used as an absolute links to our documentation.

Also modified the contact page to include links to FAQ for common resolutions

image

image

image

Added FAQ links to contact page:

image

### DIFF
--- a/src/NuGetGallery/Controllers/AuthenticationController.cs
+++ b/src/NuGetGallery/Controllers/AuthenticationController.cs
@@ -500,8 +500,8 @@ namespace NuGetGallery
                 // The identity value contains cookie non-compliant charachters like `<, >`(eg: John Doe <john@doe.com>), and which would be treated 
                 // as an HTML tags, hence these need to be replaced before encoding, the URL characters need to be endcoded.
                 TempData["RawErrorMessage"] = string.Format(Strings.ChangeCredential_Failed,
-                    HttpUtility.UrlEncode(newCredential.Identity.Replace('<', '[').Replace('>', ']')),
-                    HttpUtility.UrlEncode(UriExtensions.GetExternalUrlElement("FAQs page", GalleryConstants.FAQLinks.MSALinkedToAnotherAccount)));
+                    newCredential.Identity.Replace('<', '[').Replace('>', ']'),
+                    UriExtensions.GetExternalUrlElement("FAQs page", GalleryConstants.FAQLinks.MSALinkedToAnotherAccount));
             }
 
             return SafeRedirect(returnUrl);
--- a/src/NuGetGallery/Controllers/AuthenticationController.cs
+++ b/src/NuGetGallery/Controllers/AuthenticationController.cs
@@ -497,7 +497,7 @@ namespace NuGetGallery
             }
             else
             {
-                // The identity value contains cookie non-compliant charachters like `<, >`(eg: John Doe <john@doe.com>), and which would be treated 
+                // The identity value contains cookie non-compliant characters like `<, >`(eg: John Doe <john@doe.com>), and which would be treated 
                 // as an HTML tags, hence these need to be replaced before encoding, the URL characters need to be endcoded.
                 TempData["RawErrorMessage"] = string.Format(Strings.ChangeCredential_Failed,
                     newCredential.Identity.Replace('<', '[').Replace('>', ']'),
--- a/src/NuGetGallery/Controllers/AuthenticationController.cs
+++ b/src/NuGetGallery/Controllers/AuthenticationController.cs
@@ -497,8 +497,11 @@ namespace NuGetGallery
             }
             else
             {
-                // The identity value contains cookie non-compliant characters like `<, >`(eg: John Doe <john@doe.com>), hence these need to be encoded
-                TempData["ErrorMessage"] = string.Format(Strings.ChangeCredential_Failed, HttpUtility.UrlEncode(newCredential.Identity));
+                // The identity value contains cookie non-compliant charachters like `<, >`(eg: John Doe <john@doe.com>), and which would be treated 
+                // as an HTML tags, hence these need to be replaced before encoding, the URL characters need to be endcoded.
+                TempData["RawErrorMessage"] = string.Format(Strings.ChangeCredential_Failed,
+                    HttpUtility.UrlEncode(newCredential.Identity.Replace('<', '[').Replace('>', ']')),
+                    HttpUtility.UrlEncode(UriExtensions.GetExternalUrlElement("FAQs page", GalleryConstants.FAQLinks.MSALinkedToAnotherAccount)));
             }
 
             return SafeRedirect(returnUrl);
--- a/src/NuGetGallery/Controllers/AuthenticationController.cs
+++ b/src/NuGetGallery/Controllers/AuthenticationController.cs
@@ -497,11 +497,11 @@ namespace NuGetGallery
             }
             else
             {
-                // The identity value contains cookie non-compliant characters like `<, >`(eg: John Doe <john@doe.com>), and which would be treated 
-                // as an HTML tags, hence these need to be replaced before encoding, the URL characters need to be endcoded.
+                // The identity value contains cookie non-compliant characters like `<, >`(eg: John Doe <john@doe.com>), 
+                // These need to be replaced so that they are not treated as HTML tags
                 TempData["RawErrorMessage"] = string.Format(Strings.ChangeCredential_Failed,
-                    newCredential.Identity.Replace('<', '[').Replace('>', ']'),
-                    UriExtensions.GetExternalUrlElement("FAQs page", GalleryConstants.FAQLinks.MSALinkedToAnotherAccount));
+                    newCredential.Identity.Replace("<", "&lt;").Replace(">", "&gt;"),
+                    UriExtensions.GetExternalUrlAnchorTag("FAQs page", GalleryConstants.FAQLinks.MSALinkedToAnotherAccount));
             }
 
             return SafeRedirect(returnUrl);
--- a/src/NuGetGallery/GalleryConstants.cs
+++ b/src/NuGetGallery/GalleryConstants.cs
@@ -132,8 +132,13 @@ namespace NuGetGallery
 
         public static class FAQLinks
         {
-            public const string MSALinkedToAnotherAccount = "https://docs.microsoft.com/en-us/nuget/faqs/nuget-faq#microsoft-account-is-linked-with-another-nugetorg-account";
-            public const string EmailLinkedToAnotherMSAAccount = "https://docs.microsoft.com/en-us/nuget/faqs/nuget-faq#signing-in-with-microsoft-account-shows-me-my-email-is-linked-to-another-microsoft-account";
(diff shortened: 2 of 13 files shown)


anangaur:
LGTM

shishirx34:
@anangaur please also verify the text for transform to org FAQ link for password discontinuation flow.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -497,8 +497,11 @@ public virtual async Task<ActionResult> LinkOrChangeExternalCredential(string re
             }
             else
             {
-                // The identity value contains cookie non-compliant characters like `<, >`(eg: John Doe <john@doe.com>), hence these need to be encoded
-                TempData["ErrorMessage"] = string.Format(Strings.ChangeCredential_Failed, HttpUtility.UrlEncode(newCredential.Identity));
+                // The identity value contains cookie non-compliant charachters like `<, >`(eg: John Doe <john@doe.com>), and which would be treated
```
loic-sharma: `charachters` -> `characters`

src/NuGetGallery/Views/Authentication/LinkExternal.cshtml
```diff
@@ -39,7 +39,8 @@
                                 The account with the email @Model.SignIn.UserNameOrEmail is linked to another Microsoft account.
                             </p>
                             <p class="text-center">
-                                If you would like to update the linked Microsoft account you can do so from the account settings page.
+                                Please refer to the <a href="https://docs.microsoft.com/en-us/nuget/faqs/nuget-faq#signing-in-with-microsoft-account-shows-me-my-email-is-linked-to-another-microsoft-account" target="_blank"> FAQs page</a>
```
loic-sharma: There's an extra space between `>` and ` FAQs page`

src/NuGetGallery/Views/Pages/Contact.cshtml
```diff
@@ -14,15 +14,15 @@
     </div>
     <div class="row">
         <div class="@ViewHelpers.GetColumnClasses(ViewBag)">
-            <h2 class="ms-fontSize-xl ms-fontWeight-semibold">Having problems using a specific package?</h2>
+            <h2 class="ms-fontSize-xl ms-fontWeight-semibold">Having problems using a specific package on nuget.org?</h2>
             <p>
-                If you're having trouble installing a specific package, try contacting the owner of that package first.
-                You can reach them using the "Contact Owners" link on the package details page.
+                If you're having trouble installing or using a specific package, try contacting the owner of that package first.
+                You can reach them using the "Contact Owners" link on the package details page. Please note that nuget.org support cannot offer any assistance with regards to individual package issues.
```
loic-sharma: Instead of:

> Please note that nuget.org support cannot offer any assistance with regards to individual package issues.`

What do you think think of something like:

> Please note that nuget.org does not offer assistance for individual package issues.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -497,8 +497,11 @@ public virtual async Task<ActionResult> LinkOrChangeExternalCredential(string re
             }
             else
             {
-                // The identity value contains cookie non-compliant characters like `<, >`(eg: John Doe <john@doe.com>), hence these need to be encoded
-                TempData["ErrorMessage"] = string.Format(Strings.ChangeCredential_Failed, HttpUtility.UrlEncode(newCredential.Identity));
+                // The identity value contains cookie non-compliant characters like `<, >`(eg: John Doe <john@doe.com>), and which would be treated
```
scottbommarito: nit:

```
// The identity value contains `<` and `>` characters (e.g. 'John Doe <john@doe.com>')
// We need to replace these characters so they are not treated as HTML tags
```

src/NuGetGallery/Helpers/UriExtensions.cs
```diff
@@ -81,5 +83,10 @@ public static string AppendQueryStringToRelativeUri(string relativeUrl, IReadOnl
             builder.Query = query.ToString();
             return builder.Uri.PathAndQuery;
         }
+
+        public static string GetExternalUrlElement(string data, string link)
```
scottbommarito: Maybe `GetExternalUrlAnchorTag` instead, so that it's immediately clear this is generating an HTML `<a>` tag?

Also, shouldn't this method return `HtmlString`?

src/NuGetGallery/Views/Pages/Contact.cshtml
```diff
@@ -14,15 +14,15 @@
     </div>
     <div class="row">
         <div class="@ViewHelpers.GetColumnClasses(ViewBag)">
-            <h2 class="ms-fontSize-xl ms-fontWeight-semibold">Having problems using a specific package?</h2>
+            <h2 class="ms-fontSize-xl ms-fontWeight-semibold">Having problems using a specific package on nuget.org?</h2>
             <p>
-                If you're having trouble installing a specific package, try contacting the owner of that package first.
-                You can reach them using the "Contact Owners" link on the package details page.
+                If you're having trouble installing or using a specific package, try contacting the owner of that package first.
```
scottbommarito: We should make this as clear as possible to prevent additional spam.

I would suggest
- making "we don't provide individual package support" the very first thing in the section
- removing "first" from "try contacting them **first**"

for example
```
nuget.org does not offer any assistance for individual package issues.
If you're having trouble installing or using a specific package, try contacting the owners of that package.
You can reach them using the "Contact Owners" link on the package details page.
```

src/NuGetGallery/Views/Pages/Contact.cshtml
```diff
@@ -31,25 +31,35 @@
...
                 If you feel that a package is violating the license of software you own or is violating our terms of service,
-                use the "Report Abuse" link on the package details page to report it directly to us.
+                use the "Report" link on the package details page to report it directly to us.
             </p>
 
-            <h2 class="ms-fontSize-xl ms-fontWeight-semibold">Want to change your NuGet.org username?</h2>
+            <h2 class="ms-fontSize-xl ms-fontWeight-semibold">Have trouble logging in to nuget.org with MSA/AAD logins?</h2>
             <p>
-                No need to contact support. Create a new user and transfer ownership of your packages to it.
+                Check out our <a href="https://docs.microsoft.com/en-us/nuget/faqs/nuget-faq#nugetorg-account-management" target="_blank">nuget.org account management</a> FAQ page for solutions.
```
scottbommarito: Have you considered using your new `UriExtensions` method for generating these anchor tags?

src/NuGetGallery/GalleryConstants.cs
```diff
@@ -129,5 +129,11 @@ public static class StatisticsDimensions
             public const string ClientVersion = "ClientVersion";
             public const string Operation = "Operation";
         }
+
+        public static class FAQLinks
+        {
+            public const string MSALinkedToAnotherAccount = "https://docs.microsoft.com/en-us/nuget/faqs/nuget-faq#microsoft-account-is-linked-with-another-nugetorg-account";
```
joelverhagen: Would it make sense to turn these into aka.ms links? This means we can fix them without redeploying and I think we do this elsewhere. Worth checking.

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -5,6 +5,7 @@
         && Request.IsAuthenticated
         && Config.Current.DeprecateNuGetPasswordLogins
         && User.HasPasswordLogin();
+    var rawErrorMessage = TempData.ContainsKey("RawErrorMessage") ? TempData["RawErrorMessage"].ToString() : null;
```
joelverhagen: Can TempData["RawErrorMessage"] somehow return null? `?.ToString()` would be a bit more defensive.

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
```diff
@@ -1282,7 +1282,7 @@ public async Task GivenExistingCredential_ItSafeRedirectsToReturnUrlWithErrorMes
 
                 // Assert
                 ResultAssert.IsSafeRedirectTo(result, "theReturnUrl");
-                Assert.Equal(string.Format(Strings.ChangeCredential_Failed, identity), controller.TempData["ErrorMessage"]);
+                Assert.NotNull(controller.TempData["RawErrorMessage"]);
```
joelverhagen: This test is weaker now. Don't we know the specific message here?

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
shishirx34: Since we use `Convert.ToString` to put values in cookie, it will put empty for null values. so it won't be null. No objection on being more thorough and add null operator though.

https://docs.microsoft.com/en-us/dotnet/api/system.convert.tostring?redirectedfrom=MSDN&view=netframework-4.7.2#System_Convert_ToString_System_Object_

> Returns
> String
> The string representation of value, or Empty if value is null.

shishirx34: nevermind.. razor is using c#5 version which doesn't support null propagating operator, might need to use ternary operator for completeness which isn't required as such.. so not going to do that.

src/NuGetGallery/Helpers/UriExtensions.cs
shishirx34: renamed. this should return string because its being used from the `AuthenticationController` as well. I will add a helper function to generate `HtmlString` for view.

tests/NuGetGallery.Facts/Controllers/AuthenticationControllerFacts.cs
```diff
@@ -1282,7 +1282,15 @@ public async Task GivenExistingCredential_ItSafeRedirectsToReturnUrlWithErrorMes
 
                 // Assert
                 ResultAssert.IsSafeRedirectTo(result, "theReturnUrl");
-                Assert.NotNull(controller.TempData["RawErrorMessage"]);
+                
+                var errorMessage = controller.TempData["RawErrorMessage"];
+                var expectedMessage = string.Format(
+                    Strings.ChangeCredential_Failed,
+                    identity.Replace("<", "&lgt;").Replace(">", "&gt;"),
```
joelverhagen: `&lt;`

shishirx34: Ah, I see why it didn't fail the test with this typo. Fixed it.

PR #6887, opened 2019-02-13
Autocomplete APIs for vulnerability information

xavierdecoster:


src/NuGetGallery/Queries/CweQueryStringValidator.cs
```diff
@@ -0,0 +1,53 @@
...
+            }
+            else
+            {
+                return queryString;
+            }
+        }
+
+        private static CweQueryMethod DetermineQueryMethod(string queryString, out string numericPartString)
+        {
+            if (CweIdHelper.StartsWithCweIdPrefix(queryString))
```
scottbommarito: Similarly, have you considered using regex here as well? I think it would be simpler because you can get rid of `CweIdHelper` in favor of a couple regex calls.

I believe it would be `(CWE-)?(?<query>\d+)` (depending on the max length of a CWE's ID). If it matches, it's by ID, and the query is the value of the `query` group. If it doesn't, it's by name. You can filter both queries by a max length.

src/NuGetGallery/Queries/IAutoCompleteCveIdsQuery.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace NuGetGallery
+{
+    public interface IAutoCompleteCveIdsQuery
```
scottbommarito: Have you considered combining `IAutoCompleteCveIdsQuery` and `IAutoCompleteCweIdsQuery`? They have very similar purposes, are used by a single controller and page, and have the same implementation dependencies. I think we should try to reduce the number of services and classes in the gallery when we can.

src/NuGetGallery/Queries/IAutoCompleteCweIdsQuery.cs
```diff
@@ -0,0 +1,24 @@
...
+using System;
+using System.Collections.Generic;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public interface IAutoCompleteCweIdsQuery
+    {
+        /// <summary>
+        /// Returns the CVE Id's and (truncated) descriptions matching
```
scottbommarito: `CVE` -> `CWE`

tests/NuGetGallery.Facts/Queries/AutocompleteCveIdsQueryFacts.cs
```diff
@@ -0,0 +1,68 @@
...
+                entitiesContext.Cves.Add(expectedResult5);
+                entitiesContext.Cves.Add(notExpectedResult2);
+
+                var query = new AutocompleteCveIdsQuery(entitiesContext);
+                var queryResults = query.Execute(queryString);
+
+                Assert.NotNull(queryResults);
+                Assert.Equal(5, queryResults.Count);
+
+                Assert.All(
```
scottbommarito: IMO I'd prefer to see simpler asserts here:

```
Assert.Contains(expectedResult1, queryResults);
Assert.Contains(expectedResult2, queryResults);
Assert.Contains(expectedResult3, queryResults);
Assert.Contains(expectedResult4, queryResults);
Assert.Contains(expectedResult5, queryResults);
Assert.NotContains(notExpectedResult1, queryResults);
Assert.NotContains(notExpectedResult2, queryResults);
```

tests/NuGetGallery.Facts/Queries/AutocompleteCweIdsQueryFacts.cs
```diff
@@ -0,0 +1,115 @@
...
+                entitiesContext.Cwes.Add(notExpectedResult1);
+                entitiesContext.Cwes.Add(notExpectedResult2);
+
+                var query = new AutocompleteCweIdsQuery(entitiesContext);
+                var queryResults = query.Execute("Name A");
+
+                Assert.NotNull(queryResults);
+                Assert.Single(queryResults);
+
+                Assert.Collection(
```
scottbommarito: nit: just grab the single element from the collection and perform the asserts on it directly instead of in a "foreach"

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -230,6 +228,72 @@ public async Task<JsonResult> RemovePackageOwner(string id, string username)
...
+            }
+
+            var model = new CveAutocompleteDataViewModel();
+
+            // Get CVE data.
+            // Suggestions will be CVE Id's that start with characters entered by the user.
+            IReadOnlyCollection<CveIdAutocompleteQueryResult> suggestions;
+            try
+            {
+                suggestions = GetService<IAutoCompleteCveIdsQuery>().Execute(query);
```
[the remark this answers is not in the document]
xavierdecoster: I was taking the exact same approach here as taken for the existing `IAutoCompletePackageIdsQuery` and `IAutoCompletePackageVersionsQuery`, which are also not injected via constructor.

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -230,6 +228,72 @@ public async Task<JsonResult> RemovePackageOwner(string id, string username)
...
+
+                model.Items.AddRange(suggestions);
+            }
+            catch (FormatException formatException)
+            {
+                return Json(new { success = false, message = formatException.Message }, JsonRequestBehavior.AllowGet);
```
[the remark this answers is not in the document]
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/7b0640335e227b43d771deb33c032c71211e8dc3

src/NuGetGallery/Queries/IAutoCompleteCweIdsQuery.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/8c87a2e94bfc22027e55fe2ec1d0a783779ca8a2

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -230,6 +228,72 @@ public async Task<JsonResult> RemovePackageOwner(string id, string username)
             }
         }
 
+        [HttpGet]
+        [ActionName("CveIds")]
+        public ActionResult GetCveIds(string query)
+        {
+            if (string.IsNullOrWhiteSpace(query))
```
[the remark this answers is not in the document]
xavierdecoster: Addressed with https://github.com/NuGet/NuGetGallery/pull/6887/commits/d5d8a3c26ecdd108b95906ebd31ab1c896672b72.

The `model` will be `null` though. Do you require it to be initialized?

src/NuGetGallery/Queries/AutocompleteCveIdsQuery.cs
```diff
@@ -0,0 +1,90 @@
...
+                .ToList();
+        }
+
+        private string ValidatePartialCveIdFormat(string partialId)
+        {
+            // The user input will be interpreted as a single CVE Id.
+            // The user may choose to not type the CVE Id prefix "CVE-".
+            // Accepted user input format is "CVE-{year}-xxxxxxx" or just "{year}-xxxxxxx".
+
+            if (!partialId.Contains('-'))
```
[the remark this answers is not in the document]
xavierdecoster: Yep, great catch! Added a test case to cover for this scenario.
Fixed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/bf242de4badf5e62e0c7dc237969ee9073c04cb1

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -230,6 +228,72 @@ public async Task<JsonResult> RemovePackageOwner(string id, string username)
...
+            {
+                return new HttpStatusCodeResult(HttpStatusCode.BadRequest, "Search term cannot be empty.");
+            }
+
+            // We should wait for at least 4 numeric characters before suggesting.
+            if (query.ToUpperInvariant().Replace(Cve.IdPrefix, string.Empty).Length < 4)
```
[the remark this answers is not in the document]
xavierdecoster: I forgot to remove this one as I moved validation into the queries.
Removed the duplicate check in https://github.com/NuGet/NuGetGallery/pull/6887/commits/4be06c24b8d379674f8b94e56a53f43e72b95be2

src/NuGetGallery/Queries/IAutoCompleteCveIdsQuery.cs
xavierdecoster: The return types of the `Execute` methods are different, and they have different validation logic (leveraging different static class methods). Would prefer to keep them separate for now.

src/NuGetGallery/Queries/AutocompleteCveIdsQuery.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                throw new FormatException(Strings.AutocompleteCveIds_FormatException);
+            }
+
+            // Strip off the leading "CVE-" if present.
+            var remainingSearchTerm = CveIdHelper.RemoveCveIdPrefix(partialId);
```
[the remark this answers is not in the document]
xavierdecoster: Good idea. Slightly changed the regex for CVE ID query validation, as there's no max-length limit on the second numeric part.

Addressed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/d4f08d2724a27531e66574dbfe7865cf95466f36.

src/NuGetGallery/Queries/CweQueryStringValidator.cs
xavierdecoster: Addressed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/d4f08d2724a27531e66574dbfe7865cf95466f36.

tests/NuGetGallery.Facts/Queries/AutocompleteCveIdsQueryFacts.cs
xavierdecoster: Can't, they're different types. e.g. `Cve` versus `AutocompleteCveQueryResult`

tests/NuGetGallery.Facts/Queries/AutocompleteCweIdsQueryFacts.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/111179748971cdb7ec7c26fd0dc533849577f25d

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -230,6 +228,74 @@ public async Task<JsonResult> RemovePackageOwner(string id, string username)
             }
         }
 
+        [HttpGet]
+        [ActionName("CveIds")]
+        public JsonResult GetCveIds(string query)
```
joelverhagen: Can we move this to its own controller, e.g. `VulnerabilitiesController`? This one is getting too big.

joelverhagen: Unit tests for these methods?

src/NuGetGallery/Queries/AutocompleteCveIdsQuery.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            // The user input will be interpreted as a single CVE Id.
+            // The user may choose to not type the CVE Id prefix "CVE-".
+            // Accepted user input format is "CVE-{year}-xxxxxxx" or just "{year}-xxxxxxx".
+
+            var match = ValidationRegex.Match(partialId);
+
+            if (match.Value == string.Empty)
+            {
+                throw new FormatException(Strings.AutocompleteCveIds_FormatException);
```
joelverhagen: This is non-exceptional. Could we make it so the `CveIdAutocompleteQueryResult` can represent an error state as well and do away with non-exceptional exceptions?

src/NuGetGallery/Queries/AutocompleteCveIdsQuery.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                throw new ArgumentNullException(nameof(partialId));
+            }
+
+            var validatedPartialId = ValidatePartialCveIdFormat(partialId);
+
+            // Query the database.
+            // Only include listed CVE entities.
+            var queryResults = _entitiesContext.Cves
+                .Where(e => e.CveId.StartsWith(validatedPartialId) && e.Listed == true)
```
joelverhagen: Why `== true`?

src/NuGetGallery/Controllers/JsonApiController.cs
```diff
@@ -230,6 +228,74 @@ public async Task<JsonResult> RemovePackageOwner(string id, string username)
...
+            {
+                suggestions = GetService<IAutoCompleteCveIdsQuery>().Execute(query);
+
+                model.Items.AddRange(suggestions);
+            }
+            catch (FormatException formatException)
+            {
+                return Json(
+                    HttpStatusCode.BadRequest,
+                    new { success = false, message = formatException.Message },
```
joelverhagen: So if the user types "foo" they will get an HTTP 400?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -702,6 +702,17 @@ private static void ConfigureAutocomplete(ContainerBuilder builder, IGalleryConf
                     .As<IAutoCompletePackageVersionsQuery>()
                     .InstancePerLifetimeScope();
             }
+
+            // Vulnerability Autocomplete
+            builder.RegisterType<AutocompleteCveIdsQuery>()
+                .AsSelf()
```
joelverhagen: Why register `AsSelf()`?

src/NuGetGallery/Queries/AutocompleteCweIdsQuery.cs
```diff
@@ -0,0 +1,64 @@
...
+            var validatedSearchTerm = CweQueryStringValidator.Validate(queryString, out var queryMethod);
+
+            // Query the database.
+            // Only include listed CVE entities.
+            IReadOnlyCollection<Cwe> queryResults;
+            switch (queryMethod)
+            {
+                case CweQueryMethod.ByCweId:
+                    queryResults = _entitiesContext.Cwes
+                        .Where(e => e.CweId.StartsWith(validatedSearchTerm) && e.Listed == true)
```
joelverhagen: Why `== true`?

src/NuGetGallery/Controllers/JsonApiController.cs
joelverhagen: What is the user experience when they type something that doesn't match the regex? Like from the user-in-a-web-browser perspective?

Autocomplete probably shouldn't show errors... for example in Intellisence you never see error messages. You may see no results. Perhaps we suggest the format to the user with placeholder text in the HTML?

src/NuGetGallery/Controllers/JsonApiController.cs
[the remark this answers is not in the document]
scottbommarito: IMO we should change that as well--it doesn't match how we usually handle dependencies.

src/NuGetGallery/Controllers/JsonApiController.cs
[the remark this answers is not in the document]
scottbommarito: To clarify--if you'd like this to be as easy to handle on the client-side as possible--the body for both successful and unsuccessful responses should contain a `success` property that is either `true` or `false` regardless of the actual error code of the response. Then, unsuccessful responses can have a `message` (or `error`) property that is what is displayed to the user, and successful responses can have an `items` property that is the list of CVEs/CWEs returned to the user.

For example:

```
public class CveAutocompleteViewModel
{
    public CveAutocompleteViewModel(string errorMessage)
    {
        Success = false;
        ErrorMessage = errorMessage;
    }

    public CveAutocompleteViewModel(IReadOnlyCollection<CveIdAutocompleteQueryResult> results)
    {
        Success = true;
        Results = results;
    }

    public bool Success { get; set; }
    public string ErrorMessage { get; set; }
    public IReadOnlyCollection<CveIdAutocompleteQueryResult> Results { get; set; }
}
```

As an added benefit, with this you also don't need to create any dynamic objects in the controller methods.

src/NuGetGallery/Controllers/JsonApiController.cs
scottbommarito: IMO it might be best if we
- renamed the existing controller to `ManageOwnersJsonApiController` (that's all that's in it right now)
- created a new controller, `ManageDeprecationJsonApiController`, and move all deprecation JSON API methods (the vulnerabilities and alternate package APIs) to there

src/NuGetGallery/Controllers/JsonApiController.cs
scottbommarito: I would argue that it is an error because users are expected to input valid CVEs/CWEs...just like how inserting an email address in the forms where we ask for a username is invalid.

scottbommarito: Also, the format is in the placeholder text.

src/NuGetGallery/Queries/IAutoCompleteCveIdsQuery.cs
scottbommarito: How about we make a single `IVulnerabilityAutocompleteService` that has two methods, `FindCveIds` and `FindCweIds`. That way we can have two separate methods but a single interface/implementation.

src/NuGetGallery/Controllers/JsonApiController.cs
joelverhagen: Is there anything on the client side preventing the submission of an incorrect format?

The email address analogy is not close. The user's explicitly hit a "submit" button before submitting the form. Autocomplete automatically submits, typically.

src/NuGetGallery/Controllers/JsonApiController.cs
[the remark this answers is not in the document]
xavierdecoster: Ok, I'll change those too.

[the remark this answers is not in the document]
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/6b0936622e41a94c06ed2f6d58dd9934ddcfc160

src/NuGetGallery/Controllers/JsonApiController.cs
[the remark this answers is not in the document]
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/82e32720042d4eb8d14b3a40943a9ed041a884b0

src/NuGetGallery/Queries/IAutoCompleteCveIdsQuery.cs
xavierdecoster: I can see the convenience of introducing an `IVulnerabilityAutocompleteService` even if only wrapping the calls to these query classes, but I'd like to stay consistent with the other `Autocomplete*Query` types we have.

Introduced the service in https://github.com/NuGet/NuGetGallery/pull/6887/commits/8a3cbd9a18a4606184e0b11221d3962b5658ca5a

src/NuGetGallery/Controllers/JsonApiController.cs
xavierdecoster: We could, but that means we'd have different routes...

Currently, all of these are conveniently found under the `/json` route, e.g. `/json/cweids?query=` and `/json/removepackageowner?id=&username=`

Would prefer to limit the scope of this PR. If routes/controllers need refactoring, perhaps we should log an issue for it and handle it in a separate PR.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
xavierdecoster: Not needed indeed. Followed the same registration as for `IAutocompletePackageIdsQuery` and `IAutocompletePackageVersionsQuery` (right above this code, for which this likely is also not needed...).

Addressed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/8a3cbd9a18a4606184e0b11221d3962b5658ca5a

src/NuGetGallery/Queries/AutocompleteCweIdsQuery.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/011d47dcdfcfd5d9b22874e9a5ab4a6226b0c9bd

src/NuGetGallery/Queries/AutocompleteCveIdsQuery.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/6887/commits/011d47dcdfcfd5d9b22874e9a5ab4a6226b0c9bd

src/NuGetGallery/Queries/AutocompleteCveIdsQuery.cs
xavierdecoster: Fixed as part of https://github.com/NuGet/NuGetGallery/pull/6887/commits/c51a9cb5a13214de819f6179b82b3199b234c3a3

src/NuGetGallery/Controllers/JsonApiController.cs
xavierdecoster: In terms of user experience, it's up to the client to handle/prevent those scenarios. This PR is server-side only, and returns `HTTP 400` status combined with `success: false` and `errorMessage: "reason of validation failure"`.

src/NuGetGallery/Controllers/JsonApiController.cs
xavierdecoster: Added as part of  https://github.com/NuGet/NuGetGallery/pull/6887/commits/e6860eb3b8a9d3c2a0f21646e46a1b8324fff338

scottbommarito: We have a lot more JSON routes than those in the `JsonApiController`. Note that it's also very easy to create a `json/{...}` route in `Routes.cs` that isn't in the `JsonApiController`. I think it's fine if you add the new endpoints here but I also think you should also consider creating a new controller for these endpoints but not renaming the old one as a good option if you are concerned with the scope of this PR. I think massive controllers is an anti-pattern in the gallery and we should really try to avoid it in the future.

xavierdecoster: Ok, I'll introduce these new autocomplete endpoints in a new `ManageDeprecationJsonApiController` as part of this work. 
Renaming the current `JsonApiController` and moving the alternate package APIs is out of scope and I prefer those changes in separate PRs.

What do you think about the following route for this controller?
`/json/deprecation/{action}`

xavierdecoster: Handled in https://github.com/NuGet/NuGetGallery/pull/6887/commits/18e0ba7f8f83fa4890bfdcc67953efd70d0cc36d

scottbommarito: I will update my PR to use your new controller for the alternate package API once you've merged this one.

PR #6894, opened 2019-02-16
Add a feature flag for the atom feed and enable prerelease filtering

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/2992

If there is unexpected performance issues, we can use this as an escape hatch. Also plumb through the `prerel` parameter and give it a name consistent with the package search page.

### DIFF
--- a/src/NuGetGallery/App_Data/Files/Content/flags.json
+++ b/src/NuGetGallery/App_Data/Files/Content/flags.json
@@ -1,11 +1,14 @@
 {
   "Features": {
-    "NuGetGallery.Typosquatting": "Enabled"
+    "NuGetGallery.Typosquatting": "Enabled",
+    "NuGetGallery.PackagesAtomFeed": "Enabled"
   },
-
   "Flights": {
     "NuGetGallery.TyposquattingFlight": {
-      "All": true
+      "All": true,
+      "SiteAdmins": false,
+      "Accounts": [],
+      "Domains": []
     }
   }
 }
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/App_Data/Files/Content/flags.json
```diff
@@ -1,11 +1,14 @@
...
   "Features": {
-    "NuGetGallery.Typosquatting": "Enabled"
+    "NuGetGallery.Typosquatting": "Enabled",
+    "NuGetGallery.PackagesAtomFeed": "Enabled"
   },
-
   "Flights": {
     "NuGetGallery.TyposquattingFlight": {
-      "All": true
+      "All": true,
```
joelverhagen: This will no-op round trip now.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -181,6 +182,12 @@ public class PackagesControllerFacts
 
             licenseExpressionSplitter = licenseExpressionSplitter ?? new Mock<ILicenseExpressionSplitter>();
 
+            if (featureFlagService == null)
+            {
+                featureFlagService = new Mock<IFeatureFlagService>();
+                featureFlagService.SetReturnsDefault<bool>(true);
```
loic-sharma: This is a cool trick 👍

Issue #6570, opened 2018-10-16
Improper cleanup after package sync validation failure


agr:
It seems, Gallery is not properly cleaning up package upload state after sync validation failure:

0. Need latest dev with `Gallery.RejectPackagesWithLicense` flag in (#6562);
1. Enable the flag;
2. Start Gallery;
3. Try to upload .nupkg with <license> in it, Gallery rejects it;
4. Stop Gallery, disable the flag, start it again;
5. Go to upload page;

At this point Gallery suggests to continue with package upload (that previously failed), which means that package that was previously considered invalid was still stored. I'm not sure this is expected behavior.

agr:
Found another manifestation of probably the same issue:
1. Find a package that will fail validation;
2. Try uploading it to the Gallery
3. Notice the error message appearing under the file input with red background;
4. Click the "Upload" link at the header.
5. Notice the same error, now appearing at the page top on the grey background (on server side it revalidated the package, although no upload happened).

agr:
This happens because we save the file to the upload container before we run validations on it:

https://github.com/NuGet/NuGetGallery/blob/9f578da47bbb8d6c83112512e3ec5c6a715629b6/src/NuGetGallery/Controllers/PackagesController.cs#L523-L532

validations happen in `GetVerifyPackageView` (similar code exists for symbol packages). Validation code reads the file from the upload container.

PR #6896, opened 2019-02-16
Cleaning up upload container on failed sync validation

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/6570

Moved package contents processing (opening package, validating metadata, getting the license data) into separate method `ValidateAndProcessPackageContents`. Added code that if that method returns error the package is deleted from the upload container, preventing the situation when Gallery tries to resume package upload for a package that was deemed invalid.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -549,7 +549,6 @@ namespace NuGetGallery
             try
             {
                 packageContentData = await ValidateAndProcessPackageContents(currentUser, isSymbolsPackageUpload);
-
             }
             catch (Exception)
             {
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -545,20 +545,70 @@ namespace NuGetGallery
             bool isSymbolsPackageUpload,
             bool hasExistingSymbolsPackageAvailable)
         {
+            var packageContentData = await ValidateAndProcessPackageContents(currentUser, isSymbolsPackageUpload);
+
+            if (packageContentData.ErrorResult != null)
+            {
+                await _uploadFileService.DeleteUploadFileAsync(currentUser.Key);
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -544,21 +544,71 @@ private async Task<JsonResult> UploadPackageInternal(PackageArchiveReader packag
             PackageRegistration existingPackageRegistration,
             bool isSymbolsPackageUpload,
             bool hasExistingSymbolsPackageAvailable)
+        {
+            var packageContentData = await ValidateAndProcessPackageContents(currentUser, isSymbolsPackageUpload);
+
```
loic-sharma: Should we delete the upload file if `ValidateAndProcessPackageContents` throws?

agr: There is a catch-all in the top level entry function:
https://github.com/NuGet/NuGetGallery/blob/189ac93b7b408e5e10039c19dcf584e36b0ec3e1/src/NuGetGallery/Controllers/PackagesController.cs#L364-L367
That shows generic "Failed to read the package file" message.

agr: Well, never mind, it does not delete the file...

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -545,7 +545,17 @@ private async Task<JsonResult> UploadPackageInternal(PackageArchiveReader packag
             bool isSymbolsPackageUpload,
             bool hasExistingSymbolsPackageAvailable)
         {
-            var packageContentData = await ValidateAndProcessPackageContents(currentUser, isSymbolsPackageUpload);
+            PackageContentData packageContentData = null;
+            try
+            {
+                packageContentData = await ValidateAndProcessPackageContents(currentUser, isSymbolsPackageUpload);
+
```
loic-sharma: Please remove line 552

PR #6902, opened 2019-02-19
Allow bulk inserts/deletes through EntitiesContext/EntityRepository

scottbommarito:
precursor to https://github.com/NuGet/NuGetGallery/issues/6773

When deprecating/undeprecating many versions, we may need to do a bulk insert/delete. Unfortunately the `IDbSet.Add` and `IDbSet.Remove` that we currently use result in individual SQL operations for each call, so we need to use `DbSet.AddRange` and `DbSet.RemoveRange` to properly do bulk inserts/deletes.

Unfortunately these methods aren't on `IDbSet`, so I had to move all of our uses of `IDbSet` to `DbSet`. See https://stackoverflow.com/questions/31539750/why-theres-no-addrange-removerange-method-in-idbset-interface-in-entity-6 for some more context.


skofman1:
>     public class ValidationEntityRepository<T> : IEntityRepository<T>

why do we need to make changes to ValidationEntityRepository?

---
Refers to: src/NuGetGallery/Areas/Admin/Services/ValidationEntityRepository.cs:11 in 6f87c4c. [](commit_id = 6f87c4cd0aec5f666eb3a859da6911054940264e, deletion_comment = False)

scottbommarito:
@skofman1 - I had to change `ValidationEntityRepository` because it inherits from `IEntityRepository` which I added the new bulk insert/remove methods to.

PR #6905, opened 2019-02-20
Restructure PackageService to better support endpoints requiring all versions of an ID

scottbommarito:
precursor to https://github.com/NuGet/NuGetGallery/issues/6773

Both `PackagesController.DisplayPackage` and `PackagesController.Manage` require all versions of a package.

Previously, this was done either implicitly or lazily, specifically:
- if a version isn't specified or is missing, `FindPackageByIdAndVersion` must fetch all versions to calculate the latest one
- if a version was specified and found, the remaining versions would be loaded lazily

This is bad because our perf for these endpoints are relying on a hidden behavior of one of our services, and when this hidden behavior doesn't go our way, lazy loading is **very very very very** slow, and adding the deprecation entities associated with this feature makes it *even* worse.

So, to remedy this
- add `FindPackagesById` to get all packages by ID (with a `withDeprecations` parameter for when deprecation entities are needed)
- move logic from `FindPackageByIdAndVersion` that fetches the latest version from a list of packages and move it into `FilterLatestPackage`
    - `FindPackageByIdAndVersion` now calls `FilterLatestPackage` internally
- make DisplayPackage and Manage call `FindPackagesById` and then `FilterLatestPackage` if an exact version isn't found
- move the code from `FindAbsoluteLatestPackageById` into `DisplayPackage`

### DIFF
--- a/src/NuGetGallery.Core/Services/CorePackageService.cs
+++ b/src/NuGetGallery.Core/Services/CorePackageService.cs
@@ -285,15 +285,26 @@ namespace NuGetGallery
             }
         }
 
-        protected IQueryable<Package> GetPackagesByIdQueryable(string id)
+        protected IQueryable<Package> GetPackagesByIdQueryable(string id, bool withDeprecations = false)
         {
-            return _packageRepository
+            var packages = _packageRepository
                 .GetAll()
                 .Include(p => p.LicenseReports)
                 .Include(p => p.PackageRegistration)
                 .Include(p => p.User)
                 .Include(p => p.SymbolPackages)
                 .Where(p => p.PackageRegistration.Id == id);
+
+            if (withDeprecations)
+            {
+                packages = packages
+                    .Include(p => p.Deprecations.Select(d => d.AlternatePackage.PackageRegistration))
+                    .Include(p => p.Deprecations.Select(d => d.AlternatePackageRegistration))
+                    .Include(p => p.Deprecations.Select(d => d.Cves))
+                    .Include(p => p.Deprecations.Select(d => d.Cwes));
+            }
+
+            return packages;
         }
 
         private static Package FindPackage(IQueryable<Package> packages, Func<IQueryable<Package>, IQueryable<Package>> predicate = null)
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -650,14 +650,26 @@ namespace NuGetGallery
                 return RedirectToActionPermanent("DisplayPackage", new { id = id, version = normalized });
             }
 
-            Package package;
-            if (version != null && version.Equals(GalleryConstants.AbsoluteLatestUrlString, StringComparison.InvariantCultureIgnoreCase))
+            Package package = null;
+            // Load all packages with the ID.
+            var packages = _packageService.FindPackagesById(id);
+            if (version != null)
             {
-                package = _packageService.FindAbsoluteLatestPackageById(id, SemVerLevelKey.SemVer2);
+                if (version.Equals(GalleryConstants.AbsoluteLatestUrlString, StringComparison.InvariantCultureIgnoreCase))
+                {
+                    // The user is looking for the absolute latest version and not an exact version.
+                    package = packages.FirstOrDefault(p => p.IsLatestSemVer2);
+                }
+                else
+                {
+                    package = packages.SingleOrDefault(p => p.NormalizedVersion == NuGetVersionFormatter.Normalize(version));
+                }
             }
-            else
+
+            if (package == null)
             {
-                package = _packageService.FindPackageByIdAndVersion(id, version, SemVerLevelKey.SemVer2);
+                // If we cannot find the exact version or no version was provided, fall back to the latest version.
+                package = _packageService.FilterLatestPackage(packages, SemVerLevelKey.SemVer2, allowPrerelease: true);
             }
 
             // Validating packages should be hidden to everyone but the owners and admins.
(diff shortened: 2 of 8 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -669,14 +681,13 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
...
             }
-            
+
             var model = new DisplayPackageViewModel(package, currentUser);
 
             model.ValidatingTooLong = _validationService.IsValidatingTooLong(package);
             model.PackageValidationIssues = _validationService.GetLatestPackageValidationIssues(package);
             model.SymbolsPackageValidationIssues = _validationService.GetLatestPackageValidationIssues(model.LatestSymbolsPackage);
             model.IsCertificatesUIEnabled = _contentObjectService.CertificatesConfiguration?.IsUIEnabledForUser(currentUser) ?? false;
-            model.IsAtomFeedEnabled = _featureFlagService.IsPackagesAtomFeedEnabled();
```
loic-sharma: I'm guessing this was a mistake, please revert

scottbommarito: LOL yeah my bad

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -650,14 +650,26 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
                 return RedirectToActionPermanent("DisplayPackage", new { id = id, version = normalized });
             }
 
-            Package package;
-            if (version != null && version.Equals(GalleryConstants.AbsoluteLatestUrlString, StringComparison.InvariantCultureIgnoreCase))
+            Package package = null;
+            // Load all packages with the ID.
+            var packages = _packageService.FindPackagesById(id);
+            if (version != null)
```
loic-sharma: Should we test the performance of this on a package with thousands of versions? This will load quite a bit of information for the package's other versions

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I will also add a test so this behavior is verified.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I have a package with 1k versions locally and this works great. We were loading all this information before, just not obviously to the `PackagesController`.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -634,21 +636,49 @@ public async Task GivenAnAbsoluteLatestVersionItQueriesTheCorrectVersion()
...
-                         {
-                             Id = "Foo",
-                             Owners = new List<User>()
-                         },
-                         Version = "2.0.0",
-                         NormalizedVersion = "2.0.0",
-                         IsLatest = true,
-                         Title = "A test package!"
-                     });
+                var id = "Foo";
```
joelverhagen: Please rename the test to describe what you're really testing, which is DB sequence trumps SemVer order.

src/NuGetGallery/Services/IPackageService.cs
```diff
@@ -17,6 +17,8 @@ namespace NuGetGallery
     /// </summary>
     public interface IPackageService : ICorePackageService
     {
+        IReadOnlyCollection<Package> FindPackagesById(string id, bool withDeprecations = false);
```
skofman1: please add comments for new methods

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -170,62 +175,63 @@ public override PackageRegistration FindPackageRegistrationById(string packageId
...
+            bool allowPrerelease = true)
+        {
+            return FilterLatestPackageHelper(
+                // Filter out prereleases in the list if prereleases are not allowed.
+                packages.Where(p => allowPrerelease || !p.IsPrerelease).ToList(),
+                semVerLevelKey,
+                allowPrerelease);
+        }
+
+        private Package FilterLatestPackageHelper(
```
skofman1: this can be static

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Test was added, and we now also test that the feed is not shown when the flag is not enabled.

src/NuGetGallery.Core/Services/CorePackageService.cs
```diff
@@ -285,15 +285,26 @@ public virtual async Task UpdatePackageSigningCertificateAsync(string packageId,
             }
         }
 
-        protected IQueryable<Package> GetPackagesByIdQueryable(string id)
+        protected IQueryable<Package> GetPackagesByIdQueryable(string id, bool withDeprecations = false)
```
skofman1: something cool you could do here is add an extension method 
IQueryable<Package> WithDeprecations(this IQueryable<Package> package)
{
return  packages
                    .Include(p => p.Deprecations.Select(d => d.AlternatePackage.PackageRegistration))
                    .Include(p => p.Deprecations.Select(d => d.AlternatePackageRegistration))
                    .Include(p => p.Deprecations.Select(d => d.Cves))
                    .Include(p => p.Deprecations.Select(d => d.Cwes));
}

scottbommarito: Unfortunately extensions must be in `public static class`es or else they won't work--I tried to make an extension for similar `Include` situations before as well and was unable to. Because I only intend to use this here, I don't think it's worth making an extension class for this.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -170,62 +175,63 @@ public override PackageRegistration FindPackageRegistrationById(string packageId
...
             return package;
         }
 
-        public virtual Package FindAbsoluteLatestPackageById(string id, int? semVerLevelKey)
+        public virtual Package FilterLatestPackage(
+            IReadOnlyCollection<Package> packages,
+            int? semVerLevelKey = SemVerLevelKey.SemVer2,
+            bool allowPrerelease = true)
+        {
+            return FilterLatestPackageHelper(
```
cristinamanum: There is not a null check for the packages on this or the FilterLatestPackageHelper below.

scottbommarito: fixed and added tests

PR #6215, opened 2018-07-26
Fix (most) accessibility issues found by recent scan

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6152

This PR fixes the following issues:

[635011](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/635011) - "Skip to Content" link loses tab order
My solution here was to write a script that sets the focus on first element in the content that can be tabbed to.

[636482](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/636482) - Discontinued login modal dialog does not set focus correctly

[636486](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/636486) - Register new certificate button does not respond to keyboard events

[638489](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/638489) - Recaptcha badge in contact forms tab order is incorrect
*[(note: this currently does not fix the issue due to a bug in Recaptcha)](https://github.com/google/recaptcha/issues/160)*

[638490](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/638490) - Browse for readme button does not respond to keyboard events

[643610](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/643610) / [644048](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/644038) - Links that are styled like buttons are displaying like links in high contrast mode

[643613](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/643613) - Profile page "delete user" link tab order is incorrect
My solution here was to change the "delete user" text into an icon alongside the "manage account" breadcrumb.
image

[643615](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/643615) - Email notifications form label name is incorrect

[643617](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/643617) - Profile page "manage my organizations" breadcrumb should read the number of organizations you own

[644030](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/644030) - Owner dropdown in upload form should read correctly in screenreader

[644070](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/644070) - Number of packages returned in search should be announced by screenreader

[Functional](https://github.com/NuGet/NuGetGallery/issues/6200) - "Cancel" buttons do not work on the Account Settings page.

PR #6906, opened 2019-02-20
Improve implementation of "Skip To Content" link

scottbommarito:
When exploring custom keyboard controls for the Deprecation form, I discovered this as a substantially simpler implementation than what I had implemented in https://github.com/NuGet/NuGetGallery/pull/6215.

Basically, instead of having to traverse the DOM manually, I am now leveraging JQuery to do this heavy-lifting for me by defining a "this element can be tabbed to" selector. I also added `focusable` in case we need it.

This also fixes a bug in the previous implementation. If an element has a `tabindex` of `-1`, its children can still be tabbed to. In the previous implementation, children of elements with `tabindex="-1"` were not considered `tabbable`.

I also got rid of all the ESLint issues in the file.

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/common.js
+++ b/src/NuGetGallery/Scripts/gallery/common.js
@@ -216,21 +216,9 @@
         });
     };
 
-    nuget.isElement = function (element, type) {
-        return element.is(type);
-    };
-
-    nuget.hasAttribute = function (element, attributeName) {
-        var attribute = element.attr(attributeName);
-        return typeof attribute !== typeof undefined && attribute !== false;
-    };
-
     nuget.canElementBeFocused = function (element) {
         element = $(element);
-        var isElement = window.nuget.isElement.bind(this, element);
-        var hasAttribute = window.nuget.hasAttribute.bind(this, element);
-
-        if (!isElement(':visible')) {
+        if (!element.is(':visible')) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -208,27 +208,30 @@
...
-        }
+    nuget.hasAttribute = function (element, attributeName) {
+        var attribute = element.attr(attributeName);
+        return typeof attribute !== typeof undefined && attribute !== false;
+    };
+
+    nuget.canElementBeFocused = function (element) {
+        element = $(element);
+        var isElement = window.nuget.isElement.bind(this, element);
+        var hasAttribute = window.nuget.hasAttribute.bind(this, element);
```
joelverhagen: unused?

src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -208,27 +208,30 @@
...
+        return element.is(type);
+    };
 
-        var hasAttribute = function (attributeName) {
-            var attribute = element.attr(attributeName);
-            return typeof attribute !== typeof undefined && attribute !== false;
-        }
+    nuget.hasAttribute = function (element, attributeName) {
+        var attribute = element.attr(attributeName);
+        return typeof attribute !== typeof undefined && attribute !== false;
```
loic-sharma: Double-check this, but I think you can simplify line 225 to `!!attribute`

src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -208,27 +208,30 @@
...
+    };
+
+    nuget.canElementBeFocused = function (element) {
+        element = $(element);
+        var isElement = window.nuget.isElement.bind(this, element);
```
loic-sharma: `isElement(':visible')` could be just `element.is(':visible')`, right? If so, `isElement` seems a little overkill. `hasAttribute` is good though 😄

src/NuGetGallery/Scripts/gallery/common.js
scottbommarito: so much cleaner! thanks

src/NuGetGallery/Scripts/gallery/common.js
scottbommarito: removed both after you told me the `!!` idiom.

src/NuGetGallery/Scripts/gallery/common.js
scottbommarito: removed (it was used, but i refactored it and got rid of the method)

PR #6912, opened 2019-02-21
Add Deprecation section of the Manage Package page

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6773
https://github.com/NuGet/NuGetGallery/issues/6848
https://github.com/NuGet/NuGetGallery/issues/6849

Future PRs:
- CVE/CWE autocomplete
- tooltips (the (i) icons in the spec)

Tons of screenshots below:

image

image

image

image

image

image

image

image

image

image

image

image


src/Bootstrap/less/theme/base.less
```diff
@@ -273,6 +273,13 @@ h2, h3 {
 }
 
 @media screen and (max-width: @screen-sm) {
+  // Bootstrap doesn't set the container width when the screen is xs.
+  // We want the container width to be the width of the viewport on xs screens.
+  // Otherwise, elements that depend on the width of the container could behave oddly on xs screens.
+  .container {
```
scottbommarito: For more context--I had to add this to fix a UI bug with the multi-select dropdown on XS screens. The dropdown's button text is the list of every selected version, and when this list is greater than can display in the button, it relied on the container's width to properly overflow. At SM widths and above, the container width is set by Bootstrap, but at XS and smaller, it is set to "auto", so its width becomes equal to the longest item inside it, which caused it to take the width of the un-overflowed dropdown button. By setting the XS container width to the viewport's width, the dropdown button would then properly restrict itself to the size of the viewport.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1449,26 +1567,20 @@ public virtual ActionResult DeleteSymbols(string id, string version)
...
+            var versionsWithSymbols = package
                 .PackageRegistration
                 .Packages
                 .Where(p => p.PackageStatusKey != PackageStatus.Deleted)
-                .Select(p => p.LatestSymbolPackage())
-                .Where(sp => sp != null && sp.StatusKey == PackageStatus.Available)
-                .Select(sp => new PackageViewModel(sp.Package));
+                .Where(p => (p.LatestSymbolPackage()?.StatusKey ?? PackageStatus.Deleted) == PackageStatus.Available);
 
-            model.VersionSelectList = new SelectList(
```
scottbommarito: For more context, I made this change because I discovered that our use of `SelectList` was fairly unclean.

`SelectList` provides nothing other than a way to construct a `IEnumerable<SelectListItem>` using reflection. We have no reason to use it because there is no reason we can easily just rewrite our code to create the `IEnumerable<SelectListItem>` directly. I think this approach is much cleaner because it avoids the reflection and dynamic objects we currently use.

I discovered this while updating the `SelectList` in `ManagePackageViewModel` and updated it here as well because it was the only other place we used it.

src/NuGetGallery/Views/Packages/_ManageDocumentation.cshtml
```diff
@@ -50,18 +50,14 @@
...
-                        @Html.DropDownList("version-selection", Model.VersionSelectList,
-                            new
-                            {
-                                @class = "form-control selectpicker show-tick show-menu-arrow btn-block",
-                                @title = "Select a version",
-                                @id = "input-select-readme"
-                            })
-                    </label>
-                </div>
+                <label for="input-select-readme" id="readme-version-label">Select version</label>
```
scottbommarito: This change was a minor UI fix I noticed while working on the page. The select list in the documentation section is bolded because it is inside the label. None of our other select lists are bold, so I made this change to align it with the others.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -48,5 +55,29 @@ public JsonResult GetCweIds(string query)
                 queryResult,
                 JsonRequestBehavior.AllowGet);
         }
+
+        [HttpGet]
+        [UIAuthorize]
```
xavierdecoster: Seems I forgot the `UIAuthorize` attribute on the vulnerability APIs on this controller. Perhaps we should just add the attribute on the class?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1430,6 +1433,121 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+                {
+                    packagesToUpdate.Add(package);
+                }
+            }
+
+            cveIds = cveIds ?? Enumerable.Empty<string>();
+            var cves = _deprecationService.GetCvesById(cveIds);
+            if (cveIds.Count() != cves.Count)
+            {
+                return DeprecateErrorResponse(HttpStatusCode.NotFound, Strings.DeprecatePackage_MissingCve);
```
xavierdecoster: Per offline discussion, we need to cover for the scenario where vulnerability data is not up-to-date.

> A CVE/CWE info may not be available in the autocomplete API due to the following reasons:
> 1.	Data staleness on NuGet Gallery – We refresh every 24 hours to begin with.
> 2.	Data staleness on the CVE/CWE repository – This could happen too as sometimes the ID is provided upon report but the corresponding data record is absent till the vulnerability is being explored/validated.

In these cases, we should visually warn but accept the input from the package publisher

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1430,6 +1433,121 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+            }
+
+            cweIds = cweIds ?? Enumerable.Empty<string>();
+            var cwes = _deprecationService.GetCwesById(cweIds);
+            if (cweIds.Count() != cwes.Count)
+            {
+                return DeprecateErrorResponse(HttpStatusCode.NotFound, Strings.DeprecatePackage_MissingCwe);
```
xavierdecoster: Same here.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: I originally didn't have the attribute on my endpoint, but @skofman1 asked me to as an extra precaution. I don't feel very strongly about it, because we're not exposing any private information here and the only protection on the endpoint is that the user has an account. I'll add it to your endpoints if you'd like.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: I was originally going to postpone doing this until a future PR, but I think it probably wouldn't be too much work.

scottbommarito: After looking into it more, I will postpone this to the future CVE/CWE autocomplete integration PR to reduce the size of this one.

src/Bootstrap/less/theme/page-manage-deprecation.less
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    .deprecation-section {
+        margin-top: 40px;
+
+        .deprecation-section-header {
+            margin-bottom: 15px;
+        }
+
+        .dropdown {
```
joelverhagen: This CSS should be in a shared .less file so it can more easily be used in another place.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -48,5 +55,29 @@ public JsonResult GetCweIds(string query)
...
+            if (registration == null)
+            {
+                return Json(HttpStatusCode.NotFound, null, JsonRequestBehavior.AllowGet);
+            }
+
+            var versions = registration.Packages
+                .Where(p => p.PackageStatusKey == PackageStatus.Available)
+                .ToList()
+                .OrderByDescending(p => NuGetVersion.Parse(p.Version))
+                .Select(p => NuGetVersionFormatter.ToFullString(p.Version));
```
joelverhagen: This is parsing twice, right? You should be able to both sort and get a full version string from `NuGetVersion`.

Also, why full version? Does this look reasonable with, say, the build metadata that client team puts on their packages?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1430,6 +1433,127 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+            bool isLegacy,
+            bool isOther,
+            IEnumerable<string> cveIds,
+            decimal? cvssRating,
+            IEnumerable<string> cweIds,
+            string alternatePackageId,
+            string alternatePackageVersion,
+            string customMessage,
+            bool shouldUnlist)
+        {
```
joelverhagen: If the package registration is locked, people shouldn't be able to deprecate right? My understand is that locking prevents everything.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1430,6 +1433,127 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+                            HttpStatusCode.NotFound,
+                            string.Format(Strings.DeprecatePackage_NoAlternatePackageRegistration, alternatePackageId));
+                    }
+                }
+            }
+
+            var packagesToUpdate = new List<Package>();
+            foreach (var version in versions)
+            {
+                var package = packages.SingleOrDefault(v => v.NormalizedVersion == NuGetVersionFormatter.Normalize(version));
```
joelverhagen: You don't need to `Normalize` `version` over and over.

src/NuGetGallery/Helpers/PackageHelper.cs
```diff
@@ -170,5 +172,41 @@ public static void ValidateNuGetPackageMetadata(PackageMetadata packageMetadata)
                 throw new EntityException(Strings.NuGetPackagePropertyTooLong, "RepositoryUrl", "4000");
             }
         }
+
+        public static string GetSelectListText(Package package)
```
joelverhagen: Are there unit tests just for this?

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
```diff
@@ -0,0 +1,450 @@
...
+            visible: visible,
+            selected: ko.pureComputed(function () {
+                // If a version is checked but not visible in the UI, it is not selected.
+                return checked() && visible();
+            })
+        };
+    });
+
+    // The versions selected in the UI.
+    this.chosenVersions = ko.pureComputed(function () {
```
joelverhagen: Can we move the dropdown JS to a shareable JS file and make the variable names not version specific?

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
```diff
@@ -0,0 +1,450 @@
...
+    // Toggles whether or not all versions are selected.
+    // If the checkbox is not selected, it selects all versions visible in the UI.
+    // If the checkbox is already selected, it deselects all versions visible in the UI.
+    this.toggleVersionSelectAll = function () {
+        var checked = !self.versionSelectAllChecked();
+        ko.utils.arrayForEach(
+            self.versions,
+            function (version) {
+                if (version.visible()) {
+                    version.checked(checked);
```
joelverhagen: This is N individual observable changes which I believe each results in its own redraw. Please investigate batching these observable changes. Maybe start here?
https://knockoutjs.com/documentation/deferred-updates.html

joelverhagen: In other words, how does this perform on 1000+ versions?

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
```diff
@@ -125,5 +122,50 @@ public VersionReadMeState(string submitUrl, string getReadMeUrl, string readMe)
...
+
+                    CustomMessage = deprecation.CustomMessage;
+
+                    // It doesn't make sense to unlist packages that are already unlisted.
+                    // Additionally, if a package was not unlisted when it was deprecated, we shouldn't unlist it when its deprecation information is updated.
+                    ShouldUnlist = package.Listed && deprecation.Status == PackageDeprecationStatus.NotDeprecated;
+                }
+            }
+
+            public string Text { get; set; }
```
joelverhagen: Do all of these need `set;`? Seems like a lot of this can be read-only.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: Rewrote it to only parse once.

We use full version in all of our other version select lists, so I chose it here to be consistent. I don't mind making it normalized, but I feel like more information is better. I don't think it would look reasonable with how the client uses build metadata, but I would prefer to see how other users use build metadata as well before changing this.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: good call.

src/NuGetGallery/Helpers/PackageHelper.cs
scottbommarito: Will add.

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
scottbommarito: I'm aware of this and have been trying to fix it, but I'm not sure how or if it's worth investing into. On 1000 versions it takes probably around .5 seconds to finish, which is slow, but still very usable. I tried deferred updates and unfortunately the performance did not improve--I think it still performs the update because the updated observable changes the UI.

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
scottbommarito: I was planning on it, but it's nontrivial to do and I was still making some tweaks to the dropdown UI/behavior.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1430,6 +1433,127 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
             return View(model);
         }
 
+        [HttpPost]
+        [RequiresAccountConfirmation("deprecate a package")]
+        [ValidateAntiForgeryToken]
+        public virtual async Task<JsonResult> Deprecate(
+            string id,
```
loic-sharma: Could these new routes be moved to `ManageDeprecationJsonApiController` as these all return JSON?

src/Bootstrap/less/theme/page-manage-deprecation.less
scottbommarito: Will do this when I generalize it.

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
scottbommarito: done

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
joelverhagen: Alright, sounds good. If we've given it an honest effort and the duration is still pretty low, then that's good enough for now.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -48,5 +55,30 @@ public JsonResult GetCweIds(string query)
...
+            var versions = registration.Packages
+                .Where(p => p.PackageStatusKey == PackageStatus.Available)
+                .ToList()
+                .Select(p => NuGetVersion.Parse(p.Version))
+                .OrderByDescending(v => v)
+                .Select(v => v.ToFullString());
+
+            if (!versions.Any())
+            {
+                return Json(HttpStatusCode.NotFound, null, JsonRequestBehavior.AllowGet);
```
skofman1: >(HttpStatusCode.NotFound, [](start = 27, length = 25)

why did you chose to return NotFound instead of an empty list? Won't 404 be interpreted as "endpoint not found"?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1430,6 +1433,135 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
             return View(model);
         }
 
+        [HttpPost]
+        [RequiresAccountConfirmation("deprecate a package")]
+        [ValidateAntiForgeryToken]
+        public virtual async Task<JsonResult> Deprecate(
```
skofman1: [UIAuthorize] is missing

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
xavierdecoster: Yes, please do :) Thanks!

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1430,6 +1433,135 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+            string id,
+            IEnumerable<string> versions,
+            bool isVulnerable,
```
skofman1: I think we are missing some input checks here:
1. isVulnerable, isVulnerable, isOther - one of them must be "true"
2. customMessage - are we allowing any characters or only say "UTF-8"
3. if cvssRating provided, is it in the expected range of 0-10

Maybe more checks. All of those should be performed before we do a single DB call.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1430,6 +1433,135 @@ public virtual async Task<ActionResult> Manage(string id, string version = null)
...
+            string alternatePackageVersion,
+            string customMessage,
+            bool shouldUnlist)
+        {
+            if (versions == null || !versions.Any())
+            {
+                return DeprecateErrorResponse(HttpStatusCode.BadRequest, Strings.DeprecatePackage_NoVersions);
+            }
+
+            var registration = _packageService.FindPackageRegistrationById(id);
```
skofman1: check that the registration exists

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: 404 means "content not found", not endpoint not found.

I'm returning an error status instead of an empty list because users should not be able to chose an alternate package that doesn't exist.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: 1. All flags missing is interpreted as "undeprecate" and removes any existing deprecations.

2. We currently don't do any character filtering on user input (for example, package IDs can have Chinese characters). I don't think it makes sense to restrict the character set here when we don't restrict more important fields. If a package has an ID in a language not supported by UTF-8, I don't think it makes sense to prevent them from writing a custom message in a language not supported by UTF-8.

3. I can double check, but there's a couple levels of data restrictions (data attributes on the entities, EF restrictions) that should be preventing this. If the call does get to SQL, it will get rejected by restrictions on the column.

skofman1: Agree on (1), (2). Regarding (3) and in general, you should do as many sanity checks as possible before doing multiple DB calls. It makes failure cheaper.

scottbommarito: Added an explicit range check. Unfortunately there's no easy way to check the `decimal`'s precision AFAIK so I decided to not to write a check for it.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: added a check

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
scottbommarito: done

src/Bootstrap/less/theme/page-manage-deprecation.less
scottbommarito: done

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Will do in the PR that integrates your CVE/CWE autocomplete work with this form.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -110,6 +110,7 @@ public partial class PackagesController
         private readonly ICoreLicenseFileService _coreLicenseFileService;
         private readonly ILicenseExpressionSplitter _licenseExpressionSplitter;
         private readonly IFeatureFlagService _featureFlagService;
+        private readonly IPackageDeprecationService _deprecationService;
```
loic-sharma: >DeprecationService [](start = 33, length = 18)

Is this unused now?

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -0,0 +1,255 @@
...
+        private IEntityRepository<Cve> _cveRepository;
+        private IEntityRepository<Cwe> _cweRepository;
+
+        public PackageDeprecationService(
+           IEntityRepository<PackageDeprecation> deprecationRepository,
+           IEntityRepository<Cve> cveRepository,
+           IEntityRepository<Cwe> cweRepository)
+        {
+            _deprecationRepository = deprecationRepository;
+            _cveRepository = cveRepository;
```
loic-sharma: Don't forget null checks here

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -0,0 +1,255 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public class PackageDeprecationService : IPackageDeprecationService
+    {
+        private IEntityRepository<PackageDeprecation> _deprecationRepository;
```
loic-sharma: >private [](start = 8, length = 7)

`private readonly`

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Yup! removed.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: added

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: added!

src/Bootstrap/less/theme/base.less
```diff
@@ -273,6 +273,20 @@ h2, h3 {
...
+    // This works fine when there are no elements in the container that are wider than the viewport,
+    // however, if there are elements in the container that are wider than the viewport, 
+    // their overflow properties will not work properly because they are not bound by the size of the container.
+    // To fix this, we need to set the container width to be the width of the viewport minus the row margin on xs screens.
+    // This way, those elements that are larger than the viewport are properly bound by the size of the container and overflow properly.
+    width: calc(100vw - 15px);
+  }
+
+  .navbar-toggle {
+    margin-right: 0;
```
scottbommarito: This fixes the alignment of the `navbar-toggle` after the above change.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
skofman1: From wikipedia:
404 Not Found
The requested resource could not be found but may be available in the future. Subsequent requests by the client are permissible.


I don't think it's correct to return a 404 here. Return 200 or if you want to be extra correct 

204 No Content
The server successfully processed the request and is not returning any content.


---
In reply to: [259066156](https://github.com/NuGet/NuGetGallery/pull/6912#discussion_r259066156) [](ancestors = 259066156)

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -48,5 +62,173 @@ public JsonResult GetCweIds(string query)
...
+        public virtual JsonResult GetAlternatePackageVersions(string id)
+        {
+            var registration = _packageService.FindPackageRegistrationById(id);
+            if (registration == null)
+            {
+                return Json(HttpStatusCode.NotFound, null, JsonRequestBehavior.AllowGet);
+            }
+
+            var versions = registration.Packages
+                .Where(p => p.PackageStatusKey == PackageStatus.Available)
```
skofman1: are we allowing unlisted versions?

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -48,5 +62,173 @@ public JsonResult GetCweIds(string query)
...
+            var packagesToUpdate = new List<Package>();
+            foreach (var version in versions)
+            {
+                var normalizedVersion = NuGetVersionFormatter.Normalize(version);
+                var package = packages.SingleOrDefault(v => v.NormalizedVersion == normalizedVersion);
+                if (package == null)
+                {
+                    // This should only happen if someone hacks the form or if a version of the package is deleted while the user is filling out the form.
+                    return DeprecateErrorResponse(
+                        HttpStatusCode.NotFound,
```
skofman1: This should be BadRequest and not NotFound.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: I don't think we should restrict this unless we have a strong reason not to allow it. I think it's possible that a package author may want to choose an unlisted package as an alternate, and if we don't show these versions in the autocomplete, it would be impossible for them to do so. Perhaps an author has an alternate package they want to use, but this package exists only for the purpose of patching the old package. The author doesn't want this alternate package to be discoverable except for through the existing deprecated package, so they unlist it.

We may want to consider hiding unlisted packages from users that do not own them in this autocomplete, but I think this is fine as-is, because unlisted packages are not meant to be private, they are only meant to be more difficult to discover.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: In the past, I have always used `NotFound` to signify "you asked for something, but it doesn't exist", regardless of the context. I think this is very consistent with the definition of the status code, which is "The requested resource could not be found but may be available in the future." I wasn't aware that we returned `BadRequest` in this situation anywhere until this comment.

I don't think the definition of `BadRequest` fits this situation very well. Its definition is "The server cannot or will not process the request due to an apparent client error." I don't think this is very suitable--the client made a completely coherent request that the server cannot fulfill because the resource (package/user/etc) no longer exists. I don't think the client made an error, especially given that the server told it the list of versions and the client chose from that list. In my opinion, `BadRequest` makes a lot more sense when the user entered something invalid into the form, e.g. they entered a username when an email address was asked for.

As an additional thing to note, I see in several listings of these status code definitions that `NotFound` is retriable, but `BadRequest` is not. I think this supports my opinion as well, because if the missing version was reuploaded the request will succeed, but if the user inputted a username instead of an email address, the request will never succeed.

Given that, I think `NotFound` makes more sense, but ultimately it's very arbitrary given that this is an AJAX call. We define the protocol and we define how the client handles the response.

skofman1: Since this is indeed the backend for AJAX calls we can table this discussion for now :)

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
skofman1: please check with @anangaur

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1449,26 +1451,20 @@ public virtual ActionResult DeleteSymbols(string id, string version)
...
-            // database since the DisplayPackageViewModel(base class for DeletePackageViewModel) does not
-            // set the `LatestSymbolsPackage` data on the model(since it is an unnecessary and expensive db
-            // query). It is fine to do this here when invoking delete page. Note: this could also potentially 
-            // cause unbounded(high number) db calls based on the number of versions associated with a package.
-            var packageViewModelsForAllAvailableSymbolsPackage = package
+            // Fetch all versions of the package with symbols.
+            var versionsWithSymbols = package
                 .PackageRegistration
                 .Packages
                 .Where(p => p.PackageStatusKey != PackageStatus.Deleted)
```
joelverhagen: What part of this IEnumerable is meant to be done in SQL Server and what should be done client side?

joelverhagen:
> function MultiSelectDropdownItem(value, text, name, checked, isRed) {

use strict

---
Refers to: src/NuGetGallery/Scripts/gallery/common-multi-select-dropdown.js:1 in f162e01. [](commit_id = f162e01adffce3cd3fef7fd02c7c81c1c56d82bf, deletion_comment = False)

src/NuGetGallery/App_Data/Files/Content/flags.json
```diff
@@ -9,6 +9,12 @@
       "SiteAdmins": false,
       "Accounts": [],
       "Domains": []
+    },
+    "NuGetGallery.ManageDeprecation": {
+      "All": false,
+      "SiteAdmins": true,
```
joelverhagen: >SiteAdmins [](start = 7, length = 10)

This is fine, since this is local only, but let's remember not to turn this on for anyone (even admins) until we have done perf testing.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: This comment made me realize that this `IEnumerable` makes O(n) lazy-loading calls, which is very expensive. Will be switching this entire method to `FindPackagesById` instead of `FindPackageByIdAndVersion`.

src/NuGetGallery/App_Start/AppActivator.cs
```diff
@@ -156,6 +156,10 @@ private static void BundlingPostStart()
                 .Include("~/Scripts/gallery/stats-dimensions.js");
             BundleTable.Bundles.Add(d3ScriptBundle);
 
+            var multiSelectDropdownBundle = new ScriptBundle("~/Scripts/gallery/common-multi-select-dropdown.min.js")
+                .Include("~/Scripts/gallery/common-multi-select-dropdown.js");
+            BundleTable.Bundles.Add(multiSelectDropdownBundle);
```
shishirx34: were you able to get the OSS approval for this package?

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
shishirx34: I agree with @skofman1 shouldn't be returning 404, return empty list with 200. and view side show that no versions found.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -48,5 +62,173 @@ public JsonResult GetCweIds(string query)
...
+                return DeprecateErrorResponse(
+                    HttpStatusCode.Forbidden,
+                    string.Format(Strings.DeprecatePackage_Locked, id));
+            }
+
+            PackageRegistration alternatePackageRegistration = null;
+            Package alternatePackage = null;
+            if (!string.IsNullOrEmpty(alternatePackageId))
+            {
+                if (!string.IsNullOrEmpty(alternatePackageVersion))
```
shishirx34: I would do `!string.IsNullOrWhitespace` more restrictive and checks all failure cases for strings. Not sure if `FindpackageByIdAndVersionStrict` throws for whitespace versions or Ids.

shishirx34: here and above.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -48,5 +62,173 @@ public JsonResult GetCweIds(string query)
...
+                {
+                    packagesToUpdate.Add(package);
+                }
+            }
+
+            cveIds = cveIds ?? Enumerable.Empty<string>();
+            var cves = _deprecationService.GetCvesById(cveIds);
+            if (cveIds.Count() != cves.Count)
+            {
+                return DeprecateErrorResponse(HttpStatusCode.NotFound, Strings.DeprecatePackage_MissingCve);
```
shishirx34: Is it worth augmenting error message with the missing cves?

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -48,5 +62,173 @@ public JsonResult GetCweIds(string query)
...
+            }
+
+            cweIds = cweIds ?? Enumerable.Empty<string>();
+            var cwes = _deprecationService.GetCwesById(cweIds);
+            if (cweIds.Count() != cwes.Count)
+            {
+                return DeprecateErrorResponse(HttpStatusCode.NotFound, Strings.DeprecatePackage_MissingCwe);
```
shishirx34: same as above.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -48,5 +62,173 @@ public JsonResult GetCweIds(string query)
...
+            }
+
+            var status = PackageDeprecationStatus.NotDeprecated;
+            if (isVulnerable)
+            {
+                status = status |= PackageDeprecationStatus.Vulnerable;
```
shishirx34: doesn't simply doing `|=` do `OR` and assignment, why the double assignment to status?

shishirx34: here and below.
`status |= PackageDeprecationStatus....`

should be enough.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1449,26 +1451,20 @@ public virtual ActionResult DeleteSymbols(string id, string version)
...
+            model.VersionSelectList = versionsWithSymbols
+                .Select(versionWithSymbols => new SelectListItem
                 {
-                    text = pvm.NuGetVersion.ToFullString() + (pvm.LatestVersionSemVer2 ? " (Latest)" : string.Empty),
-                    url = Url.DeleteSymbolsPackage(pvm)
-                }), "url", "text", Url.DeleteSymbolsPackage(model));
+                    Text = PackageHelper.GetSelectListText(versionWithSymbols),
+                    Value = Url.DeleteSymbolsPackage(new TrivialPackageVersionModel(versionWithSymbols)),
+                    Selected = package == versionWithSymbols
+                });
```
shishirx34: Please ensure the delete symbols work flow runs fine after this change.

src/NuGetGallery/App_Start/AppActivator.cs
scottbommarito: That isn't an OSS package--I wrote that myself with guidance from the UI team.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: We decided that missing CVEs/CWEs will be added to the DB with an "unverified" state, but that's out of the scope of this PR.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: Intended to do just the `|=` must have mistyped it and never noticed it due to the lack of functional difference haha.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Yup!

src/NuGetGallery/Helpers/PackageHelper.cs
scottbommarito: added

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Fixed. This method now loads all packages by ID and then filters it like how `Manage` and `DisplayPackage` do. The `SymbolPackages` for the packages are `.Include`d in the result so there is no lazy-loading.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: switched to `IsNullOrWhitespace`

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: Made it return 200.

src/Bootstrap/dist/css/bootstrap-theme.css
```diff
@@ -545,6 +552,58 @@ img.reserved-indicator-icon {
...
+  border-style: inset;
+
+  -webkit-box-pack: justify;
+  -webkit-justify-content: space-between;
+  -ms-flex-pack: justify;
+          justify-content: space-between;
+  -webkit-box-align: center;
+  -webkit-align-items: center;
+  -ms-flex-align: center;
+          align-items: center;
```
agr: Inconsistent indentation.

src/Bootstrap/dist/css/bootstrap-theme.css
```diff
@@ -545,6 +552,58 @@ img.reserved-indicator-icon {
...
+}
+.multi-select-dropdown .dropdown-content .dropdown-selector {
+  height: 300px;
+  overflow: auto;
+}
+.multi-select-dropdown .dropdown-content .dropdown-selector .dropdown-filter {
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display:         flex;
```
agr: extra spaces? (few more instances of the same below).

src/Bootstrap/dist/css/bootstrap-theme.css
scottbommarito: This file is auto-generated. May be a bug or intended behavior in [LESS](https://github.com/less/less.js).

src/Bootstrap/dist/css/bootstrap-theme.css
scottbommarito: This file is auto-generated. May be a bug or intended behavior in [LESS](https://github.com/less/less.js).

src/NuGetGallery/Scripts/gallery/common-multi-select-dropdown.js
```diff
@@ -0,0 +1,147 @@
...
+    });
+
+    // If an element outside of the dropdown gains focus, close it.
+    $(document).focusin(function (event) {
+        if (!self.isElementInsideDropdown(event.target)) {
+            self.dropdownOpen(false);
+        }
+    });
+
+    this.escapeKeyCode = 27;
```
agr: Javascript has constants, if I am not mistaken. Why do we need a property?

scottbommarito: Does it? I can't find the constant online...

agr: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const ?

scottbommarito: Oh I see, I thought you meant that there was a constant for the escape keycode built into the language. I'll look into `const`s.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
xavierdecoster: We need to align on the CVE state for those that users add to our db, but haven't yet been processed by the `UpdateVulnerabilityData` job. The data refresh operation has turned into a two-way sync due to this: we need to scan for user-added CVE entries in the gallery db, and we need to scan for modified CVE entries in the external source.

The NVD uses the following states for CVE enties:

```
    public enum CveStatus
    {
        Received = 0,
        AwaitingAnalysis = 1,
        UndergoingAnalysis = 2,
        Analyzed = 3,
        Modified = 4,
        Deferred = 5,
        Rejected = 6,
        Reserved = 7,
        Disputed = 8,
        Unverifiable = 9,
        Split = 10
    }
```

I think only those with a status of `Analyzed`, or above, are returned by their JSON data feeds. (https://nvd.nist.gov/vuln)

@scottbommarito  Can we settle on using `CveStatus.Received` for those that are user-provided? That's also the initial state on NVD-side.

xavierdecoster: Actually, doesn't matter for update logic, which needs to be able to overwrite status too. Nevermind :)

src/NuGetGallery/UrlHelperExtensions.cs
```diff
@@ -758,6 +758,15 @@ public static string GetOrganizationCertificates(this UrlHelper url, string acco
             return url.PackageVersionAction(nameof(PackagesController.RevalidateSymbols), package, relativeUrl);
         }
 
+        public static string DeprecatePackage(this UrlHelper url, bool relativeUrl = true)
+        {
+            return GetActionLink(
+                url, 
+                nameof(ManageDeprecationJsonApiController.Deprecate),
+                "ManageDeprecationJsonApi",
```
loic-sharma: You have trailing whitespace on lines 764 and 766

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1449,26 +1468,18 @@ public virtual ActionResult DeleteSymbols(string id, string version)
...
-                .Select(p => p.LatestSymbolPackage())
-                .Where(sp => sp != null && sp.StatusKey == PackageStatus.Available)
-                .Select(sp => new PackageViewModel(sp.Package));
+                .Where(p => (p.LatestSymbolPackage()?.StatusKey ?? PackageStatus.Deleted) == PackageStatus.Available);
 
-            model.VersionSelectList = new SelectList(
-                packageViewModelsForAllAvailableSymbolsPackage
-                .Select(pvm => new
+            model.VersionSelectList = versionsWithSymbols
+                .Select(versionWithSymbols => new SelectListItem
```
loic-sharma: This is a good change, but it's independent of this pull request. Considering the size of this pull request, please move this change into its own pull request

src/NuGetGallery/Scripts/gallery/common-multi-select-dropdown.js
```diff
@@ -0,0 +1,147 @@
...
+    this.dropdownOpen = ko.observable(false);
+    this.toggleDropdown = function () {
+        self.dropdownOpen(!self.dropdownOpen());
+    };
+
+    this.isAncestor = function (element, ancestorSelector) {
+        var $target = $(element);
+        // '.closest' returns the list of ancestors between this element and the selector.
+        // If the selector is not an ancestor of the element, it returns an empty list.
+        return $target.closest(ancestorSelector).length;
```
loic-sharma: >$target.closest(ancestorSelector).length [](start = 15, length = 40)

This will return a number, right? Can we cast this to a bool?

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
```diff
@@ -0,0 +1,405 @@
...
+
+    // The ID that has been typed into the textbox but not yet submitted.
+    this.addId = ko.observable('');
+    this.addLabel = addLabel;
+    this.add = function () {
+        self.addedIds.push(self.addId());
+        self.addId('');
+    };
+    this.addKeyDown = function (data, event) {
+        if (event.which === 13) { /* Enter */
```
loic-sharma: >13 [](start = 28, length = 2)

Consider adding a keycode like the escape key on the dropdown page

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -2416,6 +2416,49 @@ public async Task WhenNoReadMe_ReturnsNull(User currentUser, User owner)
                 Assert.Null(model.ReadMe.ReadMe.SourceType);
                 Assert.Null(model.ReadMe.ReadMe.SourceText);
             }
+
+            public static IEnumerable<object[]> ManageDeprecationFeatureFlagIsSetInModel_Data =
+                MemberDataHelper.Combine(
+                    Owner_Data,
+                    MemberDataHelper.AsDataSet(false, true));
```
loic-sharma: >AsDataSet [](start = 37, length = 9)

Why not `BooleanDataSet`?

scottbommarito: Created `BooleanDataSet` after this test was made...will use it here as well.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -48,5 +65,168 @@ public JsonResult GetCweIds(string query)
...
+            }
+
+            cveIds = cveIds ?? Enumerable.Empty<string>();
+            var cves = _deprecationService.GetCvesById(cveIds);
+            if (cveIds.Count() != cves.Count)
+            {
+                return DeprecateErrorResponse(HttpStatusCode.NotFound, Strings.DeprecatePackage_MissingCve);
+            }
+
+            if (cvssRating.HasValue && (cvssRating < 0 || cvssRating > 10))
```
loic-sharma: >cvssRating [](start = 16, length = 10)

Can we move this validation to the top of the request? Let's not hit the database if we know the request is invalid.

scottbommarito: Good call.

src/NuGetGallery/Scripts/gallery/common-multi-select-dropdown.js
scottbommarito: added `!!`

src/NuGetGallery/UrlHelperExtensions.cs
scottbommarito: fixed

src/NuGetGallery/Controllers/PackagesController.cs

PR #6917, opened 2019-02-22
Show deprecation information for current package on the Display Package page

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6774

# Vulnerable and Legacy and CVSS collapsed

image

# Vulnerable and Legacy and CVSS expanded

image

# Other

image

# Vulnerable and CVSS

image

# Legacy collapsed

image

# Legacy expanded

image

# Vulnerable and CVSS collapsed

image

# Vulnerable and CVSS expanded

image

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -110,6 +110,7 @@ namespace NuGetGallery
         private readonly ICoreLicenseFileService _coreLicenseFileService;
         private readonly ILicenseExpressionSplitter _licenseExpressionSplitter;
         private readonly IFeatureFlagService _featureFlagService;
+        private readonly IPackageDeprecationService _deprecationService;
 
         public PackagesController(
             IPackageService packageService,
@@ -137,7 +138,8 @@ namespace NuGetGallery
             IDiagnosticsService diagnosticsService,
             ICoreLicenseFileService coreLicenseFileService,
             ILicenseExpressionSplitter licenseExpressionSplitter,
-            IFeatureFlagService featureFlagService)
+            IFeatureFlagService featureFlagService,
+            IPackageDeprecationService deprecationService)
         {
             _packageService = packageService;
             _uploadFileService = uploadFileService;
@@ -165,6 +167,7 @@ namespace NuGetGallery
             _coreLicenseFileService = coreLicenseFileService ?? throw new ArgumentNullException(nameof(coreLicenseFileService));
             _licenseExpressionSplitter = licenseExpressionSplitter ?? throw new ArgumentNullException(nameof(licenseExpressionSplitter));
             _featureFlagService = featureFlagService ?? throw new ArgumentNullException(nameof(featureFlagService));
+            _deprecationService = deprecationService ?? throw new ArgumentNullException(nameof(deprecationService));
         }
 
         [HttpGet]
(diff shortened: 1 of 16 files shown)


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -652,7 +652,7 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
 
             Package package = null;
             // Load all packages with the ID.
-            var packages = _packageService.FindPackagesById(id);
+            var packages = _packageService.FindPackagesById(id, withDeprecations: true);
```
scottbommarito: I'm planning on reverting this and adding a call to `PackageDeprecationService` after I merge in the deprecation form PR.
This works fine, but loads all deprecation state for all versions, which is unnecessary because we only want the primary package.

src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
```diff
@@ -0,0 +1,143 @@
...
+        {
+            if (Model.CvssRating >= 7)
+            {
+                subclass = "warning";
+                icon = "Warning";
+            }
+            else
+            {
+                subclass = "danger";
+                icon = "ErrorBadge";
```
xavierdecoster: For which ranges of CvssRating do we use `info`, `warning` and `danger` subclasses?

Looking at this code, I derive the following:
* when not vulnerable, we use `info`
* when vulnerable and CvssRating >= 7, we use `warning`
* when vulnerable and CvssRating < 7, we use 'danger'

I would assume that anything `< 7` would be a `warning`, and anything above would be `danger`?

scottbommarito: Good catch, as you inferred I mixed up `warning` and `danger` here.

The intention is:
- deprecations that are not vulnerable should be `info`
- vulnerable deprecations with CVSS >= 7 are danger
- vulnerable deprecations with CVSS < 7 are warning

scottbommarito: fixed

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: Let's not merge this until we only load what we need. Otherwise, we need to do some perf testing for the many version case.

scottbommarito: Talked to @anangaur and he discussed that he wants to make changes to the version list based on deprecation, so we will need all this information.

joelverhagen: Okay, sounds good.

We need to perf testing then. How does this affect showing many versions on display package page?

scottbommarito: @joelverhagen - as discussed offline, I will split version history changes into a different PR, and postpone perf testing until then. This PR only adds a single query for fetching a single deprecation and should have no effect on perf.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -737,6 +740,11 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
                 return HttpNotFound();
             }
 
+            // Load the deprecation associated with this package.
+            // We want to fetch only the deprecation for the primary package.
+            // Fetching the deprecation for every version of the package is potentially very expensive.
+            _deprecationService.GetDeprecationByPackage(package);
+
```
loic-sharma: This looks unused?

scottbommarito: Discussed offline--made `DisplayPackageViewModel` accept the deprecation returned by this method rather than relying on EF context behavior.

PR #6918, opened 2019-02-23
Add functional tests for error cases in gallery

joelverhagen:
This just adds test coverage of current behavior. Cases asserted:

1. Bad cookie
1. Missing route (simple 404)
1. Nasty URLs that IIS does not like.
1. All possible inputs for the "simulate error" endpoints.

Progress on https://github.com/NuGet/NuGetGallery/issues/6601.

### DIFF
--- a/tests/NuGetGallery.FunctionalTests/ErrorHandling/ErrorHandlingTests.cs
+++ b/tests/NuGetGallery.FunctionalTests/ErrorHandling/ErrorHandlingTests.cs
@@ -1,0 +1,212 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace NuGetGallery.FunctionalTests.ErrorHandling
+{
+    public class ErrorHandlingTests : GalleryTestBase, IDisposable
+    {
+        private readonly HttpClientHandler _httpClientHandler;
+        private readonly HttpClient _httpClient;
+
+        public ErrorHandlingTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
+        {
+            _httpClientHandler = new HttpClientHandler
+            {
+                AllowAutoRedirect = false,
+                UseCookies = false,
+            };
+            _httpClient = new HttpClient(_httpClientHandler);
+        }
+
+        /// <summary>
+        /// Verify the behavior when a corrupted cookie is sent back to the server.
+        /// </summary>
+        [Theory]
+        [InlineData("__Controller::TempData", "Message=You successfully uploaded z̡̜͍̈̍̐̃̊͋́a̜̣͍̬̞̝͉̽ͧ͗l̸̖͕̤̠̹̘͖̃̌ͤg͓̝͓̰̀ͪo͈͌ 1.0.0.")]
+        public async Task RejectedCookie(string name, string value)
+        {
+            // Arrange
+            var relativePath = $"/packages/{Constants.TestPackageId}";
+            var requestUri = GetRequestUri(relativePath);
+            using (var request = new HttpRequestMessage(HttpMethod.Get, requestUri))
+            {
+                request.Headers.TryAddWithoutValidation("Cookie", $"{name}={value}");
+
+                // Act
+                var response = await GetTestResponseAsync(relativePath, request);
+
+                // Assert
+                Validator.SimpleHtml(HttpStatusCode.BadRequest)(response);
+            }
+        }
+
+        /// <summary>
+        /// Verify the behavior when a URL with restricted characters is used.
+        /// </summary>
+        [Theory]
+        [InlineData("/api/v2/lazy/%3E=1.0.5%20%3C1.1")]
+        [InlineData("/api/v2/mysql/*")]
+        [InlineData("/api/v2/comb/%3E=0.0.2")]
+        [InlineData("/.nuGetV3/feed.json:properties")]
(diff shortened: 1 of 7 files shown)


tests/NuGetGallery.FunctionalTests/ErrorHandling/ErrorHandlingTests.cs
```diff
@@ -0,0 +1,210 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// The expected responses for a <see cref="SimulatedErrorRequest"/>. If a key is not present, the
+        /// <see cref="SimulateError(EndpointType, SimulatedErrorType)"/> test will use the
+        /// <see cref="Validator.PrettyInternalServerError"/> validator.
+        /// </summary>
+        public static IReadOnlyDictionary<SimulatedErrorRequest, Action<TestResponse>> ExpectedSimulatedErrorResponses = new Dictionary<SimulatedErrorRequest, Action<TestResponse>>
```
joelverhagen: This behavior isn't necessarily desired today but it is the current behavior.

If we change our HTTP pipeline or some IIS configuration, this will tell us.

joelverhagen: And yes, it's maddeningly unpredictable.

tests/NuGetGallery.FunctionalTests/ErrorHandling/TestResponse.cs
```diff
@@ -0,0 +1,125 @@
...
+        public HttpStatusCode StatusCode { get; }
+        public string ReasonPhrase { get; }
+        public string ContentTypeHeader { get; }
+        public string LocationHeader { get; }
+        public int ContentLength { get; }
+        public string Content { get; }
+
+        public bool IsPrettyHtml => Content.Contains("page-error")
+            && Content.Contains("Oops")
+            && Content.Contains("Get me out of here!");
```
cristinamanum: Why these strings?

joelverhagen: This is a sniff test of this page:
https://www.nuget.org/does-no-exist

tests/NuGetGallery.FunctionalTests/ErrorHandling/Validator.cs
```diff
@@ -0,0 +1,159 @@
...
+                }
+
+                Assert.Equal($"aspxerrorpath={expectedRelativePath}", queryString);
+                Assert.Contains(errorPageName, actualErrorPageName);
+
+                AssertContentIsNotPrettyHtml(response);
+            };
+        }
+
+        public static Action<TestResponse> Xml()
```
cristinamanum: Will all the responses with xml content type  have 500 status code? Is it worth to pass the expected status code as argument?

joelverhagen: Today this is the only possible state when the response is XML. If we change server side later, then we can react at that point by modifying the functional test.

tests/NuGetGallery.FunctionalTests/ErrorHandling/ErrorHandlingTests.cs
```diff
@@ -0,0 +1,210 @@
...
+        [InlineData("/api/v2/comb/%3E=0.0.2")]
+        [InlineData("/.nuGetV3/feed.json:properties")]
+        [InlineData("/.nuGetV3/registration-semver2/serilog/page/0.1.6/1.2.47.json:properties")]
+        public async Task RejectedUrl(string relativePath)
+        {
+            // Arrange & Act
+            var response = await GetTestResponseAsync(relativePath);
+
+            // Assert
+            Validator.Redirect("500")(response);
```
skofman1: The behavior of not following redirects in this case should be documented and explained why you chose to test this way. I recommend summary for the class to make it extra visible.

tests/NuGetGallery.FunctionalTests/ErrorHandling/ErrorHandlingTests.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// This is a short-hand for instantiating a <see cref="SimulatedErrorRequest"/>.
+        /// </summary>
+        private static SimulatedErrorRequest SER(EndpointType endpointType, SimulatedErrorType simulatedErrorType)
+        {
+            return new SimulatedErrorRequest(endpointType, simulatedErrorType);
+        }
+
+        private async Task<TestResponse> GetTestResponseAsync(string relativePath, HttpRequestMessage request)
+        {
+            TestOutputHelper.WriteLine($"Request:  {request.Method} {request.RequestUri.AbsoluteUri}");
```
loic-sharma: >   [](start = 49, length = 2)

Super nit: there's two spaces here :)

joelverhagen: Intentional. I like it when a list of KVPs line up with fixed-width font

joelverhagen: To be clear "Response:" has one more letter than "Request:"

tests/NuGetGallery.FunctionalTests/ErrorHandling/TestResponse.cs
```diff
@@ -0,0 +1,125 @@
...
+                {
+                    sb.Append("'");
+                }
+
+                sb.Append(value);
+
+                if (quotes)
+                {
+                    sb.Append("'");
+                }
```
agr:
```C#
if (quotes)
{
    sb.AppendFormat("'{0}'", value);
}
else
{
    sb.Append(value);
}
```
?
Looks a bit more concise.

agr: shorter but uglier:
```C#
    sb.AppendFormat("{1}{0}{1}", value, quotes ? "'" : "");
```

joelverhagen: Did the first.

PR #6929, opened 2019-02-25
Remove Atom feed functional tests that depend on 3rd party package

joelverhagen:
This behavior is asserted by unit tests so a functional test is not strictly necessary.

Unit tests:
https://github.com/NuGet/NuGetGallery/blob/651381b1506a8c29eb879831f592514bcee5a8fa/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs#L1178
https://github.com/NuGet/NuGetGallery/blob/651381b1506a8c29eb879831f592514bcee5a8fa/tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs#L1253


joelverhagen:
Note that this is a hotfix.

PR #6930, opened 2019-02-26
Site admins can see email address on profile page

joelverhagen:
This is currently already possible with the database admin view, but this makes it much more convenient. If the user is not a site admin, this information will remain private -- as it should be!

Address https://github.com/NuGet/Engineering/issues/609.

Looks like this:
image


joelverhagen:
Uses this Bootstrap for dl/dt/dd:
https://getbootstrap.com/docs/3.3/css/#description

PR #6932, opened 2019-02-26
Remove flaky download count assertion

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/1952.

This is broken since stats don't sync between search and gallery simultaneously.

### DIFF
--- a/tests/NuGetGallery.FunctionalTests.Core/Helpers/ClientSdkHelper.cs
+++ b/tests/NuGetGallery.FunctionalTests.Core/Helpers/ClientSdkHelper.cs
@@ -576,37 +576,6 @@ namespace NuGetGallery.FunctionalTests
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


scottbommarito:
Is there any assertion we can do instead that isn't flaky? Could we check for ANY download count instead of a specific value?

joelverhagen:
> Is there any assertion we can do instead that isn't flaky? 

Theoretically we could regex out the download count and make sure it is within X% of the value found in search... but is this really worth the effort? How many times has this test actually caught a real problem?

> Could we check for ANY download count instead of a specific value?

Wouldn't this just be asserting the shape of the HTML? How is this beneficial?

In short, I am not convinced that the test coverage (past observed) that this test provides is worth the pain we have experienced from it.

joelverhagen:
@scottbommarito, tracked here: https://github.com/NuGet/NuGetGallery/issues/6933.

PR #6937, opened 2019-02-26
Minor fixes to the GET DeleteSymbols endpoint

scottbommarito:
A couple fixes that were done while working on https://github.com/NuGet/NuGetGallery/pull/6912 but moved to here because they are separate to it

Two issues fixed here:

1. [`DeleteSymbols` requires all versions of the package but doesn't call `FindPackagesById`](https://github.com/NuGet/NuGetGallery/pull/6905)
2. `DeleteSymbols` uses `SelectList` (which uses reflection and dynamic objects) instead of `IEnumerable<SelectListItem>`

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1526,7 +1526,8 @@ namespace NuGetGallery
             // Fetch all versions of the package with symbols.
             var versionsWithSymbols = packages
                 .Where(p => p.PackageStatusKey != PackageStatus.Deleted)
-                .Where(p => (p.LatestSymbolPackage()?.StatusKey ?? PackageStatus.Deleted) == PackageStatus.Available);
+                .Where(p => (p.LatestSymbolPackage()?.StatusKey ?? PackageStatus.Deleted) == PackageStatus.Available)
+                .OrderByDescending(p => new NuGetVersion(p.Version));
 
             model.VersionSelectList = versionsWithSymbols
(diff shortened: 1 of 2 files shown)


scottbommarito:
(Test will be fixed after I pull in https://github.com/NuGet/NuGetGallery/pull/6912.)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1507,26 +1523,18 @@ public virtual ActionResult DeleteSymbols(string id, string version)
...
-            var packageViewModelsForAllAvailableSymbolsPackage = package
-                .PackageRegistration
-                .Packages
+            // Fetch all versions of the package with symbols.
+            var versionsWithSymbols = packages
                 .Where(p => p.PackageStatusKey != PackageStatus.Deleted)
-                .Select(p => p.LatestSymbolPackage())
-                .Where(sp => sp != null && sp.StatusKey == PackageStatus.Available)
-                .Select(sp => new PackageViewModel(sp.Package));
+                .Where(p => (p.LatestSymbolPackage()?.StatusKey ?? PackageStatus.Deleted) == PackageStatus.Available);
```
agr: What's the point of `?? PackageStatus.Deleted`? `null` by itself is not equal to any enum value.

scottbommarito: `enum`s are `int`s internally, which means they don't support `null`.
Therefore, I must convert the `null` to a valid `enum` value before I can compare.

PR #6943, opened 2019-02-27
Unknown CVEs should be created when the deprecation form is submitted

scottbommarito:
Right now, if we haven't yet ingested a CVE or CWE and users specify it in the form, we will fail the request with `NotFound`. With this change, these CVEs or CWEs are instead created in an unlisted state (they do not appear in autocomplete). When we eventually ingest the CVE or CWE, its entity will be updated with the full data.

I insert the CVE/CWE with null fields except `CveId`/`CweId`, `Status` (which is set to the new value of `Unknown`), and `Listed (which is set to false). As a result, I had to remove some DB constraints that expected fields to be non-null.

We verify that the CVE and CWE IDs are valid before we create them. I will be adding validation to the form as well, so a user will have to hack the form in order to submit it with an invalid CVE/CWE.


src/NuGetGallery/Strings.resx
```diff
@@ -1109,11 +1109,11 @@ The {1} Team</value>
   <data name="DeprecatePackage_NoVersions" xml:space="preserve">
     <value>You must select at least one version to deprecate!</value>
   </data>
-  <data name="DeprecatePackage_MissingCve" xml:space="preserve">
-    <value>At least one CVE specified could not be found.</value>
+  <data name="DeprecatePackage_InvalidCve" xml:space="preserve">
+    <value>The CVE ID '{0}' is invalid.</value>
```
skofman1: The error message can be friendlier. It should contain what's the expected format and perhaps even link to documentation

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -16,6 +17,14 @@ namespace NuGetGallery
     public partial class ManageDeprecationJsonApiController
         : AppController
     {
+        private static readonly TimeSpan RegexTimeout = TimeSpan.FromMinutes(1);
+
+        public const string CveIdRegexPattern = @"CVE-\d{4}-\d+";
+        private static readonly Regex CveIdRegex = GetRegexFromPattern(CveIdRegexPattern);
+
+        public const string CweIdRegexPattern = @"CWE-\d+";
```
xavierdecoster: As CWE IDs `1-9` are already taken, should we enforce minimum length for digits when creating new CWE's? Newest CWE IDs are > 1000, so 4 characters seems like a good minimum?

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -16,6 +17,14 @@ namespace NuGetGallery
     public partial class ManageDeprecationJsonApiController
         : AppController
     {
+        private static readonly TimeSpan RegexTimeout = TimeSpan.FromMinutes(1);
+
+        public const string CveIdRegexPattern = @"CVE-\d{4}-\d+";
```
xavierdecoster: For CVE IDs, we should also validate the `{year}` part. Should be a reasonable number. CVE ID listing starts with `CVE-1999-0001`, so the year part should not be `< 1999`. Nor should it be `> {DateTime.UtcNow.Year}`.

src/NuGetGallery/Strings.resx
xavierdecoster: For inspiration, see `AutocompleteCveIds_ValidationError` and `AutocompleteCweIds_ValidationError` in `Strings`.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: Good call.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: By the time this is live, we should have all the CWEs imported, so it shouldn't be possible for users to create CWEs that low, correct?

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: Is it possible that someone may want to post a CVE/CWE that is going to be posted in the near future? For example, I applied for a CVE/CWE, but it won't be made official for the next week. I was thinking maybe `> (DateTime.UtcNow.Year + 1)` would be more resilient.

xavierdecoster: AFAIK, the `{year}` part is assigned the moment the ID is reserved/provided, and doesn't change anymore, even if it's being disclosed in the year after.

This is also a good resource on CVE ID syntax and validation guidance: https://cve.mitre.org/cve/identifiers/tech-guidance.html

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
xavierdecoster: Not true. There are CWE IDs lower that aren't imported, because they are of type `Category` or `View`. CWE data has a tree structure, and we only import the leaves without children.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: Added.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: I don't think we can stop invalid CVEs or CWEs from being created regardless of what we do. If we restrict to `> 1000`, someone could always do `99999`. If we restrict to between a range, we'll eventually have to update the range when CWEs overflow. I think this is fine for now, especially given that the feature is behind a flag.

src/NuGetGallery/Strings.resx
scottbommarito: Improved!

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: Decided to restrict CWEs as discussed.

PR #6944, opened 2019-02-27
Integrate Deprecate form with CVE/CWE autocomplete

scottbommarito:
# View without autocomplete:

image

# CVE autocomplete:

image

# CWE autocomplete:

image

# Invalid ID:

image

# Unknown ID:

image

### DIFF
--- a/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
+++ b/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
@@ -17,9 +17,6 @@ namespace NuGetGallery
     public partial class ManageDeprecationJsonApiController
         : AppController
     {
-        public const string GetCveIdsActionName = "CveIds";
-        public const string GetCweIdsActionName = "CweIds";
-
         private static readonly TimeSpan RegexTimeout = TimeSpan.FromMinutes(1);
 
         public const string CveIdRegexYearGroupName = "year";
@@ -48,7 +45,7 @@ namespace NuGetGallery
 
         [HttpGet]
         [UIAuthorize]
-        [ActionName(GetCveIdsActionName)]
+        [ActionName(ActionName.GetCveIds)]
         public JsonResult GetCveIds(string query)
         {
             // Get CVE data.
@@ -64,7 +61,7 @@ namespace NuGetGallery
 
         [HttpGet]
         [UIAuthorize]
-        [ActionName(GetCweIdsActionName)]
+        [ActionName(ActionName.GetCweIds)]
         public JsonResult GetCweIds(string query)
         {
             // Get CWE data.
--- a/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
+++ b/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
@@ -17,6 +17,9 @@ namespace NuGetGallery
     public partial class ManageDeprecationJsonApiController
         : AppController
     {
+        public const string GetCveIdsActionName = "CveIds";
+        public const string GetCweIdsActionName = "CweIds";
+
         private static readonly TimeSpan RegexTimeout = TimeSpan.FromMinutes(1);
 
         public const string CveIdRegexYearGroupName = "year";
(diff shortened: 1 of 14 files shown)


xavierdecoster:
Out of curiosity, have you tried using real data to see how the UI looks like with longer descriptions and titles? 
(you can use the importer locally from https://github.com/NuGet/NuGet.Jobs/pull/722)

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
```diff
@@ -163,13 +163,45 @@ public VersionDeprecationState(Package package, string text)
...
             public string AlternatePackageVersion { get; }
             public string CustomMessage { get; }
             public bool ShouldUnlist { get; }
+
+            /// <remarks>
+            /// Ideally, the fields of this class would be pascal-case.
+            /// Unfortunately, however, the serializer that MVC uses interally (JavaScriptSerializer) does not support changing the serialized name of a property.
+            /// The JS on the client-side depends on the exact naming of these fields, and it would add unnecessary complexity to do the conversion on the client-side.
+            /// </remarks>
+            public class VulnerabilityDetailState
```
xavierdecoster: Should we have unit tests ensuring these property names never change casing?

scottbommarito:
@xavierdecoster - that's a good call, I'll probably have to add an `overflow` to the autocomplete dropdown.

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
scottbommarito: Not sure--I don't think it provides much value, because we don't have tests on the Javascript side that assure that the code there uses the same property names. I don't think it's worthwhile to test this until we invest in JS testing.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -16,6 +16,9 @@ namespace NuGetGallery
     public partial class ManageDeprecationJsonApiController
         : AppController
     {
+        public const string GetCveIdsActionName = "CveIds";
```
skofman1: Should be in [https://github.com/NuGet/NuGetGallery/blob/651381b1506a8c29eb879831f592514bcee5a8fa/src/NuGetGallery/RouteName.cs](https://github.com/NuGet/NuGetGallery/blob/651381b1506a8c29eb879831f592514bcee5a8fa/src/NuGetGallery/RouteName.cs)

Like we did for organizations:
[https://github.com/NuGet/NuGetGallery/blob/bf7e98f696e346b55cb1eba38866682ae936160f/src/NuGetGallery/Controllers/UsersController.cs#L122](https://github.com/NuGet/NuGetGallery/blob/bf7e98f696e346b55cb1eba38866682ae936160f/src/NuGetGallery/Controllers/UsersController.cs#L122)

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
skofman1: Not sure having both CVE and CWE data in a single class is the best solution... they do share some data but not all data (like name, cvss is different).
I recommend splitting to two classes (with maybe shared base class).

scottbommarito: This class represents the model used to share data between the server and the client-side JS. The client side treats CVEs and CWEs identically, but with slightly different UI based on the presence of name/description. IMO adding additional class structure here serves no purpose and complicates the code. Perhaps I can add a comment to make the purpose of these classes (all classes in this file named `State`) more clear?

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: As discussed offline, `RouteName`s and `ActionName`s are different concepts, and I think it's confusing to group them together. I do agree with moving these to a shared file, so I will create `ActionName.cs` for this purpose.

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
scottbommarito: Added more clear comments.

scottbommarito: Added clear comments.

PR #6951, opened 2019-03-01
Action names should not be inline strings

scottbommarito:
(The idea for this PR was derived from https://github.com/NuGet/NuGetGallery/pull/6944#discussion_r261384308)

Our uses of actions is very brittle. There is no structural connection between the names of methods and our routing, which means that if controller names are changed either accidentally or purposefully, we may not catch it when the routing breaks. When preparing this PR, I noticed a couple (legacy) routes that were completely broken.

This PR ensures that all routing either refers to actions by a constant in `ActionName` or by a `nameof(...)` call. Additionally, we are incorrectly referring to action names as route names in a couple places. This includes some of the uses of `RouteName.TransformToOrganization` and a couple methods in `UrlHelper`, so I fixed those while working on this.

The process for creating this PR was the following:
- Where we used `ActionNameAttribute`, replace the hard-coded string value with a const in `ActionName.cs` (similar to `RouteName.cs`)
- Update `Routes.cs` to use proper action names (either a value in `ActionName.cs` or a `nameof(...)` call
- Update all calls to `Url.Action(...)` to use proper action names
- Update all calls to `Url.BeginForm(...)` to use proper action names
- Update all calls to `RedirectToAction` to use proper action names
- Verified that no actions with `ActionName.cs` constants were referred to by `nameof(...)` calls

Unfortunately, MVC won't allow us to use `nameof(...)` in views, so action names referred to in views needed to have `ActionName.cs` constants created for them. Additionally, I made sure that all values of `ActionName.cs` are consistent with the existing name of the action so that there should be no functional change.

As a side note, this problem still exists with controller names and view names, but I was unsure how to solve it there, but I think it should be considered out of scope of this PR.

For controllers, controller names are not the names of their classes, but instead are their name with the "controller" suffix removed. Furthermore, there is no `ControllerNameAttribute` that could be used to move these strings into constants. I could make a helper function that removes "Controller" for the end of controller names, but ultimately this isn't a full link to the name of the controller.

For views, the name refers to a filename, although that filename is usually the name of an action. AFAIK, there isn't a way to link a variable to the name of a file. Given this, I don't see any way I could improve the situation.


src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -104,11 +104,6 @@ public static void RegisterUIRoutes(RouteCollection routes)
                 "json/deprecation/{action}",
                 new { controller = "ManageDeprecationJsonApi" });
 
-            routes.MapRoute(
```
scottbommarito: I believe this is leftover from before the rework, when we rewrote how the `About` page loaded site contributors.

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -591,62 +586,57 @@ public static void RegisterUIRoutes(RouteCollection routes)
...
                 constraints: new { httpMethod = new HttpMethodConstraint("GET") });
 
             routes.MapRoute(
                 "v1" + RouteName.PushPackageApi,
                 "v1/PackageFiles/{apiKey}/nupkg",
-                defaults: new { controller = "Api", action = "PushPackageApi" },
+                defaults: new { controller = "Api", action = ActionName.PushPackageApi },
                 constraints: new { httpMethod = new HttpMethodConstraint("POST") });
 
-            routes.MapRoute(
```
scottbommarito: Appears to be a leftover V1 delete package route that no longer works (there is no `DeletePackages` method on ApiController). I verified that it hasn't been called in the last week.

scottbommarito:
Discussed offline with some people. This is a good change, but currently untestable. I tried adding unit tests around it but was unable to write testing that was both quick to write and thorough. We should follow these conventions in the future but as a bulk change, this is unmanageable.

PR #6955, opened 2019-03-05
Unlisting a package when deprecating it should update LastEdited

scottbommarito:
https://github.com/NuGet/Engineering/issues/2082

This is a bug fix for the behavior of the deprecation form on the manage package page. Without setting `LastEdited`, the change is not pushed to V3.

In the future, when deprecation information is pushed to V3 as well, this route should always set `LastEdited`, but right now, because deprecation information is not push to V3, it only needs to update `LastEdited` when we are unlisting.

### DIFF
--- a/src/NuGetGallery/Services/PackageDeprecationService.cs
+++ b/src/NuGetGallery/Services/PackageDeprecationService.cs
@@ -15,15 +15,21 @@ namespace NuGetGallery
         private readonly IEntityRepository<PackageDeprecation> _deprecationRepository;
         private readonly IEntityRepository<Cve> _cveRepository;
         private readonly IEntityRepository<Cwe> _cweRepository;
+        private readonly IPackageService _packageService;
+        private readonly IIndexingService _indexingService;
 
         public PackageDeprecationService(
            IEntityRepository<PackageDeprecation> deprecationRepository,
            IEntityRepository<Cve> cveRepository,
-           IEntityRepository<Cwe> cweRepository)
+           IEntityRepository<Cwe> cweRepository,
+           IPackageService packageService,
+           IIndexingService indexingService)
         {
             _deprecationRepository = deprecationRepository ?? throw new ArgumentNullException(nameof(deprecationRepository));
             _cveRepository = cveRepository ?? throw new ArgumentNullException(nameof(cveRepository));
             _cweRepository = cweRepository ?? throw new ArgumentNullException(nameof(cweRepository));
+            _packageService = packageService ?? throw new ArgumentNullException(nameof(packageService));
+            _indexingService = indexingService ?? throw new ArgumentNullException(nameof(indexingService));
         }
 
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -102,6 +102,7 @@ public class PackageDeprecationService : IPackageDeprecationService
                     if (shouldUnlist)
```
skofman1: If you review the code in ApiController DeletePackage
[https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Controllers/ApiController.cs#L799](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Controllers/ApiController.cs#L799)

you will see it does many additional things:
1. Updates LastUpdated
2. auditing
3. telemetry
4. Updates Lucene index for local Gallery scenario

Unlisting as part of deprecation should also do all those things. Look into reusing the same code. Also, consider the "List" package scenario, when we are "undeprecating"

scottbommarito: Great feedback. Will reuse the code from `PackageService.MarkPackageUnlisted`.

One thing though: "undeprecating" does not relist packages.

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -101,7 +107,7 @@ public class PackageDeprecationService : IPackageDeprecationService
 
                     if (shouldUnlist)
                     {
-                        package.Listed = false;
+                        await _packageService.MarkPackageUnlistedAsync(package, false);
```
skofman1: named param please

cristinamanum: This will not commit the changes. Where are the changes committed? It would be great to add a comment about this.

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -116,6 +122,11 @@ public class PackageDeprecationService : IPackageDeprecationService
             }
 
             await _deprecationRepository.CommitChangesAsync();
+
+            foreach (var package in packages)
```
cristinamanum: What you like to achieve through this ? Please add comments.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: See the `deprecationRepository.CommitChangesAsync` call below.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: We are updating the indexing of the packages we just updated above.
I will add a comment describing why we are doing it here.

PR #6961, opened 2019-03-05
Improve deprecation information UI on Display Package page with multiple reasons

scottbommarito:
@anangaur had some additional feedback for https://github.com/nuget/nugetgallery/issues/6774

If you remember before, we merged a single string for the reason text and then had text/icons centered with the reason text. We now have separate text for each reason and text/icons aligned with the top line.

We also changed the text for legacy only.

# Vulnerable and Legacy and Other with CVSS collapsed

image

# Vulnerable and Legacy and Other with CVSS expanded

image

# Vulnerable and Legacy with CVSS

image

# Vulnerable and Legacy

image

# Vulnerable and Legacy collapsed

image

# Vulnerable and Legacy expanded

image

# Legacy collapsed

image

# Legacy expanded

image

### DIFF
--- a/src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
+++ b/src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
@@ -33,61 +33,62 @@
     <div class="icon-text alert alert-warning">
         <div id="show-deprecation-content-container" class="deprecation-expander" tabindex="0" data-toggle="collapse" data-target="#deprecation-content-container" aria-expanded="false" aria-controls="deprecation-content-container" aria-labelledby="deprecation-container-label">
             <div class="deprecation-expander">
-                <div class="deprecation-expander-icon-container">
+                <div class="deprecation-expander-container">
                     <i class="deprecation-expander-icon ms-Icon ms-Icon--Warning" aria-hidden="true"></i>
+
+                    <div id="deprecation-container-label" class="deprecation-expander-info-right">
+                        @{
+                            var isVulnerable = Model.DeprecationStatus.HasFlag(PackageDeprecationStatus.Vulnerable);
+                            var isLegacy = Model.DeprecationStatus.HasFlag(PackageDeprecationStatus.Legacy);
+                            if (isVulnerable || isLegacy)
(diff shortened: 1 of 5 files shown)


anangaur:
The messaging if in the second line could be better for legacy. We could just say `This package is no longer maintained`.

src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
```diff
@@ -33,61 +33,58 @@
...
+                <div class="deprecation-expander-container">
                     <i class="deprecation-expander-icon ms-Icon ms-Icon--Warning" aria-hidden="true"></i>
+
+                    <div id="deprecation-container-label" class="deprecation-expander-info-right">
+                        @{
+                            var isVulnerable = Model.DeprecationStatus.HasFlag(PackageDeprecationStatus.Vulnerable);
+                            var isLegacy = Model.DeprecationStatus.HasFlag(PackageDeprecationStatus.Legacy);
+                            if (isVulnerable || isLegacy)
+                            {
+                                if (isVulnerable)
```
joelverhagen: Why nest ifs? Can't you just do `if / else if / else`?

scottbommarito: Good catch! Fixed it.

src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
```diff
@@ -33,61 +33,59 @@
...
                         {
-                            if (isVulnerable)
-                            {
-                                deprecationReasonString += " and ";
-                            }
-
-                            deprecationReasonString += "is no longer maintained by its owners";
+                            ratingString = "Medium";
+                        }
+                        else if (Model.CvssRating < 9)
```
joelverhagen: This should be unit tested. Please move this logic for the View model or something

scottbommarito: Good call, this is very easy to do.

PR #6965, opened 2019-03-05
Remove unlist checkbox from Manage Deprecation form

scottbommarito:
We have decided to remove the "should unlist" checkbox from the Manage Deprecation form. This PR removes it from the form.

Because this PR was also small, I also made a small UI fix that @anangaur asked for--the CVSS rating in the deprecation form should be bolded.

If we leave "should unlist" in the form, we'd need to do testing on whether or not bulk unlisting works. This is non-trivial and may require large V3 changes.

Additionally, we decided to separate "deprecation" and "listing" in the UI, and having a checkbox that bridges the two is confusing, especially when multiple versions are selected.

### DIFF
--- a/src/NuGetGallery/Views/Packages/_ManageDeprecation.cshtml
+++ b/src/NuGetGallery/Views/Packages/_ManageDeprecation.cshtml
@@ -98,12 +98,6 @@
             <label for="customMessage" class="deprecation-section-header">Provide custom message</label>
             <textarea id="customMessage" name="customMessage" class="form-control full-line" data-bind="textInput: customMessage" rows="2"></textarea>
         </div>
-        <div class="deprecation-section form-group">
-            <label>
-                <input type="checkbox" value="true" data-bind="checked: shouldUnlist" />
-                Unlist this package from search results
-            </label>
-        </div>
     </div>
 
     <div data-bind="visible: submitError">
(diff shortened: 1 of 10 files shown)


src/NuGetGallery/Views/Packages/_ManageDeprecation.cshtml
```diff
@@ -59,7 +59,7 @@
                 </div>
                 <div class="security-detail-input">
                     <input id="selectedCvssRating" class="form-control cvss-input" name="selectedCvssRating" type="text" placeholder="0.0" size="4" maxlength="4" data-bind="textInput: selectedCvssRating, disable: !hasCvss()" />
-                    <span data-bind="text: cvssRatingLabel, css: { 'text-danger': cvssRatingIsInvalid }"></span>
+                    <span class="text-" data-bind="text: cvssRatingLabel, css: { 'text-danger': cvssRatingIsInvalid, 'cvss-bold': !cvssRatingIsInvalid() }"></span>
```
joelverhagen: `text-`?

scottbommarito: woops--was trying to find an existing CSS class that bolds the text but was unable to

PR #6968, opened 2019-03-06
Snupkgs without symbols will be rejected.

cristinamanum:
Snupkgs without symbols will be rejected.

### DIFF
--- a/src/NuGetGallery/Services/SymbolPackageService.cs
+++ b/src/NuGetGallery/Services/SymbolPackageService.cs
@@ -156,7 +156,7 @@ namespace NuGetGallery
 
             if (!CheckForPDBFiles(symbolPackage))
             {
-                throw new InvalidDataException(string.Format(Strings.SymbolsPackage_InvalidFiles, PDBExtension));
+                throw new InvalidDataException(string.Format(Strings.SymbolsPackage_NoSymbols));
             }
         }
 
--- a/src/NuGetGallery/Services/SymbolPackageService.cs
+++ b/src/NuGetGallery/Services/SymbolPackageService.cs
@@ -153,6 +153,11 @@ namespace NuGetGallery
             {
                 throw new InvalidDataException(string.Format(Strings.SymbolsPackage_InvalidFiles, PDBExtension));
             }
+
+            if (!CheckForPDBFiles(symbolPackage))
+            {
+                throw new InvalidDataException(string.Format(Strings.SymbolsPackage_InvalidFiles, PDBExtension));
+            }
         }
 
         private static bool CheckForAllowedFiles(PackageArchiveReader symbolPackage)
@@ -169,6 +174,19 @@ namespace NuGetGallery
             }
 
             return true;
+        }
+
+        /// <summary>
+        /// If there are not any pdb files in the snupkg the publish action should be blocked.
+        /// </summary>
+        /// <param name="symbolPackage"></param>
+        /// <returns></returns>
+        private static bool CheckForPDBFiles(PackageArchiveReader symbolPackage)
+        {
+            return symbolPackage.GetFiles()
+                .Select( filePath => new FileInfo(filePath))
+                .Where( fileInfo => !string.IsNullOrEmpty(fileInfo.Name) && fileInfo.Extension == PDBExtension)
+                .Any();
         }
 
         private static bool IsPortable(string pdbFile)
--- a/src/NuGetGallery/Services/SymbolPackageService.cs
+++ b/src/NuGetGallery/Services/SymbolPackageService.cs
@@ -181,7 +181,7 @@ namespace NuGetGallery
         /// </summary>
         /// <param name="symbolPackage"></param>
         /// <returns></returns>
-        internal static bool CheckForPDBFiles(PackageArchiveReader symbolPackage)
+        private static bool CheckForPDBFiles(PackageArchiveReader symbolPackage)
         {
             return symbolPackage.GetFiles()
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -171,6 +176,19 @@ private static bool CheckForAllowedFiles(PackageArchiveReader symbolPackage)
...
 
+        /// <summary>
+        /// If there are not any pdb files in the snupkg the publish action should be blocked.
+        /// </summary>
+        /// <param name="symbolPackage"></param>
+        /// <returns></returns>
+        private static bool CheckForPDBFiles(PackageArchiveReader symbolPackage)
+        {
+            return symbolPackage.GetFiles()
+                .Select( filePath => new FileInfo(filePath))
```
loic-sharma: Can this code be simplified by using `Path.GetExtension` instead of `FileInfo`?

loic-sharma: Also, can we add a unit test verifying this behavior?

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -171,6 +176,19 @@ private static bool CheckForAllowedFiles(PackageArchiveReader symbolPackage)
...
+                .Where( fileInfo => !string.IsNullOrEmpty(fileInfo.Name) && fileInfo.Extension == PDBExtension)
```
joelverhagen: Is a `.PDB` accepted? Are we consistent across the board on accepting/denying .PDB or some other casing of the file extension?

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -171,6 +176,19 @@ private static bool CheckForAllowedFiles(PackageArchiveReader symbolPackage)
             return true;
         }
 
+        /// <summary>
+        /// If there are not any pdb files in the snupkg the publish action should be blocked.
+        /// </summary>
+        /// <param name="symbolPackage"></param>
+        /// <returns></returns>
+        private static bool CheckForPDBFiles(PackageArchiveReader symbolPackage)
```
joelverhagen: Need unit test.

src/NuGetGallery/Services/SymbolPackageService.cs
cristinamanum: Agree with Path.GetExtension. Added unit test.

src/NuGetGallery/Services/SymbolPackageService.cs
cristinamanum: No, we accept only .pdbs. We are consistent with lower case file extensions.

src/NuGetGallery/Services/SymbolPackageService.cs
cristinamanum: Added unit test.

src/NuGetGallery/Strings.resx
```diff
@@ -1138,4 +1138,7 @@ The {1} Team</value>
     <value>https://cwe.mitre.org/data/definitions/{0}.html</value>
     <comment>{0} is the CWE's ID number (e.g. CWE-999 has an ID number of 999)</comment>
   </data>
+  <data name="SymbolsPackage_NoSymbols" xml:space="preserve">
+    <value>The package does not contain any symbols.</value>
```
shishirx34: `... any symbol files (.pdb)` perhaps?

cristinamanum: Updated.

scottbommarito:
@cristinamanum - yes, they are separate issues, but the value of adding this validation is decreased when the validation error doesn't show properly on the client side. I would look into it, see if there's an easy fix, and if there is, apply it with this PR. If the fix is nontrivial, then I agree that we should fix it separately, but if there is a quick fix, it should be done alongside this work.

src/NuGetGallery/Services/SymbolPackageService.cs
```diff
@@ -171,6 +176,18 @@ private static bool CheckForAllowedFiles(PackageArchiveReader symbolPackage)
...
+        /// <summary>
+        /// If there are not any pdb files in the snupkg the publish action will be rejected.
+        /// </summary>
+        /// <param name="symbolPackage"></param>
+        /// <returns></returns>
+        private static bool CheckForPDBFiles(PackageArchiveReader symbolPackage)
+        {
+            return symbolPackage.GetFiles()
+                .Where(filePath => !string.IsNullOrEmpty(Path.GetFileName(filePath)) && Path.GetExtension(filePath) == PDBExtension)
+                .Any();
```
agr: `Any` accepts predicates the same way `Where` does, so `Where` call is redundand here.

tests/NuGetGallery.Facts/Services/SymbolPackageServiceFacts.cs
```diff
@@ -48,12 +49,16 @@ private static List<ClientPackageType> CreateSymbolPackageTypesObject()
         }
 
         private static Action<ZipArchive> CreatePopulatePackageAction(string extension)
+        {
+            return CreatePopulatePackageAction(new string[]{ extension });
+        }
+
+        private static Action<ZipArchive> CreatePopulatePackageAction(string[] extensions)
```
agr: This can be made 
```C#
CreatePopulatePackageAction(params string[] extensions)
```
There would be no need for `CreatePopulatePackageAction(string extension)` overload and calling it in general becomes easier.

cristinamanum: Updated

Issue #6954, opened 2019-03-04
Show deprecation status in the version history on the Display Package page


scottbommarito:
In addition to the information for the current package on the top of the page, we should show basic deprecation information in the version history.

Suggestions:
- make the version red
- strike-out the version
- gray out the version
- Add another column with "Listed", "Deprecated" or "Deprecated and Unlisted"

PR #6973, opened 2019-03-07
Mark deprecated versions in Display Package page's version history

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6954

Current implementation: icons on right with row border between

Regarding SQL performance, the `DisplayPackage` endpoint now loads the `PackageDeprecation`s associated with each `Package` as well, but not any entities linked to those deprecations (like the CVEs, CWEs, alternate packages, etc). In my testing, this added load seemed entirely insignificant. I am continuing to look into potential impact but I think I'm not sure what else we could do to decrease the risk of impact.

# As owner

image

with certs

image

# As non-owner

image

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -529,15 +529,18 @@
                                 rowCount++;
                                 @VersionListDivider(rowCount, versionsExpanded)
                                 <tr class="@(packageVersion.IsCurrent(Model) ? "bg-info" : null)">
-                                    <td class="signature-info-cell">
-                                        @if (Model.IsCertificatesUIEnabled && !string.IsNullOrEmpty(packageVersion.SignatureInformation))
+                                    <td class="package-info-cell">
+                                        @if (packageVersion.DeprecationStatus != PackageDeprecationStatus.NotDeprecated)
                                         {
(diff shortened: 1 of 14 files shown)


joelverhagen:
I think the strike-out is hard to read. What are @anangaur's thoughts in the area? I would prefer one of the other options that doesn't make it hard to read the versions. Maybe I have old man eyes.

anangaur:
I want them in red. Stroked off suggests ‘deleted’ which is not true

scottbommarito:
@joelverhagen @anangaur - I made the deprecated versions red as requested but I'm not sure how much I like it. I agree that it's easier to read than the strike-through, but red usually indicates an error, and deprecation is more of a warning.

scottbommarito:
@joelverhagen @anangaur - updated the PR with our agreed on UI

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -556,37 +551,75 @@
                                     {
                                         var packageStatusSummary = packageVersion.PackageStatusSummary;
 
-                                        if (packageStatusSummary == PackageStatusSummary.Listed ||
-                                            packageStatusSummary == PackageStatusSummary.Unlisted)
-                                        {
-                                            <td>
+                                        <td>
+                                            @if (packageStatusSummary == PackageStatusSummary.Listed ||
+                                                packageStatusSummary == PackageStatusSummary.Unlisted)
```
joelverhagen: nit: indent 1 more space

PR #6974, opened 2019-03-08
Improve version list display for deleted packages

scottbommarito:
Noticed this while working on https://github.com/NuGet/NuGetGallery/pull/6973

The version list display for deleted packages is odd. The signature information column is improperly aligned, and we have "(deleted" in the same column as the version, when there is a status column that we could easily leverage.

New:

image

with certificate icons

image

Old:

image

with certificate icons

image


scottbommarito:
Combined this into https://github.com/NuGet/NuGetGallery/pull/6973/files

PR #6975, opened 2019-03-08
Upper bound individual retries, move configurations in web config.

cristinamanum:
Two changes in this PR:
1. Move configurations in web.config.
2. Fix a potential issue with a case like:
- one single query will return 500+
- multiple queries will be fine
The circuit breaker will not be closed and the failing query will keep retrying until timing out. 
With this change only a certain amount of retries will be allowed.

### DIFF
--- a/src/NuGetGallery/Configuration/IAppConfiguration.cs
+++ b/src/NuGetGallery/Configuration/IAppConfiguration.cs
@@ -409,16 +409,16 @@ namespace NuGetGallery.Configuration
         /// <summary>
         /// The wait time in milliseconds for the WaitAndRetry policy.
         /// </summary>
-        int WaitAndRetryIntervalInMilliseconds { get; set; }
+        int SearchCircuitBreakerWaitAndRetryIntervalInMilliseconds { get; set; }
 
         /// <summary>
-        /// A request will fail after this number of retries.
+        /// A request will fail after this number of retries. In total a request will fail after this number of retries + 1.
         /// </summary>
-        int WaitAndRetryCount { get; set; }
+        int SearchCircuitBreakerWaitAndRetryCount { get; set; }
 
         /// <summary>
         /// CircuitBreaker will open after this number of consecutive failed requests.
         /// </summary>
-        int CircuitBreakerBreakAfterCount { get; set; }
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/Configuration/IAppConfiguration.cs
```diff
@@ -405,5 +405,20 @@ public interface IAppConfiguration : IMessageServiceConfiguration
...
         /// </summary>
         int SearchCircuitBreakerDelayInSeconds { get; set; }
+
+        /// <summary>
+        /// The wait time in milliseconds for the WaitAndRetry policy.
+        /// </summary>
+        int WaitAndRetryIntervalInMilliseconds { get; set; }
+
+        /// <summary>
+        /// A request will fail after this number of retries.
```
joelverhagen: Is this "attempts" or "retries"? Meaning, if the first attempt fails, will it try this many more times?

src/NuGetGallery/Web.config
```diff
@@ -138,6 +138,9 @@
     <add key="Gallery.SearchServiceUriPrimary" value=""/>
     <add key="Gallery.SearchServiceUriSecondary" value=""/>
     <add key="Gallery.SearchCircuitBreakerDelayInSeconds" value="600"/>
+    <add key="Gallery.WaitAndRetryIntervalInMilliseconds" value="500"/>
+    <add key="Gallery.WaitAndRetryCount" value="3"/>
```
joelverhagen: These config names should say something about "search"

skofman1: Agree. Preferably prefix with SearchCircuitBreaker

---
In reply to: [264359495](https://github.com/NuGet/NuGetGallery/pull/6975#discussion_r264359495) [](ancestors = 264359495)

src/NuGetGallery/Configuration/IAppConfiguration.cs
cristinamanum: Updated the comments. A request will be max executed retry + 1.

src/NuGetGallery/Web.config
cristinamanum: Updated. Prefix all with SearchCircuitBreaker.

PR #6994, opened 2019-03-13
Fix minor bugs in Deprecation UI found in bug bash

scottbommarito:
Some quick bugs found in https://github.com/nuget/engineering/issues/2168

- multi-select dropdown should not be larger than content

image

- "skip to content" does not tab to the deprecation warning on the package details page (non-focusable elements with `tabIndex` set should be considered `:focusable` and `:tabbable`)

- Fixed page when package is locked or feature flag is disabled

site admin:

image

owner:

image

- deprecation warning on package details page should show pointer cursor

- custom message in deprecation warning on package details page should have better text

- alternate package text in deprecation warning on package details page should be shorter

image

### DIFF
--- a/src/NuGetGallery/Views/Packages/_ManageDeprecation.cshtml
+++ b/src/NuGetGallery/Views/Packages/_ManageDeprecation.cshtml
@@ -1,26 +1,25 @@
 ﻿@model ManagePackageViewModel
 
 <div class="page-manage-deprecation">
-    <span>Deprecating a package will warn all consumers of the package who use it in their projects.</span>
-
     @if (!Model.VersionSelectList.Any())
     {
-        @ViewHelpers.AlertInfo(
-                @<text>
-                    <span>There are no versions that can be deprecated.</span>
-                </text>)
+        @ViewHelpers.AlertInfo(@<text>There are no versions that can be deprecated.</text>)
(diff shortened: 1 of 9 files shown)


src/NuGetGallery/Views/Packages/_ManageDeprecation.cshtml
```diff
@@ -1,27 +1,26 @@
...
-        @ViewHelpers.AlertWarning(@<text>You do not have permission to deprecate this package.</text>)
+        @ViewHelpers.AlertDanger(@<text>You do not have permission to deprecate this package.</text>)
+    }
+    else if (Model.IsLocked)
+    {
+        @ViewHelpers.AlertDanger(@<text>Locked packages cannot be deprecated.</text>)
     }
     else
     {
+        <span>Deprecating a package will warn all consumers of the package who use it in their projects.</span>
```
shishirx34: should this be attributed with `AlertInfo`

scottbommarito: No, it's meant to appear as plain text at the top of the section as a description. We could make it an `AlertInfo` but I'm not sure if it's necessary or if it would look better.

image

PR #6999, opened 2019-03-15
Sign the NuGet packages produced by the build

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/2244

### DIFF
--- a/.nuget/packages.config
+++ b/.nuget/packages.config
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="MicroBuild.Core" version="0.3.0" />
   <package id="Microsoft.CodeAnalysis.BinSkim" version="1.3.6" />
   <package id="xunit.runner.console" version="2.1.0" />
   <package id="xunit.runner.visualstudio" version="2.1.0" />
--- a/build.ps1
+++ b/build.ps1
@@ -10,7 +10,7 @@ param (
     [string]$PackageSuffix,
     [string]$Branch,
     [string]$CommitSHA,
-    [string]$BuildBranch = '5fd8377a9abf3ff411918dbb973948a6677432db'
+    [string]$BuildBranch = '2d8feecabe3aeaed7f5b4d50b9be78c94faf39ec'
 )
 
 Set-StrictMode -Version 1.0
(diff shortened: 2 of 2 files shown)


.nuget/packages.config
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="MicroBuild.Core" version="0.3.0" />
```
skofman1: where is this hosted? I didn't find it on nuget.org..
Found this one though: https://www.nuget.org/packages/Microsoft.VisualStudioEng.MicroBuild.Core/

scottbommarito: https://www.nuget.org/packages/microbuild.core/

We should migrate to the new package if it's not additional work.

joelverhagen: It's unlisted. We use this version everywhere:
https://www.nuget.org/packages/microbuild.core/0.3.0

joelverhagen: I can create a backlog item for moving if you want.

skofman1: will it be a problem to use the new one in Gallery? (since it's a new dependency)
Backlog item to fix the rest will be great.

scottbommarito: It's a new dependency in the `.nuget/packages.config`, but it's already included in all our projects (`NuGet.Services.Entities`, `NuGetGallery.Core`, `NuGetGallery`).

joelverhagen: https://github.com/NuGet/Engineering/issues/2278

joelverhagen: I'd rather be consistent now and handle any breaking changes all at once.

skofman1: Ok

build.ps1
```diff
@@ -99,6 +99,11 @@ Invoke-BuildStep 'Creating artifacts' { `
 } `
 -ev +BuildErrors
 
+Invoke-BuildStep 'Signing the packages' {
+    Sign-Packages -Configuration $Configuration -BuildNumber $BuildNumber -MSBuildVersion "15" `
+} `
+-ev +BuildErrors
```
cristinamanum: Just wondering - is this a async operation? Probably it will introduce some delay in the build time, as the local builds do not need to have the packages signed maybe for local builds can it be turned off .

joelverhagen: It is not asynchronous. The build does not end until all of the packages are signed. We already have this concept for code signing, by the way.

Local builds will not take any extra time since they don't have the MicroBuild plugins that enable signing. These plugins _do_ exist on our build agents. This build will basically take zero time for local builds since it actually does nothing.

PR #7017, opened 2019-03-26
Hotfix: FeatureFlagClient should return the defaultValue when user is null

scottbommarito:
`FeatureFlagClient` throws if the user is `null`.

Null user means "not logged in", so we should return the default value. This is the same logic the permissions service has.

(Note: https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs#L61 throws when `user` is `null`.)

### DIFF
--- a/src/NuGetGallery.Core/Features/FeatureFlagClientExtensions.cs
+++ b/src/NuGetGallery.Core/Features/FeatureFlagClientExtensions.cs
@@ -17,7 +17,9 @@ namespace NuGetGallery.Features
         {
             if (user == null)
             {
-                throw new ArgumentNullException(nameof(user));
+                // The current user is null when not logged in.
+                // Return the default value if the user is not logged in.
+                return defaultValue;
             }
 
             return client.IsEnabled(flight, new FlightUser(user), defaultValue);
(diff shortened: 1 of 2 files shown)


skofman1:
>         public static bool IsEnabled(

another solution can be adding a different method that doesn't take user as input and uses a different FeatureFlagClient API: [https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs#L22](https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs#L22)

---
Refers to: src/NuGetGallery.Core/Features/FeatureFlagClientExtensions.cs:12 in 0629d51. [](commit_id = 0629d51a07479207e13a0f53d1f59ed5421487f4, deletion_comment = True)

loic-sharma:
I talked offline to @scottbommarito, I don't think this is the best solution. If the flight is on for all accounts, the flight should return `true` for a `null` user even if the `defaultValue` is `false`. I think it'd be better if the flight client handles `null` values properly.

scottbommarito:
Closing this and pursuing another temporary option. Permanent fix is handled by https://github.com/NuGet/ServerCommon/pull/290

PR #7020, opened 2019-03-26
Hotfix: ManageDeprecation is not enabled for logged out users

scottbommarito:
Unfortunately, https://github.com/NuGet/NuGetGallery/pull/7016 throws when the user is null (e.g. logged out)

This is a temporary fix--https://github.com/NuGet/ServerCommon/pull/290 is the long-term fix

### DIFF
--- a/src/NuGetGallery/Services/FeatureFlagService.cs
+++ b/src/NuGetGallery/Services/FeatureFlagService.cs
@@ -45,11 +45,6 @@ namespace NuGetGallery
 
         public bool IsManageDeprecationEnabled(User user)
         {
-            if (user == null)
-            {
-                return false;
-            }
-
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Services/FeatureFlagService.cs
```diff
@@ -45,6 +45,11 @@ public bool IsPackagesAtomFeedEnabled()
 
         public bool IsManageDeprecationEnabled(User user)
         {
+            if (user == null)
```
skofman1: UTs!

scottbommarito: There are no UTs on this or any of the other `FeatureFlagService` methods.

scottbommarito: moved to `PackagesController.DisplayPackage` and added a unit test

Issue #7019, opened 2019-03-26
[Feature Flags] Handle anonymous users in flights


loic-sharma:
* [x] Update servercommon: https://github.com/NuGet/ServerCommon/pull/290
* [x] Create new packages https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2542231
* [x] Pin dependencies
* [x] Update ServerCommon dependencies in NuGetGallery

PR #7021, opened 2019-03-26
[Feature Flags] Handle anonymous users

loic-sharma:
* [X] Depends on https://github.com/NuGet/ServerCommon/pull/290
* [x] Build ServerCommon [#2534292](https://devdiv.visualstudio.com/DevDiv/_build%2Fresults?buildId=2534292&_a=summary)
* [x] Pin MyGet dependencies

See: https://github.com/NuGet/NuGetGallery/issues/7019

### DIFF
--- a/src/NuGetGallery.Core/Features/FeatureFlagClientExtensions.cs
+++ b/src/NuGetGallery.Core/Features/FeatureFlagClientExtensions.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using NuGet.Services.Entities;
 using NuGet.Services.FeatureFlags;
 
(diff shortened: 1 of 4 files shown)


src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -219,19 +219,19 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="NuGet.Services.Entities">
-      <Version>2.43.0</Version>
+      <Version>2.44.0-loshar-anonymous-flights-2530168</Version>
```
loic-sharma: TODO: Update this when https://github.com/NuGet/ServerCommon/pull/290 is merged

src/NuGetGallery.Core/Features/FeatureFlagClientExtensions.cs
```diff
@@ -15,12 +14,12 @@ public static class FeatureFlagClientExtensions
             User user,
             bool defaultValue)
         {
-            if (user == null)
-            {
-                throw new ArgumentNullException(nameof(user));
-            }
+            // The user object is null if the user isn't logged in.
+            var flightUser = (user != null)
+                ? new FlightUser(user)
```
joelverhagen: This could be simplified by making the ctor private and adding a static `.CreateOrNull` method. Seems like a null flight user is an expected state now so this could be more descriptive of this allowance.

joelverhagen: (nit)

loic-sharma: Talked offline. The class is a nested private class, so this method wouldn't add too much value.

Issue #7026, opened 2019-03-28
Delete account endpoints are slow


scottbommarito:
Both the page for requesting delete as a user and the page for deleting an account as a site admin take too long

Average on PROD for the last week: 
Delete Request - 3.960 seconds
Delete Account - 9.277 seconds

These endpoints should not take so long and it's annoying for the support request on-call.

scottbommarito:
Backlogging and unassigning because this was not as trivial as I thought.

PR #7028, opened 2019-03-28
Performance improvements for Delete User endpoints

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/7026

The changes here fit into several categories:
- Added a `PackageRegistrations` field on `User` so that package registrations owned by a user can easily be found instead of requiring a query. This did not require a migration because the associated index already exists for `PackageRegistration.Owners`.
- Used `RouteUrlTemplate`s on the delete page. (This could probably be done on some other pages in the table below for better performance.)
- Included more fields in `UserService.FindByKey` and `UserService.FindByUsername` requests so that less additional queries need to be made to determine organizations of the user requested.

### DIFF
--- a/src/NuGet.Services.Entities/User.cs
+++ b/src/NuGet.Services.Entities/User.cs
@@ -52,6 +52,7 @@ namespace NuGet.Services.Entities
             Roles = new List<Role>();
             Username = username;
             UserCertificates = new List<UserCertificate>();
+            PackageRegistrations = new List<PackageRegistration>();
         }
 
         /// <summary>
@@ -136,6 +137,8 @@ namespace NuGet.Services.Entities
         /// Gets or sets the collection of user certificates.
         /// </summary>
         public virtual ICollection<UserCertificate> UserCertificates { get; set; }
+
+        public virtual ICollection<PackageRegistration> PackageRegistrations { get; set; }
 
         public void ConfirmEmailAddress()
         {
--- a/src/NuGetGallery.Core/Entities/EntitiesContext.cs
+++ b/src/NuGetGallery.Core/Entities/EntitiesContext.cs
@@ -113,7 +113,7 @@ namespace NuGetGallery
                 .WillCascadeOnDelete(false);
 
             modelBuilder.Entity<Scope>()
-                .HasRequired<Credential>(sc => sc.Credential)
+                .HasRequired(sc => sc.Credential)
                 .WithMany(cr => cr.Scopes)
                 .HasForeignKey(sc => sc.CredentialKey)
                 .WillCascadeOnDelete(true);
@@ -133,12 +133,12 @@ namespace NuGetGallery
                 .HasKey(u => u.Key);
 
             modelBuilder.Entity<User>()
-                .HasMany<EmailMessage>(u => u.Messages)
+                .HasMany(u => u.Messages)
                 .WithRequired(em => em.ToUser)
                 .HasForeignKey(em => em.ToUserKey);
 
             modelBuilder.Entity<User>()
-                .HasMany<Role>(u => u.Roles)
+                .HasMany(u => u.Roles)
                 .WithMany(r => r.Users)
                 .Map(c => c.ToTable("UserRoles")
                            .MapLeftKey("UserKey")
@@ -194,7 +194,7 @@ namespace NuGetGallery
                 .HasKey(u => u.Key);
 
             modelBuilder.Entity<UserSecurityPolicy>()
-                .HasRequired<User>(p => p.User)
+                .HasRequired(p => p.User)
                 .WithMany(cr => cr.SecurityPolicies)
                 .HasForeignKey(p => p.UserKey)
                 .WillCascadeOnDelete(true);
@@ -203,14 +203,14 @@ namespace NuGetGallery
                 .HasKey(p => p.Key);
 
             modelBuilder.Entity<ReservedNamespace>()
-                .HasMany<PackageRegistration>(rn => rn.PackageRegistrations)
+                .HasMany(rn => rn.PackageRegistrations)
                 .WithMany(pr => pr.ReservedNamespaces)
                 .Map(prrn => prrn.ToTable("ReservedNamespaceRegistrations")
                                 .MapLeftKey("ReservedNamespaceKey")
                                 .MapRightKey("PackageRegistrationKey"));
 
             modelBuilder.Entity<ReservedNamespace>()
-                .HasMany<User>(pr => pr.Owners)
+                .HasMany(pr => pr.Owners)
                 .WithMany(u => u.ReservedNamespaces)
                 .Map(c => c.ToTable("ReservedNamespaceOwners")
                            .MapLeftKey("ReservedNamespaceKey")
@@ -223,7 +223,7 @@ namespace NuGetGallery
                 .HasKey(em => em.Key);
 
             modelBuilder.Entity<EmailMessage>()
-                .HasOptional<User>(em => em.FromUser)
+                .HasOptional(em => em.FromUser)
                 .WithMany()
                 .HasForeignKey(em => em.FromUserKey);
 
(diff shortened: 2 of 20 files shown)


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -257,16 +257,20 @@ public IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted,
...
         }
 
-        private IEnumerable<Package> GetPackagesForOwners(IEnumerable<int> ownerKeys, bool includeUnlisted, bool includeVersions)
+        private IEnumerable<Package> GetPackagesForOwners(IEnumerable<User> owners, bool includeUnlisted, bool includeVersions)
         {
-            IQueryable<Package> packages = _packageRepository.GetAll()
-                .Where(p => p.PackageRegistration.Owners.Any(o => ownerKeys.Contains(o.Key)));
+            var packageRegistrationKeys = owners
+                .SelectMany(o => o.PackageRegistrations)
+                .Select(pr => pr.Key);
```
shishirx34: get distinct here so that you can improve performance below by a little bit.

scottbommarito: Added! great catch

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -257,16 +257,20 @@ public IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted,
             bool includeUnlisted,
             bool includeVersions = false)
         {
-            var ownerKeys = user.Organizations.Select(org => org.OrganizationKey).ToList();
-            ownerKeys.Insert(0, user.Key);
-
-            return GetPackagesForOwners(ownerKeys, includeUnlisted, includeVersions);
+            var owners = new List<User> { user };
+            owners.AddRange(user.Organizations.Select(o => o.Organization));
```
skofman1: you dropped the ToList(). Intentional?

src/NuGetGallery/Services/UserService.cs
```diff
@@ -383,9 +382,18 @@ public virtual User FindByKey(int key, bool includeDeleted = false)
...
+            return IncludeUserFields(users)
+                .SingleOrDefault(u => u.Key == key);
+        }
+
+        private IQueryable<User> IncludeUserFields(IQueryable<User> users)
+        {
             return users.Include(u => u.Roles)
                 .Include(u => u.Credentials)
-                .SingleOrDefault(u => u.Key == key);
+                .Include(u => u.PackageRegistrations)
```
skofman1: A single user (like MSFT) can have thousands of registrations. This info is not needed in all scenarios. Add an option to have "slim" user vs. "full"

scottbommarito: Yes, but in my experiments with a local account with thousands of registrations, always loading these registrations was a performance improvement because it reduces the number of SQL calls in general. I think this makes sense, because most of the time, when we are loading a user individually, we are usually trying to fetch their packages or organizations as well. I will deploy to DEV to confirm.

src/NuGetGallery/Services/PackageService.cs
scottbommarito: `AddRange` materializes the `IEnumerable`, no need to do it twice.

src/NuGetGallery/Services/UserService.cs
skofman1: Not every scenario that loads user needs all good package registrations. For example, login flow. We should be cautious and make the load change only in tested scenarios.

src/NuGet.Services.Entities/User.cs
```diff
@@ -137,6 +138,8 @@ public string LastSavedEmailAddress
         /// </summary>
         public virtual ICollection<UserCertificate> UserCertificates { get; set; }
 
+        public virtual ICollection<PackageRegistration> PackageRegistrations { get; set; }
```
cristinamanum: Private set?

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -133,12 +133,12 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
...
 
             modelBuilder.Entity<User>()
-                .HasMany<EmailMessage>(u => u.Messages)
+                .HasMany(u => u.Messages)
                 .WithRequired(em => em.ToUser)
                 .HasForeignKey(em => em.ToUserKey);
 
             modelBuilder.Entity<User>()
-                .HasMany<Role>(u => u.Roles)
+                .HasMany(u => u.Roles)
```
cristinamanum: Is this change(and other same changes) necessary? Maybe add few comments.

src/NuGet.Services.Entities/User.cs
joelverhagen: Check `Add-Migration`. Does this add an index?

src/NuGetGallery/Services/UserService.cs
joelverhagen: I agree with @skofman1, unless we perf test all of the code paths... on PROD... on varying package registration counts... we should make a more targeted fix.

src/NuGetGallery/UrlHelperExtensions.cs
```diff
@@ -617,10 +616,28 @@ public static string GetOrganizationCertificates(this UrlHelper url, string acco
...
+        {
+            var routesGenerator = new Dictionary<string, Func<string, object>>
+            {
+                { "username", u => u }
+            };
+
+            Func<RouteValueDictionary, string> linkGenerator = rv => GetActionLink(
+                url,
+                "Profiles",
+                "Users",
```
joelverhagen: can these use `nameof`?

src/NuGetGallery/Views/Users/_UserOrganizationsListForDeletedAccount.cshtml
```diff
@@ -26,6 +26,9 @@
                     <tbody>
                         @foreach (var organization in Model)
                         {
+                            var usernameTemplate = Url.UsernameTemplate();
```
joelverhagen: can you explain why we separate template initialization from URL initialization below?

src/NuGetGallery.Core/Entities/EntitiesContext.cs
joelverhagen: These are changes VS has been suggesting for ages. The type parameter is implied by the type of `u.Roles`.

scottbommarito: Yeah, I was annoyed by it so i just did the fix. The only change to this file is the `HasMany(u => u.PackageRegistrations)`

src/NuGet.Services.Entities/User.cs
scottbommarito: I did test `Add-Migration`. The index already exists.

src/NuGetGallery/Views/Users/_UserOrganizationsListForDeletedAccount.cshtml
scottbommarito: I'm dumb--the templates need to be initialized outside of the for-loop so that they are reused. As a result of fixing this, delete page performance is now down to 900ms on average!

src/NuGet.Services.Entities/User.cs
scottbommarito: I was keeping the pattern with the rest of the file...not sure if EF requires a public `set` or not.

src/NuGetGallery/UrlHelperExtensions.cs
scottbommarito: the controller names can't but the rest can

src/NuGetGallery/Services/UserService.cs
scottbommarito: Deployed this to DEV and tested against the DEV DB and including the `PackageRegistrations` field was actually much slower--despite being faster locally. Removing the include.

scottbommarito:
Closing this. Was unable to get the performance gains on DEV that I experienced locally, and I don't want to spend too much more time on this issue as it is low priority. IMO, the way our services are structured, which heavily encourages lazy-loading, is the cause of this issue, and it is nontrivial to fix this issue, especially in pages like delete that require mass loading of many entities.

PR #7034, opened 2019-04-02
Only commit once in DeleteAccountService.DeleteAccountAsync

scottbommarito:
https://github.com/nuget/engineering/issues/2264

`DeleteAccountService.DeleteAccountAsync` was so slow because it calls a lot of other services, most of which internally called `EntitiesContext.SaveChangesAsync`. Because the entire method is wrapped in a transaction, none of these calls are necessary and thus slow down the process substantially. Each call to `EntitiesContext.SaveChangesAsync` forces EF to do a lot of work behind the scenes--even if the transaction isn't going to be committed to the DB yet.

This was fixed by
- passing `commitChanges: false` to services that already supported it
- modifying services so that they allow `commitChanges: false`

We had a couple services with a `commitAsTransaction` parameter, which committed regardless of whether or not it was in a transaction or not. I changed these so that it only committed if it was wrapped in a transaction, because if the caller doesn't want the commit in a transaction, it's likely they don't want the commit anyway.

### DIFF
--- a/src/NuGetGallery/Services/DeleteAccountService.cs
+++ b/src/NuGetGallery/Services/DeleteAccountService.cs
@@ -33,20 +33,20 @@ namespace NuGetGallery
         private readonly IAuditingService _auditingService;
         private readonly ITelemetryService _telemetryService;
 
-        public DeleteAccountService(IEntityRepository<AccountDelete> accountDeleteRepository,
-                                    IEntityRepository<User> userRepository,
-                                    IEntityRepository<Scope> scopeRepository,
-                                    IEntitiesContext entitiesContext,
-                                    IPackageService packageService,
-                                    IPackageOwnershipManagementService packageOwnershipManagementService,
-                                    IReservedNamespaceService reservedNamespaceService,
-                                    ISecurityPolicyService securityPolicyService,
-                                    AuthenticationService authService,
-                                    ISupportRequestService supportRequestService,
-                                    IEditableFeatureFlagStorageService featureFlagService,
-                                    IAuditingService auditingService,
-                                    ITelemetryService telemetryService
-            )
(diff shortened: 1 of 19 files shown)


src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -114,72 +115,79 @@ private async Task DeleteAccountImplAsync(User userToBeDeleted, User userToExecu
...
             };
+
             _accountDeleteRepository.InsertOnCommit(accountDelete);
-            await _accountDeleteRepository.CommitChangesAsync();
         }
 
         private async Task RemoveUserCredentials(User user)
         {
             // Remove any credential owned by this user.
-            foreach (var uc in user.Credentials.ToArray())
```
scottbommarito: Previously, we deleted credentials owned by the user, committed, and then queried for any remaining credentials scoped to the user. I noticed that when I removed the commit, the query below (`var userScopes = ...`) was now including some of the credentials owned by the user, and, as a result, failing. I had to rewrite this section so that it fetches all credentials to delete first and then deletes them at once.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -64,7 +65,6 @@ ITelemetryService telemetryService
 
         public async Task<DeleteUserAccountStatus> DeleteAccountAsync(User userToBeDeleted,
             User userToExecuteTheDelete,
-            bool commitAsTransaction,
```
cristinamanum: Why removing the parameter?

scottbommarito: It's unused and adds more complexity to the code.

PR #7041, opened 2019-04-03
Allow site admins to delete organizations they are not a part of

scottbommarito:
https://github.com/nuget/engineering/issues/2157

Most of the changes here consist of refactoring code that was in `UsersController` so that it can be reused by `OrganizationsController`.

# Link to delete as site admin is in the same place as it is for regular user accounts:
image

# Page looks like this:
image


karann-msft:
should it also expire any api keys where the organization is the owner?

scottbommarito:
@karann-msft - yes, and it already does this

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteOrganizationAccount.cshtml
```diff
@@ -0,0 +1,58 @@
...
+        <div class="form-group">
+            @ViewHelpers.Breadcrumb(true, @<text><a href="@Url.User(@Model.User)">@Model.AccountName</a></text>, @<text>Delete organization</text>)
+
+            @ViewHelpers.AlertDanger(@<text>
+                <b class="keywords">Once the organization is deleted, it cannot be undone!</b> <br />
+                Deleting the organization will: <br />
+                1. Remove the account as an owner of any child packages.<br />
+                2. Delete any API keys that are scoped to this organization.<br />
+                3. Dissociate all previously existent ID prefix reservations with this account.<br />
+                4. Remove all of its members.<br />
```
cristinamanum: Do we delete accounts that are members in the org or is only the membership link removed?

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteOrganizationAccount.cshtml
```diff
@@ -0,0 +1,58 @@
...
+            }
+        </div>
+    </div>
+</section>
+
+@section BottomScripts {
+    <script type="text/javascript">
+        $(function () {
+            $('#delete-form').submit(function (e) {
+                if (!confirm('Are you sure you want to continue to delete this account?')) {
```
cristinamanum: " Are you sure ... this account' or 'Are you sure ... this organization' ?

src/NuGetGallery/ViewModels/DeleteOrganizationViewModel.cs
```diff
@@ -15,10 +15,17 @@ public class DeleteOrganizationViewModel : DeleteAccountViewModel<Organization>
...
+                .ToList();
+
             AdditionalMembers = organizationToDelete.Members
                 .Where(m => !m.Member.MatchesUser(currentUser))
-                .Select(m => new OrganizationMemberViewModel(m));
+                .Select(m => new OrganizationMemberViewModel(m))
+                .ToList();
         }
+
+        public IEnumerable<OrganizationMemberViewModel> Members { get; set; }
```
loic-sharma: >set [](start = 71, length = 3)

Does this need a setter?

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteUserAccount.cshtml
```diff
@@ -26,7 +26,12 @@
                 @Html.Partial("_UserOrganizationsListForDeletedAccount", Model.Organizations)
             </div>
         </div>
-        @Html.Partial("_DeleteUserAccountForm", new DeleteAccountAsAdminViewModel(Model))
+        <div class="form-group danger-zone">
+            @using (Html.BeginForm("Delete", "Users", FormMethod.Post, new { id = "delete-form" }))
```
cristinamanum: What is the benefit of moving this part from the _DeleteUserAccountForm (below) here ?

src/NuGetGallery/Views/Users/Profiles.cshtml
```diff
@@ -46,9 +46,10 @@
                     {
                         <a href="@(Model.UserIsOrganization ? Url.ManageMyOrganization(Model.Username) : Url.AccountSettings())" aria-label="Manage this account"><i class="ms-Icon ms-Icon--Edit ms-font-xl"></i></a>
                     }
-                    @if (User.IsAdministrator() && !Model.UserIsOrganization)
+                    @if (User.IsAdministrator())
                     {
-                        <a href="@Url.AdminDeleteAccount(@Model.Username)" aria-label="Delete this account"><i class="ms-Icon ms-Icon--Delete ms-font-xl"></i></a>
+                        var deleteUrl = Model.UserIsOrganization ? Url.AdminDeleteOrganization(Model.Username) : Url.AdminDeleteAccount(Model.Username);
+                        <a href="@deleteUrl" aria-label="Delete this account"><i class="ms-Icon ms-Icon--Delete ms-font-xl"></i></a>
```
loic-sharma: >class [](start = 81, length = 5)

Do we need to put `aria-hidden="true"` on icons? Or is that no longer a requirement?

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteOrganizationAccount.cshtml
cristinamanum: What will happen with in flight membership requests?

scottbommarito: Yeah this is unclear...we only remove the membership link.

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteUserAccount.cshtml
scottbommarito: `DeleteUserAccount.cshtml` and `DeleteOrganizationAccount.cshtml` need to submit to different endpoints--`UsersController` handles deleting users and `OrganizationsController` handles deleting organizations. Rather than making changes to the model, I decided to move the `BeginForm` call outside the partial into the parent view.

src/NuGetGallery/ViewModels/DeleteOrganizationViewModel.cs
scottbommarito: Probably not, but I was keeping up with the pattern of this file and its base class. I'd rather not remove the setter for this field given that some MVC behavior depends on having setters.

src/NuGetGallery/Areas/Admin/Views/DeleteAccount/DeleteOrganizationAccount.cshtml
scottbommarito: They are deleted. Will add comment.

src/NuGetGallery/Views/Users/Profiles.cshtml
scottbommarito: added it

PR #7042, opened 2019-04-04
Add index to UserSecurityPolicy

xavierdecoster:
Adds a unique constraint to the `UserKey`, `Name` and `Subscription` columns of `UserSecurityPolicy`.


src/NuGet.Services.Entities/UserSecurityPolicy.cs
```diff
@@ -31,10 +32,11 @@ public UserSecurityPolicy(string name, string subscription, string value = null)
         /// Policy key.
         /// </summary>
         public int Key { get; set; }
-        
+
         /// <summary>
         /// User key.
         /// </summary>
+        [Index("IX_UserSecurityPolicy_UserKeyNameSubscription", IsUnique = true, Order = 0)]
```
loic-sharma: Most of our entities are configured using fluent in the `EntitiesContext`. Would it be difficult to move all the attributes over?

xavierdecoster: Changed this PR to use fluent api as well.

PR #7043, opened 2019-04-04
Reduce telemetry log count. 

cristinamanum:
It would reduce a daily average of 43,559,911.125 traces. Change tested on dev.


joelverhagen:
hotfix?

PR #7047, opened 2019-04-04
Fix error behavior for 400s

skofman1:
Current behavior: certain paths that cause HttpException / 400 result are redirected to 500 page.
With this fix they will get a page with "Bad Request" message. Will reduce log noise substantially.

### DIFF
--- a/src/NuGetGallery/Web.config
+++ b/src/NuGetGallery/Web.config
@@ -330,6 +330,7 @@
     <customErrors mode="RemoteOnly" defaultRedirect="~/Errors/500" redirectMode="ResponseRedirect">
       <!-- Adding ? at the end of the redirect URL prevents the illegal request to be passed
            as a query parameter to the redirect URL and causing additional failures -->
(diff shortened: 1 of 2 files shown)


joelverhagen:
Related to https://github.com/NuGet/NuGetGallery/issues/6601

src/NuGetGallery/Web.config
```diff
@@ -330,6 +330,7 @@
     <customErrors mode="RemoteOnly" defaultRedirect="~/Errors/500" redirectMode="ResponseRedirect">
       <!-- Adding ? at the end of the redirect URL prevents the illegal request to be passed
            as a query parameter to the redirect URL and causing additional failures -->
+      <error statusCode="400" redirect="~/Errors/400"/>
```
scottbommarito: Can you post a screenshot of the BadRequest page? I forgot what it looks like, and it hasn't been exposed since the redesign AFAIK.

skofman1:
I will merge to

PR #7052, opened 2019-04-05
UI for Feature Flags

scottbommarito:
https://github.com/NuGet/Engineering/issues/2334

This UI should make it much easier to edit feature flags without making mistakes.

I tried to keep the implementation as simple as I could (no AJAX, minimal knockout) while still making it functional.

# Full page

image

image

# Clicking "Update" for flights expands table row to allow editing

image

# Clicking "All" in flight update form hides the rest of the form

image

# Cannot add a user that doesn't exist to a flight

image


src/NuGetGallery/Areas/Admin/Views/Features/_EditFeature.cshtml
```diff
@@ -0,0 +1,34 @@
...
+        }
+    </td>
+    <td>
+        @using (Html.BeginForm("DeleteFeature", "Features", FormMethod.Post))
+        {
+            @Html.AntiForgeryToken()
+            @Html.Hidden("ContentId", Model.ContentId)
+            @Html.Hidden("Name", Model.Name)
+
+            <input id="Delete" type="submit" class="btn btn-danger" value="Delete" />
```
loic-sharma: Please add confirmation dialog on deletion actions

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -30,59 +35,330 @@ public class FeaturesController : AdminControllerBase
...
+            return MergeChangesAndTrySave(
+                flight,
+                GetDelete(
+                    flight,
+                    ValidateFlight,
+                    GetFlights,
+                    PrettyFlightName));
+        }
+
+        private delegate string ApplyChange(FeatureFlagsViewModel model);
```
loic-sharma: This controller is getting a little big. Can we more the core logic to services?

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -30,59 +35,330 @@ public class FeaturesController : AdminControllerBase
...
 
-                switch (result.Type)
-                {
-                    case FeatureFlagSaveResultType.Ok:
-                        // The flags have been persisted. Refresh this instance's cache immediately.
-                        await _cache.RefreshAsync();
+            var errorMessages = ModelState
+                .SelectMany(s => s.Value.Errors)
+                .Where(e => !string.IsNullOrWhiteSpace(e.ErrorMessage))
+                .Select(e => e.ErrorMessage);
```
loic-sharma: If you switch lines 291 and 290 you can slightly simplify the where to `.Where(e => !string.IsNullOrWhiteSpace(e));`

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -30,59 +35,330 @@ public class FeaturesController : AdminControllerBase
...
 
-            TimeSpan? timeSinceLastRefresh = null;
-            if (lastUpdated.HasValue)
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public Task<ActionResult> AddFeature(ModifyFeatureFlagsFeatureViewModel feature)
+        {
+            return MergeChangesAndTrySave(
+                feature,
+                GetAdd(
```
loic-sharma: This pattern of building up an `ApplyChange` using small components is really powerful, but it's also really hard to understand. Can we make this more straightforward?

loic-sharma: To understand how a feature is added we first have to understand what `MergeChangesAndTrySave` - this essentially calls some delegate returned by `GetAdd` on the inputted model. Next we have to understand what `GetAdd` does. To do that, we have to understand what `GetApplyChange` does. Okay, `GetApplyChange` returns the delegate that was called by `MergeChangesAndTrySave`.  That delegate first calls `validateChanges`, which was first passed by `GetAdd`, which was set to `ValidateFeature` by `AddFeature`. It turns out `ValidateFeature` doesn't do anything.

... I have to repeat this process for all arguments of `MergeChangesAndTrySave` for every caller of `MergeChangesAndTrySave`.

loic-sharma: My recommendation:

1. I would avoid passing delegates around as this results in high cognitive load
2. I would avoid merging the features and flights code paths as generics are another source of complexity

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -30,59 +35,330 @@ public class FeaturesController : AdminControllerBase
...
+                : "The model submitted was invalid.";
+        }
 
-                        TempData["Message"] = $"Your feature flags have been saved! It may take up to {refreshSeconds} seconds for this change to propagate everywhere.";
-                        return Redirect(Url.Action(actionName: "Index", controllerName: "Features"));
+        private async Task<string> TrySaveFlags(FeatureFlagsViewModel model, string contentId)
+        {
+            var flags = new FeatureFlags(
+                model.Features.ToDictionary(f => f.Name, f => f.Status),
+                model.Flights.ToDictionary(f => f.Name, GetFlightFromModel));
```
loic-sharma: Thoughts on adding a `AsFlight` method to the model?

src/NuGetGallery/Areas/Admin/ViewModels/FeatureFlagsFlightViewModel.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        [Required]
+        public string Name { get; set; }
+
+        public bool All { get; set; }
+
+        [Display(Name = "Site Admins")]
+        public bool SiteAdmins { get; set; }
+
+        public IEnumerable<string> Accounts { get; set; }
```
loic-sharma: Can these be `IReadOnlyList`?

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
scottbommarito: nice catch

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
scottbommarito: I'm going to move the delegates away from function parameters and put them as methods in the model and see how if it improves things.

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
scottbommarito: Sounds good.

src/NuGetGallery/Areas/Admin/ViewModels/FeatureFlagsFlightViewModel.cs
scottbommarito: Probably not--MVC doesn't like `IReadOnlyList` and refuses to correctly handle them when they're in model.

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
scottbommarito: Finished! Looks great except I wasn't able to get rid of the generics on a couple methods but there's no more delegates anywhere.

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
scottbommarito: It's smaller now with the delegate changes. I don't think it needs to be split up now.

src/NuGetGallery/Areas/Admin/Views/Features/_EditFeature.cshtml
scottbommarito: Sounds good.

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -30,59 +35,219 @@ public class FeaturesController : AdminControllerBase
...
+
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(type));
+            }
         }
 
-        [HttpPost]
-        [ValidateAntiForgeryToken]
-        public async Task<ActionResult> Index(FeatureFlagsViewModel model)
+        private async Task<ActionResult> MergeChangesAndTrySave<TModify, TBase>(
```
loic-sharma: >MergeChangesAndTrySave [](start = 41, length = 22)

Nit: please place this method before `ApplyChange` to follow call ordering

src/NuGetGallery/Areas/Admin/ViewModels/IFeatureFlagsViewModel.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Areas.Admin.ViewModels
+{
+    public interface IFeatureFlagsViewModel
```
loic-sharma: >IFeatureFlagsViewModel [](start = 21, length = 22)

Super nit: `FeatureFlagsViewModel` does not implement `IFeatureFlagsViewModel`. This is a little confusing, consider renaming this type. Maybe `IFeatureFlagsObjectViewModel`?

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
```diff
@@ -30,59 +35,219 @@ public class FeaturesController : AdminControllerBase
...
+            TModify change,
+            ChangeType type)
+            where TModify : IModifyFeatureFlagsViewModel<TBase>, TBase
+            where TBase : IFeatureFlagsViewModel
+        {
+            var validationError = change.GetValidationError(_userService);
+            if (validationError != null)
             {
-                timeSinceLastRefresh = DateTimeOffset.UtcNow.Subtract(lastUpdated.Value);
+                return validationError;
```
loic-sharma: The user's inputs aren't saved when validation errors happen. Is this hard to fix?

src/NuGetGallery.Core/Features/FeatureFlagSaveResult.cs
```diff
@@ -8,7 +8,7 @@ namespace NuGetGallery.Features
     /// </summary>
     public class FeatureFlagSaveResult
     {
-        private FeatureFlagSaveResult(FeatureFlagSaveResultType type, string message = null)
+        public FeatureFlagSaveResult(FeatureFlagSaveResultType type, string message = null)
```
loic-sharma: The `message` is unused now, right? Can we remove the `FeatureFlagSaveResult` type and use just `FeatureFlagSaveResultType`? If so, I would rename `FeatureFlagSaveResultType` to `FeatureFlagSaveResult`

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
scottbommarito: Unfortunately, it is nontrivial to fix. There's no concept of a "pending" change right now so I would have to add it. I agree that it's not ideal, but I don't think the time investment warrants it.

src/NuGetGallery/Areas/Admin/ViewModels/IFeatureFlagsViewModel.cs
scottbommarito: Yeah, I wasn't sure what the best term for "a thing that is stored in the feature flags" is. I think "object" works fine so I'll rename it to that.

src/NuGetGallery.Core/Features/FeatureFlagSaveResult.cs
scottbommarito: Good call.

src/NuGetGallery/Areas/Admin/Controllers/FeaturesController.cs
scottbommarito: done!

src/NuGetGallery/Areas/Admin/Views/Features/_EditFlight.cshtml
```diff
@@ -0,0 +1,81 @@
...
+                    <div class="flight-array col-sm-3" name="Accounts" data-items="@string.Join(";", Model.Accounts ?? new string[0])" data-bind="template: { name: 'flight-array', data: $data }"></div>
+                    <div class="flight-array col-sm-3" name="Domains" data-items="@string.Join(";", Model.Domains ?? new string[0])" data-bind="template: { name: 'flight-array', data: $data }"></div>
+                </div>
+
+                <div class="col-sm-2 @(Model.All ? "col-sm-offset-7" : "")" id="@rowFormSubmitId">
+                    <div class="form-group">
+                        <input id="Update" type="submit" class="btn btn-primary form-control" value="Update" />
+                    </div>
+                    <div class="form-group">
+                        <button type="button" class="btn btn-default form-control" onclick="$('.@rowClass').show(); $('#@rowFormId').hide()">Cancel</button>
```
joelverhagen: >button [](start = 25, length = 6)

Does cancel reset the changes you've made? If not, it should probably refresh the page to do this.

src/NuGetGallery/Areas/Admin/ViewModels/FeatureFlagsFeatureViewModel.cs
```diff
@@ -0,0 +1,35 @@
...
+        public FeatureFlagsFeatureViewModel()
+        {
+        }
+
+        public FeatureFlagsFeatureViewModel(
+            FeatureFlagsFeatureViewModel feature)
+        {
+            Name = feature.Name;
+            Status = feature.Status;
+        }
```
skofman1: where is this ctor used?

src/NuGetGallery/Areas/Admin/Views/Web.config
```diff
@@ -10,13 +10,16 @@
 
   <system.web.webPages.razor>
     <host factoryType="System.Web.Mvc.MvcWebRazorHostFactory, System.Web.Mvc, Version=5.2.3.0, Culture=neutral, PublicKeyToken=31BF3856AD364E35" />
-    <pages pageBaseType="System.Web.Mvc.WebViewPage">
+    <pages pageBaseType="NuGetGallery.Views.NuGetViewBase">
```
skofman1: what's the impact of this change?

src/NuGetGallery/Branding/Views/web.config
```diff
@@ -18,8 +18,9 @@
         <add namespace="System.Web.Mvc.Html" />
         <add namespace="System.Web.Routing" />
         <add namespace="System.Web.Optimization" />
-        <add namespace="NuGetGallery"/>
-        <add namespace="NuGetGallery.Helpers"/>
+        <add namespace="NuGetGallery" />
+        <add namespace="NuGetGallery.Helpers" />
+        <add namespace="NuGet.Services.Entities" />
```
skofman1: what's the impact of this change?

src/NuGetGallery/Views/web.config
```diff
@@ -16,11 +16,11 @@
...
         <add namespace="System.Web.Mvc.Html" />
-        <add namespace="System.Web.Routing" />
         <add namespace="System.Web.Optimization" />
-        <add namespace="NuGetGallery"/>
-        <add namespace="NuGetGallery.Helpers"/>
-        <add namespace="NuGet.Services.Entities"/>
+        <add namespace="System.Web.Routing" />
+        <add namespace="NuGetGallery" />
+        <add namespace="NuGetGallery.Helpers" />
+        <add namespace="NuGet.Services.Entities" />
```
skofman1: nothing changed here..

src/NuGetGallery/Areas/Admin/Views/Features/_EditFlight.cshtml
scottbommarito: Right now it changes the edit view back to the view with the unedited changes, so no refresh is necessary--it already shows the unedited content.

src/NuGetGallery/Areas/Admin/Views/Web.config
scottbommarito: Oh I forgot to mention why the `web.config` files were changed, but basically:

[In the `web.config` for all views in the normal area of the site, we set `NuGetGallery.Views.NuGetViewBase` to be the base of the view](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Views/web.config#L13), just like I am doing here. Doing this does a couple things, but specifically why I did it as part of this PR was that [it configures the `ViewHelpers.Section` method](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Views/NuGetViewBase.cs#L64) that we use all over the Gallery and also on this feature flag UI.

Apparently, as you can see in this diff, the `web.config` for the views in the admin area of the site (and also the branding area) were NOT setting the view base as expected. I noticed this because when I added calls to `ViewHelpers.Section` in the feature flag UI, it didn't work. As part of this PR, so I could use `ViewHelpers.Section` in the feature flag UI, I fixed this issue.

src/NuGetGallery/Branding/Views/web.config
scottbommarito: Views in the branding section of the site (we currently have none) will not need to include `NuGet.Services.Entities` when they use types from them. This was done to align the section with the rest of the site.

src/NuGetGallery/Views/web.config
scottbommarito: Alphabetized and aligned with `" />` instead of `"/>

PR #7055, opened 2019-04-08
DeleteAccountService should delete the row in the Users table

scottbommarito:
Part of https://github.com/NuGet/Engineering/issues/2358

Currently, when the `DeleteAccountService` deletes a user, it keeps the user's row in the `Users` table but scrubs of the data associated with that user. This has proven to be error-prone, because if DB relationships are not properly cleaned up, the `DeleteAccountService` does not fail, and these relationships still silently exist and may cause issues.

With this change, when a user is deleted, its row in the `Users` table is deleted and the username is persisted in the `AccountDelete`s table instead. When creating a new user or organization, this table is checked for username conflicts as well. Similarly to before, for unconfirmed accounts, we will not create a row in the `AccountDelete` table so that the username is not reserved.

I believe this is a positive change because
- if there is a bug with the `DeleteAccountService`, instead of silently failing and not cleaning up the deleted user data, it now will fail so that we will know we need to fix a problem
- the code is less complex, as there is no longer a `User.IsDeleted` field that we need to check everywhere or an `includeDeleted` parameter on `IUserService.FindByUsername` and `IUserService.FindByKey`--the only places that need to be aware of deleted account usernames are the register and add organization flows

I will follow this PR with some additional PRs that fix issues in the `DeleteAccountService`.


src/NuGetGallery/Migrations/201904081336296_DeletingAccountShouldDeleteUser.cs
```diff
@@ -0,0 +1,61 @@
...
+            Sql(@"
+DELETE o 
+FROM Organizations o
+JOIN Users u ON o.[Key] = u.[Key]
+WHERE u.IsDeleted = 1
+
+DELETE FROM Users
+WHERE IsDeleted = 1");
+
+            DropColumn("dbo.Users", "IsDeleted");
```
loic-sharma: How will we deploy this? If you run this migration you'll break the existing Gallery. If you deploy the new code without this migration, you'll break the new Gallery deployment. What's the plan?

Also do we really need to drop these columns? Normally our migrations don't have breaking changes and can be deployed with no code changes.

src/NuGetGallery/Migrations/201904081336296_DeletingAccountShouldDeleteUser.cs
```diff
@@ -0,0 +1,61 @@
...
+            DropColumn("dbo.AccountDeletes", "DeletedAccountKey");
+        }
+        
+        public override void Down()
+        {
+            AddColumn("dbo.AccountDeletes", "DeletedAccountKey", c => c.Int(nullable: false));
```
loic-sharma: We need to be prepared to rollback the Gallery. We should be able to repopulate the tables with deleted accounts

skofman1:
@scottbommarito , since this is a non-trivial change, please write a short dev spec for this and https://github.com/NuGet/Engineering/issues/2357
In the spec also call out the rollout plan for the new DB schema. Since you are working remotely this week, the team can review the spec and send comments by mail.

src/NuGetGallery/Migrations/201904081336296_DeletingAccountShouldDeleteUser.cs
scottbommarito: Good catch.

IMO we should drop these columns--we shouldn't have columns in the DB that don't serve any purpose--but it was incorrect to both migrate the data into new columns and remove the old columns in the same migration.

I think it'd be better if we split the migration into two, with one where we add the new columns and another where we delete the old columns. This will allow us to deploy the Gallery first in a way that does not break existing behavior in case we rollback, and then, after the new behavior is verified, drop the old rows.

I will flesh out this idea more in the spec.

src/NuGetGallery/Migrations/201904081336296_DeletingAccountShouldDeleteUser.cs
scottbommarito: Sounds reasonable. I think, ideally, the migration would insert new User rows with the username for every `AccountDelete` row, and then link the `AccountDelete` row to it. I don't think it's possible for us to completely re-inflate all information that was deleted (for example various rows that were deleted through a cascade) but it should definitely revert to a state that works with the previous implementation.

scottbommarito:
Abandoning this in favor of a lower cost alternative.

PR #7058, opened 2019-04-10
Add auditing to feature flags

scottbommarito:
Gallery side portion of https://github.com/nuget/engineering/issues/2333

### DIFF
--- a/src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFeature.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFeature.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using NuGet.Services.FeatureFlags;
+
+namespace NuGetGallery.Auditing.AuditedEntities
+{
+    public class AuditedFeatureFlagFeature
+    {
+        public string Name { get; private set; }
+        public FeatureStatus Status { get; private set; }
+
+        public static AuditedFeatureFlagFeature[] CreateFrom(FeatureFlags flags)
+        {
+            return flags.Features?
+                .Select(f => CreateFrom(f.Key, f.Value))
+                .ToArray() ?? new AuditedFeatureFlagFeature[0];
+        }
+
+        public static AuditedFeatureFlagFeature CreateFrom(string name, FeatureStatus status)
+        {
+            return new AuditedFeatureFlagFeature
+            {
+                Name = name,
+                Status = status
+            };
+        }
+    }
+}
--- a/src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using NuGet.Services.FeatureFlags;
+
+namespace NuGetGallery.Auditing.AuditedEntities
+{
+    public class AuditedFeatureFlagFlight
+    {
+        public string Name { get; private set; }
+        public bool All { get; private set; }
+        public bool SiteAdmins { get; private set; }
+        public string[] Accounts { get; private set; }
+        public string[] Domains { get; private set; }
+
+        public static AuditedFeatureFlagFlight[] CreateFrom(FeatureFlags flags)
+        {
+            return flags.Flights?
+                .Select(f => CreateFrom(f.Key, f.Value))
+                .ToArray() ?? new AuditedFeatureFlagFlight[0];
+        }
+
+        public static AuditedFeatureFlagFlight CreateFrom(string name, Flight flight)
+        {
+            return new AuditedFeatureFlagFlight
+            {
+                Name = name,
+                All = flight.All,
+                SiteAdmins = flight.SiteAdmins,
+                Accounts = flight.Accounts?.ToArray() ?? new string[0],
+                Domains = flight.Domains?.ToArray() ?? new string[0]
+            };
+        }
+    }
+}
--- a/src/NuGetGallery.Core/Auditing/FeatureFlagsAuditRecord.cs
+++ b/src/NuGetGallery.Core/Auditing/FeatureFlagsAuditRecord.cs
@@ -1,0 +1,48 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using NuGet.Services.FeatureFlags;
+using NuGetGallery.Auditing.AuditedEntities;
+using NuGetGallery.Features;
+
+namespace NuGetGallery.Auditing
+{
+    public sealed class FeatureFlagsAuditRecord : AuditRecord<AuditedFeatureFlagsAction>
+    {
+        public AuditedFeatureFlagFeature[] Features { get; }
+        public AuditedFeatureFlagFlight[] Flights { get; }
+        public string ContentId { get; }
+        public FeatureFlagSaveResult Result { get; }
+
+        public FeatureFlagsAuditRecord(
+            AuditedFeatureFlagsAction action, 
+            FeatureFlags flags,
+            string contentId, 
+            FeatureFlagSaveResult result)
+            : base(action)
+        {
+            if (flags == null)
+            {
+                throw new ArgumentNullException(nameof(flags));
+            }
+
+            ContentId = contentId ?? throw new ArgumentNullException(nameof(contentId));
+
+            Features = AuditedFeatureFlagFeature.CreateFrom(flags);
+            Flights = AuditedFeatureFlagFlight.CreateFrom(flags);
+            
+            Result = result;
+
+            // Group feature flags changes by the month they occurred in.
+            var currentTime = DateTime.UtcNow;
+            _path = $"{currentTime.Year}-{currentTime.Month}";
+        }
+
+        private readonly string _path;
+        public override string GetPath()
+        {
+            return _path;
+        }
+    }
+}
--- a/src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
+++ b/src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
@@ -10,9 +10,9 @@ using Microsoft.Extensions.Logging;
 using Microsoft.WindowsAzure.Storage;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
-using Newtonsoft.Json.Linq;
 using NuGet.Services.Entities;
 using NuGet.Services.FeatureFlags;
+using NuGetGallery.Auditing;
 
 namespace NuGetGallery.Features
 {
@@ -23,6 +23,7 @@ namespace NuGetGallery.Features
         private static readonly JsonSerializer Serializer;
 
         private readonly ICoreFileStorageService _storage;
+        private readonly IAuditingService _auditing;
         private readonly ILogger<FeatureFlagFileStorageService> _logger;
 
         static FeatureFlagFileStorageService()
@@ -40,9 +41,11 @@ namespace NuGetGallery.Features
 
         public FeatureFlagFileStorageService(
             ICoreFileStorageService storage,
+            IAuditingService auditing,
             ILogger<FeatureFlagFileStorageService> logger)
         {
             _storage = storage ?? throw new ArgumentNullException(nameof(storage));
+            _auditing = auditing ?? throw new ArgumentNullException(nameof(auditing));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
@@ -118,6 +121,19 @@ namespace NuGetGallery.Features
         }
 
         public async Task<FeatureFlagSaveResult> TrySaveAsync(FeatureFlags flags, string contentId)
+        {
+            var result = await TrySaveInternalAsync(flags, contentId);
+            await _auditing.SaveAuditRecordAsync(
+                new FeatureFlagsAuditRecord(
+                    AuditedFeatureFlagsAction.Update, 
+                    flags, 
+                    contentId, 
+                    result));
+
+            return result;
+        }
+
+        private async Task<FeatureFlagSaveResult> TrySaveInternalAsync(FeatureFlags flags, string contentId)
         {
             var accessCondition = AccessConditionWrapper.GenerateIfMatchCondition(contentId);
 
--- a/src/NuGetGallery.Core/Auditing/AuditedFeatureFlagsAction.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditedFeatureFlagsAction.cs
@@ -1,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Auditing
+{
+    public enum AuditedFeatureFlagsAction
+    {
+        Update
+    }
+}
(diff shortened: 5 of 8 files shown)


src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
```diff
@@ -0,0 +1,36 @@
...
+using NuGet.Services.FeatureFlags;
+
+namespace NuGetGallery.Auditing.AuditedEntities
+{
+    public class AuditedFeatureFlagFlight
+    {
+        public string Name { get; private set; }
+        public bool All { get; private set; }
+        public bool SiteAdmins { get; private set; }
+        public int AccountsCount { get; private set; }
```
scottbommarito: Note that I'm auditing the number of accounts and domains in the flight rather than the actual values--I didn't want to put extra effort into making the account/domain names GDPR compliant.

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFeature.cs
```diff
@@ -0,0 +1,29 @@
...
+        public string Name { get; private set; }
+        public FeatureStatus Status { get; private set; }
+
+        public static AuditedFeatureFlagFeature[] CreateFrom(FeatureFlags flags)
+        {
+            return flags.Features
+                .Select(f => CreateFrom(f.Key, f.Value))
+                .ToArray();
+        }
+        public static AuditedFeatureFlagFeature CreateFrom(string name, FeatureStatus status)
```
loic-sharma: Please add a new line above here

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
loic-sharma: @cristinamanum We can put PII in auditing data as it's ephemeral, right?

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
```diff
@@ -118,6 +121,19 @@ public async Task RemoveUserAsync(User user)
         }
 
         public async Task<FeatureFlagSaveResult> TrySaveAsync(FeatureFlags flags, string contentId)
+        {
+            var result = await TrySaveInternalAsync(flags, contentId);
+            await _auditing.SaveAuditRecordAsync(
```
loic-sharma: If there's an exception in between auditing and saving the flags, what data would we rather lose? It seems to me that we should guarantee that all update attempts are audited.

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
scottbommarito: I noticed that some of the other auditing classes have `Obfuscate` attributes on them. My problem with having an array of usernames in this audit record is that this obfuscation attribute doesn't currently support arrays.

src/NuGetGallery.Core/Auditing/FeatureFlagsAuditRecord.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            if (flags == null)
+            {
+                throw new ArgumentNullException(nameof(flags));
+            }
+
+            Features = AuditedFeatureFlagFeature.CreateFrom(flags);
+            Flights = AuditedFeatureFlagFlight.CreateFrom(flags);
+
+            ContentId = contentId ?? throw new ArgumentNullException(nameof(contentId));
```
skofman1: move this validation to the beginning of the ctor (or right after flags check)

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFeature.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace NuGetGallery.Auditing.AuditedEntities
+{
+    public class AuditedFeatureFlagFeature
+    {
+        public string Name { get; private set; }
+        public FeatureStatus Status { get; private set; }
+
+        public static AuditedFeatureFlagFeature[] CreateFrom(FeatureFlags flags)
+        {
+            return flags.Features
```
skofman1: >Features [](start = 25, length = 8)

Features can be null

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
skofman1: yes

---
In reply to: [274024465](https://github.com/NuGet/NuGetGallery/pull/7058#discussion_r274024465) [](ancestors = 274024465)

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
```diff
@@ -0,0 +1,36 @@
...
+        public int DomainsCount { get; private set; }
+
+        public static AuditedFeatureFlagFlight[] CreateFrom(FeatureFlags flags)
+        {
+            return flags.Flights
```
skofman1: this can be null

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
skofman1: it's ok to put usernames in audit, as long as we don't audit to blob storage but only to IFxAudit

---
In reply to: [273941956](https://github.com/NuGet/NuGetGallery/pull/7058#discussion_r273941956) [](ancestors = 273941956)

src/NuGetGallery.Core/Auditing/FeatureFlagsAuditRecord.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace NuGetGallery.Auditing
+{
+    public sealed class FeatureFlagsAuditRecord : AuditRecord<AuditedFeatureFlagsAction>
+    {
+        public AuditedFeatureFlagFeature[] Features { get; }
+        public AuditedFeatureFlagFlight[] Flights { get; }
+        public string ContentId { get; }
+        public FeatureFlagSaveResult Result { get; }
+
+        public FeatureFlagsAuditRecord(
```
skofman1: we also want to know who was the user that performed the update

scottbommarito: I was under the impression that the "audit actor", which is automatically injected into the audit log, contains this information. I will double check and add it if not.

scottbommarito: I was under the impression that the "audit actor", which is automatically injected into the audit log, contains this information. I will double check and add it if not.

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
scottbommarito: Will add a new obfuscation type (`UserNameList`) that obfuscates as `new [] { Obfuscator.ObfuscatedUserName }` and then include the full list of usernames in the audit record.

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
scottbommarito: I'm not sure. AI will already have record of these attempts and the exceptions they throw. Many bad requests will fail before we even get to save the flags--should we log those too? I was under the impression that these auditing changes were more for seeing the change history and the user that made the change. I would argue that failed changes are much less important for this.

Furthermore, I looked across the gallery. It seems we're very inconsistent regarding whether or not to audit the change before or after the actual change occurs, but I don't see any cases in which we try-catch-finally and always audit the change.

src/NuGetGallery.Core/Auditing/FeatureFlagsAuditRecord.cs
scottbommarito: In the JSON blob, this is the case. I will double-check on IFXAudit before merging.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
```diff
@@ -41,10 +44,30 @@ protected virtual PropertyInfo ResolvePath(string path)
             var currentInstance = Instance;
             foreach (var part in path.Split('.'))
             {
-                property = currentInstance.GetType().GetProperty(part);
-                if (property == null) { return null; }
+                var propertyName = part;
```
scottbommarito: Obfuscation was not being done for fields of objects that were in an array.

For example:
```
{
    hello: [
        {
            "someField": "this is not being obfuscated correctly!"
        }
    ]
}
```

This change fixes it, but ultimately I think this entire obfuscation converter should be rewritten in a way that doesn't require resolving the path of the object being serialized. There are a lot of scenarios here that are untested--dictionaries, arrays with multiple indices, etc.

src/NuGetGallery.Core/Auditing/FeatureFlagsAuditRecord.cs
```diff
@@ -0,0 +1,48 @@
...
+            ContentId = contentId ?? throw new ArgumentNullException(nameof(contentId));
+
+            Features = AuditedFeatureFlagFeature.CreateFrom(flags);
+            Flights = AuditedFeatureFlagFlight.CreateFrom(flags);
+            
+            Result = result;
+
+            // Group feature flags changes by the month they occurred in.
+            var currentTime = DateTime.UtcNow;
+            _path = $"{currentTime.Year}-{currentTime.Month}";
```
skofman1: retention policy for IFxAudit is 1 week, so this will not be useful.  You might as well use a const value.

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
```diff
@@ -0,0 +1,40 @@
...
+using NuGetGallery.Auditing.Obfuscation;
+
+namespace NuGetGallery.Auditing.AuditedEntities
+{
+    public class AuditedFeatureFlagFlight
+    {
+        public string Name { get; private set; }
+        public bool All { get; private set; }
+        public bool SiteAdmins { get; private set; }
+        [Obfuscate(ObfuscationType.UserNameList)]
```
skofman1: >i [](start = 27, length = 1)

I'm confused: this audit should go ONLY to IfxAudit. And there we don't need to obfuscate. So why is this needed?

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
cristinamanum: @loic-sharma and @skofman1 are correct, we do not need to obfuscate the data that goes audit logs.

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
cristinamanum: We need to obfuscate only if this data is persisted to blob storage or anything else than ifx audit.

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public static AuditedFeatureFlagFlight CreateFrom(string name, Flight flight)
+        {
+            return new AuditedFeatureFlagFlight
+            {
+                Name = name,
+                All = flight.All,
+                SiteAdmins = flight.SiteAdmins,
+                Accounts = flight.Accounts?.ToArray() ?? new string[0],
```
cristinamanum: This could mean a large number of accounts. Maybe is worth a manual testing for how it behaves for large number of accounts.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscationType.cs
```diff
@@ -8,6 +8,7 @@ public enum ObfuscationType
         UserKey,
         UserName,
         Authors,
-        IP
+        IP,
+        UserNameList
```
cristinamanum: If the logs go only in IFXAudit these changes are not needed right?

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
scottbommarito: There are three implementations of the auditing service--file, cloud, and IFX. Each implementation decides whether or not to obfuscate the data, and only cloud does it today. There is no mechanism for saying "this should only be audited if we are doing IFX audit". As a result, we need to make sure it has the proper obfuscation attribute in case we are not using the IFX audit implementation.

As a side note, cloud is the only auditing service implementation that processes the obfuscation attribute--the implementation that writes to the local file system does not obfuscate. As it is meant to be a debugging tool it seems fine to me but I thought it was worth mentioning.

src/NuGetGallery.Core/Features/FeatureFlagFileStorageService.cs
cristinamanum: I think that the TrySaveInternalAsync exception data is more of diagnostics data rather than Audit data.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscationType.cs
scottbommarito: There is no mechanism today to prevent them from going to another source other than IFX audit. I would rather not break the pattern of the existing code--which is to use the `Obfuscate` attribute on any audit property that should be obfuscated.

src/NuGetGallery.Core/Auditing/FeatureFlagsAuditRecord.cs
scottbommarito: I believe this is mostly used by the local file system auditing service. It uses this string to determine where to save the audit record file.

scottbommarito:
@cristinamanum - yup, shims changes coming soon

scottbommarito:
@skofman1 @cristinamanum - it seems like I was wrong and that there is a mechanism that prevents any audit logs other than package deletion audit records from going to blob storage, so I can go ahead and remove all obfuscation changes from this PR

src/NuGetGallery.Core/Auditing/FeatureFlagsAuditRecord.cs
scottbommarito: It is the case on IFXAudit.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscationType.cs
scottbommarito: Determined that I was wrong.

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
scottbommarito: This log will not go to blob storage, so no obfuscation is needed.

src/NuGetGallery.Core/Auditing/Obfuscation/ObfuscatorJsonConverter.cs
scottbommarito: no longer necessary

scottbommarito:
Shims changes at https://nuget.visualstudio.com/NuGetMicrosoft/_git/NuGet.Shims/pullrequest/899?_a=files

src/NuGetGallery.Core/Auditing/AuditedEntities/AuditedFeatureFlagFlight.cs
scottbommarito: no need to obfuscate. putting full names in audit logs

PR #7059, opened 2019-04-10
DeleteAccountService should delete the user and then recreate it

scottbommarito:
https://github.com/nuget/engineering/issues/2358

This guarantees that the deleted user is completely purged from the DB while not requiring any schema changes.

### DIFF
--- a/src/NuGetGallery/Services/DeleteAccountService.cs
+++ b/src/NuGetGallery/Services/DeleteAccountService.cs
@@ -116,21 +116,7 @@ namespace NuGetGallery
                 RemoveMembers(organizationToBeDeleted);
             }
 
-            if (!userToBeDeleted.Confirmed)
-            {
-                // Unconfirmed users should be hard-deleted.
-                // Another account with the same username can be created.
-                RemoveUser(userToBeDeleted);
-            }
-            else
-            {
-                // Confirmed users should be soft-deleted.
-                // Another account with the same username cannot be created.
-                RemoveUserDataInUserTable(userToBeDeleted);
-                InsertDeleteAccount(
-                    userToBeDeleted, 
-                    userToExecuteTheDelete);
-            }
+            RemoveUser(userToBeDeleted, userToExecuteTheDelete);
 
             if (commitChanges)
             {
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -285,19 +271,33 @@ private void RemoveMembers(Organization organization)
...
+        private void RemoveUser(User userToBeDeleted, User userToExecuteTheDelete)
         {
-            _userRepository.DeleteOnCommit(user);
+            var username = userToBeDeleted.Username;
+            var wasConfirmed = userToBeDeleted.Confirmed;
+
+            // Completely remove the user to guarantee that all user data is purged.
+            _userRepository.DeleteOnCommit(userToBeDeleted);
+
+            if (wasConfirmed)
```
loic-sharma: Please add a test case for the unconfirmed account deletion

joelverhagen:
Does this actually clear the existing relationships? In your design spec you mentioned the "pushed by user" relationship on the "Packages" table. How is that getting cleared out?

Do we have some custom SQL in migrations already that is doing the proper cascade in all cases?

scottbommarito:
@joelverhagen - The relationship doesn't get cleared out--fixing bugs with the `DeleteAccountService` will be done separately from this change. What this change guarantees is that if there are any FK relationships on the user row that was deleted, the delete commit will fail because the new row will have a different key.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -285,19 +271,33 @@ private void RemoveMembers(Organization organization)
...
 
-        private void RemoveUser(User user)
+        private void RemoveUser(User userToBeDeleted, User userToExecuteTheDelete)
         {
-            _userRepository.DeleteOnCommit(user);
+            var username = userToBeDeleted.Username;
+            var wasConfirmed = userToBeDeleted.Confirmed;
+
+            // Completely remove the user to guarantee that all user data is purged.
+            _userRepository.DeleteOnCommit(userToBeDeleted);
```
cristinamanum: Do we have a plan to transition the current data to follow this model?

cristinamanum: We will also need to consider cases when new entities are added but data relationship is not cleared before this call. Fixing it will mostly require gallery redeployment.

scottbommarito: The current data follows this model, so no changes are needed, but we will need to do a pass through the existing entities and clean out any unexpected DB relationships.

We will either have to hotfix the gallery or do a manual DB change to delete the user. I think this is fine, however, because this is typical for bugs in production, and I think requiring us to fix a bug before deleting a user is better than having a data issue.

Also, yes, with this design, the Gallery.Maintenance job will delete both the "ghost" user row and the account delete row.

cristinamanum: Automation of this scenario may be impacted by this as well.

scottbommarito: Automation of account deletes will be affected by this change, as there's a chance that attempting to delete the user row will fail, but our jobs that automatically delete accounts could fail to delete an account for any number of reasons. Although it's true that this change may make it more likely to fail, I think that's acceptable because by failing, we are preventing data from getting into a bad state. We already need to add monitoring and resilience to our jobs that automatically delete accounts, and this change doesn't change that.

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: We already have tests for unconfirmed user and org deletion. I didn't make any changes to it because the behavior is the same.

scottbommarito:
Closing this. We should reopen if we determine that we would like to do this.

PR #7076, opened 2019-04-12
DeleteAccountService should clean up Package.User and PackageDeprecation.DeprecatedBy

scottbommarito:
https://github.com/nuget/engineering/issues/2358

These relationships aren't being cleaned up.

We also aren't cleaning up `AccountDelete.DeletedBy` and `PackageDelete.DeletedBy`, but I'm not doing them as part of this change as fixing them will require a DB migration.

### DIFF
--- a/src/NuGetGallery/Services/DeleteAccountService.cs
+++ b/src/NuGetGallery/Services/DeleteAccountService.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Data.Entity;
 using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
@@ -26,6 +25,7 @@ namespace NuGetGallery
         private readonly IReservedNamespaceService _reservedNamespaceService;
         private readonly ISecurityPolicyService _securityPolicyService;
         private readonly AuthenticationService _authService;
+        private readonly IEntityRepository<PackageDeprecation> _deprecationRepository;
         private readonly IEntityRepository<User> _userRepository;
         private readonly IEntityRepository<Scope> _scopeRepository;
         private readonly ISupportRequestService _supportRequestService;
@@ -35,6 +35,7 @@ namespace NuGetGallery
 
         public DeleteAccountService(
             IEntityRepository<AccountDelete> accountDeleteRepository,
+            IEntityRepository<PackageDeprecation> deprecationRepository,
             IEntityRepository<User> userRepository,
             IEntityRepository<Scope> scopeRepository,
             IEntitiesContext entitiesContext,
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -209,6 +214,17 @@ private async Task RemovePackageOwnership(User user, User requestingUser, Accoun
...
         }
 
+        private void RemovePackagePushedBy(User user)
+        {
+            foreach (var package in _entitiesContext
+                .Packages
+                .Where(p => p.UserKey == user.Key)
+                .ToList())
+            {
+                package.User = null;
```
skofman1: does the code that uses this field behave correctly when it sees a null in the user column? 
Also, add a comment above this field in the Package class that it can be null if the user is deleted. Add similar comments in all the cases you fixed.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -218,15 +234,32 @@ private List<Package> GetPackagesOwnedByUser(User user)
...
         }
+
+        private void RemovePackageDeprecatedBy(User user)
+        {
+            foreach (var deprecation in _deprecationRepository
+                .GetAll()
+                .Where(d => d.DeprecatedByUserKey == user.Key)
+                .ToList())
+            {
+                deprecation.DeprecatedByUser = null;
```
skofman1: same as above. Does the code that reads this field handles nulls correctly?

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -368,6 +368,13 @@ public async Task RemovePackageOwnerAsync(PackageRegistration package, User user
         {
             // To support the delete account scenario, the admin can delete the last owner of a package.
             package.Owners.Remove(user);
+
+            // If the package's required signer is the user, remove them.
+            if (package.RequiredSigners.Any(s => s.MatchesUser(user)))
+            {
+                await SetRequiredSignerAsync(package, null, false);
```
skofman1: use named params

cristinamanum: Will this remove all the required signers? Is this ok from functionality point of view?

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: Yes, this field didn't always exist so older packages have `null` instead of an actual user. The UI handles it fine.

src/NuGetGallery/Services/PackageService.cs
scottbommarito: Removing this from the PR--discussed with @dtivel and this is not a good solution.

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: Nothing uses this today, but I would like to fix it as part of this change.

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: Added comments.

src/NuGetGallery/Services/PackageService.cs
scottbommarito: This change was removed, so no longer relevant.

Issue #6624, opened 2018-11-05
Message duplication causes orchestrator messages dead-letter 


joelverhagen:
Orchestrator has logic that clones messages and re-enqueues them so that we can try a message more than the maximum attempts set on the subscription (20). In short, orchestrator does this:

1. Gets a message.
1. Process the message.
1. If the validation set is not done, re-enqueue
1. Complete the message.

If the "complete the message" step fails, that means there are now two messages for the one validation set. When the validation set finally completes, one message "wins" by copying the package from to the public container. The other message "loses" because it can't copy the package from the validation set location to the public container because the first message deleted the validation set copy.

This causes dead-lettering.

This issue is a continuation of https://github.com/NuGet/NuGetGallery/issues/6515.

In my on-call week I had 34+ dead-letter messages.

loic-sharma:
```
traces
| where cloud_RoleName == "NuGet.Services.Validation.Orchestrator" 
| extend CallGuid = tostring(customDimensions.CallGuid)
| summarize max(timestamp), min(timestamp) by CallGuid
| extend Duration = (max_timestamp  - min_timestamp)/1s
| join (
  exceptions
  | where cloud_RoleName == "NuGet.Services.Validation.Orchestrator" 
  | where outerMessage contains "The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue."
  | extend CallGuid = tostring(customDimensions.CallGuid)
  | project CallGuid
) on CallGuid
| join (
  traces
  | where cloud_RoleName == "NuGet.Services.Validation.Orchestrator"
  | extend CallGuid = tostring(customDimensions.CallGuid)
  | extend ValidationSetId = tostring(customDimensions.ValidationSetId)
  | where ValidationSetId <> ""
  | summarize any(ValidationSetId) by CallGuid 
  | extend ValidationSetId = any_ValidationSetId
  | project ValidationSetId, CallGuid
) on CallGuid
| join (
  traces
  | where cloud_RoleName == "NuGet.Services.Validation.Orchestrator" 
  | where message contains "Before calling FetchAttributesAsync(), the source blob 'validation-sets/"
  | extend ValidationSetId = tostring(customDimensions.ValidationSetId)
  | summarize count() by ValidationSetId
  | extend CopyFailures = count_
) on ValidationSetId 
| project ValidationSetId, CallGuid, Duration, CopyFailures
```

loic-sharma:
[...]

PR #7084, opened 2019-04-17
Sending validation request right before committing changes to DB

agr:
Related to #6624 

Targeted fix for the "DuplicatePushesAreRejectedAndNotDeleted" tests causing deadlettering in validation pipeline. Validation request ServiceBus message now sent right before committing changes to DB, which in case of that test means there will be no surge of validation request for the same package.

Since the post validation request package status is used in `PackageUploadService` to determine the course of action, updating the package status was extracted into separate method, so it can be done without sending the validation request. It, of course, introduces an issue that we can forget to call the actual method that sends the message, but the exposure is very limited (one call for regular packages and another one for the symbol packages), and we are likely to notice very soon that validations are not running.

### DIFF
--- a/src/NuGetGallery/Services/IPackageValidationInitiator.cs
+++ b/src/NuGetGallery/Services/IPackageValidationInitiator.cs
@@ -15,12 +15,14 @@ namespace NuGetGallery
         /// <summary>
         /// Returns the package status that package should go into when validation is started.
         /// Async validations typically return <see cref="PackageStatus.Validating"/>.
-        /// Sync, non-blocking or no validaiton typically return <see cref="PackageStatus.Available"/>.
+        /// Sync, non-blocking or no validation typically return <see cref="PackageStatus.Available"/>.
         /// Caller still must call <see cref="IPackageValidationInitiator{TPackageEntity}.StartValidationAsync(TPackageEntity)"/>
         /// to start the actual validation.
         /// </summary>
         /// <param name="package">The <see cref="TPackageEntity"/> to get future validation status for.</param>
-        /// <returns></returns>
+        /// <returns>The new package status assuming the validation will be started later.</returns>
+        /// <remarks>This method validates the argument the same way <see cref="StartValidationAsync(TPackageEntity)"/> does
+        /// and will throw on invalid input allowing to fail earlier, before expensive operations are performed.</remarks>
         PackageStatus GetPackageStatus(TPackageEntity package);
 
         /// <summary>
--- a/src/NuGetGallery/Services/IPackageValidationInitiator.cs
+++ b/src/NuGetGallery/Services/IPackageValidationInitiator.cs
@@ -13,6 +13,17 @@ namespace NuGetGallery
         where TPackageEntity: IPackageEntity
     {
         /// <summary>
+        /// Returns the package status that package should go into when validation is started.
+        /// Async validations typically return <see cref="PackageStatus.Validating"/>.
+        /// Sync, non-blocking or no validaiton typically return <see cref="PackageStatus.Available"/>.
+        /// Caller still must call <see cref="IPackageValidationInitiator{TPackageEntity}.StartValidationAsync(TPackageEntity)"/>
+        /// to start the actual validation.
+        /// </summary>
+        /// <param name="package">The <see cref="TPackageEntity"/> to get future validation status for.</param>
+        /// <returns></returns>
+        PackageStatus GetPackageStatus(TPackageEntity package);
+
+        /// <summary>
         /// Starts the validation for the specified IPackageEntity. The validation can be done asynchronously with respect
         /// to the gallery and therefore may not be complete when the returned <see cref="Task"/> completes. This
         /// pending validation state is indicated by the package having the <see cref="PackageStatus.Validating"/>
--- a/src/NuGetGallery/Services/ImmediatePackageValidator.cs
+++ b/src/NuGetGallery/Services/ImmediatePackageValidator.cs
@@ -13,6 +13,9 @@ namespace NuGetGallery
     public class ImmediatePackageValidator<TPackageEntity> : IPackageValidationInitiator<TPackageEntity> 
         where TPackageEntity: IPackageEntity
     {
+        public PackageStatus GetPackageStatus(TPackageEntity package)
+            => PackageStatus.Available;
+
         public Task<PackageStatus> StartValidationAsync(TPackageEntity package)
         {
             return Task.FromResult(PackageStatus.Available);
--- a/src/NuGetGallery/Services/PackageUploadService.cs
+++ b/src/NuGetGallery/Services/PackageUploadService.cs
@@ -674,7 +674,7 @@ namespace NuGetGallery
 
         public async Task<PackageCommitResult> CommitPackageAsync(Package package, Stream packageFile)
         {
-            await _validationService.StartValidationAsync(package);
+            await _validationService.UpdatePackageAsync(package);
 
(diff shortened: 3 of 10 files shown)


src/NuGetGallery/Services/IPackageValidationInitiator.cs
```diff
@@ -12,6 +12,17 @@ namespace NuGetGallery
     public interface IPackageValidationInitiator<TPackageEntity> 
         where TPackageEntity: IPackageEntity
     {
+        /// <summary>
+        /// Returns the package status that package should go into when validation is started.
+        /// Async validations typically return <see cref="PackageStatus.Validating"/>.
+        /// Sync, non-blocking or no validaiton typically return <see cref="PackageStatus.Available"/>.
```
loic-sharma: >validaiton [](start = 37, length = 10)

validation

src/NuGetGallery/Services/IPackageValidationInitiator.cs
```diff
@@ -12,6 +12,17 @@ namespace NuGetGallery
...
+        /// Caller still must call <see cref="IPackageValidationInitiator{TPackageEntity}.StartValidationAsync(TPackageEntity)"/>
+        /// to start the actual validation.
+        /// </summary>
+        /// <param name="package">The <see cref="TPackageEntity"/> to get future validation status for.</param>
+        /// <returns></returns>
+        PackageStatus GetPackageStatus(TPackageEntity package);
```
loic-sharma: >GetPackageStatus [](start = 22, length = 16)

Let's call out that this method may throw

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -92,6 +103,34 @@ public IReadOnlyList<ValidationIssue> GetLatestPackageValidationIssues(SymbolPac
...
+        }
+
+        public async Task RevalidateAsync(SymbolPackage symbolPackage)
+        {
+            await _symbolPackageValidationInitiator.StartValidationAsync(symbolPackage);
+
+            _telemetryService.TrackSymbolPackageRevalidate(symbolPackage.Id, symbolPackage.Version);
+        }
+
+        private async Task UpdatePackageInternalAsync(Package package, PackageStatus packageStatus)
```
loic-sharma: >UpdatePackageInternalAsync [](start = 27, length = 26)

Is this method necessary? I would inline this

src/NuGetGallery/Services/ValidationService.cs
```diff
@@ -50,14 +50,25 @@ public class ValidationService : IValidationService
             }
         }
 
+        public async Task UpdatePackageAsync(Package package)
```
loic-sharma: >UpdatePackageAsync [](start = 26, length = 18)

It is not at all clear that this does not commit changes to the database. I would accept a `commitChanges` parameter that your callers set to `false` to make this explicit.

loic-sharma: Same with `StartValidationAsync`, even though that's not new code.

---
In reply to: [276305264](https://github.com/NuGet/NuGetGallery/pull/7084#discussion_r276305264) [](ancestors = 276305264)

src/NuGetGallery/Services/ValidationService.cs
agr: It is called from two locations and I wanted to eliminate the possibility of being inconsistent with the `commitChanges` argument.

shishirx34: same for `UpdateSymbolPackageInternalAsync` below.

src/NuGetGallery/Services/ImmediatePackageValidator.cs
```diff
@@ -13,6 +13,9 @@ namespace NuGetGallery
     public class ImmediatePackageValidator<TPackageEntity> : IPackageValidationInitiator<TPackageEntity> 
         where TPackageEntity: IPackageEntity
     {
+        public PackageStatus GetPackageStatus(TPackageEntity package)
+            => PackageStatus.Available;
```
cristinamanum: Why always available? Comments?

src/NuGetGallery/Services/SymbolPackageUploadService.cs
```diff
@@ -206,6 +206,8 @@ public async Task<PackageCommitResult> CreateAndUploadSymbolsPackage(Package pac
                     await _symbolPackageFileService.SavePackageFileAsync(symbolPackage.Package, symbolPackageFile, overwrite);
                 }
 
+                await _validationService.StartValidationAsync(symbolPackage);
```
shishirx34: could you ensure that this doesn't cause any consistency issues with the above code of saving/overwriting validation packages for symbols.. I don't think it should but another set of eyes should be helpful.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -761,6 +761,8 @@ public async Task<PackageCommitResult> CommitPackageAsync(Package package, Strea
                 return PackageCommitResult.Conflict;
             }
 
+            await _validationService.StartValidationAsync(package);
```
cristinamanum: Can we add comments that the order is intentional?

cristinamanum: If an exception happens before this line and after the package was saved in the validation container can the package be resubmitted?

src/NuGetGallery/Services/ValidationService.cs
agr: There are no conditions under which this method would commit, hence no parameter is needed.

src/NuGetGallery/Services/ImmediatePackageValidator.cs
agr: Comment on the class covers it:

> A validation initiator that immediately marks the package as validated. In other words, no asynchronous validation is performed by this implementation.

src/NuGetGallery/Services/PackageUploadService.cs
agr: The only code between saving the package into validation container and sending the message are exception handlers, but sending itself may throw and we need to clean up in this case, will move the call into the try/catch below.

agr: Done

src/NuGetGallery/Services/SymbolPackageUploadService.cs
agr: `StartValidationAsync` does two things:
1. Updates the package status (which is already done by the previous call to `UpdatePackageAsync`);
2. Sends the validation request to the service bus topic.
So, with regards to the file operations, there are no changes in the execution flow.

agr: The only potential issue I foresee is `StartValidationAsync` throwing, as Cristina suggested below, so I moved it into the try/catch together with `SaveChangesAsync` call, so it is properly cleaned up if fails.

src/NuGetGallery/Services/ValidationService.cs
loic-sharma: @agr A comment would be better if the goal is to keep two methods consistent.

src/NuGetGallery/Services/ValidationService.cs
loic-sharma: That’s true for a lot of methods that accept `commitChanges` parameter. The goal is to make that decision clear when the method is called. I would either add this parameter, or I would document this method and call out it does not commit

PR #7095, opened 2019-04-24
The new DB migration job

zhhyu:
The new DB migration job for Gallery.


src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Data.SqlClient;
+using System.Data.Entity.Migrations;
+using NuGetGallery;
+using NuGetGallery.Migrations;
+using NuGet.Jobs;
+using NuGet.Jobs.Configuration;
+
+namespace DatabaseMigrationTools
```
scottbommarito: IMO `NuGetGallery.DatabaseMigrationTools` or something similar that has this job as a subnamespace under `NuGetGallery` would be better. This project isn't generic and is specific for migrating the NuGetGallery DB.

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,83 @@
...
+                var migrationsConfiguration = new MigrationsConfiguration();
+                var migrator = new DbMigrator(migrationsConfiguration);
+
+                ExecuteDatabaseMigration(migrator, sqlConnection, accessToken);
+            }
+        }
+
+        private void ExecuteDatabaseMigration(DbMigrator migrator, SqlConnection sqlConnection, string accessToken)
+        {
+            var historyRepository = typeof(DbMigrator).GetField(
```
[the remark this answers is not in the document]
zhhyu: This is a bug for entity framework's migration to use AAD token for database connection. We contact EF team, and they suggest that we have to use this way to overcome the problem effectively and safely.
The key is that we need to overwrite the private database connection field.
EF team will fix this issue in the next version.
I will leave some comments there.

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,83 @@
...
+    class Job : JsonConfigurationJob
+    {
+        public override async Task Run()
+        {
+            using (var sqlConnection = await _serviceProvider.GetRequiredService<ISqlConnectionFactory<GalleryDbConfiguration>>().CreateAsync())
+            {
+                var accessToken = sqlConnection.AccessToken;
+                DbContextFactory.EntitiesContextFactory = () =>
+                {
+                    if (sqlConnection.State == ConnectionState.Closed)
```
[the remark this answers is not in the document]
zhhyu: Once the connection is closed, the secret access token will be cleaned up. In my understanding, considering the behavior of "DbMigrator", we will have to reuse the same connection from the connection pool for multiple times, so we have to set the access token again if the connection is closed and build the connection with authorization.

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,83 @@
...
+                    {
+                        sqlConnection.AccessToken = accessToken;
+                    }
+
+                    return new EntitiesContext(sqlConnection, true);
```
shishirx34: add the parameter for `true`

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        private void RunDatabaseMigration(SqlConnection sqlConnection, string accessToken)
+        {
+            SetDbContextFactory(sqlConnection, accessToken);
+            var migrator = SetDbMigrator();
+
+            ExecuteDatabaseMigration(migrator, sqlConnection, accessToken);
+        }
+
+        private DbMigrator SetDbMigrator()
```
shishirx34: why create these private methods when its used only once? Only refactor if the code size is getting big or logically needs to be separate or will be reused.

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            // Reset the access token if the connection is closed to ensure connection authorization.
+            DbContextFactory.EntitiesContextFactory = () =>
+            {
+                if (sqlConnection.State == ConnectionState.Closed)
+                {
+                    sqlConnection.AccessToken = accessToken;
+                }
+
+                return new EntitiesContext(sqlConnection, true);
```
shishirx34: nit: parameter variable

src/DatabaseMigrationTools/Job.cs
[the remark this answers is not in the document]
shishirx34: add the comments about the bug if possible.

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,98 @@
...
+            // Overwrite the database connection of DbMigrator.
+            // Consider updating this section when the new Entity Framework 6.3 or higher version is released.
+            var historyRepository = typeof(DbMigrator).GetField(
+                                    "_historyRepository",
+                                    BindingFlags.NonPublic | BindingFlags.Instance)
+                                    .GetValue(migrator);
+
+            var connectionField = historyRepository.GetType().BaseType.GetField(
+                                        "_existingConnection",
+                                        BindingFlags.NonPublic | BindingFlags.Instance);
```
shishirx34: nit: indentation

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,98 @@
...
+                return new EntitiesContext(sqlConnection, true);
+            };
+        }
+
+        private void ExecuteDatabaseMigration(DbMigrator migrator, SqlConnection sqlConnection, string accessToken)
+        {
+            // Overwrite the database connection of DbMigrator.
+            // Consider updating this section when the new Entity Framework 6.3 or higher version is released.
+            var historyRepository = typeof(DbMigrator).GetField(
+                                    "_historyRepository",
```
shishirx34: nit: may be set these values as private const fields?

shishirx34: also, indentation.

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            sqlConnection.AccessToken = accessToken;
+            sqlConnection.Open();
+            connectionField.SetValue(historyRepository, sqlConnection);
+
+            var pendingMigrations = migrator.GetPendingMigrations();
+            if (pendingMigrations.Count() > 0)
+            {
+                Logger.LogInformation("Executing pending migrations: \n {PendingMigrations}", String.Join("\n", pendingMigrations));
+                migrator.Update();
```
shishirx34: does migrator publish the "Verbose" details of the migrations it is running?

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -1,15 +1,32 @@
...
+    /// 1. Run migration using DbConnection; (For DatabaseMigrationTools with AAD token)
+    /// 2. Run migration using connection string;
+    /// 3. Run migration using default connection string ("Gallery.SqlServer") in a web.config; (For command-line migration with integrated AAD/username+password)
+    /// </summary>
+    public class DbContextFactory : IDbContextFactory<EntitiesContext>
+    {
+        public static Func<EntitiesContext> EntitiesContextFactory;
+        public EntitiesContext Create()
+        {
+            return EntitiesContextFactory == null ? new EntitiesContext("Gallery.SqlServer", false) : EntitiesContextFactory();
```
shishirx34: parameter variable.

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        public override async Task Run()
+        {
+            using (var sqlConnection = await _serviceProvider.GetRequiredService<ISqlConnectionFactory<GalleryDbConfiguration>>().CreateAsync())
+            {
+                RunDatabaseMigration(sqlConnection, sqlConnection.AccessToken);
+            }
+        }
+
+        private void RunDatabaseMigration(SqlConnection sqlConnection, string accessToken)
```
shishirx34: I don't see the need for a different method here, since you are passing `SqlConnection` to the method.

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,98 @@
...
+                                    "_historyRepository",
+                                    BindingFlags.NonPublic | BindingFlags.Instance)
+                                    .GetValue(migrator);
+
+            var connectionField = historyRepository.GetType().BaseType.GetField(
+                                        "_existingConnection",
+                                        BindingFlags.NonPublic | BindingFlags.Instance);
+
+            sqlConnection.AccessToken = accessToken;
+            sqlConnection.Open();
```
shishirx34: do you need to explicitly call Open here? You are `using` in top level methods.

src/DatabaseMigrationTools/Job.cs
zhhyu: It will log the migration file's name which are executed.

shishirx34: but not the output of the migration that is run? no SQL output?

src/DatabaseMigrationTools/Job.cs
zhhyu: I don't want to touch this place because these steps are suggested by EF team and it works here. 
We can get rid of this when the new version of EF is released in the future.

src/DatabaseMigrationTools/Job.cs
zhhyu: It will output the same verbose info of SQL scripts as the command line.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -1,15 +1,32 @@
...
+    /// 1. Run migration using DbConnection; (For DatabaseMigrationTools with AAD token)
+    /// 2. Run migration using connection string;
+    /// 3. Run migration using default connection string ("Gallery.SqlServer") in a web.config; (For command-line migration with integrated AAD/username+password)
+    /// </summary>
+    public class DbContextFactory : IDbContextFactory<EntitiesContext>
+    {
+        public static Func<EntitiesContext> EntitiesContextFactory;
+        public EntitiesContext Create()
+        {
+            return EntitiesContextFactory == null ? new EntitiesContext("Gallery.SqlServer", readOnly: false) : EntitiesContextFactory();
```
shishirx34: This will explicitly run only Gallery DB migrations. I think we should have this passed in as parameter so that we can run the job with `SupportRequestDb` migrations as well.

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,109 @@
...
+            OverWriteSqlConnection(migratorForScripting, sqlConnection, accessToken);
+
+            var pendingMigrations = migrator.GetPendingMigrations();
+            if (pendingMigrations.Count() > 0)
+            {
+                Logger.LogInformation("Target database is: {DataSource}/{Database}", sqlConnection.DataSource, sqlConnection.Database);
+                Logger.LogInformation("Applying pending migrations: \n {PendingMigrations}", String.Join("\n", pendingMigrations));
+
+                var migratorScripter = new MigratorScriptingDecorator(migratorForScripting);
+                var migrationScripts = migratorScripter.ScriptUpdate(null, null);
```
shishirx34: parameter variable.

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,109 @@
...
+using NuGet.Jobs.Configuration;
+using System.Data.Entity.Migrations.Infrastructure;
+
+namespace NuGetGallery.DatabaseMigrationTools
+{
+    class Job : JsonConfigurationJob
+    {
+        public override async Task Run()
+        {
+            using (var sqlConnection = await _serviceProvider.GetRequiredService<ISqlConnectionFactory<GalleryDbConfiguration>>().CreateAsync())
```
shishirx34: how will we run the migrations on `Validation` database? This is only configured for Gallery database. Will we have a separate job for running validation database migrations?

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,109 @@
...
+                    Logger.LogInformation("Executing migrations...");
+                    migrator.Update();
+                }
+                catch (Exception e)
+                {
+                    Logger.LogError(0, e, "Failed to execute migrations on the target database {DataSource}/{Database}", sqlConnection.DataSource, sqlConnection.Database);
+                    throw;
+                }
+
+                Logger.LogInformation("Finished executing {pendingMigrationsCount} migrations successfully.", pendingMigrations.Count());
```
shishirx34: move this to the `try` block?

src/DatabaseMigrationTools/Job.cs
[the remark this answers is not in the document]
shishirx34: why would the connection be closed though? When would this be invoked? Its not like we are keeping the connection open forever, in the above line you are `using` which should open the connection, right? I do not see in the code below that we are doing a `Thread.Sleep` or looping forever or doing multiple Sql operations to reconnect. 
Can you set a breakpoint in here and see when this code gets invoked?

src/NuGetGallery.Core/Entities/EntitiesContext.cs
shishirx34: you can also test this locally with supportrequest DB migrations

src/DatabaseMigrationTools/Job.cs
```diff
@@ -0,0 +1,109 @@
...
+                    {
+                        // Reset the access token if the connection is closed to ensure connection authorization.
+                        sqlConnection.AccessToken = accessToken;
+                    }
+
+                    return new EntitiesContext(sqlConnection, readOnly: false);
+                };
+
+                Logger.LogInformation("Initializing DbMigrator...");
+                var migrationsConfiguration = new MigrationsConfiguration();
```
shishirx34: how do you figure out which Migrations configuration file to be used here?

shishirx34: talked offline: the name of the class is the configuration file here.

zhhyu:
The validation database migration still looks ambiguous.

src/DatabaseMigrationTools/Job.cs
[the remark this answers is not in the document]
zhhyu: I think this is due to the behavior of Migrator and our codes, especially when we initialize the Migrator and get pending migration files. If we really want to understand details of this behavior, we need to have a deeper investigation on the entity framework and ADO.NET code base. 
In my understanding, the connection is picked up from the connection pool, so the cost to open/close the connection should be not very expensive.

src/DatabaseMigrationTools/IMigrationContext.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Data.Entity.Migrations;
+using System.Data.SqlClient;
+
+namespace NuGetGallery.DatabaseMigrationTools
+{
+    public interface IMigrationContext
+    {
+        SqlConnection SqlConnection { get; }
+        string SqlConnectionAccessToken { get; }
+        DbMigrator Migrator { get; }
```
joelverhagen: Alternatively, you could just have a Func<DbMigrator> property.

src/DatabaseMigrationTools/SupportRequestDbMigrationContext.cs
```diff
@@ -0,0 +1,40 @@
...
+        public Func<DbMigrator> GetDbMigrator { get; }
+
+        public SupportRequestDbMigrationContext(SqlConnection sqlConnection)
+        {
+            SqlConnection = sqlConnection ?? throw new ArgumentNullException(nameof(sqlConnection));
+            SqlConnectionAccessToken = sqlConnection.AccessToken;
+
+            SupportRequestDbContextFactory.SupportRequestEntitiesContextFactory = () =>
+            {
+                if (SqlConnection.State == ConnectionState.Closed)
```
joelverhagen: >SqlConnection [](start = 20, length = 13)

this `if` block seems duplicated. Can we move it to a shared place?

src/DatabaseMigrationTools/SupportRequestDbMigrationContext.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Data.SqlClient;
+using NuGet.Services.DatabaseMigration;
+using NuGetGallery.Areas.Admin;
+using NuGetGallery.Areas.Admin.Models;
+
+namespace NuGetGallery.DatabaseMigrationTools
+{
+    public class SupportRequestDbMigrationContext : IMigrationContext
+    {
+        public SqlConnection SqlConnection { get; }
```
joelverhagen: These definitions could be de-duplicated if you make a BaseMigrationContext class.

src/DatabaseMigrationTools/MigrationContextFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+using NuGet.Services.DatabaseMigration;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace NuGetGallery.DatabaseMigrationTools
+{
+    public class MigrationContextFactory : IMigrationContextFactory
+    {
+        private IDictionary<string, Func<IServiceProvider, Task<IMigrationContext>>> _dictionary = new Dictionary<string, Func<IServiceProvider, Task<IMigrationContext>>>
```
joelverhagen: static readonly, right?

Also, `IReadOnlyDictionary`

src/NuGet.Services.DatabaseMigration/NuGet.Services.DatabaseMigration.csproj
```diff
@@ -0,0 +1,95 @@
...
+      <Version>1.1.2</Version>
+    </PackageReference>
+    <PackageReference Include="NuGet.Jobs.Common">
+      <Version>4.1.0-master-2602271</Version>
+    </PackageReference>
+    <PackageReference Include="NuGet.Services.Configuration">
+      <Version>2.46.0</Version>
+    </PackageReference>
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
```
joelverhagen: This need signing.

src/NuGet.Services.DatabaseMigration/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+//
+//      Major Version
+//      Minor Version
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
```
joelverhagen: These should be generated at build time. These should come from `AssemblyInfo.g.cs`

src/NuGet.Services.DatabaseMigration/NuGet.Services.DatabaseMigration.csproj
```diff
@@ -0,0 +1,95 @@
...
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="IMigrationContext.cs" />
+    <Compile Include="IMigrationContextFactory.cs" />
+    <Compile Include="Job.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
```
joelverhagen: <Compile Include="Properties\AssemblyInfo.*.cs" />

src/DatabaseMigrationTools/MigrationContextFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+                {
+                    var sqlConnection = await serviceProvider.GetRequiredService<ISqlConnectionFactory<GalleryDbConfiguration>>().CreateAsync();
+                    return new GalleryDbMigrationContext(sqlConnection);
+                }
+            }
+        };
+
+        public async Task<IMigrationContext> CreateMigrationContextAsync(string migrationTargetDatabase, IServiceProvider serviceProvider)
+        {
+            return await _dictionary[migrationTargetDatabase](serviceProvider);
```
joelverhagen: Should use `TryGetValue` and throw `ArgumentException` if the `migrationTargetDatabase` is not a valid key

joelverhagen:
Update build.ps1 to generate AssemblyInfo.g.cs and to pack

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -0,0 +1,121 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Autofac;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using System.Threading.Tasks;
+using System;
+using System.Linq;
+using System.Reflection;
```
joelverhagen: System.* at the top

src/NuGet.Services.DatabaseMigration/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("NuGet.Services.DatabaseMigration")]
+[assembly: AssemblyCopyright("Copyright ©  2019")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+
+// Setting ComVisible to false makes the types in this assembly not visible
+// to COM components.  If you need to access a type in this assembly from
+// COM, set the ComVisible attribute to true on that type.
+[assembly: ComVisible(false)]
```
joelverhagen: This AssemblyInfo.cs has different info than the other one in your PR. Could we try to get some consistency with the existing projects? Perhaps use NuGet.Services.Entities or NuGetGallery.Core as inspiration.

src/NuGet.Services.DatabaseMigration/IMigrationContext.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Data.Entity.Migrations;
+using System.Data.SqlClient;
+
+namespace NuGet.Services.DatabaseMigration
+{
+    public interface IMigrationContext
```
agr: Please, add documentation for class and/or methods.

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -0,0 +1,121 @@
...
+            else
+            {
+                Logger.LogInformation("There are no pending migrations to execute.");
+            }
+        }
+
+        // Overwrite the database connection of DbMigrator.
+        // Hit the bug:  https://github.com/aspnet/EntityFramework6/issues/522
+        // Consider deleting/updating this section when the new Entity Framework 6.3 or higher version is released.
+        private void OverWriteSqlConnection(DbMigrator migrator, SqlConnection sqlConnection, string accessToken)
```
agr: `OverwriteSqlConnection`. Don't capitalize `W`: 'overwrite' is a single word.

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            var historyRepository = typeof(DbMigrator).GetField(
```
agr: This code modifies private fields, right? This might break as a result of EF package upgrade. We might want to be more defensive here. Maybe add a test that verifies that those fields exist in `DbMigrator`?

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -0,0 +1,121 @@
...
+            }
+            else
+            {
+                Logger.LogInformation("There are no pending migrations to execute.");
+            }
+        }
+
+        // Overwrite the database connection of DbMigrator.
+        // Hit the bug:  https://github.com/aspnet/EntityFramework6/issues/522
+        // Consider deleting/updating this section when the new Entity Framework 6.3 or higher version is released.
```
agr: Do we have an issue open to check if future versions of EF have this fixed?

src/NuGet.Services.DatabaseMigration/Job.cs
agr: Or add a test that checks if specific version is used, so package upgrade would involve some verification steps...

src/NuGet.Services.DatabaseMigration/Job.cs
zhhyu: https://github.com/nuget/engineering/issues/2392

zhhyu:
Only build.ps1 is left now.

src/NuGet.Services.DatabaseMigration/Properties/AssemblyInfo.cs
zhhyu: I find that "NuGet.Services.Entities" doesn't have any "AssemblyInfo.g.cs" file, but building script contains that?

src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
```diff
@@ -0,0 +1,105 @@
...
+    <ProjectReference Include="..\NuGetGallery.Core\NuGetGallery.Core.csproj">
+      <Project>{097b2cdd-9623-4c34-93c2-d373d51f5b4e}</Project>
+      <Name>NuGetGallery.Core</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\NuGetGallery\NuGetGallery.csproj">
+      <Project>{1dacf781-5cd0-4123-8bac-cd385d864be5}</Project>
+      <Name>NuGetGallery</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
```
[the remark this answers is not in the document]
zhhyu: signing is added now.

tests/NuGet.Services.DatabaseMigration.Facts/DatabaseMigrationFacts.cs
```diff
@@ -0,0 +1,50 @@
...
+        public DatabaseMigrationFacts()
+        {
+            var sln = TryFindFile(Directory.GetCurrentDirectory(), "NuGetGallery.sln");
+            var csproj = new FileInfo(Path.Combine(sln.Directory.FullName, @"src\NuGet.Services.DatabaseMigration\NuGet.Services.DatabaseMigration.csproj"));
+
+            _csprojDocument = XDocument.Load(csproj.FullName);
+        }
+
+        [Fact]
+        public void VerifyEntityFrameworkVersion()
```
agr: Need a comment regarding why this test exists, so the person who would look into why it fails knows the reasoning behind requiring specific EF version.

src/DatabaseMigrationTools/BaseDbMigrationContext.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Data;
+using System.Data.Entity.Migrations;
+using System.Data.SqlClient;
+using NuGet.Services.DatabaseMigration;
+
+namespace NuGetGallery.DatabaseMigrationTools
```
shishirx34: do you think may be the namespace shouldn't be "NuGetGallery"? If this is supposed to be referenced by the validation?

src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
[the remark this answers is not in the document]
shishirx34: You also need to add package reference for MicroBuild.Core project.. checkout one of the other project which has signing setup, otherwise you will get signing failures during build.

src/DatabaseMigrationTools/JobArgumentNames.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.DatabaseMigrationTools
+{
+    internal static class JobArgumentNames
+    {
+        // avoids value duplication, avoids annoying namespace conflicts in this job
+        public const string GalleryDatabase = NuGet.Jobs.JobArgumentNames.GalleryDatabase;
+        // need to add these database argument names to NuGet.Jobs
```
shishirx34: is this a TODO? May be create an issue for tracking and add in the comment here, so it isn't missed.

src/DatabaseMigrationTools/JobArgumentNames.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.DatabaseMigrationTools
+{
+    internal static class JobArgumentNames
```
shishirx34: rename this class to be `GalleryDatabaseNames` or something appropriate. JobArgumentNames is kind of too generic.

src/DatabaseMigrationTools/MigrationContextFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+                }
+            }
+        };
+
+        public async Task<IMigrationContext> CreateMigrationContextAsync(string migrationTargetDatabase, IServiceProvider serviceProvider)
+        {
+            Func<IServiceProvider, Task<IMigrationContext>> migrationContext;
+            if (_dictionary.TryGetValue(migrationTargetDatabase, out migrationContext))
+            {
+                return await _dictionary[migrationTargetDatabase](serviceProvider);
```
shishirx34: why not `return await migrationContext(serviceProvider)`?

src/DatabaseMigrationTools/Program.cs
```diff
@@ -0,0 +1,18 @@
...
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            var migrationContextFactory = new MigrationContextFactory();
+            var job = new Job(migrationContextFactory);
+            JobRunner.RunOnce(job, args).GetAwaiter().GetResult();
+        }
+    }
+}
```
shishirx34: I am assuming you will have something similar job in ValidationServices?

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        public override async Task Run()
+        {
+            Logger.LogInformation("Initializing database migration context...");
+            var migrationContext = await _migrationContextFactory.CreateMigrationContextAsync(_migrationTargetDatabase, _serviceProvider);
+
+            ExecuteDatabaseMigration(migrationContext.GetDbMigrator,
+                                     migrationContext.SqlConnection,
+                                     migrationContext.SqlConnectionAccessToken);
```
shishirx34: nit: indentation

src/DatabaseMigrationTools/JobArgumentNames.cs
zhhyu: https://github.com/nuget/engineering/issues/2405

src/DatabaseMigrationTools/BaseDbMigrationContext.cs
zhhyu: This one will not be used by validation database.
It sticks to the Gallery and Support request databases.
Validation migration context will exist in another project in Server.Common.

src/DatabaseMigrationTools/JobArgumentNames.cs
zhhyu: Rename to "MigrationTargetDatabaseArgumentNames" now.

src/DatabaseMigrationTools/Program.cs
zhhyu: That one will exist with Server.common

tests/NuGet.Services.DatabaseMigration.Facts/DatabaseMigrationFacts.cs
zhhyu: Add the comment now.

src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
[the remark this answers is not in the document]
zhhyu: Shared package NuGet.Services.DatabaseMigration uses MicroBuild.Core now.

src/DatabaseMigrationTools/Program.cs
```diff
@@ -0,0 +1,18 @@
...
+using NuGet.Jobs;
+using NuGet.Services.DatabaseMigration;
+
+namespace NuGetGallery.DatabaseMigrationTools
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            var migrationContextFactory = new MigrationContextFactory();
```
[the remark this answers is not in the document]
zhhyu: This is used for decoupling. The implementation of "MigrationContextFactory()" is tightly coupled with the target database while the "Job" is inside the shared package.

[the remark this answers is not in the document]
scottbommarito: Sounds good. Maybe rename this to `NuGetGalleryMigrationContextFactory` instead so it's clear that this is only for gallery/support requests?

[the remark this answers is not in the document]
zhhyu: Rename to "GalleryMigrationContextFactory".

build.ps1
```diff
@@ -79,7 +79,8 @@ Invoke-BuildStep 'Set version metadata in AssemblyInfo.cs' {
     $Paths = `
         (Join-Path $PSScriptRoot "src\NuGetGallery\Properties\AssemblyInfo.g.cs"), `
         (Join-Path $PSScriptRoot "src\NuGetGallery.Core\Properties\AssemblyInfo.g.cs"), `
-        (Join-Path $PSScriptRoot "src\NuGet.Services.Entities\Properties\AssemblyInfo.g.cs")
+        (Join-Path $PSScriptRoot "src\NuGet.Services.Entities\Properties\AssemblyInfo.g.cs"), `
+        (Join-Path $PSScriptRoot "src\NuGet.Services.DatabaseMigration\Properties\AssemblyInfo.g.cs")
```
joelverhagen: Need AssemblyInfo.g.cs for DatabaseMigrationTools.

zhhyu:
Decide to run migration for validation database here, considering the difficult to put it within server common.

src/DatabaseMigrationTools/DatabaseMigration.Validation.nuspec
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
```
agr: Are there any differences between `DatabaseMigration.Gallery`, `DatabaseMigration.SupportRequest` and `DatabaseMigration.Validation` besides the package ID?

zhhyu: The package ID will be used to distinguish different jobs for EV2 pipeline.
Other things are the same things.

Issue #7097, opened 2019-04-25
Remember tab selection in package details page


akoeplinger:
Now that https://github.com/NuGet/NuGetGallery/issues/5670 landed it would be nice if the gallery UI would remember the selected tab instead of always reverting back to `Package Manager`.

I never use the default tab since I prefer the `PackageReference` option so I always need to switch to that tab before I can copy the string.

scottbommarito:
@khellang has proposed a great fix. I have assigned myself so I can be in charge of pushing it through our deployment process.

PR #7098, opened 2019-04-25
Store users preferred tab in localStorage

khellang:
Closes https://github.com/NuGet/NuGetGallery/issues/7097

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -90,17 +90,17 @@ $(function () {
 
     var storage = window['localStorage'];
     if (storage) {
-		var key = 'preferred_tab';
-		
+        var key = 'preferred_tab';
+
         // Restore preferred tab selection from localStorage.
         var preferredTab = storage.getItem(key);
         if (preferredTab) {
-            $('.nav-tabs a[aria-controls="' + preferredTab + '"]').tab('show');
+            $('#' + preferredTab).tab('show');
         }
 
         // Make sure we save the user's preferred tab to localStorage.
-        $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
-            storage.setItem(key, e.target.getAttribute("aria-controls"));
+        $('.package-manager-tab').on('shown.bs.tab', function (e) {
+            storage.setItem(key, e.target.id);
         });
     }
 
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -88,6 +88,22 @@ $(function () {
         $(this).closest('form').submit();
     })
 
+    var storage = window['localStorage'];
+    if (storage) {
+		var key = 'preferred_tab';
+		
+        // Restore preferred tab selection from localStorage.
+        var preferredTab = storage.getItem(key);
+        if (preferredTab) {
+            $('.nav-tabs a[aria-controls="' + preferredTab + '"]').tab('show');
+        }
+
+        // Make sure we save the user's preferred tab to localStorage.
+        $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
+            storage.setItem(key, e.target.getAttribute("aria-controls"));
+        });
+    }
+
     // Emit a Google Analytics event when the user expands or collapses the Dependencies section.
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -88,6 +88,22 @@ $(function () {
         $(this).closest('form').submit();
     })
 
+    var storage = window['localStorage'];
+    if (storage) {
+		var key = 'preferred_tab';
```
khellang: Went with `var` as opposed to `const` or `let` because I didn't know your supported browser targets and saw it was used elsewhere.

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -88,6 +88,22 @@ $(function () {
...
+		
+        // Restore preferred tab selection from localStorage.
+        var preferredTab = storage.getItem(key);
+        if (preferredTab) {
+            $('.nav-tabs a[href="#' + preferredTab + '"]').tab('show');
+        }
+
+        // Make sure we save the user's preferred tab to localStorage.
+        $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
+            storage.setItem(key, e.target.getAttribute("aria-controls"));
```
khellang: I's not the cleanest solution to use the `aria-controls` attribute here (and the `href` attribute above), but it works without having to also modify `DisplayPackage.cshtml` to add an `id` attribute to the element.

scottbommarito: IMO I'd rather see the change to `DisplayPackage.cshtml` to give the element an `id` (maybe `@packageManager.Id + "-tab"`), primarily because it simplifies the code here. The JQuery call above to show the tab becomes simpler (`$('#' + preferredTab).tab('show');`) and setting the local storage becomes simpler as well (`storage.setItem(key, e.target.id);`).

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -88,6 +88,22 @@ $(function () {
...
+		var key = 'preferred_tab';
+		
+        // Restore preferred tab selection from localStorage.
+        var preferredTab = storage.getItem(key);
+        if (preferredTab) {
+            $('.nav-tabs a[aria-controls="' + preferredTab + '"]').tab('show');
+        }
+
+        // Make sure we save the user's preferred tab to localStorage.
+        $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
```
scottbommarito: I worry that this selector applies to too many elements. If we add any other elements with `data-toggle="tab"` to our layout or elsewhere on this page, they will be saved to local storage as well. I would rather give all of the `CommandTab` `<a>` tags a class (maybe `package-manager-tab`) and then using that in the selector.

khellang:
@scottbommarito Thanks for great feedback! Updated 👍🏻

src/NuGetGallery/Scripts/gallery/page-display-package.js
loic-sharma: Please use spaces here

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -88,6 +88,22 @@ $(function () {
...
     })
 
+    var storage = window['localStorage'];
+    if (storage) {
+		var key = 'preferred_tab';
+		
+        // Restore preferred tab selection from localStorage.
+        var preferredTab = storage.getItem(key);
+        if (preferredTab) {
+            $('#' + preferredTab).tab('show');
```
loic-sharma: Out of curiosity, what happens here if the tab no longer exists?

khellang: Just tested... nothing 😄

scottbommarito: My understanding was that the `$(...)` call returns what is essentially an empty array, and the `.tab` call does nothing because it has no elements to operate on.

Similarly, if the `$(...)` returned multiple elements that match the criteria, the `.tab` call would be applied to all of them.

scottbommarito:
Pulled down the branch locally and it works great. All unit tests pass. Also tried removing tabs and rearranged tabs and the saved tab still works.

PR #7104, opened 2019-04-25
AccountDelete.DeletedBy and PackageDelete.DeletedBy should be set to null when the User is deleted

scottbommarito:
https://github.com/NuGet/Engineering/issues/2358


src/NuGet.Services.Entities/AccountDelete.cs
```diff
@@ -30,7 +30,7 @@ public class AccountDelete
         /// <summary>
         /// The User(admin) key that executed the delete action.
         /// </summary>
-        public int DeletedByKey { get; set; }
+        public int? DeletedByKey { get; set; }
```
loic-sharma: >DeletedByKey [](start = 20, length = 12)

Please add a comment as to why this row may be null. Same thing for `PackageDelete.DeletedByKey`

shishirx34: will be null for self deletes?

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -109,6 +112,7 @@ private async Task DeleteAccountImplAsync(User userToBeDeleted, User userToExecu
             await RemoveSecurityPolicies(userToBeDeleted);
             await RemoveUserCredentials(userToBeDeleted);
             await RemovePackageOwnershipRequests(userToBeDeleted);
+            ResetPackagesAndAccountsDeletedBy(userToBeDeleted);
```
loic-sharma: >ResetPackagesAndAccountsDeletedBy [](start = 12, length = 33)

What would happen if the account deletes a package after `ResetPackagesAndAccountsDeletedBy` is called but before `SaveChangesAsync`?

src/NuGet.Services.Entities/AccountDelete.cs
scottbommarito: Right now this is set to the site admin who executes the user delete. When this flow is automated, it depends on how it is implemented by @ryuyu

scottbommarito: Done.

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: The deletion process happens in a single transaction so it's not possible for a `User` delete a package (or do any action) at any point in the deletion process.

src/NuGetGallery/Services/DeleteAccountService.cs
```diff
@@ -318,6 +322,25 @@ private void RemoveMembers(Organization organization)
...
         }
 
+        private void ResetPackagesAndAccountsDeletedBy(User user)
+        {
+            foreach (var deletedPackage in _packageDeleteRepository
+                .GetAll()
+                .Where(d => d.DeletedByKey == user.Key)
+                .ToList())
+            {
+                deletedPackage.DeletedBy = null;
```
zhhyu: Don't you also clean up "DeleteByKey"?

tests/NuGetGallery.Facts/Services/DeleteAccountServiceFacts.cs
```diff
@@ -141,6 +141,8 @@ public async Task DeleteHappyUser(bool isPackageOrphaned, AccountDeletionOrphanP
                     Assert.NotEmpty(testUser.OrganizationMigrationRequests);
                     Assert.NotEmpty(testUser.OrganizationRequests);
                     Assert.NotEmpty(testUser.Organizations);
+                    Assert.NotNull(testableService.PackageDeletedByUser.DeletedBy);
```
zhhyu: I think you should also check that "DeletedByKey" is set to null or not for all these test cases?

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -356,7 +356,7 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
                 .HasRequired(a => a.DeletedAccount);
 
             modelBuilder.Entity<AccountDelete>()
-                .HasRequired(a => a.DeletedBy)
+                .HasOptional(a => a.DeletedBy)
```
zhhyu: Do we have any test cases to cover this change?

zhhyu:
Should we have any functional tests to cover the whole picture of deleting accounts?

tests/NuGetGallery.Facts/Services/DeleteAccountServiceFacts.cs
```diff
@@ -533,6 +544,8 @@ public DeleteAccountTestService(User user, PackageRegistration userPackagesRegis
                     NewOwner = _user
                 });
 
+                AccountDeletedByUser = new AccountDelete { DeletedBy = _user, DeletedByKey = _user.Key };
```
zhhyu: Maybe it's better that we include some test cases which contain empty "AccountDeletedByUser" or "PackageDeletedByUser" to improve test coverage? It's just a suggestion, not a must.

zhhyu:
Should we have any functional tests/E2E tests to cover the whole picture of deleting accounts including this change?

src/NuGetGallery/Services/DeleteAccountService.cs
scottbommarito: If you set one, EF will set the other automatically when you commit changes.

scottbommarito:
@zhhyu - we have wanted to add functional tests for this, but unfortunately it is difficult to make a test around this primarily because deleting accounts requires being a site admin, and there are difficulties associated with giving a test account site admin.

tests/NuGetGallery.Facts/Services/DeleteAccountServiceFacts.cs
scottbommarito: I'd rather not, for two reasons:
- If one is set to null, the other must be set to the same thing or EF is not working properly
- `DeletedByKey` won't even be set to null because the EF behavior that sets it to null does not execute in unit tests

tests/NuGetGallery.Facts/Services/DeleteAccountServiceFacts.cs
scottbommarito: Good idea. Will add!

PR #7106, opened 2019-04-26
Add support for package deprecation reason 'unusable'

xavierdecoster:


src/NuGet.Services.Entities/PackageDeprecationStatus.cs
```diff
@@ -26,10 +26,9 @@ public enum PackageDeprecationStatus
         /// </summary>
         Legacy = 2,
 
-        // Note: the below is commented out for future reference so we keep track of the original enum value for PackageDeprecationStatus.Vulnerable = 4
         /// <summary>
-        /// The package is deprecated because it is vulnerable.
+        /// The package is deprecated because it has bugs/issues that makes it unusable.
         /// </summary>
-        // Vulnerable = 4
+        Unusable = 4
```
xavierdecoster: I checked our data, nothing is using enum value 4 yet. We can reinstate the `PackageDeprecationStatus.Vulnerable` enum with a value of `8` without issues.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -582,7 +582,12 @@
                                         {
                                             var deprecationTitle = packageVersion.Version;
                                             var isLegacy = packageVersion.DeprecationStatus.HasFlag(PackageDeprecationStatus.Legacy);
-                                            if (isLegacy)
+                                            var isUnusable = packageVersion.DeprecationStatus.HasFlag(PackageDeprecationStatus.Unusable);
+                                            if (isUnusable)
+                                            {
+                                                deprecationTitle += " is deprecated because it has bugs/issues that makes it unusable.";
+                                            }
+                                            else if (isLegacy)
```
xavierdecoster: Have an open question about precedence [here](https://github.com/NuGet/Engineering/issues/2382#issuecomment-487043359)...

src/NuGetGallery/Helpers/PackageHelper.cs
```diff
@@ -191,6 +191,11 @@ public static string GetSelectListText(Package package)
                     deprecationReasons.Add("Legacy");
                 }
 
+                if (deprecation.Status.HasFlag(PackageDeprecationStatus.Unusable))
+                {
+                    deprecationReasons.Add("Unusable");
```
xavierdecoster: Figured `Unusable` was the keyword here... let me know if you prefer an alternative label for this deprecation reason.

src/NuGet.Services.Entities/PackageDeprecationStatus.cs
```diff
@@ -26,10 +26,9 @@ public enum PackageDeprecationStatus
         /// </summary>
         Legacy = 2,
 
-        // Note: the below is commented out for future reference so we keep track of the original enum value for PackageDeprecationStatus.Vulnerable = 4
         /// <summary>
-        /// The package is deprecated because it is vulnerable.
+        /// The package is deprecated because it has critical bugs.
         /// </summary>
-        // Vulnerable = 4
+        CriticalBugs = 4
```
xavierdecoster: I checked our data, nothing is using enum value `4` yet. We can reinstate the `PackageDeprecationStatus.Vulnerable` enum with a value of `8` without issues.

xavierdecoster:
Screenshots using a random package as an example:

image

image

anangaur:
Some proposed changes:
* All versions -> All current versions
* Move alternate version selection down just before the custom text as it’s valid for any of the deprecation reasons

xavierdecoster:
> All versions -> All current versions

@scottbommarito I think this requires changes to the multi-select-dropdown JS, right? I noticed the control is pretty generic, and automatically generates the labels based on a `pluralItemTitle` (in `common-multi-select-dropdown.js`) Suggested approaches for this change?

xavierdecoster:
> Move alternate version selection down just before the custom text as it’s valid for any of the deprecation reasons

@anangaur does this mean it should always be visible?

scottbommarito:
@xavierdecoster 

For `All current versions`, I would add a `allSelectedTitle` parameter to the dropdown constructor and then choose that for the title. This means that we will need to change the JS of the vulnerability stuff when we add it back, but that probably won't be too hard to do.

Regarding the alternate version selection, it is intended to be always visible now.

src/NuGetGallery/Helpers/PackageHelper.cs
```diff
@@ -191,6 +191,11 @@ public static string GetSelectListText(Package package)
                     deprecationReasons.Add("Legacy");
                 }
 
+                if (deprecation.Status.HasFlag(PackageDeprecationStatus.CriticalBugs))
+                {
+                    deprecationReasons.Add("Critical Bugs");
```
shishirx34: Feel like we should have these values defined in a class as const since we are adding more and more in here.

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
```diff
@@ -76,8 +78,8 @@ function ManageDeprecationViewModel(id, versionDeprecationStateDictionary, defau
 
     // The list of options in the alternate package version dropdown.
     this.alternatePackageVersions = ko.pureComputed(function () {
-        // Include an "Any Version" option in case users want to select the package registration.
-        return [strings_AnyVersion].concat(self.alternatePackageVersionsCached());
+        // Include a "Latest" option in case users want to select the package registration.
+        return [strings_Latest].concat(self.alternatePackageVersionsCached());
```
scottbommarito: Should probably rename this string variable to something that describes what the variable is for rather than what the variable contains, e.g. `strings_SelectAlternateVersionOption`, in which case I'd also suggest updating the comments to not have the value (`"Latest"`) as well.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -582,9 +582,21 @@
...
                                             var deprecationTitle = packageVersion.Version;
                                             var isLegacy = packageVersion.DeprecationStatus.HasFlag(PackageDeprecationStatus.Legacy);
-                                            if (isLegacy)
+                                            var hasCriticalBugs = packageVersion.DeprecationStatus.HasFlag(PackageDeprecationStatus.CriticalBugs);
+                                            if (hasCriticalBugs)
                                             {
-                                                deprecationTitle += " is deprecated because it is legacy and is no longer maintained.";
+                                                if (isLegacy)
+                                                {
+                                                    deprecationTitle += " is deprecated because it's legacy. It also has critical bugs.";
```
scottbommarito: I would suggest `it's legacy and has critical bugs`. Adding a new sentence is unnecessary verbage IMO. Also this aligns more with what it looked like before vulnerable was removed.

src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
```diff
@@ -9,10 +9,22 @@
...
                         @{
+                            var hasCriticalBugs = Model.DeprecationStatus.HasFlag(PackageDeprecationStatus.CriticalBugs);
                             var isLegacy = Model.DeprecationStatus.HasFlag(PackageDeprecationStatus.Legacy);
-                            if (isLegacy)
+                            if (hasCriticalBugs)
                             {
-                                @:This package has been deprecated because it is <b>legacy</b> and is no longer maintained.
+                                if (isLegacy)
+                                {
+                                    @:This package has been deprecated because it's <b>legacy</b>. It also has <b>critical bugs</b>.
```
scottbommarito: Right now, we break into two lines: https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml#L50

Confirm with @anangaur that only one line is okay.

Also, if we only have it one line, I'd rather see ` it's <b>legacy</b> and has <b>critical bugs</b>.` as described above.

anangaur:
I am fine with the text rename to fit into one line. 

And yes the alternate package option will always be visible with “optional” text. Refer to the screenshot on the spec.

src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
xavierdecoster: Handled in https://github.com/NuGet/NuGetGallery/pull/7106/commits/3ae7173dffb67556447837d9068d6c13b0392347

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
xavierdecoster: Handled in https://github.com/NuGet/NuGetGallery/pull/7106/commits/3ae7173dffb67556447837d9068d6c13b0392347

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
xavierdecoster: Handled in https://github.com/NuGet/NuGetGallery/pull/7106/commits/aa05cdc825387b2725420a094c388cfe7786cf42

PR #7108, opened 2019-04-29
Custom message should be required if deprecating package due to "Other" reason

scottbommarito:
Part of https://github.com/NuGet/Engineering/issues/2382

image

### DIFF
--- a/src/NuGetGallery/Views/Packages/_ManageDeprecation.cshtml
+++ b/src/NuGetGallery/Views/Packages/_ManageDeprecation.cshtml
@@ -77,6 +77,9 @@
             <label for="customMessage" class="deprecation-section-header">Provide custom message</label>
             <textarea id="customMessage" name="customMessage" class="form-control full-line" data-bind="textInput: customMessage" rows="2"></textarea>
         </div>
+        <div data-bind="visible: requiresCustomMessage()">
+            @ViewHelpers.AlertDanger(@<text>You must add a custom message if you are deprecating a package due to another reason!</text>)
+        </div>
     </div>
 
     <div data-bind="visible: submitError">
@@ -84,6 +87,6 @@
     </div>
 
     <div class="deprecation-section form-group">
-        <button class="btn btn-primary full-line" type="submit" aria-label="Save package deprecation information" data-bind="click: submit">Save</button>
+        <button class="btn btn-primary full-line" type="submit" aria-label="Save package deprecation information" data-bind="click: submit, disable: requiresCustomMessage">Save</button>
     </div>
 </script>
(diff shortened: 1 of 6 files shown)


scottbommarito:
@anangaur - to clarify, do you still want custom message required if "other" is selected but legacy/bugs is also selected?

anangaur:
Anytime `Other` is selected, the custom message should become mandatory IMO. In other cases, the header itself should have **`Provide custom message `**`(optional)`. On selection of `Other`, it should become: **`Provide custom message `**

src/NuGetGallery/Views/Packages/_ManageDeprecation.cshtml
```diff
@@ -77,13 +77,16 @@
             <label for="customMessage" class="deprecation-section-header">Provide custom message</label>
             <textarea id="customMessage" name="customMessage" class="form-control full-line" data-bind="textInput: customMessage" rows="2"></textarea>
         </div>
+        <div data-bind="visible: requiresCustomMessage()">
+            @ViewHelpers.AlertDanger(@<text>You must add a custom message if you are deprecating a package due to another reason!</text>)
```
loic-sharma: >You must add a custom message if you are deprecating a package due to another reason! [](start = 44, length = 85)

What do you think of the message:

`A custom message is required if <b>other</b> is selected`

It makes it clearer which field added this requirement.

tests/NuGetGallery.Facts/Controllers/ManageDeprecationJsonApiControllerFacts.cs
```diff
@@ -108,6 +108,20 @@ public void ReturnsAllAvailableVersionsInReverseVersionOrder()
...
         {
+            [Fact]
+            public async Task ReturnsBadRequestIfOtherAndNoCustomMessage()
+            {
+                // Arrange
+                var controller = GetController<ManageDeprecationJsonApiController>();
+
+                // Act
+                var result = await controller.Deprecate(
+                    "id", null, false, false, true, null, null, null);
```
loic-sharma: You can use named properties here, right? If so, please add those in here :)

src/NuGetGallery/Views/Packages/_ManageDeprecation.cshtml
scottbommarito: agreed.

scottbommarito:
@anangaur - Agreed with requiring it whenever `other` is selected. I don't think `(optional)` is necessary...the message below when `other` is selected is pretty explicit

tests/NuGetGallery.Facts/Controllers/ManageDeprecationJsonApiControllerFacts.cs
scottbommarito: will do--i think this will improve test resiliency

anangaur:
LGTM. Can the “other” option be just “Other”? I guess that’s much ore easier to read.

PR #7111, opened 2019-04-29
Add PackageDeprecation.LastEdited column and db trigger

xavierdecoster:


joelverhagen:
Why do we do it this way? Isn't doing it from C# enough?

scottbommarito:
Although we discussed having this column, I still don't see how it's necessary given what we've decided on for the MVP--especially for deprecation and not vulnerability.

We should finalize our design and then have a meeting with the team before we proceed with this PR.

xavierdecoster:
> Why do we do it this way? Isn't doing it from C# enough?

Same approach as for `Package.LastEdited`.

@scottbommarito Not needed for MVP indeed, but very likely needed for the API. I had the code already, but happy to park the code and this PR for now until the API requirements have been spec-ed and the design finalized. Can reopen when needed.

joelverhagen:
> Same approach as for `Package.LastEdited`.

We added this to improve the chance that package edits/reflows/transitions to available occur in a sequential order. Do we have this need here?

PR #7118, opened 2019-04-30
Add auditing for package (un)deprecate actions

xavierdecoster:


src/NuGetGallery.Core/Auditing/PackageAuditRecord.cs
```diff
@@ -14,12 +15,15 @@ public class PackageAuditRecord : AuditRecord<AuditedPackageAction>
 
         public AuditedPackage PackageRecord { get; }
         public AuditedPackageRegistration RegistrationRecord { get; }
+        public AuditedPackageDeprecation[] DeprecationRecords { get; }
```
scottbommarito: I think just having one `AuditedPackageDeprecation` for now is sufficient.
The only reason the relationship is 1-many right now was because it gives us more flexibility in the future in case the relationship changes, but we don't have this concern with auditing because we can very easily change the structure with no repercussions.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -157,7 +161,18 @@ public virtual JsonResult GetAlternatePackageVersions(string id)
                 status,
                 alternatePackageRegistration,
                 alternatePackage,
-                customMessage);
+                customMessage,
+                currentUser);
+
+            foreach (var packageToUpdate in packagesToUpdate)
+            {
+                await _auditingService.SaveAuditRecordAsync(
```
scottbommarito: I noticed we aren't verifying this is called in the unit tests--can we add it?

xavierdecoster: Handled in https://github.com/NuGet/NuGetGallery/pull/7118/commits/042a7cbaefa3bb95f162faeed8fe264914e661d2

src/NuGetGallery.Core/Auditing/PackageAuditRecord.cs
xavierdecoster: Handled in https://github.com/NuGet/NuGetGallery/pull/7118/commits/042a7cbaefa3bb95f162faeed8fe264914e661d2

PR #7124, opened 2019-05-02
Add development dependency metadata to the database

### LINKED ISSUE #5666: Persist development dependency information in the DB
This is useful for the `<PackageReference>` tab on the package details since the code snippet is has some additional properties when the package is a development dependency.

See: https://github.com/NuGet/NuGetGallery/pull/5656

To roll this out we would need to turn it on for existing packages then perform some kind of data fix for the existing development dependencies.

Note that this data is in the catalog leaves.

khellang:
This is what it looks like after I ran the migration and update tool:

image

Closes #5666


src/GalleryTools/Commands/BackfillDevelopmentDependencyCommand.cs
```diff
@@ -0,0 +1,518 @@
...
+    /// 1. To collect development dependency metadata:
+    ///     a. Configure app.config with DB information and service index URL
+    ///     b. Run this tool with: GalleryTools.exe filldevdeps -c
+    /// This will create a file developmentDependencyMetadata.txt with all collected data. You can stop the job anytime and restart. cursor.txt contains current position.    
+    /// 
+    /// 2. To update DB:
+    ///     a. Run GalleryTools.exe filldevdeps -u  
+    /// This will update DB from file developmentDependencyMetadata.txt. You can stop the job anytime and restart.
+    /// </summary>
+    public class BackfillDevelopmentDependencyCommand
```
khellang: This is just an adapted version of [`BackfillRepositoryMetadataCommand`](https://github.com/NuGet/NuGetGallery/blob/0ca0368afbfc1bbf7133e2d72389fc168d35f6ce/src/GalleryTools/Commands/BackfillRepositoryMetadataCommand.cs).

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
```diff
@@ -105,7 +105,7 @@ public VerifyPackageRequest(PackageMetadata packageMetadata, IEnumerable<User> p
         public string Copyright { get; set; }
         public string Description { get; set; }
         public DependencySetsViewModel Dependencies { get; set; }
-        public string DevelopmentDependency { get; set; }
+        public bool DevelopmentDependency { get; set; }
```
khellang: I'm not 100% sure this is a safe change to make, but it looks like it's only used for display purposes so it shouldn't really matter.

src/Bootstrap/dist/css/bootstrap-theme.css
```diff
@@ -977,9 +977,6 @@ img.reserved-indicator-icon {
   font-weight: 600;
   text-decoration: underline;
 }
-.page-package-details .install-tabs .tab-content {
-  height: 79px;
```
khellang: This hardcoded height didn't work well with the additional lines for dev dependencies. It made the content slide under the "Dependencies" expander below.

Does anyone know why it was there in the first place?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -125,11 +131,7 @@
 
     <div role="tabpanel" class="tab-pane @(active ? "active" : string.Empty)" id="@packageManager.Id">
         <div>
-            <div class="install-script" id="@packageManager.Id-text">
-                <span>
-                    @packageManager.InstallPackageCommand
-                </span>
-            </div>
+            <pre class="install-script" id="@packageManager.Id-text">@packageManager.InstallPackageCommand</pre>
```
khellang: I had to change this to a `pre` in order to properly preserve the whitespace (tried all kinds of CSS combos to no avail), which might've slightly changed the formatting. I think the font size might be a tad bigger now, but I think it looks great overall.

khellang: Pushed an update to make sure the font-size matches what's currently live in the gallery.

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
scottbommarito: This affects the website upload flow. This request model is used when the UI shows you the metadata of the package you've uploaded before you confirm and submit the package. Did you check that the website upload flow still works? Otherwise you'll probably need to make some minor UI changes.

src/GalleryTools/Commands/BackfillDevelopmentDependencyCommand.cs
```diff
@@ -0,0 +1,523 @@
...
```
scottbommarito: Do you think you could share the logic for these two commands so that they aren't duplicated? I see a lot of classes in this file and most of them appear to be very similar to the same classes in `BackfillRepositoryMetadataCommand`. Doing this would reduce the maintenance cost of this code and it would make it very easy for someone to create a new "backfill" command.

khellang: Yeah, the whole copy-paste thing felt kinda gross, but at the same time, I figured it's for a one-off task (hopefully :wink) so it wasn't really that important. If you still feel it's worth it, I have no problem refactoring it to share more code between the two commands.

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
khellang: Yeah, I tested that the `_VerifyMetadata.cshtml` partial (which is part of `_VerifyForm.cshtml` and `UploadPackage.cshtml
`) works as intended. The code is the same as `RequiresLicenseAcceptance`, which is also a `bool`:

https://github.com/NuGet/NuGetGallery/blob/1eff45eac940922495af949ea30fda4daee6045f/src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml#L153-L156

https://github.com/NuGet/NuGetGallery/blob/1eff45eac940922495af949ea30fda4daee6045f/src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml#L235-L238

Is there anywhere else I need to test?

khellang:
Also, just a side note... The whole `PrivateAssets`/`IncludeAssets` thing is quite the mouthful. What would it take to "teach" `PackageReference` about `DevelopmentDependency`? So that

```xml
<PackageReference Include="Nerdbank.GitVersioning" Version="2.3.138">
  <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
</PackageReference>
```

becomes just

```xml
<PackageReference Include="Nerdbank.GitVersioning" Version="2.3.138" DevelopmentDependency="True" />
```

Just an idea...

scottbommarito:
@khellang - I agree that it's quite the mouthful, but in terms of why it was implemented like this, that would be a question for the client team. https://github.com/NuGet/Home would be the place to ask.

src/NuGetGallery/RequestModels/VerifyPackageRequest.cs
scottbommarito: Yeah I think that's good then!

src/GalleryTools/Commands/BackfillDevelopmentDependencyCommand.cs
scottbommarito: I think at this point, because we've done it twice so far, it's pretty likely that we will need to do it again, in which case I think it's worth doing the effort.

src/Bootstrap/dist/css/bootstrap-theme.css
joelverhagen: My guess is it was a hack to make it look consistent across browsers. Could you try IE, Firefox, Chrome, your phone browser, and Safari on OS X if possible to make sure things look okay? Also make sure non-maximized window works alright and "zoom in" on your desktop browser.

---
In reply to: [280319928](https://github.com/NuGet/NuGetGallery/pull/7124#discussion_r280319928) [](ancestors = 280319928)

joelverhagen:
>             public async Task WillReadThePropertiesFromTheNuGetPackageWhenCreatingANewPackage()

There should be a test that covers when there is a) no dev dependency node b) false dev dependency.

Another thing, from this line in client it looks like a string that's not parseable as bool will throw:
[https://github.com/NuGet/NuGet.Client/blob/f7e4a189c85fc6cfcc2d3f6b3d94079125adfe91/src/NuGet.Core/NuGet.Packaging/Core/NuspecCoreReaderBase.cs#L126](https://github.com/NuGet/NuGet.Client/blob/f7e4a189c85fc6cfcc2d3f6b3d94079125adfe91/src/NuGet.Core/NuGet.Packaging/Core/NuspecCoreReaderBase.cs#L126)

This means client should reject such a package. Please verify the client behavior for non-parseable bools and update validation accordingly. 

---
Refers to: tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs:200 in 45b0783. [](commit_id = 45b0783ed2387445165d559703102d677c350d05, deletion_comment = False)

src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -0,0 +1,461 @@
...
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml;
+using System.Xml.Linq;
+
+namespace GalleryTools.Commands
+{
+    public abstract class BackfillCommand<TMetadata>
```
joelverhagen: >BackfillCommand [](start = 26, length = 15)

Nice one!

joelverhagen:
Could you update the readme with instructions on how to use this tool?
[https://github.com/NuGet/NuGetGallery/blob/master/src/GalleryTools/Readme.md](https://github.com/NuGet/NuGetGallery/blob/master/src/GalleryTools/Readme.md)

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
joelverhagen: As long as you test this across various browsers, viewport sizes, and zooms, this looks like better markup.

joelverhagen:
> What would it take to "teach" `PackageReference` about `DevelopmentDependency`?

> @khellang - I agree that it's quite the mouthful, but in terms of why it was implemented like this, that would be a question for the client team. https://github.com/NuGet/Home would be the place to ask.

Given it was not originally implemented this way, we would have two sets of client versions to worry about. One that supports the concise property, one that does not. As mentioned by @scottbommarito, perhaps you can open an issue against the client team but we would still need to think through client versioning w.r.t. what we show on nuget.org.

joelverhagen:
Nice work @khellang! Just a couple small comment. Also, can you rebase/merge latest `dev`? There are some merge conflicts showing up.

src/Bootstrap/dist/css/bootstrap-theme.css
joelverhagen: Oh, and make sure long package IDs and versions that require wrapping still look okay.

khellang: So, here are the results of my testing:

# Galaxy S7

GS7

# iPhone 6S

IP6S

# Safari 12 - MacOS

MacOS - Safari 12 - Large
MacOS - Safari 12 - Small

# Chrome 74 - Win7

Win7 - Chrome 74 - Large
Win7 - Chrome 74 - Small

# Firefox 66 - Win7

Win7 - FF66 - Large
Win7 - FF66 - Small

# IE11 - Win7

Win7 - IE11 - Large
Win7 - IE11 - Small

# Edge 18 - Win10

Win10 - Edge 18 - Large

Most of the browsers was tested on 1920x1080 and 1024x768 resolution. IMO they all look decent. It would be very hard to get the wrapping perfect. The only issue I can see is the "copy to clipboard" button not growing vertically in some browsers (Safari (both mobile and desktop) and Firefox). I'll look into that.

khellang: Managed to tweak the copy button so it looks a bit better on Firefox. Still not 100% on Safari, but I think it's good enough:

## iOS

image

## MacOS

image

## Firefox 66 - Win10

image

khellang: Was finally able to make the copy button fill the entire vertical space in all browsers:

# iOS

image

# Chrome

image

# Firefox

image

# Edge 

image

# IE 11

image

# Safari (Desktop)

image

# Galaxy S7

image

khellang:
> Another thing, from this line in client it looks like a string that's not parseable as bool will throw:
> [NuGet/NuGet.Client:src/NuGet.Core/NuGet.Packaging/Core/NuspecCoreReaderBase.cs@`f7e4a18`#L126](https://github.com/NuGet/NuGet.Client/blob/f7e4a189c85fc6cfcc2d3f6b3d94079125adfe91/src/NuGet.Core/NuGet.Packaging/Core/NuspecCoreReaderBase.cs#L126)
> 
> This means client should reject such a package. Please verify the client behavior for non-parseable bools and update validation accordingly.

@joelverhagen I'm getting

> The package manifest contains an invalid boolean value for metadata element: 'developmentDependency'. The value should be 'true' or 'false'.

So I'm assuming everything is validated either `true`, `false` or missing entirely, which would default to `false`.

khellang:
Filed https://github.com/NuGet/Home/issues/8098 for tracking `DevelopmentDependency="True"` on `PackageReference`.

khellang:
I think all current feedback has been addressed if you want to do another pass 😄

src/Bootstrap/dist/css/bootstrap-theme.css
khellang: The only browsers I can't get it to show properly now is IE8-10, but I'm not sure those are supported?

image

khellang: And just for the hell of it, here's IE8 on XP:

image

Not bad, eh? 😂

joelverhagen: The IE10 look is 100% acceptable given that it is not the latest IE. In fact, it still quite usable. GREAT work on your analysis. I'm quite comfortable moving forward with this.

joelverhagen:
> @joelverhagen I'm getting
> 
> > The package manifest contains an invalid boolean value for metadata element: 'developmentDependency'. The value should be 'true' or 'false'.
> 
> So I'm assuming everything is validated either `true`, `false` or missing entirely, which would default to `false`.

Okay, awesome! This exactly what we need.

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -253,6 +254,26 @@ public async Task WillReadTheLanguagePropertyFromThePackage()
                 Assert.Equal("fr", package.Language);
             }
 
+            [Theory]
+            [InlineData(true, true)]
+            [InlineData(false, false)]
+            [InlineData(null, false)]
+            public async Task WillReadDevelopmentDependencyFromPackage(bool? developmentDependency, bool expected)
+            {
```
joelverhagen: There should be a unit test that verifies that an invalid bool is rejected.

joelverhagen:
Needs one last test for verify bad dev dependency bools are rejected, then I'll run the CI and merge this to `dev`.

src/Bootstrap/dist/css/bootstrap-theme.css
scottbommarito: To be honest, this is more analysis than I think we've ever done for a change like this. Phenomenal work!

khellang: I happen to have a BrowserStack subscription for testing our sites in different browsers on different devices. Makes it a breeze to test changes like these, even from your local machine 😅👍

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
khellang: Added a test that should verify behavior for invalid `developmentDependency` nodes below 👍

joelverhagen: Thanks! I'll move forward with getting this rolled out. My current plan is:

1. Run the DB migration
1. Update gallery so new packages have this bool set
1. Run the back-fill too.

joelverhagen:
@khellang, looks like there is a build issue:

```
Services\PackageServiceFacts.cs(290,58): error CS0246: The type or namespace name 'PackagingException' could not be found (are you missing a using directive or an assembly reference?) [C:\vsts-agent\_work\605\s\tests\NuGetGallery.Facts\NuGetGallery.Facts.csproj]
```

khellang:
@joelverhagen Ah, sorry about that. Try now 😉

Issue #7080, opened 2019-04-16
NuGet.org asks to re-login when browser is closed and re-opened


anangaur:
Repro:

1. Login to nuget.org
2. Close the Browser.
3. Open the broser and navigate to nuget.org

**Expected**: It should just be signed in already to my account

**Actual**: It asks to sign in again.

Note that if you close the tab and keep the browser running, the experience is as expected. May be the cookie expiration is set to renew for each browser session?

shishirx34:
is your browser setting to destroy cookies on close perhaps? I'll need to investigate if some other settings need to be applied for cookie persistence, we do have expiration of 6 hours.

anangaur:
Don’t remember such setting as it just works fine for GitHub

skofman1:
https://stackoverflow.com/questions/31946582/how-ispersistent-works-in-owin-cookie-authentication/46659752#46659752

shishirx34:
assigning @scottbommarito to verify the work in INT and PROD

weitzhandler:
I'm also experiencing this issue.
NuGet is the website that I'll always see myself as logged out when revisiting. Pretty annoying. It should keep my logged in for at least some weeks of idle time before disconnecting me.
And no, my browser doesn't delete its cookies upon closing. I think the issue happens even not closing browser.

PR #7127, opened 2019-05-02
[NuGet Gallery Login] Add isPersistent flag to cookie

shishirx34:
Addresses #7080 - This will persist the session cookie even after the browser is closed, until the session expires. Verified with local sessions.


anangaur:
Verified it works now on dev

Issue #7140, opened 2019-05-07
Current entity model does not match migrations


joelverhagen:
In `dev`, the following migration gets created if you do `Add-Migration`:

```csharp
public override void Up()
{
    DropForeignKey("dbo.PackageDeprecationCves", "PackageDeprecation_Key", "dbo.PackageDeprecations");
    DropForeignKey("dbo.PackageDeprecationCves", "Cve_Key", "dbo.Cves");
    DropForeignKey("dbo.PackageDeprecationCwes", "PackageDeprecation_Key", "dbo.PackageDeprecations");
    DropForeignKey("dbo.PackageDeprecationCwes", "Cwe_Key", "dbo.Cwes");
    DropIndex("dbo.Cves", new[] { "CveId" });
    DropIndex("dbo.Cwes", new[] { "CweId" });
    DropIndex("dbo.PackageDeprecationCves", new[] { "PackageDeprecation_Key" });
    DropIndex("dbo.PackageDeprecationCves", new[] { "Cve_Key" });
    DropIndex("dbo.PackageDeprecationCwes", new[] { "PackageDeprecation_Key" });
    DropIndex("dbo.PackageDeprecationCwes", new[] { "Cwe_Key" });
    DropColumn("dbo.PackageDeprecations", "CvssRating");
    DropTable("dbo.Cves");
    DropTable("dbo.Cwes");
    DropTable("dbo.PackageDeprecationCves");
    DropTable("dbo.PackageDeprecationCwes");
}
```

/cc @scottbommarito @xavierdecoster 

This would have been caught with https://github.com/NuGet/NuGetGallery/issues/4934.

joelverhagen:
I think this may be due to https://github.com/NuGet/NuGetGallery/pull/7124. Investigating.

PR #7141, opened 2019-05-07
Recreate development dependency migration on latest EF model

joelverhagen:
Addresses https://github.com/NuGet/NuGetGallery/issues/7140.
Related to https://github.com/NuGet/NuGetGallery/issues/5666.

The `DevelopmentDependency` migration was based on the EF model prior to https://github.com/NuGet/NuGetGallery/blob/dev/src/NuGetGallery/Migrations/201904292139301_RemoveVulnerabilityDataFromPackageDeprecationModel.cs.

It needs to be recreated because the EDMX in the migration I deleted still has the removed vulnerability models.

/cc @khellang - this was simply due to two people add a migration in parallel. You made yours first but it was checked in after someone else had already made another migration. No harm, just wanted to let you know as it relates to your PR. This PR will bring everything to a good state.


khellang:
Ugh, I'd forgotten how bad it is to work with EF migrations in a team. Anyway, thanks for fixing my :shit: @joelverhagen 😊

Issue #4934, opened 2017-10-31
Add unit test to check for pending schema changes


chenriksson:
Suggested by @joelverhagen to keep migrations and schema changes in sync

Should be possible to programmatically create migration and check. See:
- https://github.com/aspnet/EntityFrameworkCore/blob/dev/test/ef.Tests/SimpleProjectTest.cs
- https://github.com/aspnet/EntityFrameworkCore/blob/dev/src/ef/Commands/MigrationsAddCommand.cs

skofman1:
@chenriksson , which DB do we want to test this on?

chenriksson:
@skofman1 This isn't testing migrations against a DB instance. The test would check if EF would detect schema changes if a new migration were added... comparing the source code against the latest EF migration snapshot.

Will keep, but put at P3

joelverhagen:
Done on gallery DB, support requests DB, and validation DB.

PR #7143, opened 2019-05-07
Add pending migration test

joelverhagen:
This adds a unit test that fails if there are any pending changes in the entity model. It asserts that if someone was to run `Add-Migration`, an empty migration would be added (indicating the latest migration matches the entity model).

Found with https://github.com/NuGet/NuGetGallery/issues/7140.
Addresses https://github.com/NuGet/NuGetGallery/issues/4934.

### DIFF
--- a/src/DatabaseMigrationTools/MigrationTargetDatabaseArgumentNames.cs
+++ b/src/DatabaseMigrationTools/MigrationTargetDatabaseArgumentNames.cs
@@ -3,7 +3,7 @@
 
 namespace NuGetGallery.DatabaseMigrationTools
 {
-    internal static class MigrationTargetDatabaseArgumentNames
+    public static class MigrationTargetDatabaseArgumentNames
     {
         public const string GalleryDatabase = "GalleryDatabase";
         public const string SupportRequestDatabase = "SupportRequestDatabase";
--- a/tests/NuGet.Services.DatabaseMigration.Facts/NuGet.Services.DatabaseMigration.Facts.csproj
+++ b/tests/NuGet.Services.DatabaseMigration.Facts/NuGet.Services.DatabaseMigration.Facts.csproj
@@ -44,9 +44,13 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="DatabaseMigrationFacts.cs" />
+    <Compile Include="PendingMigrationsFacts.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Moq">
+      <Version>4.8.2</Version>
+    </PackageReference>
     <PackageReference Include="xunit">
       <Version>2.4.1</Version>
     </PackageReference>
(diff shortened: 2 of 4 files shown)


tests/NuGetGallery.Facts/Migrations/PendingMigrationsFacts.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        [Fact]
+        public void NoPendingMigrations()
+        {
+            var initialMigrationType = typeof(Initial);
+            var connectionStringInfo = new DbConnectionInfo(
+                $@"Data Source=(localdb)\mssqllocaldb; Initial Catalog={_databaseName}; Integrated Security=True; MultipleActiveResultSets=True",
+                "System.Data.SqlClient");
+            var migrationName = $"TestMigration{DateTimeOffset.UtcNow:yyyyMMddHHmmssFFFFFFF}";
```
joelverhagen: This should be using `_currentTimestamp`.

tests/NuGetGallery.Facts/Migrations/PendingMigrationsFacts.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            var result = migrationScaffolder.Scaffold(migrationName);
+
+            _output.WriteLine("Migration content:");
+            _output.WriteLine(new string('-', 60));
+            _output.WriteLine(result.UserCode);
+            _output.WriteLine(new string('-', 60));
+
+            Assert.Equal(
+                $@"namespace NuGetGallery.Migrations
```
agr: That looks fragile.

tests/NuGetGallery.Facts/Migrations/PendingMigrationsFacts.cs
```diff
@@ -0,0 +1,97 @@
...
+            var migrationName = $"TestMigration{DateTimeOffset.UtcNow:yyyyMMddHHmmssFFFFFFF}";
+
+            var migrationsConfiguration = new DbMigrationsConfiguration
+            {
+                ContextType = typeof(EntitiesContext),
+                MigrationsAssembly = initialMigrationType.Assembly,
+                MigrationsNamespace = initialMigrationType.Namespace,
+                TargetDatabase = connectionStringInfo,
+            };
+
```
zhhyu: You could directly depend on the "MigrationsConfiguration.cs" and set the connection string through the factory like mine.

zhhyu: You can also try to directly depend on the "MigrationsConfiguration.cs" and use the factory to set the connection string like mine.

tests/NuGetGallery.Facts/Migrations/PendingMigrationsFacts.cs
zhhyu: If we could get the explicit SQL which will be put on the database side similar like using "MigratorScriptingDecorator", it will may be an empty string. But I am not sure whether we can achieve that.

joelverhagen: @agr, this should be consistent per EntityFramework version.

tests/NuGetGallery.Facts/Migrations/PendingMigrationsFacts.cs
joelverhagen: This is what specifies `NuGetGallery.Migrations` for the generated migration code. I'll try using the factory.

joelverhagen: I am not going to use the factory since it depends on a static. I don't need that in this case which allows tests to run in parallel.

joelverhagen: I decided to use the factory after all, which allowed me to easily test support requests DB as well.

PR #7145, opened 2019-05-08
Remove the old search.

cristinamanum:
There are two changes in this PR:
1. Remove the code related with the previous search functionality.
2. Make the Search http client timeout configurable in web.config.


src/NuGetGallery/Configuration/AppConfiguration.cs
```diff
@@ -87,16 +87,6 @@ public class AppConfiguration : IAppConfiguration
         /// </summary>
         public Uri ServiceDiscoveryUri { get; set; }
```
cristinamanum: Still needed for LuceneController

src/NuGetGallery/Web.config
```diff
@@ -129,18 +129,16 @@
     <!-- Set this value to use a remote search service. This overrides ALL other Lucene-related settings and disables indexing jobs. -->
     <!-- Example:
         <add key="Gallery.ServiceDiscoveryUri" value="https://api.nuget.org/v3/index.json" />
-        <add key="Gallery.SearchServiceResourceType" value="SearchGalleryQueryService/3.0.0-rc" />
-        <add key="Gallery.AutocompleteServiceResourceType" value="SearchAutocompleteService/3.0.0-rc" />
         -->
     <add key="Gallery.ServiceDiscoveryUri" value=""/>
```
cristinamanum: Still needed for LuceneController.

scottbommarito: Same as above...try to remove if possible.

src/NuGetGallery/Configuration/AppConfiguration.cs
cristinamanum: It is used only for Admin page. I would not cut this feature in this change. Talked offline with @joelverhagen and this page is needed to support custom deployments of Gallery.

src/NuGetGallery/Web.config
cristinamanum: Removed - please see above.

src/NuGetGallery/Web.config
```diff
@@ -126,21 +126,15 @@
...
-    <!-- Set this value to use a remote search service. This overrides ALL other Lucene-related settings and disables indexing jobs. -->
-    <!-- Example:
-        <add key="Gallery.ServiceDiscoveryUri" value="https://api.nuget.org/v3/index.json" />
-        <add key="Gallery.SearchServiceResourceType" value="SearchGalleryQueryService/3.0.0-rc" />
-        <add key="Gallery.AutocompleteServiceResourceType" value="SearchAutocompleteService/3.0.0-rc" />
-        -->
-    <add key="Gallery.ServiceDiscoveryUri" value=""/>
-    <add key="Gallery.SearchServiceResourceType" value=""/>
-    <add key="Gallery.AutocompleteServiceResourceType" value=""/>
+    <!-- Set this values to use a remote search service. This overrides ALL other Lucene-related settings and disables indexing jobs. -->
```
skofman1: >this [](start = 13, length = 4)

these

src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
```diff
@@ -20,22 +18,15 @@ namespace NuGetGallery.Infrastructure.Search
...
         private readonly ISearchClient _searchClient;
-        private readonly IFeatureFlagService _featureFlagService;
 
         private JObject _diagCache;
 
-        public Uri ServiceUri { get; private set; }
-
         protected IDiagnosticsSource Trace { get; private set; }
 
         public string IndexPath
```
skofman1: >IndexPath [](start = 22, length = 9)

why is this still needed?

src/NuGetGallery/Services/StatusService.cs
```diff
@@ -43,20 +43,17 @@ public async Task<ActionResult> GetStatus()
         {
             bool sqlAzureAvailable =  IsSqlAzureAvailable();
             bool? storageAvailable = await IsAzureStorageAvailable();
-            bool? searchServiceAvailable = await IsSearchServiceAvailable();
 
             bool galleryServiceAvailable =
                 sqlAzureAvailable
-                && (!storageAvailable.HasValue || storageAvailable.Value) // null == true for this condition.
-                && (!searchServiceAvailable.HasValue || searchServiceAvailable.Value);
+                && (!storageAvailable.HasValue || storageAvailable.Value); // null == true for this condition.
```
skofman1: with the current behavior, when primary blob storage is down we will failover to read-only mode because index.json load will fail (this hides this bug: [https://github.com/NuGet/Engineering/issues/1980](https://github.com/NuGet/Engineering/issues/1980)). 
When u remove the index.json load from the code here, the bug will be exposed. 
We need to fix the bug before merging this code.

src/NuGetGallery/Web.config
cristinamanum: Updated.

src/NuGetGallery/Infrastructure/Lucene/ExternalSearchService.cs
cristinamanum: Is needed for the IIndexingService.

src/NuGetGallery/Services/StatusService.cs
cristinamanum: Fixed the https://github.com/NuGet/Engineering/issues/1980 issue

skofman1:
>         bool AdminPanelDatabaseAccessEnabled { get; set; }

this is from Peter's changes.. 

---
Refers to: src/NuGetGallery/Configuration/IAppConfiguration.cs:101 in be02a22. [](commit_id = be02a222d8c7ed728ffe94c247d0e44e45afb6f4, deletion_comment = False)

Issue #7147, opened 2019-05-08
Gallery corrupts packages in local file system when they contain license files


agr:
If gallery is set up to store packages in local file system it corrupts packages that contain license files in them.

zhhyu:
What will be the behavior of Gallery for this case?

PR #7148, opened 2019-05-08
Fix for the package corruption when license file is extracted by Gallery

agr:
Addresses #7147.

The issue was that the `_coreLicenseFileService.ExtractAndSaveLicenseFileAsync` call in `PackageUploadService` left the package stream with `Position` pointing to the middle of the stream, so `_packageFileService.SavePackageFileAsync` saved the package starting from that position.

### DIFF
--- a/src/NuGetGallery/Services/PackageUploadService.cs
+++ b/src/NuGetGallery/Services/PackageUploadService.cs
@@ -692,6 +692,21 @@ namespace NuGetGallery
 
         public async Task<PackageCommitResult> CommitPackageAsync(Package package, Stream packageFile)
         {
+            if (package == null)
+            {
+                throw new ArgumentNullException(nameof(package));
+            }
+
+            if (packageFile == null)
+            {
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -762,6 +777,7 @@ public async Task<PackageCommitResult> CommitPackageAsync(Package package, Strea
                     }
                     try
                     {
+                        packageFile.Seek(0, SeekOrigin.Begin);
```
cristinamanum: Are there any current corrupted package files ?

agr: It only happens if async validation is disabled, which is not the case for us. So, I do not expect to have any files corrupted because of this issue.

joelverhagen:
Is this blocked on anything? I think the sooner we get this in, the better it is for people standing up their own gallery.

PR #7150, opened 2019-05-09
[Database Migration] Fix CI and Enhance protection. 

zhhyu:
Need to generate the artifacts for different target databases in the build script.
Wait for the logger to send telemetry.


src/DatabaseMigrationTools/Program.cs
```diff
@@ -13,6 +14,8 @@ static void Main(string[] args)
             var migrationContextFactory = new MigrationContextFactory();
             var job = new Job(migrationContextFactory);
             JobRunner.RunOnce(job, args).GetAwaiter().GetResult();
+
+            Thread.Sleep(60000); // wait for the logger
```
joelverhagen: This shouldn't be necessary, due to:
https://github.com/NuGet/NuGet.Jobs/blob/8ead6ced512a1ac88f1be5e883d31b5381ddd41f/src/NuGet.Jobs.Common/JobRunner.cs#L165

joelverhagen: To be clear, I meant to suggest this code as an investigation step, not for check-in.

zhhyu: The Flush() method is not reliable based on the following issues:
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/281
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/407
https://stackoverflow.com/questions/53768107/how-to-flush-applicationinsightstracelistener-in-a-console-application-on-ex
We need to consider updating NuGet.jobs.common for this usage.

zhhyu:
Find something interesting.

src/DatabaseMigrationTools/Program.cs
joelverhagen: Gotcha, thanks for the context.

src/DatabaseMigrationTools/Scripts/PostDeploy.ps1
```diff
@@ -1 +1,2 @@
-.\RunJob.cmd
\ No newline at end of file
+.\RunJob.cmd
+if ($LastExitCode -ne 0) {throw "Database Migration Job Fails"}
```
joelverhagen: Nit: Put the `throw` on its own line

`Failed` instead of `Fails` (past tense is more appropriate here)

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -32,20 +34,39 @@ public Job(IMigrationContextFactory migrationContextFactory)
 
         public override void Init(IServiceContainer serviceContainer, IDictionary<string, string> jobArgsDictionary)
         {
-            base.Init(serviceContainer, jobArgsDictionary);
-            _migrationTargetDatabase = JobConfigurationManager.GetArgument(jobArgsDictionary, MigrationTargetDatabaseArgument);
+            try
+            {
+                base.Init(serviceContainer, jobArgsDictionary);
+                _migrationTargetDatabase = JobConfigurationManager.GetArgument(jobArgsDictionary, MigrationTargetDatabaseArgument);
+            } catch (Exception)
```
joelverhagen: `catch` on a new line

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -91,6 +125,48 @@ private void ExecuteDatabaseMigration(Func<DbMigrator> getMigrator, SqlConnectio
...
         }
 
+        private void CheckIsValidMigration(DbMigrator migrator)
+        {
+            var databaseMigrations = migrator.GetDatabaseMigrations().ToList();
+            databaseMigrations.Reverse();
+            var localMigrations = migrator.GetLocalMigrations().ToList();
+            if (databaseMigrations.Count == 0)
+            {
+                throw new Exception("Migration validation failed: Unexpected empty history of database migrations");
```
joelverhagen: Don't use `Exception`. It makes the catch sites too generic. `InvalidOperationException` is an alternative.

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -91,6 +125,48 @@ private void ExecuteDatabaseMigration(Func<DbMigrator> getMigrator, SqlConnectio
...
+            }
+            if (localMigrations.Count == 0)
+            {
+                throw new Exception("Migration validation failed: Unexpected empty history of local migrations");
```
joelverhagen: Same. Don't use `Exception`.

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -91,6 +125,48 @@ private void ExecuteDatabaseMigration(Func<DbMigrator> getMigrator, SqlConnectio
...
+                if (_migrationTargetDatabase.Equals(MigrationTargetDatabaseArgumentNames.GalleryDatabase) &&
+                    databaseMigrations[databaseMigrationsCursor].Equals(SkipGalleryDatabaseMigrationFile))
+                {
+                    databaseMigrationsCursor++;
+                }
+                else
+                {
+                    if (!databaseMigrations[databaseMigrationsCursor].Equals(localMigrations[localMigrationsCursor]))
+                    {
+                        throw new Exception($"Migration validation failed: Mismatch local migration file: {localMigrations[localMigrationsCursor]}");
```
joelverhagen: Don't use `Exception`.

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -91,6 +125,48 @@ private void ExecuteDatabaseMigration(Func<DbMigrator> getMigrator, SqlConnectio
...
+                    }
+
+                    localMigrationsCursor++;
+                    databaseMigrationsCursor++;
+                }
+            }
+
+            if (databaseMigrationsCursor < databaseMigrations.Count)
+            {
+                throw new Exception("Migration validation failed: Database migrations are ahead of local migrations.");
```
joelverhagen: Don't use `Exception`.

sign.thirdparty.targets
```diff
@@ -0,0 +1,64 @@
+
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <ItemGroup>
+        <SignFilesDependsOn Include="EnumerateThirdPartyBinariesToSign" />
+    </ItemGroup>
+    <Target Name="EnumerateThirdPartyBinariesToSign" AfterTargets="AfterBuild" Condition="'$(SignType)' != 'none'">
+        <ItemGroup>
+            <ThirdPartyBinaries Include="AnglicanGeek.MarkdownMailer.dll" />
```
loic-sharma: Do these show up in Visual Studio's solution explorer? If so, add `Visible="false"`. For example, https://github.com/NuGet/NuGet.Services.Metadata/blob/dev/src/NuGet.Services.SearchService/NuGet.Services.SearchService.csproj#L118

zhhyu: No, it seems that it's not shown in Visual Studio solution explorer. 
I use this trick for the "PowerShellScriptsToNotSign"/"ExecutablesToNotSign" in the "proj" of "DatabaseMigrationTools".

scottbommarito: Separate `targets` files don't show up solution explorer AFAIK.

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -20,8 +20,10 @@ namespace NuGet.Services.DatabaseMigration
 {
     public class Job : JsonConfigurationJob
     {
-        private const string MigrationTargetDatabaseArgument = "MigrationTargetDatabase";
+        public int ExitCode { get; set; }
 
+        private const string MigrationTargetDatabaseArgument = "MigrationTargetDatabase";
+        private const string SkipGalleryDatabaseMigrationFile = "201304262247205_CuratedPackagesUniqueIndex";
```
scottbommarito: Add a comment around why we're doing this.

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -56,32 +129,49 @@ private void ExecuteDatabaseMigration(Func<DbMigrator> getMigrator, SqlConnectio
...
             if (pendingMigrations.Count() > 0)
             {
-                Logger.LogInformation("Applying pending migrations: \n {PendingMigrations}", String.Join("\n", pendingMigrations));
+                try
+                {
+                    var databaseMigrations = migrator.GetDatabaseMigrations().ToList();
+                    databaseMigrations.Reverse();
+                    var localMigrations = migrator.GetLocalMigrations().ToList();
+
+                    CheckIsValidMigration(databaseMigrations: databaseMigrations, localMigrations: localMigrations);
```
joelverhagen: Why named parameters?

Issue #6803, opened 2019-01-08
[Azure Search] Add Google Analytics telemetry to support search KPIs (e.g. search result click index)


joelverhagen:

joelverhagen:
This is needed for A/B testing.

PR #7157, opened 2019-05-14
Add Google Analytics events for search

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6803.

Adds two new events to Google Analytics:

1. `search-selection`. This is emitted when someone clicks a search result. This includes the absolute index of the selection. In other words, with page size 20 and you click the 4th result on the 3rd page, the index will be `(20 * (3 - 1)) + (4 - 1) = 43`.
1. `search-page`. This is emitted when someone navigates to a page of search results. The page index is emitted with this event.

There is parity between these events and the client event where appropriate.

These events are only emitted when there is a search term. Our main search ranking efforts are going into relevancy towards the search query, not to the default page of popular packages. I would rather not add noise to the events related to people not even performing a search.

### DIFF
--- a/src/NuGetGallery/Views/Shared/ListPackages.cshtml
+++ b/src/NuGetGallery/Views/Shared/ListPackages.cshtml
@@ -61,9 +61,13 @@
     }
 
     <div class="list-packages" role="list">
+        @{ var itemIndex = Model.PageIndex * Model.PageSize; }
         @foreach (var package in Model.Items)
         {
-            @Html.Partial("_ListPackage", package)
+            // Only set an item index when there is a search query, for now.
+            var thisItemIndex = string.IsNullOrWhiteSpace(Model.SearchTerm) ? null : (int?) itemIndex;
+            @Html.Partial("_ListPackage", package, new ViewDataDictionary { { "itemIndex", thisItemIndex } })
+            itemIndex++;
         }
     </div>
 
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -61,9 +61,13 @@
     }
 
     <div class="list-packages" role="list">
+        @{ var itemIndex = Model.PageIndex * Model.PageSize; }
```
loic-sharma: 💯

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -61,9 +61,13 @@
...
         @foreach (var package in Model.Items)
         {
-            @Html.Partial("_ListPackage", package)
+            // Only set an item index when there is a search query, for now.
+            var thisItemIndex = string.IsNullOrWhiteSpace(Model.SearchTerm) ? null : (int?) itemIndex;
+            @Html.Partial("_ListPackage", package, new ViewDataDictionary{ { "itemIndex", thisItemIndex } })
```
loic-sharma:
```suggestion
            @Html.Partial("_ListPackage", package, new ViewDataDictionary { { "itemIndex", thisItemIndex } })
```

Feel free to ignore if this is intentional

Issue #7154, opened 2019-05-12
Version component of package details page URL is case sensitive


joelverhagen:
Compare the version selected between these two URLs:

https://www.nuget.org/packages/NUnit/3.0.0-rc-3
https://www.nuget.org/packages/NUnit/3.0.0-RC-3

Both should show `3.0.0-rc-3` selected.

This line is the problem.
https://github.com/NuGet/NuGetGallery/blob/9501b3539f893e6011a89120fd5a9b65a8420d86/src/NuGetGallery/Controllers/PackagesController.cs#L723

It is operating on a materialized list and uses a function that can't be mapped to SQL. Both are indicators that the string comparison is done in app code (case sensitive) not SQL Server (case insensitive in our collation).

scottbommarito:
Looked into this. The work that introduced this bug intended to make this comparison in app code. The fix is to make the app code comparison case-insensitive.

PR #7161, opened 2019-05-15
DisplayPackage, Manage, and DeleteSymbols pages should find exact version case-insensitively

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7154

### DIFF
--- a/src/NuGetGallery/Services/PackageService.cs
+++ b/src/NuGetGallery/Services/PackageService.cs
@@ -183,6 +183,16 @@ namespace NuGetGallery
             return package;
         }
 
+        public virtual Package FilterExactPackage(
+            IReadOnlyCollection<Package> packages,
+            string version)
+        {
+            return packages.SingleOrDefault(p => string.Equals(
+                p.NormalizedVersion,
+                NuGetVersionFormatter.Normalize(version),
+                StringComparison.OrdinalIgnoreCase));
+        }
+
         public virtual Package FilterLatestPackage(
             IReadOnlyCollection<Package> packages,
             int? semVerLevelKey = SemVerLevelKey.SemVer2,
--- a/src/NuGetGallery/Services/PackageService.cs
+++ b/src/NuGetGallery/Services/PackageService.cs
@@ -187,9 +187,10 @@ namespace NuGetGallery
             IReadOnlyCollection<Package> packages,
             string version)
         {
+            var normalizedVersion = NuGetVersionFormatter.Normalize(version);
             return packages.SingleOrDefault(p => string.Equals(
                 p.NormalizedVersion,
-                NuGetVersionFormatter.Normalize(version),
+                normalizedVersion,
                 StringComparison.OrdinalIgnoreCase));
         }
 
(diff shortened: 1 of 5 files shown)


src/NuGetGallery/Services/PackageService.cs
```diff
@@ -183,6 +183,16 @@ public override PackageRegistration FindPackageRegistrationById(string packageId
             return package;
         }
 
+        public virtual Package FilterExactPackage(
+            IReadOnlyCollection<Package> packages,
+            string version)
+        {
+            return packages.SingleOrDefault(p => string.Equals(
+                p.NormalizedVersion,
+                NuGetVersionFormatter.Normalize(version),
```
loic-sharma: Should you normalize the version before `SingleOrDefault` to avoid normalizing the string a bunch of times?

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -183,6 +183,16 @@ public override PackageRegistration FindPackageRegistrationById(string packageId
             return package;
         }
 
+        public virtual Package FilterExactPackage(
```
loic-sharma: Since this may return null, what do you think of this name?

```suggestion
        public virtual Package FilterExactPackageOrNull(
```

src/NuGetGallery/Services/PackageService.cs
scottbommarito: LINQ converts `NuGetVersionFormatter.Normalize(version)` to a constant before running it on all of the packages.

I could move it to make it a little easier to read however.

joelverhagen: Wait really? Are you sure it's not LINQ-to-SQL/EF that's doing this?

src/NuGetGallery/Services/PackageService.cs
scottbommarito: Not sure...if this were `GetExactPackage` I would agree, but because I'm using the term `Filter` here I don't think adding `OrNull` makes as much sense. As we have `FilterLatestPackage` already which doesn't have `OrNull`.

src/NuGetGallery/Services/PackageService.cs
scottbommarito: I was _pretty_ sure of this, given how I've seen `Moq` handle these sort of statements in `Setup` calls, but maybe I'm wrong. I'll go ahead and move it anyway to be safe.

joelverhagen: `Moq` accepts and expression. This accepts a delegate. You can inspect an expression at runtime. AFAIK you can't do this to a delegate.

src/NuGetGallery/Services/PackageService.cs
scottbommarito: Added some xmldocs to define the behavior of the method more clearly.

src/NuGetGallery/Services/PackageService.cs
scottbommarito: @joelverhagen -- yup, that's what I was thinking. Moved it outside anyway.

PR #7165, opened 2019-05-16
StorageStatusService

cristinamanum:
Address issue https://github.com/NuGet/Engineering/issues/1980. Gallery status service will verify the storage status for primary or secondary depending on the read only mode.

### DIFF
--- a/src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Threading.Tasks;
+using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Blob;
 
 namespace NuGetGallery
@@ -34,9 +35,9 @@ namespace NuGetGallery
             return new CloudBlobWrapper(_blobContainer.GetBlockBlobReference(blobAddressUri));
         }
 
-        public Task<bool> ExistsAsync()
+        public Task<bool> ExistsAsync(BlobRequestOptions blobRequestOptions, OperationContext context)
         {
-            return _blobContainer.ExistsAsync();
+            return _blobContainer.ExistsAsync(blobRequestOptions, context);
         }
 
         public async Task<bool> DeleteIfExistsAsync()
--- a/src/NuGetGallery/Services/CloudBlobFileStorageService.cs
+++ b/src/NuGetGallery/Services/CloudBlobFileStorageService.cs
@@ -6,14 +6,12 @@ using System.Diagnostics;
 using System.Threading.Tasks;
 using System.Web;
 using System.Web.Mvc;
-using Microsoft.WindowsAzure.Storage;
-using Microsoft.WindowsAzure.Storage.Blob;
 using NuGetGallery.Configuration;
 using NuGetGallery.Diagnostics;
 
 namespace NuGetGallery
 {
-    public class CloudBlobFileStorageService : CloudBlobCoreFileStorageService, IFileStorageService, ICloudStorageStatusDependency
+    public class CloudBlobFileStorageService : CloudBlobCoreFileStorageService, IFileStorageService
     {
         private readonly IAppConfiguration _configuration;
         private readonly ISourceDestinationRedirectPolicy _redirectPolicy;
@@ -97,19 +95,10 @@ namespace NuGetGallery
             return urlBuilder.Uri;
         }
 
-        public async Task<bool> IsAvailableAsync(BlobRequestOptions options, OperationContext operationContext)
-        {
-            var container = await GetContainerAsync(CoreConstants.Folders.PackagesFolderName);
-            return await container.ExistsAsync(options, operationContext);
-        }
-
-        /// <summary>
-        /// IFileStorageService.IsAvailableAsync 
-        /// </summary>
-        /// <returns></returns>
         public async Task<bool> IsAvailableAsync()
         {
-            return await IsAvailableAsync(options: null, operationContext: null);
+            var container = await GetContainerAsync(CoreConstants.Folders.PackagesFolderName);
+            return await container.ExistsAsync(options: null, operationContext: null);
         }
     }
 }
--- a/src/NuGetGallery/Services/CloudBlobFileStorageService.cs
+++ b/src/NuGetGallery/Services/CloudBlobFileStorageService.cs
@@ -6,6 +6,8 @@ using System.Diagnostics;
 using System.Threading.Tasks;
 using System.Web;
 using System.Web.Mvc;
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
 using NuGetGallery.Configuration;
 using NuGetGallery.Diagnostics;
 
@@ -95,10 +97,19 @@ namespace NuGetGallery
             return urlBuilder.Uri;
         }
 
-        public async Task<bool> IsAvailableAsync()
+        public async Task<bool> IsAvailableAsync(BlobRequestOptions options, OperationContext operationContext)
         {
             var container = await GetContainerAsync(CoreConstants.Folders.PackagesFolderName);
-            return await container.ExistsAsync();
+            return await container.ExistsAsync(options, operationContext);
+        }
+
+        /// <summary>
+        /// IFileStorageService.IsAvailableAsync 
+        /// </summary>
+        /// <returns></returns>
+        public async Task<bool> IsAvailableAsync()
+        {
+            return await IsAvailableAsync(options: null, operationContext: null);
         }
     }
 }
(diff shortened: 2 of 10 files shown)


src/NuGetGallery/Services/StatusService.cs
```diff
@@ -91,7 +93,9 @@ private bool IsSqlAzureAvailable()
             try
             {
                 // Check Storage Availability
-                var tasks = _cloudStorageAvailabilityChecks.Select(s => s.IsAvailableAsync());
+                BlobRequestOptions options = new BlobRequestOptions();
+                options.LocationMode = _config.ReadOnlyMode ? LocationMode.SecondaryOnly : LocationMode.PrimaryOnly;
+                var tasks = _cloudStorageAvailabilityChecks.Select(s => s.IsAvailableAsync(options, operationContext : null));
```
joelverhagen: Why do we plumb `operationContext` through if it's always null?

joelverhagen: Actually, it's been rather painful in the past adding overload after overload as we need more options from the storage SDK. I think this is the most future-proof.

tests/NuGetGallery.Facts/Services/StatusServiceFacts.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace NuGetGallery.Services
+{
+    public class StatusServiceFacts
+    {
+        [Fact]
+        public void ValidateServicesThatWillBeInStatusValidation()
+        {
+            // Arrange
+            var cloudBlobFileStorageServiceIsStatusParticipant = typeof(ICloudStorageStatusDependency).IsAssignableFrom(typeof(CloudBlobFileStorageService));
```
joelverhagen: You can do `Assert.IsType<T>`

joelverhagen: Or `Assert.IsAssignableFrom<T>`

src/NuGetGallery/Services/StatusService.cs
joelverhagen: Unit tests for this?

src/NuGetGallery/Services/StatusService.cs
```diff
@@ -91,7 +93,9 @@ private bool IsSqlAzureAvailable()
             try
             {
                 // Check Storage Availability
-                var tasks = _cloudStorageAvailabilityChecks.Select(s => s.IsAvailableAsync());
+                BlobRequestOptions options = new BlobRequestOptions();
+                options.LocationMode = _config.ReadOnlyMode ? LocationMode.SecondaryOnly : LocationMode.PrimaryOnly;
```
skofman1: I think that if we are in ReadOnlyMode, we should use LocationMode.PrimaryThanSecondary, because this is how our blob access is configured. i.e. if secondary blob storage is down, read-only Gallery is still up, since it reads from Primary blob storage, so there's no need to advertise it's status as unavailable.

src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
```diff
@@ -34,9 +35,9 @@ public ISimpleCloudBlob GetBlobReference(string blobAddressUri)
             return new CloudBlobWrapper(_blobContainer.GetBlockBlobReference(blobAddressUri));
         }
 
-        public Task<bool> ExistsAsync()
+        public Task<bool> ExistsAsync(BlobRequestOptions blobRequestOptions, OperationContext context)
```
skofman1: I didn't see any usages of this method where OperationContext is not null. Do we need this param?

tests/NuGetGallery.Facts/Services/StatusServiceFacts.cs
```diff
@@ -0,0 +1,23 @@
...
+            var cloudAuditingServiceIsStatusParticipant = typeof(ICloudStorageStatusDependency).IsAssignableFrom(typeof(CloudAuditingService));
```
skofman1: Those two services are configured with the same storage account, so by having both implementing ICloudStorageStatusDependency both will hit the same blob storage. 
Here's a related bug: [https://github.com/NuGet/Engineering/issues/2095](https://github.com/NuGet/Engineering/issues/2095)

src/NuGetGallery/Services/CloudBlobFileStorageService.cs
```diff
@@ -95,10 +97,19 @@ internal Uri GetRedirectUri(Uri requestUrl, Uri blobUri)
...
             var container = await GetContainerAsync(CoreConstants.Folders.PackagesFolderName);
-            return await container.ExistsAsync();
+            return await container.ExistsAsync(options, operationContext);
+        }
+
+        /// <summary>
+        /// IFileStorageService.IsAvailableAsync 
+        /// </summary>
+        /// <returns></returns>
+        public async Task<bool> IsAvailableAsync()
```
scottbommarito: Any reason we can't make `options` and `operationContext` optional in the method above and get rid of this additional method?

src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
scottbommarito: From @joelverhagen:
> Actually, it's been rather painful in the past adding overload after overload as we need more options from the storage SDK. I think this is the most future-proof.

I agree with him...might as well leave this parameter here and avoid pains later.

src/NuGetGallery/Services/StatusService.cs
cristinamanum: Added.

tests/NuGetGallery.Facts/Services/StatusServiceFacts.cs
cristinamanum: Any of these options require an object instance.

src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
cristinamanum: It was a comment earlier. I think that it will support future cases,

src/NuGetGallery/Services/StatusService.cs
cristinamanum: Using PrimaryThanSecondary will  do an extra call to primary that is not necessary. Also it will not actually check the health of the secondary storage.

src/NuGetGallery/Services/CloudBlobFileStorageService.cs
cristinamanum: I do not think that is relevant for this bug.

src/NuGetGallery/Services/CloudBlobFileStorageService.cs
```diff
@@ -95,10 +97,19 @@ internal Uri GetRedirectUri(Uri requestUrl, Uri blobUri)
...
-        public async Task<bool> IsAvailableAsync()
+        public async Task<bool> IsAvailableAsync(BlobRequestOptions options, OperationContext operationContext)
         {
             var container = await GetContainerAsync(CoreConstants.Folders.PackagesFolderName);
-            return await container.ExistsAsync();
+            return await container.ExistsAsync(options, operationContext);
+        }
+
+        /// <summary>
+        /// IFileStorageService.IsAvailableAsync
```
loic-sharma: >IFileStorageService.IsAvailableAsync  [](start = 12, length = 37)

Is this comment unintentional? If not, please remove

tests/NuGetGallery.Facts/Services/StatusServiceFacts.cs
```diff
@@ -0,0 +1,221 @@
...
+                var cloudStorageStatusDependency1 = CloudStorageStatusDependencyFactory.GetCloudStorageStatusDependency(ICloudStorageStatusDependencyTestType.Available, appCopnfiguration.Object);
+                var cloudStorageStatusDependency2 = CloudStorageStatusDependencyFactory.GetCloudStorageStatusDependency(ICloudStorageStatusDependencyTestType.Available, appCopnfiguration.Object);
+
+                var statusTestService = new StatusService(entities: null,
+                    cloudStorageAvailabilityChecks: new ICloudStorageStatusDependency[] { cloudStorageStatusDependency1, cloudStorageStatusDependency2 },
+                    config: appCopnfiguration.Object);
+                var result = await statusTestService.IsAzureStorageAvailable();
+
+                // Assert
+                Assert.True(result.Value);
```
loic-sharma: >          [](start = 42, length = 9)

There's extra whitespace at the end of this line

cristinamanum: Updated.

src/NuGetGallery/Services/CloudBlobFileStorageService.cs
cristinamanum: Updated

tests/NuGetGallery.Facts/Services/StatusServiceFacts.cs
cristinamanum: Let only the CloudAuditingService to participate in StatusCheck.

Issue #7166, opened 2019-05-16
[Azure Search] Implement randomized A/B test for new search service


joelverhagen:
Very simple approach.

1. Randomly return a cookie to the user: `previewSearch = true | false`
    - Get the randomness percentage from JSON content file
1. With this cookie as `true`, use the preview search instead
    - Need a cookie so pagination isn't totally crazy and a single user has a consistent experience
1. Add feature flag to turn this off if things go bad.

This should happen once we have decently acceptable results.

Consider putting a Usabilla button on the treatment page.

scottbommarito:
Note: will need to add new cookie to cookie compliance list.

joelverhagen:
We should have an opt-out button and ask for feedback as they are opting out.

joelverhagen:
If the Azure Search query takes too long or fails, fall back to control.

joelverhagen:
I've prototyped something simple but would like to start the onboarding process with the internal A/B testing system before shipping something we cook ourselves.

Issue #7152, opened 2019-05-10
[Azure Search] Add page to gallery than can be used from compare new and old search results


joelverhagen:
This is the initial work related to https://github.com/nuget/nugetgallery/issues/7057.

The page will show both new and old search results.

It will have a "smile" and "frowny" button.

It will have a text area that allows custom feed back. The user can optionally provide their email so we can reach back to them. Logged in users have the email field auto-filled in.

karann-msft:
Can we look at a quick mock before we make too much engineering investment?

joelverhagen:
Yes.

PR #7171, opened 2019-05-20
Add capability of a preview ISearchService

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7152.

- Add config for two "preview" search endpoints.
- Add an `ISearchService` instance that hits these preview endpoints
  - This is available with a binding key. The default `ISearchService` has not changed.

This basic infrastructure is for the side-by-side search page and can be adapted for https://github.com/NuGet/NuGetGallery/issues/7166 later.

To do this, I needed to move the DI wire-up of `ISearchService` to Autofac, which has binding key support.

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -50,6 +50,7 @@ using NuGetGallery.Infrastructure.Search;
 using NuGetGallery.Infrastructure.Search.Correlation;
 using NuGetGallery.Security;
 using SecretReaderFactory = NuGetGallery.Configuration.SecretReader.SecretReaderFactory;
+using Microsoft.Extensions.Http;
 
 namespace NuGetGallery
 {
@@ -62,6 +63,8 @@ namespace NuGetGallery
             public const string PackageValidationEnqueuer = "PackageValidationEnqueuerBindingKey";
             public const string SymbolsPackageValidationEnqueuer = "SymbolsPackageValidationEnqueuerBindingKey";
             public const string EmailPublisherTopic = "EmailPublisherBindingKey";
+
+            public const string PreviewSearchClient = "PreviewSearchClientBindingKey";
         }
 
         protected override void Load(ContainerBuilder builder)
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -671,60 +673,104 @@ private static DbConnection CreateDbConnection(ISqlConnectionFactory connectionF
                 .InstancePerLifetimeScope();
         }
 
-        private static void ConfigureSearch(ContainerBuilder builder, IGalleryConfigurationService configuration)
+        private static List<(string name, Uri searchUri)> GetSearchClientsFromConfiguration(IGalleryConfigurationService configuration)
```
scottbommarito: You could probably combine this and the preview search clients method by allowing the primary and secondary config strings to be passed in.

Also this new tuple syntax is awesome but I'd make sure it works on our build machines haha.

joelverhagen: Named tuple usage is pre-existing.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -671,60 +673,104 @@ private static DbConnection CreateDbConnection(ISqlConnectionFactory connectionF
...
+            }
+        }
+
+        private static void RegisterSearchService(
+            ILoggerFactory loggerFactory,
+            IGalleryConfigurationService configuration,
+            ITelemetryService telemetryService,
+            ServiceCollection services,
+            ContainerBuilder builder,
+            List<(string name, Uri searchUri)> searchClients,
```
skofman1: how does this behave when there are no preview search clients configured?

joelverhagen: A `ResilientSearchHttpClient` will be activated with no HTTP clients. The code handles this by faking a 503 response message with empty JSON. This is very similar to the case when both circuit breakers are open. This will never happen for the default search client though since the whole path is gated on `searchClients.Count >= 1`.

Issue #7115, opened 2019-04-29
Updating a package's deprecation information should update its LastEdited and LastUpdated


scottbommarito:
We initially removed it so that deprecation was independent of V3, but now that we are piping deprecation to V3, we need to add it back.

scottbommarito:
Can't verify on PROD due to feature flag.

PR #7173, opened 2019-05-21
Deprecation information should flow into V3

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7115

This PR updates the package's `LastEdited` and `LastUpdated` timestamps when a package is deprecated so that these changes will appear in the feed and be consumed by DB2Catalog.


src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
```diff
@@ -132,10 +135,10 @@ function ManageDeprecationViewModel(id, versionDeprecationStateDictionary, defau
 
     // The alternate package ID to submit with the form.
     this.alternatePackageId = ko.pureComputed(function () {
-        if (self.isLegacy()) {
+        if (self.hasReason()) {
```
scottbommarito: This fixes a bug with the form right now--it doesn't submit the alternate package with the form if Legacy is not selected.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -483,7 +483,7 @@ public async Task MarkPackageUnlistedAsync(Package package, bool commitChanges =
             // NOTE: LastEdited will be overwritten by a trigger defined in the migration named "AddTriggerForPackagesLastEdited".
             package.LastEdited = DateTime.UtcNow;
 
-            if (package.IsLatest || package.IsLatestStable)
+            if (package.IsLatest || package.IsLatestStable || package.IsLatestSemVer2 || package.IsLatestStableSemVer2)
```
scottbommarito: This appears to be an existing bug with the `PackageService`--when a package is unlisted, if it is latest SemVer2 but not latest non-semver2, is latest is not updated.

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -72,6 +84,14 @@ public class PackageDeprecationService : IPackageDeprecationService
 
                     deprecation.CustomMessage = customMessage;
                 }
+
+                package.LastUpdated = DateTime.UtcNow;
+                package.LastEdited = DateTime.UtcNow;
```
xavierdecoster: Perhaps call `DateTime.UtcNow` only once? You could set `package.LastEdited = package.LastUpdated = DateTime.UtcNow;` in one go.

src/NuGetGallery/Services/PackageService.cs
xavierdecoster: Do we have tests covering this?

scottbommarito: Improved the existing one.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: I love that syntax. Functionally it won't change anything, though--`LastEdited` is overwritten by a trigger.

scottbommarito:
I discovered that this PR does not currently set `LastEdited` properly--all packages should have their `LastEdited` set to the same value, but currently they are being set differently.

I am fixing this issue and will update the PR when I've done it.

src/NuGetGallery/Services/PackageService.cs
```diff
@@ -445,7 +445,7 @@ public async Task MarkPackageListedAsync(Package package, bool commitChanges = t
                 throw new InvalidOperationException("A package that failed validation should never be listed!");
             }
 
-            if (!package.Listed && (package.IsLatestStable || package.IsLatest))
+            if (!package.Listed && (package.IsLatestStable || package.IsLatest || package.IsLatestSemVer2 || package.IsLatestStableSemVer2))
```
scottbommarito: will add test for this

src/NuGetGallery/Views/Packages/Manage.cshtml
```diff
@@ -80,22 +81,22 @@
         var strings_RemovingSelf = "@Html.Raw(Strings.ManagePackageOwners_RemovingSelf)";
 
         // Set up delete section
-        var versionListedState = @Html.Raw(Json.Encode(Model.VersionListedStateDictionary));
+        var versionListedState = @Html.Raw(JsonConvert.SerializeObject(Model.VersionListedStateDictionary));
 
         // Set up deprecation section
         var strings_SelectAlternateVersionOption = "Latest";
-        var versionDeprecationState = @Html.Raw(Json.Encode(Model.VersionDeprecationStateDictionary));
+        var versionDeprecationState = @Html.Raw(JsonConvert.SerializeObject(Model.VersionDeprecationStateDictionary));
```
scottbommarito: With 10k versions, this overflowed our maximum JSON length limit with `Json.Encode`.

`JsonConvert` has no limit. This is a pattern we already use on the [Manage Packages page](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Views/Users/Packages.cshtml#L459).

scottbommarito:
PR is good for review now!

src/NuGetGallery/Services/BulkPackageUpdateService.cs
```diff
@@ -0,0 +1,99 @@
...
+            if (setListed.HasValue)
+            {
+                // Update listed state first to minimize the amount of time between setting LastEdited and committing the transaction.
+                await UpdatePackagesListedAsync(packages, setListed.Value);
+            }
+
+            await UpdatePackagesTimestampsAsync(packages);
+        }
+
+        private async Task UpdatePackagesListedAsync(IEnumerable<Package> packages, bool setListed)
```
joelverhagen: You're enumerating this many times over. Can you just accept an `IReadOnlyList`?

src/NuGetGallery/Services/BulkPackageUpdateService.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            var query = string.Format(
+                BaseQueryFormat,
+                string.Join(", ", packages.Select(p => p.Key)));
+
+            var result = await _entitiesContext
+                .GetDatabase()
+                .ExecuteSqlCommandAsync(query);
+
+            // The query updates each row twice--once for the initial commit and a second time due to the trigger on LastEdited.
```
joelverhagen: How are we sure that the trigger is applying the same value to all rows? Is there documentation supporting this? The basis of the entire design depends on this fact. I personally feel that local tests are not enough for full validation. For example, what if the DB is under load?

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -26,64 +32,82 @@ public class PackageDeprecationService : IPackageDeprecationService
...
+                {
+                    _deprecationRepository.DeleteOnCommit(deprecations);
+                }
+                else
+                {
+                    _deprecationRepository.InsertOnCommit(deprecations);
+                }
+
+                // Save deprecation changes before bulk updating the packages.
+                await _deprecationRepository.CommitChangesAsync();
```
joelverhagen: Have you manually tested that this works exactly as expected? I am thinking the following steps:

1. Open SSMS against target DB (I think DEV Azure SQL is probably a good first test).
1. Query the last edited on the affected packages. Capture the output in csv or something.
1. Set a breakpoint at this line (`CommitChangesAsync`)
1. Run this code and wait for the breakpoint to hit here.
1. Repeat step 2. Diff with previous result. There should be no change.
1. Step over.
1. Repeat step 5.
1. Step over the call to the bulk package update service.
1. Repeat step 5.
1. Step over the transaction commit. Assert all last edited values have changed to the same value.

Recommended with 10K versions.

src/NuGetGallery.Core/Entities/EntityRepository.cs
```diff
@@ -47,5 +47,10 @@ public IQueryable<T> GetAll()
         {
             return _entities.Set<T>();
         }
+
+        public IDatabase GetDatabase()
```
joelverhagen: I think we should just import the entities context here instead. This abstraction shouldn't leak anymore IMO.

src/NuGetGallery/Services/BulkPackageUpdateService.cs
scottbommarito: good call!

src/NuGetGallery.Core/Entities/EntityRepository.cs
scottbommarito: In retrospect I think I added this because I needed it but I no longer do. Can probably take this out entirely.

src/NuGetGallery/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -103,36 +106,48 @@ public void UpdatePackage(Package package)
                 var packageRegistrationKey = package.PackageRegistrationKey;
                 var updateTerm = new Term("PackageRegistrationKey", packageRegistrationKey.ToString(CultureInfo.InvariantCulture));
 
-                if (!package.IsLatest || !package.IsLatestStable)
+                if (!package.IsLatest || !package.IsLatestStable || !package.IsLatestSemVer2 || !package.IsLatestStableSemVer2)
```
skofman1: UTs?

scottbommarito: I will probably revert the entire `LuceneIndexingService` changes and then replace `UpdatePackageRegistration` with an `UpdatePackage(packages.First())` call.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: I believe I have done this but I will do it again.

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -26,64 +32,82 @@ public class PackageDeprecationService : IPackageDeprecationService
...
+            var registration = packages.First().PackageRegistration;
+            if (packages.Select(p => p.PackageRegistrationKey).Distinct().Count() > 1)
+            {
+                throw new ArgumentException("All packages to deprecate must have the same ID.", nameof(packages));
+            }
+
             if (user == null)
             {
                 throw new ArgumentNullException(nameof(user));
             }
```
skofman1: the cheaper verification should always come first

src/NuGetGallery/Services/BulkPackageUpdateService.cs
```diff
@@ -0,0 +1,99 @@
...
+                package.Listed = setListed;
+            }
+
+            if (!packagesByRegistration.Any(p => p.IsLatest || p.IsLatestStable || p.IsLatestSemVer2 || p.IsLatestStableSemVer2))
+            {
+                // Don't need to update latest if we haven't affected any latest packages.
+                return;
+            }
+
+            await _packageService.UpdateIsLatestAsync(packagesByRegistration.Key, false);
```
joelverhagen: name bool param

src/NuGetGallery/Services/BulkPackageUpdateService.cs
```diff
@@ -0,0 +1,99 @@
...
+        public BulkPackageUpdateService(
+            IEntitiesContext entitiesContext,
+            IPackageService packageService)
+        {
+            _entitiesContext = entitiesContext ?? throw new ArgumentNullException(nameof(entitiesContext));
+            _packageService = packageService ?? throw new ArgumentNullException(nameof(packageService));
+        }
+
+        public async Task UpdatePackagesAsync(IEnumerable<Package> packages, bool? setListed = null)
+        {
```
joelverhagen: Can we assert that there is a transaction?

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -26,64 +32,82 @@ public class PackageDeprecationService : IPackageDeprecationService
...
-            var deprecations = new List<PackageDeprecation>();
-            foreach (var package in packages)
+            using (var strategy = new SuspendDbExecutionStrategy())
+            using (var transaction = _deprecationRepository.GetDatabase().BeginTransaction())
             {
-                var deprecation = package.Deprecations.SingleOrDefault();
-                if (shouldDelete)
+                var shouldDelete = status == PackageDeprecationStatus.NotDeprecated;
+                var deprecations = new List<PackageDeprecation>();
+                var deprecationTime = DateTime.UtcNow;
```
joelverhagen: unused?

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -26,64 +32,82 @@ public class PackageDeprecationService : IPackageDeprecationService
...
-                        package.Deprecations.Add(deprecation);
-                        deprecations.Add(deprecation);
-                    }
+                            package.Deprecations.Add(deprecation);
+                            deprecations.Add(deprecation);
+                        }
 
-                    deprecation.Status = status;
-                    deprecation.DeprecatedByUser = user;
+                        deprecation.Status = status;
```
joelverhagen: just curious why we modify an existing deprecation instead of creating a new one?

what were the considerations when making this decision? is there semantic significance? perf implications?

skofman1:
>             _telemetryService.TrackPackageUnlisted(package);

we should reduce code duplication between this and bulk service. Unlist functionality should be in one place

---
Refers to: src/NuGetGallery/Services/PackageService.cs:493 in 568c2ab. [](commit_id = 568c2ab02bc9dc88a305e1c40223f353116ec736, deletion_comment = False)

src/NuGetGallery/Services/PackageDeprecationService.cs
joelverhagen: Seems like we could potentially save work if some of the existing deprecations have not changed... but then we are updating last edited on all packages

why not just update last edited on deprecations that changed? wouldn't that reduce work downstream?

src/NuGetGallery/Services/BulkPackageUpdateService.cs
```diff
@@ -0,0 +1,99 @@
...
+            if (packages == null || !packages.Any())
+            {
+                throw new ArgumentException(nameof(packages));
+            }
+
+            if (setListed.HasValue)
+            {
+                // Update listed state first to minimize the amount of time between setting LastEdited and committing the transaction.
+                await UpdatePackagesListedAsync(packages, setListed.Value);
```
skofman1: >UpdatePackagesListedAsync [](start = 22, length = 25)

This method doesn't do anything "bulky", so can be taken out of this service and moved to deprecation service.

scottbommarito: It's shared logic that will be needed for both bulk deprecation and bulk unlisting.

src/NuGetGallery/Services/BulkPackageUpdateService.cs
scottbommarito: Not sure...I was considering having a transaction as a parameter but it wouldn't be used in any way.

src/NuGetGallery/Services/BulkPackageUpdateService.cs
skofman1: A class should do one thing. By adding unlisting logic to the class it adds unnecessary complexity.

src/NuGetGallery/Services/BulkPackageUpdateService.cs
scottbommarito: The SQL server docs seem to **heavily** imply that 1 statement results in 1 trigger execution, but I haven't been able to find any documentation that explicitly states this. I have reached out to some discussion aliases for clarification.

src/NuGetGallery/Services/BulkPackageUpdateService.cs
joelverhagen: Perhaps there is something on the entities context / db object we can assert? Basically this should always be executed inside a transaction right?

src/NuGetGallery/Services/BulkPackageUpdateService.cs
scottbommarito: Discussion alias verified that trigger will only execute once.

joelverhagen:
There are merge conflicts.

src/NuGetGallery/Services/BulkPackageUpdateService.cs
agr: `IReadOnlyCollection` is the most basic interface that can be enumerated multiple times I think. Does not matter much for the private method, though.

src/NuGetGallery.Core/Entities/EntityRepository.cs
scottbommarito: It is used...will consider using entities context instead.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: I think functionally there isn't much difference between modifying an existing deprecation rather than creating a new one, but I like this way because it results in less work for the DB. Additionally, it stops the `[Key]` field from increasing.

We decided to avoid "diffing" deprecations to avoid additional complexity. Only updating deprecations that have changed requires diffing, which we wanted to avoid. I don't think the additional overheard is likely to be very significant--the number of versions owned by the affected ID scales the work to do more than the number of versions deprecated.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: The implementation has changed and now `LastUpdated` and `LastEdited` are no longer set here.

src/NuGetGallery.Core/Entities/EntityRepository.cs
scottbommarito: Switched everything to `EntitiesContext`.

src/NuGetGallery/Infrastructure/Lucene/LuceneIndexingService.cs
scottbommarito: Reverted.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: done

src/NuGetGallery/Services/BulkPackageUpdateService.cs
scottbommarito: Doesn't look like there is.

To mitigate this, I've changed the code so that the transaction is committed inside `PackageUpdateService` (the replacement for `BulkPackageUpdateService`), so a transaction must be passed into the method.

src/NuGetGallery/Services/BulkPackageUpdateService.cs
scottbommarito: Did a heavy refactor to rearrange things.

We now have `PackageUpdateService`, which handles all unlisting/updating of existing packages.

I moved the existing unlist/list methods into the class and added a method to unlist/list/update in bulk.

scottbommarito:
@joelverhagen - Oddly, `git` was locally able to resolve all merge conflicts without any manually changes.

scottbommarito:
@joelverhagen - fixed the conflicts

src/NuGetGallery.Services/PackageManagement/IIndexingService.cs
```diff
@@ -20,8 +17,6 @@ public interface IIndexingService
         Task<int> GetDocumentCount();
         Task<long> GetIndexSizeInBytes();
 
-        void RegisterBackgroundJobs(IList<IJob> jobs, IAppConfiguration configuration);
```
scottbommarito: Due to the `NuGetGallery.Services` restructuring, I needed to
- move `IIndexingService` into `NuGetGallery.Services`
- move `RegisterBackgroundJobs` from `IIndexingService` to `IIndexingJobFactory`

I chose to do this instead of pulling in `WebBackgrounder` and `IAppConfiguration` into `NuGetGallery.Services`.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: I've done this a couple times and can assure you that no changes are persisted to the DB until `transaction.Commit()` is called.

scottbommarito:
@joelverhagen @skofman1 - I've handled all comments. Please review again!

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
```diff
@@ -0,0 +1,185 @@
...
+using NuGet.Services.Entities;
+using NuGetGallery.Auditing;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace NuGetGallery
+{
+    public class PackageUpdateService : IPackageUpdateService
```
scottbommarito: I originally created this class because `UpdatePackagesAsync` could unlist/relist packages as well as update them, so I wanted a shared class for all of these operations.

I would like to keep this restructuring so that in the future I can easily add back the unlist/relist capabilities, which is necessary for bulk unlist/relist operations.

src/NuGetGallery/Services/PackageDeprecationService.cs
joelverhagen: I think it's a guess to say that the updates would perform better than the inserts. But I agree it would be more complex.

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
```diff
@@ -0,0 +1,185 @@
...
+            {
+                // The indexing service will find the latest version of the package to index--it doesn't matter what package we pass in.
+                _indexingService.UpdatePackage(packages.First());
+            }
+        }
+
+        private const string UpdateBulkPackagesQueryFormat = @"
+UPDATE [dbo].Packages
+SET LastEdited = GETUTCDATE(), LastUpdated = GETUTCDATE()
+WHERE [Key] IN ({0})";
```
agr: Are there limits on query length? Can it run over it?

agr: This also reeks SQL injection...

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        /// <remarks>
+        /// Normally we would use a large parameterized SQL query for this.
+        /// Unfortunately, however, there is a maximum number of parameters for a SQL query (around 2,000-3,000).
+        /// By writing a query containing the package keys directly we can remove this restriction.
+        /// Furthermore, package keys are not user data, so there is no risk to writing a query in this way.
+        /// </remarks>
+        private async Task UpdatePackagesInBulkAsync(IReadOnlyList<Package> packages)
```
agr: This function don't have a good rollback story and must run in a transaction. I suggest passing the transaction as an argument and check inside that it's active to enforce that.

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
```diff
@@ -0,0 +1,185 @@
...
+                await _entitiesContext.SaveChangesAsync();
+            }
+
+            if (updateIndex)
+            {
+                _indexingService.UpdatePackage(package);
+            }
+        }
+
+        public async Task UpdatePackagesAsync(IReadOnlyList<Package> packages, IDbContextTransaction transaction, bool updateIndex = true)
```
agr: `null` check for `transaction`.

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -12,83 +12,101 @@ namespace NuGetGallery
...
+                }
+                else
+                {
+                    _entitiesContext.Deprecations.AddRange(deprecations);
+                }
+
+                await _entitiesContext.SaveChangesAsync();
 
-            await _deprecationRepository.CommitChangesAsync();
+                await _packageUpdateService.UpdatePackagesAsync(packages, transaction: transaction);
```
agr: I'd expect transaction to be committed in this method and not somewhere deep down the stack.

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
scottbommarito: Locally, with 10,000 versions, I didn't hit any issues.

Users can't control their `Package.[Key]` so I don't see any potential for SQL injection. Rewriting this query in a way that gets around the parameter limit without doing this is very difficult--I thought this was an acceptable workaround.

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -12,83 +12,101 @@ namespace NuGetGallery
...
+            }
```
agr: There are no guarantees that [`DbTransaction.Dispose()`](https://docs.microsoft.com/en-us/dotnet/api/system.data.common.dbtransaction.dispose?redirectedfrom=MSDN&view=netframework-4.6.2#System_Data_Common_DbTransaction_Dispose) will rollback, it's better to have explicit one.
On the other hand, since we target only SQL server that might not matter much, since I'd expect it does rollback.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: Okay, so to summarize a couple of your comments:
- pass in a transaction to `UpdatePackagesAsync`
- `UpdatePackagesAsync` should check that the transaction is active and not null
- `PackageDeprecationService` should commit the transaction

Sounds good to me.

src/NuGetGallery/Views/Packages/Manage.cshtml
agr: Maybe add a comment in code about what to use and what not to use and why?

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: AFAIK we rely on the `Dispose` rollback in a number of other places. In fact, `IDbContextTransaction`, our wrapper around `DbTransaction`, doesn't even have a `Rollback` method on it. Given this, I think it's fine to continue that pattern.

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
agr: Consider passing just a list of package keys: `IReadOnlyCollection<int>`, instead of list of packages, that's extra data in a method that builds raw SQL...

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
agr: Have you considered [table-valued parameters](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql/table-valued-parameters)?

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
scottbommarito: will do. great idea

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
scottbommarito: That's awesome--I will change this to a table-valued parameter.

scottbommarito: Unfortunately, I looked into this and realized using a TVP would be far too heavy-handed. The main blocker is that I would need to add a new table type to the DB as part of a migration (and not as part of this transaction) in order to pass in a table parameter with that shape. I don't think this query is worth this investment.

src/NuGetGallery/Services/PackageDeprecationService.cs
scottbommarito: I realized I don't even need a transaction in this method anymore due to the remove of listing/unlisting from this function!

src/NuGetGallery/Views/Packages/Manage.cshtml
scottbommarito: Comment added.

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
scottbommarito: No longer required to run in a transaction.

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        public async Task UpdatePackagesAsync(IReadOnlyList<Package> packages, bool updateIndex = true)
+        {
+            if (packages == null || !packages.Any())
+            {
+                throw new ArgumentException(nameof(packages));
+            }
+
+            var registration = packages.First().PackageRegistration;
```
loic-sharma: >registration [](start = 16, length = 12)

Is this variable used?

scottbommarito: It was before, but no longer...will remove

PR #7175, opened 2019-05-21
[SxS 1] Add initial infrastructure for search side-by-side (SxS) page

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7152.

- Add feature flag
- Add an email message that will be sent when people leave feedback
  - Email address and comments should not be recorded in AI and therefore take a path like support requests.
- Add telemetry to record SxS search and feedback in AI
  - Just "metadata" not comments or email address
- Minor style issues in existing code (VS was nagging me)

This is what the email looks like:

image

Look at the view model for an idea of the kind of data we are going to pass to the view and get back from the feedback form. Here's a sneak-peak of the UI (PR forthcoming) to help you understand the goal. Please leave UI comments to a future PR.

image

PR #7176, opened 2019-05-21
[SxS 2] Add SearchSideBySideService to do the two searches

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7152.
Depends on https://github.com/NuGet/NuGetGallery/pull/7175.

This is the service layer code to actually perform the search against the two search services.

### DIFF
--- a/src/NuGetGallery/Services/ISearchSideBySideService.cs
+++ b/src/NuGetGallery/Services/ISearchSideBySideService.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// A service which two search queries in parallel and return the results for the purposes of side-by-side
+    /// comparison.
+    /// </summary>
+    public interface ISearchSideBySideService
+    {
+        Task<SearchSideBySideViewModel> SearchAsync(string searchTerm, User currentUser);
+        Task RecordFeedbackAsync(SearchSideBySideViewModel viewModel, string searchUrl);
+    }
+}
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Services/SearchSideBySideService.cs
```diff
@@ -0,0 +1,117 @@
...
+            _messageService = messageService ?? throw new ArgumentNullException(nameof(messageService));
+            _messageServiceConfiguration = messageServiceConfiguration ?? throw new ArgumentNullException(nameof(messageServiceConfiguration));
+        }
+
+        public async Task<SearchSideBySideViewModel> SearchAsync(string q, User currentUser)
+        {
+            SearchSideBySideViewModel viewModel;
+            if (!string.IsNullOrWhiteSpace(q))
+            {
+                q = (q ?? string.Empty).Trim();
```
agr: `q` is guaranteed to be non-null here, right? No need to coalesce then.

src/NuGetGallery/Services/ISearchSideBySideService.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// A service which two search queries in parallel and return the results for the purposes of side-by-side
+    /// comparison.
+    /// </summary>
+    public interface ISearchSideBySideService
+    {
+        Task<SearchSideBySideViewModel> SearchAsync(string q, User currentUser);
```
agr: Should we use `query` instead of `q`?

src/NuGetGallery/Services/SearchSideBySideService.cs
joelverhagen: Oh yeah, good point.

src/NuGetGallery/Services/ISearchSideBySideService.cs
joelverhagen: Yes, I think that's better. I'll align with `searchTerm` since that is used in `SearchFilters`.

PR #7182, opened 2019-05-23
RemoveVulnerabilityDataFromPackageDeprecationModel.Down should recreate dbo.PackageDeprecations_CVSSRating constraint

xavierdecoster:


src/NuGetGallery/Migrations/201904292139301_RemoveVulnerabilityDataFromPackageDeprecationModel.cs
```diff
@@ -74,6 +74,7 @@ public override void Down()
                 .PrimaryKey(t => t.Key);
             
             AddColumn("dbo.PackageDeprecations", "CvssRating", c => c.Decimal(precision: 3, scale: 1));
+            Sql("ALTER TABLE [dbo].[PackageDeprecations] WITH CHECK ADD CONSTRAINT [dbo.PackageDeprecations_CVSSRating] CHECK (([CvssRating]>=(0) AND [CvssRating]<=(10)))");
```
joelverhagen: nit, this line should look exactly like:
https://github.com/NuGet/NuGetGallery/blob/aa628079fd17c94848700ad36f454c021503b2fe/src/NuGetGallery/Migrations/201901142143060_AddDeprecationEntities.cs#L36

right?

scottbommarito: The only difference is the extra parenthesis, right?

joelverhagen: And the spaces

xavierdecoster: Fixed in https://github.com/NuGet/NuGetGallery/pull/7182/commits/bc9d872dc819611115b1acb7eabced6d756c1a45

PR #7198, opened 2019-05-31
Wrap the Status endpoint in a Razor view

scottbommarito:
https://github.com/NuGet/Engineering/issues/2493

By wrapping our status response in a Razor view, we now force this endpoint to also wait for Razor initialization as well, which means that this endpoint will no longer return available before the instance is ready to handle Razor requests.

It also adds testing to the `HealthProbe` endpoint, which had the same change made to it as part of a hotfix.

### DIFF
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -265,6 +265,10 @@ namespace NuGetGallery
             return new RedirectResult(NuGetExeUrl, permanent: false);
         }
 
+        /// <remarks>
+        /// This endpoint is used by our traffic manager to determine whether or not this instance is available.
+        /// We are returning a Razor view because it forces this request to wait for Razor to be initialized before becoming available.
+        /// </remarks>
         [HttpGet]
         [ActionName("StatusApi")]
         public virtual async Task<ActionResult> Status()
@@ -279,6 +283,10 @@ namespace NuGetGallery
             return View(status);
         }
 
+        /// <remarks>
+        /// This endpoint is used by our load balancer to determine whether or not this instance is available.
+        /// We are returning a Razor view because it forces this request to wait for Razor to be initialized before becoming available.
+        /// </remarks>
         [HttpGet]
         [ActionName("HealthProbeApi")]
         public ActionResult HealthProbe()
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -13,7 +13,6 @@ using System.Net;
 using System.Threading.Tasks;
 using System.Web;
 using System.Web.Mvc;
-using System.Web.UI.WebControls;
 using Newtonsoft.Json.Linq;
 using NuGet.Frameworks;
 using NuGet.Packaging;
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -13,6 +13,7 @@ using System.Net;
 using System.Threading.Tasks;
 using System.Web;
 using System.Web.Mvc;
+using System.Web.UI.WebControls;
 using Newtonsoft.Json.Linq;
 using NuGet.Frameworks;
 using NuGet.Packaging;
@@ -273,7 +274,10 @@ namespace NuGetGallery
             {
                 return new HttpStatusCodeResult(HttpStatusCode.ServiceUnavailable, "Status service is unavailable");
             }
-            return await StatusService.GetStatus();
+
+            var status = await StatusService.GetStatus();
+            Response.StatusCode = (int)(status.GalleryServiceAvailable ? HttpStatusCode.OK : HttpStatusCode.ServiceUnavailable);
+            return View(status);
         }
 
         [HttpGet]
--- a/src/NuGetGallery/Services/StatusService.cs
+++ b/src/NuGetGallery/Services/StatusService.cs
@@ -4,16 +4,13 @@
 using System;
 using System.Collections.Generic;
 using System.Data.Entity;
-using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
-using System.Web.Mvc;
 using Microsoft.WindowsAzure.Storage.Blob;
 using Microsoft.WindowsAzure.Storage.RetryPolicies;
 using NuGetGallery.Configuration;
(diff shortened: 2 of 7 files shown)


src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -13,6 +13,7 @@
 using System.Threading.Tasks;
 using System.Web;
 using System.Web.Mvc;
+using System.Web.UI.WebControls;
```
joelverhagen: What is in this namespace that we need?

src/NuGetGallery/Views/Api/StatusApi.cshtml
```diff
@@ -0,0 +1,19 @@
...
+}
+
+<section role="main" class="container main-container">
+    <div class="row">
+        <div class="@ViewHelpers.GetColumnClasses(ViewBag)">
+            <h1>
+                Status
+            </h1>
+            <p>
+                NuGet Gallery instance @HostMachine.Name is @AvailabilityString(Model.GalleryServiceAvailable). SQL is @AvailabilityString(Model.SqlAzureAvailable). Storage is @AvailabilityString(Model.StorageAvailable).
```
joelverhagen: `StorageAvailable` can be null so this may say `Storage is .` right?

tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
```diff
@@ -15,6 +15,7 @@
 using System.Web.Mvc;
 using System.Web.Routing;
 using Autofac;
+using Microsoft.Ajax.Utilities;
```
joelverhagen: What is in this namespace that we need?

joelverhagen:
I like this solution. It is simple.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -273,14 +274,17 @@ public virtual async Task<ActionResult> Status()
             {
                 return new HttpStatusCodeResult(HttpStatusCode.ServiceUnavailable, "Status service is unavailable");
             }
-            return await StatusService.GetStatus();
+
+            var status = await StatusService.GetStatus();
```
joelverhagen: Can you leave a comment about why we are returning a view here? It's important this keeps using Razor.

tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
scottbommarito: no idea how it came in...will remove

src/NuGetGallery/Views/Api/StatusApi.cshtml
scottbommarito: nope--`@AvailabilityString(bool? isAvailable)` returns `available` if `isAvailable` is null

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: good call!

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -265,6 +265,10 @@ public virtual ActionResult GetNuGetExe()
             return new RedirectResult(NuGetExeUrl, permanent: false);
         }
 
+        /// <remarks>
+        /// This endpoint is used by our traffic manager to determine whether or not this instance is available.
+        /// We are returning a Razor view because it forces this request to wait for Razor to be initialized before becoming available.
+        /// </remarks>
         [HttpGet]
         [ActionName("StatusApi")]
         public virtual async Task<ActionResult> Status()
```
skofman1: as far as I understand, there is no issue with Status endpoint and with Traffic Manager availability. Why is this change needed? Changing HealthProbe should be sufficient.

skofman1:
To mitigate the current incident it's sufficient to release only the HealthProbe() change and deploy this as a hotfix

src/NuGetGallery/Services/StatusService.cs
```diff
@@ -41,22 +38,11 @@ public class StatusService : IStatusService
...
-            return new HttpStatusCodeWithBodyResult(AvailabilityStatusCode(galleryServiceAvailable),
-                String.Format(CultureInfo.InvariantCulture,
-                    StatusMessageFormat,
-                    AvailabilityMessage(galleryServiceAvailable),
-                    AvailabilityMessage(sqlAzureAvailable),
-                    AvailabilityMessage(storageAvailable),
-                    HostMachine.Name));
+            return new StatusViewModel(
+                IsSqlAzureAvailable(), 
+                await IsAzureStorageAvailable());
```
cristinamanum: The traffic manager endpoint check timeout is set now to 10 seconds. With this change are there any chances that this will take longer?

scottbommarito: 10 seconds should be plenty of time for a healthy instance to return the result. I don't think any change to the timeout is needed.

src/NuGetGallery/Controllers/ApiController.cs
scottbommarito: This issue will theoretically happen with our traffic manager as well, but it's far less likely.

I would still like to apply this fix to the endpoint as well. I will open another hotfix PR with the `HealthProbe` related changes only, and leave this one against dev.

scottbommarito: https://github.com/NuGet/NuGetGallery/pull/7199

scottbommarito:
@skofman1 - Opened hotfix via https://github.com/NuGet/NuGetGallery/pull/7199

scottbommarito:
@skofman1 - I have updated this PR with the hotfix

scottbommarito:
@skofman1 brought up the point offline that with the new load balancer behavior, an instance should never be in rotation if it doesn't respond to Razor requests, in which case these changes are unnecessary and just introduce risk. Although this is technically true, I still believe, from a theoretical perspective, that our traffic manager should fail over a region if it does not respond to Razor requests, and that we should enforce this in case the behavior of our load balancer changes without a corresponding traffic manager change. Given this, I would like to wait for the load balancer change to be deployed to PROD for a couple days and to verify that it has caused no issues or perf impact before I go ahead with this PR.

scottbommarito:
I looked into how the load balancer probe changes have affected INT.  The endpoint has gone from taking 9 ms to load on average to taking 12 ms to load. I think this is a fairly insignificant change. I will reanalyze when it hits PROD.

PR #7218, opened 2019-06-01
ODataReadOnlySupport

cristinamanum:
Add ODataReadOnlySupport


src/NuGetGallery.Core/Entities/ReadOnlyEntitiesContext.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Linq;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public class ReadOnlyEntitiesContext : IReadOnlyEntitiesContext
+    {
+        private readonly EntitiesContext _entitiesContext; 
+        
+        public ReadOnlyEntitiesContext(DbConnection connection)
```
skofman1: Is there value in making this testable? If so consider using GalleryDbContextFactory

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -53,7 +53,7 @@ public EntitiesContext(string connectionString, bool readOnly)
         public EntitiesContext(DbConnection connection, bool readOnly)
             : base(connection, contextOwnsConnection: true)
         {
-            ReadOnly = readOnly;
+           ReadOnly = readOnly;
```
joelverhagen: 4 spaces for indentation

src/NuGetGallery/Controllers/ODataV2FeedController.cs
```diff
@@ -597,5 +603,14 @@ where isSemVerLevelCompliant(p)
 
             return updates;
         }
+
+        private IQueryable<Package> GetAll()
```
skofman1: Is there a test to verify this code?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -562,6 +564,28 @@ private static DbConnection CreateDbConnection(ISqlConnectionFactory connectionF
...
+            ISecretInjector secretInjector)
+        {
+            var galleryDbReadOnlyReplicaConnectionFactory = CreateDbConnectionFactory(
+                diagnostics,
+                nameof(ReadOnlyEntitiesContext),
+                configuration.Current.SqlReadOnlyReplicaConnectionString ?? configuration.Current.SqlConnectionString,
+                secretInjector);
+
+            builder.Register(c => new ReadOnlyEntitiesContext(CreateDbConnection(galleryDbReadOnlyReplicaConnectionFactory)))
+                .AsSelf()
```
joelverhagen: An we not register `ReadOnlyEntitiesContext`? IMO you should only inject interfaces.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -562,6 +564,28 @@ private static DbConnection CreateDbConnection(ISqlConnectionFactory connectionF
...
+                .As<IReadOnlyEntitiesContext>()
+                .InstancePerLifetimeScope();
+
+            builder.RegisterType<ReadOnlyEntityRepository<Package>>()
+                .AsSelf()
```
joelverhagen: Same, I don't see why a service would import `ReadOnlyEntityRepository<Package>`

src/NuGetGallery.Core/Entities/IReadOnlyEntitiesContext.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Data.Entity;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public interface IReadOnlyEntitiesContext : IDisposable
+    {
+        DbSet<Package> Packages { get; set; }
+
+        DbSet<T> Set<T>() where T : class;
```
agr: Need to document the purpose of the method. Also name seems confusing: it's called `Set` but does not set anything, instead it returns something.

src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -281,5 +287,14 @@ public IHttpActionResult GetPropertyFromPackages(string propertyName, string id,
             var searchResults = await Search(options, searchTerm, targetFramework);
             return searchResults.FormattedAsCountResult<V1FeedPackage>();
         }
+
+        private IQueryable<Package> GetAll()
+        {
+            if (_featureFlagService.IsODataDatabaseReadOnlyEnabled())
```
agr: This should be resolved by DI, isn't it?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -562,6 +564,28 @@ private static DbConnection CreateDbConnection(ISqlConnectionFactory connectionF
...
+            ConfigurationService configuration,
+            ISecretInjector secretInjector)
+        {
+            var galleryDbReadOnlyReplicaConnectionFactory = CreateDbConnectionFactory(
+                diagnostics,
+                nameof(ReadOnlyEntitiesContext),
+                configuration.Current.SqlReadOnlyReplicaConnectionString ?? configuration.Current.SqlConnectionString,
+                secretInjector);
+
+            builder.Register(c => new ReadOnlyEntitiesContext(CreateDbConnection(galleryDbReadOnlyReplicaConnectionFactory)))
```
agr: DI should fallback to providing `EntitiesContext` instances when `IReadOnlyEntitiesContext` is requested when feature flag is disabled, so we would make this choice in a single location. Same for `ReadOnlyEntityRepository<Package>` below.

src/NuGetGallery.Core/Entities/ReadOnlyEntitiesContext.cs
cristinamanum: I do not think that it is worth. It is only redirecting to the EntitiesContext.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
cristinamanum: Fixed.

src/NuGetGallery/Controllers/ODataV1FeedController.cs
cristinamanum: After the feature flag will be removed this method will go away. If the feature flag is not on the read-write will be used.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
cristinamanum: The feature flags may be changed multiple times after startup. For this reason the decision is done in the controller.

Issue #7185, opened 2019-05-27
Package validations can be made much quicker with a "queue-back"


joelverhagen:
Downstream validators can queue back to orchestrator when they are done. This can save minutes. From testing on DEV, it can save on average roughly 2 minutes and 30 seconds per initial package validation.

1. [x] Apply Validation DB migration.
1. [x] Deploy orchestrators that handle the new message type.
    - https://github.com/NuGet/NuGetGallery/issues/7394
1. [x] Deploy symbol validator and symbol ingester.
1. [x] Turn on the feature flag.
1. [x] E2E tests and manual testing of symbol validation.
1. [x] Turn off the feature flag.
1. [x] Deploy revalidate certificate job.
1. [x] Deploy package validators.
1. [x] Deploy scan and sign submitter.
1. [x] Enable feature flag monitoring 

# Results

Over the past 7 days (4 days with queue-back, 3 days without queue-back), this is the improvement of validation times at a variety of percentiles. In short, we shaved off 2 minutes of validation time!

jobName | p50 | p90 | p95 | p99
-- | -- | -- | -- | --
Validation.SymbolsOrchestrator | 43.53355 | 134.0543 | 165.6078 | 190.0541
Validation.Orchestrator | 129.1221 | 158.1307 | 184.3262 | 216.7317

This is the P95 validation time, per 6 hours over the past 7 days.

![image](https://user-images.githubusercontent.com/94054/65446774-ed4ee580-dde9-11e9-9496-51fc500daf64.png)

## Queries

```
let enabledData = todatetime('2019-09-19T23:11:46.558');
let summary = materialize (
    customMetrics
    | where timestamp > ago(7d)
    | where name == "Orchestrator.TotalValidationDurationSeconds"
    | extend instanceName = tostring(customDimensions.InstanceName)
    | extend jobName = iff(instanceName startswith "Validation.Orchestrator", "Validation.Orchestrator", "Validation.SymbolsOrchestrator")
    | extend hasQueueBack = timestamp > enabledData
    | summarize avg(value), percentiles(value, 50, 90, 95, 99) by jobName, hasQueueBack
);
summary
| where hasQueueBack == false
| join (
   summary
   | where hasQueueBack == true
) on jobName
| project jobName,
          p50 = percentile_value_50 - percentile_value_501,
          p90 = percentile_value_90 - percentile_value_901,
          p95 = percentile_value_95 - percentile_value_951,
[...]

PR #7219, opened 2019-06-03
Update ServerCommon version to pull in latest validation DB migration

joelverhagen:
The new DB migration in validation DB adds a status column to validation sets. Also, polish the validation page a little bit.

Progress on https://github.com/NuGet/NuGetGallery/issues/7185.


joelverhagen:
Related to https://github.com/NuGet/NuGetGallery/issues/7223.

PR #7230, opened 2019-06-04
Moving ContentObject, User, Package Services

ryuyu:
This PR pulls the ContentObjectService, UserService, PackageService, and their dependencies into the library.

Please give this a careful look as there were merge conflicts.
I will be working closer to dev in the future, so hopefully, this will be the last one like this.

This was rebased onto dev off of a branch wayyyyy down from dev, so ignore the excess commits as they are already in dev, and squashing made them reappear here.


src/NuGetGallery.Services/Constants/GitConstants.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NuGetGallery.Services.Constants
+{
+    public static class GitConstants
```
scottbommarito: This file is very specific. I don't see us making any more `GitConstants`, so it seems a little unnecessary to make a whole file for this string. Is there another generic constants file that we can put this into? `ServiceConstants`?

ryuyu: You are absolutely right this should go into ServicesConstants.
Bit of background: this was initially moved here as I was planning on splitting the constants up a little bit, but that started to get unwieldy, so I just started moving stuff into ServicesConstants, and forgot to move this little bit. Good catch.

src/NuGetGallery.Services/Properties/AssemblyInfo.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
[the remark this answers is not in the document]
ryuyu: Tried, but there doesn't appear to be anything there for me. Formatted document and there were no changes either.

src/NuGetGallery.Services/Configuration/ILoginDiscontinuationConfiguration.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.Entities;
+using NuGetGallery.Authentication;
+
+namespace NuGetGallery
+{
+    public interface ILoginDiscontinuationConfiguration
```
agr: Documentation for class and/or methods?

src/NuGetGallery.Services/Properties/AssemblyInfo.cs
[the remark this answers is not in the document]
agr: Windows line break VS Unix one?

PR #7238, opened 2019-06-05
Enforcing thread safe implementation of `RandomNumberGenerator` for V3Hasher

agr:
Addresses https://github.com/NuGet/Engineering/issues/2265.

The [`RandomNumberGenerator.GetBytes`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rngcryptoserviceprovider.getbytes?view=netframework-4.6.2#System_Security_Cryptography_RNGCryptoServiceProvider_GetBytes_System_Byte___) is not marked as thread safe. The default implementation of that abstract class *is* [thread safe](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rngcryptoserviceprovider.getbytes?view=netframework-4.6.2#System_Security_Cryptography_RNGCryptoServiceProvider_GetBytes_System_Byte___).

This change enforces it.


scottbommarito:
Did you ever consider just creating a new `RandomNumberGenerator` every time that method is called?

Why do we need to reuse the same instance across threads?
Is there a high cost to setting up the object?

If possible, I would prefer we just keep things simple and avoid the issue of thread-safety entirely.

agr:
@scottbommarito, I didn't. This code is borrowed from some other project and was security reviewed, I didn't want to make any functional changes to it.

Issue #7130, opened 2019-05-02
Status page degraded icon is invisible with a white background


karann-msft:
v2 protocol is degraded and it has a ❗️ but it is hard to see.
![image](https://user-images.githubusercontent.com/16904420/57106736-0348f600-6ce3-11e9-83bc-20a4130dfe5b.png)

PR #7244, opened 2019-06-05
Merge Status page degraded icon should be orange for subcomponents (#7243) to master

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7130

This change has no functional difference for the Gallery--I would just like to keep the `NuGet.Services.Status` submodule reference on `master` instead of `dev`.


scottbommarito:
Just kidding-- @ryuyu merged his changes to `dev` and this PR is now broken.

Issue #6854, opened 2019-01-26
Add telemetry for package deprecation


scottbommarito:
![image.png](https://images.zenhubusercontent.com/58054e1495e42b6c5baa9ebb/387f6b91-7223-4d4b-945d-8a44bb48301f)

scottbommarito:
I have discovered that we need to add a new event to the `ITelemetryService` for deprecation to appear on the `PackageEvents` dashboard.

scottbommarito:
Can't verify on PROD due to feature flag.

PR #7245, opened 2019-06-06
Add "PackageDeprecate" telemetry event

scottbommarito:
https://github.com/nuget/nugetgallery/issues/6854

We need to emit a new event so that it can show in the PackageEvents dashboard (https://github.com/nuget/engineering/issues/2477).

### DIFF
--- a/src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
+++ b/src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
@@ -34,6 +34,13 @@ namespace NuGetGallery
 
         void TrackPackageRevalidate(Package package);
 
+        void TrackPackageDeprecate(
+            IReadOnlyList<Package> packages,
+            PackageDeprecationStatus status,
+            bool hasAlternateRegistration,
+            bool hasAlternatePackage,
+            bool hasCustomMessage);
+
         void TrackPackageReadMeChangeEvent(Package package, string readMeSourceType, PackageEditReadMeState readMeState);
 
         void TrackCreatePackageVerificationKeyEvent(string packageId, string packageVersion, User user, IIdentity identity);
--- a/src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
+++ b/src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
@@ -37,8 +37,8 @@ namespace NuGetGallery
         void TrackPackageDeprecate(
             IReadOnlyList<Package> packages,
             PackageDeprecationStatus status,
-            bool hasAlternateRegistration,
-            bool hasAlternatePackage,
+            PackageRegistration alternateRegistration,
+            Package alternatePackage,
             bool hasCustomMessage);
 
         void TrackPackageReadMeChangeEvent(Package package, string readMeSourceType, PackageEditReadMeState readMeState);
--- a/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
+++ b/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
@@ -17,18 +17,15 @@ namespace NuGetGallery
     public partial class ManageDeprecationJsonApiController
         : AppController
     {
-        private readonly IAuditingService _auditingService;
         private readonly IPackageService _packageService;
         private readonly IPackageDeprecationService _deprecationService;
         private readonly IFeatureFlagService _featureFlagService;
 
         public ManageDeprecationJsonApiController(
-            IAuditingService auditingService,
             IPackageService packageService,
             IPackageDeprecationService deprecationService,
             IFeatureFlagService featureFlagService)
         {
-            _auditingService = auditingService ?? throw new ArgumentNullException(nameof(auditingService));
             _packageService = packageService ?? throw new ArgumentNullException(nameof(packageService));
             _deprecationService = deprecationService ?? throw new ArgumentNullException(nameof(deprecationService));
             _featureFlagService = featureFlagService ?? throw new ArgumentNullException(nameof(featureFlagService));
@@ -168,16 +165,6 @@ namespace NuGetGallery
                 alternatePackage,
                 customMessage,
                 currentUser);
-
-            foreach (var packageToUpdate in packagesToUpdate)
-            {
-                await _auditingService.SaveAuditRecordAsync(
-                    new PackageAuditRecord(
-                        packageToUpdate,
-                        status == PackageDeprecationStatus.NotDeprecated ? AuditedPackageAction.Undeprecate : AuditedPackageAction.Deprecate,
-                        status == PackageDeprecationStatus.NotDeprecated ? PackageUndeprecatedVia.Web : PackageDeprecatedVia.Web));
-
-            }
 
             return Json(HttpStatusCode.OK);
         }
--- a/src/NuGetGallery/Services/PackageDeprecationService.cs
+++ b/src/NuGetGallery/Services/PackageDeprecationService.cs
@@ -113,7 +113,7 @@ namespace NuGetGallery
                     status,
                     alternatePackageRegistration != null,
                     alternatePackage != null,
-                    string.IsNullOrWhiteSpace(customMessage));
+                    !string.IsNullOrWhiteSpace(customMessage));
 
                 foreach (var package in packages)
                 {
--- a/src/NuGetGallery/Services/PackageDeprecationService.cs
+++ b/src/NuGetGallery/Services/PackageDeprecationService.cs
@@ -111,8 +111,8 @@ namespace NuGetGallery
                 _telemetryService.TrackPackageDeprecate(
                     packages,
                     status,
-                    alternatePackageRegistration != null,
-                    alternatePackage != null,
+                    alternatePackageRegistration,
+                    alternatePackage,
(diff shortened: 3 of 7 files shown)


src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -169,16 +166,6 @@ public virtual JsonResult GetAlternatePackageVersions(string id)
                 customMessage,
                 currentUser);
 
-            foreach (var packageToUpdate in packagesToUpdate)
```
scottbommarito: I felt like it makes more sense to emit the auditing in the `PackageDeprecationService` instead of in the controller, to be more consistent with `MarkPackageListedAsync` and `MarkPackageUnlistedAsync` in `PackageService`.

xavierdecoster:
Changes LGTM. Some tests are failing though.

src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
```diff
@@ -34,6 +34,13 @@ public interface ITelemetryService
 
         void TrackPackageRevalidate(Package package);
 
+        void TrackPackageDeprecate(
+            IReadOnlyList<Package> packages,
+            PackageDeprecationStatus status,
+            bool hasAlternateRegistration,
```
joelverhagen: Why not just record the registration and alternate package and do the `!= null` check in AI? This is not PII and allows detailed information.

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -900,5 +960,10 @@ protected virtual void TrackMetric(string metricName, double value, Action<Dicti
 
             _telemetryClient.TrackMetric(metricName, value, telemetryProperties);
         }
+
+        private string BuildArrayProperty(IEnumerable<string> list)
+        {
+            return string.Join(", ", list);
```
joelverhagen: Why not do a JSON array? This is indexable on AI but a comma separated list is not.

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -648,6 +675,39 @@ private static string GetApiKeyCreationDate(User user, IIdentity identity)
             });
         }
 
+        private void TrackMetricForPackages(
```
loic-sharma: Could you explain why you made these helper methods? Considering no one else is using these today, can we just inline all of these helpers' code into `TrackPackageDeprecate`?

scottbommarito: It's very likely they will be used for bulk unlist/relist.

loic-sharma: I would reassess when if/we do that work.

src/NuGetGallery/Services/TelemetryService.cs
scottbommarito: I tried to do what was the existing pattern in the code. Any time we have an array, we serialize it by comma-separating it.

I agree with you that this is better, however. Do you think I should change it for all the events that serialize arrays or just the new one?

src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
scottbommarito: I did it like this so that we can surface the information in Geneva, which doesn't support infinite cardinality on dimensions. I was under the impression that any dimensions that are infinitely cardinal must be removed from our AI -> Geneva connector. Is this incorrect? Maybe I should add both a `hasAlternateRegistration` field and a `alternateRegistrationId` field?

src/NuGetGallery/Services/TelemetryService.cs
scottbommarito: I would rather keep it like this, primarily because it mirrors the call structure from the individual package event path (e.g. `TrackPackagePushEvent` calls `TrackMetricForPackage`). I worry that if someone wants to make another metric for multiple packages, they will not see the deprecate method and will instead create their own multiple package event, whereas keeping this helper method will make it more difficult for that to happen.

src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
joelverhagen: This is only true when we make pre-aggregates on that dimension, AFAIK.

scottbommarito: That was my plan...in retrospect, I think it's likely this data will only be queried through AI anyway, in which case the dimension concern is no longer valid. I will change this to log the ID and version.

src/NuGetGallery/Services/TelemetryService.cs
joelverhagen: Just the new one.

joelverhagen: Well, maybe ask the owners of the existing events to get opinions.

scottbommarito: Will ask the owners for confirmation.

scottbommarito: Got confirmation from @cristinamanum and @zhhyu.

scottbommarito:
@joelverhagen @loic-sharma - I've handled your comments, please take a look again.

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -900,5 +961,10 @@ protected virtual void TrackMetric(string metricName, double value, Action<Dicti
 
             _telemetryClient.TrackMetric(metricName, value, telemetryProperties);
         }
+
+        private string BuildArrayProperty(IEnumerable<string> list)
+        {
+            return JsonConvert.SerializeObject(list);
```
joelverhagen: just make sure this isn't pretty JSON. No point in that.

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -648,6 +676,39 @@ private static string GetApiKeyCreationDate(User user, IIdentity identity)
             });
         }
 
+        private void TrackMetricForPackages(
+            string metricName,
+            IReadOnlyList<Package> packages,
+            Action<Dictionary<string, string>> addProperties = null)
+        {
+            if (packages == null || !packages.Any() || packages.Select(p => p.PackageRegistrationKey).Distinct().Count() > 1)
```
cristinamanum: Should not be ....Count() >= 1?

src/NuGetGallery/Services/TelemetryService.cs
```diff
@@ -648,6 +676,39 @@ private static string GetApiKeyCreationDate(User user, IIdentity identity)
...
+            }
+
+            TrackMetricForPackages(
+                metricName,
+                packages.First().PackageRegistration.Id,
+                packages.Select(p => p.NormalizedVersion).ToList(),
+                addProperties);
+        }
+
+        private void TrackMetricForPackages(
```
cristinamanum: Should it be called TrackMetricForPackage or TrackMetricForPackageVersions instead of TrackMetricForPackages?

src/NuGetGallery/Services/TelemetryService.cs
scottbommarito: This check verifies that, of the list of packages passed in (`IReadOnlyList<Package>`), these packages must have the same registration.

Doing `Count() >= 1` would fail lists of packages with the same registration as well.

src/NuGetGallery/Services/TelemetryService.cs
scottbommarito: I like `TrackMetricForPackageVersions` because it conveys that they must be the same ID. I will rename it to that.

src/NuGetGallery/Services/TelemetryService.cs
scottbommarito: I confirmed that `JsonConvert` returns "ugly JSON"--no whitespace.

src/NuGetGallery/Services/PackageDeprecationService.cs
```diff
@@ -84,6 +91,23 @@ public class PackageDeprecationService : IPackageDeprecationService
...
+                alternatePackage,
+                !string.IsNullOrWhiteSpace(customMessage));
+
+            foreach (var package in packages)
+            {
+                await _auditingService.SaveAuditRecordAsync(
+                    new PackageAuditRecord(
+                        package,
+                        status == PackageDeprecationStatus.NotDeprecated ? AuditedPackageAction.Undeprecate : AuditedPackageAction.Deprecate,
+                        status == PackageDeprecationStatus.NotDeprecated ? PackageUndeprecatedVia.Web : PackageDeprecatedVia.Web));
```
agr: Is there non "via.web" variant?

scottbommarito: It's very subtle, but one of the options is `UndeprecatedVia` and the other is `DeprecatedVia`.

I think part of the motivation behind this was that we will probably have CLI deprecation eventually, at which point we would need a `DeprecatedVia.Cli` option or something.

src/NuGetGallery/Services/TelemetryService.cs
cristinamanum: Right

PR #7247, opened 2019-06-06
Add deployment ID value to Application Insights telemetry

joelverhagen:
This can help us distinguish between USNC-USSC, production-staging slot. This is the same value we see in our internal telemetry system for "tenant".

The actual name of the slot and cloud service resource is not directly available. See:
https://stackoverflow.com/questions/24522781/get-cloud-service-name-in-a-web-role

I am in the process of confirming this on DEV.


joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7250.

Issue #7248, opened 2019-06-06
Google Analytics click events are not emitted on middle-mouse click


joelverhagen:
This is a common way to open pages in a new tab.

PR #7249, opened 2019-06-06
Emit Google Analytics click event on middle-mouse click

joelverhagen:
Discovered when testing out search click index telemetry. With this change, we now emit our "click" event when someone uses the middle-mouse click on a tracked link.

I left the existing logic for left-click.

Address https://github.com/NuGet/NuGetGallery/issues/7248

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/common.js
+++ b/src/NuGetGallery/Scripts/gallery/common.js
@@ -406,7 +406,7 @@
         if (window.nuget.isGaAvailable()) {
             ga('send', 'event', category, action, label, eventValue, options);
         }
-    }
+    };
 
     window.nuget = nuget;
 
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -447,25 +447,37 @@
...
+                return;
+            }
+
+            var href = $(this).attr('href');
+            var category = $(this).data().track;
+            var trackValue = $(this).data().trackValue;
+            if (href && category) {
+                if (emitDirectly) {
+                    window.nuget.sendAnalyticsEvent(category, 'click', href, trackValue);
+                } else {
```
loic-sharma: What is the purpose of this `else` branch? I know it's existing code, just curious :)

loic-sharma: It may be worth adding a comment if this is something fancy

joelverhagen: When you left click a link it changes the current tab and any running JavaScript (e.g. `sendAnalyticsEvent`) is interrupted. This fork emits the telemetry before navigating to the new URL.

joelverhagen: Added comment.

Issue #7222, opened 2019-06-03
[Symbols Improvement] Unclear UI workflow on new symbols publish failure 


cristinamanum:
When new symbols package is uploaded and the upload fails it is not clear what happened with the old correct published symbols. 

The current behavior:
1. The user has the error message that the symbols failed to be published. 
2. The Download Symbols button is not available anymore. 

While this is fine for new symbol packages, if previous symbols were already published it is not clear if the old symbols are still available. 
The old symbols are still available but the UI could lead the user to the conclusion that they are not. 

More details see issue:
https://github.com/NuGet/NuGetGallery/issues/7188

304NotModified:
Also, "Download Symbols" is then unavailable.

shishirx34:
@304NotModified - I am simplifying this issue with appropriate message and making the download button available, see the screenshot below. Does this help in removing any confusion you might have with regards to this flow of symbols upload?

![image](https://user-images.githubusercontent.com/1646506/59068522-2da14c80-8869-11e9-94dc-b724ead432ee.png)

shishirx34:
Note: this message and download button will also be available when the newer uploaded symbols package are being indexed and validated.

304NotModified:
> - I am simplifying this issue with appropriate message and making the download button available, see the screenshot below. Does this help in removing any confusion you might have with regards to this flow of symbols upload?

That's a real improvement! Would it be possible to only show the warning to the (logged in) owners of the package?

304NotModified:
Another option would be to remove the warning and just show the last datetime when the symbols are updated.

shishirx34:
@304NotModified - I think its fair to show this message only to the logged in owners of the package and not everyone.

Owners:
![image](https://user-images.githubusercontent.com/1646506/59069739-f03ebe00-886c-11e9-978d-636516799616.png)

Everyone else (note: download symbols is visible):
![image](https://user-images.githubusercontent.com/1646506/59069752-fdf44380-886c-11e9-9519-9ed27ad010dd.png)

304NotModified:
Great!

PR #7252, opened 2019-06-06
Allow published symbols to be downloaded after re-upload failures and Blog updates

shishirx34:
Allow the pre-published symbols to be downloaded after newer symbols are being published/failed validation.

See https://github.com/NuGet/NuGetGallery/issues/7222#issuecomment-499686460 snapshots

Also, updated the link for "Blog" header to point to new blogs pages.

### DIFF
--- a/src/NuGet.Services.Entities/Extensions/PackageExtensions.cs
+++ b/src/NuGet.Services.Entities/Extensions/PackageExtensions.cs
@@ -14,10 +14,7 @@ namespace NuGet.Services.Entities
         /// <returns>The latest symbol package if present, null otherwise</returns>
         public static SymbolPackage LatestSymbolPackage(this Package package)
         {
-            return package
-                .SymbolPackages?
-                .OrderByDescending(sp => sp.Created)
-                .FirstOrDefault();
+            return package.GetSymbolsPackagesInReverseOrder()?.FirstOrDefault();
         }
 
         /// <summary>
@@ -25,12 +22,14 @@ namespace NuGet.Services.Entities
         /// </summary>
         /// <param name="package"><see cref="Package"/> for which latest symbol package is to be retrieved.</param>
         /// <returns>The last available symbol package if present, null otherwise</returns>
-        public static SymbolPackage LastAvailableSymbolPackage(this Package package)
+        public static SymbolPackage LatestAvailableSymbolPackage(this Package package)
         {
-            return package
-                .SymbolPackages?
-                .OrderByDescending(sp => sp.Created)
-                .FirstOrDefault(sp => sp.StatusKey == PackageStatus.Available);
+            return package.GetSymbolsPackagesInReverseOrder()?.FirstOrDefault(sp => sp.StatusKey == PackageStatus.Available);
+        }
+
+        public static IOrderedEnumerable<SymbolPackage> GetSymbolsPackagesInReverseOrder(this Package package)
+        {
+            return package.SymbolPackages?.OrderByDescending(sp => sp.Created);
         }
 
         /// <summary>
--- a/src/NuGet.Services.Entities/Extensions/PackageExtensions.cs
+++ b/src/NuGet.Services.Entities/Extensions/PackageExtensions.cs
@@ -27,7 +27,7 @@ namespace NuGet.Services.Entities
             return package.GetSymbolsPackagesInReverseOrder()?.FirstOrDefault(sp => sp.StatusKey == PackageStatus.Available);
         }
 
-        public static IOrderedEnumerable<SymbolPackage> GetSymbolsPackagesInReverseOrder(this Package package)
+        private static IOrderedEnumerable<SymbolPackage> GetSymbolsPackagesInReverseOrder(this Package package)
         {
             return package.SymbolPackages?.OrderByDescending(sp => sp.Created);
         }
--- a/src/NuGet.Services.Entities/Extensions/PackageExtensions.cs
+++ b/src/NuGet.Services.Entities/Extensions/PackageExtensions.cs
@@ -21,6 +21,19 @@ namespace NuGet.Services.Entities
         }
 
         /// <summary>
+        /// Get the last successfully uploaded symbols package for the given package.
+        /// </summary>
+        /// <param name="package"><see cref="Package"/> for which latest symbol package is to be retrieved.</param>
+        /// <returns>The last available symbol package if present, null otherwise</returns>
+        public static SymbolPackage LastAvailableSymbolPackage(this Package package)
+        {
+            return package
+                .SymbolPackages?
+                .OrderByDescending(sp => sp.Created)
+                .FirstOrDefault(sp => sp.StatusKey == PackageStatus.Available);
+        }
+
+        /// <summary>
         /// Returns true if there exists a symbol package which is latest and is in
         /// available state, false otherwise.
         /// </summary>
--- a/src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
+++ b/src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
@@ -35,6 +35,9 @@ namespace NuGetGallery
             PackageFileSize = package.PackageFileSize;
 
             LatestSymbolsPackage = package.LatestSymbolPackage();
+            LastAvailableSymbolsPackage = LatestSymbolsPackage.StatusKey == PackageStatus.Available
+                ? LatestSymbolsPackage
+                : package.LastAvailableSymbolPackage();
 
             if (packageHistory.Any())
             {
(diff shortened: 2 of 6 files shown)


src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -114,6 +117,7 @@ private DisplayPackageViewModel(Package package, User currentUser, string pushed
         public int TotalDaysSinceCreated { get; private set; }
         public long PackageFileSize { get; private set; }
         public SymbolPackage LatestSymbolsPackage { get; private set; }
+        public SymbolPackage LastAvailableSymbolsPackage { get; private set; }
```
scottbommarito: I would suggest naming this `LatestAvailableSymbolsPackage` instead, for consistency with `LatestSymbolsPackage`

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -323,6 +323,12 @@
                             <strong>The symbols for this package have not been indexed yet.</strong> They are not available
                             for download from the NuGet.org symbol server. Symbols will be available for download after both validation and indexing are complete. 
                             Symbols validation and indexing may take up to an hour. <a href="https://aka.ms/NuGetSymbolsPackageValidation">Read more</a>.
+
+                            @if(hasSymbolsPackageAvailable)
```
scottbommarito: Can you get rid of this duplication here? I would suggest you make a `@helper` that emits this and call it in both `if (Model.LatestSymbolsPackage.StatusKey == PackageStatus.Validating)` and `else if (Model.LatestSymbolsPackage.StatusKey == PackageStatus.FailedValidation)`

src/NuGet.Services.Entities/Extensions/PackageExtensions.cs
```diff
@@ -20,6 +20,19 @@ public static SymbolPackage LatestSymbolPackage(this Package package)
                 .FirstOrDefault();
         }
 
+        /// <summary>
+        /// Get the last successfully uploaded symbols package for the given package.
+        /// </summary>
+        /// <param name="package"><see cref="Package"/> for which latest symbol package is to be retrieved.</param>
+        /// <returns>The last available symbol package if present, null otherwise</returns>
+        public static SymbolPackage LastAvailableSymbolPackage(this Package package)
```
scottbommarito: `LatestAvailableSymbolPackage`, for consistency?

src/NuGet.Services.Entities/Extensions/PackageExtensions.cs
```diff
@@ -20,6 +20,19 @@ public static SymbolPackage LatestSymbolPackage(this Package package)
...
+        {
+            return package
```
scottbommarito: Could you share this LINQ code between `LatestSymbolPackage` and this method?

E.g.
```
public static SymbolPackage LatestSymbolPackage(this Package package)
{
    return package.GetSymbolsPackagesInReverseOrder().FirstOrDefault();
}

public static SymbolPackage LastAvailableSymbolPackage(this Package package)
{
    return package.GetSymbolsPackagesInReverseOrder().FirstOrDefault(sp => sp.StatusKey == PackageStatus.Available);
}

public static IEnumerable<SymbolPackage> GetSymbolsPackagesInReverseOrder(this Package package)
{
    return package.SymbolPackages?.OrderByDescending(sp => sp.Created);
}
```

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -86,9 +86,9 @@
                         {
                             @DisplayNavigationItem("Admin", Url.Admin())
                         }
-                        @DisplayNavigationItem("Documentation", "https://docs.microsoft.com/en-us/nuget/what-is-nuget")
+                        @DisplayNavigationItem("Documentation", "https://docs.microsoft.com/en-us/nuget/what-is-nuget", newWindow: true)
                         @DisplayNavigationItem("Downloads", Url.Downloads())
-                        @DisplayNavigationItem("Blog", "https://blog.nuget.org/")
+                        @DisplayNavigationItem("Blog", "https://devblogs.microsoft.com/nuget/", newWindow: true)
```
loic-sharma: Please move this to its own PR. I would also put this behind a feature flag.

shishirx34: Why? we want to go live with it the next time we deploy. On Monday we are going to be going live with CDN redirection for blogs as is.

scottbommarito: I don't think it's a huge deal either way, but it's a little easier to approve these changes if they are separate. Additionally, if it's important that the blog changes go out on Monday, the published symbols PR could potentially hold it back from being merged and vice versa.

loic-sharma: For "why should it be in its own PR?": Scott covered all the major points. I'd add that it's easier to track the work items that need to be verified if the dev->master merge has a commit for each item that needs to be verified.

For "why should it have a feature flag?": If we don't have a feature flag and the blog has issues, we'll have to rollback the gallery. All other Gallery changes will be affected. As we said on SharePoint, "don't stop the train".

shishirx34: To be clear, the redirection CDN changes are going live on Monday, the gallery changes can go out whenever. This PR isn't going to hold off that deployment. I do not think that the simple blog link is a big feature change to warrant a feature flag. I am comfortable with merging it with this PR.

src/NuGet.Services.Entities/Extensions/PackageExtensions.cs
```diff
@@ -14,10 +14,22 @@ public static class PackageExtensions
...
+        /// Get the last successfully uploaded symbols package for the given package.
+        /// </summary>
+        /// <param name="package"><see cref="Package"/> for which latest symbol package is to be retrieved.</param>
+        /// <returns>The last available symbol package if present, null otherwise</returns>
+        public static SymbolPackage LatestAvailableSymbolPackage(this Package package)
+        {
+            return package.GetSymbolsPackagesInReverseOrder()?.FirstOrDefault(sp => sp.StatusKey == PackageStatus.Available);
+        }
+
+        public static IOrderedEnumerable<SymbolPackage> GetSymbolsPackagesInReverseOrder(this Package package)
```
joelverhagen: Can be private

src/NuGetGallery/Views/Shared/Gallery/Header.cshtml
```diff
@@ -35,10 +35,10 @@
     </div>
 }
 
-@helper DisplayNavigationItem(string tab, string url, bool bold = false, string title = null)
+@helper DisplayNavigationItem(string tab, string url, bool bold = false, string title = null, bool newWindow = false)
 {
     <li class="@(ViewBag.Tab == tab ? "active" : string.Empty)" role="presentation">
-        <a role="tab" aria-selected="@(ViewBag.Tab == tab ? "true" : "false")" href="@url" title="@title">
+        <a role="tab" target= "@(newWindow ? "_blank" : "")" aria-selected="@(ViewBag.Tab == tab ? "true" : "false")" href="@url" title="@title">
```
joelverhagen: Why target "_blank"? This is still our property... at least from the user's perspective. IMO we should be sparing with opening in a new tab since it weakens the user's options to middle click/etc to open in a new window or left click to open in the same window. We don't even have target blank on project URL/license URL/repo URL.

Interesting conversation:
https://ux.stackexchange.com/questions/5397/whats-your-take-on-target-blank

In short I think we should avoid it.

shishirx34: Interesting read, surprisingly counter intuitive. Here's my take on why to open a new window. There is a breaking change with the new blogs such that the new site will not have the header to navigate back to the nuget.org website(or the links in the headers), which we had previously, only way would be to go back with several back clicks if browsing multiple blog posts. To keep the experience same, I think we can keep "Documentation" open in the same window, but I still feel like opening blogs in new window should be useful for the user since now its technically an external site(not even same domain). @karann-msft thoughts?

agr: Firefox even have an "advanced" setting that makes it ignore all `target="_blank"`. Opening links in new tabs is annoying and takes control from user.

scottbommarito: I agree with @joelverhagen and @agr: I think the small improvement in experience in case the user wants to navigate back to nuget.org quickly is not worth the confusion and loss of control that the user will experience.

Even if they **do** want to navigate back to nuget.org, they're going to not expect the new window when they click the tab at the top of the page. All of the other tabs on the page open the in the same window, and the difference in behavior is not going to be clear to users until they click it. Furthermore, users that have gone to the blog before will expect it to open in the same window will be confused when it opens a new window.

Additionally, I don't even think that navigating to the blog and then navigating back to the site is a common scenario. I would expect that most people that want to go to the blog just want to read articles on the blog--they don't actually want to do anything on nuget.org but follow the link to the blog, at which point opening the blog in a new window becomes a nuisance to them.

At least with Documentation it seems possible that someone might be going to the docs just to read about something they're doing on nuget.org, and then go back to nuget.org to do that thing. I don't see this scenario with blogs.

Issue #7203, opened 2019-05-31
[GitHub Usage] Add initial Gallery UI


loic-sharma:
Add UI on Gallery, hidden behind feature flag. Add Google Analytics.

PR #7257, opened 2019-06-10
[GiHub Usage] Add Gallery UI

mogah:
This PR adds only the first part of the UI. The second part would include Google Analytics.
image

https://github.com/NuGet/NuGetGallery/issues/7203


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -517,6 +517,49 @@
...
+                        <p>
+                            About @(Model.GitHubDependenciesInformation.TotalRepos.ToKiloFormat()) popular GitHub repositories depend on
+                            @(Model.Id), including the following:
+                        </p>
+
+                        @foreach (var dep in Model.GitHubDependenciesInformation.Repos)
+                        {
+                            <div class="row top-buffer">
+                                <div class="col-xs-12">
+                                    <button class="btn btn-gh-repo text-left btn-block"
```
scottbommarito: I would use `<a href="@dep.Url"` instead--needing to have Javascript in the `onclick` is a little ugly.

Also, make sure this has a `title` property so that screen readers have a description of where this links to.

tests/NuGetGallery.Core.Facts/GitHub/GitHubUsageConfigurationFacts.cs
```diff
@@ -0,0 +1,167 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace NuGetGallery.GitHub
+{
+    public class GitHubUsageConfigurationFacts
+    {
+        private static RepositoryInformation GenRepo(string id, int stars = 100, params string[] nugetDependencies)
```
scottbommarito: I would suggest `CreateRepo` instead of `GenRepo`--I haven't seen that abbreviation in our codebase before and less acronyms is less confusion.

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery.GitHub
```
scottbommarito: Typically what we do is put all of the files in this project into the same `namespace`.
I don't really have a problem with changing this (in fact I think @ryuyu was thinking of doing the same thing) but we should agree, as a team, that this is the right approach.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -504,11 +504,17 @@ public async Task GivenAValidatingPackageThatTheCurrentUserDoesNotOwnThenHideIt(
                 var deprecationService = new Mock<IPackageDeprecationService>();
                 var httpContext = new Mock<HttpContextBase>();
                 var httpCachePolicy = new Mock<HttpCachePolicyBase>();
+                var contentObjectService = new Mock<IContentObjectService>();
```
scottbommarito: Do you have any tests that check that the repository information is properly passed on to the model if the feature flag is enabled?

tests/NuGetGallery.Facts/Extensions/NumberExtensionsFacts.cs
```diff
@@ -29,5 +29,38 @@ public void FormatsUsingExpectedUnit(long bytes, string expected)
...
+            [InlineData(9999, "9.99K")]
+            [InlineData(10_000, "10.0K")]
+            [InlineData(10_990, "10.9K")]
+            [InlineData(99_990, "99.9K")]
+            [InlineData(100_000, "100K")]
+            [InlineData(100_990, "100K")]
+            [InlineData(100_999, "100K")]
+            [InlineData(999_000, "999K")]
+            [InlineData(999_999, "999K")]
+            [InlineData(1_000_000, "1.00M")]
```
scottbommarito: Is `1_000_000` a way to write `1000000` in C#?
Never seen that before, looks pretty cool.

src/NuGetGallery.Core/GitHub/IGitHubUsageConfiguration.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        /// <summary>
+        /// Returns a NuGetPackageGitHubInformation object that contains the information about a NuGet package.
+        /// NOTE: If a packageId has no information, the NuGetPackageGitHubInformation's TotalRepos will be 0
+        /// and the Repos list will be empty
+        /// 
+        /// throws an ArgumentException if the packageId is null
+        /// </summary>
+        /// <param name="packageId">NuGet package id, cannot be null</param>
+        /// <returns></returns>
```
scottbommarito: Either add some information to this tag or remove it.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -752,6 +753,9 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
             model.IsCertificatesUIEnabled = _contentObjectService.CertificatesConfiguration?.IsUIEnabledForUser(currentUser) ?? false;
             model.IsAtomFeedEnabled = _featureFlagService.IsPackagesAtomFeedEnabled();
             model.IsPackageDeprecationEnabled = _featureFlagService.IsManageDeprecationEnabled(currentUser);
+            model.IsGitHubUsageEnabled = _featureFlagService.IsGitHubUsageEnabled(currentUser);
+
+            model.GitHubDependenciesInformation = _contentObjectService.GitHubUsageConfiguration.GetPackageInformation(id);
```
scottbommarito: Have you considered not putting this data in the model if the feature flag is disabled? I think this is fine but just a thought.

joelverhagen:
Screenshot please

src/NuGetGallery/Extensions/NumberExtensions.cs
```diff
@@ -66,5 +67,40 @@ public static string ToUserFriendlyBytesLabel(this long bytes)
 
             return "0 bytes";
         }
+
+        /// <summary>
+        /// Format the number to a 3 digit representation plus a letter to represent the scale (K for kilo, M for mega, or G for giga)
+        /// </summary>
+        /// <param name="number"></param>
```
joelverhagen: Fill out param and returns or remove

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: I think it's best to make the feature flag more powerful in this regard. 

---
In reply to: [292238816](https://github.com/NuGet/NuGetGallery/pull/7257#discussion_r292238816) [](ancestors = 292238816)

src/NuGetGallery/Services/IFeatureFlagService.cs
```diff
@@ -52,5 +52,13 @@ public interface IFeatureFlagService
         /// Whether the user is able to access the search side-by-side experiment.
         /// </summary>
         bool IsSearchSideBySideEnabled(User user);
+
+        /// <summary>
+        /// Whether a user can see the "GitHub Usage" section in a package's display page as well as 
+        /// well as the added "GitHub Usage count" in the "Statistics" section
+        /// </summary>
+        /// <param name="user"></param>
```
joelverhagen: delete/fill out param and returns

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
joelverhagen: I think we should be consistent with our current approach (most everything in NuGetGallery namespace) until we design good buckets

---
In reply to: [292237967](https://github.com/NuGet/NuGetGallery/pull/7257#discussion_r292237967) [](ancestors = 292237967)

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Collections.Generic;
```
joelverhagen: license header

joelverhagen:
> namespace NuGetGallery.GitHub

license header

---
Refers to: src/NuGetGallery.Core/GitHub/IGitHubUsageConfiguration.cs:1 in 229d173. [](commit_id = 229d17330ec98397f82642bcd29744ceadacad8c, deletion_comment = False)

src/NuGetGallery.Core/GitHub/RepositoryInformation.cs
```diff
@@ -0,0 +1,46 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+
+namespace NuGetGallery.GitHub
+{
+    public class RepositoryInformation
+    {
+        public RepositoryInformation(
```
joelverhagen: Wait you don't need [JsonConstructor] here?

src/NuGetGallery.Core/GitHub/NuGetPackageGitHubInformation.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            if( repos == null)
+            {
+                throw new ArgumentNullException(nameof(repos));
+            }
+
+            TotalRepos = totalRepos;
+            Repos = repos
+                .OrderByDescending(x => x.Stars)
+                .ThenBy(x => x.Id)
```
joelverhagen: nit: StringComparer.OrdinalIgnoreCase

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Linq;
+
+namespace NuGetGallery.GitHub
+{
+    public class GitHubUsageConfiguration : IGitHubUsageConfiguration
+    {
+        private readonly Dictionary<string, NuGetPackageGitHubInformation> _nuGetPackagesGitHubDependencies;
```
joelverhagen: >Dictionary [](start = 25, length = 10)

nit: IReadOnlyDictionary

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                return value;
+            }
+
+            return NuGetPackageGitHubInformation.Empty;
+        }
+
+        private static Dictionary<string, NuGetPackageGitHubInformation> GetNuGetPackagesDependents(IReadOnlyList<RepositoryInformation> repositories)
+        {
+            var dependentsPerPackage = new Dictionary<string, List<RepositoryInformation>>(StringComparer.InvariantCultureIgnoreCase);
```
joelverhagen: >InvariantCultureIgnoreCase [](start = 106, length = 26)

OrdinalIgnoreCase

src/NuGetGallery.Core/GitHub/NuGetPackageGitHubInformation.cs
```diff
@@ -0,0 +1,38 @@
...
+                throw new IndexOutOfRangeException(string.Format("{0} cannot have a negative value!", nameof(totalRepos)));
+            }
+
+            if( repos == null)
+            {
+                throw new ArgumentNullException(nameof(repos));
+            }
+
+            TotalRepos = totalRepos;
+            Repos = repos
```
joelverhagen: Nit: maybe do this sorting and list truncation in `GitHubUsageConfiguration` so that all the "business logic" is in one and and this model is just dumb

tests/NuGetGallery.Facts/Extensions/NumberExtensionsFacts.cs
```diff
@@ -29,5 +29,38 @@ public void FormatsUsingExpectedUnit(long bytes, string expected)
...
+            [InlineData(1_999_000, "1.99M")]
+            [InlineData(99_990_000, "99.9M")]
+            [InlineData(100_990_000, "100M")]
+            [InlineData(999_990_000, "999M")]
+            [InlineData(1_000_000_000, "1.00G")]
+            [InlineData(1_999_000_000, "1.99G")]
+            public void FormatsUsingExpectedUnit(int number, string expected)
```
joelverhagen: I would expect K. M seems relatively common. G seems weird though. i have never seen that for abbreviating quantities. Maybe it's just thinking of financial journals but  I would expect "B" for billion.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -517,6 +517,49 @@
...
+                                            onclick="window.open('@dep.Url');">
```
joelverhagen: Why not just a link `<a>`?

Project URL, license URL, description URLs etc just do a standard navigation

Also, I would recommend rel=nofollow

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -688,11 +700,17 @@ public async Task GivenAnAbsoluteLatestVersionItReturnsTheFirstLatestSemVer2()
                 var packageService = new Mock<IPackageService>();
                 var indexingService = new Mock<IIndexingService>();
                 var deprecationService = new Mock<IPackageDeprecationService>();
+                var contentObjectService = new Mock<IContentObjectService>();
+                contentObjectService
```
joelverhagen: perhaps CreateController should do this by default or the app code should be resilient to null

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -823,13 +847,20 @@ public async Task GivenAValidPackageWithNoVersionThatTheCurrentUserDoesNotOwnItD
...
                     GetConfigurationService(),
                     packageService: packageService,
                     indexingService: indexingService,
-                    deprecationService: deprecationService);
+                    deprecationService: deprecationService,
+                    contentObjectService: contentObjectService);
                 controller.SetCurrentUser(TestUtility.FakeUser);
 
+
                 var package = new Package()
```
joelverhagen: no extra space

joelverhagen:
>             private async Task<ActionResult> GetActionResultForPackageStatusAsync(

Maybe I missed it but where do you actually assert the value from the content object service is the same as the value put into the view model

---
Refers to: tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs:496 in 229d173. [](commit_id = 229d17330ec98397f82642bcd29744ceadacad8c, deletion_comment = False)

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -517,6 +517,49 @@
...
+                        <span>GitHub Usage</span>
+                    </a>
+                </h2>
+            
+                if (Model.GitHubDependenciesInformation.Repos.Any())
+                {
+                    <div class="github-usage panel-collapse collapse" aria-expanded="true" id="github-usage">
+                        <p>
+                            About @(Model.GitHubDependenciesInformation.TotalRepos.ToKiloFormat()) popular GitHub repositories depend on
+                            @(Model.Id), including the following:
```
joelverhagen: @karann-msft / @anangaur could you check the phrasing here?

src/NuGetGallery/Extensions/NumberExtensions.cs
```diff
@@ -66,5 +67,40 @@ public static string ToUserFriendlyBytesLabel(this long bytes)
...
+        /// <returns></returns>
+        public static string ToKiloFormat(this int number)
+        {
+            if (number >= 1_000_000_000)
+            {
+                return new StringBuilder((number / 1_000_000_000.0f).ToString("F3")) { [4] = 'G' }.ToString();
```
joelverhagen: Don't think the string builder adds anything useful here. String interpolation will probably be more readly like:

$"{(number / 1_000_000_000.0f):F3}G"

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
agr: Is code in Gallery GitHub-specific? Should we keep 'GitHub' out of class names everywhere?

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery.GitHub
+{
+    public class GitHubUsageConfiguration : IGitHubUsageConfiguration
```
agr: Is code in Gallery GitHub-specific? Should we keep 'GitHub' out of class names everywhere?

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            if (_nuGetPackagesGitHubDependencies.TryGetValue(packageId, out var value))
+            {
+                return value;
+            }
+
+            return NuGetPackageGitHubInformation.Empty;
+        }
+
+        private static Dictionary<string, NuGetPackageGitHubInformation> GetNuGetPackagesDependents(IReadOnlyList<RepositoryInformation> repositories)
```
agr: If I understand code correctly this whole function can be rewritten as a single LINQ query. Something along those lines (I didn't actually check that it compiles and works as intended):
```C#
return repositories
    .SelectMany(r => r.Dependencies.Select(d => new {Dependency = d, RepositoryInformation: r}))
    .GroupBy(x => x.Dependency, x => x.RepositoryInformation)
    .ToDictionary(g => g.Key, g => new NuGetPackageGitHubInformation(g.Count, g));
```
(code needs sprinkling with `StringComparer.OrdinalIgnoreCase`).

src/NuGetGallery.Core/GitHub/NuGetPackageGitHubInformation.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public class NuGetPackageGitHubInformation
+    {
+        public const int ReposPerPackage = 10;
+
+        public readonly static NuGetPackageGitHubInformation Empty = new NuGetPackageGitHubInformation(
+                                                            0,
+                                                            new List<RepositoryInformation>());
+
+        public NuGetPackageGitHubInformation(int totalRepos, IReadOnlyList<RepositoryInformation> repos)
```
agr: `totalRepos` duplicates `repos.Count` here isn't it?

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery.GitHub
+{
+    public class GitHubUsageConfiguration : IGitHubUsageConfiguration
+    {
+        private readonly Dictionary<string, NuGetPackageGitHubInformation> _nuGetPackagesGitHubDependencies;
+
+        public GitHubUsageConfiguration(IReadOnlyList<RepositoryInformation> repositories)
```
agr: Please prefer `IReadOnlyCollection` over `IReadOnlyList` everywhere where you don't need access by index.

src/NuGetGallery.Core/GitHub/NuGetPackageGitHubInformation.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            TotalRepos = totalRepos;
+            Repos = repos
+                .OrderByDescending(x => x.Stars)
+                .ThenBy(x => x.Id)
+                .Take(ReposPerPackage)
+                .ToList();
+        }
+
+        public int TotalRepos { get; }
```
agr: Same comment as for the argument in constructor. It seems it is not needed or should be implemented as ` => Repos.Count`.

src/NuGetGallery.Services/Storage/ContentObjectService.cs
```diff
@@ -46,9 +49,14 @@ public async Task Refresh()
             TyposquattingConfiguration =
                await Refresh<TyposquattingConfiguration>(ServicesConstants.ContentNames.TyposquattingConfiguration) ??
                new TyposquattingConfiguration();
+
+            var reposCache = await Refresh<IReadOnlyList<RepositoryInformation>>(ServicesConstants.ContentNames.NuGetPackagesGitHubDependencies) ??
+                             Array.Empty<RepositoryInformation>();
```
agr: Layout/indentation is inconsistent with previous statement.

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
loic-sharma: Don't we use `ToLowerInvariant` for package IDs for things like the flat container? How do you choose between `OrdinalIgnoreCase` vs `InvariantCultureIgnoreCase`?

src/NuGetGallery.Core/GitHub/NuGetPackageGitHubInformation.cs
mogah: Loïc and I felt like the `ReposPerPackage` and the filtering would be more appropriate in the class since it's all information that is tightly related to it

src/NuGetGallery.Core/GitHub/NuGetPackageGitHubInformation.cs
mogah: No this is not, since the Repos collection is a trimmed version of the collection received in the constructor

src/NuGetGallery.Core/GitHub/RepositoryInformation.cs
mogah: Apparently not 😮

src/NuGetGallery/Extensions/NumberExtensions.cs
```diff
@@ -66,5 +67,40 @@ public static string ToUserFriendlyBytesLabel(this long bytes)
...
         }
+
+        /// <summary>
+        /// Format the number to a 3 digit representation plus a letter to represent the scale (K for kilo, M for mega, or G for giga)
+        /// </summary>
+        /// <param name="number"></param>
+        /// <returns></returns>
+        public static string ToKiloFormat(this int number)
+        {
+            if (number >= 1_000_000_000)
```
agr: The list of values you compare against can be explicitly declared, might increase the readability:
```C#
var data  = new[]{  // this can be static
    new { Threshold = 1_000_000_000, Transform = n => n / 1_000_000_000f, Format = "{0:G3}G" }, // should it be "B" for billions? What's "G"?
    new { Threshold = 1_000_000, Transform = n => n / 1_000_000f, Format = "{0:G3}M" },
    new { Threshold = 1_000, Transform = n => n / 1_000f, Format = "{0:G3}K" },
    new { Threshold = int.MinValue, Transform = n => n, Format = "{0}" },
};

return data
    .First(d => number >= d.Threshold)
    .Select(d => string.Format(d.Format, d.Transform(number)));
```

src/NuGetGallery/Extensions/NumberExtensions.cs
```diff
@@ -66,5 +67,40 @@ public static string ToUserFriendlyBytesLabel(this long bytes)
...
+            {
+                return new StringBuilder((number / 1_000_000_000.0f).ToString("F3")) { [4] = 'G' }.ToString();
+            }
+
+            if (number >= 100_000_000)
+            {
+                return (number / 1_000_000) + "M";
```
agr: Integer division uses `floor` to round while float division with subsequent formatting with `F2` uses regular rounding.
I.e. 123_567_890 will produce `123M` while 12_356_789 will produce `12.4M`. That looks inconsistent.

src/NuGetGallery/Extensions/NumberExtensions.cs
```diff
@@ -66,5 +67,40 @@ public static string ToUserFriendlyBytesLabel(this long bytes)
...
+            {
+                return (number / 1_000) + "K";
+            }
+
+            if (number >= 1000)
+            {
+                return new StringBuilder((number / 1_000.0f).ToString(number >= 10_000 ? "F2" : "F3")) { [4] = 'K' }.ToString();
+            }
+
+            return number.ToString("#,0");
```
agr: This will produce number with thousand separators, all other options don't use them.

tests/NuGetGallery.Facts/Extensions/NumberExtensionsFacts.cs
agr: Missing tests for negative numbers.

agr: I might be nitpicking here, but I have concerns regarding the readability of the output: when you presents a column of numbers with decimal points, you'd want those decimal points aligned:
```
23.4M
1.23M
```
There is an order of magnitude difference between numbers which is not immediately obvious.

agr: Probably need tests to define/validate rounding rules.

tests/NuGetGallery.Core.Facts/GitHub/GitHubUsageConfigurationFacts.cs
```diff
@@ -0,0 +1,170 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace NuGetGallery
+{
+    public class GitHubUsageConfigurationFacts
+    {
+        private static RepositoryInformation GenRepo(string id, int stars = 100, params string[] nugetDependencies)
```
agr: Does it compile in our build pipelines? We don't have the latest compiler there, default argument in the middle of argument list might not be supported.

src/NuGetGallery/Extensions/NumberExtensions.cs
mogah: This is technically an F2, but I use F3 since there are weird rounding problems and then override the last character with the Unit scale

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
mogah: What do you mean?

src/NuGetGallery.Core/GitHub/NuGetPackageGitHubInformation.cs
loic-sharma: Yup, especially since consumers of this class expect `Repos` to be ordered.

tests/NuGetGallery.Facts/Extensions/NumberExtensionsFacts.cs
mogah: Yeah it's nice. I used to do it in Java, tried it in C#, it worked 😄

src/NuGetGallery/Extensions/NumberExtensions.cs
mogah: 12_356_789 produces 12.3M

agr: Missed the fact that last digit was overwritten with K/M/G

src/NuGetGallery/Extensions/NumberExtensions.cs
```diff
@@ -75,32 +76,22 @@ public static string ToUserFriendlyBytesLabel(this long bytes)
...
-            }
-
-            if (number >= 100_000)
-            {
-                return (number / 1_000) + "K";
-            }
+                new { Threshold = 1_000_000_000, Transform = new Func<int, float>((x) => x / 1_000_000_000f), Format = "{0:F3}B", Trim = true },
+                new { Threshold = 100_000_000, Transform =  new Func<int, float>((x) => x / 1_000_000), Format = "{0:F0}M", Trim = false},
+                new { Threshold = 10_000_000, Transform =  new Func<int, float>((x) => x / 1_000_000f), Format = "{0:F2}M", Trim = true},
+                new { Threshold = 1_000_000, Transform =  new Func<int, float>((x) => x / 1_000_000f), Format = "{0:F3}M", Trim = true},
```
agr: Why not `G3` format? This will leave 3 significant digits and will cover 1M/10M/100M cases in a single line.

src/NuGetGallery/Extensions/NumberExtensions.cs
mogah: I was just following the file style. The param is the int the method is applied to since it is an extension

src/NuGetGallery.Services/ServicesConstants.cs
```diff
@@ -39,6 +39,7 @@ public static class ContentNames
             public static readonly string CertificatesConfiguration = "Certificates-Configuration";
             public static readonly string SymbolsConfiguration = "Symbols-Configuration";
             public static readonly string TyposquattingConfiguration = "Typosquatting-Configuration";
+            public static readonly string NuGetPackagesGitHubDependencies = "GitHubUsage.v1";
```
shishirx34: why `.v1`, what other iterations are planned? why not name it `GitHubUsage-Configuration` same as all other configs?

src/NuGetGallery.Services/Storage/ContentObjectService.cs
```diff
@@ -46,9 +48,15 @@ public async Task Refresh()
             TyposquattingConfiguration =
                await Refresh<TyposquattingConfiguration>(ServicesConstants.ContentNames.TyposquattingConfiguration) ??
                new TyposquattingConfiguration();
+
+            var reposCache = 
+                await Refresh<IReadOnlyCollection<RepositoryInformation>>(ServicesConstants.ContentNames.NuGetPackagesGitHubDependencies) ??
```
shishirx34: I guess the background job will be writing the repos cache to this location?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -517,6 +517,50 @@
...
 
+            @if(Model.IsGitHubUsageEnabled)
+            {
+                <h2>
+                    <a role="button" data-toggle="collapse" data-target="#github-usage"
+                       aria-expanded="false" aria-controls="github-usage" id="show-github-usage">
+                        <i class="ms-Icon ms-Icon--ChevronRight" aria-hidden="true"></i>
+                        <span>GitHub Usage</span>
+                    </a>
+                </h2>
```
shishirx34: I am assuming this section will also be collapsed by default, what do you think of adding an overview details to the header to the right...

Something like

```
> GitHub Usage                                                      1377 Repositories
```
The overview details can be augmented with other details if needed.

It would also be a good idea to add this to other sections(#dependencies of the package for Dependencies section, #versions for Version History section.

@karann-msft, @anangaur - thoughts?

src/NuGetGallery/Extensions/NumberExtensions.cs
joelverhagen: We have been inconsistent/low quality in the past. This is a place where being inconsistent is OK IMO.

src/NuGetGallery.Core/GitHub/NuGetPackageGitHubInformation.cs
joelverhagen: Sounds good.

src/NuGetGallery.Services/ServicesConstants.cs
mogah: @loic-sharma Thought that it might be better to name it this way (same format as downloads.v1.json file) since we might change the format in the futur

src/NuGetGallery.Services/Storage/ContentObjectService.cs
mogah: yes

src/NuGetGallery.Core/GitHub/GitHubUsageConfiguration.cs
mogah: The whole project is GitHub specific. Indexing other Git repos hosting providers was a non-goal

tests/NuGetGallery.Facts/Extensions/NumberExtensionsFacts.cs
mogah: Yeah K is for Kilo, M for Mega, so G is for Giga, but you're the 3rd person to notice this, so I guess B is more used than G ^^

src/NuGetGallery.Services/ServicesConstants.cs
loic-sharma: This file is closer to `downloads.v1.json` and `owners.v1.json` as it is generated periodically by a background job.

We didn't add a `-Configuration` suffix as this file contains data exported from GitHub (it's not configs).

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
loic-sharma: This is an interesting idea. I think I'd prefer the information to be next to the title:

```
> GitHub Usage (1377 Repositories)
```

We'll ask the UX expert about this suggestion, thanks!

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -517,6 +517,50 @@
...
+                                </div>
+                            </div>
+
+                        }
+                    </div>
+                }
+                else
+                {
+                    <p class="github-usage panel-collapse collapse" aria-expanded="true" id="github-usage">
+                        This package is not used by any popular GitHub repository.
```
loic-sharma:
```suggestion
                        This package is not used by any popular GitHub repositories.
```

tests/NuGetGallery.Facts/Extensions/NumberExtensionsFacts.cs
loic-sharma: @joelverhagen Why are you reading financial journals with billions? 😆

PR #7263, opened 2019-06-13
[GH Usage] Added analytics

mogah:
This PR adds analytics for the GitHub Usage section. 
The 2 metrics are:
- Clicks on the GitHub Usage section (show/hide)
- Clicks on the GitHub repo links (which index was clicked/how many total clicks)

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -105,14 +105,12 @@ $(function () {
         });
     }
 
-    // Emit a Google Analytics event when the user expands or collapses the Dependencies section.
     if (window.nuget.isGaAvailable()) {
+        // Emit a Google Analytics event when the user expands or collapses the Dependencies section.
         $("#dependency-groups").on('hide.bs.collapse show.bs.collapse', function (e) {
             ga('send', 'event', 'dependencies', e.type);
         });
-    }
 
-    if (window.nuget.isGaAvailable()) {
         // Emit a Google Analytics event when the user expands or collapses the GitHub Usage section.
         $("#github-usage").on('hide.bs.collapse show.bs.collapse', function (e) {
             ga('send', 'event', 'github-usage', e.type);
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -111,4 +111,24 @@ $(function () {
             ga('send', 'event', 'dependencies', e.type);
         });
     }
+
+    if (window.nuget.isGaAvailable()) {
+        // Emit a Google Analytics event when the user expands or collapses the GitHub Usage section.
+        $("#github-usage").on('hide.bs.collapse show.bs.collapse', function (e) {
+            ga('send', 'event', 'github-usage', e.type);
+        });
+
+        // Emit a Google Analytics event when the user clicks on a repo link in the GitHub Usage section.
+        $(".btn-gh-repo").on('click', function (e) {
+            let elem = e.delegateTarget.parentElement.parentElement;
+            let linkIndex = 0;
+            while ((elem = elem.previousSibling) != null) {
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -111,4 +111,24 @@ $(function () {
...
+    if (window.nuget.isGaAvailable()) {
+        // Emit a Google Analytics event when the user expands or collapses the GitHub Usage section.
+        $("#github-usage").on('hide.bs.collapse show.bs.collapse', function (e) {
+            ga('send', 'event', 'github-usage', e.type);
+        });
+
+        // Emit a Google Analytics event when the user clicks on a repo link in the GitHub Usage section.
+        $(".btn-gh-repo").on('click', function (e) {
+            let elem = e.delegateTarget.parentElement.parentElement;
+            let linkIndex = 0;
```
loic-sharma: Let's try to use `data-track` attributes. Example:

https://github.com/NuGet/NuGetGallery/blob/5b7c505086fae24912f3b0d3df58ffb5a9d1e8b8/src/NuGetGallery/Views/Shared/_ListPackage.cshtml#L43

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -111,4 +111,21 @@ $(function () {
...
+    if (window.nuget.isGaAvailable()) {
+        // Emit a Google Analytics event when the user expands or collapses the GitHub Usage section.
+        $("#github-usage").on('hide.bs.collapse show.bs.collapse', function (e) {
+            ga('send', 'event', 'github-usage', e.type);
+        });
+
+        // Emit a Google Analytics event when the user clicks on a repo link in the GitHub Usage section.
+        $(".btn-gh-repo").on('click', function (elem) {
+            if (!elem.delegateTarget.dataset.indexNumber) {
+                console.error("indexNumber property doesn't exist!");
```
joelverhagen: Why console.error? Will this show up somewhere for us to see like in AI?

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -111,4 +111,21 @@ $(function () {
...
+            } else {
+                let linkIndex = elem.delegateTarget.dataset.indexNumber;
+                ga('send', 'event', 'github-usage', 'link-click-' + linkIndex);
```
joelverhagen: Why isn't the link index a value?

joelverhagen: (don't feel strongly one way or another, just curious how this will be used)

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -111,4 +111,21 @@ $(function () {
             ga('send', 'event', 'dependencies', e.type);
         });
     }
+
+    if (window.nuget.isGaAvailable()) {
```
joelverhagen: nit you can group this together with the dependency-groups if block

src/NuGetGallery/Scripts/gallery/page-display-package.js
mogah: I didn't find a way to easily get the different values count and also get the count for each value individually. By making them actions, that becomes way easier

src/NuGetGallery/Scripts/gallery/page-display-package.js
mogah: I made this since it's the easiest way to notice that there is a problem. Imagine someone modifies the HTML code and deletes the data-index-number attribute, this will show an error on the console to notify them. I feel that if it gets logged in AI, this will get burried in the exceptions.

src/NuGetGallery/Scripts/gallery/page-display-package.js
joelverhagen: Sounds good. This is okay since there is a fixed cardinality.

PR #7268, opened 2019-06-17
[DB Migration] Fix the package dependency issue and some updates

zhhyu:
Mainly contain the following updates:
1. Fix the package dependency issue due to the conflict among NuGet.Jobs/Server Common/Gallery and recent package version upgrade.  
2. Depend on the new job to return ExitCode to make it clean;
3. Improve test coverage and telemetry messages.

No code logic is changed.


tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
```diff
@@ -0,0 +1,131 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+using Xunit;
+
+namespace NuGet.Services.DatabaseMigration.Facts
```
joelverhagen: Let's talk through the goal of these tests. This seems like a maintenance head-ache.

src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
```diff
@@ -54,6 +54,9 @@
     <None Include="App.config" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Autofac.Extensions.DependencyInjection">
```
joelverhagen: Why is this a direct dependency?

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
```diff
@@ -0,0 +1,131 @@
...
+                yield return new object[] { "Microsoft.Extensions.Options", "1.1.2" };
+            }
+        }
+
+        /// <summary>
+        /// 1. Verify that the version of EntityFramework equals to 6.2.0;
+        /// We hit a bug of EntityFramework 6.2.0, and We use some special ways to fix it which only targets version 6.2.0.
+        /// <see cref="NuGet.Services.DatabaseMigration.Job"/>: "NuGet.Services.DatabaseMigration.Job/OverwriteSqlConnection"
+        /// 2. Verify that the version of Microsoft.Extensions.* package equals to 1.1.2;
+        /// 3. Verify that the version of Autofac package equals to 4.2.0.
```
agr: Why do we want to enforce Autofac version?

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
```diff
@@ -0,0 +1,131 @@
...
+                yield return new object[] { "Microsoft.Extensions.Logging", "1.1.2" };
+                yield return new object[] { "Microsoft.Extensions.Options", "1.1.2" };
+            }
+        }
+
+        /// <summary>
+        /// 1. Verify that the version of EntityFramework equals to 6.2.0;
+        /// We hit a bug of EntityFramework 6.2.0, and We use some special ways to fix it which only targets version 6.2.0.
+        /// <see cref="NuGet.Services.DatabaseMigration.Job"/>: "NuGet.Services.DatabaseMigration.Job/OverwriteSqlConnection"
+        /// 2. Verify that the version of Microsoft.Extensions.* package equals to 1.1.2;
```
agr: An explanation about why we want 1.x specifically (like the one above, for the `EntityFramework` package) would be nice to have here.

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// 3. Verify that the version of Autofac package equals to 4.2.0.
+        /// </summary>
+        [Theory]
+        [MemberData(nameof(NuGet_Services_DatabaseMigration_PackageNameAndVersion))]
+        public void VerifyPackageVersionForNuGet_Services_DatabaseMigration(string packageName, string packageVersion)
+        {
+            var csproj = new FileInfo(Path.Combine(_sln.Directory.FullName, @"src\NuGet.Services.DatabaseMigration\NuGet.Services.DatabaseMigration.csproj"));
+            var csprojPackageReferences = XDocument.Load(csproj.FullName).Root.Descendants().Where(e => e.Name.LocalName == "PackageReference").ToList();
+            var packageElement = csprojPackageReferences.Where(d => d.Attribute("Include").Value.Equals(packageName, StringComparison.OrdinalIgnoreCase)).Single();
```
agr: `queryable.Where(predicate).Single()` can generally be expressed as `queryable.Single(delegate)`, no need for extra calls.

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
```diff
@@ -0,0 +1,131 @@
...
+                yield return new object[] { "Autofac.Extensions.DependencyInjection", "4.2.0" };
+                yield return new object[] { "EntityFramework", "6.2.0" };
+                yield return new object[] { "Microsoft.Extensions.Configuration", "1.1.2" };
+                yield return new object[] { "Microsoft.Extensions.Configuration.Abstractions", "1.1.2" };
+                yield return new object[] { "Microsoft.Extensions.Configuration.Binder", "1.1.2" };
+                yield return new object[] { "Microsoft.Extensions.Configuration.EnvironmentVariables", "1.1.2" };
+                yield return new object[] { "Microsoft.Extensions.Configuration.FileExtensions", "1.1.2" };
+                yield return new object[] { "Microsoft.Extensions.Configuration.Json", "1.1.2" };
+                yield return new object[] { "Microsoft.Extensions.Logging", "1.1.2" };
+                yield return new object[] { "Microsoft.Extensions.Options", "1.1.2" };
```
agr: Do we need all those `Microsoft.Extensions.Configuration.*` here? Would listing `Microsoft.Extensions.Logging` and `Microsoft.Extensions.Options` only be enough?

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
joelverhagen: Spoke offline.

Let's try to verify the runtime behavior without executing real migrations. A start could be override `Job.Run`.

src/DatabaseMigrationTools/App.config
```diff
@@ -1,6 +1,14 @@
...
-    <startup> 
-        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.2" />
-    </startup>
+  <startup>
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.2" />
+  </startup>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="Microsoft.Extensions.Logging" publicKeyToken="adb9793829ddae60" culture="neutral" />
```
joelverhagen: Does `AutoGenerateBindingRedirects` fix this?

joelverhagen: It's worth a try at least :)

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
zhhyu: The Gallery depends on Autofac 4.3.1, and jobs.common depends on 4.2.0. And it seems that they change the API. 
So if we depend on Autofac 4.3.1, we will meet the "System.MissingMethodException: Method not found" error.

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
zhhyu: Runtime verification is hard to implement, because it depends on the NuGet.Jobs to read the configuration and inject the secrets. 
DLL version check is easily to be broken, so remove it now. 
Decide to verify that the package version is set up in csproj. The break last time is because of the manual deletion of these package dependencies.

src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
zhhyu: Same as below.

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
zhhyu: Add it now.

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
zhhyu: Done.

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
zhhyu: Yes, you are right. Remove these dependencies now.

src/DatabaseMigrationTools/App.config
zhhyu: No, it doesn't work. I am not sure why.

tests/NuGet.Services.DatabaseMigration.Facts/PackageDependencyFacts.cs
zhhyu: If we want to solve this for a long-time goal, there is a need to align the package version among NuGet.Jobs/Server Common/Gallery.

zhhyu:
Remove the version check now. 
Create two issues for this:
https://github.com/nuget/engineering/issues/2528
https://github.com/nuget/engineering/issues/2529

tests/NuGet.Services.DatabaseMigration.Facts/DatabaseMigrationFacts.cs
```diff
@@ -1,55 +1,28 @@
...
-        private static FileInfo TryFindFile(string startDirectory, string filename)
+        public void AssertDbMigratorPrivateFieldNotNull()
         {
-            var path = new DirectoryInfo(startDirectory);
+            var historyRepository = typeof(DbMigrator).
+                GetField("_historyRepository",
+                BindingFlags.NonPublic | BindingFlags.Instance);
 
-            FileInfo[] files = null;
+            Assert.NotNull(historyRepository);
```
agr: Do you want to check the type besides the presence of the property?

zhhyu: Check the type now.

Issue #7236, opened 2019-06-04
Improve github issues template for Gallery and Engineering repos


skofman1:
Purpose: make it more usable for customers and team members (state problems, customer impact, possible solutions, complexity of solution. )

PR #7270, opened 2019-06-17
Update GitHub templates

loic-sharma:
These templates are loosely based off AspNetCore's: https://github.com/aspnet/AspNetCore/tree/master/.github

Part of https://github.com/nuget/nugetgallery/issues/7236


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,24 @@
+---
+name: Bug report
+about: Report issues when hosting the NuGet Gallery
+---
+
+### For issues regarding the NuGet client tools (the Visual Studio extension, NuGet.exe command line tool, etc.), file a bug on [NuGet Home](https://github.com/nuget/home/issues).
+
+### Describe the bug
```
agr: Are users expected to keep the structure of document headings? If yes, they should be formulated as proper headings: "Bug description", "Reproduction steps", etc.

loic-sharma: > Are users expected to keep the structure of document headings? 

Users will likely keep the headings as we do not have any strict requirements here.

> If yes, they should be formulated as proper headings: "Bug description", "Reproduction steps", etc.

We use GitHub issues for two buckets:

1. Customer reports of bugs and issues
2. Our own work tracking

Your suggestion gears the templates closer to that second bucket. I'd rather the templates are customer oriented to guide users to create great issues. We, the maintainers, can modify the issue after the fact to store whatever information we feel is necessary to track the work.

Issue #7092, opened 2019-04-19
[Icons] Gallery should serve icons from flat container when available


agr:
Related to #7066

Issue #7064, opened 2019-04-11
[Icons] DB changes


agr:
Add a flag to the packages container indicating the presence of the icon file in flat container.

PR #7271, opened 2019-06-18
Initial Gallery support for icon files

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/7064 and https://github.com/NuGet/NuGetGallery/issues/7092.

Feature is not publicly available (behind feature flag), no icon file validation present and no upload UI: both will come later. No non-hijacked V2 support.

Added new `Package` flag `UsesIconFromFlatContainer` indicating the presence of embedded icon file in the package.
Added two new configuration properties:
* `IgnoreIconUrl` makes Gallery ignore the `iconUrl` metadata for UI purposes (URL is still saved in DB);
* `InternalIconUrlBaseAddress` specifies the base URL for internal icon storage;

Added the `IIconUrlProvider` to abstract Icon URL generation. Two methods there are for convenience, to reduce boilerplate null checks when we need `Uri` objects instead of strings.

We have fairly deep view-model inheritance trees in certain places and have an unfortunate tendency to bring business logic into them. I am not going to address that as a part of the work, since that would require major updates. Instead I pass `IIconUrlProvider` around. I added an engineering work item linking to this issue to track the view-model fix.


src/NuGetGallery/Infrastructure/PackageIndexEntity.cs
```diff
@@ -15,11 +15,12 @@ public class PackageIndexEntity
 
         public Package Package { get; set; }
 
-        public PackageIndexEntity() { }
```
agr: I am a little bit confused about the need for the default constructor. It seems to have been used in a single place and that usage there was not needed. Am I missing something?

tests/NuGetGallery.Facts/Controllers/ODataV2CuratedFeedControllerFacts.cs
```diff
@@ -442,7 +444,8 @@ public async Task SearchCount_IncludesSemVerV2PackageVersionsWhenSemVerLevel2OrH
             IEntityRepository<Package> packagesRepository,
             IGalleryConfigurationService configurationService,
             ISearchService searchService,
-            ITelemetryService telemetryService)
+            ITelemetryService telemetryService,
+            IIconUrlProvider iconUrlProvider)
```
agr: not needed

tests/NuGetGallery.Facts/Controllers/ODataV1FeedControllerFacts.cs
```diff
@@ -93,7 +93,8 @@ public async Task SearchCount_FiltersSemVerV2PackageVersions()
             IEntityRepository<Package> packagesRepository,
             IGalleryConfigurationService configurationService,
             ISearchService searchService,
-            ITelemetryService telemetryService)
+            ITelemetryService telemetryService,
+            IIconUrlProvider iconUrlProvider)
```
agr: Unused

tests/NuGetGallery.Facts/Controllers/ODataV2FeedControllerFacts.cs
```diff
@@ -406,7 +406,8 @@ public async Task GetUpdatesCount_IncludesSemVerV2PackageVersionsWhenSemVerLevel
             IEntityRepository<Package> packagesRepository,
             IGalleryConfigurationService configurationService,
             ISearchService searchService,
-            ITelemetryService telemetryService)
+            ITelemetryService telemetryService,
+            IIconUrlProvider iconUrlProvider)
```
agr: Unused

src/NuGetGallery.Core/Packaging/PackageMetadata.cs
```diff
@@ -100,6 +100,7 @@ private void SetPropertiesFromMetadata()
             Title = GetValue(PackageMetadataStrings.Title, (string)null);
             Tags = GetValue(PackageMetadataStrings.Tags, (string)null);
             Language = GetValue(PackageMetadataStrings.Language, (string)null);
+            Icon = GetValue(PackageMetadataStrings.Icon, (string)null);
```
joelverhagen: Is there an icon "type" that we need to observe? Db2Catalog i believe expects no type or type "file"

src/NuGetGallery.Core/IAppConfiguration.cs
```diff
@@ -430,5 +430,15 @@ public interface IAppConfiguration : IMessageServiceConfiguration
...
+
+        /// <summary>
+        /// Indicates whether the contents of the IconUrl package metadata should be ignored.
+        /// </summary>
+        bool IgnoreIconUrl { get; set; }
+
+        /// <summary>
+        /// Base URL for the internal storage if package icons.
+        /// </summary>
+        string InternalIconUrlBaseAddress { get; set; }
```
joelverhagen: I don't understand what "Internal" means here. Can we just say `IconStorageBaseUrl` or something like that?

src/NuGetGallery/Web.config
```diff
@@ -156,6 +156,8 @@
     <add key="Gallery.RejectSignedPackagesWithNoRegisteredCertificate" value="false"/>
     <add key="Gallery.RejectPackagesWithTooManyPackageEntries" value="false"/>
     <add key="Gallery.BlockSearchEngineIndexing" value="false"/>
+    <add key="Gallery.IgnoreIconUrl" value ="false"/>
+    <add key="Gallery.InternalIconUrlBaseAddress" value="https://apidev.nugettest.org/v3/flatcontainer/" />
```
joelverhagen: nit no trailing space

src/NuGetGallery.Core/Packaging/PackageMetadata.cs
```diff
@@ -134,6 +135,12 @@ private void SetPropertiesFromMetadata()
         /// </summary>
         public LicenseMetadata LicenseMetadata { get; }
 
+        /// <summary>
+        /// Contains the embedded icon filename taken from the 'icon' node of the nuspec file.
+        /// Null if not specified.
+        /// </summary>
+        public string Icon { get; private set; }
```
joelverhagen: nit: iconFile to be more consistent across our system

src/NuGet.Services.Entities/Package.cs
```diff
@@ -274,5 +274,11 @@ public bool HasReadMe
         /// See <see cref="PackageDeprecation.AlternatePackage"/>.
         /// </summary>
         public virtual ICollection<PackageDeprecation> AlternativeOf { get; set; }
+
+        /// <summary>
+        /// A flag that indicates that the icon in flat container is assumed to exist and must be
+        /// used everywhere instead of the <see cref="IconUrl"/>.
+        /// </summary>
+        public bool UsesIconFromFlatContainer { get; set; }
```
joelverhagen: This name doesn't make much sense for stand-alone deployments. Can we make this "UsesIconFromStorage" or something like that?

src/NuGetGallery.Core/IAppConfiguration.cs
joelverhagen: How does this work on stand-alone deployments?

joelverhagen: What do you think about making this a URL pattern like @xavierdecoster did with package URL pattern? Then we could default the pattern to the gallery endpoint that enables this on stand-alone endpoints or something like that

joelverhagen: https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Catalog/Helpers/PackageContentUriBuilder.cs

agr: The plan is to have Gallery endpoint that would serve an icon for the package. Then if this option is not set to anything `IIconUrlProvider` will point to that Gallery endpoint.

agr: There is no code currently to support standalone deployments, but the work is planned.

src/NuGetGallery.Core/Packaging/PackageMetadata.cs
agr: The type was dropped from the [current spec](https://github.com/NuGet/Engineering/pull/1925). And I did not expect to have other types in the first place.

src/NuGetGallery.Services/Configuration/IAppConfiguration.cs
```diff
@@ -430,5 +430,15 @@ public interface IAppConfiguration : IMessageServiceConfiguration
...
         /// </summary>
         int SearchHttpClientTimeoutInMilliseconds { get; set; }
+
+        /// <summary>
+        /// Indicates whether the contents of the IconUrl package metadata should be ignored.
+        /// </summary>
+        bool IgnoreIconUrl { get; set; }
+
+        /// <summary>
+        /// Base URL for the internal storage if package icons.
```
loic-sharma:
```suggestion
        /// Base URL for the internal storage of package icons.
```

src/NuGetGallery/Infrastructure/PackageIndexEntity.cs
```diff
@@ -15,11 +15,12 @@ public class PackageIndexEntity
...
+        private readonly IIconUrlProvider _iconUrlProvider;
```
loic-sharma: Please place this field before the `Package` property.

src/NuGetGallery.Core/IAppConfiguration.cs
agr: "Internal" means "related to the service operators". As opposing to any "external" storage operated by third parties.

joelverhagen: I don't think that's very clear. The URL is externally accessible. IMO when I hear "external" vs "internal" URL I think access levels or network not who is maintaining it. But I see your point now.

src/NuGet.Services.Entities/Package.cs
skofman1: Since it's part of the DB schema, wouldn't "HasSelfContainedIcon" be a more descriptive name?

---
In reply to: [296406860](https://github.com/NuGet/NuGetGallery/pull/7271#discussion_r296406860) [](ancestors = 296406860)

src/NuGetGallery/Web.config
```diff
@@ -156,6 +156,8 @@
     <add key="Gallery.RejectSignedPackagesWithNoRegisteredCertificate" value="false"/>
     <add key="Gallery.RejectPackagesWithTooManyPackageEntries" value="false"/>
     <add key="Gallery.BlockSearchEngineIndexing" value="false"/>
+    <add key="Gallery.IgnoreIconUrl" value ="false"/>
+    <add key="Gallery.InternalIconUrlBaseAddress" value=""/>
```
skofman1: can you provide an example value for this config? I wonder if it should be hard coded.

src/NuGetGallery/Web.config
```diff
@@ -156,6 +156,8 @@
     <add key="Gallery.RejectSignedPackagesWithNoRegisteredCertificate" value="false"/>
     <add key="Gallery.RejectPackagesWithTooManyPackageEntries" value="false"/>
     <add key="Gallery.BlockSearchEngineIndexing" value="false"/>
+    <add key="Gallery.IgnoreIconUrl" value ="false"/>
```
skofman1: regarding "IgnoreIconUrl", can you explain how it's going to be used and when?

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -45,7 +46,8 @@ public partial class UsersController
             ISecurityPolicyService securityPolicyService,
             ICertificateService certificateService,
             IContentObjectService contentObjectService,
-            IMessageServiceConfiguration messageServiceConfiguration)
+            IMessageServiceConfiguration messageServiceConfiguration,
+            IIconUrlProvider iconUrlProvider)
```
skofman1: move IIconUrlProvider to the base (AccountsController)

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -383,6 +383,14 @@ protected override void Load(ContainerBuilder builder)
                 .As<IPackageUpdateService>()
                 .InstancePerLifetimeScope();
 
+            builder.RegisterType<ConfigurationIconFileProvider>()
+                .As<IIconUrlProvider>()
+                .InstancePerLifetimeScope();
+
+            builder.RegisterType<IconUrlTemplateProcessor>()
+                .As<IIconUrlTemplateProcessor>()
+                .InstancePerLifetimeScope();
```
skofman1: why not singleton?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -383,6 +383,14 @@ protected override void Load(ContainerBuilder builder)
                 .As<IPackageUpdateService>()
                 .InstancePerLifetimeScope();
 
+            builder.RegisterType<ConfigurationIconFileProvider>()
+                .As<IIconUrlProvider>()
+                .InstancePerLifetimeScope();
```
skofman1: why not singleton?

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
```diff
@@ -12,8 +12,15 @@ namespace NuGetGallery
...
-        public ManagePackageViewModel(Package package, User currentUser, IReadOnlyList<ReportPackageReason> reasons, UrlHelper url, string readMe, bool isManageDeprecationEnabled)
-            : base(package, currentUser)
+        public ManagePackageViewModel(
+            Package package,
+            User currentUser,
+            IReadOnlyList<ReportPackageReason> reasons,
+            UrlHelper url,
+            string readMe,
+            bool isManageDeprecationEnabled,
+            string overrideIconUrl)
```
skofman1: Many of the viewmodels now have "overrideIconUrl". I wonder if this pattern can be avoided.

skofman1:
>         /// These are properties generated in the V3 pipeline (feed2catalog job) and could collide if the .nuspec

can you rename to Db2Catalog? :)

---
Refers to: src/NuGetGallery.Core/Packaging/PackageMetadata.cs:17 in 50d5307. [](commit_id = 50d5307ec472ad9ff10b10e90f91bdb12cdcd426, deletion_comment = False)

src/NuGetGallery.Core/IAppConfiguration.cs
agr: I had it renames to `EmbeddedIconUrlTemplate` and made it a template instead of base url.

src/NuGet.Services.Entities/Package.cs
agr: This flag will be set for all packages with icons eventually, even ones that don't have embedded icon (when we ingest external icons into our storage), so the name `HasSelfContainedIcon` is a bit too specific for the usage.

src/NuGetGallery/ViewModels/ManagePackageViewModel.cs
agr: Yeah, sure, I filed an issue to address that: https://github.com/NuGet/Engineering/issues/2510. It's just not going to be an easy fix.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
agr: As a general rule, I prefer not to have singletons unless we really need them. Allows to avoid a host of potential issues with multi-threading.

skofman1: On the other hand creating an instance on every request creates additional CPU+memory load..

src/NuGet.Services.Entities/Package.cs
agr: Spoke offline. Will make a change to have that flag to reflect the presence of an embedded icon metadata in .nuspec and will handle the external URLs after ingestion in a different way. Will rename to `HasEmbeddedIcon` to align with the existing `EmbeddedLicenseType` property.

src/NuGetGallery.Core/Services/StringReplaceTemplateProcessor.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Processes templates by running a series of <see cref="string.Replace(string, string)"/> calls against a template.
+    /// Not very efficient but simple.
+    /// </summary>
+    /// <typeparam name="TInput">The data source type for placeholder substitution.</typeparam>
+    /// <remarks>
+    /// Known issue:
```
cristinamanum: Is this comment intentionally?

src/NuGetGallery/ViewModels/DeleteAccountViewModel.cs
```diff
@@ -53,8 +55,9 @@ public class DeleteAccountListPackageItemViewModel : ListPackageItemViewModel
             Package package, 
             User userToDelete, 
             User currentUser, 
-            IPackageService packageService)
-            : base(package, currentUser)
+            IPackageService packageService,
+            string overrideIconUrl)
+            : base(package, currentUser, overrideIconUrl)
```
cristinamanum: Just an idea. Would not be simpler to have the packageService deal with the icon provider and the construct the iconUrl? In this way maybe the IconUrlProvider  is not needed to be added to many of the View Models?

src/NuGetGallery/Infrastructure/PackageIndexEntity.cs
```diff
@@ -13,13 +13,14 @@ public class PackageIndexEntity
...
-        public Package Package { get; set; }
+        private readonly IIconUrlProvider _iconUrlProvider;
 
-        public PackageIndexEntity() { }
+        public Package Package { get; set; }
 
-        public PackageIndexEntity(Package package)
+        public PackageIndexEntity(Package package, IIconUrlProvider iconUrlProvider)
         {
             this.Package = package;
```
skofman1: >this [](start = 12, length = 4)

we don't use "this." in the code base.. might as well remove from here if you are touching the code.

src/NuGetGallery/Migrations/201906262145129_EmbeddedIconFlag.cs
```diff
@@ -0,0 +1,18 @@
+namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class EmbeddedIconFlag : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.Packages", "HasEmbeddedIcon", c => c.Boolean(nullable: false));
```
skofman1: should we provide a default value here or EF auto sets it?
What about packages that are pushed with old code and new schema (after migration but before deployment), will the have this value set to false?

src/NuGetGallery/Services/ConfigurationIconFileProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+                // never fall back to iconUrl if HasEmbeddedIcon is true
+                return GetIconUrlFromTemplate(package);
+            }
+            else
+            {
+                if (string.IsNullOrWhiteSpace(package.IconUrl))
+                {
+                    return null;
+                }
+                if (_configuration.IgnoreIconUrl)
```
skofman1: >IgnoreIconUrl [](start = 35, length = 13)

IgnoreIconUrl is a little confusing. How about "UseHostedIcon"?

src/NuGetGallery/Services/IconUrlTemplateProcessor.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using NuGet.Services.Entities;
+using NuGetGallery.Configuration;
+
+namespace NuGetGallery
+{
+    public class IconUrlTemplateProcessor : StringReplaceTemplateProcessor<Package>, IIconUrlTemplateProcessor
```
skofman1: Using DI and Inheritance here makes the code complex and less readable. What is the benefit here? I recommend moving this to a static class and keeping it simple.

src/NuGetGallery.Core/Services/StringReplaceTemplateProcessor.cs
agr: Yes.

src/NuGetGallery/ViewModels/DeleteAccountViewModel.cs
agr: Mutating models with values that don't match the DB state sounds like a bad idea in the long run. Instead I changed the approach to instantiating the ViewModels.

src/NuGetGallery/Services/ConfigurationIconFileProvider.cs
agr: The idea is that when this property is set, Gallery will ignore any value passed in `IconUrl` metadata property. It does not enforce anything else.
Would `IgnoreIconUrlMetadata` be better?

src/NuGetGallery/Services/IconUrlTemplateProcessor.cs
agr: I'm not sure I follow. Making it static creates a hard dependency between the controller that uses it and the implementation of template processing, which complicates the unit testing (essentially making the template processing part of the controller code that cannot be tested in isolation).
I also didn't want to have template processing that is hard-coded for the purpose of icon URL specifically, so I added the generic processor and its specialization for icon URL template.

agr: `IIconUrlTemplateProcessor` is essentially a marker interface to simplify the DI setup.

src/NuGetGallery/Migrations/201906262145129_EmbeddedIconFlag.cs
agr: EF implicitly creates a default constraint for this column with `false` being the default value. After the migration is applied and before new code is deployed all new rows in `Packages` table will have it set to `false`.

src/NuGetGallery/Infrastructure/PackageIndexEntity.cs
```diff
@@ -13,13 +13,14 @@ public class PackageIndexEntity
...
         internal static readonly char[] IdSeparators = new[] { '.', '-' };
 
-        public Package Package { get; set; }
+        private readonly IIconUrlProvider _iconUrlProvider;
 
-        public PackageIndexEntity() { }
+        public Package Package { get; set; }
 
-        public PackageIndexEntity(Package package)
+        public PackageIndexEntity(Package package, IIconUrlProvider iconUrlProvider)
```
joelverhagen: This is is taking a different approach than the view models. Why?

src/NuGetGallery/Services/IInternalIconUrlProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Abstracts the logic of getting the package's icon URL.
+    /// </summary>
+    public interface IIconUrlProvider
```
joelverhagen: >IIconUrlProvider [](start = 21, length = 16)

File name doesn't match type

src/NuGetGallery/Services/ConfigurationIconFileProvider.cs
joelverhagen: >IgnoreIconUrl [](start = 35, length = 13)

IgnoreIconUrl doesn't seem strictly necessary to implement this yet since the icon mirroring does not have a design spec yet.

src/NuGetGallery.Core/Services/StringReplaceTemplateProcessor.cs
```diff
@@ -0,0 +1,69 @@
...
+        private readonly string _template;
+        private readonly IReadOnlyDictionary<string, Func<TInput, string>> _placeholderProcessors;
+
+        /// <summary>
+        /// Initializes the instance.
+        /// </summary>
+        /// <param name="template">String template. Can be null, in that case <see cref="Process(TInput)"/> will always produce null, too.</param>
+        /// <param name="placeholderProcessors">The map of placeholders to delegates that extract the appropriate value 
+        /// to substitute with from the <typeparamref name="TInput"/> object. Placeholders are case-sensitive.</param>
+        public StringReplaceTemplateProcessor(string template, IReadOnlyDictionary<string, Func<TInput, string>> placeholderProcessors)
```
joelverhagen: >StringReplaceTemplateProcessor [](start = 15, length = 30)

Is this really better than two string replaces?

agr: It is reusable.

joelverhagen: Okay, if you think it's worth the additional code. Honestly I (personally) would probably not think of finding an existing reusable component for such a simple thing. However maybe I am oversimplifying some edge cases that can be covered in this shared thing. Your call just wanted to ask the question.

agr: I received a feedback that [PackageContentUriBuilder](https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Catalog/Helpers/PackageContentUriBuilder.cs) does essentially template processing and was advised to try and reuse it. There is not much to reuse though and we'd end up with the copy of the class with some minimal changes, instead I chose to generalize it, so we would have actual reusable code.

PR #7272, opened 2019-06-18
First Draft Delete Job

ryuyu:
This is the first look at the new background delete job.

Includes:
 - Starter suite of tests
 - Working structure
 - Almost full external configuration.

Missing:
Ability to actually delete. This is requiring more refactoring on the side of gallery and will be added a bit later.

Any comments or concerns are appreciated. Thanks.


src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+                if (baseEmailBuilder != null)
+                {
+                    var recipientEmail = await _accountManager.GetEmailAddresForUser(username);
+
+                    var toEmail = new List<MailAddress>();
+
+                    var configuration = _options.Value;
+                    var senderAddress = configuration.EmailConfiguration.GalleryOwner;
+                    var ccEmail = new List<MailAddress>();
+                    // toEmail.Add(new MailAddress(recipientEmail)); // Temporarily disable sending to end user while we are in phase 1.
```
scottbommarito: I forgot what "phase 1" and "phase 2" refer to. Can you explain it again?

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using Microsoft.Extensions.Logging;
+using NuGet.Services.Entities;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class AggregateEvaluator : IUserEvaluator
```
scottbommarito: To be honest, I don't think we need a very complex system for user evaluation. I don't expect us to be doing anything more than "does the user have any packages?" and "does the user have any organizations?", which are one line statements. Adding a system of several `IUserEvaluator`s and an `AggregateEvaluator` to tie them together just so that we can split the single-line checks into different files doesn't seem necessary to me.

src/AccountDeleter/AccountDeleteMessage.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public class AccountDeleteMessage
+    {
+        public AccountDeleteMessage(string subject, string source)
+        {
+            Subject = subject;
+            Source = source;
+        }
+
+        public string Subject { get; }
```
scottbommarito: This is the username to be deleted, right? This is very confusing when you consider that "subject" is a very specific term for emails, and this job sends a lot of emails. I would rather just see this parameter named `Username`.

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async Task<bool> HandleAsync(AccountDeleteMessage command)
+        {
+            var messageProcessed = true;
+            var username = command.Subject;
+            _logger.LogInformation("Processing Request from Source {Source}", command.Source);
+
+            var source = command.Source;
```
scottbommarito: I don't really see need the need for `source`s. Why do we need to distinguish between messages sent by the gallery or by the DSR job? Is the delete process not going to be the same? If it's not completely necessary to distinguish between different message sources, I think we should get rid of it.

Furthermore, this seems like an improper way to do a queue...if these messages need to be handled in different ways, they should be in different queues. Imagine if we had multiple dequeuers that handled messages in different ways, and the wrong dequeuer happened to dequeue and fail a message so many times that it got deadlettered.

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+            var deleteSuccess = await _accountManager.DeleteAccount(username);
+
+            try
+            {
+                var baseEmailBuilder = _emailBuilderFactory.GetEmailBuilder(source, deleteSuccess);
+                if (baseEmailBuilder != null)
+                {
+                    var recipientEmail = await _accountManager.GetEmailAddresForUser(username);
```
scottbommarito: `Addres` -> `Address`

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                // Should definitely log if source isn't expected. Should we even send mail? or log and alert?
+                // Log unknown source and fail.
+                _logger.LogError("Unknown message source detected: {Source}.", command.Source);
+                messageProcessed = false;
+            }
+            catch (EmailContactNotAllowedException)
+            {
+                // Should we not send? or should we ignore the setting.
+                _logger.LogWarning("User did not allow Email Contact.");
```
scottbommarito: This is not a warning--it's an entirely expected and valid state.

src/AccountDeleter/GalleryAccountManager.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            _logger.LogInformation("Attempting delete...");
+            var user = _userService.FindByUsername(username);
+            if (user == null)
+            {
+                _logger.LogWarning("Requested user was not found in DB. Assuming delete was already done.");
+                return true;
+            }
+
+            if (_userEvaluator.CanUserBeDeleted(user))
```
scottbommarito: I would reduce nesting here by inverting this check (to `if (!_userEvaluator.CanUserBeDeleted(user))`).

src/AccountDeleter/AccountDeleteMessageHandler.cs
scottbommarito: Haven't we already deleted the user by this point (with `await _accountManager.DeleteAccount(username)`)? This call with never work when `_accountManager.DeleteAccount` actually deletes an account.

src/AccountDeleter/GalleryAccountManager.cs
```diff
@@ -0,0 +1,89 @@
...
+                }
+            }
+
+            _logger.LogInformation("User was not able to be automatically deleted. Criteria check failed.");
+            return false;
+        }
+
+        public Task<string> GetEmailAddresForUser(string username)
+        {
+            var user = _userService.FindByUsername(username);
```
scottbommarito: We have already grabbed the user by this point in `DeleteAccount(string username)`.

Why don't we instead have
- `AccountDeleteMessageHandler` should get the `User` object itself
- `DeleteAccount` takes in a `User` instead of a `username`
- `GetEmailAddressForUser` takes in a `User` instead of a `username`

src/AccountDeleter/Messaging/DisposableEmailBuilder.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.Messaging.Email;
+using System;
+using System.Net.Mail;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class DisposableEmailBuilder : IEmailBuilder
```
scottbommarito: `Disposable`? what does this mean in this context?

src/AccountDeleter/Messaging/ITemplater.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.AccountDeleter
+{
+    public interface ITemplater
```
scottbommarito: What complex templates do you expect to see? I expect that you will only be passing in the user's username to the email template. I think you could simplify all this stuff substantially.

tests/AccountDeleter.Facts/EvaluatorFacts.cs
```diff
@@ -0,0 +1,167 @@
...
+using NuGetGallery;
+using NuGetGallery.AccountDeleter;
+using System;
+using System.Collections.Generic;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace NuGetGallery.AccountDeleter.Facts
+{
+    public class EvaluatorFacts
```
scottbommarito: I think it would be more readable if you split up tests by class that's being tested.

src/AccountDeleter/AccountDeleteMessage.cs
```diff
@@ -0,0 +1,18 @@
...
+
+        public string Source { get; }
```
cristinamanum: Please add comments.

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        public async Task<bool> HandleAsync(AccountDeleteMessage command)
+        {
+            var messageProcessed = true;
+            var username = command.Subject;
+            _logger.LogInformation("Processing Request from Source {Source}", command.Source);
+
+            var source = command.Source;
+            var deleteSuccess = await _accountManager.DeleteAccount(username);
```
cristinamanum: What happens if this throws? How is the exception handled?

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+                    var configuration = _options.Value;
+                    var senderAddress = configuration.EmailConfiguration.GalleryOwner;
+                    var ccEmail = new List<MailAddress>();
+                    // toEmail.Add(new MailAddress(recipientEmail)); // Temporarily disable sending to end user while we are in phase 1.
+                    toEmail.Add(new MailAddress(senderAddress)); // Remove this when we switch to phase 2.
+                    ccEmail.Add(new MailAddress(senderAddress));
+
+                    var recipients = new EmailRecipients(toEmail, ccEmail);
+                    var emailBuilder = new DisposableEmailBuilder(baseEmailBuilder, recipients, username, new AccountDeleteTemplater(_options));
+                    await _messenger.SendMessageAsync(emailBuilder);
```
cristinamanum: Is an email sent to the user always? I do not think that we need this. We do not have this behavior right now.

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+                // Should we not send? or should we ignore the setting.
+                _logger.LogWarning("User did not allow Email Contact.");
+            }
+            catch (Exception e)
+            {
+                _logger.LogError("An unknown exception occured: {ExceptionMessage}", e.Message);
+                throw e;
+            }
+
+            return messageProcessed;
```
cristinamanum: If the delete fails but the email action does not throw this method will return true. Is this expected ?

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: Source is so that the message follow-up can be different. As of today, we have different message templates that we use for gallery deletes and DSR deletes. Having this allows us to differentiate without having to add a new queue to listen to every time we want to add a source. The messages aren't really handled in different ways, they go through the same path, the only switch is when retrieving message contents from settings.

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: It is still a warning in this context. It is perfectly valid state in the context of a user. It is a warning here if we didn't send a mail, but expected to.

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: Phase 1 was getting the job in place and running with dSR, but not adding any new behaviour. This means that we want to limit mail sending to just us (nugetteam) for right now since DSR has this behaviour.

Phase 2 will be enabling delete and sending mail to end user from DSR source.

src/AccountDeleter/AccountDeleteMessageSerializer.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.ServiceBus;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class AccoundDeleteMessageSerializer : IBrokeredMessageSerializer<AccountDeleteMessage>
```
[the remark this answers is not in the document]
ryuyu: Fixed.

src/AccountDeleter/AccountDeleteMessage.cs
ryuyu: I used subject to be consistent with how DSR was calling the identifying information, but i see your point. we can call it username.

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: fixed.

ryuyu: That is an excellent point. This should be retrieved BEFORE we attempt the delete. Good catch.

src/AccountDeleter/AccountDeleteMessageHandler.cs
scottbommarito: Is it necessary for us to have different messages from different sources though? What are the messages that you expect to emit for gallery deletes and DSR deletes?

Did you consider having the message template as part of the queue message instead, so this job doesn't need to be updated every time a message source is added or changed?

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: Good idea. done.

src/AccountDeleter/Messaging/DisposableEmailBuilder.cs
ryuyu: Disposable as in we will throw it away after this run. Better name might be scoped.

src/AccountDeleter/Messaging/ITemplater.cs
ryuyu: We will need to at least fill in URL in the email (saying "we delete profile https://stuff/Username"). Also allows us to add more stuff later that can switch by environment if we want without having to rebuild binaries.

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: If this throws, the thread dies and the message is returned to queue and retried later. I'm not sure if it gets logged, I will have to check.

src/AccountDeleter/Messaging/ITemplater.cs
scottbommarito: The URL in the email is going to be the same but with the username different as well...so the only templated parameter you need is the username.

I would argue we shouldn't write this type of system until we know we need more than a single parameter.

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: For now, the only source will be DSR, and the only mail will be to us, so we will send always.
Settings exposes a setting that will allow us to send mail only if the message could not be handled automatically, and allows us to choose to send a message on delete completion or not.

src/AccountDeleter/Messaging/DisposableEmailBuilder.cs
scottbommarito: I would argue that `TemplatedEmailBuilder` is a better name.

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: Sure

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: Discussed offline. If delete fails but does not throw exception, this means that the criteria check failed. if this happens, we want to send mail, and as long as this process completes, message can be considered processed (since we don't want to send duplicate mail).

joelverhagen:
This PR is massive. Could it be split into a two or more PRs?

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+                var username = command.Username;
+                _logger.LogInformation("Processing Request from Source {Source}", command.Source);
+
+                var source = command.Source;
+                _options.Value.VerifySource(source);
+
+                var user = _userService.FindByUsername(username);
+                var recipientEmail = await _accountManager.GetEmailAddressForUser(user);
+                var deleteSuccess = await _accountManager.DeleteAccount(user);
+
```
cristinamanum: Could we have a metric for the "deleteSuccess". It will be interesting to see how many success / failed delete happened.

src/AccountDeleter/Configuration/GalleryConfiguration.cs
```diff
@@ -0,0 +1,97 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Mail;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class GalleryConfiguration : IAppConfiguration
```
cristinamanum: What is the need for this class?

src/AccountDeleter/Configuration/SourceConfiguration.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.AccountDeleter
+{
+    // Should evaluator config be in here? should we even allow different evaluator config for different sources?
+    public class SourceConfiguration
```
cristinamanum: Please add comments for this class.

src/AccountDeleter/Job.cs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            if (IsDebugMode)
+            {
+                services.AddSingleton<IDeleteAccountService, EmptyDeleteAccountService>();
+                services.AddSingleton<IUserService, EmptyUserService>();
+            }
+            else
+            {
+                services.AddSingleton<IDeleteAccountService, EmptyDeleteAccountService>();
+                //services.AddSingleton<IDeleteAccountService, DeleteAccountService>();
```
cristinamanum: Commented code?

src/AccountDeleter/EmptyDeleteAccountService.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+using NuGetGallery.Areas.Admin.ViewModels;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class EmptyDeleteAccountService : IDeleteAccountService
```
cristinamanum: Why is the EmptyDeleteAccountService needed?

src/AccountDeleter/EmptyIndexingService.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class EmptyIndexingService : IIndexingService
```
cristinamanum: What is the purpose of the EmptyIndexingService ?

src/AccountDeleter/EmptyUserService.cs
```diff
@@ -0,0 +1,138 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class EmptyUserService : IUserService
```
cristinamanum: Is this intended for Unit testing? If yes why not added in a different UT project?

src/AccountDeleter/Evaluators/AlwaysAllowEvaluator.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+using NuGet.Services.Entities;
+using System;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class AlwaysAllowEvaluator : IUserEvaluator
```
cristinamanum: Why needed here. It seems used in testing. Why not adding it at the test project level?

src/AccountDeleter/Evaluators/AlwaysRejectEvaluator.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+using NuGet.Services.Entities;
+using System;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class AlwaysRejectEvaluator : IUserEvaluator
```
cristinamanum: Same as above.

src/AccountDeleter/Evaluators/UserOrganizationEvaluator.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+using NuGet.Services.Entities;
+using System;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class UserOrganizationEvaluator : IUserEvaluator
```
cristinamanum: This is not used? Will it be used in the future. Why not adding it when it will be used?

src/AccountDeleter/GalleryAccountManager.cs
```diff
@@ -0,0 +1,85 @@
...
+    public class GalleryAccountManager : IAccountManager
+    {
+        private readonly IOptionsSnapshot<AccountDeleteConfiguration> _options;
+        private readonly IDeleteAccountService _deleteAccountService;
+        private readonly IUserEvaluator _userEvaluator;
+        private readonly IAccountDeleteTelemetryService _telemetryService;
+        private readonly ILogger<GalleryAccountManager> _logger;
+
+        public GalleryAccountManager(
+            IOptionsSnapshot<AccountDeleteConfiguration> options,
```
cristinamanum: nit: Why called options and not accountDeleteConfiguration or similar?

src/AccountDeleter/GalleryAccountManager.cs
```diff
@@ -0,0 +1,85 @@
...
+                _logger.LogInformation("User was not able to be automatically deleted. Criteria check failed.");
+                return false;
+            }
+
+            _logger.LogInformation("All criteria passed.");
+            var result = await _deleteAccountService.DeleteAccountAsync(user, user, AccountDeletionOrphanPackagePolicy.UnlistOrphans);
+            if (result.Success)
+            {
+                _logger.LogInformation("Deleted user successfully.");
+                _telemetryService.TrackAccountDelete();
```
cristinamanum: Should not be TrackAccountDelete with a status ?

src/AccountDeleter/GalleryAccountManager.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+
+            if (!_userEvaluator.CanUserBeDeleted(user))
+            {
+                _logger.LogInformation("User was not able to be automatically deleted. Criteria check failed.");
+                return false;
+            }
+
+            _logger.LogInformation("All criteria passed.");
+            var result = await _deleteAccountService.DeleteAccountAsync(user, user, AccountDeletionOrphanPackagePolicy.UnlistOrphans);
```
cristinamanum: user, user , .. are sued for args, is this intended? Can we have named args?

src/AccountDeleter/GalleryAccountManager.cs
```diff
@@ -0,0 +1,85 @@
...
+                _logger.LogError("Criteria passed but delete failed.");
+                _telemetryService.TrackAccountDelete();
+                return false;
+            }
+        }
+
+        public Task<string> GetEmailAddressForUser(User user)
+        {
+            // We may want to ignore this setting, but respect contact for now
+            if(!user.EmailAllowed)
```
cristinamanum: Is a null check needed ?

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+            try
+            {
+                var username = command.Username;
+                _logger.LogInformation("Processing Request from Source {Source}", command.Source);
+
+                var source = command.Source;
+                _options.Value.VerifySource(source);
+
+                var user = _userService.FindByUsername(username);
```
cristinamanum: If the user is null, because it was deleted the GetEmailAddressForUser will throw NullArg Exception that will go on line 907 and the message will be re-queued. 

Would not be better if in case of null user to return true (and record some telemetry metrics)?

src/AccountDeleter/GalleryAccountManager.cs
```diff
@@ -0,0 +1,85 @@
...
+                {
+                    throw new EmailContactNotAllowedException();
+                }
+                else
+                {
+                    _logger.LogWarning("Ignoring EmailAllowed due to configuration.");
+                }
+            }
+
+            return Task.FromResult(user.EmailAddress ?? user.UnconfirmedEmailAddress);
```
cristinamanum: We should not send emails to unconfirmed email addresses.

src/AccountDeleter/Job.cs
```diff
@@ -0,0 +1,181 @@
...
+    public class Job : SubcriptionProcessorJob<AccountDeleteMessage>
+    {
+        private const string AccountDeleteConfigurationSectionName = "AccountDeleteSettings";
+        private const string DebugArgumentName = "Debug";
+
+        private bool IsDebugMode { get; set; }
+
+        public override void Init(IServiceContainer serviceContainer, IDictionary<string, string> jobArgsDictionary)
+        {
+            IsDebugMode = jobArgsDictionary.ContainsKey(DebugArgumentName);
```
cristinamanum: Do we need to deploy the job in "Debug" mode? Why do we need this mode?

src/AccountDeleter/Job.cs
```diff
@@ -0,0 +1,181 @@
...
+                    var alwaysAllow = sp.GetRequiredService<AlwaysAllowEvaluator>();
+
+                    evaluator.AddEvaluator(alwaysReject);
+                    evaluator.AddEvaluator(alwaysAllow);
+                }
+                else
+                {
+                    // Configure evaluators here.
+                    var alwaysReject = sp.GetRequiredService<AlwaysRejectEvaluator>();
+                    evaluator.AddEvaluator(alwaysReject);
```
cristinamanum: If I understand correctly AlwaysRejectEvaluator + Any other evaluator == AlwaysRejectEvaluator. Why do we need AlwaysRejectEvaluator?

src/AccountDeleter/Configuration/GalleryConfiguration.cs
ryuyu: This is being used so we can create an IAppConfiguration for DI, and only provide exactly the settings we have need. I wasn't sure how to use AppConfiguration with a bunch of nulls, so i create this shell that only defines the ones we need. I'm happy to drop it if we can get AppConfiguration to work with partial configuration.

src/AccountDeleter/Job.cs
ryuyu: This was intentionally commented as it is the line we eventually want to use. However, we don't want the job to do any actual deleting in phase 1, and the DI hydration for using a real DeleteAccount service was a much larger effort than i anticipated. I left this here so it was clear that the EmptyDeleteAccount was eventually NOT what we wanted. I can comment the comment for clarity.

src/AccountDeleter/EmptyDeleteAccountService.cs
ryuyu: This is to have an IDeleteAccountService that is usable while DeleteAccountService is missing things in DI.

src/AccountDeleter/EmptyIndexingService.cs
ryuyu: This was part of an effort to stub out some of the services that DeleteAccountService needed. I think we don't need it anymore. I will check and remove.

src/AccountDeleter/EmptyUserService.cs
ryuyu: This was used for debugging, and is used when the -Debug flag is passed in. This is a pattern that we used in DSR that i continued here. It is technically not used at real runtime.

src/AccountDeleter/Evaluators/AlwaysAllowEvaluator.cs
ryuyu: We could move it to the test project. I was using this during my debugging, so it seemed natural to add it here as I needed it at the time. We can move it.

src/AccountDeleter/Evaluators/AlwaysRejectEvaluator.cs
ryuyu: This we actually want to use for now. We don't want the job (at this time) to actually delete any account, so by adding this evaluator, we can guarantee that all requests will flag as "cannot be auto deleted"

src/AccountDeleter/Evaluators/UserOrganizationEvaluator.cs
ryuyu: It will be used in the future. When i was making these, I stubbed out all the evaluators that I thought we would be using. Implementation is not complete. I can delete until we need if you thikn that is cleaner.

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: This was suggested naming by @joelverhagen . 

I can change it to something like accountDeleteConfigurationAccessor or something if that is more clear.

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: Yes. Telemetry implementation is not yet complete.

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: Yes we can add named args. This was intended. The call the DeleteAccountAsync requests user being delete, and also the user doing the delete (to check permissions).

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: I changed it.

src/AccountDeleter/Configuration/SourceConfiguration.cs
ryuyu: I added a comment to the top indicating what this class was for.

src/AccountDeleter/Job.cs
ryuyu: I'll add a comment about this here. We are adding the always reject evaluator for right now because we never want the job to attempt to delete at this time. During phase 2, when we enable real deletes, we will remove this evaluator and add in real evaluators.

src/AccountDeleter/Job.cs
ryuyu: Debug mode is for running locally for debugging purposes. Christy started this pattern in DSR agent and I used it here. With real configuration finished being set up. the value is limited, so we can remove it if you think it is cluttering things.

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: Good point. Removed.

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: Hmm. This is an excellent point. I will add this.

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: Added null check.

src/AccountDeleter/AccountDeleteMessageHandler.cs
scottbommarito: If the user does not allow email contact, the expected behavior is to not send a mail, so I don't see how this is a warning. Not a big deal regardless.

src/AccountDeleter/GalleryAccountManager.cs
scottbommarito: missing space before `(`

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: Added

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+                }
+            }
+            catch (UnknownSourceException)
+            {
+                // Should definitely log if source isn't expected. Should we even send mail? or log and alert?
+                // Log unknown source and fail.
+                _logger.LogError("Unknown message source detected: {Source}.", command.Source);
+                messageProcessed = false;
+            }
+            catch (EmailContactNotAllowedException)
```
scottbommarito: If the evaluators say the user shouldn't be deleted, but the user doesn't allow email contact, should we delete the user anyway? As far as I can tell, the current flow for this code is that the user is not deleted and the email is not sent. How will we know to clean up the user?

src/AccountDeleter/Evaluators/IUserEvaluator.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.Entities;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public interface IUserEvaluator
```
scottbommarito: Maybe `IAccountDeletionEvaluator` would be a better name?

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: This is a good point. We should figure out exactly what the behaviour should be in this case.
For DSR, the DSR support request would still be filed, and we can use this as flag to clean up the user. From gallery, we could do soemthing similar (ie continue to let the flow log the support request).

The configs will cause us to ignore the AllowEmailConfig setting for now, so the path will not be hit. But it's a good point that we should figure this out.

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+                var baseEmailBuilder = _emailBuilderFactory.GetEmailBuilder(source, deleteSuccess);
+                if (baseEmailBuilder != null)
+                {
+                    var toEmail = new List<MailAddress>();
+
+                    var configuration = _accountDeleteConfigurationAccessor.Value;
+                    var senderAddress = configuration.EmailConfiguration.GalleryOwner;
+                    var ccEmail = new List<MailAddress>();
+                    // toEmail.Add(new MailAddress(recipientEmail)); // Temporarily disable sending to end user while we are in phase 1.
+                    toEmail.Add(new MailAddress(senderAddress)); // Remove this when we switch to phase 2.
```
agr: >// Remove this when we switch to phase 2. [](start = 65, length = 41)

Do we have an issue tracking Phase 2? Put a link here maybe?

src/AccountDeleter/EmptyDeleteAccountService.cs
agr: Document that maybe?

---
In reply to: [295953365](https://github.com/NuGet/NuGetGallery/pull/7272#discussion_r295953365) [](ancestors = 295953365)

src/AccountDeleter/GalleryAccountManager.cs
```diff
@@ -0,0 +1,91 @@
...
+            if (result.Success)
+            {
+                _logger.LogInformation("Deleted user successfully.");
+                _telemetryService.TrackAccountDelete();
+                return true;
+            }
+            else
+            {
+                _logger.LogError("Criteria passed but delete failed.");
+                _telemetryService.TrackAccountDelete();
```
agr: >_telemetryService.TrackAccountDelete(); [](start = 16, length = 39)

Should it be some other metric if delete failed?

src/AccountDeleter/IAccountManager.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// </summary>
+        /// <param name="user">User entity to act on</param>
+        /// <returns>True if the account delete completes, false if account delete did not proceed for some reason.</returns>
+        Task<bool> DeleteAccount(User user); //Maybe userkey here? or an internal structure. Depends on when we resolve the user name against DB. same below. We want to limit db lookups if possible (re: minimize calls to services where possible).
+
+        /// <summary>
+        /// Try to get a contact email address for a user
+        /// </summary>
+        /// <param name="usere">User entity to act on</param>
+        /// <returns>An email address. Throw <see cref="EmailContactNotAllowedException"/> if email contact was disallowed</returns>
```
agr: >Throw <see cref="EmailContactNotAllowedException"/> if email contact was disallowed [](start = 39, length = 83)

Please, use `<exception>` for exceptions.

agr: >An email address. [](start = 21, length = 17)

The current implementation can return `null`. The documentation should probably cover it.

src/AccountDeleter/IAccountManager.cs
```diff
@@ -0,0 +1,25 @@
...
+        Task<string> GetEmailAddressForUser(User user);
```
agr: I'm not sure this method belongs here. Essentially, what it does is decides whether email should be sent or not (which is not reflected in the name by the way) as user's email is readily available to the caller through the `user.EmailAddress` property. Also, the decision is communicated in different ways: return value can be `null` or non-`null` and besides that, it can throw exception, which essentially mean "don't delete", right? That suspiciously looks like an attempt to do flow control with exceptions. I would suggest to move the method to the class that calls into it as it looks like the part of business logic belonging to that class, and maybe get rid of the exception?

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+                _telemetryService.TrackSource(source);
+                _accountDeleteConfigurationAccessor.Value.VerifySource(source);
+
+                var user = _userService.FindByUsername(username);
+                if (user == null)
+                {
+                    throw new UserNotFoundException();
+                }
+
+                var recipientEmail = await _accountManager.GetEmailAddressForUser(user);
```
agr: >await _accountManager.GetEmailAddressForUser(user); [](start = 37, length = 51)

If this line throws `_accountManager.DeleteAccount` will not be called, but if it returns `null` it will. Is that expected behavior?

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: Yes. Telemetry is not yet complete. I'll update this now. There is a new metric that was added earlier that should be appropriate here.

src/AccountDeleter/MicroUserService.cs
```diff
@@ -0,0 +1,147 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data.Entity;
+using System.Linq;
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class MicroUserService : IUserService
```
agr: >public class MicroUserService [](start = 4, length = 29)

Please, add the documentation regarding why it exists and implemented the way it is implemented. Maybe file an issue to consider splitting the `IUserService` into multiple more fine grained ones?

src/AccountDeleter/IAccountManager.cs
ryuyu: Done

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                return _id.ToString();
+            }
+        }
+
+        public bool CanUserBeDeleted(User user)
+        {
+            var evaluators = _evaluatorList.Values;
+            _logger.LogInformation("Running {EvaluatorCount} evaluators.", evaluators.Count);
+            return evaluators.All(e => e.CanUserBeDeleted(user));
```
agr: >           return evaluators.All(e => e.CanUserBeDeleted(user)); [](start = 1, length = 64)

Do we want to log or indicate is some other way which evaluator caused the method to return `false` when this happens?

src/AccountDeleter/IAccountManager.cs
ryuyu: Got rid of this method. Since it was updated to take a User type, it doesn't make a whole lot of sense to go to another thing to get the email. Really, this was left over from when we passed a string username around, and not the user object.

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
agr: >public class AggregateEvaluator [](start = 4, length = 31)

Please, add documentation for all evaluators.

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: Removed this as the metric is logged up the call stack.

src/AccountDeleter/Evaluators/UserOrganizationEvaluator.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            get
+            {
+                return _id.ToString();
+            }
+        }
+
+        public bool CanUserBeDeleted(User user)
+        {
+            return false;
```
agr: This looks awfully like `AlwaysRejectEvaluator`.

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: Removed this altogether and moved the check to respect email contact setting into this method.

src/AccountDeleter/MicroUserService.cs
ryuyu: Done

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
ryuyu: At this time, individual evaluators are responsible for logging results. As it is, not all the evaluators get called; it appears to stop as soon as one fails. I'm not 100% about that behaviour as we may want the result of all evaluators in logs.

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
ryuyu: Done

src/AccountDeleter/Evaluators/UserOrganizationEvaluator.cs
ryuyu: This evaluator is currently incomplete. I believe this was one of the criteria that will eventually be used, so i stubbed it here to track it.

ryuyu: Chagned to throw NotImplementedException.

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.Extensions.Logging;
+using NuGet.Services.Entities;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetGallery.AccountDeleter
+{
+    /// <summary>
+    /// Evaluator that runs other evaluators and returns a single result
```
agr: >Evaluator that runs other evaluators and returns a single result [](start = 8, length = 64)

It's probably worth mentioning here that it implements the "AND" logic.

ryuyu: Good Call. I'll add a note. May need to expand in the future, but for now, this will work.

src/AccountDeleter/MicroUserService.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery.AccountDeleter
+{
+    /// <summary>
+    /// This class implements the minimal function of the full <see cref="UserService"/> that we currently need for AccountDeleter
+    /// This is done for now in order to not pollute DI with things that we won't use.
+    /// </summary>
+    public class MicroUserService : IUserService
```
scottbommarito: We should be using the `ICoreUserService` and `IUserService` pattern that we established to allow the validation pipeline to use smaller implementations of other gallery services.

src/AccountDeleter/Configuration/SourceConfiguration.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.AccountDeleter
+{
+    // Should evaluator config be in here? should we even allow different evaluator config for different sources?
```
loic-sharma: Looks like this should be removed

src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            catch (UserNotFoundException)
+            {
+                _logger.LogWarning("User was not found. They may have already been deleted.");
+                _telemetryService.TrackUserNotFound(source);
+                messageProcessed = true;
+            }
+            catch (Exception e)
+            {
+                _logger.LogError("An unknown exception occured: {ExceptionMessage}", e.Message);
```
loic-sharma: Prefer `e.ToString()` over `e.Message` as the real error may be in an inner exception. In any case, I would use the logging framework:

```suggestion
                _logger.LogError(0, e, "An unknown exception occurred");
```

src/AccountDeleter/GalleryAccountManager.cs
```diff
@@ -0,0 +1,63 @@
...
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async Task<bool> DeleteAccount(User user)
+        {
+            _logger.LogInformation("Attempting delete...");
+            if (user == null)
+            {
+                _logger.LogWarning("Requested user was not found in DB. Delete was probalby already done.");
```
loic-sharma:
```suggestion
                _logger.LogWarning("Requested user was not found in DB. Delete was probably already done.");
```

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: Ah yes this is what should be used here. I think i did it somewhere else the right way.

src/AccountDeleter/Configuration/SourceConfiguration.cs
ryuyu: Removed.

tests/AccountDeleter.Facts/EmailBuilderFacts.cs
```diff
@@ -0,0 +1,65 @@
...
+    public class EmailBuilderFacts
+    {
+        private Mock<IEmailBuilder> _accountDeleteEmailBuilder;
+
+        public EmailBuilderFacts()
+        {
+            _accountDeleteEmailBuilder = new Mock<IEmailBuilder>();
+        }
+
+
```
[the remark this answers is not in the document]
ryuyu: Removed

src/AccountDeleter/MicroUserService.cs
ryuyu: I don't know if that is how we even want this split to go. This is a go between until we can decide how the services in gallery should break down. That is not part of this PR, and will be addressed later. This, as noted above, is minimalist function for this usage of the userService and this usage only.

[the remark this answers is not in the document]
ryuyu: Again, it's a hold over until we can decide exactly how it should break down. The notes in the heading explain what the purpose is.

src/AccountDeleter/Messaging/EmptyMessenger.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+using NuGet.Services.Messaging.Email;
+using System;
+using System.Threading.Tasks;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class EmptyMessenger : IMessageService
```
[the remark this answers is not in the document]
ryuyu: Renamed to DebugMessageService. Messenger was used before IMessageService interface was consumed (there was an IMessenger interface that was replaced with IMessageService).

src/AccountDeleter/AccountDeleteMessageHandler.cs
ryuyu: Code path is not broken. It is currently behaving as intended. The question is whether the intended behaviour is correct. Better behaviour might be to fail the request 10 times and let it deadletter.

src/AccountDeleter/MicroUserService.cs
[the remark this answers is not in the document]
scottbommarito: Ideally, the purpose of the class should be made relatively clear by its name alone, and the XML docs should not be necessary to do so. I think it's very unclear what `MicroUserService` is for. I would suggest `AccountDeleterUserService` because it conveys the fact that this is an `IUserService` for use by the `AccountDeleter`.

scottbommarito: Okay, sounds good. I would suggest either the `ICoreUserService` model or splitting this class into several more specific services when you do solve the breakdown.

src/AccountDeleter/Configuration/GalleryConfiguration.cs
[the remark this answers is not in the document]
ryuyu: There is no confusion already. The "real gallery configuration objects" are called AppConfiguration. In the context of AccountDeleter, being called GalleryConfiguration makes sense, as it is the configuration required to emulate gallery. Likely, more of this will need to be filled out as we require use of more of the gallery services. As it is, we want it to throw as we should know if we are using something that we aren't expecting to use. We could change the name to this one, but that would mean introducing another one later when we actually start needing configuration.

src/AccountDeleter/EmptyUserService.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public IList<User> FindByUnconfirmedEmailAddress(string unconfirmedEmailAddress, string optionalUsername)
+        {
+            throw new NotImplementedException();
+        }
+
+        public User FindByUsername(string username, bool includeDeleted = false)
```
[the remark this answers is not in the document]
ryuyu: Sure.

src/AccountDeleter/IAccountManager.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace NuGetGallery.AccountDeleter
+{
+    public interface IAccountManager
+    {
+        /// <summary>
+        /// Try to delete the account of the specified username
+        /// </summary>
+        /// <param name="user">User entity to act on</param>
+        /// <returns>True if the account delete completes, false if account delete did not proceed for some reason.</returns>
+        Task<bool> DeleteAccount(User user); //Maybe userkey here? or an internal structure. Depends on when we resolve the user name against DB. same below. We want to limit db lookups if possible (re: minimize calls to services where possible).
```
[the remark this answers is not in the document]
ryuyu: Removed.

src/AccountDeleter/IAccountManager.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.Entities;
+using System.Threading.Tasks;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public interface IAccountManager
```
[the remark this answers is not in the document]
ryuyu: It is not managing account deletes. It is managing an account. Currently, the only function we need is delete, so that is all we have here. Any further account management actions would be found here (were they to be necessary).

src/AccountDeleter/Job.cs
```diff
@@ -0,0 +1,181 @@
...
+            services.AddSingleton<AggregateEvaluator>();
+            services.AddTransient<AlwaysRejectEvaluator>();
+            services.AddTransient<AlwaysAllowEvaluator>();
+            services.AddTransient<UserPackageEvaluator>();
+
+            services.AddSingleton<IUserEvaluator>(sp =>
+            {
+                var evaluator = sp.GetRequiredService<AggregateEvaluator>();
+
+                if (IsDebugMode)
```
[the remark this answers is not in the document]
ryuyu: Good question. maybe evaluators should stay the same.
The idea with the flag is the switching over to a mode where it is easy to debug things would be nice to have. That being said, what you are trying to debug ends up being important in the case of trying to figure it out. If you don't care about debugging evaluators, it shouldn't be a blocker. However, if you wanted to debug evaluators, it wouldn't make sense to change them here.

That is on the debugger (re: person) to make the decision about what they are debugging and make adjustments to the debug paths as necessary.

src/AccountDeleter/Job.cs
```diff
@@ -0,0 +1,181 @@
...
+                //services.AddSingleton<IDeleteAccountService, DeleteAccountService>();
+
+                services.AddSingleton<IUserService, MicroUserService>();
+                services.AddSingleton<IDiagnosticsService, LoggerDiagnosticsService>();
+
+                services.AddSingleton<IPackageService, PackageService>();
+
+                services.AddSingleton<ITelemetryService, GalleryTelemetryService>();
+                services.AddSingleton<ISecurityPolicyService, SecurityPolicyService>();
+                services.AddSingleton<IAuditingService>(sp => { return AuditingService.None; }); //Replace with real when we start doing deletes. For now, we are a readonly operation.
```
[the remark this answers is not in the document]
ryuyu: As per the comment, we aren't doing any deletes right now.
Also, the auditing service only audits when we call it.
We do this in the gallery service implmentations, but it is not done nearly anywhere else.
If there was a bug causing other things to happen, it is unlikely that it would have been caught by this auditing in the first place, unless it happened to be called through a service that was calling the auditing in the first place.

src/AccountDeleter/AccountDeleteMessageHandler.cs
scottbommarito: Regardless of what the "intended behavior" is, we never expect to hit it because it's blocked by a config. We've agreed that we need to rethink how the code behaves here, but we don't even need this code in this PR because we plan for it to never execute.

I would remove any concept of the `AllowEmailConfig` setting from this PR and then add it back in a future PR that does the behavior in a way we can agree on is better.

src/AccountDeleter/Configuration/AccountDeleteConfiguration.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// Nested configurations for sending email messages.
+        /// </summary>
+        public EmailConfiguration EmailConfiguration { get; set; }
+
+        /// <summary>
+        /// Default template replacements list.
+        /// </summary>
+        public Dictionary<string, string> TemplateReplacements { get; set; }
+
+        public IEmailBuilder GetEmailBuilder(string source, bool success = false)
```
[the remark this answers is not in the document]
ryuyu: Move to EmailBuilderFactory. VerifySource will stay here as it is a concern of the configuration, and should live with the configuration instance.

PR #7273, opened 2019-06-18
Fix for "Resend confirmation email" button to use the correct email address

agr:
Attempt to address https://github.com/NuGet/Engineering/issues/2512

### DIFF
--- a/src/NuGetGallery/Controllers/AccountsController.cs
+++ b/src/NuGetGallery/Controllers/AccountsController.cs
@@ -111,10 +111,10 @@ namespace NuGetGallery
                 return new HttpStatusCodeResult(HttpStatusCode.Forbidden, Strings.Unauthorized);
             }
 
-            var alreadyConfirmed = account.UnconfirmedEmailAddress == null;
+            var hasUnconfirmedEmailAddress = account.UnconfirmedEmailAddress == null;
 
             ConfirmationViewModel model;
-            if (!alreadyConfirmed)
+            if (!hasUnconfirmedEmailAddress)
             {
                 if (account.EmailAddress == null)
                 {
--- a/src/NuGetGallery/Controllers/AccountsController.cs
+++ b/src/NuGetGallery/Controllers/AccountsController.cs
@@ -111,10 +111,10 @@ namespace NuGetGallery
                 return new HttpStatusCodeResult(HttpStatusCode.Forbidden, Strings.Unauthorized);
             }
 
-            var hasUnconfirmedEmailAddress = account.UnconfirmedEmailAddress == null;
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Controllers/AccountsController.cs
```diff
@@ -116,7 +116,14 @@ public virtual async Task<ActionResult> ConfirmationRequiredPost(string accountN
             ConfirmationViewModel model;
             if (!alreadyConfirmed)
```
scottbommarito: IMO this variable name is confusing

I would either
- inline the check (`account.UnconfirmedEmailAddress != null`)
- rename the variable to `hasUnconfirmedEmailAddress`

tests/NuGetGallery.Facts/Controllers/OrganizationsControllerFacts.cs
```diff
@@ -273,6 +273,24 @@ public async Task WithNonOrganizationAdmin_ReturnsForbidden(Func<Fakes, User> ge
                 Assert.NotNull(result);
                 Assert.Equal((int)HttpStatusCode.Forbidden, result.StatusCode);
             }
+
+            [Fact]
+            public async Task ResendsProperConfirmationEmail()
```
scottbommarito: We should add a test as well that verifies that a `NewAccountMessage` is sent when `account.EmailAddress` is null.

agr: Done

src/NuGetGallery/Controllers/AccountsController.cs
agr: Renamed the variable.

src/NuGetGallery/Controllers/AccountsController.cs
```diff
@@ -111,12 +111,19 @@ public virtual async Task<ActionResult> ConfirmationRequiredPost(string accountN
                 return new HttpStatusCodeResult(HttpStatusCode.Forbidden, Strings.Unauthorized);
             }
 
-            var alreadyConfirmed = account.UnconfirmedEmailAddress == null;
+            var hasUnconfirmedEmailAddress = account.UnconfirmedEmailAddress == null;
```
cristinamanum: When the account.UnconfirmedEmailAddress == null it means that the email is confirmed. Why the name changed?

agr: The rename was suggested by @scottbommarito: https://github.com/NuGet/NuGetGallery/pull/7273#discussion_r295047300

Issue #7281, opened 2019-06-19
[Deprecation] Packages show up in db2catalog slowly with big deprecation messages


joelverhagen:
Likely mitigated with
https://github.com/NuGet/NuGetGallery/issues/7280

I deprecated 80 packages. Each had a deprecation message of 1,702,389 characters. This look 218 seconds for db2catalog to process.

joelverhagen:
I think the SQL was slow:
![image](https://user-images.githubusercontent.com/94054/59790193-3724ae80-9284-11e9-8efe-1f3a3442d6ce.png)

Issue #7280, opened 2019-06-19
[Deprecation] Deprecation message has no reasonable length restriction


joelverhagen:
I put a deprecation message that has 1,702,389 characters.

scottbommarito:
Probably going to limit this to 4,000 characters, like how other similar fields are restricted (https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Helpers/PackageHelper.cs#L75).

joelverhagen:
Consider a number based on expected command line experience

joelverhagen:
Lower is better for starters. It's easy to increase the limit later.

Issue #7279, opened 2019-06-19
[Deprecation] URLs are not linkified in deprecation message


joelverhagen:
Deprecation message:
```
This is what you should see: https://www.nuget.org. Okay?
```

It should be linkified like description.

![image](https://user-images.githubusercontent.com/94054/59788353-5faaa980-9280-11e9-9df7-d8a4171ab0b4.png)

Issue #7278, opened 2019-06-19
[Deprecation] New lines in the deprecation message are not respected in HTML


joelverhagen:
I used this deprecation message:

```
Line 1

Line 3
```

This is the HTML. This behaves differently than package description.

![image](https://user-images.githubusercontent.com/94054/59788237-23774900-9280-11e9-9ded-9c8a5cd6c055.png)

PR #7284, opened 2019-06-19
Fix deprecation custom message issues

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/7277
https://github.com/NuGet/NuGetGallery/issues/7278
https://github.com/NuGet/NuGetGallery/issues/7279
https://github.com/NuGet/NuGetGallery/issues/7280
https://github.com/NuGet/NuGetGallery/issues/7281

The bulk of these changes were to move the deprecation request parameters to a model so I can add `[AllowHtml]` to the custom message. Otherwise, the changes were fairly trivial.

### DIFF
--- a/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
+++ b/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
@@ -6,10 +6,12 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
+using System.Web;
 using System.Web.Mvc;
 using NuGet.Services.Entities;
 using NuGet.Versioning;
 using NuGetGallery.Filters;
+using NuGetGallery.RequestModels;
 
 namespace NuGetGallery
 {
@@ -51,30 +53,23 @@ namespace NuGetGallery
         [RequiresAccountConfirmation("deprecate a package")]
         [ValidateAntiForgeryToken]
         public virtual async Task<JsonResult> Deprecate(
-            string id,
-            IEnumerable<string> versions,
-            bool isLegacy,
-            bool hasCriticalBugs,
-            bool isOther,
-            string alternatePackageId,
-            string alternatePackageVersion,
-            string customMessage)
+            DeprecatePackageRequest request)
         {
             var status = PackageDeprecationStatus.NotDeprecated;
 
-            if (isLegacy)
+            if (request.IsLegacy)
             {
                 status |= PackageDeprecationStatus.Legacy;
             }
 
-            if (hasCriticalBugs)
+            if (request.HasCriticalBugs)
             {
                 status |= PackageDeprecationStatus.CriticalBugs;
             }
 
-            if (isOther)
+            if (request.IsOther)
             {
-                if (string.IsNullOrWhiteSpace(customMessage))
+                if (string.IsNullOrWhiteSpace(request.CustomMessage))
                 {
                     return DeprecateErrorResponse(HttpStatusCode.BadRequest, Strings.DeprecatePackage_CustomMessageRequired);
                 }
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -6,17 +6,20 @@
...
-using NuGetGallery.Auditing;
 using NuGetGallery.Filters;
+using NuGetGallery.RequestModels;
 
 namespace NuGetGallery
 {
     public partial class ManageDeprecationJsonApiController
         : AppController
     {
+        private const int MaxCustomMessageLength = 4000;
```
scottbommarito: Chose this because it aligns with the restriction similar package metadata fields (https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Helpers/PackageHelper.cs#L75). I'm open to suggestions however.

xavierdecoster:
Changes look good but tests fail.

scottbommarito:
@xavierdecoster - fixed the issue with the tests

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
joelverhagen: How does 4000 look in the @xavierdecoster's client PR? 4000 seems pretty high. 

P95 of description length is 487.
P99 is 978.

Maybe start with 1000 assuming it looks like on the UI with a reasonable number of newlines.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -50,56 +53,62 @@ public virtual JsonResult GetAlternatePackageVersions(string id)
...
+            if (request.CustomMessage != null)
+            {
+                if (request.CustomMessage.Length > MaxCustomMessageLength)
+                {
+                    return DeprecateErrorResponse(
+                        HttpStatusCode.BadRequest,
+                        string.Format(Strings.DeprecatePackage_CustomMessageTooLong, MaxCustomMessageLength));
+                }
+
+                customMessage = HttpUtility.HtmlEncode(request.CustomMessage);
```
joelverhagen: We should not be storing encoded strings in the DB. HTML encoding is a view concern not a data concern.

src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
```diff
@@ -63,7 +63,7 @@
             @if (!string.IsNullOrEmpty(Model.CustomMessage))
             {
                 <b>Additional Details</b>
-                <p>@Model.CustomMessage</p>
+                <p>@Html.PreFormattedText(Model.CustomMessage)</p>
```
joelverhagen: HTML encode here, not in the data

We don't want HTML entities to flow into the catalog and into WPF eventually

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
joelverhagen: How does your 10k test perform if each message has the max size?

scottbommarito: sure

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: Sounds good.

src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
scottbommarito: MVC properly encodes it automatically before displaying it when I don't encode it before putting it in the DB.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -50,56 +52,62 @@ public virtual JsonResult GetAlternatePackageVersions(string id)
...
 
                 status |= PackageDeprecationStatus.Other;
             }
 
+            string customMessage = null;
+            if (request.CustomMessage != null)
+            {
+                if (request.CustomMessage.Length > MaxCustomMessageLength)
+                {
+                    return DeprecateErrorResponse(
```
joelverhagen: What's the behavior when they put in a message too long? Do they lose it all after submit?

scottbommarito: Nope, it's ajax so it just shows a warning message on the page. They can easily reduce the message and resubmit.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: Local gallery took roughly the same amount of time to process the request as before.

On DEV, I just tested the E2E perf with a 1k messages. I observed a full E2E (including Azure search) in 41 minutes, which is slightly slower than before. The major impacted party appears to be Db2Catalog. I am testing the E2E perf without empty messages and will report back on how it performs.

scottbommarito: After running both tests (ignoring the slower Azure Search region):

|        | With Messages | Without Messages |
|--------|---------------|------------------|
| Global | 35            | 24               |
| China  | 55            | 44               |

In other words, with large messages, we are slower, but not slow enough to be a blocker IMO.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
joelverhagen: Oh nice.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
joelverhagen: We don't have those crystal clear "SLA" numbers for ingestion time so it's hard to say. But yeah, it's not even a 2x factor so I agree.

Issue #7159, opened 2019-05-15
Add feature flag for deprecations flowing into V3


scottbommarito:
In case bulk deprecations overflow the V3 pipeline, we should be able to temporarily turn them off until the pipeline can catch up.

Potential metrics: number of versions being deprecated, number of total versions of package being deprecated.

scottbommarito:
We may not need this--we can consider using the existing feature flag around the entire feature when feature is misbehaving in V3.

scottbommarito:
We have added a feature flag that does not allow users to deprecate packages with more than 500 versions when it is disabled.

scottbommarito:
can't verify prod cuz feat flag

PR #7285, opened 2019-06-20
Add feature flag for deprecating a package with many versions

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7159

Basically, if the feature goes haywire, we want to be able to turn it off for packages with many versions while not affecting people with few versions.

Right now I've set the cut-off at 500. I think this is sufficiently low given the performance I've seen on DEV--even doing a bulk update on a thousand versions doesn't seem to have a large effect.

Also, if the existing feature flag is enabled but the "many versions" flag is not configured, I have it returning true. This means that if the existing flag is enabled, we would need to explicitly disable it for many versions to turn it off for many versions. I think this is ideal because it will make it harder for us to not notice when it's turned on for few versions and turned off for many versions.

### DIFF
--- a/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
+++ b/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
@@ -81,12 +81,6 @@ namespace NuGetGallery
                 status |= PackageDeprecationStatus.Other;
             }
 
-            var currentUser = GetCurrentUser();
-            if (!_featureFlagService.IsManageDeprecationEnabled(GetCurrentUser()))
-            {
-                return DeprecateErrorResponse(HttpStatusCode.Forbidden, Strings.DeprecatePackage_Forbidden);
-            }
-
             if (versions == null || !versions.Any())
             {
                 return DeprecateErrorResponse(HttpStatusCode.BadRequest, Strings.DeprecatePackage_NoVersions);
@@ -100,6 +94,12 @@ namespace NuGetGallery
                 return DeprecateErrorResponse(
                     HttpStatusCode.NotFound,
                     string.Format(Strings.DeprecatePackage_MissingRegistration, id));
+            }
+
+            var currentUser = GetCurrentUser();
+            if (!_featureFlagService.IsManageDeprecationEnabled(GetCurrentUser(), registration))
+            {
+                return DeprecateErrorResponse(HttpStatusCode.Forbidden, Strings.DeprecatePackage_Forbidden);
             }
 
             if (ActionsRequiringPermissions.DeprecatePackage.CheckPermissionsOnBehalfOfAnyAccount(currentUser, registration) != PermissionsCheckResult.Allowed)
--- a/src/NuGetGallery/Services/FeatureFlagService.cs
+++ b/src/NuGetGallery/Services/FeatureFlagService.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
 using NuGet.Services.Entities;
 using NuGet.Services.FeatureFlags;
 using NuGetGallery.Features;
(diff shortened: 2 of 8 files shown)


src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -81,12 +81,6 @@ public virtual JsonResult GetAlternatePackageVersions(string id)
                 status |= PackageDeprecationStatus.Other;
             }
 
-            var currentUser = GetCurrentUser();
```
loic-sharma: What if the `FindPackagesById` query is too expensive? I would keep keep the previous flight on the entire feature.

src/NuGetGallery/Services/FeatureFlagService.cs
```diff
@@ -46,9 +48,20 @@ public bool IsPackagesAtomFeedEnabled()
             return _client.IsEnabled(PackagesAtomFeedFeatureName, defaultValue: false);
         }
 
-        public bool IsManageDeprecationEnabled(User user)
+        /// <summary>
+        /// The number of versions a package needs to have before it should be flighted using <see cref="ManageDeprecationForManyVersionsFeatureName"/> instead of <see cref="ManageDeprecationFeatureName"/>.
+        /// </summary>
+        private const int _manageDeprecationForManyVersionsThreshold = 500;
```
loic-sharma: If it turns out that packages with 300+ versions are problematic, we'd have to hotfix this code and create a new build. Should we make this an app setting?

src/NuGetGallery/Services/FeatureFlagService.cs
```diff
@@ -46,9 +48,20 @@ public bool IsPackagesAtomFeedEnabled()
...
+
+        public bool IsManageDeprecationEnabled(User user, PackageRegistration registration)
         {
-            return _client.IsEnabled(ManageDeprecationFeatureName, user, defaultValue: false);
+            if (!_client.IsEnabled(ManageDeprecationFeatureName, user, defaultValue: false))
+            {
+                return false;
+            }
+
+            return registration.Packages.Count() < _manageDeprecationForManyVersionsThreshold
```
joelverhagen: So if we turn off "many" then the package owner won't see the UI at all? This could be rather startling. Could we show a support link with a message saying "sorry, contact support" or something?

src/NuGetGallery/Services/FeatureFlagService.cs
scottbommarito: In my experiences, packages with even 1,000 versions flow without any problems quickly on DEV--I already divided that threshold in half for the limit I proposed in this PR. It's only at the many-thousand versions level that things really begin to slow down. I think if we start seeing issues on packages with this few versions, IMO we should turn the feature off entirely--or we should turn off deprecations flowing to V3.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
scottbommarito: I'd argue that the database load from these queries is going to be trivial for a couple reasons
- I can't see this feature being used more than 1 time per minute (at least on release)...compared to the several thousand times that `DisplayPackage` is called per minute at its peak, which makes the same query
- In order to hit this endpoint with the feature disabled, you would either have to be able to navigate to the page right before the feature was disabled or construct the request yourself, both of which seem very unlikely

I prefer the increased maintainability and testability of the single call.

src/NuGetGallery/Services/FeatureFlagService.cs
scottbommarito: The same thing will happen if we disable the feature for everyone. I'd argue that if we'd want to have a special message when the feature is disabled only for packages with many versions, we should consider having a message when the feature is disabled for all users as well. I was considering this to feature flag to be a stopgap for us to use in the unlikely case that the deprecation feature causes massive load on V3, so I didn't want to invest too much in changing the UI. I'm not too opposed to adding a message, but I thought one wasn't necessary.

joelverhagen: Okay, sounds good.

Issue #7287, opened 2019-06-20
[Azure Search] Emit search selection telemetry to AI


joelverhagen:
This will allow better integration and queryability.

PR #7288, opened 2019-06-20
Emit Application Insights metric from browser for search selection

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7287

I will keep it side-by-side with Google Analytics so we can understand if we are missing data. I think this is a better model overall and works great for click events performed within our site.

For outgoing links we would probably need to investigate the beacon API in AI to avoid events dropped during page transition.

Issue https://github.com/NuGet/NuGetGallery/issues/7289 tracks removing GA later.

### DIFF
--- a/src/NuGetGallery/Views/Shared/ListPackages.cshtml
+++ b/src/NuGetGallery/Views/Shared/ListPackages.cshtml
@@ -80,7 +80,12 @@
 
 @section bottomScripts {
     <script type="text/javascript">
+        // Used to track how long the user waited before clicking a search selection.
+        var pageLoadTime = Date.now();
 
+        // Used to track how many selections were made on this page. Multiple selections can happen if the user opens
+        // a search selection in a new tab, instead of navigating away from this page.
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -111,7 +122,46 @@
...
                 window.location.href = url;
-            })
+            });
+
+            var emitAiClickEvent = function () {
+                if (!window.nuget.isAiAvailable()) {
+                    return;
+                }
+
+                var href = $(this).attr('href');
```
scottbommarito: Maybe save `$(this)` into a variable (`$this`) so you don't have to keep calling it?

Also you don't really need to declare all of these variables. You could just save `href` and `category`, check them, and then just put the expressions in the `sendAiMetric` call directly.

joelverhagen: You're right that does look better

PR #7301, opened 2019-06-25
Limiting the id + version length to 160 characters.

agr:
User friendly mitigation for https://github.com/NuGet/Engineering/issues/2481

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
@@ -134,6 +134,11 @@ namespace NuGetGallery
                 throw new EntityException(ServicesStrings.NuGetPackagePropertyTooLong, "Language", "20");
             }
 
+            if (packageMetadata.Id.Length + (packageMetadata.Version?.ToFullString().Length ?? 0) > 160)
+            {
+                throw new EntityException(ServicesStrings.NuGetPackageIdVersionCombinedTooLong, "160");
+            }
+
             // Validate dependencies
             if (packageMetadata.GetDependencyGroups() != null)
             {
--- a/src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
@@ -77,7 +77,7 @@ namespace NuGetGallery
             // TODO: Change this to use DataAnnotations
             if (packageMetadata.Id.Length > NuGet.Services.Entities.Constants.MaxPackageIdLength)
             {
-                throw new EntityException(ServicesStrings.NuGetPackagePropertyTooLong, "Id", NuGet.Services.Entities.Constants.MaxPackageIdLength);
+                throw new EntityException(ServicesStrings.NuGetPackagePropertyTooLong, "ID", NuGet.Services.Entities.Constants.MaxPackageIdLength);
             }
             if (packageMetadata.Authors != null && packageMetadata.Authors.Flatten().Length > 4000)
             {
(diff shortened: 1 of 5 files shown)


src/NuGetGallery.Services/ServicesStrings.resx
```diff
@@ -1118,4 +1118,7 @@ The {1} Team</value>
   <data name="UploadPackage_EmbeddedIconNotAccepted" xml:space="preserve">
     <value>The &lt;icon&gt; element is not currently supported.</value>
   </data>
+  <data name="NuGetPackageIdVersionCombinedTooLong" xml:space="preserve">
+    <value>A nuget package's id and version properties combined may not be more than {0} characters long.</value>
```
joelverhagen: `ID` not `id`

> **define id**
> the part of the mind in which innate instinctive impulses and primary processes are manifest.

src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
```diff
@@ -134,6 +134,11 @@ public static void ValidateNuGetPackageMetadata(PackageMetadata packageMetadata)
                 throw new EntityException(ServicesStrings.NuGetPackagePropertyTooLong, "Language", "20");
             }
 
+            if (packageMetadata.Id.Length + (packageMetadata.Version?.ToFullString().Length ?? 0) > 160)
```
joelverhagen: So with "." and the ".nupkg" are subtracted from the actual limit to get 160? So that actual limit is 167?

src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
```diff
@@ -134,6 +134,11 @@ public static void ValidateNuGetPackageMetadata(PackageMetadata packageMetadata)
...
+            {
+                throw new EntityException(ServicesStrings.NuGetPackageIdVersionCombinedTooLong, "160");
```
joelverhagen: the number should be a const so these two are the same value

agr: I followed the pattern above...

src/NuGetGallery.Services/ServicesStrings.resx
agr: Do you think I should fix it here as well:
https://github.com/NuGet/NuGetGallery/blob/1b7f3da0f88d0a5a229026db1e4f1a2d5e05e680/src/NuGetGallery.Services/PackageManagement/PackageHelper.cs#L80
?

joelverhagen: If it's outside of your changed lines I wouldn't worry about it.

src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
joelverhagen: it's so bad though. follow your heart

src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
agr: The actual limit is 175, but, apparently, it has a history of fluctuating. I opted for a smaller round number.

src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
agr: I'll make a work item to have it fixed. I started changing all, but now I'm afraid I might have messed up the values, so we need to write tests to cover those and with tests it starts to get too much time to fix.

PR #7307, opened 2019-06-28
Update to Application Insights 2.3.0

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/2527.
Depends on https://github.com/NuGet/ServerCommon/pull/302.

This allows us to avoid https://github.com/microsoft/ApplicationInsights-dotnet/issues/480.

PR #7309, opened 2019-07-01
[Hotfix] Update to Application Insights 2.3.0

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/2527.

This is a version of https://github.com/NuGet/NuGetGallery/pull/7307. I think it is more prudent to deploy this as a hotfix. Deploying this change along with product changes bit us last time. It would be foolish to not be cautious given that experience.

Additionally, this will allow us to get this change out sooner.

### DIFF
--- a/src/NuGetGallery.Services/NuGetGallery.Services.csproj
+++ b/src/NuGetGallery.Services/NuGetGallery.Services.csproj
@@ -170,13 +170,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.ApplicationInsights">
-      <Version>2.2.0</Version>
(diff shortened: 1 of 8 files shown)


src/NuGetGallery.Services/NuGetGallery.Services.csproj
```diff
@@ -170,13 +170,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.ApplicationInsights">
-      <Version>2.2.0</Version>
-    </PackageReference>
-    <PackageReference Include="Microsoft.AspNet.Mvc">
```
loic-sharma: Nice cleanup!

PR #7310, opened 2019-07-01
[A/B Test 1] Add A/B test content file, flight, and telemetry

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7166

1. Add a config type to content object service to define test percentage
1. Add a flight to turn the whole thing off if necessary
1. Add telemetry when A/B tests are evaluated and when users are enrolled

More PRs to come.

PR #7311, opened 2019-07-01
[A/B Test 2] Add A/B test enrollment service and cookie

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7166
Depends on https://github.com/NuGet/NuGetGallery/pull/7310

This is the meat of the assignment logic for A/B testing.

### DIFF
--- a/src/NuGetGallery/Infrastructure/ABTestEnrollment.cs
+++ b/src/NuGetGallery/Infrastructure/ABTestEnrollment.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// This class stores a specified user's configuration about what variants of A/B tests they are participating in.
+    /// Aside from the <see cref="State"/> and <see cref="SchemaVersion"/> properties, the other properties are about
+    /// which bucket number the user falls into for different A/B tests.
+    /// 
+    /// For each A/B test, an integer in the range [0, 99] is stored. This is a bucket number. An A/B test is
+    /// configured to include a specific percentage of these buckets. For example, suppose we want to point 50% of
+    /// users to a new preview search experience. If a "PreviewSearch" property has a value of less than 50, that means
+    /// this user along with all other bucket values in range [0, 49] will get the new preview experience.
+    /// 
+    /// The reason a bucket value is stored instead of a simple boolean (i.e. UsesPreviewSearch = true | false) is to
+    /// make percentage increase transitions consistent for users. If we start with 2% of users on preview search and
+    /// then later want 10% of users, the original 2% should still see the preview search. Storing the bucket value
+    /// means that each user's placement on the range of percentage thresholds remains consistent.
+    /// </summary>
+    public class ABTestEnrollment
+    {
+        public ABTestEnrollment(ABTestEnrollmentState state, int schemaVersion, int previewSearchBucket)
+        {
+            State = state;
+            SchemaVersion = schemaVersion;
+            PreviewSearchBucket = previewSearchBucket;
+        }
+
+        public ABTestEnrollmentState State { get; }
+        public int SchemaVersion { get; }
+        public int PreviewSearchBucket { get; }
+    }
+}
--- a/src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
+++ b/src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
@@ -1,0 +1,121 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Cryptography;
+using System.Threading;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace NuGetGallery
+{
+    public class ABTestEnrollmentFactory : IABTestEnrollmentFactory
+    {
+        private const int SchemaVersion1 = 1;
+
+        private static readonly RNGCryptoServiceProvider _secureRng = new RNGCryptoServiceProvider();
+        private static readonly ThreadLocal<byte[]> _bytes = new ThreadLocal<byte[]>(() => new byte[sizeof(ulong)]);
+
+        private readonly ITelemetryService _telemetryService;
+        private readonly ILogger<ABTestEnrollmentFactory> _logger;
+
+        public ABTestEnrollmentFactory(
+            ITelemetryService telemetryService,
+            ILogger<ABTestEnrollmentFactory> logger)
+        {
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public ABTestEnrollment Initialize()
+        {
+            var enrollment = new ABTestEnrollment(
+                ABTestEnrollmentState.FirstHit,
+                SchemaVersion1,
+                previewSearchBucket: GetRandomWholePercentage());
+
+            _telemetryService.TrackABTestEnrollmentInitialized(
+                enrollment.SchemaVersion,
+                enrollment.PreviewSearchBucket);
+
+            return enrollment;
+        }
+
+        /// <summary>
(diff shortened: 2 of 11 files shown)


src/NuGetGallery/Infrastructure/CookieBasedABTestService.cs
```diff
@@ -0,0 +1,145 @@
...
+            ILogger<CookieBasedABTestService> logger)
+        {
+            _httpContext = httpContext ?? throw new ArgumentNullException(nameof(httpContext));
+            _featureFlagService = featureFlagService ?? throw new ArgumentNullException(nameof(featureFlagService));
+            _enrollmentFactory = enrollmentFactory ?? throw new ArgumentNullException(nameof(enrollmentFactory));
+            _contentObjectService = contentObjectService ?? throw new ArgumentNullException(nameof(contentObjectService));
+            _cookieComplianceService = cookieComplianceService ?? throw new ArgumentNullException(nameof(cookieComplianceService));
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _lazyEnrollment = new Lazy<ABTestEnrollment>(() => DetermineEnrollment());
```
loic-sharma: >new Lazy<ABTestEnrollment>(() => DetermineEnrollment()) [](start = 30, length = 55)

Does this work? `... new Lazy<ABTestEnrollment>(DetermineEnrollment);`?

src/NuGetGallery/Infrastructure/CookieBasedABTestService.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        public bool IsPreviewSearchEnabled(User user)
+        {
+            return IsActive(
+                nameof(Enrollment.PreviewSearchBucket),
+                user,
+                e => e.PreviewSearchBucket,
```
loic-sharma: >e => e.PreviewSearchBucket [](start = 16, length = 26)

Do we need to pass `Func`s down? Could we instead calculate `testPercentage` and then pass that value into `IsActive`?

Also, what do you think of expanding `e` to `enrollment` to clarify this code a little

src/NuGetGallery/Infrastructure/CookieBasedABTestService.cs
joelverhagen: Yes, nice catch.

src/NuGetGallery/Infrastructure/CookieBasedABTestService.cs
joelverhagen: I'll take a look at the code and see if it's better.

joelverhagen: Ah I chose not too since I didn't want to activate the `Enrollment` lazy if the cookie compliance service or feature flag say "always return false". I can expand the variable name though.

src/NuGetGallery/Infrastructure/ABTestEnrollment.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// 
+    /// For each A/B test, an integer in the range [0, 99] is stored. This is a bucket number. An A/B test is
+    /// configured to include a specific percentage of these buckets. For example, suppose we want to point 50% of
+    /// users to a new preview search experience. If a "PreviewSearch" property has a value of less than 50, that means
+    /// this user along with all other bucket values in range [0, 49] will get the new preview experience.
+    /// 
+    /// The reason a bucket value is stored instead of a simple boolean (i.e. UsesPreviewSearch = true | false) is to
+    /// make percentage increase transitions consistent for users. If we start with 2% of users on preview search and
+    /// then later want 10% of users, the original 2% should still see the preview search. Storing the bucket value
+    /// means that each user's placement on the range of percentage thresholds remains consistent.
```
shishirx34: smart :) me likey.

src/NuGetGallery/Infrastructure/ABTestEnrollmentState.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    public enum ABTestEnrollmentState
+    {
+        /// <summary>
+        /// The user was enrolled for A/B testing in the current request.
```
shishirx34: could you explain a bit more here? unless this is documented else where in the code.

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+            var enrollment = new ABTestEnrollment(
+                ABTestEnrollmentState.FirstHit,
+                SchemaVersion1,
+                previewSearchBucket: GetRandomWholePercentage());
+
+            _telemetryService.TrackABTestEnrollmentInitialized(
+                enrollment.SchemaVersion,
+                enrollment.PreviewSearchBucket);
+
+            _logger.LogInformation("Initialized A/B test enrollment: {Enrollment}", JsonConvert.SerializeObject(enrollment));
```
shishirx34: wouldn't this be too much noise in app insights? may be not if its executed only once.

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+                enrollment = new ABTestEnrollment(
+                    ABTestEnrollmentState.Active,
+                    v1.SchemaVersion,
+                    v1.PreviewSearchBucket);
+
+                return true;
+            }
+            catch (JsonException)
+            {
+                return false;
```
shishirx34: may be log deserialization failures?

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            return input < 1 || input > 100;
+        }
+
+        private class StateVersion1
+        {
+            [JsonProperty("v", Required = Required.Always)]
+            public int SchemaVersion { get; set; }
+
+            [JsonProperty("ps", Required = Required.Always)]
```
shishirx34: nit: "psb" for `PreviewSearchBucket`? not super important.

src/NuGetGallery/Infrastructure/ABTestEnrollmentState.cs
shishirx34: nevermind.. got it from the code.

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
joelverhagen: I don't think it will be too much noise, it is once per unique users (with the caveat of users that ignore cookies like `HttpClient`). However I will remove it since the metric right above covers this.

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
joelverhagen: Good idea.

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
joelverhagen: All the values beside schema version should be buckets, so I left the `b` out.

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
joelverhagen: Oh wait it is logged at a higher level already.

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
joelverhagen: Fixed.

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
joelverhagen: https://github.com/NuGet/NuGetGallery/blob/ebdea994e8fa5e7fc8d1b5d6cae3a6a39450b55f/src/NuGetGallery/Infrastructure/CookieBasedABTestService.cs#L71

PR #7313, opened 2019-07-01
[A/B Test 3] Integrate A/B test service with package search

joelverhagen:
Depends on https://github.com/NuGet/NuGetGallery/pull/7311
Address https://github.com/NuGet/NuGetGallery/issues/7166

This toggles the search client used for package search based on the user's cookie. This also toggles the telemetry emitted.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -90,6 +90,7 @@ namespace NuGetGallery
         private readonly IPackageUpdateService _packageUpdateService;
         private readonly IPackageFileService _packageFileService;
         private readonly ISearchService _searchService;
+        private readonly ISearchService _previewSearchService;
         private readonly IUploadFileService _uploadFileService;
         private readonly IUserService _userService;
         private readonly IEntitiesContext _entitiesContext;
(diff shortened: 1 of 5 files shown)


tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -317,6 +327,30 @@ private static Mock<IPackageUploadService> GetValidPackageUploadService(string p
...
+            {
+                var type = typeof(PackagesController);
+
+                var constructors = type.GetConstructors();
+                var constructor = Assert.Single(constructors);
+
+                var parameters = constructor.GetParameters();
+                var parameter = Assert.Single(
+                    parameters,
+                    p => p.Name == DefaultDependenciesModule.ParameterNames.PackagesController_PreviewSearchService);
```
loic-sharma: Nice 👍

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -120,6 +122,7 @@ public partial class PackagesController
             IUserService userService,
             IMessageService messageService,
             ISearchService searchService,
+            ISearchService previewSearchService,
```
shishirx34: may be add a comment to not accidentally modify this name since its keyed to the DI?

shishirx34: I see the test, nice.

joelverhagen: Yeah I will leave the comment out since there is a test.

PR #7316, opened 2019-07-02
Additional Statistics Telemetry

ryuyu:
Adding additional telemetry around statistics in gallery.
Adds metrics for the following:
Interceptor replaces a gallery values with a smaller values from downloads for package
Interceptor replaces a gallery values with a smaller values from downloads for package registration
Request to get stats for package fails
Request to get stats for package registration fails
Duration of downloads.v1.json refresh
Update of downloads.v1.json causes a decrease in download count.

Also rejects updates to internal download count in service if it would lower the download count of a package.

@cristinamanum

### DIFF
--- a/src/NuGetGallery/Services/CloudDownloadCountService.cs
+++ b/src/NuGetGallery/Services/CloudDownloadCountService.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net;
@@ -17,6 +18,17 @@ namespace NuGetGallery
 {
     public class CloudDownloadCountService : IDownloadCountService
     {
+        private const string TelemetrySourcePrefix = "CloudDownloadCountService.";
+
+        private const string AdditionalInfoDimensionName = "AdditionalInfo";
+        private const string PackageIdDimensionName = "PackageId";
+        private const string PackageVersionDimensionName = "PackageVersion";
+        private const string TelemetryOriginDimensionName = "Origin";
+
+        private const string DownloadCountDecreasedMetricName = TelemetrySourcePrefix + "DownloadCountDecreased";
+        private const string DownloadJsonRefreshDurationMetricName = TelemetrySourcePrefix + "DownloadJsonRefreshDuration";
+        private const string GetDownloadCountFailedMetricName = TelemetrySourcePrefix + "GetDownloadCountFailed";
+
         private const string StatsContainerName = "nuget-cdnstats";
         private const string DownloadCountBlobName = "downloads.v1.json";
         private const string TelemetryOriginForRefreshMethod = "CloudDownloadCountService.Refresh";
@@ -52,6 +64,8 @@ namespace NuGetGallery
                 return true;
             }
 
+            _telemetryClient.TrackMetric(GetDownloadCountFailedMetricName, 0, GetTelemetryDimensions(id, "", TelemetrySourcePrefix + "TryGetDownloadCountForPackageRegistration"));
+
             downloadCount = 0;
             return false;
         }
@@ -74,6 +88,8 @@ namespace NuGetGallery
                 return true;
             }
 
+            _telemetryClient.TrackMetric(GetDownloadCountFailedMetricName, 0, GetTelemetryDimensions(id, version, TelemetrySourcePrefix + "TryGetDownloadCountForPackage"));
+
             downloadCount = 0;
             return false;
         }
@@ -94,7 +110,14 @@ namespace NuGetGallery
             {
                 try
                 {
+                    var stopwatch = Stopwatch.StartNew();
                     RefreshCore();
+                    stopwatch.Stop();
+                    _telemetryClient.TrackMetric(DownloadJsonRefreshDurationMetricName, stopwatch.Elapsed.TotalSeconds, new Dictionary<string, string>
+                        {
+                            { TelemetryOriginDimensionName, TelemetryOriginForRefreshMethod }
+                        });
+
                 }
                 catch (WebException ex)
                 {
@@ -192,7 +215,14 @@ namespace NuGetGallery
                                                 var version = token[0].ToString();
                                                 var downloadCount = token[1].ToObject<int>();
 
-                                                versions.AddOrSet(version, downloadCount);
+                                                if (versions.ContainsKey(version) && downloadCount < versions[version])
+                                                {
+                                                    _telemetryClient.TrackMetric(DownloadCountDecreasedMetricName, downloadCount - versions[version], GetTelemetryDimensions(id, version));
+                                                }
+                                                else
+                                                {
+                                                    versions.AddOrSet(version, downloadCount);
+                                                }
                                             }
                                         }
                                     }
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/Services/CloudDownloadCountService.cs
```diff
@@ -74,6 +88,8 @@ public bool TryGetDownloadCountForPackage(string id, string version, out int dow
                 return true;
             }
 
+            _telemetryClient.TrackMetric(GetDownloadCountFailedMetricName, 0, GetTelemetryDimensions(id, version, TelemetrySourcePrefix + "TryGetDownloadCountForPackage"));
```
skofman1: follow the existing pattern and move those to TelemetryService

src/NuGetGallery/Services/CloudDownloadCountService.cs
```diff
@@ -52,6 +64,8 @@ public bool TryGetDownloadCountForPackageRegistration(string id, out int downloa
                 return true;
             }
 
+            _telemetryClient.TrackMetric(GetDownloadCountFailedMetricName, 0, GetTelemetryDimensions(id, "", TelemetrySourcePrefix + "TryGetDownloadCountForPackageRegistration"));
```
cristinamanum: Named args for the GetTelemetryDimensions. What is the empty value? Is it intended to be empty?

ryuyu: It was intended to be empty. I couldn't think of a great way to indicate a package registration without separating the name completely. I can do that if you think it would be cleaner.

src/NuGetGallery/Services/CloudDownloadCountService.cs
```diff
@@ -94,7 +110,14 @@ public void Refresh()
             {
                 try
                 {
+                    var stopwatch = Stopwatch.StartNew();
                     RefreshCore();
+                    stopwatch.Stop();
+                    _telemetryClient.TrackMetric(DownloadJsonRefreshDurationMetricName, stopwatch.Elapsed.TotalSeconds, new Dictionary<string, string>
```
skofman1: Milliseconds.. I doubt the refresh will take seconds.

src/NuGetGallery/Services/CloudDownloadCountService.cs
```diff
@@ -192,7 +215,14 @@ private void RefreshCore()
                                                 var version = token[0].ToString();
                                                 var downloadCount = token[1].ToObject<int>();
 
-                                                versions.AddOrSet(version, downloadCount);
+                                                if (versions.ContainsKey(version) && downloadCount < versions[version])
+                                                {
+                                                    _telemetryClient.TrackMetric(DownloadCountDecreasedMetricName, downloadCount - versions[version], GetTelemetryDimensions(id, version));
```
skofman1: >downloadCount - versions[version] [](start = 115, length = 33)

might be beneficial to get the numbers themselves as you can try to use them for further reverse engineering

src/NuGetGallery/Services/DownloadCountObjectMaterializedInterceptor.cs
```diff
@@ -36,6 +48,16 @@ protected void InterceptPackageMaterialized(Package package)
             int downloadCount;
             if (_downloadCountService.TryGetDownloadCountForPackage(package.PackageRegistration.Id, packageNormalizedVersion, out downloadCount))
             {
+                if (downloadCount < package.DownloadCount)
+                {
+                    _telemetryClient.TrackMetric(GalleryDownloadCountLargerThanDownloadJsonByMetricName, package.DownloadCount - downloadCount, new Dictionary<string, string>
+                        {
+                            { TelemetryOriginDimensionName, TelemetrySourcePrefix + "InterceptPackageMaterialized" },
+                            { PackageIdDimensionName, package.PackageRegistration.Id },
```
cristinamanum: Will it be useful to track also the two download counts? In case that someone reports a specific value seen at a moment of time?

src/NuGetGallery/Services/DownloadCountObjectMaterializedInterceptor.cs
```diff
@@ -50,6 +72,16 @@ protected void InterceptPackageRegistrationMaterialized(PackageRegistration pack
             int downloadCount;
             if (_downloadCountService.TryGetDownloadCountForPackageRegistration(packageRegistration.Id, out downloadCount))
             {
+                if (downloadCount < packageRegistration.DownloadCount)
+                {
+                    _telemetryClient.TrackMetric(GalleryDownloadCountLargerThanDownloadJsonByMetricName, packageRegistration.DownloadCount - downloadCount, new Dictionary<string, string>
+                        {
+                            { TelemetryOriginDimensionName,  TelemetrySourcePrefix + "InterceptPackageRegistrationMaterialized" },
+                            { PackageIdDimensionName, packageRegistration.Id },
```
cristinamanum: Same comment as above.

src/NuGetGallery/Services/DownloadCountObjectMaterializedInterceptor.cs
```diff
@@ -50,6 +72,16 @@ protected void InterceptPackageRegistrationMaterialized(PackageRegistration pack
             int downloadCount;
             if (_downloadCountService.TryGetDownloadCountForPackageRegistration(packageRegistration.Id, out downloadCount))
             {
+                if (downloadCount < packageRegistration.DownloadCount)
+                {
+                    _telemetryClient.TrackMetric(GalleryDownloadCountLargerThanDownloadJsonByMetricName, packageRegistration.DownloadCount - downloadCount, new Dictionary<string, string>
+                        {
```
cristinamanum: Maybe use a different name and not add the empty version below?

src/NuGetGallery/Services/CloudDownloadCountService.cs
ryuyu: I separated them with the telemetry service change.

cristinamanum: I thing the we can use new Dictionary ... and not have a method.  Also the value above should be 1 not 0 right?

src/NuGetGallery/Services/DownloadCountObjectMaterializedInterceptor.cs
ryuyu: Yup decided to split these.

src/NuGetGallery/Services/DownloadCountObjectMaterializedInterceptor.cs
ryuyu: Spoke offline. We decided to do the following:
Log Metric with Name with Value of 1 to treat as an event.
Log this metric with properties that include the old and new counts.

src/NuGetGallery/Services/CloudDownloadCountService.cs
ryuyu: Updated.

src/NuGetGallery/Services/CloudDownloadCountService.cs
ryuyu: We decided to do the following:
Log Metric with Name with Value of 1 to treat as an event.
Log this metric with properties that include the old and new counts.

src/NuGetGallery/Services/DownloadCountObjectMaterializedInterceptor.cs
```diff
@@ -2,18 +2,22 @@
...
     public class DownloadCountObjectMaterializedInterceptor
         : IObjectMaterializedInterceptor
     {
+        private readonly ITelemetryService _telemetryService;
         private readonly IDownloadCountService _downloadCountService;
 
-        public DownloadCountObjectMaterializedInterceptor(IDownloadCountService downloadCountService)
+        public DownloadCountObjectMaterializedInterceptor(IDownloadCountService downloadCountService, ITelemetryService telemetryService)
         {
             _downloadCountService = downloadCountService;
```
skofman1: add null checks

ryuyu: Done

Issue #7305, opened 2019-06-28
[Deprecation UI] blue frame around deprecation status 


skofman1:
When clicking on the expander the blue frame appears

![image](https://user-images.githubusercontent.com/16807822/60363196-b0fd1c00-9997-11e9-8d5a-0dfe27c710e0.png)

scottbommarito:
Only needed to add `outline: none`.

scottbommarito:
Can't verify on PROD due to feature flag.

PR #7326, opened 2019-07-09
Small Deprecation fixes

scottbommarito:
https://github.com/nuget/engineering/issues/2541
https://github.com/nuget/engineering/issues/2536
https://github.com/nuget/nugetgallery/issues/7305
https://github.com/nuget/engineering/issues/2542
https://github.com/nuget/engineering/issues/2540

The new error message for having a package the alternate of itself looks like
image

### DIFF
--- a/src/Bootstrap/dist/css/bootstrap-theme.css
+++ b/src/Bootstrap/dist/css/bootstrap-theme.css
@@ -882,6 +882,7 @@ img.reserved-indicator-icon {
   justify-content: space-between;
   vertical-align: middle;
   width: 100%;
+  outline: none;
 }
 .page-package-details .deprecation-container .deprecation-expander .deprecation-expander-container {
   display: -webkit-box;
--- a/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
+++ b/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
@@ -38,10 +38,10 @@ namespace NuGetGallery
         public virtual JsonResult GetAlternatePackageVersions(string id)
         {
             var versions = _packageService.FindPackagesById(id)
-                .Where(p => p.PackageStatusKey == PackageStatus.Available)
+                .Where(p => p.PackageStatusKey == PackageStatus.Available && p.Listed)
                 .Select(p => NuGetVersion.Parse(p.Version))
                 .OrderByDescending(v => v)
-                .Select(v => v.ToFullString())
+                .Select(v => v.ToNormalizedString())
                 .ToList();
 
             return Json(HttpStatusCode.OK, versions, JsonRequestBehavior.AllowGet);
@@ -162,6 +162,13 @@ namespace NuGetGallery
                 {
                     packagesToUpdate.Add(package);
                 }
+            }
+
+            if (alternatePackage != null && packagesToUpdate.Any(p => p == alternatePackage))
+            {
+                return DeprecateErrorResponse(
+                    HttpStatusCode.BadRequest,
+                    Strings.DeprecatePackage_AlternateOfSelf);
             }
 
             await _deprecationService.UpdateDeprecation(
--- a/src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
+++ b/src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
@@ -154,7 +154,7 @@ function ManageDeprecationViewModel(id, versionDeprecationStateDictionary, defau
             error: function () {
                 if (self.alternatePackageId() === id) {
                     self.alternatePackageVersionsCached.removeAll();
-                    self.chosenAlternatePackageIdServerError("An unknown occurred when searching for alternate package '" + id + "'.");
+                    self.chosenAlternatePackageIdServerError("An unknown error occurred when searching for alternate package '" + id + "'.");
                 }
             }
         });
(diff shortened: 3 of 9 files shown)


src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
```diff
@@ -92,13 +106,26 @@ function ManageDeprecationViewModel(id, versionDeprecationStateDictionary, defau
...
+        return null;
+    }, this);
 
     // When a new alternate package ID is entered, load the list of versions from the server.
+    this.chosenAlternatePackageIdServerError = ko.observable();
     this.chosenAlternatePackageId.subscribe(function (id) {
         if (!id) {
             // If the user hasn't input an ID, don't query the server.
-            self.chosenAlternatePackageIdError(null);
+            self.alternatePackageVersionsCached.removeAll();
```
scottbommarito: This line is fix for https://github.com/nuget/engineering/issues/2541

src/Bootstrap/dist/css/bootstrap-theme.css
```diff
@@ -882,6 +882,7 @@ img.reserved-indicator-icon {
   justify-content: space-between;
   vertical-align: middle;
   width: 100%;
+  outline: none;
```
scottbommarito: This line (and `page-display-package.less`) is the fix for https://github.com/nuget/nugetgallery/issues/7305

src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
```diff
@@ -15,16 +15,16 @@
                             {
                                 if (isLegacy)
                                 {
-                                    @:This package has been deprecated because it's <b>legacy</b> and has <b>critical bugs</b>.
+                                    @:This package has been deprecated as it is <b>legacy</b> and has <b>critical bugs</b>.
```
scottbommarito: Changes to this file fix https://github.com/nuget/engineering/issues/2542

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -38,10 +38,10 @@ public partial class ManageDeprecationJsonApiController
         public virtual JsonResult GetAlternatePackageVersions(string id)
         {
             var versions = _packageService.FindPackagesById(id)
-                .Where(p => p.PackageStatusKey == PackageStatus.Available)
+                .Where(p => p.PackageStatusKey == PackageStatus.Available && p.Listed)
```
scottbommarito: This line fixes https://github.com/nuget/engineering/issues/2540

src/NuGetGallery/Strings.resx
```diff
@@ -1121,4 +1121,7 @@ The {1} Team</value>
   <data name="DeprecatePackage_CustomMessageTooLong" xml:space="preserve">
     <value>Your custom message is too long. It must be under {0} characters.</value>
   </data>
+  <data name="DeprecatePackage_AlternateOfSelf" xml:space="preserve">
+    <value>A package cannot be the alternate package of itself!</value>
```
joelverhagen: nit: I find the exclamation mark rather unprofessional.

scottbommarito: removed it.

anangaur:
Can’t I deprecate old versions of a package and propose newer versions of the same package as alternative?

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
```diff
@@ -92,13 +106,26 @@ function ManageDeprecationViewModel(id, versionDeprecationStateDictionary, defau
...
     // E.g. the package does not exist or cannot be chosen as an alternate.
-    this.chosenAlternatePackageIdError = ko.observable();
+    this.chosenAlternatePackageIdError = ko.pureComputed(function () {
+        var serverError = self.chosenAlternatePackageIdServerError();
+        if (serverError) {
+            return serverError;
+        }
+
+        if (self.isAlternateOfSelf()) {
+            return "A package cannot be the alternate package of itself. Please choose a different alternate package.";
```
xavierdecoster: This message may be a bit misleading, as one should be able to chose an alternate version of the same package registration.

How about this subtle difference, which suggests selecting a different version is supported as well?
```
A package cannot be the alternate package of itself. Please choose a different alternate package (version).
```

tests/NuGetGallery.Facts/Controllers/ManageDeprecationJsonApiControllerFacts.cs
```diff
@@ -603,6 +617,65 @@ public async Task ReturnsNotFoundIfSomeVersionMissing(User currentUser, User own
...
+                var controller = GetController<ManageDeprecationJsonApiController>();
+                controller.SetCurrentUser(currentUser);
+
+                // Act
+                var result = await controller.Deprecate(
+                    CreateDeprecatePackageRequest(
+                        id, 
+                        new[] { version }, 
+                        alternatePackageId: id, 
+                        alternatePackageVersion: version));
```
xavierdecoster: Do we have a test that asserts that same alternate package id but different version works as expected and doesn't error as being alternate of self?

scottbommarito:
@anangaur - Yes. This PR only blocks when one of the versions you are deprecated is the alternate package version you selected.

tests/NuGetGallery.Facts/Controllers/ManageDeprecationJsonApiControllerFacts.cs
scottbommarito: Nope, good call

src/NuGetGallery/Scripts/gallery/page-manage-deprecation.js
scottbommarito: To be honest, I think it's confusing either way--the difference between "packages", "package registrations", and "package versions" is confusing in general. I think it's fine as-is because after typing in the alternate package ID, it only shows the error after you select a specific version, which should suggest to users that it's the specific version, not the ID, that's the problem.

PR #7338, opened 2019-07-11
Move more services to Library And enable real deletes

ryuyu:
Moves more of the implementations of required services into NuGetGallery.Services library.
Updates AccountDeleter to consume these changes.

Merge changes and test fixes coming soon.

Please look at changes by commit to make this more managable. Thanks.


src/AccountDeleter/Telemetry/GalleryTelemetryService.cs
```diff
@@ -29,7 +29,7 @@ public void TrackABTestEvaluated(string name, bool isActive, bool isAuthenticate
 
         public void TrackAccountDeletionCompleted(User deletedUser, User deletedBy, bool success)
         {
-            throw new NotImplementedException();
+            //no-op for now
```
shishirx34: is there a task to add telemetry here? may be add the TODO link to task

ryuyu: I switched us over in the newest commit to using the real gallery telemetry service, and have deleted by dev stub.

src/AccountDeleter/EmptyIndexingService.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class EmptyIndexingService : IIndexingService
+    {
+        public EmptyIndexingService()
```
cristinamanum: Is this needed?

src/AccountDeleter/Evaluators/BaseUserEvaluator.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public abstract class BaseUserEvaluator : IUserEvaluator
+    {
+        protected readonly Guid _id;
```
cristinamanum: It does not need to be protected. Private would be fine right?

src/AccountDeleter/GalleryAccountManager.cs
```diff
@@ -56,6 +56,7 @@ public async Task<bool> DeleteAccount(User user)
             else
             {
                 _logger.LogError("Criteria passed but delete failed.");
+                _logger.LogError(result.Description);
```
cristinamanum: Do we record metrics for users deleted and users failed to be deleted?

src/AccountDeleter/Job.cs
```diff
@@ -57,9 +53,10 @@ protected override void ConfigureJobServices(IServiceCollection services, IConfi
             services.AddTransient<ISubscriptionProcessorTelemetryService, AccountDeleteTelemetryService>();
 
             services.AddScoped<AggregateEvaluator>();
-            services.AddTransient<AlwaysRejectEvaluator>();
-            services.AddTransient<AlwaysAllowEvaluator>();
-            services.AddTransient<UserPackageEvaluator>();
+            services.AddScoped<AlwaysRejectEvaluator>();
```
cristinamanum: Why Scoped ? Maybe add comments?

src/AccountDeleter/Job.cs
```diff
@@ -120,17 +114,56 @@ protected void ConfigureGalleryServices(IServiceCollection services, IConfigurat
             }
             else
             {
-                services.AddScoped<IDeleteAccountService, EmptyDeleteAccountService>();
-                //services.AddScoped<IDeleteAccountService, DeleteAccountService>();
+                services.AddScoped<IEntitiesContext>(sp =>
+                {
+                    var connectionFactory = sp.GetRequiredService<ISqlConnectionFactory<GalleryDbConfiguration>>();
+                    var connection = connectionFactory.CreateAsync().GetAwaiter().GetResult();
```
cristinamanum: Why not await connectionFactory.CreateAsync() ?

src/NuGetGallery.Services/ServicesConstants.cs
```diff
@@ -33,6 +35,17 @@ public static class ServicesConstants
         public const string EmailValidationRegex = "^" + EmailValidationRegexFirstPart + "@" + EmailValidationRegexSecondPart + "$";
         public static TimeSpan EmailValidationRegexTimeout = TimeSpan.FromSeconds(5);
 
+        /// <summary>
+        /// Parameters for calculating account lockout period after 
+        /// wrong password entry.
+        /// </summary>
+        public const double AccountLockoutMultiplierInMinutes = 10;
+        public const double AllowedLoginAttempts = 10;
```
cristinamanum: Is this new ? I do not see it removed from other file ?

src/AccountDeleter/Evaluators/BaseUserEvaluator.cs
ryuyu: Yes i think you are right. Private would be fine since we can use EvaluatorId in derived classes.

src/AccountDeleter/Job.cs
ryuyu: I copied this from the way we hydrate these connections in our other jobs. I don't think i can use await here as it would force the rest of everything here to become asynchronous, which i don't think we currently support.

src/AccountDeleter/EmptyIndexingService.cs
ryuyu: One of the guys that we are hydrating with DI requires an IIndexingService in order, but we don't use this in our current system since we index over V3 which directly reads DB to its updates. This is so the service that requests this can make the calls it wants, but we hijack them and no-op.

src/AccountDeleter/GalleryAccountManager.cs
ryuyu: Yes we track an event using TrackDeleteResult. It just happens where this gets called above in the MessageHandler.

src/NuGetGallery.Services/ServicesConstants.cs
ryuyu: These should have been removed from GalleryConstants. I have updated this. Good catch.

src/AccountDeleter/EmptyIndexingService.cs
cristinamanum: Maybe comments.

src/AccountDeleter/Configuration/GalleryConfiguration.cs
```diff
@@ -10,6 +10,16 @@ namespace NuGetGallery.AccountDeleter
 {
     public class GalleryConfiguration : IAppConfiguration
     {
+        public string SiteRoot
+        {
+            get
+            {
+                return "";
+            }
```
shishirx34: why return empty string?

ryuyu: This guy is called by someone during a constructor, but as of today, AccountDeleter doesn't care about this value. However, the guy that calls it expects to get a non-null result.

src/AccountDeleter/Evaluators/UserOrganizationEvaluator.cs
```diff
@@ -10,30 +10,19 @@ namespace NuGetGallery.AccountDeleter
...
-            get
-            {
-                return _id.ToString();
-            }
-        }
-
-        public bool CanUserBeDeleted(User user)
+        public override bool CanUserBeDeleted(User user)
         {
             throw new NotImplementedException();
```
shishirx34: what are the plans to use this evaluator? Is it needed?

src/AccountDeleter/GalleryAccountManager.cs
shishirx34: is the metric tracking inside the `DeleteAccountAsync` method?

src/AccountDeleter/Job.cs
shishirx34: Also, don't see the AccountOrganization evaluator, is that needed to be included anywhere? may be eventually?

src/AccountDeleter/Configuration/AccountDeleteConfiguration.cs
```diff
@@ -28,6 +28,11 @@ public class AccountDeleteConfiguration
         /// </summary>
         public Dictionary<string, string> TemplateReplacements { get; set; }
 
+        /// <summary>
+        /// Storage container connection string where gallery content can be found
+        /// </summary>
+        public string GalleryStorageConnectionString { get; set; }
```
skofman1: why is this needed?

ryuyu: Feature flags clean up requires access to the gallery blob where it is held.

src/AccountDeleter/EmptyIndexingService.cs
shishirx34: I see `PackageUpdateService` uses it. I know in gallery we used the `LuceneIndexingService` as the default, doesn't sound like it should be moved here. 👍 for this empty service.

src/AccountDeleter/Job.cs
```diff
@@ -120,17 +114,56 @@ protected void ConfigureGalleryServices(IServiceCollection services, IConfigurat
...
+                    var optionsSnapshot = options.Value;
+
+                    return new CloudBlobClientWrapper(optionsSnapshot.GalleryStorageConnectionString, readAccessGeoRedundant: true);
+                });
 
-                services.AddScoped<ITelemetryService, GalleryTelemetryService>();
+                services.AddScoped<ITelemetryService, TelemetryService>();
                 services.AddScoped<ISecurityPolicyService, SecurityPolicyService>();
-                services.AddScoped<IAuditingService>(sp => { return AuditingService.None; }); //Replace with real when we start doing deletes. For now, we are a readonly operation.
+                services.AddScoped<IAuditingService>(sp => { return AuditingService.None; });
```
shishirx34: I would leave the comment, doesn't look like we have auditing integrated yet?

shishirx34: unless auditing is being tracked with an issue for deletions.

src/AccountDeleter/Configuration/GalleryConfiguration.cs
agr: Maybe use expression body, like in the setter below? `get => "";`

src/AccountDeleter/Evaluators/UserOrganizationEvaluator.cs
ryuyu: Might not be needed. We will be discussing exactly what evaluations we want to run in the next few weeks. At that point, we will either complete this implementation, or remove to evaluator altogether.

src/AccountDeleter/Job.cs
ryuyu: According to @skofman1, we don't have this auditing in this path. If you know otherwise, let's file an issue and link here and we can move that too.

Issue #7119, opened 2019-05-01
[Azure Search] Test gallery circuit breaker against new search service


joelverhagen:
Both the region down case and the intermittent failure case.

loic-sharma:
Also, we should ensure we have good visibility on this through monitoring.

PR #7353, opened 2019-07-17
Consider a 403 from search in the circuit breaker configuration

joelverhagen:
Stopped Azure web apps return HTTP 403. We should consider this case for the circuit breaker configuration in light of moving search to web app (for Azure Search).

Note that we do not retry in 403 case as we do with HTTP 500+.

Progress on https://github.com/NuGet/NuGetGallery/issues/7119.

### DIFF
--- a/tests/NuGetGallery.Facts/Infrastructure/Lucene/SearchPolicyFacts.cs
+++ b/tests/NuGetGallery.Facts/Infrastructure/Lucene/SearchPolicyFacts.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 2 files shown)


tests/NuGetGallery.Facts/Infrastructure/Lucene/SearchPolicyFacts.cs
```diff
@@ -192,10 +241,17 @@ public async Task TestWaitAndRetryForInvalidRequests()
             Assert.Equal(HttpStatusCode.ServiceUnavailable, r.StatusCode);
         }
 
-        private static ILogger<ResilientSearchHttpClient> GetLogger()
+        private static TestHttpHandler MakeTestHttpHandler()
         {
-            var mockConfiguration = new Mock<ILogger<ResilientSearchHttpClient>>();
-            return mockConfiguration.Object;
+            var handler = new TestHttpHandler();
```
joelverhagen: Previously this test was going online. No longer.

PR #7356, opened 2019-07-18
Move search timeout from HttpClient to Polly timeout policy

joelverhagen:
When a timeout occurs, consider this as an error for the circuit breaker.
Progress on https://github.com/NuGet/NuGetGallery/issues/7119

Previously everyone would always encounter the timeout case on USNC and the circuit breaker would never open.

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -52,6 +52,7 @@ using NuGetGallery.Security;
 using SecretReaderFactory = NuGetGallery.Configuration.SecretReader.SecretReaderFactory;
 using Microsoft.Extensions.Http;
 using NuGetGallery.Infrastructure.Lucene;
+using System.Threading;
 
 namespace NuGetGallery
 {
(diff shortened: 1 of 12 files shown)


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -838,23 +839,29 @@ private static List<(string name, Uri searchUri)> GetPreviewSearchClientsFromCon
...
                         {
                             c.BaseAddress = searchClient.searchUri;
-                            c.Timeout = TimeSpan.FromMilliseconds(configuration.Current.SearchHttpClientTimeoutInMilliseconds);
+
+                            // Here we calculate a timeout for HttpClient that allows for all of the retries to occur.
+                            // This  is not strictly necessary today since the timeout exception is not a case that is
+                            // retried but it's best to set this right now instead of the default (100) or something
+                            // too small. The timeout on HttpClient is not really used. Instead, we depend on a timeout
+                            // policy from Polly.
+                            var perRetryMs = configuration.Current.SearchHttpClientTimeoutInMilliseconds;
```
cristinamanum: Maybe we should change the name from SearchHttpClientTimeoutInMilliseconds to SearchRequestTimeoutInMilliseconds or something similar ?

joelverhagen: Done.

PR #7364, opened 2019-07-22
Move business logic out of ViewModels related to upcoming embedded icon changes

agr:
Addresses https://github.com/NuGet/Engineering/issues/2599.

`ViewModelExtensions` class was added and all the initialization code from the constructors was moved there. All the properties of affected VMs were made read-write where possible, so controllers are now have freedom to set the to whatever they like without potentially complicated (in case of deep inheritance trees) changes to VM constructors.
`BasicUserViewModel` class was introduced to be used instead of passing `User` model into VMs, helpers were updated to accommodate to the change.


src/NuGetGallery/ViewModels/DeletePackageViewModel.cs
```diff
@@ -11,28 +11,15 @@ namespace NuGetGallery
...
-                        package.PackageRegistration.Id,
-                        package.Version)
-                },
-                ReasonChoices = reasons
-            };
-
-            IsLocked = package.PackageRegistration.IsLocked;
+            // TODO: remove
+            this.Setup(package, currentUser, reasons);
         }
```
loic-sharma: Should this be removed?

src/NuGetGallery/Helpers/ViewModelExtensions.cs
```diff
@@ -0,0 +1,609 @@
...
+using System.Linq;
+using System.Web.Mvc;
+using NuGet.Services.Entities;
+using NuGet.Services.Licenses;
+using NuGet.Versioning;
+using NuGetGallery.Security;
+
+namespace NuGetGallery
+{
+    public static class ViewModelExtensions
```
skofman1: this is a huge class. Consider spliting into a class per ViewMode / few related view models

src/NuGetGallery/ViewModels/PackageListViewModel.cs
```diff
@@ -24,7 +24,7 @@ public class PackageListViewModel
             bool isPreviewSearch)
         {
             // TODO: Implement actual sorting
```
skofman1: in some VMs you completely removed the ctor, but not here. Why?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -956,7 +956,7 @@ public virtual async Task<ActionResult> License(string id, string version)
                 throw;
             }
 
-            var model = new DisplayLicenseViewModel(package, licenseExpressionSegments, licenseFileContents);
+            var model = new DisplayLicenseViewModel().Setup(package, licenseExpressionSegments, licenseFileContents);
```
cristinamanum: Why not private ctor and Setup static on each ViewModel instead of extension?

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -13,98 +13,7 @@ namespace NuGetGallery
 {
     public class DisplayPackageViewModel : ListPackageItemViewModel
     {
-        public DisplayPackageViewModel(Package package, User currentUser, PackageDeprecation deprecation)
-            : this(package, currentUser, (string)null)
-        {
-            HasSemVer2Version = NuGetVersion.IsSemVer2;
-            HasSemVer2Dependency = package.Dependencies.ToList()
```
cristinamanum: Where was all this logic moved to?

src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs
```diff
@@ -84,45 +59,56 @@ public bool UseVersion
...
         public bool CanDeprecate { get; set; }
 
-        private static bool CanPerformAction(User currentUser, Package package, ActionRequiringPackagePermissions action)
+        public void SetShortDescriptionFrom(string fullDescription)
+        {
+            ShortDescription = fullDescription.TruncateAtWordBoundary(_descriptionLengthLimit, _omissionString, out var wasTruncated);
+            IsDescriptionTruncated = wasTruncated;
+        }
+
+        public void UpdateSignatureInformation(IReadOnlyCollection<string> signerUsernames, string sha1Thumbprint)
```
cristinamanum: Is this method used anywhere?

src/NuGetGallery/Controllers/PackagesController.cs
agr: A lot of code that was moved out contained genuine business logic. With services and models. That code simply does not belong to the ViewModels.

src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs
agr: https://github.com/NuGet/NuGetGallery/blob/3932e085983e4220c716fb5fa00fb75f9cdeea66/src/NuGetGallery/Helpers/ViewModelExtensions.cs#L117

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
agr: https://github.com/NuGet/NuGetGallery/blob/3932e085983e4220c716fb5fa00fb75f9cdeea66/src/NuGetGallery/Helpers/ViewModelExtensions.cs#L322-L325

src/NuGetGallery/ViewModels/PackageListViewModel.cs
agr: I was not trying to address _all_ VMs, only the ones related to icons. This one did not made it into my list, I am having a second thoughts now, though.

src/NuGetGallery/ViewModels/DeletePackageViewModel.cs
agr: Missed that.

src/NuGetGallery/ViewModels/PackageListViewModel.cs
agr: I modified the code a bit so there is no need to pass is models anymore, but I won't touch it any further.

src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs
```diff
@@ -84,45 +59,56 @@ public bool UseVersion
...
-            if (_package.Certificate == null)
+            if (_signerUsernames == null)
             {
                 return null;
             }
 
-            var owners = _package.PackageRegistration?.Owners ?? Enumerable.Empty<User>();
-            var signers = owners.Where(owner => owner.UserCertificates.Any(uc => uc.CertificateKey == _package.CertificateKey));
-            var signersCount = signers.Count();
+            var signersCount = _signerUsernames.Count();
```
joelverhagen: nit: you can use `Count` here

src/NuGetGallery/ViewModels/ListPackageItemViewModel.cs
```diff
@@ -15,40 +16,14 @@ public class ListPackageItemViewModel : PackageViewModel
...
-            CanDeleteSymbolsPackage = CanPerformAction(currentUser, package, ActionsRequiringPermissions.DeleteSymbolPackage);
-            CanDeprecate = CanPerformAction(currentUser, package, ActionsRequiringPermissions.DeprecatePackage);
-        }
+        private IReadOnlyCollection<string> _signerUsernames;
+        private string _sha1Thumbprint;
 
         public string Authors { get; set; }
-        public ICollection<User> Owners { get; set; }
-        public IEnumerable<string> Tags { get; set; }
+        public ICollection<BasicUserViewModel> Owners { get; set; }
```
joelverhagen: Maybe now's a good time to switch to `IReadOnlyCollection` where possible?

src/NuGetGallery/ViewModels/ListPackageItemRequiredSignerViewModel.cs
```diff
@@ -1,170 +1,29 @@
...
     {
-        // username must be an empty string because <select /> option values are based on username
-        // and this "user" must be distinguishable from an account named "Any" and any other user;
-        // null would be ideal, but null won't work as a <select /> option value.
-        private static readonly SignerViewModel AnySigner = 
-            new SignerViewModel(username: "", displayText: "Any");
-
         public SignerViewModel RequiredSigner { get; set; }
-        public string RequiredSignerMessage { get; set; }
+        public string RequiredSignerMessage { get; private set; }
```
joelverhagen: Is it really important for us to make this private unlike the rest of the properties?

agr: There is some presentation logic hidden in the `GetRequiredSignerMessage` method (although one might argue that this logic should belong to View, not ViewModel). I wanted to make it non-trivial to work around it. We can always drop that `private` when (and if) we need it.

joelverhagen: Sounds good.

agr: Should I turn it to a `GetRequiredSignerMessage` method instead?

src/NuGetGallery/Helpers/ViewModelExtensions/DeleteAccountListPackageItemViewModelExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public static class DeleteAccountListPackageItemViewModelExtensions
```
scottbommarito: I'm not really sure I see the need to move the initialization logic for these view models into extensions. I understand the desire to allow view models to be constructed independently without this complexity, a view model's initialization logic is still heavily tied to the view model itself.

If someone wants to see how a view model is intended to be constructed, they now need to find the associated extensions file. This makes the code less maintainable, and you may start seeing people initializing view models on their own because they didn't realize there was a `Setup` method hidden somewhere else.

Furthermore, the syntax is much nicer if it's a static method on the class:
```
var model = new ViewModel().Setup(...);
```
vs
```
var model = ViewModel.Create(...);
```

I understand that some of these constructors are rather huge and maybe should belong in their own file. In that case, could we make static methods in the class that call these extensions?

src/NuGetGallery/ViewModels/BasicUserViewModel.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    public class BasicUserViewModel
```
scottbommarito: I think you could just keep this as `UserViewModel`. I don't see any sort of complex hierarchy of view models about users so the `Basic` isn't very necessary.

src/NuGetGallery/ViewModels/BasicUserViewModel.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        public string EmailAddress { get; set; } // we might want to make it private and expose Gravatar URL instead to reduce chance to leak the email address to public
```
scottbommarito: I'd put it this comment in `<remarks>` on an XML doc for this property.

src/NuGetGallery/UrlHelperExtensions.cs
```diff
@@ -632,10 +632,29 @@ public static string GetOrganizationCertificates(this UrlHelper url, string acco
             return new RouteUrlTemplate<User>(linkGenerator, routesGenerator);
         }
 
+        public static RouteUrlTemplate<BasicUserViewModel> BasicUserTemplate(
```
scottbommarito: I think this method name should have `ViewModel` in it...otherwise it's fairly easy to confuse with the `UserTemplate` method that takes in actual `User` objects.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -931,9 +931,9 @@ foreach (var owner in Model.Owners)
             <li>
                 @if (!String.IsNullOrEmpty(owner.EmailAddress))
     {
-            <a href="@Url.User(owner)" title="@owner.Username">@ViewHelpers.GravatarImage(owner.EmailAddress, owner.Username, GalleryConstants.GravatarElementSize)</a>
+            <a href="@Url.User(owner.Username)" title="@owner.Username">@ViewHelpers.GravatarImage(owner.EmailAddress, owner.Username, GalleryConstants.GravatarElementSize)</a>
```
scottbommarito: Could you make a method on `UrlExtensions` that takes in a `UserViewModel` and calls the normal `User` method with its `Username`? Would remove the need for you to touch these files.

src/NuGetGallery/Helpers/ViewModelExtensions/DeleteAccountListPackageItemViewModelExtensions.cs
agr: Models and especially services should not cross the Controller-ViewModel boundary. The code that fills the contents of VMs belongs to controllers. Ideally, those should be private methods on controllers that do things specific to the callers. Unfortunately we reuse VMs quite a lot, so hiding that initialization in private controller methods is no good. So I put them into static methods, then made them extensions, so that initialization can be kept on a single line.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
agr: One only needs username to construct a link to a user profile page. Why would we want to pass any extra information to that helper? I would rather eliminate the overload that accepts `User` objects and use `@Url.User(owner.Username)` everywhere, but that's a bit out of scope.

src/NuGetGallery/ViewModels/BasicUserViewModel.cs
agr: We already have `UserViewModel` in admin pages. It's in its own namespace, but I am hesitant to introduce another class with the same name.

src/NuGetGallery/ViewModels/ListPackageItemRequiredSignerViewModel.cs
joelverhagen: No, I think this is fine.

src/NuGetGallery/ViewModels/BasicUserViewModel.cs
agr: I decided to just drop it.

PR #7365, opened 2019-07-22
Externalizing container information out of `CloudBlobCoreFileStorageService`

agr:
`CloudBlobCoreFileStorageService` had a hard-coded list of supported containers (mostly for the purpose of setting up proper public access if it needs to create said container), which limited its reuse in other projects.
That list was moved into its own class(`GalleryCloudBlobFolderInformationProvider : ICloudBlobFolderInformationProvider`), so there is no strict dependency anymore.

### DIFF
--- a/src/NuGetGallery.Core/Services/GalleryCloudBlobFolderInformationProvider.cs
+++ b/src/NuGetGallery.Core/Services/GalleryCloudBlobFolderInformationProvider.cs
@@ -29,6 +29,70 @@ namespace NuGetGallery
             CoreConstants.Folders.PackagesContentFolderName,
         };
 
+        public string GetCacheControl(string folderName)
+        {
+            switch (folderName)
+            {
+                case CoreConstants.Folders.PackagesFolderName:
+                case CoreConstants.Folders.SymbolPackagesFolderName:
+                case CoreConstants.Folders.ValidationFolderName:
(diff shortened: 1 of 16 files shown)


src/NuGetGallery.Core/Services/GalleryCloudBlobFolderInformationProvider.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace NuGetGallery
+{
+    public class GalleryCloudBlobFolderInformationProvider : ICloudBlobFolderInformationProvider
```
joelverhagen: Spoke offline. This should include content type too I think.

joelverhagen: Maybe the name "ContainerProvider" is more appropriate?

agr: Done

agr: Done.

PR #7368, opened 2019-07-23
Initial implementation async account deletion feature flagged.

ryuyu:
Adds support for async account deletes in gallery behind a feature flag.


src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -0,0 +1,141 @@
...
+                        || ex is MessagingEntityNotFoundException
+                        || ex is MessageNotFoundException
+                        || ex is MessageLockLostException
+                        || ex is SessionLockLostException
+                        || ex is MessagingEntityAlreadyExistsException
+                        || ex is RuleActionException
+                        || ex is FilterException
+                        || ex is TransactionSizeExceededException
+                        || ex is NoMatchingSubscriptionException
+                        || ex is MessageSizeExceededException)
```
agr: Please, consider moving that expression into its own function for readibility.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -0,0 +1,141 @@
...
+                Source = GalleryAccountDeleteMessageSourceName
+            };
+
+            var message = _serializer.Serialize(messageData);
+
+            var enqueueSuccess = false;
+            var canRetry = true;
+            var retryCount = 0;
+
+            while (!enqueueSuccess && retryCount < MAX_RETRY_COUNT && canRetry)
```
scottbommarito: I feel like this code is stuff that should be inherent to our service bus APIs in general--we should have some sort of `TopicClientWrapperWithRetry` in `ServerCommon` that we can put everywhere. I looked in a ton of places, and nowhere can I find any retrying. Could you add retries to our abstraction and keep this code simple?

For example, [Validation messages are enqueued like this](https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.Validation/PackageValidationEnqueuer.cs#L30), and [ValidateCertificate's enqueuer looks like this](https://github.com/NuGet/NuGet.Jobs/blob/master/src/Validation.PackageSigning.RevalidateCertificate/ValidateCertificateEnqueuer.cs#L30).

src/NuGetGallery.Services/ServicesStrings.resx
```diff
@@ -1121,4 +1121,13 @@ The {1} Team</value>
   <data name="NuGetPackageIdVersionCombinedTooLong" xml:space="preserve">
     <value>A nuget package's ID and version properties combined may not be more than {0} characters long.</value>
   </data>
+  <data name="AsyncAccountDelete_InsufficientPermissions" xml:space="preserve">
+    <value>{0} is not allowed to delete account '{1}'</value>
+  </data>
+  <data name="AsyncAccountDelete_NoRetryError" xml:space="preserve">
+    <value>There was an exception that caused retrys to be aborted. Please check logs for details.</value>
```
scottbommarito: `retrys` -> `retries`

src/NuGetGallery.Services/AccountManagement/SwitchingDeleteAccountService.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            _syncDeleteService = syncDeleteService ?? throw new ArgumentException(nameof(syncDeleteService));
+            _asyncDeleteService = asyncDeleteService ?? throw new ArgumentException(nameof(asyncDeleteService));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _featureFlagService = featureFlagService;
+        }
+
+        public async Task<DeleteAccountStatus> DeleteAccountAsync(User userToBeDeleted, User userToExecuteTheDelete, AccountDeletionOrphanPackagePolicy orphanPackagePolicy = AccountDeletionOrphanPackagePolicy.DoNotAllowOrphans)
+        {
+            if (_featureFlagService != null && _featureFlagService.IsAsyncAccountDeleteEnabled())
```
scottbommarito: Have you considered using DI to switch between these two?

I think [you can give Autofac a Func](https://autofaccn.readthedocs.io/en/latest/register/registration.html#selection-of-an-implementation-by-parameter-value) and call the feature flag service in it and then switch between implementations. It would get rid of the need to have this class.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -0,0 +1,141 @@
...
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Sends a message to a service bus queue when an account delete is requested.
+    /// A request is considered to be successful if the message is enqueued to the service bus successfully.
+    /// Retrys will be attempted until the limit is reached or an exception that cannot be retried is recieved
+    /// </summary>
+    public class AsynchronousDeleteAccountService : IDeleteAccountService
+    {
+        private const int MAX_RETRY_COUNT = 5;
```
joelverhagen: nit `MaxRetryCount` is more idiomatic

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
joelverhagen: I agree. The impact of a failure here means the caller will see a 500 maybe. Then they can just try again.

If we have data to support the need for a retry (say, `SendAsync` fails 1% of the time) then we should consider it but in a common place (NuGet.Services.ServiceBus) or something.

src/NuGetGallery.Services/AccountManagement/SwitchingDeleteAccountService.cs
```diff
@@ -0,0 +1,45 @@
...
+        public SwitchingDeleteAccountService(
+            IDeleteAccountService syncDeleteService, 
+            IDeleteAccountService asyncDeleteService, 
+            IFeatureFlagService featureFlagService, 
+            ILogger<SwitchingDeleteAccountService> logger)
+        {
+            _syncDeleteService = syncDeleteService ?? throw new ArgumentException(nameof(syncDeleteService));
+            _asyncDeleteService = asyncDeleteService ?? throw new ArgumentException(nameof(asyncDeleteService));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _featureFlagService = featureFlagService;
```
joelverhagen: Why would this be null?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -451,6 +452,56 @@ private void RegisterABTestServices(ContainerBuilder builder)
...
+                    (pi, ctx) => pi.ParameterType == typeof(ITopicClient),
+                    (pi, ctx) => ctx.ResolveKeyed<ITopicClient>(BindingKeys.AccountDeleterTopic)));
+
+            builder.RegisterType<DeleteAccountService>();
+
+            builder
+                .RegisterType<SwitchingDeleteAccountService>()
+                .WithParameters(new List<Parameter> {
+                    new ResolvedParameter(
+                        (pi, ctx) => pi.Name == "syncDeleteService",
```
joelverhagen: There should be a dumb unit test reflecting on `SwitchingDeleteAccountService` asserting these parameter names are expected. They can use a constant defined above.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -451,6 +452,56 @@ private void RegisterABTestServices(ContainerBuilder builder)
...
+                        (pi, ctx) => ctx.Resolve<DeleteAccountService>()
+                    ),
+                    new ResolvedParameter(
+                        (pi, ctx) => pi.Name == "asyncDeleteService",
```
joelverhagen: nit: check parameter type too `IDeleteAccountService`

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -0,0 +1,141 @@
...
+        private ILogger<AsynchronousDeleteAccountService> _logger;
+
+        public AsynchronousDeleteAccountService(ITopicClient topicClient, ILogger<AsynchronousDeleteAccountService> logger)
+        {
+            _topicClient = topicClient ?? throw new ArgumentNullException(nameof(topicClient));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _serializer = new BrokeredMessageSerializer<AccountDeleteMessageData>();
+        }
+
+        public async Task<DeleteAccountStatus> DeleteAccountAsync(User userToBeDeleted, User userToExecuteTheDelete, AccountDeletionOrphanPackagePolicy orphanPackagePolicy = AccountDeletionOrphanPackagePolicy.DoNotAllowOrphans)
```
skofman1: this is not used in the code

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        /// <summary>
+        /// Determine if the user requesting the delete is allowed to request a delete of the account.
+        /// This passes if the user requests their own delete, or if the user requesting is an admin.
+        /// </summary>
+        /// <param name="userToBeDeleted"></param>
+        /// <param name="userRequestingDelete"></param>
+        /// <returns></returns>
+        private bool UserIsAllowedToDelete(User userToBeDeleted, User userRequestingDelete)
```
skofman1: why is this check required? The only place I found that triggers delete is [https://github.com/NuGet/NuGetGallery/blob/4ebf904ed765e1db64eebf640c61a5c70353d4f1/src/NuGetGallery/Controllers/AccountsController.cs#L342](https://github.com/NuGet/NuGetGallery/blob/4ebf904ed765e1db64eebf640c61a5c70353d4f1/src/NuGetGallery/Controllers/AccountsController.cs#L342)

which authenticates the user and verifies he is an admin. This seems redundant.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
cristinamanum: Maybe it is worth using the list of exceptions that can be retried instead of the ones that cannot.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+
+            return false;
+        }
+
+        [Schema(Name = AccountDeleteMessageSchemaName, Version = 1)]
+        private struct AccountDeleteMessageData
+        {
+            public string Username { get; set; }
+            public string Source { get; set; }
```
cristinamanum: Please add comments for Source

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -451,6 +452,56 @@ private void RegisterABTestServices(ContainerBuilder builder)
                 .As<IABTestService>();
         }
 
+        private static void RegisterDeleteAccountService(ContainerBuilder builder, ConfigurationService configuration)
+        {
+            if (configuration.Current.AsynchronousDeleteAccountServiceEnabled)
```
cristinamanum: Why the feature flag is not enough? Register the SwitchingDeleteAccountService and the feature flag should redirect to one or the other.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -0,0 +1,141 @@
...
+using NuGetGallery.Areas.Admin.ViewModels;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Sends a message to a service bus queue when an account delete is requested.
+    /// A request is considered to be successful if the message is enqueued to the service bus successfully.
+    /// Retrys will be attempted until the limit is reached or an exception that cannot be retried is recieved
+    /// </summary>
+    public class AsynchronousDeleteAccountService : IDeleteAccountService
```
cristinamanum: I did not see any unit tests.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
ryuyu: I'll do what scott said here. The list of retryable ones is just as long.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// <summary>
+        /// Determine if the user requesting the delete is allowed to request a delete of the account.
+        /// This passes if the user requests their own delete, or if the user requesting is an admin.
+        /// </summary>
+        /// <param name="userToBeDeleted"></param>
+        /// <param name="userRequestingDelete"></param>
+        /// <returns></returns>
+        private bool UserIsAllowedToDelete(User userToBeDeleted, User userRequestingDelete)
+        {
+            if (userRequestingDelete.MatchesUser(userToBeDeleted) || userRequestingDelete.IsAdministrator)
```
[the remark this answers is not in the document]
ryuyu: Sure we can make it simpler. The idea here was to separate whatever we use to determine permissions here so it's easier to change it later if necessary. That being said, there is no reason for it not to directly return the comparison in the if statement. Done.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
ryuyu: doen

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
ryuyu: We can drop it here. It just seemed odd to me for us to pass through a userToExecuteDelete and NOT expect DeleteAccountAsync to check it. I'm happy to drop it if we are making the call here to assume all permissions are correct.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
ryuyu: Removed this class entirely in favor of doing the switching at DI time.

src/NuGetGallery.Services/AccountManagement/SwitchingDeleteAccountService.cs
ryuyu: Removed class completely. Was considering the case where someone might not hydrate a feature flag service for whatever reason.

src/NuGetGallery.Services/AccountManagement/SwitchingDeleteAccountService.cs
ryuyu: Swapped to doing this.

src/NuGetGallery.Services/ServicesStrings.resx
ryuyu: Fixed.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
ryuyu: This is defined on the interface. we just dont' use it here.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
ryuyu: Removed it after verifying usage pattern in the DeleteAccountService.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
ryuyu: This is similar to the pattern we followed for Validation where there are 2 switches. 1 to enable the usage of the feature at all, and one to be able to hot swap (feature flag). This allows consumers that don't have the async pipeline to not have to define the configurations that only apply to that.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
ryuyu: Makes sense to me. I'll dump retires on this for now then.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
ryuyu: Done

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
cristinamanum: From the link mentioned in the source code (https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messaging-exceptions) there are 7 exceptions that may be worth retrying and 17 do not need to be retried.
I think that is better to be explicit with what we retry. If an exception is thrown and not captured in the non retry exception list the retry will happen without any reason. 

Some exceptions(MessagingCommunicationException) mention the retry after a certain interval of time.

ryuyu: We talked about the retrying in this thread (https://github.com/NuGet/NuGetGallery/pull/7368#discussion_r306568589) and have since decided to pull the retry logic altogether from this for now.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -0,0 +1,86 @@
...
+                _logger.LogError(0, ex, "Failed to enqueue to AccountDeleter.");
+                result.Success = false;
+                result.Description = ServicesStrings.AsyncAccountDelete_Fail;
+            }
+
+            return result;
+        }
+
+        [Schema(Name = AccountDeleteMessageSchemaName, Version = 1)]
+        private struct AccountDeleteMessageData
```
skofman1: why is this private? Shouldn't it be a shared struct between client and server?

ryuyu: hmm. this is a good point. I used private here because the other job that used this pattern used a private struct for this, but i see no reason that it couldn't be completely shared. Let me try to change to that.

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -0,0 +1,86 @@
...
+{
+    /// <summary>
+    /// Sends a message to a service bus queue when an account delete is requested.
+    /// A request is considered to be successful if the message is enqueued to the service bus successfully.
+    /// Retrys will be attempted until the limit is reached or an exception that cannot be retried is recieved
+    /// </summary>
+    public class AsynchronousDeleteAccountService : IDeleteAccountService
+    {
+        private const string AccountDeleteMessageSchemaName = "AccountDeleteMessageData";
+        private const string GalleryAccountDeleteMessageSourceName = "Gallery";
```
skofman1: we discussed that Gallery has actually 2 scenarios when a delete is sent: user initiated, admin initiated, with different behavior for each. One way of signaling the job about the different scenarios is the Source value. 
If this is the way you plan to implement this, better set a good name here already, like GalleryAdmin/GalleryUser

ryuyu: Yes this was before we made this decision. I'll make the updates here to support the naming, but the actual changes to enable self service delete will be in a future PR.

PR #7369, opened 2019-07-23
Add a simple NuGetDeleteEvaluator

ryuyu:
Add a new simple evaluator for us to use to put/test policies.

this will wait to go out until we have the Unconfirmed Deletes flowing through first.


src/AccountDeleter/Evaluators/NuGetDeleteEvaluator.cs
```diff
@@ -0,0 +1,54 @@
...
+using NuGet.Services.Entities;
+using System;
+
+namespace NuGetGallery.AccountDeleter
+{
+    /// <summary>
+    /// Simple single (non aggregate) evaluator for easy use
+    /// Criteria is user must own zero (0) packages, and be admin on zero (0) organizations. OR User is unconfirmed
+    /// </summary>
+    public class NuGetDeleteEvaluator : BaseUserEvaluator
```
skofman1: UTs?

scottbommarito: So, this class has merged together the `UserOrganizationEvaluator` and the `UserPackageEvaluator`, but we haven't gotten rid of the existing classes? We shouldn't have duplicate code. It'd be very easy for this class and the existing classes to become out of sync.

I would suggest either
- deleting those classes, because they are unused
- deleting this class and injecting in the existing classes into the aggregate evaluator in DI
- Moving the logic shared between this classes and other classes into a shared place.

To be honest, I prefer the second, because if you already have an aggregate evaluator you might as well keep the checks split up.

src/AccountDeleter/Evaluators/NuGetDeleteEvaluator.cs
```diff
@@ -0,0 +1,54 @@
...
+                _logger.LogWarning("{Evaluator}:{EvaluatorId} User cannot be deleted because they owned packages.", nameof(NuGetDeleteEvaluator), EvaluatorId);
+                deleteUser = false;
+            }
+
+            var userOrgMemberships = user.Organizations;
+            var userIsAdminOnOrgs = userOrgMemberships.AnySafe(m => m.IsAdmin);
+
+            if (userIsAdminOnOrgs)
+            {
+                _logger.LogWarning("{Evaluator}:{EvaluatorId} User cannot be deleted because they are administrator on an organization.", nameof(UserOrganizationEvaluator), EvaluatorId);
```
scottbommarito: This log should say `NuGetDeleteEvaluator` instead of `UserOrganizationEvaluator`. Also `on an organization` should be `of an organization`

Have you considered changing how these errors are bubbled up? Repeating `{Evaluator}:{EvaluatorId}` everywhere seems redundant and error-prone. You could consider having the child evaluators throw exceptions for the aggregate evaluator to process and wrap the message in.

```
if (userIsAdminOnOrgs)
{
    throw new AccountCannotBeDeletedException("User is an administrator of an organizations!");
}
```

```
try
{
    evaluator.CanUserBeDeleted(user);
}
catch (AccountCannotBeDeletedException e)
{
    _logger.LogInformation("{Evaluator}:{EvaluatorId} {EvaluatorMessage}", evaluator.GetType(), evaluator.EvaluatorId, e.Message};
}
```

src/AccountDeleter/Evaluators/NuGetDeleteEvaluator.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using Microsoft.Extensions.Logging;
+using NuGet.Services.Entities;
+using System;
+
+namespace NuGetGallery.AccountDeleter
+{
+    /// <summary>
+    /// Simple single (non aggregate) evaluator for easy use
+    /// Criteria is user must own zero (0) packages, and be admin on zero (0) organizations. OR User is unconfirmed
```
scottbommarito: `Criteria is user must own zero (0) packages and be admin on zero (0) organizations OR be unconfirmed.`

src/AccountDeleter/Evaluators/UserOrganizationEvaluator.cs
```diff
@@ -12,19 +12,28 @@ namespace NuGetGallery.AccountDeleter
...
         {
-            throw new NotImplementedException();
+            var userOrgMemberships = user.Organizations;
+
+            var userIsAdminOnOrgs = userOrgMemberships.AnySafe(m => m.IsAdmin);
+
+            if (userIsAdminOnOrgs)
+            {
+                _logger.LogWarning("{Evaluator}:{EvaluatorId} User cannot be deleted because they are administrator on an organization.", nameof(UserOrganizationEvaluator), EvaluatorId);
+                return false;
```
scottbommarito: No need to have multiple returns...just return `!userIsAdminOnOrgs`

src/AccountDeleter/Evaluators/NuGetDeleteEvaluator.cs
ryuyu: This is currently done this way so we can tune this without too many changes until we get what we need. I can remove the other classes for now if we think it will be cleaner for now, but this will likely stay as is (for now).

src/AccountDeleter/Evaluators/NuGetDeleteEvaluator.cs
ryuyu: Evaulator and EvaluatorId are used so that message from a single evaluator can be associated together. AggregateEvalutor (as of today) will stop as soon as a single evaluator evaluates false, so bubble and wrapping doesn't make a ton of sense unless we change to run everything all the time. Messaging should be the responsibility of the individual evaluators. We could consider having a method in BaseUserEvaluator taht will always append the evaluatorID to mitigate forgetting this.

and yes UserOrganizationEvalutor here should be NuGetDeleteEvaluator.

src/AccountDeleter/Evaluators/UserOrganizationEvaluator.cs
ryuyu: Done

src/AccountDeleter/Evaluators/NuGetDeleteEvaluator.cs
ryuyu: Added some UTs for this evaluator.

ryuyu: Removed some fo the unused evaluators.

tests/AccountDeleter.Facts/EvaluatorFacts.cs
```diff
@@ -27,134 +27,138 @@ public EvaluatorFacts(ITestOutputHelper output)
...
-            var result = aggregateEvalutor.CanUserBeDeleted(testUser);
+            var result = evaluator.CanUserBeDeleted(testUser);
 
             // Assert
             Assert.False(result);
         }
 
         [Fact]
-        public void AggregateEvaluatorAcceptsIfAllAccept()
+        public void AlwayAllowEvaulatorReturnsTrue()
```
scottbommarito: `Evaulator` -> `Evaluator`

This applies to most of the test names in this file.

src/AccountDeleter/Evaluators/NuGetDeleteEvaluator.cs
```diff
@@ -0,0 +1,54 @@
...
+            if (!user.Confirmed)
+            {
+                return deleteUser;
+            }
+
+            var userPackages = _packageService.FindPackagesByOwner(user, includeUnlisted: true);
+
+            if (userPackages.AnySafe())
+            {
+                _logger.LogWarning("{Evaluator}:{EvaluatorId} User cannot be deleted because they owned packages.", nameof(NuGetDeleteEvaluator), EvaluatorId);
```
scottbommarito: `owned` -> `own`

Technically, because they haven't been deleted yet, they still own packages, so present tense applies.

PR #7377, opened 2019-07-25
Add Auditing support to AccountDeleter

ryuyu:
Enables real auditing service to be hydrated in AccountDeleter.


src/AccountDeleter/Job.cs
```diff
@@ -235,5 +241,80 @@ protected override void ConfigureAutofacServices(ContainerBuilder containerBuild
                 .As<IEntityRepository<PackageOwnerRequest>>()
                 .InstancePerLifetimeScope();
         }
+
+        private void RegisterAuditingServices(IServiceCollection services)
```
scottbommarito: What are the differences between this and the existing Gallery DI for auditing? This looks very similar from my memory.

Could we combine the two? Put these functions in a shared place and then call it on both DIs.

src/AccountDeleter/RuntimeServiceProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition.Hosting;
+using System.IO;
+using System.Linq;
+
+namespace NuGetGallery
+{
+    internal sealed class RuntimeServiceProvider : IDisposable
```
scottbommarito: Was this code copied from [here](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/App_Start/RuntimeServiceProvider.cs)?

Please share the implementations between the two.

src/AccountDeleter/Job.cs
ryuyu: Gallery DI uses Autofac to set up auditing.
We are using Microsoft.Extensions DI to set this up.

src/AccountDeleter/RuntimeServiceProvider.cs
ryuyu: I would be reluctant to do this until I understand why this was made internal sealed to begin with. @dtivel Do you have any insight here? If there is no reason for this to be internal, and can instead be moved out to Core or Services where the other providers are, I'm more than happy to share implementations.

src/AccountDeleter/Job.cs
scottbommarito: Bummer.

Still, the only part of the 100 lines of code here that needs to be different are the `AddSingleton` calls.. Everything else should be able to be shared.

src/AccountDeleter/Job.cs
```diff
@@ -23,6 +23,9 @@
 using NuGetGallery.Areas.Admin.Models;
 using NuGet.Jobs;
 using NuGetGallery.Infrastructure.Authentication;
+using System.IO;
```
skofman1: sort usings

src/AccountDeleter/Job.cs
```diff
@@ -235,5 +241,80 @@ protected override void ConfigureAutofacServices(ContainerBuilder containerBuild
...
+        private AuditingService GetAuditingServiceForLocalFileSystem()
+        {
+            var auditingPath = Path.Combine(
+                AppDomain.CurrentDomain.BaseDirectory,
+                FileSystemAuditingService.DefaultContainerName);
+
+            return new FileSystemAuditingService(auditingPath, AuditActor.GetAspNetOnBehalfOfAsync);
+        }
+
+        private AuditingService GetAuditingServiceForAzureStorage(string auditStorageConnectionString, bool readAccessGeoRedundant = true)
```
skofman1: there is no need to support blob storage auditing. It's just confusing, and definitely shouldn't be used. I can see the value in local storage auditing for test purposes...

ryuyu: Ok i can remove this. I was trying to mirror gallery and it looked like gallery did one or the other here, but i'm happy to drop it if we know it doesn't.

src/AccountDeleter/RuntimeServiceProvider.cs
dtivel: Making a type public and/or subclassable should be a deliberate decision.  There was no need for the original type being public or subclassable, so I defaulted to `internal sealed`.

src/AccountDeleter/Job.cs
ryuyu: Sorted.

src/AccountDeleter/Job.cs
```diff
@@ -235,5 +241,67 @@ protected override void ConfigureAutofacServices(ContainerBuilder containerBuild
...
+            services.AddSingleton<IAuditingService>(sp =>
+            {
+                var addInAuditingServices = GetAddInServices<IAuditingService>();
+                var auditingServices = new List<IAuditingService>(addInAuditingServices);
+
+                try
+                {
+                    auditingServices.Add(sp.GetRequiredService<AuditingService>());
+                }
+                catch
```
dtivel: Swallowing all exceptions is generally discouraged.

What are the exceptions you're trying to handle?  Can you handle it more targetedly?  For example, when no default auditing service was registered, does `sp.GetRequiredService<AuditingService>()` throw or return `null`?  If the former, catch that specific exception.  If the latter, remove the blind `try...catch` and add a `null` check.

src/AccountDeleter/RuntimeServiceProvider.cs
ryuyu: Cool. Then I will move this into one of the shared gallery libraries, make it public, and share the implementation. Thanks!

dtivel: If you make the original public, be sure to add tests.

src/AccountDeleter/Job.cs
ryuyu: Updated this to catch InvalidOperationException as specified that this method throws by https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceproviderserviceextensions.getrequiredservice?view=aspnetcore-1.1

src/AccountDeleter/RuntimeServiceProvider.cs
ryuyu: There already appear to be tests for this at https://github.com/NuGet/NuGetGallery/blob/master/tests/NuGetGallery.Facts/App_Start/RuntimeServiceProviderTests.cs. I'll take a look and see if there is anything to add.

src/AccountDeleter/Job.cs
cristinamanum: Agree with @scottbommarito to share as much as possible.

PR #7387, opened 2019-07-29
Add page for managing site admins

scottbommarito:
Currently, we don't have a UI that allows us to manage site admins. Previously, we had the database editor, which we would use to do this, but we recently removed it. As a result, now, whenever we need to onboard someone to the team, we have to gain database access and make a SQL query to make them a site admin.

This PR adds a simple UI to the admin section of the site that allows us to easily and quickly manage site admins.

# Link to page in Admins index

image

# Site Admins page

image

### DIFF
--- a/src/NuGet.Services.Entities/Role.cs
+++ b/src/NuGet.Services.Entities/Role.cs
@@ -20,7 +20,7 @@ namespace NuGet.Services.Entities
 
         public bool Is(string roleName)
         {
-            return Name.Equals(roleName, StringComparison.OrdinalIgnoreCase);
+            return string.Equals(Name, roleName, StringComparison.OrdinalIgnoreCase);
         }
     }
 }
--- a/src/NuGet.Services.Entities/Role.cs
+++ b/src/NuGet.Services.Entities/Role.cs
@@ -18,9 +18,9 @@ namespace NuGet.Services.Entities
             Users = new HashSet<User>();
         }
 
-        public bool Is(string name)
+        public bool Is(string roleName)
         {
-            return Name.Equals(name, StringComparison.OrdinalIgnoreCase);
+            return Name.Equals(roleName, StringComparison.OrdinalIgnoreCase);
         }
     }
 }
--- a/src/NuGet.Services.Entities/Role.cs
+++ b/src/NuGet.Services.Entities/Role.cs
@@ -17,10 +17,5 @@ namespace NuGet.Services.Entities
         {
             Users = new HashSet<User>();
         }
-
-        public bool Is(string roleName)
-        {
-            return string.Equals(Name, roleName, StringComparison.OrdinalIgnoreCase);
-        }
     }
 }
--- a/src/NuGet.Services.Entities/Role.cs
+++ b/src/NuGet.Services.Entities/Role.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 
 namespace NuGet.Services.Entities
@@ -15,6 +16,11 @@ namespace NuGet.Services.Entities
         public Role()
         {
             Users = new HashSet<User>();
+        }
+
+        public bool Is(string name)
+        {
+            return Name.Equals(name, StringComparison.OrdinalIgnoreCase);
         }
     }
 }
--- a/src/NuGetGallery.Services/UserManagement/IUserService.cs
+++ b/src/NuGetGallery.Services/UserManagement/IUserService.cs
@@ -56,5 +56,9 @@ namespace NuGetGallery
         Task<bool> CancelTransformUserToOrganizationRequest(User accountToTransform, string token);
 
         Task<Organization> AddOrganizationAsync(string username, string emailAddress, User adminUser);
+
+        IReadOnlyList<User> GetSiteAdmins();
+
+        Task SetIsAdministrator(User user, bool isAdmin);
     }
 }
--- a/src/NuGetGallery.Services/UserManagement/UserService.cs
+++ b/src/NuGetGallery.Services/UserManagement/UserService.cs
@@ -25,6 +25,8 @@ namespace NuGetGallery
 
         public IEntityRepository<User> UserRepository { get; protected set; }
 
+        public IEntityRepository<Role> RoleRepository { get; protected set; }
+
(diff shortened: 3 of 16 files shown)


src/NuGet.Services.Entities/Role.cs
```diff
@@ -16,5 +17,10 @@ public Role()
         {
             Users = new HashSet<User>();
         }
+
+        public bool Is(string name)
```
agr: `name` -> `roleName`?

agr: This also could have been an extension.

scottbommarito: done

src/NuGet.Services.Entities/Role.cs
```diff
@@ -16,5 +17,10 @@ public Role()
...
+        {
+            return Name.Equals(name, StringComparison.OrdinalIgnoreCase);
```
agr: This will throw if `Name` is `null`. How about using `string.Equals(Name, name, StringComparison.OrdinalIgnoreCase)`?

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
```diff
@@ -147,7 +147,7 @@
         </li>
         <li>
             <h2>
-                <a href="@Url.Action(actionName: "Index", controllerName: "Features")">
+                <a href="@Url.Action(actionName: " Index", controllerName: "Features" )">
```
joelverhagen: Why the space before index? Won't this mean there's a little underline tail?

src/NuGetGallery.Services/UserManagement/IUserService.cs
```diff
@@ -56,5 +56,9 @@ public interface IUserService
         Task<bool> CancelTransformUserToOrganizationRequest(User accountToTransform, string token);
 
         Task<Organization> AddOrganizationAsync(string username, string emailAddress, User adminUser);
+
+        IReadOnlyList<User> GetSiteAdmins();
+
+        Task SetIsAdministrator(User user, bool isAdmin);
```
agr: `MakeAdministrator`?

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
scottbommarito: I think the awful CSHTML auto-format did this...will fix.

src/NuGetGallery/Areas/Admin/Views/SiteAdmins/Index.cshtml
```diff
@@ -0,0 +1,38 @@
...
+                <div class="col-xs-6">
+                    <a href="@Url.User(adminUsername)"
+                        title="View profile of @adminUsername">
+                        @adminUsername
+                    </a>
+                </div>
+                <div class="col-xs-6">
+                    @Html.AntiForgeryToken()
+                    @Html.Hidden("username", adminUsername)
+                    <a href="#" onclick="$(this).closest('form').submit()">
```
joelverhagen: Can't you use a submit input with class `btn-link`?

joelverhagen: There should be a confirm or you should make the UI much clearer which one you are removing, e.g. alternating colors of rows, button closer, etc.

src/NuGetGallery.Services/UserManagement/UserService.cs
```diff
@@ -662,5 +666,48 @@ public async Task<bool> CancelTransformUserToOrganizationRequest(User accountToT
...
+            }
+
+            return RoleRepository.CommitChangesAsync();
+        }
+
+        private Role GetAdminRole()
+        {
+            return RoleRepository.GetAll()
+                .Include(r => r.Users)
+                .ToList()
```
agr: This will pull the complete user list from the DB that belong to at least one Role, right? Does not seem very effective...

src/NuGetGallery.Services/UserManagement/UserService.cs
```diff
@@ -662,5 +666,48 @@ public async Task<bool> CancelTransformUserToOrganizationRequest(User accountToT
...
+                .Single(r => r.Is(Constants.AdminRoleName));
```
agr: This should happen before `.ToList()`, you won't be able to use the `Is` method then, though, but that would limit amount of data retrieved from DB.

src/NuGetGallery/Areas/Admin/Views/SiteAdmins/Index.cshtml
scottbommarito: I'll try putting it in a table. If that doesn't look good, I'll add a confirmation

src/NuGetGallery.Services/UserManagement/UserService.cs
scottbommarito: I wanted to put this before the `ToList` but it's unsupported by link, probably because it's a method on the class. I think if I put it as an extension and it will work.

src/NuGetGallery.Services/UserManagement/UserService.cs
scottbommarito: Initially, I had it like this because LINQ could not do the method on the class before the list is materialized, and I felt the risk was small of pulling all role members as we only have the site admin role now.

src/NuGetGallery/Areas/Admin/Views/SiteAdmins/Index.cshtml
scottbommarito: How's this?

image

scottbommarito: yup! done

src/NuGetGallery.Services/UserManagement/UserService.cs
scottbommarito: Still didn't work. I'll remove the `Include` to prevent loading large amounts of users.

src/NuGetGallery.Services/UserManagement/UserService.cs
scottbommarito: Still didn't work. I'll remove the `Include` to prevent loading large amounts of users.

src/NuGet.Services.Entities/Role.cs
scottbommarito: done

PR #7395, opened 2019-07-31
"Include prerelease" checkbox should be a link

scottbommarito:
Currently, it is a checkbox that refreshes the page upon checking/unchecking, which is very confusing for screen readers. Functionally, it behaves like a link, so we have changed it to a link.

Alternatively, we considered changing the form to ajax, but ultimately that is a very nontrivial change.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/910827
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/905602

# Hide prerelease

image

# Show prerelease

image


scottbommarito:
Got approval from UI team that this change is ok.

PR #7398, opened 2019-08-01
Icon related view-model changes. Take 2

agr:
Progress on https://github.com/NuGet/Engineering/issues/2599.

After trying to apply [previous](https://github.com/NuGet/NuGetGallery/pull/7364) changes to Icon branch, I realized that it would be better to have an actual class that produces view models. That would allow to pass services it uses as constructor arguments, simplifying actual `Create` (or `Setup`) calls.

One day, we can have it in DI (and we'd have to shuffle around the Unit tests as a result), but that would be quite a massive change by itself.


src/NuGetGallery/Helpers/ViewModelExtensions/DeleteAccountListPackageItemViewModelHelper.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public class DeleteAccountListPackageItemViewModelHelper
```
agr: This was the `DeletePackageViewModelExtensions.cs` file, git didn't picked up the rename for some reason.

src/NuGetGallery/Helpers/ViewModelExtensions/DeletePackageViewModelHelper.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Globalization;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public class DeletePackageViewModelHelper
```
agr: This was the `DeletePackageViewModelExtensions.cs` file, git didn't picked up the rename for some reason.

joelverhagen:
My only concern with this is that the view model helpers are perhaps a bit too granular. The number of dependencies on `PackagesController` will go up by, what, 6 when we DI them?

src/NuGetGallery/Controllers/AccountsController.cs
```diff
@@ -52,6 +52,8 @@ public class ViewMessages
 
         public IDeleteAccountService DeleteAccountService { get; }
 
+        private readonly DeleteAccountListPackageItemViewModelHelper _deleteAccountListPackageItemViewModelHelper;
```
loic-sharma: What're your thoughts on renaming `Helper` to `Builder`?

```suggestion
        private readonly DeleteAccountListPackageItemViewModelBuilder _deleteAccountListPackageItemViewModelBuilder;
```

scottbommarito: Have you considered `Factory` as well?

agr: I think [Builder](https://en.wikipedia.org/wiki/Builder_pattern) is an established pattern when one create a stateful builder object, throws bunch of items at it, then asks it to produce an instance of some target object. These guys are essentially stateless.

scottbommarito:
@joelverhagen - I think that's primarily a consequence of `PackagesController` (and most of our other controllers).

src/NuGetGallery/Controllers/AccountsController.cs
```diff
@@ -74,6 +76,8 @@ public class ViewMessages
             ContentObjectService = contentObjectService ?? throw new ArgumentNullException(nameof(contentObjectService));
             MessageServiceConfiguration = messageServiceConfiguration ?? throw new ArgumentNullException(nameof(messageServiceConfiguration));
             DeleteAccountService = deleteAccountService ?? throw new ArgumentNullException(nameof(deleteAccountService));
+
+            _deleteAccountListPackageItemViewModelHelper = new DeleteAccountListPackageItemViewModelHelper(PackageService);
```
scottbommarito: I'd be very cautious of doing non-DI things in these constructors due to how the unit tests are configured--you will likely need to make a large amount of changes to unit tests so that the helpers are set up properly. To mitigate this, I would make these fields `Lazy` and then set them up inline.

```
private readonly Lazy<DeleteAccountListPackageItemViewModelHelper> _deleteAccountListPackageItemViewModelHelper = new Lazy<DeleteAccountListPackageItemViewModelHelper>(() => new DeleteAccountListPackageItemViewModelHelper(PackageService));
```

src/NuGetGallery/Helpers/ViewModelExtensions/DeleteAccountListPackageItemViewModelHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+            return Setup(viewModel, package, userToDelete, currentUser);
+        }
+
+        private DeleteAccountListPackageItemViewModel Setup(DeleteAccountListPackageItemViewModel viewModel, Package package, User userToDelete, User currentUser)
+        {
+            _listPackageItemViewModelHelper.Setup(viewModel, package, currentUser);
+            return SetupInternal(viewModel, package, userToDelete);
+        }
+
+        private DeleteAccountListPackageItemViewModel SetupInternal(DeleteAccountListPackageItemViewModel viewModel, Package package, User userToDelete)
```
scottbommarito: Any reason why this can't be added to the `Setup` method? Is it ever called separately? This applies to all of the helper classes.

agr:
@joelverhagen cramming all of them into a single class will help mostly in `PackagesController`, but will cause oversubscription to services by that single class (it would have to depend on the union of all dependencies individual `Create` methods it would contain), which would make unit testing a bit painful. Right now it's not an issue, but eventually might become. At the very least I want to move the following code: https://github.com/NuGet/NuGetGallery/blob/b775690ecbfa1f8b523f45aa3996fad67c54b0c2/src/NuGetGallery/Controllers/PackagesController.cs#L766-L832

into `DisplayPackageViewModelHelper`, and that pulls bunch services needed only for that view-model setup.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -750,7 +761,7 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
             }
 
             var deprecation = _deprecationService.GetDeprecationByPackage(package);
-            var model = new DisplayPackageViewModel().Setup(package, currentUser, deprecation);
+            var model = _displayPackageViewModelHelper.Create(package, currentUser, deprecation, await _readMeService.GetReadMeHtmlAsync(package));
 
             model.ValidatingTooLong = _validationService.IsValidatingTooLong(package);
```
scottbommarito: Any reason why these properties aren't set up by the `DisplayPackageViewModelHelper`? Now they can pull in services.

src/NuGetGallery/Controllers/AccountsController.cs
agr: All the changes made so far change the way the code that sets up the view model is called. It is the same code that was in VM constructors before, just called in a different way. The services are not accessed unless `Create` method is called, so there is no change in behavior.

src/NuGetGallery/Controllers/PackagesController.cs
agr: There is a class that inherit from `DisplayPackageViewModel` and which don't have those set up (and that's one of the reasons why we should think really careful about inheritance). I don't want to deal with it now.

agr: That would be a functional change, which I didn't want to make.

src/NuGetGallery/Controllers/AccountsController.cs
scottbommarito: To clarify, I was referring to the constructor of the controllers, not the VM factory constructors. We do some strange and awful things with Moq in our unit tests that can cause test failures when the constructors of certain tested classes change, but, given that your unit tests have passed, the changes you made are not affected by this.

src/NuGetGallery/Helpers/ViewModelExtensions/DeleteAccountListPackageItemViewModelHelper.cs
agr: I wanted to make it really obvious that setup is multi-step process: we set up base portion of the class, then set up "own" portion with the `SetupInternal` method. Previously, a call to set up base portion was barely visible among dozens of other lines.

src/NuGetGallery/Controllers/AccountsController.cs
agr: Sold on "Factory"

src/NuGetGallery/Helpers/ViewModelExtensions/DeleteAccountListPackageItemViewModelHelper.cs
scottbommarito: I would argue that having this method didn't actually improve the visibility of the base call for me--it just made me confused what the difference between the two methods was. I think any number of other solutions could call attention to the importance of the single line without adding another method. I think even just adding a line between the base call and the rest of the code should be sufficient, but you could also add comments as well.

PR #7408, opened 2019-08-02
Some more accessibility issues found in last scan

scottbommarito:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/910474
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/917900
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/910502

### DIFF
--- a/src/NuGetGallery/App_Code/ViewHelpers.cshtml
+++ b/src/NuGetGallery/App_Code/ViewHelpers.cshtml
@@ -639,15 +639,10 @@ var hlp = new AccordionHelper(name, formModelStatePrefix, expanded, page);
         <script src='https://www.google.com/recaptcha/api.js' async defer></script>
         <script type="text/javascript">
             var $submit = $('@("#" + submitId)');
-            $submit.before('<div id="recaptcha"></div>');
-            var $recaptcha = $('#recaptcha');
-            $recaptcha.addClass("g-recaptcha");
-            $recaptcha.attr({
-                'data-sitekey': "@reCaptchaPublicKey",
-                'data-size': "invisible",
-                'data-callback': "onSubmit",
-                'data-tabindex': "-1"
-            });
+            $submit.before(
+                '<div id="recaptcha" class="g-recaptcha" '
+                + 'data-sitekey="@reCaptchaPublicKey" data-size="invisible" '
+                + 'data-callback="onSubmit" data-tabindex="-1"></div>');
 
             var $form = $submit.parents('form')
             $submit.click(function (e) {
--- a/src/NuGetGallery/App_Code/ViewHelpers.cshtml
+++ b/src/NuGetGallery/App_Code/ViewHelpers.cshtml
@@ -632,7 +632,7 @@ var hlp = new AccordionHelper(name, formModelStatePrefix, expanded, page);
     @BreadcrumbWithProfile(url, currentUser, showDivider, newSegments);
 }
 
-@helper RecatpchaScripts(string reCaptchaPublicKey, string submitId)
+@helper RecaptchaScripts(string reCaptchaPublicKey, string submitId)
 {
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -632,22 +632,33 @@ var hlp = new AccordionHelper(name, formModelStatePrefix, expanded, page);
...
 {
     if (!string.IsNullOrEmpty(reCaptchaPublicKey))
     {
         <script src='https://www.google.com/recaptcha/api.js' async defer></script>
         <script type="text/javascript">
             var $submit = $('@("#" + submitId)');
-            $submit.addClass("g-recaptcha");
-            $submit.attr({
+            $submit.before('<div id="recaptcha"></div>');
+            var $recaptcha = $('#recaptcha');
```
scottbommarito: These ReCaptcha related fixes are for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/917900

The gist is that ReCaptcha is currently being executed before form validation, which ruins the form validation for screen readers. By changing how we set up ReCaptcha, we now wait until after form validation, which allows screen readers to properly read the error messages.

src/NuGetGallery/ExtensionMethods.cs
```diff
@@ -202,7 +202,9 @@ public static HtmlString ShowCheckboxFor<TModel>(this HtmlHelper<TModel> html, E
                 htmlAttributes["class"] = "form-control";
             }
 
-            if (metadata.IsRequired)
+            // If the property is required, mark it as required unless it's a bool (which are usually checkboxes).
+            // Hearing that a checkbox is required is very confusing to screen readers.
+            if (metadata.IsRequired && metadata.ModelType != typeof(bool))
```
scottbommarito: Fix for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/910502

`aria-required` makes screen reader say "required", which generally makes users think that they need to leave the checkbox checked, which is not the point

src/NuGetGallery/Views/Packages/_ManageOwners.cshtml
```diff
@@ -45,7 +45,7 @@
                     <!-- /ko -->
                     <!-- ko ifnot: $parent.IsAllowedToRemove($data) -->
                     <div class="col-md-3 text-right remove-owner-disabled">
-                        <a class="icon-link"
+                        <a role="button" aria-disabled="true" class="icon-link"
```
scottbommarito: An additional improvement for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/910474, which was improved in the last PR but was still not announcing the link as a disabled button.

src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -632,22 +632,33 @@ var hlp = new AccordionHelper(name, formModelStatePrefix, expanded, page);
...
+            $recaptcha.addClass("g-recaptcha");
+            $recaptcha.attr({
```
loic-sharma: Could you instead all the class and attributes in the `.before` call?

src/NuGetGallery/ExtensionMethods.cs
loic-sharma: Would it be better to add an `isCheckbox` parameter to `GetHtmlAttributes` instead of checking the `ModelType`?

src/NuGetGallery/App_Code/ViewHelpers.cshtml
scottbommarito: done

src/NuGetGallery/ExtensionMethods.cs
scottbommarito: Yup. Made the change.

PR #7418, opened 2019-08-06
Unwrap error messages generated by jQuery Validation

scottbommarito:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/905677

When the jQuery Validation plugin finds an issue with an input field, it adds a `span` containing the error message to a `span` that has `role=alert`. As a result, Narrator reads `alert group <the error message>`. This is because Narrator thinks the `div` is a `group` containing multiple elements instead of just one. Furthermore, [the jQuery Validation plugin only supports a single error message for every field](https://github.com/jquery-validation/jquery-validation/issues/2146), so there will never be more than one error message anyway, so wrapping the error message in a container is invalid.

To fix this, I have added some post-processing logic that gets rid of the nested `span` and moves the error message to the parent span.

To demonstrate what is going on here...

Before post-processing:
```
<input name="a" aria-describedby="error" />
<span id="validation-container">
    <span id="error">Invalid input!</span>
</span>
```

After post-processing:
```
<input name="a" aria-describedby="validation-container" />
<span id="validation-container">Invalid input!</span>
```

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/common.js
+++ b/src/NuGetGallery/Scripts/gallery/common.js
@@ -35,6 +35,7 @@
 
         // Source: https://stackoverflow.com/questions/18754020/bootstrap-3-with-jquery-validation-plugin
         // Set the JQuery validation plugin's defaults to use classes recognized by Bootstrap.
+        var validatorErrorClass = 'help-block';
         $.validator.setDefaults({
             highlight: function (element) {
                 $(element).closest('.form-group').addClass('has-error');
@@ -43,7 +44,7 @@
                 $(element).closest('.form-group').removeClass('has-error');
             },
             errorElement: 'span',
-            errorClass: 'help-block',
+            errorClass: validatorErrorClass,
             errorPlacement: function (error, element) {
                 if (element.parent('.input-group').length) {
                     error.insertAfter(element.parent());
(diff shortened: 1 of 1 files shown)


joelverhagen:
Does anything depend on that "error" ID?

scottbommarito:
@joelverhagen - as far as I can tell, the only thing that depends on the error ID (`error` in the example) is the `aria-describedby` in the `input`, which I properly update to point to the ID of the validation container.

src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -59,30 +60,55 @@
...
+    function removeInvalidAriaDescribedBy(element, validatorErrorClass) {
         var describedBy = element.getAttribute("aria-describedby");
         if (!describedBy) {
             return;
         }
 
-        var ids = describedBy.split(" ")
+        var uniqueIds = [];
+        var ids = describedBy
+            .split(" ")
```
joelverhagen: what is there is multiple consecutive spaces or newlines in the attribute? It seems like jQuery should have a helper to do this since it's selector engine has the same needs.

src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -59,30 +60,55 @@
...
                 for (i = 0; this.successList[i]; i++) {
-                    removeInvalidAriaDescribedBy(this.successList[i]);
+                    removeInvalidAriaDescribedBy(this.successList[i], validatorErrorClass);
                 }
             }
         });
     }
 
-    function removeInvalidAriaDescribedBy(element) {
+    function removeInvalidAriaDescribedBy(element, validatorErrorClass) {
```
joelverhagen: This method does more than what the name describes now. Consider something about fixing up accessibility issues in general

src/NuGetGallery/Scripts/gallery/common.js
scottbommarito: I tried searching to see if there is a better way to do this, and, according to several StackOverflow posts, jQuery does not have any sort of helper method and this split call seems to be the best way to do it.

joelverhagen: What is there are consecutive spaces or other forms of whitespace?

scottbommarito: Other forms of whitespace are properly parsed. Consecutive spaces seems to result in empty entries. Will fix.

scottbommarito: Actually, it looks like the `if (!describedById)` check below should handle this case.

Issue #7415, opened 2019-08-05
Locked out, unable to convert nuget.org account to use Microsoft sign in


adamjcooper:
### Describe the bug

Our corporate nuget.org account username is "accusoft". This is currently a nuget.org account, not a Microsoft account.

We can sign in using our NuGet.org account, but after signing in we are shown the home page with a warning banner telling us we need to go to Account Settings and switch to using a Microsoft account to sign in. However, we can't actually open our account settings. In fact, nearly the entire UI of the home page is disabled (links don't work, cursor changes to indicate a clickable link or text field in strange places, etc.). 

After signing in via NuGet.org, we've tried manually navigating to https://www.nuget.org/account, but this simply redirects back to the home page.

We seem to be locked out of our account, without any way forward to convert to using a Microsoft account for sign in.

### To Reproduce

1. Navigate to https://www.nuget.org (using private/incognito mode to ensure no cookies are set)
2. Click the "sign in" link in the top right
3. Because we do not yet have a Microsoft account setup, click "Sign in using NuGet.org account"
4. Sign in with our username and password
5. Observe that we are signed in and see this at the top of the page when it loads:

![image](https://user-images.githubusercontent.com/443020/62481472-85ccd000-b780-11e9-929a-76ea2930b178.png)

6. Try to click the "Account Settings" link and observe it does not function. Try doing anything significant on this page and notice it does not function (open the profile menu in top right, click on links in top nav bar, etc.).
7. Manually navigate to https://www.nuget.org/account. Observe that the account page does not load. Instead, the home page is loaded. 

### Expected behavior
We should be able to sign in with our NuGet.org account and edit our account settings to switch over to using a Microsoft account.

### Screenshots
![image](https://user-images.githubusercontent.com/443020/62481598-d6442d80-b780-11e9-9a58-e22b17f27db0.png)

![image](https://user-images.githubusercontent.com/443020/62481626-ea882a80-b780-11e9-9ba1-23d03e5ba7e6.png)

![image](https://user-images.githubusercontent.com/443020/62481662-068bcc00-b781-11e9-9f61-24d38b0e52d7.png)

![image](https://user-images.githubusercontent.com/443020/62481717-3044f300-b781-11e9-8bb5-393e672fa8c8.png)

Additionally, manually navigating to https://www.nuget.org/account does not load the account details page. Instead, replies with an HTTP 302, redirecting to the (still non-functional) home page.

skofman1:
[...]

PR #7421, opened 2019-08-07
[Hotfix]Update modal dialog show binding.

ryuyu:
jQuery3.x has fully removed the $(document).on("ready") binding for binding events to page ready (see https://api.jquery.com/ready/ for details).

This caused the modal dialog "show" method to never actually get called.

This appears to be the only place we use the document.on('ready') pattern, so update it to the proper, non-deprecated jQuery3.x ready binding.

Thanks to @adamjcooper for catching and reporting this (https://github.com/NuGet/NuGetGallery/issues/7415)

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-home.js
+++ b/src/NuGetGallery/Scripts/gallery/page-home.js
@@ -22,11 +22,9 @@ $(function () {
     });
 
     function showModal() {
-        $(function (e) {
-            $("#popUpModal").modal({
-                show: true,
-                focus: true
-            });
+        $("#popUpModal").modal({
+            show: true,
+            focus: true
         });
     };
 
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Scripts/gallery/page-home.js
```diff
@@ -22,12 +22,12 @@ $(function () {
     });
 
     function showModal() {
-        $(document).on('ready', function (e) {
+        $(function (e) {
```
scottbommarito: The code in which `showModal` is called is already wrapped in a `$(function (e) { ... })` call. Is this additional call necessary?

ryuyu: That would be a behaviour change from the working thing before. I am trying to just fix the regression introduced by jQuery3.x change without touching behaviour for now.
You are correct in that it looks like it might be ok to remove the wrap here completely.

ryuyu: Looks like it still works without the wrapping. We can remove it if you want.

scottbommarito: I would remove the wrapping then.

PR #7422, opened 2019-08-07
Add support for custom default request options in CloudBlobClientWrapper

joelverhagen:
Related to https://github.com/NuGet/Engineering/issues/2633

See https://github.com/Azure/azure-storage-net/issues/470#issuecomment-306245417 for more information.

There is no timeout by default. This API change will allow Azure Search Service to set a reasonable timeout instead of being unbounded.

### DIFF
--- a/src/NuGetGallery.Core/Services/CloudBlobClientWrapper.cs
+++ b/src/NuGetGallery.Core/Services/CloudBlobClientWrapper.cs
@@ -12,6 +12,7 @@ namespace NuGetGallery
     public class CloudBlobClientWrapper : ICloudBlobClient
     {
         private readonly string _storageConnectionString;
+        private readonly BlobRequestOptions _defaultRequestOptions;
         private readonly bool _readAccessGeoRedundant;
         private CloudBlobClient _blobClient;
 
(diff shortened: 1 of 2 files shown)


src/NuGetGallery.Core/Services/CloudBlobClientWrapper.cs
```diff
@@ -21,6 +22,12 @@ public CloudBlobClientWrapper(string storageConnectionString, bool readAccessGeo
             _readAccessGeoRedundant = readAccessGeoRedundant;
         }
 
+        public CloudBlobClientWrapper(string storageConnectionString, BlobRequestOptions defaultRequestOptions)
```
scottbommarito: Why not add `BlobRequestOptions defaultRequestOptions = null` to the previous constructor?

With this constructor, you can't set `_readAccessGeoRedundant`.

joelverhagen: You can set `_readAccessGeoRedundant` on the `BlobRequestOptions`. In general this API should have been written this way to begin this way to be more extensible. I don't want to change the existing constructor since this is an API change.

Issue #7400, opened 2019-08-02
[Azure Search] Split traffic for hijack by percentage


joelverhagen:
### Updates

9/24/2019 @ 1:10 PM PST - Raised to 100%
9/24/2019 @ 1:05 PM PST - Updated autoscale configs to scale up if CPU% > 30, down if CPU% < 15, to a maximum of 15 instances
9/23/2019 @ 1:55 PM PST - Raised to 75%
9/23/2019 @ 10:21 PM PST - Raised to 40%
9/20/2019 @ 11:40 PM - Updated autoscale configs to scale up if CPU% > 15, and down if CPU% < 5
9/19/2019 @ 1:10 PM PST - Raised to 20%
9/18/2019 @ 8:45 AM PST - Raised to 5%
9/18/2019 @ 3:48 PM PST - Enabled OData hijack feature flag
9/18/2019 @ 2:00 PM PST - Enabled OData hijack to Azure Search at 1%

### Application Insights query 

[...]

PR #7424, opened 2019-08-07
[Azure Search] Split traffic for hijack by percentage

loic-sharma:
Addresses https://github.com/nuget/nugetgallery/issues/7400

Test build: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=2936530
Test release: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=413946

### DIFF
--- a/src/NuGetGallery/App_Data/Files/Content/flags.json
+++ b/src/NuGetGallery/App_Data/Files/Content/flags.json
@@ -1,7 +1,8 @@
 {
   "Features": {
     "NuGetGallery.Typosquatting": "Enabled",
-    "NuGetGallery.PackagesAtomFeed": "Enabled"
+    "NuGetGallery.PackagesAtomFeed": "Enabled",
+    "NuGetGallery.PreviewHijack": "Enabled"
   },
   "Flights": {
     "NuGetGallery.TyposquattingFlight": {
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -17,11 +17,12 @@ using System.Web.Hosting;
 using System.Web.Mvc;
 using AnglicanGeek.MarkdownMailer;
 using Autofac;
-using Autofac.Extensions.DependencyInjection;
 using Autofac.Core;
+using Autofac.Extensions.DependencyInjection;
 using Elmah;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.WindowsAzure.ServiceRuntime;
 using NuGet.Services.Entities;
@@ -45,15 +46,13 @@ using NuGetGallery.Diagnostics;
 using NuGetGallery.Features;
 using NuGetGallery.Infrastructure;
 using NuGetGallery.Infrastructure.Authentication;
+using NuGetGallery.Infrastructure.Lucene;
 using NuGetGallery.Infrastructure.Mail;
 using NuGetGallery.Infrastructure.Search;
 using NuGetGallery.Infrastructure.Search.Correlation;
 using NuGetGallery.Security;
-using SecretReaderFactory = NuGetGallery.Configuration.SecretReader.SecretReaderFactory;
-using Microsoft.Extensions.Http;
-using NuGetGallery.Infrastructure.Lucene;
-using System.Threading;
 using Role = NuGet.Services.Entities.Role;
+using SecretReaderFactory = NuGetGallery.Configuration.SecretReader.SecretReaderFactory;
 
 namespace NuGetGallery
 {
@@ -881,6 +880,16 @@ namespace NuGetGallery
                     c.Resolve<IMessageServiceConfiguration>(),
                     c.Resolve<IIconUrlProvider>()))
                 .As<ISearchSideBySideService>()
+                .InstancePerLifetimeScope();
+
+            builder
+                .Register(c => new HijackSearchServiceFactory(
+                    c.Resolve<HttpContextBase>(),
+                    c.Resolve<IFeatureFlagService>(),
+                    c.Resolve<IContentObjectService>(),
+                    c.Resolve<ISearchService>(),
+                    c.ResolveKeyed<ISearchService>(BindingKeys.PreviewSearchClient)))
+                .As<IHijackSearchServiceFactory>()
                 .InstancePerLifetimeScope();
 
             builder
--- a/src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
+++ b/src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
@@ -15,7 +15,7 @@ namespace NuGetGallery
         /// to avoid performance issues from creating <see cref="SHA256"/> objects.
         /// </summary>
         [ThreadStatic]
-        private static readonly SHA256 Hasher = SHA256.Create();
+        private static SHA256 Hasher;
 
         private readonly HttpContextBase _httpContext;
         private readonly IFeatureFlagService _featureFlags;
@@ -42,6 +42,13 @@ namespace NuGetGallery
             if (!_featureFlags.IsPreviewHijackEnabled())
             {
                 return _search;
+            }
+
+            // Initialize this thread's hasher if necessary.
+            // See: https://docs.microsoft.com/en-us/dotnet/api/system.threadstaticattribute?view=netframework-4.8#remarks
+            if (Hasher == null)
+            {
+                Hasher = SHA256.Create();
             }
 
             var testBucket = GetClientBucket();
--- a/src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
+++ b/src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
@@ -1,0 +1,66 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Cryptography;
+using System.Text;
+using System.Web;
+
+namespace NuGetGallery
+{
+    public class HijackSearchServiceFactory : IHijackSearchServiceFactory
+    {
+        /// <summary>
+        /// The hasher that maps clients to test buckets. There is a single hasher per thread
+        /// to avoid performance issues from creating <see cref="SHA256"/> objects.
+        /// </summary>
+        [ThreadStatic]
+        private static readonly SHA256 Hasher = SHA256.Create();
+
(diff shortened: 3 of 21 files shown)


src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+                isActive,
+                testBucket,
+                testPercentage);
+
+            return isActive ? _previewSearch : _search;
+        }
+
+        public void Dispose() => _hasher.Dispose();
+
+        private int GetClientBucket()
```
loic-sharma: I tested this algorithm using a Chi Square test on a sample of IP addresses from IIS logs. See: http://onlinestatbook.com/2/chi_square/one-way.html

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -871,6 +873,17 @@ private static List<(string name, Uri searchUri)> GetPreviewSearchClientsFromCon
                 .As<ISearchSideBySideService>()
                 .InstancePerLifetimeScope();
 
+            builder
```
loic-sharma: This works even with local lucene search as that is also keyed using the preview search key. See:

https://github.com/NuGet/NuGetGallery/blob/fcb0aa132ca06bfe764e6292e9c68031903c8288/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L797

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+            ITelemetryService telemetryService)
+        {
+            _httpContext = httpContext ?? throw new ArgumentNullException(nameof(httpContext));
+            _featureFlags = featureFlags ?? throw new ArgumentNullException(nameof(featureFlags));
+            _contentObjectService = contentObjectService ?? throw new ArgumentNullException(nameof(contentObjectService));
+            _search = search ?? throw new ArgumentNullException(nameof(search));
+            _previewSearch = previewSearch ?? throw new ArgumentNullException(nameof(previewSearch));
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+
+            _hasher = SHA256.Create();
```
joelverhagen: `SHA256` is not thread safe. This type is registered as lifetime per instance but there is nothing preventing a single request to use parallelism. In general the hashers should be instantiated and disposed as a method local variable for simplicity and safety. The complexity added with the `IDisposable` is not worth it.

Alternatively you could make the hasher `[ThreadStatic]`.

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+            if (!_featureFlags.IsPreviewHijackEnabled())
+            {
+                return _search;
+            }
+
+            var testBucket = GetClientBucket();
+            var testPercentage = _contentObjectService.ABTestConfiguration.PreviewHijackPercentage;
+            var isActive = testBucket <= testPercentage;
+
+            _telemetryService.TrackHijackTestEvaluated(
```
joelverhagen: This may be very frequent right? Didn't we have some AI overload issues on the hijack path? ... like when we did circuit breaker? Maybe non-issue since I forget the specifics.

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            var hashedBytes = _hasher.ComputeHash(Encoding.ASCII.GetBytes(clientData));
+            var value = BitConverter.ToUInt64(hashedBytes, startIndex: 0);
+
+            return (int)(value % 100) + 1;
+        }
+
+        private string GetClientIpAddress()
+        {
+            // If the Gallery is behind a proxy server, the IP address will be in the
```
joelverhagen: We tested this on DEV APIM + Cloud Service?

src/NuGetGallery/App_Data/Files/Content/flags.json
```diff
@@ -1,7 +1,8 @@
 {
   "Features": {
     "NuGetGallery.Typosquatting": "Enabled",
-    "NuGetGallery.PackagesAtomFeed": "Enabled"
+    "NuGetGallery.PackagesAtomFeed": "Enabled",
+    "NuGetGallery.PreviewHijack":  "Enabled"
```
joelverhagen: nit remove double space

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -54,6 +54,8 @@
 using NuGetGallery.Infrastructure.Lucene;
 using System.Threading;
 using Role = NuGet.Services.Entities.Role;
+using System.ServiceModel.Configuration;
+using Microsoft.ApplicationInsights.Channel;
```
joelverhagen: Necessary?

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
loic-sharma: Not yet, but I will before merging this PR. For reference, I copied from `AuditActor`:

https://github.com/NuGet/NuGetGallery/blob/fcb0aa132ca06bfe764e6292e9c68031903c8288/src/NuGetGallery.Core/Auditing/AuditActor.cs#L66-L75

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
skofman1: you can use V3Hasher:
[https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery.Services/Authentication/V3Hasher.cs](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery.Services/Authentication/V3Hasher.cs)

It's static so no need to create each time

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
loic-sharma: I talked to Joel offline. We can get a weakened form of this telemetry using the dependencies table.

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
loic-sharma: V3Hasher is computationally expensive to prevent brute force attacks (see [this](https://github.com/NuGet/NuGetGallery/blob/fcb0aa132ca06bfe764e6292e9c68031903c8288/src/NuGetGallery.Services/Authentication/V3Hasher.cs#L40)). I'd prefer using SHA-256 directly in a hot path like hijacking.

joelverhagen: Yeah I agree with @loic-sharma. This is not the same adversarial situation as password hashing. We just want to bucket things. I would have suggested MD5/SHA1 since they are faster than SHA-256 if I didn't want to get dinged by security requirements.

loic-sharma: Nice catch. I went with method local variable as that's easiest

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
loic-sharma: I'm not sure how those `using`s got pulled in... Fixed!

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Text;
+using System.Web;
+
+namespace NuGetGallery
+{
+    public class HijackSearchServiceFactory : IHijackSearchServiceFactory
+    {
+        /// <summary>
+        /// The hasher that maps clients to test buckets. There is a single hasher per thread
+        /// to avoid performance issues from creating <see cref="SHA256"/> objects.
```
loic-sharma: Performance concern is due to `SHA256.Create`'s dependency on a global lock. See:

https://github.com/dotnet/corefx/blob/9c4bf29c76b21b0b7204a501129d510c7084b5f1/src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/SHA256.cs#L21

https://github.com/dotnet/corefx/blob/9c4bf29c76b21b0b7204a501129d510c7084b5f1/src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/CryptoConfig.cs#L336-L342

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
loic-sharma: We found that IP addresses could fluctuate for a single client across requests and moved away from this approach.

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+            ISearchService previewSearch)
+        {
+            _httpContext = httpContext ?? throw new ArgumentNullException(nameof(httpContext));
+            _featureFlags = featureFlags ?? throw new ArgumentNullException(nameof(featureFlags));
+            _contentObjectService = contentObjectService ?? throw new ArgumentNullException(nameof(contentObjectService));
+            _search = search ?? throw new ArgumentNullException(nameof(search));
+            _previewSearch = previewSearch ?? throw new ArgumentNullException(nameof(previewSearch));
+        }
+
+        public ISearchService GetService()
```
loic-sharma: This method does not emit any telemetry because search hijacking is a very hot path. We will use the `dependencies` Application Insights table to determine success.

loic-sharma:
🔔 This is ready for re-review! /cc @joelverhagen

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
joelverhagen: We don't hijack on local Lucene search, right? Not sure if I understand.

src/NuGetGallery/Infrastructure/HijackSearchServiceFactory.cs
joelverhagen: I see so worst case scenario is we increase by 1% in the config but we really get 11.9% of traffic?

joelverhagen:
Have we set this to 100% on DEV and ran functional tests?

loic-sharma:
I've deployed with the 100% of hijacked queries sent to the preview search, this passed end-to-end and functional tests on the latest changes: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=413946

PR #7425, opened 2019-08-08
[Delete]Progress to enable selfservice delete

ryuyu:
Some requisite changes for allowing self service delete.

Updates accountDeleter to allow configuration of evaluators by source.
Updates gallery to send 2 different source monikers for admin delete action vs user requested action.

Update to enable user self delete will come after.


src/AccountDeleter/Job.cs
```diff
@@ -253,8 +256,13 @@ private void RegisterAuditingServices(IServiceCollection services)
                 }
                 catch (InvalidOperationException)
                 {
+                    auditingServices.Add(GetAuditingServiceForLocalFileSystem());
```
skofman1: use this syntax to avoid code duplication:
catch (Exception ex) when (
    ex is ...
    || ex is ...
    || ex is ...

src/AccountDeleter/Evaluators/EvaluatorKeys.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public static class EvaluatorKeys
```
skofman1: enum?

src/NuGetGallery.Services/AccountManagement/AsynchronousDeleteAccountService.cs
```diff
@@ -18,7 +18,8 @@ namespace NuGetGallery
     /// </summary>
     public class AsynchronousDeleteAccountService : IDeleteAccountService
     {
-        private const string GalleryAccountDeleteMessageSourceName = "Gallery";
+        private const string GalleryAccountDeleteMessageSourceName = "GalleryUser";
```
skofman1: GalleryUserAccountDeleteMessageSourceName

src/AccountDeleter/Job.cs
ryuyu: Ok cool. I like that.

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public override bool CanUserBeDeleted(User user)
+        {
+            var evaluators = _evaluatorList.Values;
+            return evaluators.All(e => e.CanUserBeDeleted(user));
+        }
+
+        public bool AddEvaluator(IUserEvaluator userEvaluator)
+        {
+            if (_evaluatorList.ContainsKey(userEvaluator.EvaluatorId))
```
scottbommarito: I think if you used something like a `HashSet` where the comparator uses `EvaluatorId` you could get rid of all this code and just use the `HashSet` directly.

src/AccountDeleter/Configuration/SourceConfiguration.cs
```diff
@@ -27,5 +29,10 @@ public class SourceConfiguration
         /// Mail template for notifying a user that the account has been deleted.
         /// </summary>
         public MailTemplateConfiguration DeletedMailTemplate { get; set; }
+
+        /// <summary>
+        /// List of evaluators to use for this source. See <see cref="EvaluatorKeys"/> for valid values.
+        /// </summary>
+        public List<string> Evaluators { get; set; }
```
scottbommarito: I think something like `EvaluatorIds` or `EvaluatorKeys` would be clearer. Just `Evaluators` implies that this list contains the actual objects.

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            if (configEvaluators.Count > 1)
+            {
+                var result = new AggregateEvaluator();
+                foreach(var evaluatorKey in configEvaluators)
+                {
+                    IUserEvaluator userEvaluator;
+                    if(!_evaluatorMap.TryGetValue(evaluatorKey, out userEvaluator))
+                    {
+                        throw new UnknownEvaluatorException(evaluatorKey, source);
```
scottbommarito: This exception isn't really correct--either the evaluator was unknown OR the same evaluator was added twice.
I don't really see too much trouble with adding the same evaluator twice, however; you could just de-dupe and reduce down this code a lot.

If you'd like to throw when you see duplicate evaluator keys, I think a better approach would be to throw when loading the configuration.

src/AccountDeleter/Job.cs
```diff
@@ -60,17 +62,18 @@ protected override void ConfigureJobServices(IServiceCollection services, IConfi
             services.AddScoped<AccountConfirmedEvaluator>();
             services.AddScoped<NuGetDeleteEvaluator>();
 
-            services.AddScoped<IUserEvaluator>(sp =>
+            services.AddScoped<IUserEvaluatorFactory>(sp =>
```
scottbommarito: Maybe try adding a `Func<string, IUserEvaluator>` to DI first, then adding the `IUserEvaluatorFactory`, and make the `IUserEvaluatorFactory` use the function?

https://stackoverflow.com/questions/39174989/how-to-register-multiple-implementations-of-the-same-interface-in-asp-net-core

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+                    }
+
+                    result.AddEvaluator(userEvaluator);
+                }
+
+                return result;
+            }
+            else
```
scottbommarito: I think it'd be better if you'd always create an `AggregateEvaluator`, but if you found only a single evaluator, you unwrapped it. That way you don't have to duplicate the logic for fetching evaluators from configuration.

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+        public UserEvaluatorFactory(IOptionsSnapshot<AccountDeleteConfiguration> options)
+        {
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+            _evaluatorMap = new Dictionary<string, IUserEvaluator>();
+        }
+
+        public IUserEvaluator GetEvaluatorForSource(string source)
+        {
+            var configuration = _options.Value;
+
```
scottbommarito: nit: the whitespace between these lines seems unnecessary.

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+                if (!_evaluatorMap.TryGetValue(evaluatorKey, out userEvaluator))
+                {
+                    throw new UnknownEvaluatorException(evaluatorKey, source);
+                }
+
+                return userEvaluator;
+            }
+        }
+
+        public bool AddEvaluatorByKey(string key, IUserEvaluator evaluator)
```
scottbommarito: You have two levels of "add evaluator to dictionary/list"...could we reduce this? It doesn't seem very necessary for `AggregateEvaluator` to perform this check. Can `AggregateEvaluator` just take in a list of evaluators in its constructor?

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public AggregateEvaluator()
+            : base()
+        {
+            _evaluatorList = new Dictionary<string, IUserEvaluator>();
+        }
+
+        public override bool CanUserBeDeleted(User user)
+        {
+            var evaluators = _evaluatorList.Values;
```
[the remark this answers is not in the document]
ryuyu: This is easier to read and debug. Also compiler should likely make this optimization for us. I would rather leave it this way here.

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
ryuyu: Not sure what you mean here. The only case this will get thrown is when config requests an evaluator we don't know about.

src/AccountDeleter/Configuration/AccountDeleteConfiguration.cs
```diff
@@ -33,13 +32,18 @@ public class AccountDeleteConfiguration
...
 
-        public void VerifySource(string source)
+        /// <summary>
+        /// This method get the configuration for a source.
+        /// </summary>
+        /// <param name="source"></param>
+        /// <exception cref="UnknownSourceException">Throws <see cref="UnknownSourceException"/> when source requested is not present in configuration.</exception>
+        public SourceConfiguration GetSourceConfiguration(string source)
         {
             foreach (var sourceConfig in SourceConfigurations)
```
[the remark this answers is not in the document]
ryuyu: I generally find linq to be harder to be debug. If you care a lot about it, I can change it.

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
ryuyu: Good idea. Swapped.

src/AccountDeleter/Configuration/SourceConfiguration.cs
ryuyu: From the config perspective, you are specifying the evaluators you want to use. The fact that we use short names to map them to the actual objects is an implementation detail that the config should be agnostic of.

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
ryuyu: Seems easier to just always use the aggregate since it just wraps one if it comes through. Changed.

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
ryuyu: We could, but then we would have to create a list of evaluators, and then pass them to the aggregate (since what we get from config is a list of keys). The first level is populating the factory with all possible evaluators; the second level is populating an aggregate with all the requested evaluators (which DI is giving us scoped instances of).

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
scottbommarito: My bad, misread this.

src/AccountDeleter/Configuration/AccountDeleteConfiguration.cs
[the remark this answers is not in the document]
scottbommarito: We generally use LINQ in our code when it's something this simple, but it's up to you.

src/AccountDeleter/Configuration/SourceConfiguration.cs
scottbommarito: It's an implementation detail that is very relevant to the configuration: you configure the evaluators you wants to use by writing their short names. The fact that you link to the `EvaluatorKeys` enum is indicative of that. It's up to you, but I think `EvaluatorKeys` is clearer.

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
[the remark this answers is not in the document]
scottbommarito: I don't see how this is easier to read or debug--mousing over `_evaluatorList.Values` will give you whatever object `evaluators` contained--but it's up to you.

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
scottbommarito: That approach is okay as well. The main thing is that we shouldn't have the evaluator-grabbing logic in two places.

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
scottbommarito: That's exactly what I'm suggesting. It hugely reduces the code here.

```
public IUserEvaluator GetEvaluatorForSource(string source)
{
    var sourceConfig = _options.Value.GetSourceConfiguration(source);
    var evaluators = sourceConfig.Evaluators.Select(GetEvaluatorFromKey);
    return new AggregateEvaluator(evaluators);
}
```

You will have to add this helper method if you don't put a `Func<string, IUserEvaluator>` in DI as I suggested as well.
```
private IUserEvaluator GetEvaluatorFromKey(string evaluatorKey)
{
    if (!_evaluatorMap.TryGetValue(evaluatorKey, out var userEvaluator))
    {
        throw new UnknownEvaluatorException(evaluatorKey, source);
    }

    return userEvaluator;
}
```

src/AccountDeleter/Configuration/SourceConfiguration.cs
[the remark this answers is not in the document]
agr: I suggest to avoid using `IEnumerable` in properties: they are potentially stateful (cannot be enumerated twice), which can lead to amusing bugs later.

src/AccountDeleter/Evaluators/EvaluatorKeys.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
```
agr: Missing the copyright header.

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                    result.AddEvaluator(userEvaluator);
+                }
+
+                return result;
+            }
+            else
+            {
+                var evaluatorKey = configEvaluators[0];
+                IUserEvaluator userEvaluator;
```
agr: You can now declare those right after the `out` keyword and the variable scope will be the same.

ryuyu: Oh I did not know that. I will update. Thanks.

src/AccountDeleter/Evaluators/EvaluatorKeys.cs
ryuyu: Good catch. Fixed.

src/AccountDeleter/Job.cs
agr: We can sprinkle a bit of Autofac on top of this:
```
containerBuilder.Register(c => evaluatorKey => c.ResolveKeyed<IUserEvaluator>(evaluatorKey));
```
Then when registering individual evaluators register them keyed:
```
containerBuilder.RegisterType<AccountConfirmedEvaluator>.Keyed<IUserEvaluator>(EvaluatorKeys.AccountConfirmed);
```

That will save on maintaining the map of string to type explicitly.

scottbommarito: I was going to suggest Autofac as well but I think he's just using base `Microsoft.Extensions.DependencyInjection` right now. I don't think it's worth pulling in the new library just for this.

agr: Try finding the `ConfigureAutofacServices` method in this file. Autofac is definitely present.

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
ryuyu: Someone has to populate the _evaluatorMap. Also, i don't see how this reduces code here since we are just moving the resolution inside the foreach into a separate function, and otherwise keeping everything the same. Is your suggestion just to use linq instead of a for each?

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// Evaluator that runs other evaluators and returns a single result. Runs "AND" logic between evaluators.
+    /// </summary>
+    public class AggregateEvaluator : BaseUserEvaluator
+    {
+        private HashSet<IUserEvaluator> _evaluatorList;
+
+        public AggregateEvaluator()
+            : base()
```
loic-sharma: Is this necessary?

src/AccountDeleter/Job.cs
ryuyu: I like the idea. Updated.

src/AccountDeleter/Configuration/SourceConfiguration.cs
ryuyu: Configuration name will stay Evaluators.

src/AccountDeleter/Evaluators/AggregateEvaluator.cs
scottbommarito: I think you can move the declaration of the `_evaluatorList` inline and get rid of this constructor entirely.

ryuyu: The "base()" call, or the entire class?

src/AccountDeleter/Evaluators/UserEvaluatorFactory.cs
scottbommarito: I'm not suggesting you change how you populate `_evaluatorMap`, although I am recommending that you move it into DI instead of having it in this class. **My primary suggestion here is that you remove the dictionary/map entirely from `AggregateEvaluator` and just make it use a `IReadOnlyCollection` or `IEnumerable`.** Regarding LINQ, I am using it in the example above but if you are opposed to it then you can continue to use a `foreach`.

ryuyu: Spoke offline. We are ok with this.

src/AccountDeleter/Configuration/SourceConfiguration.cs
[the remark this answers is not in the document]
scottbommarito: Which is why I mentioned the other two first...sometimes (for example, when working with ASP.NET MVC) you can't use the read only abstractions.

PR #7426, opened 2019-08-08
Reduce flakiness of functional tests

joelverhagen:
This fixes two issues observed:

- `PackagesAppearInFeedInOrderTest` sometimes failed because too many pushes happened from other caller during the test. This causes it to not be able to find a package in the first 100 packages after an arbitrary timestamp.
- `DateTime.[Utc]Now.Ticks` is not a good way to generate unique identifiers. I've observed some 409s. See https://stackoverflow.com/a/2143163. As a side note, I think we had an SDL task about this but perhaps missed/skipped this test code.

### DIFF
--- a/tests/NuGetGallery.FunctionalTests.Core/Helpers/ODataHelper.cs
+++ b/tests/NuGetGallery.FunctionalTests.Core/Helpers/ODataHelper.cs
@@ -59,36 +59,46 @@ namespace NuGetGallery.FunctionalTests
             }
         }
 
-        public async Task<DateTime?> GetTimestampOfPackageFromResponse(string url, string propertyName, string packageId, string version = "1.0.0")
+        public async Task<DateTime?> GetTimestampOfPackageFromResponse(
+            string packageId,
+            string version,
+            string timestampPropertyName)
         {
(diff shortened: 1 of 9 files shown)


tests/NuGetGallery.FunctionalTests.Core/Helpers/ODataHelper.cs
```diff
@@ -59,11 +59,15 @@ public async Task<string> TryDownloadPackageFromFeed(string packageId, string ve
             }
         }
 
-        public async Task<DateTime?> GetTimestampOfPackageFromResponse(string url, string propertyName, string packageId, string version = "1.0.0")
+        public async Task<DateTime?> GetTimestampOfPackageFromResponse(
+            string packageId,
+            string version,
+            string propertyName)
```
loic-sharma: Thoughts on renaming this `timestampPropertyName` like in `GetPackagesAppearInFeedInOrderUrl`? (I think this makes the code a little clearer)

joelverhagen: Yup, sounds good.

PR #7437, opened 2019-08-09
[Delete][Feature] Flagged Self Service Delete

ryuyu:
Introduces a new flag that allows self service account deletion.

testing is still in progress. Thanks.


src/NuGetGallery.Services/Configuration/FeatureFlagService.cs
```diff
@@ -38,6 +39,11 @@ public bool IsAsyncAccountDeleteEnabled()
             return _client.IsEnabled(AsyncAccountDeleteFeatureName, false);
         }
 
+        public bool IsSelfServiceAccountDeleteEnabled()
+        {
+            return _client.IsEnabled(SelfServiceAccountDeleteFeatureName, false);
```
loic-sharma:
```suggestion
            return _client.IsEnabled(SelfServiceAccountDeleteFeatureName, defaultValue: false);
```

src/NuGetGallery.Services/Configuration/IFeatureFlagService.cs
```diff
@@ -15,6 +15,14 @@ public interface IFeatureFlagService
         /// <returns></returns>
         bool IsAsyncAccountDeleteEnabled();
 
+        /// <summary>
+        /// Whether account deletes requested by user are handled automatically or not.
+        /// If true, user account deletes will be deleted automatically if possible
+        /// and send alert mail to user if not.
+        /// </summary>
+        /// <returns></returns>
```
loic-sharma: You can delete this line if you'd like (we should've done this for the other xmldocs, but not a big deal)

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -367,11 +373,30 @@ public override async Task<ActionResult> RequestAccountDeletion(string accountNa
                 return RedirectToAction(nameof(DeleteRequest));
             }
 
-            var result = await DeleteAccountService.DeleteAccountAsync(account, currentUser);
+            DeleteAccountStatus result;
+            string successMessage;
+            if (FeatureFlagService.IsAsyncAccountDeleteEnabled())
```
loic-sharma: Do we want to support the synchronous account deletion feature forever for people self hosting the Gallery? If so, let's add a config (similar to the validation pipeline) as feature flags should be ephemeral.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -334,33 +335,55 @@ public override async Task<ActionResult> RequestAccountDeletion(string accountNa
             }
             TelemetryService.TrackRequestForAccountDeletion(user);
 
-            if (!user.Confirmed)
+            if (FeatureFlagService.IsSelfServiceAccountDeleteEnabled())
```
cristinamanum: What are the cases when we want this enabled?

src/NuGetGallery/Controllers/OrganizationsController.cs
ryuyu: Good call. I'll add a config for this too.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -367,11 +373,30 @@ public override async Task<ActionResult> RequestAccountDeletion(string accountNa
...
+            {
+                var isSupportRequestCreated = await SupportRequestService.TryAddDeleteSupportRequestAsync(account);
+                if (!isSupportRequestCreated)
+                {
+                    TempData["ErrorMessage"] = $"There was an issue deleting your organization '{accountName}'. Please contact support for assistance.";
+
+                    return RedirectToAction(nameof(DeleteRequest));
+                }
+
+                successMessage = $"Your organization, '{accountName}', was queued to be deleted!";
```
scottbommarito: As far as I can tell, this PR doesn't actually implement any sort of "queuing"--it allows users and organizations to delete themselves without admin intervention. We should hold off on using this language until the `DeleteAccountService` actually behaves in that way.

src/NuGetGallery/Controllers/OrganizationsController.cs
scottbommarito: This should be `IsSelfServiceAccountDeleteEnabled`.

src/NuGetGallery/Controllers/OrganizationsController.cs
scottbommarito: Deleting this comment because I was wrong, and after further investigation, we do have a `AsynchronousDeleteAccountService`

src/NuGetGallery/Controllers/OrganizationsController.cs
scottbommarito: My bad, I think this is correct.

However, it's unclear to me if this method should handle `IsSelfServiceAccountDeleteEnabled` as well--whether or not it allows people to delete orgs more freely instead of just requiring no packages/members, etc.

src/NuGetGallery/Controllers/UsersController.cs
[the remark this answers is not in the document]
ryuyu: I don't think passing a Func here makes this much harder to understand, and it is a pattern that is not unique to this.
The given solution keeps the behaviour for feature flag on/off together, which I believe to be more valuable to readability.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -334,33 +335,55 @@ public override async Task<ActionResult> RequestAccountDeletion(string accountNa
...
+                return RedirectToAction("DeleteRequest");
             }
+            OwinContext.Authentication.SignOut();
+            return SafeRedirect(Url.Home(false));
+        }
 
-            var isSupportRequestCreated = await _supportRequestService.TryAddDeleteSupportRequestAsync(user);
+        private async Task<ActionResult> CreateSupportRequestAndTakeAction(User user, Func<Task<ActionResult>> func)
+        {
+            var isSupportRequestCreated = await SupportRequestService.TryAddDeleteSupportRequestAsync(user);
```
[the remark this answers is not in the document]
ryuyu: They won't be broken. They just won't create a support request. Which is exactly the state we were in before.

src/NuGetGallery/Controllers/OrganizationsController.cs
ryuyu: That's a completely different question and is outside of the scope of these changes (for now).

src/NuGetGallery/Controllers/UsersController.cs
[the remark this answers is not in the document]
scottbommarito: Which is bad, because a support request is the "source of truth" behind a queue message to the account delete job. If the support request isn't filed, we could lose the record that an account was supposed to be deleted.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -334,33 +335,55 @@ public override async Task<ActionResult> RequestAccountDeletion(string accountNa
...
+                return await CreateSupportRequestAndTakeAction(user, () => DeleteAndCheckSuccess(user));
+            }
+            else
+            {
+                if (!user.Confirmed)
                 {
-                    TempData["RequestFailedMessage"] = Strings.AccountSelfDelete_Fail;
-                    return RedirectToAction("DeleteRequest");
+                    // Unconfirmed users can be deleted immediately without creating a support request.
+                    return await DeleteAndCheckSuccess(user);
```
scottbommarito: If the delete is asynchronous, we will not file a support request here and there will be no "source of truth" for the deletion.

src/NuGetGallery.Services/Configuration/FeatureFlagService.cs
ryuyu: Updated this oen and the one above it.

src/NuGetGallery.Services/Configuration/IFeatureFlagService.cs
ryuyu: Cleaned up all the empty ones like this in this file.

src/NuGetGallery/Controllers/UsersController.cs
ryuyu: Once we have the account deleter fully in place, we will turn this on so when users request a delete of their account, the first step (determining if they can be deleted or not) will be passed to the job instead of handled by support engineer.

src/NuGetGallery/Controllers/UsersController.cs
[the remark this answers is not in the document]
ryuyu: Spoke offline. Moved support filing to Service. Note this may incur multiple support requests.

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -367,11 +373,21 @@ public override async Task<ActionResult> RequestAccountDeletion(string accountNa
                 return RedirectToAction(nameof(DeleteRequest));
             }
 
+            string successMessage;
+            if (FeatureFlagService.IsAsyncAccountDeleteEnabled())
```
scottbommarito: If you are strongly opposed to changing the return type of `DeleteAccountAsync` so that the asynchronous delete service can communicate to its callers that it is only queuing deletes and not performing them, then I'd rather just get rid of this feature flags call and always say the same message. This controller is essentially making guesses about the implementation of the `DeleteAccountService`, which means the potential for de-sync between the two is very real.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -334,33 +335,47 @@ public override async Task<ActionResult> RequestAccountDeletion(string accountNa
             }
             TelemetryService.TrackRequestForAccountDeletion(user);
 
-            if (!user.Confirmed)
+            if (_config.SelfServiceAccountDeleteEnabled && FeatureFlagService.IsSelfServiceAccountDeleteEnabled())
```
scottbommarito: A couple thoughts here
- We should probably move the config check to the `FeatureFlagService` call itself so that it's not possible to check one without checking the other.
- By requiring both the config flag and the `FeatureFlagService` to return the same value in order for self-service deletes to be enabled, it seems very likely that someone could specify one but forget to specify the other. I'd expect to see an OR here instead of an AND, so that if either flag is specified, self-service is enabled.
- We should add unit tests to this method so that we know that this method behaves as expected with/without self-service deletes enabled.

src/NuGetGallery/Controllers/AccountsController.cs
```diff
@@ -48,6 +49,10 @@ public class ViewMessages
 
         public IContentObjectService ContentObjectService { get; }
 
+        public IFeatureFlagService FeatureFlagService { get; }
+
+        public ISupportRequestService SupportRequestService { get; }
```
scottbommarito: `OrganizationsController` doesn't need this, so it should be safe to move just to `UsersController`.

If you follow my suggestion to remove the `FeatureFlagService` call in `OrganizationsController` as well, I'd move the `FeatureFlagService` field as well.

src/NuGetGallery/Controllers/UsersController.cs
ryuyu: We want this to at least always check the setting, and the feature flag allows us to hotswap. When the feature flag is removed, we still want to check the setting. So it doesn't make sense for the feature flag itself to check the setting as that is not the purpose. And the whole point is that someone must specify both to get the defined behaviour.

src/NuGetGallery/Controllers/OrganizationsController.cs
ryuyu: I would prefer to not change the messaging over change the expected return type of DeleteAccountStatus.

src/NuGetGallery/Controllers/AccountsController.cs
ryuyu: Removed.

src/NuGetGallery/Controllers/AccountsController.cs
```diff
@@ -10,6 +10,7 @@
 using System.Web.Mvc;
 using NuGet.Services.Entities;
 using NuGet.Services.Messaging.Email;
+using NuGetGallery.Areas.Admin;
```
skofman1: the using is the only thing changed in this file. Can it be removed?

src/NuGetGallery/Controllers/OrganizationsController.cs
```diff
@@ -9,6 +9,8 @@
 using System.Web.Mvc;
 using NuGet.Services.Entities;
 using NuGet.Services.Messaging.Email;
+using NuGetGallery.Areas.Admin;
+using NuGetGallery.Areas.Admin.ViewModels;
```
skofman1: remove

src/NuGetGallery/Controllers/AccountsController.cs
ryuyu: yes it can. It is a vestige of an older change.

src/NuGetGallery/Controllers/OrganizationsController.cs
ryuyu: Removed.

Issue #6852, opened 2019-01-26
E2E test for package deprecation


scottbommarito:

scottbommarito:
Changes are good to go.

Must wait to merge this until https://github.com/NuGet/NuGetGallery/commit/9c3410ed69d1650484322a9a4bafcf544d919ab6 goes into PROD.

scottbommarito:
Will need to add `NuGetTestAccount` to the `ManageDeprecation` and `ManageDeprecationApi` feature flags.

scottbommarito:
Merged the end-to-end tests and added `NuGetTestAccount` to the flags in all environments.

PR #7449, opened 2019-08-13
Add API for deprecating packages

scottbommarito:
https://github.com/nuget/nugetgallery/issues/6852

As part of adding end-to-end tests for deprecate, we need a programmatic way to deprecate/undeprecate packages. I felt that adding a public API that is simply a mirror of the web's API would be the best approach for dealing with this rather than trying to create a hidden or less-functional API.

I'll be scheduling a short design meeting to discuss this work, but basically:
- I added `ApiController.DeprecatePackage`, which responds to `/api/v2/package/{id}/deprecations` and expects the following JSON data, where everything except `versions` is optional:
```
{
    "versions": [ "1.0.0", "2.0.0" ],
    "isLegacy": false/true,
    "hasCriticalBugs": false/true,
    "isOther": false/true,
    "alternatePackageId": "alternate-package-ID",
    "alternatePackageVersion": "1.0.0-alternate-package-version",
    "message": "customMessage"
}
```
- I moved most of the code from `ManageDeprecationJsonApiController` into `PackageDeprecationManagementService` so it can be shared by both controllers.

### DIFF
--- a/src/NuGetGallery/App_Start/Routes.cs
+++ b/src/NuGetGallery/App_Start/Routes.cs
@@ -689,7 +689,7 @@ namespace NuGetGallery
                 "v2" + RouteName.DeprecatePackageApi,
                 "api/v2/package/{id}/deprecations",
                 new { controller = "Api", action = RouteName.DeprecatePackageApi },
-                constraints: new { httpMethod = new HttpMethodConstraint("POST") });
+                constraints: new { httpMethod = new HttpMethodConstraint("PUT") });
 
             routes.MapRoute(
                 "v2CuratedFeeds" + RouteName.DownloadPackage,
--- a/src/NuGetGallery/App_Start/Routes.cs
+++ b/src/NuGetGallery/App_Start/Routes.cs
@@ -728,6 +728,12 @@ namespace NuGetGallery
                 constraints: new { httpMethod = new HttpMethodConstraint("POST") });
 
             routes.MapRoute(
+                "v2" + RouteName.DeprecatePackageApi,
+                "api/v2/package/{id}/{version}/deprecate",
+                new { controller = "Api", action = "DeprecatePackageApi" },
+                constraints: new { httpMethod = new HttpMethodConstraint("POST") });
+
+            routes.MapRoute(
                 "v2PackageIds",
                 "api/v2/package-ids",
                 new { controller = "Api", action = "PackageIDs" });
--- a/src/NuGetGallery/App_Start/Routes.cs
+++ b/src/NuGetGallery/App_Start/Routes.cs
@@ -729,7 +729,7 @@ namespace NuGetGallery
 
             routes.MapRoute(
                 "v2" + RouteName.DeprecatePackageApi,
-                "api/v2/package/{id}/deprecate",
+                "api/v2/package/{id}/deprecations",
                 new { controller = "Api", action = RouteName.DeprecatePackageApi },
                 constraints: new { httpMethod = new HttpMethodConstraint("POST") });
 
--- a/src/NuGetGallery/App_Start/Routes.cs
+++ b/src/NuGetGallery/App_Start/Routes.cs
@@ -686,6 +686,12 @@ namespace NuGetGallery
                 });
 
             routes.MapRoute(
+                "v2" + RouteName.DeprecatePackageApi,
+                "api/v2/package/{id}/deprecations",
+                new { controller = "Api", action = RouteName.DeprecatePackageApi },
+                constraints: new { httpMethod = new HttpMethodConstraint("POST") });
+
+            routes.MapRoute(
                 "v2CuratedFeeds" + RouteName.DownloadPackage,
                 "api/v2/curated-feeds/package/{id}/{version}",
                 defaults: new { controller = "Api", action = "GetPackageApi", version = UrlParameter.Optional },
@@ -725,12 +731,6 @@ namespace NuGetGallery
                 "v2" + RouteName.PublishPackageApi,
                 "api/v2/package/{id}/{version}",
                 new { controller = "Api", action = "PublishPackageApi" },
-                constraints: new { httpMethod = new HttpMethodConstraint("POST") });
-
-            routes.MapRoute(
-                "v2" + RouteName.DeprecatePackageApi,
-                "api/v2/package/{id}/deprecations",
-                new { controller = "Api", action = RouteName.DeprecatePackageApi },
                 constraints: new { httpMethod = new HttpMethodConstraint("POST") });
 
             routes.MapRoute(
--- a/src/NuGetGallery/App_Start/Routes.cs
+++ b/src/NuGetGallery/App_Start/Routes.cs
@@ -729,8 +729,8 @@ namespace NuGetGallery
 
             routes.MapRoute(
                 "v2" + RouteName.DeprecatePackageApi,
-                "api/v2/package/{id}/{version}/deprecate",
-                new { controller = "Api", action = "DeprecatePackageApi" },
+                "api/v2/package/{id}/deprecate",
+                new { controller = "Api", action = RouteName.DeprecatePackageApi },
                 constraints: new { httpMethod = new HttpMethodConstraint("POST") });
 
             routes.MapRoute(
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -920,7 +920,6 @@ namespace NuGetGallery
             return new EmptyResult();
         }
 
-
         public virtual async Task<ActionResult> Team()
         {
             var team = await ContentService.GetContentItemAsync(GalleryConstants.ContentNames.Team, TimeSpan.FromHours(1));
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -926,7 +926,7 @@ namespace NuGetGallery
             var error = await PackageDeprecationManagementService.UpdateDeprecation(
                 apiScopeEvaluationResult.Owner,
                 id,
-                versions,
+                versions.ToList(),
                 isLegacy,
                 hasCriticalBugs,
                 isOther,
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -890,7 +890,7 @@ namespace NuGetGallery
             return new EmptyResult();
         }
 
-        [HttpPost]
+        [HttpPut]
         [ApiAuthorize]
         [ApiScopeRequired(NuGetScopes.PackageUnlist)]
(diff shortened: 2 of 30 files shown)


joelverhagen:
Why made this a public feature if there is no scenario for it? Once you support feature publicly you have to support it forever. I think a safer, cheaper solution would be add the API and flight it on the test user. This was we don't really need the additional scope and we could make the "unlist" scope allow this as well.

Seems like this is more like a test hook right now. Or does @anangaur or @karann-msft have a specific experience in mind with this?

scottbommarito:
@joelverhagen - we had talked about making a CLI command earlier on with the feature (`nuget deprecate ...`), but eventually we decided it was a P2. I figured that if we already wanted it, and some form of it is a requirement of E2E tests, I would rather implement this than a private or incomplete API, which we also discussed. We could easily flight this and only make it accessible to the test user as well--it's already behind the normal deprecation flag.

joelverhagen:
>  I figured that if we already wanted it, and some form of it is a requirement of E2E tests, I would rather implement this than a private or incomplete API, which we also discussed.

Makes sense, but incomplete can be a benefit here. Until the CLI experience is nailed down we shouldn't promise a certain API. 

We can get by with something very simple for E2E tests. Nothing fancy or scoped is required.

scottbommarito:
@joelverhagen - I would agree, but the work to implement this was only slightly more than an "incomplete" API because this API calls the same existing backend as the website UI. I could remove the new API key scope and cut out some work there but, ultimately this was fairly cheap to implement overall. We can discuss more in the design meeting.

scottbommarito:
Summary of design meeting:
- change endpoint to ".../deprecations" to match REST better
- remove new API key scope and rely on existing unlist scope
- if API goes public, remember that we will need to apply throttling this endpoint

src/NuGetGallery/Services/PackageDeprecationManagementService.cs
```diff
@@ -0,0 +1,182 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+using NuGet.Versioning;
+
+namespace NuGetGallery
+{
+    public class PackageDeprecationManagementService : IPackageDeprecationManagementService
```
scottbommarito: Note that almost all of this code was moved from `ManageDeprecationJsonApiController`.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -860,6 +890,60 @@ public virtual async Task<ActionResult> PublishPackage(string id, string version
             return new EmptyResult();
         }
 
+        [HttpPost]
+        [ApiAuthorize]
+        [ApiScopeRequired(NuGetScopes.PackageUnlist)]
+        [ActionName(RouteName.DeprecatePackageApi)]
+        public virtual async Task<ActionResult> DeprecatePackage(
+            string id, 
+            [ModelBinder(typeof(ArrayModelBinder<string>))] IEnumerable<string> versions,
```
scottbommarito: Strangely, it seems like ASP.NET MVC doesn't support array parameters in JSON, so I had to write this custom model binder to make this work.

As far as I can tell, all our other actions with array parameter receive form data and not JSON.

scottbommarito: Note that we could remove the ability to specify multiple versions from this endpoint, but I don't think that's a good idea--I think the end-to-end tests should also test multiple versions being deprecated simultaneously.

scottbommarito:
This PR is ready for review after the changes from the design meeting!

src/NuGetGallery/Infrastructure/ArrayModelBinder.cs
```diff
@@ -0,0 +1,42 @@
...
+            if (!isListType && !isArrayType)
+            {
+                throw new InvalidOperationException($"{nameof(ArrayModelBinder<T>)} does not support models of type {bindingContext.ModelType.FullName}");
+            }
+
+            var values = new List<T>();
+            var i = 0;
+            string nextItemName;
+            while (bindingContext.ValueProvider.ContainsPrefix((nextItemName = $"{bindingContext.ModelName}[{i++}]")))
+            {
```
loic-sharma: Can this have an upper bound? Is there an upper limit to the number of versions you can deprecate? Also, I think this `while` loop is a little compact. How about something like:

```
for (var i = 0; i < UpperBound; i++)
{
  var nextItemName = ...
  if (!...ContainsPrefix(nextItemName))
  {
    break;
  }

  // ...
}
```

scottbommarito: There's no upper bound. The endpoint used by the website also doesn't have one. The closest thing to one is that we have a feature flag that blocks deprecating a package that has more than 500 versions.

If there was an upper bound or some other way to iterate the list of parameters, I would have written it in the way you described, but given that there isn't, I wanted to avoid doing a `while (true)`. In general I avoid doing assignments in conditions as well, but I felt that this assignment was simple enough to allow, especially when it allows me to avoid `while (true)`.

src/NuGetGallery/Services/PackageDeprecationManagementService.cs
```diff
@@ -0,0 +1,182 @@
...
+                .OrderByDescending(v => v)
+                .Select(v => v.ToNormalizedString())
+                .ToList();
+        }
+
+        public async Task<UpdateDeprecationError> UpdateDeprecation(
+            User currentUser,
+            string id, 
+            IEnumerable<string> versions, 
+            bool isLegacy = false,
```
agr: The code does multiple enumeration of this `IEnumerable`. Not all implementations can do that. Please use `IReadOnlyCollection`.

scottbommarito: Done.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -54,137 +37,23 @@ public virtual JsonResult GetAlternatePackageVersions(string id)
...
-            {
-                return DeprecateErrorResponse(
-                    HttpStatusCode.Forbidden,
-                    string.Format(Strings.DeprecatePackage_Locked, request.Id));
-            }
+            var error = await _deprecationManagementService.UpdateDeprecation(
+                GetCurrentUser(),
+                request.Id,
+                request.Versions.ToList(),
+                request.IsLegacy,
```
agr: Will anything break if `IReadOnlyCollection` is propagated all the way to `DeprecatePackageRequest`?

scottbommarito: I believe so--as demonstrated by the JSON array shenanigans discussed above, MVC can have some very touchy model binding behavior. I remember encountering issues trying making request properties `IReadOnlyCollection`s before, because MVC only supports `IEnumerable`, so I think it's unlikely that it will support it here.

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -685,6 +685,12 @@ public static void RegisterApiV2Routes(RouteCollection routes)
                     version = UrlParameter.Optional
                 });
 
+            routes.MapRoute(
+                "v2" + RouteName.DeprecatePackageApi,
+                "api/v2/package/{id}/deprecations",
+                new { controller = "Api", action = RouteName.DeprecatePackageApi },
+                constraints: new { httpMethod = new HttpMethodConstraint("POST") });
```
joelverhagen: After some reading `PUT` may be more idiomatic /cc @agr

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -860,6 +890,60 @@ public virtual async Task<ActionResult> PublishPackage(string id, string version
...
+            // Check if the current user's scopes allow deprecating/undeprecating the current package ID
+            var apiScopeEvaluationResult = EvaluateApiScope(ActionsRequiringPermissions.DeprecatePackage, registration, NuGetScopes.PackageUnlist);
+            if (!apiScopeEvaluationResult.IsSuccessful())
+            {
+                return GetHttpResultFromFailedApiScopeEvaluation(apiScopeEvaluationResult, id, versionString: null);
+            }
+
+            if (!FeatureFlagService.IsManageDeprecationApiEnabled(apiScopeEvaluationResult.Owner))
+            {
+                return new HttpStatusCodeWithBodyResult(HttpStatusCode.Forbidden, Strings.ApiKeyNotAuthorized);
```
joelverhagen: Please 404 in this case. The endpoint should always return 404 for non-allowed users. It does not exist for them.

scottbommarito: Should I make this change for the other feature flag as well?

joelverhagen: Actually, I think this is unnecessary. The `ApiScopeRequired` attribute will return non-404 now so it's clear that it exists. What you have is fine.

src/NuGetGallery/Infrastructure/ArrayModelBinder.cs
```diff
@@ -0,0 +1,42 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Web.Mvc;
+
+namespace NuGetGallery
+{
+    public class ArrayModelBinder<T> : IModelBinder
+    {
+        public object BindModel(ControllerContext controllerContext, ModelBindingContext bindingContext)
```
skofman1: one concern here: should this method ever throw? I'm looking at the example here: [https://www.hanselman.com/blog/SplittingDateTimeUnitTestingASPNETMVCCustomModelBinders.aspx](https://www.hanselman.com/blog/SplittingDateTimeUnitTestingASPNETMVCCustomModelBinders.aspx)

Why this might be a problem? 1. Unintended error behavior for user (we should return 400, do we? Or do we return 500?). 2. We might log exceptions that are not actually exceptional.

scottbommarito: I will double-check how this handles with bad input.

scottbommarito: I tested it and it appears that if the model is missing or invalid, it returns an empty array (and not 400, which is consistent with the default model binder). If this model binder is used to bind a model that is not an enumerable, it throws and the request returns 500.

I think the 500 is appropriate here because if it's thrown, it indicates one of two things:
- somebody wrote a `ModelBinderProvider` that told MVC to use this model binder when it shouldn't, which is a server error and not an issue with the request.
- somebody specified this model binder specifically in an attribute, which again indicates a server error and not an issue with the request.

src/NuGetGallery/App_Start/Routes.cs
scottbommarito: PUT appears to be correct. Technically, this should be changed in the website API as well, but I'd rather not make that change with this PR.

Issue #7063, opened 2019-04-11
[Icons] Gallery embedded icon validation


agr:
* File size check;
* File format check;
* zip bomb check;
* tests;

PR #7450, opened 2019-08-13
Icon validation

agr:
Addresses https://github.com/nuget/nugetgallery/issues/7063.

Time to reopen https://github.com/NuGet/Engineering/issues/1883 ?


src/NuGetGallery/Strings.resx
```diff
@@ -1022,22 +1022,23 @@ The {1} Team</value>
     <value>The package has no license information specified.</value>
   </data>
   <data name="UploadPackage_InvalidLicenseFileExtension" xml:space="preserve">
-    <value>The license file has invalid extension '{0}'. Extension must be either empty or one of the following: {1}.</value>
+    <value>The license file has an invalid extension '{0}'. Extension must be either empty or one of the following: {1}.</value>
```
zhhyu: Good one!

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -333,6 +344,120 @@ private async Task<PackageValidationResult> CheckLicenseMetadataAsync(PackageArc
...
+            return await FileMatchesPredicate(nuGetPackage, iconPath, IsPngAsync);
+        }
+
+        private async Task<bool> IsJpegAsync(PackageArchiveReader nuGetPackage, string iconPath)
+        {
+            return await FileMatchesPredicate(nuGetPackage, iconPath, IsJpegAsync);
+        }
+
+        private static async Task<bool> StreamStartsWithAsync(Stream stream, byte[] expectedBytes)
+        {
```
zhhyu: Maybe move this function to StreamHelper class?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -402,6 +527,14 @@ private List<LicenseData> GetLicenseList(NuGetLicenseExpression licenseExpressio
...
+        private static async Task<bool> IsStreamLengthMatchesReportedAsync(PackageArchiveReader nuGetPackage, string path, long reportedLength)
+        {
+            using (var stream = await nuGetPackage.GetStreamAsync(path, CancellationToken.None))
+            {
+                return await IsStreamLengthMatchesReportedAsync(stream, reportedLength);
+            }
+        }
+
         private static async Task<bool> IsStreamLengthMatchesReportedAsync(Stream licenseFileStream, long reportedLength)
         {
```
zhhyu: licenseFileStream -> fileStream?

src/NuGetGallery/Helpers/StreamHelper.cs
```diff
@@ -55,6 +56,40 @@ public static async Task<TruncatedStream> GetTruncatedStreamWithMaxSizeAsync(thi
             }
         }
 
+        /// <summary>
+        /// Checks whether stream's next bytes are the same as <paramref name="expectedBytes"/>.
```
joelverhagen: nit: maybe comment that it's the caller's responsibility to seek to the proper position.

Also, this method assumes that the position is 0 because "starts with" suggests the beginning of the stream not the current position.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -30,12 +31,22 @@ public class PackageUploadService : IPackageUploadService
...
+            ".png"
+        };
+
         private static readonly IReadOnlyCollection<string> AllowedLicenseTypes = new HashSet<string>
         {
             LicenseType.File.ToString(),
             LicenseType.Expression.ToString()
         };
 
+        private static readonly byte[] PngHeader = new byte[] { 0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A };
```
joelverhagen: Can you link to specs here please?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -333,6 +344,107 @@ private async Task<PackageValidationResult> CheckLicenseMetadataAsync(PackageArc
...
+            {
+                return PackageValidationResult.Invalid(Strings.UploadPackage_EmbeddedIconNotAccepted);
+            }
+
+            if (HasChildElements(iconElement))
+            {
+                return PackageValidationResult.Invalid(string.Format(Strings.UploadPackage_NodeContainsChildren, IconNodeName));
+            }
+
+            var iconFilePath = FileNameHelper.GetZipEntryPath(iconElement.Value);
```
joelverhagen: Ideally this stuff comes from a client API. Is there a tracking issue to switch to that?

agr: https://github.com/NuGet/NuGetGallery/issues/7459

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -333,6 +359,107 @@ private async Task<PackageValidationResult> CheckLicenseMetadataAsync(PackageArc
...
+            if (!AllowedIconFileExtensions.Contains(iconFileExtension, StringComparer.OrdinalIgnoreCase))
+            {
+                return PackageValidationResult.Invalid(
+                    string.Format(
+                        Strings.UploadPackage_InvalidIconFileExtension,
+                        iconFileExtension,
+                        string.Join(", ", AllowedIconFileExtensions.Where(x => x != string.Empty).Select(extension => $"'{extension}'"))));
+            }
+
+            var iconFileEntry = nuGetPackage.GetEntry(iconFilePath);
```
skofman1: nuGetPackage.GetStreamAsync is called multiple times during the execution of this method. Does this impact perf for large packages?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -333,6 +359,107 @@ private async Task<PackageValidationResult> CheckLicenseMetadataAsync(PackageArc
             return null;
         }
 
+        private async Task<PackageValidationResult> CheckIconMetadataAsync(PackageArchiveReader nuGetPackage, User user)
```
skofman1: any code sharing potential between license and icons? Both need to get a file stream from a nupkg, verify extension, check length, etc. ?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -333,6 +359,107 @@ private async Task<PackageValidationResult> CheckLicenseMetadataAsync(PackageArc
...
+        {
+            return await FileMatchesPredicate(nuGetPackage, iconPath, IsPngAsync);
+        }
+
+        private async Task<bool> IsJpegAsync(PackageArchiveReader nuGetPackage, string iconPath)
+        {
+            return await FileMatchesPredicate(nuGetPackage, iconPath, IsJpegAsync);
+        }
+
+        private async Task<bool> IsPngAsync(Stream stream)
```
skofman1: nit: those methods can be inlined

tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
```diff
@@ -1140,6 +1141,174 @@ public async Task AcceptsPackagesWithEmbeddedIconForFlightedUsers()
...
+                Assert.Contains(iconFilename, result.Message.PlainTextMessage);
+                Assert.Empty(result.Warnings);
+            }
+
+            [Theory]
+            [InlineData("icons/main.png")]
+            [InlineData("./other/something.jpg")]
+            [InlineData("media\\icon.jpeg")]
+            [InlineData(".\\data\\awesome.png")]
+            public async Task AcceptsPackagesWithIconsInSubdirectories(string iconFilename)
```
skofman1: what the path goes outside the directory? 
..\..\system\icon.png? :)

src/NuGetGallery/Helpers/StreamHelper.cs
```diff
@@ -55,6 +56,45 @@ public static async Task<TruncatedStream> GetTruncatedStreamWithMaxSizeAsync(thi
...
+        /// <param name="stream">Stream to read from.</param>
+        /// <param name="expectedBytes">Expected bytes.</param>
+        /// <returns>True if it manages to read the same amount of bytes as in <paramref name="expectedBytes"/> and 
+        /// the bytes read are the same as in <paramref name="expectedBytes"/> array, false otherwise.</returns>
+        /// <remarks>
+        /// The method assumes that stream is properly positioned before calling by the caller. It will not seek anywhere.
+        /// If the length of <paramref name="expectedBytes"/> is 0, method returns true.
+        /// </remarks>
+        public static async Task<bool> NextBytesMatchAsync(this Stream stream, byte[] expectedBytes)
+        {
```
zhhyu: Do we have the test coverage for this public method?

src/NuGetGallery/Services/PackageUploadService.cs
agr: Most of the individual validation logic is already moved to their own methods and shared between those two. So in both cases validation is pretty much a sequence of `if` statements. Trying to share more code, I'm afraid, will lead to more complicated code as a result.

tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
agr: What exactly do you want to be tested? The library that we use to create nuget packages for test is capable of creating zip file entries with double dots: `..\somedir\somefile`. And the code that we use to read data from zip files is able to read those as well, so if I simply add `[InlineData("..\\data\\awesome.png")]` here, it passes.
If you want to test that we actually read the file from zip file and don't try to read it from disk using the same path, that's a different story and I'd say out of scope.

agr: The case of paths that end up outside of package should have been covered by the following change: https://github.com/NuGet/NuGetGallery/pull/6073

agr: And tests that cover that functionality in client libraries are still there: https://github.com/NuGet/NuGet.Client/blob/a37bdcfcdf54ad2db09d80540d63d97819666455/test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageArchiveReaderTests.cs#L888-L978

agr: Added tests for the case when `iconUrl` is present, no associated zip file entry is present, but there is a file in local file system with the same name.

PR #7456, opened 2019-08-15
[Azure Search] Allow default searches to go to preview search

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7455

Also, add a GUID to correlate clicks and searches more precisely than search term.

### DIFF
--- a/src/NuGetGallery/Views/Shared/ListPackages.cshtml
+++ b/src/NuGetGallery/Views/Shared/ListPackages.cshtml
@@ -62,9 +62,7 @@
         }
         @foreach (var package in Model.Items)
         {
-            // Only set an item index when there is a search query, for now.
-            var thisItemIndex = string.IsNullOrWhiteSpace(Model.SearchTerm) ? null : (int?) itemIndex;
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -84,15 +82,17 @@
         // Used to track how many selections were made on this page. Multiple selections can happen if the user opens
         // a search selection in a new tab, instead of navigating away from this page.
         var sincePageLoadCount = 0;
-        @if (!string.IsNullOrWhiteSpace(Model.SearchTerm) && (Model.PageIndex == 0 || Model.Items.Count() > 0))
```
scottbommarito: What does removing the `Model.PageIndex == 0 || Model.Items.Count() > 0` check do here?

joelverhagen: I was previously protecting against emitting an event when the user is on, say, the second page but there are no items. This can be filtered out at the telemetry side and should never happen in practice so I wanted to simplify the code.

Issue #7461, opened 2019-08-18
TempData messages are no longer working


joelverhagen:
For example, as an admin, when you reflow a package you don't see the grey bar message on the subsequent page that tells you the reflow is starting.

This is a regression and likely affects customer scenarios.

scottbommarito:
Explanation of the issue, for posterity:

This was caused by @joelverhagen re-registering a controller in DI. Controllers are usually wired up into DI by MVC automatically, but he wanted to modify the wiring so he replaced the default registration. This controller was then, for an unknown reason, repeatedly given the same instance of a `CookieTempDataProvider`. Because it was never reconstructed for each request, this `CookieTempDataProvider` was never given the current requests' `HttpContext`, so any attempts to access `TempData` failed.

PR #7462, opened 2019-08-19
[Hotfix] Fix bug causing TempData messages to not work

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7461
Regressed in https://github.com/NuGet/NuGetGallery/pull/7313

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -897,11 +897,10 @@ namespace NuGetGallery
                 .InstancePerLifetimeScope();
 
             builder
-                .RegisterType<PackagesController>()
-                .WithParameter(new ResolvedParameter(
-                    (pi, ctx) => pi.ParameterType == typeof(ISearchService) && pi.Name == ParameterNames.PackagesController_PreviewSearchService,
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -897,11 +897,10 @@ private static List<(string name, Uri searchUri)> GetPreviewSearchClientsFromCon
                 .InstancePerLifetimeScope();
 
             builder
-                .RegisterType<PackagesController>()
```
joelverhagen: This DI wire-up was the culprit.

src/NuGetGallery/Infrastructure/ISearchServiceFactory.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    public interface ISearchServiceFactory
```
joelverhagen: I could have refactored the `IHijackSearchServiceFactory` but chose not to as to keep a small hotfix.

Issue #7239, opened 2019-06-05
[Icons] Gallery embedded icon preview for validate package page


agr:
Icon should be displayed on the validate package page right after package upload.

Issue #7093, opened 2019-04-19
[Icons] Icon-related Gallery UI while package is being published


agr:
Need to ensure proper fallbacks are in place and message is displayed in Gallery for the package owner regarding the package availability while indexing.

agr:
What's currently in dev has bugs.

PR #7465, opened 2019-08-20
Embedded icon display on verify package page and indexing message

agr:
Addresses #7239, #7093.

Moved the image type identification bits to the `ImageExtensions` to be shared between controllers and services.
Updated `PackagesController` to identify the image type from the image signature and generate data: url with proper content type.

On a verify package page it looks like this:
image
(the icon in this case is 1024x1024 reduced by browser to our "regular size")

### DIFF
--- a/src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml
+++ b/src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml
@@ -201,11 +201,13 @@
                 <div data-bind="template: { name: 'display-data-or-default', data: { DisplayText: $data.RepositoryUrl }}"></div>
             </div>
 
+            <!-- ko if: !$data.IconUrl || !$data.IconUrl.startsWith('data:') -->
             <div class="verify-package-field">
                 <!-- Bind an on change event here to update the preview below-->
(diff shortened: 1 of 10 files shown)


src/NuGetGallery/Views/Packages/_VerifyMetadata.cshtml
```diff
@@ -201,11 +201,13 @@
                 <div data-bind="template: { name: 'display-data-or-default', data: { DisplayText: $data.RepositoryUrl }}"></div>
             </div>
 
+            <!-- ko if: !$data.IconUrl || !$data.IconUrl.startsWith('data:') -->
```
scottbommarito: Should we show some UI explaining that the icon is inside the package and cannot be displayed here when someone tries to upload an embedded icon?

agr: The icon will be displayed on the verify package page. That code removes the portion of the page that displays the Icon URL, because that would be just a long base64 string in case of embedded icon.

scottbommarito: Oh, I see. Looking good!

Issue #7033, opened 2019-04-01
Deleted users are locked out from site until their session token expires


scottbommarito:
1. Log into account A.
2. Delete account A **without logging out**
    - Have a site admin delete you
    - Delete the account from the DB
3. Attempt to browse the gallery.

Any request that loads the current user (`GetCurrentUser`) will throw at https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/ExtensionMethods.cs#L481

AraHaan:
hmm when they delete their account, why not automatically invalidate their session token.

scottbommarito:
https://github.com/NuGet/NuGetGallery/pull/7469

PR #7469, opened 2019-08-20
Fix Error when Authed, but deleted

ryuyu:
In the case when a session was authenticated as a user, but the user was deleted BEFORE the session was logged out, every subsequent to the gallery would 500.

This fix makes it so if we get into this state, the user will be logged out, redirected to the home page, and shown a message indicating that their account was deleted.

Preview:
image


src/NuGetGallery/Strings.resx
```diff
@@ -1141,4 +1141,7 @@ The {1} Team</value>
   <data name="UploadPackage_UnsupportedIconImageFormat" xml:space="preserve">
     <value>Unsupported icon image format. Only PNG and JPG images are supported.</value>
   </data>
+  <data name="LoggedInUserDeleted" xml:space="preserve">
+    <value>Your account is no longer active and we have logged you out. Please contact support if this is unexpected.</value>
```
joelverhagen: Should we put the support email in there? i think there is a config value for it.

src/NuGetGallery/Infrastructure/UserDeletedErrorFilter.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public sealed class UserDeletedErrorFilter : FilterAttribute, IExceptionFilter
+    {
+        public void OnException(ExceptionContext context)
+        {
+            if (context.Exception is UserDeletedException)
+            {
+                context.HttpContext.Response.Clear();
+                context.HttpContext.Response.TrySkipIisCustomErrors = true;
+                context.HttpContext.Response.StatusCode = 401;
```
joelverhagen: why set 401? I guess the `RedirectResult` below is taking precedence.

src/NuGetGallery/Infrastructure/UserDeletedErrorFilter.cs
```diff
@@ -0,0 +1,28 @@
...
+
+                context.Controller.TempData = new TempDataDictionary
+                {
+                    { "Message", Strings.LoggedInUserDeleted }
+                };
+
+                context.Result = new RedirectResult("/");
```
joelverhagen: This is a temporary redirect right? :)

src/NuGetGallery/Infrastructure/UserDeletedErrorFilter.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+using System.Web.Mvc;
+
+namespace NuGetGallery.Infrastructure
+{
+    public sealed class UserDeletedErrorFilter : FilterAttribute, IExceptionFilter
+    {
+        public void OnException(ExceptionContext context)
+        {
+            if (context.Exception is UserDeletedException)
```
joelverhagen: Can we verify that the user is indeed logged in at the beginning of the request?

src/NuGetGallery/Strings.resx
scottbommarito: I think `no longer active` is not clear enough. People are going to think that we have different concepts of "inactive" and "deleted" and that's simply not true. If somebody's account was unintentionally deleted, they need to know that it was so they can reach out to us as soon as possible, and using the term "inactive" might confuse people. I think this should just say `Your account was deleted`.

src/NuGetGallery/Infrastructure/UserDeletedErrorFilter.cs
scottbommarito: I think this case is exceptional enough that we should have some telemetry about it.  I know we have another error filter that logs exceptions. We should have at least some sort of log here.

src/NuGetGallery/Infrastructure/UserDeletedErrorFilter.cs
scottbommarito: What's the default behavior when we return 401? Doesn't it redirect to sign in? Should we consider doing that instead of redirecting to the home page?

src/NuGetGallery.Services/AccountManagement/UserDeletedException.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+using System.Web;
+
+namespace NuGetGallery
+{
+    public class UserDeletedException : HttpException
```
scottbommarito: I think we can have a more clear name here.

`CurrentUserWasDeleted` or `LoggedInUserWasDeleted` are clearer about the problem. Don't forget to change the name of the error filter as well.

Also, consider adding some XML docs to this class.

src/NuGetGallery.Services/Extensions/IOwinContextExtensions.cs
```diff
@@ -47,7 +47,8 @@ public static User GetCurrentUser(this IOwinContext self)
             {
                 // Unauthorized! If we get here it's because a valid session token was presented, but the
                 // user doesn't exist any more. So we just have a generic error.
-                throw new HttpException(401, ServicesStrings.Unauthorized);
+                self.Authentication.SignOut();
```
scottbommarito: Could you move this sign out call to the error filter?

The reason I say this is that regardless of where we discover the current user was deleted, we should always log them out. If we discover other code paths in which we find the current user has been deleted, we shouldn't have to add this sign out call there too.

Not a big deal though, as it's unlikely we'll find another place.

skofman1:
How was this tested?

src/NuGetGallery/Infrastructure/UserDeletedErrorFilter.cs
ryuyu: It probably is. This was a vestige of something i was trying earlier. It should be safe to remove it entirely.

src/NuGetGallery/Infrastructure/UserDeletedErrorFilter.cs
ryuyu: Looks like it redirects with 302, so should be temporary.

ryuyu: Removing the 401 status set here.

src/NuGetGallery.Services/Extensions/IOwinContextExtensions.cs
ryuyu: Probably not. We don't have access to the IOwinContext inside the error filter. I imagine we could emulate it by manually clearing cookies, but I find making this call (the one we use other places to log out) more comfortable.

scottbommarito: Agreed. Was just curious if you could.

src/NuGetGallery.Services/AccountManagement/UserDeletedException.cs
ryuyu: UserDeletedException -> CurrentUserDeletedException

src/NuGetGallery/Strings.resx
ryuyu: Technically, if a user logged in with MSA/AAD and the account was deleted, they can still log in with the same MSA/AAD and it will link to a new NuGet Account. I think "no longer active" better describes this state (ie the specific NuGet Account that was part of your identity, isn't there anymore for one reason or another), but I don't feel strongly about it. I will defer to @anangaur @karann-msft

src/NuGetGallery/Infrastructure/UserDeletedErrorFilter.cs
ryuyu: They won't be at this point. We log them out before we throw the exception.
The GetCurrentUser extension method on the IOwinContext already verifies that a identity was present to us at the beginning of the call.

TL;DR:
If there was no logged in session, we would never hit this path to begin with.

src/NuGetGallery/Strings.resx
scottbommarito: That's an interesting edge-case, but I still believe `Your NuGet account was deleted` is the best thing to say there. It removes any confusion over MSA vs AAD vs NuGet while also communicating what happened.

ryuyu:
Tested locally for both password auth and MSA auth accounts.

scottbommarito:
Does this PR fix https://github.com/NuGet/NuGetGallery/issues/7033?

src/NuGetGallery/Strings.resx
loic-sharma: Also, consider inverting the first sentence: `You have been logged out as ...`

ryuyu: updated.

shishirx34: I don't suppose we have the state of the user deletion? as in how was the account deleted? was it DSR or user request to delete. If we can specify the reason it will be helpful in customer requests if someone were to see this and dispute.

ryuyu: We have no idea at this point. We just know that we got a valid user session, and the user it specified didn't exist.

src/NuGetGallery/Infrastructure/UserDeletedErrorFilter.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+using System.Web.Mvc;
+
+namespace NuGetGallery.Infrastructure
+{
+    public sealed class UserDeletedErrorFilter : FilterAttribute, IExceptionFilter
```
scottbommarito: Rename this to `CurrentUserDeletedErrorFilter`?

PR #7471, opened 2019-08-21
Add WasUsernameReleased field to AccountDelete

scottbommarito:
Part of https://github.com/NuGet/Engineering/issues/1303

The idea is that there will be a job (an additional task in Gallery.Maintenance) that queries for all accounts that were deleted more than a year in the past and renames them to a Guid or something similar. We need to have an additional column on the `AccountDelete`s table so that we can remember what accounts we've renamed because otherwise we won't be able to reliably know what accounts we've renamed.

Note that the new column will be initialized to `default(bool)` aka `false`, so all existing columns will have the value we expect.


src/NuGet.Services.Entities/AccountDelete.cs
```diff
@@ -43,9 +43,15 @@ public class AccountDelete
...
+        /// <summary>
+        /// The signature of the admin.
+        /// </summary>
+        public string Signature { get; set; }
+
+        /// <summary>
+        /// Whether or not the username of the deleted account was released.
+        /// The username of a deleted account should be eventually released so that new users can take the username.
+        /// </summary>
+        public bool WasUsernameReleased { get; set; }
```
agr: We are going to run some queries involving checking the value of that property, do we need some index on it?

scottbommarito: Is it worth adding an index for a query that's run once a day?

agr: Depends on expected table size. No index means we'd have to scan the whole table. Probably wouldn't matter until we have tens of thousands of rows there.

drewgillies:
This was approved and unmerged some time back. Is there a reason this hasn't proceeded, and if so, should it be closed? @scottbommarito @agr @loic-sharma @joelverhagen

joelverhagen:
Let's close it. We'll take this later if it needs attention.

PR #7480, opened 2019-08-23
Add message to self service delete.

ryuyu:
Small PR to improve the self service account delete scenario.
Adds a message to the post redirect to indicate that an action was taken.


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -379,6 +379,10 @@ private async Task<ActionResult> DeleteAndCheckSuccess(User user)
                 return RedirectToAction("DeleteRequest");
             }
             OwinContext.Authentication.SignOut();
+            TempData = new TempDataDictionary
```
joelverhagen: Can we use `TempData["Message"] = ...`

PR #7502, opened 2019-09-04
Add token expiration of 14 days

joelverhagen:
Tokens are used for:
 - adding owners to packages
 - adding members to organizations
 - transforming users to organizations

Progress on https://github.com/NuGet/Engineering/issues/2679

When a token is older than 14 days, it will no longer work. This means that "accept" / "reject" links in email will stop working 14 days after they are sent. All of these action links have counterparts in the web UI. When the request is older than 14 days, the recipient of the request will see "(request expired)".

The action links ("accept" / "reject") are not visible when the request is expired. In other words, the recipient sees that the request is expired for informational purposes only. I spoke offline with @karann-msft and we agreed that we don't need to implement a "dismiss" (essentially a POST-based ) button for the recipient. If we hear significant feedback from users, then we can revisit this decision later and likely add a `PostLink` to a new "dismiss" or "cancel" endpoint that supports POST.

## Existing state (no expiration concept)

Operation | Sender can cancel? | Sender can cancel? | Recipient can cancel? | Recipient can cancel?
-- | -- | -- | -- | --
&nbsp; | **Before expiration** | **After Expiration** | **Before Expiration** | **After Expiration**
Add owner to package |  ✅ | ✅ | ✅ | ✅
Add member to organization |  ✅ | ✅ | ✅ | ✅
Transform organization |  ❌ | ❌ | ✅ | ✅

Note that the recipient of the "Transform organization" operation is the user that was specified as the admin of the new organization. There is currently no way for the sender of a "transform organization" operation to cancel.

## After this PR

Operation | Sender can cancel? | Sender can cancel? | Recipient cancel? | Recipient cancel?
-- | -- | -- | -- | --
&nbsp; | **Before expiration** | **After Expiration** | **Before Expiration** | **After Expiration**
Add owner to package |  ✅ | ✅ | ✅ | ❌
Add member to organization |  ✅ | ✅ | ✅ | ❌
Transform organization |  ❌ | ❌ | ✅ | ❌

## Sender: expired package ownership request

image

image

## Recipient: expired package ownership request

image

## Sender: expired organization membership request

image

## Recipient: expired organization membership request

image

## Recipient: expired transform organization

image

### DIFF
--- a/src/NuGetGallery.Services/Extensions/RequestEntityExtensions.cs
+++ b/src/NuGetGallery.Services/Extensions/RequestEntityExtensions.cs
@@ -1,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public static class RequestEntityExtensions
+    {
+        private static readonly TimeSpan RequestTokenLifetime = TimeSpan.FromDays(14);
+
+        public static bool IsExpired(this OrganizationMigrationRequest request)
+        {
+            return IsExpired(request.RequestDate);
+        }
+
+        public static bool IsExpired(this MembershipRequest request)
+        {
+            return IsExpired(request.RequestDate);
+        }
+
+        public static bool IsExpired(this PackageOwnerRequest request)
+        {
+            return IsExpired(request.RequestDate);
+        }
+
+        private static bool IsExpired(DateTime requestDate)
+        {
+            return DateTime.UtcNow - requestDate > RequestTokenLifetime;
+        }
+    }
+}
--- a/src/NuGetGallery.Services/UserManagement/UserService.cs
+++ b/src/NuGetGallery.Services/UserManagement/UserService.cs
@@ -181,6 +181,12 @@ namespace NuGetGallery
                 throw new InvalidOperationException("No such membership request exists!");
             }
 
+            if (request != null && confirmationToken != null && request.IsExpired())
+            {
+                throw new EntityException(string.Format(CultureInfo.CurrentCulture,
+                    ServicesStrings.RejectMembershipRequest_Expired, memberName));
+            }
+
             var pendingMember = request.NewMember;
 
             organization.MemberRequests.Remove(request);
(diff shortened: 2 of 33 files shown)


src/NuGetGallery.Services/UserManagement/UserService.cs
```diff
@@ -538,16 +550,28 @@ public bool CanTransformUserToOrganization(User accountToTransform, User adminUs
...
             await SubscribeOrganizationToTenantPolicyIfTenantIdIsSupported(accountToTransform, adminUser);
             var result = await EntitiesContext.TransformUserToOrganization(accountToTransform, adminUser, token);
             if (result)
             {
                 await Auditing.SaveAuditRecordAsync(new UserAuditRecord(accountToTransform, AuditedUserAction.TransformOrganization, adminUser, affectedMemberIsAdmin: true));
             }
 
-            return result;
+            return result ? TransformOrganizationResult.Success : TransformOrganizationResult.Failure;
         }
```
loic-sharma: Nit: consider returning `TransformOrganizationResult.Success` after line 571 (inside the `if (result)`'s block) and just returning `.Failure` here

joelverhagen: Yeah that is better

src/NuGetGallery.Services/Extensions/RequestEntityExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    public static class RequestEntityExtensions
+    {
+        private static readonly TimeSpan RequestTokenLifetime = TimeSpan.FromDays(14);
+
+        public static bool IsExpired(this OrganizationMigrationRequest request)
+        {
+            return IsExpired(request.RequestDate);
+        }
+
+        public static bool IsExpired(this MembershipRequest request)
```
scottbommarito: Could you make an interface and make `MembershipRequest`, `PackageOwnerRequest`, etc extend it? So that we don't have to have clones of this method for each type? Similar to like `IEntity`? It shouldn't affect the EF scaffolding.

joelverhagen:
/cc @karann-msft 

If we go down this route, we need to update the notification emails to say "this request expires in 14 days".

joelverhagen:
Closed for now. I'm working on a more POST-based approach instead of expiration.

PR #7508, opened 2019-09-05
[ReleasePrep] RI dev to master

agr:
Apparently, previous dev -> master was squash-merged, so there is more things here than should be.


joelverhagen:
Could you wait for https://github.com/NuGet/NuGetGallery/pull/7505 please? I am verifying in DEV right now.

Issue #7512, opened 2019-09-06
Investigate using AppInsights directly instead of System.Diagnostics.Trace in DiagnosticsService


xavierdecoster:
Related to ongoing efforts to upgrade Application Insights to v2.10 in the gallery, and unblock heartbeat efforts which require v2.10 dependency in ServerCommon.

loic-sharma:
I verified that the content object service's traces (which depend on the `DiagnosticsService`) flow to Application Insights

PR #7513, opened 2019-09-06
Use AppInsights instead of System.Diagnostics.Tracing in DiagnosticsService

xavierdecoster:
Summary of the changes:

* `DiagnosticsService` is not tracing through `System.Diagnostics.TraceSource` anymore but directly through Application Insights `TelemetryClient.TrackTrace` API
* Kept `Trace.AutoFlush = true` for now, as I believe we still want that. This setting does cause the AI Trace Listener to `Flush()` each trace through AI separately, which previously lead to lock contention on flushing the buffer too often. However, the intended change of this PR is to no longer trace through `System.Diagnostics.Trace`, so the rare events that are traced through `System.Diagnostics.Trace` may as well be flushed immediately. However, we should be less exposed to any potential perf impact introduced by auto-flushing behavior of AI Trace Listener.

This change should loosen our dependency on `Microsoft.ApplicationInsights.TraceListener`, although I'm not removing it completely. 
* Some trace events, e.g. during `WebRole.OnStop` still go through `System.Diagnostics.Trace`, and (external) dependencies may be using the `System.Diagnostics.Trace` APIs. 
* We also still wire up the [`DiagnosticMonitorTraceListener`](https://docs.microsoft.com/en-us/azure/cloud-services/cloud-services-dotnet-diagnostics-trace-flow) in `AppActivator`, [line 113](https://github.com/NuGet/NuGetGallery/blob/f2c21be69b34e3f8fc1cc192d4e1ca09bcc1a9d5/src/NuGetGallery/App_Start/AppActivator.cs#L113). 

If you think I missed a spot, or have concerns, let's discuss!

Addresses #7512


src/NuGetGallery/App_Start/AppActivator.cs
```diff
@@ -39,6 +39,8 @@ public static class AppActivator
 
         public static void PreStart()
         {
+            Trace.AutoFlush = true;
```
joelverhagen: Do we really need this? This seems a bit scary.

src/NuGetGallery.Core/Diagnostics/IDiagnosticsSource.cs
```diff
@@ -12,10 +11,10 @@ namespace NuGetGallery.Diagnostics
     public interface IDiagnosticsSource : ILogger
     {
         void ExceptionEvent(Exception exception);
-        
-        void TraceEvent(TraceEventType type, int id, string message,
+
+        void TraceEvent(LogLevel logLevel, EventId eventId, string message,
```
joelverhagen: This is kind of the wrong interface for `Microsoft.Extensions.Logging`. `ILogger` works better when you don't combine the parameters with the message yourself. We can punt this part if you want but it shouldn't be too bad if you add an `params object[]` parameter.

src/NuGetGallery/App_Start/AppActivator.cs
xavierdecoster: We already had it [here](https://github.com/NuGet/NuGetGallery/pull/7513/files#diff-0b70d5e18960c2ab4f8b24f71be12222L18), I just moved it around to application pre-start outside of the constructor of `DiagnosticsService`.

Happy to remove it completely, but it's hard to determine if we miss telemetry by removing this.

src/NuGetGallery.Core/Diagnostics/IDiagnosticsSource.cs
xavierdecoster: Perhaps that work should be done as part of a separate work item. Want to log an issue for that and describe the change you'd like to see? This PR just changes the type of the first two arguments on this particular interface method.

src/NuGetGallery/App_Start/AppActivator.cs
joelverhagen: Oh, okay. My mistake.

src/NuGetGallery.Services/Diagnostics/TraceDiagnosticsSource.cs
```diff
@@ -111,20 +87,7 @@ bool ILogger.IsEnabled(LogLevel logLevel)
 
         IDisposable ILogger.BeginScope<TState>(TState state)
```
skofman1: I assume this is not used?

src/NuGetGallery.Core/Diagnostics/IDiagnosticsSource.cs
joelverhagen: I'll leave it be for now. If we find this painful in the future we can open an issue. Thanks for clarifying.

src/NuGetGallery.Services/Diagnostics/TraceDiagnosticsSource.cs
xavierdecoster: I think not. Scopes are not supported in AI traces AFAIK, so `NotSupportedException` seems appropriate?

agr: Logging shouldn't throw, should it?

xavierdecoster: @agr What's your suggested way to implement this unsupported API?

PR #7420, opened 2019-08-07
Click on "show more" should not scroll page to top

304NotModified:
Currently clicking on "+  Show more" (under  Documentation) scrolls the page to the top. This is very visible on mobile. 

Package example: https://www.nuget.org/packages/NLog/4.6.6

Return false will fix this

PR #7516, opened 2019-09-06
[Hotfix] Revert "Click on "show more" should not scroll page to top (#7420)"

joelverhagen:
This reverts commit 8c5b4e8836f57b050dff8ef98d458cb82fb4379a.

This causes the documentation of more than 10 lines to stay collapsed even if you click "Show more"

recording

/cc @304NotModified, I'll look into the root cause later. For now we are reverting it to unblock our deployment.


304NotModified:
Oops sorry for that.

I guess `e.PreventDefault()` is a safer fix.

PR #7521, opened 2019-09-07
Fix timezone issues

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/2499

Sorry @xavierdecoster. Should we move our CIs to CET 😄? Seems like the negative offsets from the prime meridian are a great edge case for our tests.


xavierdecoster:
Thanks! UTC all the way! ;) CET would introduce the reverse effect.

PR #7523, opened 2019-09-09
Add codeowners for the NuGetGallery repo to get reviews requested automatically

nkolev92:

* Scott was a fan of our code owners file on client side. 

https://github.com/NuGet/NuGet.Client/blob/dev/.github/CODEOWNERS

I think server gallery should have it as well :) 

Example: https://github.com/NuGet/NuGet.Client/pull/3053
Look at the key near most of the names in the reviewers :) 

//cc @scottbommarito


joelverhagen:
Cool!

Issue #7536, opened 2019-09-10
Error 500 during posting and listing 


psolovie:
NuGet.org returns error 500 during posting and listing via api and nuget.org UI. Error is spontaneous and usually can be fixed by a retry of the action.

This behavior breaks continues delivery automation.

![image](https://user-images.githubusercontent.com/55137163/64606079-1aed6500-d3ce-11e9-96ec-12305096387f.png)

loic-sharma:
Thank you for reporting this, we will investigate this issue.

loic-sharma:
Do you have a package that reproduces this issue? I believe this is because we're returning a HTTP Status string that's [longer than 512 characters](https://docs.microsoft.com/en-us/dotnet/api/system.web.httpresponse.statusdescription?redirectedfrom=MSDN&view=netframework-4.8#exceptions).

psolovie:
Hello, stable reproduce with listing for package https://www.nuget.org/packages/intelmkl.devel.linux-x64/2019.5.281/Manage

psolovie:
When I make api call to list the package I get an error 'An unlisted package should never be latest or latest stable'

psolovie:
Error is related to block of code 

if (!package.Listed && (package.IsLatestStable || package.IsLatest || package.IsLatestSemVer2 || package.IsLatestStableSemVer2))
            {
                throw new InvalidOperationException("An unlisted package should never be latest or latest stable!");
            }

in https://github.com/NuGet/NuGetGallery/blob/e3ae678fd8bc9f1065a17ccfb0ab3be82328bf55/src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs

I have an idea what's going on: for some reason after package version  https://www.nuget.org/packages/intelmkl.devel.linux-x64/2019.5.281/Manage was unlisted, current version for package https://www.nuget.org/packages/intelmkl.devel.linux-x64 is still 2019.5.281

so we have:

1. Unlist operation for some reason brake data consistency for package - does not update latest version property - that's the root cause of the bug
2. api returns a valid error
3. UI crashes because of message length limitations

I ask you to fix data issue ASAP to allow me to mark package version  https://www.nuget.org/packages/intelmkl.devel.linux-x64/2019.5.281/Manage as listed. That blocks project deliverable

loic-sharma:
Your package was indeed in a bad state. I went ahead and fixed that up, you should be able to list your package now. Please let me know if you run into any additional issues!

[...]

Issue #5818, opened 2018-04-18
Design an error details protocol for NuGet HTTP client-server communication


joelverhagen:
## Today

Today, we have a very simplistic approach for error messages in the NuGet HTTP API. There are two facets today:

1. HTTP status code - this is the only documented indicator of HTTP failure. For example, in the [push and delete resource](https://docs.microsoft.com/en-us/nuget/api/package-publish-resource#response) only `201`, `202`, `400`, and `409` are document indications of failures.

1. Reason phrase, a.k.a status description. This is an _undocumented_ way that some NuGet servers (such as [NuGetGallery](https://github.com/NuGet/NuGetGallery/blob/67373eb24d14207969ecb80f0c9c20a8cc17f7fd/src/NuGetGallery/Controllers/ApiController.cs#L380)) pass an error message to the user. The official NuGet client depends on the implementation of `HttpResponseMessage.EnsureSuccessStatusCode` which includes both the status code and the reason phrase in an exception message.

For example, this clearly wrong push command to NuGet.org yields this client output:

```
> nuget.exe push Package.1.0.0.nupkg -source https://api.nuget.org/v3/index.json -apikey "foo"
```
```
Pushing Package.1.0.0.nupkg to 'https://www.nuget.org/api/v2/package'...
  PUT https://www.nuget.org/api/v2/package/
  Forbidden https://www.nuget.org/api/v2/package/ 500ms
Response status code does not indicate success: 403 (The specified API key is invalid, has expired, or does not have permission to access the specified package.).
```

The client output is pretty good. It states the problem and allows the user to self resolve.

## We can do better

1. The reason phrase thing is undocumented. The API docs only talk about status codes.

1. The reason phrase has a maximum length. I think it's related to the [maximum header size](https://stackoverflow.com/a/686243), which varies per client and server.

1. The reason phrase is just a string. We don't have a way to return an error code that provides anything more granular than the HTTP status code. We have good error codes [on the client side](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings), couldn't we have the same for server communications?

1. It's unclear about how this works with localization.

1. And most importantly, reason phrases has been [removed from (or made nullable?) HTTP 2.0](https://github.com/http2/http2-spec/issues/202). Thanks for the heads up, @satamas, https://github.com/NuGet/Home/issues/6844.

For this reason, we should design a schema for an error response body. For example,

```json
{
    "statusCode": 403,
    "errorCode": "NU4031",
[...]

PR #7540, opened 2019-09-10
Mitigate "Internal Server Error" on package uploads

loic-sharma:
`HttpStatusCodeResult` throws if the `StatusDescription` property is longer than 512 characters (see [this documentation](https://docs.microsoft.com/en-us/dotnet/api/system.web.httpresponse.statusdescription?redirectedfrom=MSDN&view=netframework-4.8#exceptions)), which results in a HTTP 500 response. I wasn't able to determine which code path affected the user as a few suspicious paths append an exception message to the description.

This change mitigates the issue by truncating HTTP Status Descriptions if they are longer than 512 characters.

Addresses https://github.com/NuGet/NuGetGallery/issues/7536


joelverhagen:
We need https://github.com/NuGet/NuGetGallery/issues/5818 😢

Issue #5216, opened 2017-12-21
Show dotnet install command for template packages


nojaf:
It would be nice if the site shows the `dotnet new -i PackageName` when the nuget package is a dotnet cli template, for example [Giraffe](https://www.nuget.org/packages/giraffe-template/).

![image](https://user-images.githubusercontent.com/2621499/34246832-7d9912ea-e630-11e7-8121-eb7f024bc3d7.png)

joelverhagen:
@am11 this change is deployed to nuget.org. Thanks again for your contribution.

am11:
Thanks for the production deployment! :)

Just tried out this template and noticed that CLI does not list the installed version of templates, so I have opened an issue: https://github.com/dotnet/cli/issues/12805.

PR #7548, opened 2019-09-12
Render dotnet-new-install for template packages

am11:
With this change, the donet CLI template packages will show the `dotnet-new-install` command.

image

Addresses https://github.com/NuGet/NuGetGallery/issues/5216

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -28,13 +28,27 @@
     {
         packageManagers = new PackageManagerViewModel[]
         {
-            new PackageManagerViewModel(".NET CLI")
+            new PackageManagerViewModel(".NET CLI Tool")
             {
                 Id = "dotnet-cli",
                 CommandPrefix = "> ",
                 InstallPackageCommand = string.Format("dotnet tool install --global {0} --version {1}", Model.Id, Model.Version),
                 AlertLevel = AlertLevel.Info,
-                AlertMessage = string.Format("This package contains a <a href=\"{0}\">.NET Core Global Tool</a> you can call from the shell/command line.", "https://aka.ms/global-tools"),
+                AlertMessage = "This package contains a <a href='https://aka.ms/global-tools'>.NET Core Global Tool</a> you can call from the shell/command line.",
+            }
+        };
+    }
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -28,13 +28,27 @@
     {
         packageManagers = new PackageManagerViewModel[]
         {
-            new PackageManagerViewModel(".NET CLI")
+            new PackageManagerViewModel(".NET CLI Tool")
```
joelverhagen: /cc @nkolev92

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -28,13 +28,27 @@
...
+    {
+        packageManagers = new PackageManagerViewModel[]
+        {
+            new PackageManagerViewModel(".NET CLI Template")
+            {
+                Id = "dotnet-cli",
+                CommandPrefix = "> ",
+                InstallPackageCommand = string.Format("dotnet new --install {0}::{1}", Model.Id, Model.Version),
+                AlertLevel = AlertLevel.Info,
+                AlertMessage = "This package contains a <a href='https://aka.ms/dotnet-new'>.NET Core Template Package</a> you can call from the shell/command line.",
```
joelverhagen: I think a better link may be: 
https://docs.microsoft.com/en-us/dotnet/core/tools/custom-templates

Thoughts @karann-msft, @anangaur ?

anangaur: Seems like the right place but better to get the URL from the PM of the product. @KathleenDollard 😄

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
nkolev92: I don't have a preference either way. 

I'd let @anangaur, @karann-msft , @kathleenDollard chime in.

nkolev92: I think leaving it .NET CLI is inline with the rest of the tabs though. 
The tab represents the tool not the action.

am11: Fix pushed in next update.

PR #7554, opened 2019-09-13
Secure database access on AutocompleteDatabasePackage*Query classes.

dannyjdev:
To be compliant with our processes I checked for any database access that can be compromised.

* AutocompleteDatabase* classes were updated to use IReadOnlyEntityRepository.
* dynamic sql queries removed from those classes.

Address: https://github.com/NuGet/Engineering/issues/2713


src/NuGetGallery/Queries/AutocompleteDatabasePackageIdsQuery.cs
```diff
@@ -1,66 +1,75 @@
...
         {
-            // Create SQL filter on SemVerLevel
-            // By default, we filter out SemVer v2.0.0 package versions.
-            var semVerLevelSqlFilter = "p.[SemVerLevelKey] IS NULL";
-            if (!string.IsNullOrEmpty(semVerLevel))
+            var query = _packageRepository.GetAll()
+                .Include(p => p.PackageRegistration);
+            
+            // SemVerLevel filter
+            if (SemVerLevelKey.ForSemVerLevel(semVerLevel) == SemVerLevelKey.SemVer2)
```
cristinamanum: What happens if semVerLevel is null? Should null check first?

src/NuGetGallery/Queries/AutocompleteDatabasePackageIdsQuery.cs
```diff
@@ -1,66 +1,75 @@
...
+                query = query.Where(p => !p.IsPrerelease);
             }
 
-            var sql = string.Format(CultureInfo.InvariantCulture, _partialIdSqlFormat, semVerLevelSqlFilter, prereleaseFilter);
+            var ids = new List<string>();
+
+            // filters added for partialId
+            if (!string.IsNullOrWhiteSpace(partialId))
+            {
+                ids = query.Where(p => p.PackageRegistration.Id.StartsWith(partialId))
```
cristinamanum: Will the changes have any perf impact (positive or negative)?

src/NuGetGallery/Queries/AutocompleteDatabasePackageIdsQuery.cs
```diff
@@ -1,66 +1,75 @@
...
+                    .GroupBy(p => p.PackageRegistration.Id)
+                    .Select(group => group.Key)
+                    .OrderBy(id => id)
+                    .Take(30)
```
cristinamanum: Should 30 be a constant?

tests/NuGetGallery.Facts/Queries/AutocompleteDatabasePackageIdsQueryFacts.cs
```diff
@@ -0,0 +1,307 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Moq;
+using NuGet.Services.Entities;
+using Xunit;
+
+namespace NuGetGallery.Queries
+{
+    public class AutocompleteDatabasePackageIdsQueryFacts
```
cristinamanum: Nice tests!

src/NuGetGallery/Queries/AutocompleteDatabasePackageVersionsQuery.cs
```diff
@@ -32,26 +28,33 @@ public AutocompleteDatabasePackageVersionsQuery(IEntitiesContext entities)
...
-                prereleaseFilter = "AND p.IsPrerelease = 0";
+                query = query.Where(p => !p.IsPrerelease);
             }
-            
-            return RunSqlQuery(string.Format(CultureInfo.InvariantCulture, _sqlFormat, semVerLevelSqlFilter, prereleaseFilter), id);
+
+            var versions = query.Select(p => p.Version).ToList();
+
+            // return the result of the query
+            return Task.FromResult<IReadOnlyList<string>>(versions);
```
scottbommarito: Do you need the `<IReadOnlyList<string>>`? I would assume it casts `versions`, which is a `List<string>`, automatically.

src/NuGetGallery/Queries/AutocompleteDatabasePackageIdsQuery.cs
scottbommarito: They're only used by local gallery, although they probably do a have a small perf impact as EF is slower than raw SQL.

src/NuGetGallery/Queries/AutocompleteDatabasePackageIdsQuery.cs
joelverhagen: Can you use `SemVerLevelKey.IsPackageCompliantWithSemVerLevelPredicate`? This is what non-hijacked V2 OData does when querying the DB.

dannyjdev: In the case semVerLevel is null, the method ForSemVerLevel will return Unknown and that comparison will be false.

src/NuGetGallery/Queries/AutocompleteDatabasePackageVersionsQuery.cs
dannyjdev: `ToList()` returns `List<string>,` since `Task.FromResult()` expects to have `Task<IReadOnlyList<string>>` and not `Task<List<string>>` this needs to be explicit on the method.

src/NuGetGallery/Queries/AutocompleteDatabasePackageIdsQuery.cs
dannyjdev: @joelverhagen  I've changed the SemVerLevel filter to use that method, in the case of PackageIdsQuery the behavior is a little different (not a problem with PackageVersionsQuery), but I think it should have the same logic as the non-hijacked V2 Odata.

src/NuGetGallery/Queries/AutocompleteDatabasePackageIdsQuery.cs
```diff
@@ -1,66 +1,75 @@
...
+                    .ToList();
+            }
+            else
+            {
+                ids = query.GroupBy(p => p.PackageRegistration.Id)
```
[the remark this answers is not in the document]
dannyjdev: Updated.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -994,7 +994,7 @@ protected internal virtual Stream ReadPackageFromRequest()
         {
             return new JsonResult
             {
-                Data = (await _autocompletePackageIdsQuery.Execute(partialId, includePrerelease, semVerLevel)).ToArray(),
+                Data = (await _autocompletePackageIdsQuery.Execute(partialId, includePrerelease, semVerLevel)),
```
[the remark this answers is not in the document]
dannyjdev: Updated.

src/NuGetGallery/Queries/AutocompleteDatabasePackageIdsQuery.cs
dannyjdev: Updated.

joelverhagen:
This should have gone into dev right? This should not need to be hotfix speed which is the case where we go to master directly.

Issue #7478, opened 2019-08-23
[Icons] Gallery should warn when IconUrl is used


agr:
We need to get a warning when package specifies `iconUrl` but no `icon`.

> &lt;iconUrl&gt; element will be deprecated, please consider switching to specifying the icon in the package. [Learn more](https://aka.ms/deprecateIconUrl).

PR #7556, opened 2019-09-13
IconUrl deprecation message.

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/7478

Browser:
iconUrlDepr

Command line:
iconUrlDeprCL

### DIFF
--- a/src/NuGetGallery/Services/IconUrlDeprecationValidationMessage.cs
+++ b/src/NuGetGallery/Services/IconUrlDeprecationValidationMessage.cs
@@ -3,9 +3,6 @@
 
 namespace NuGetGallery
 {
-    /// <summary>
-    /// IconUrl element deprecation message.
-    /// </summary>
     public class IconUrlDeprecationValidationMessage : IValidationMessage
     {
         public string PlainTextMessage => $"{Strings.UploadPackage_IconUrlDeprecated} https://aka.ms/deprecateIconUrl";
--- a/src/NuGetGallery/Services/IconUrlDeprecationValidationMessage.cs
+++ b/src/NuGetGallery/Services/IconUrlDeprecationValidationMessage.cs
@@ -1,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// IconUrl element deprecation message.
+    /// </summary>
+    public class IconUrlDeprecationValidationMessage : IValidationMessage
+    {
+        public string PlainTextMessage => $"{Strings.UploadPackage_IconUrlDeprecated} https://aka.ms/deprecateIconUrl";
+
+        public bool HasRawHtmlRepresentation => true;
+
+        public string RawHtmlMessage => $"{Strings.UploadPackage_IconUrlDeprecated.Replace("<", "&lt;").Replace(">", "&gt;")} <a href=\"https://aka.ms/deprecateIconUrl\">{Strings.UploadPackage_LearnMore}</a>.";
+    }
+}
--- a/src/NuGetGallery/Strings.resx
+++ b/src/NuGetGallery/Strings.resx
@@ -1006,7 +1006,7 @@ The {1} Team</value>
     <value>Unknown error!</value>
   </data>
   <data name="UploadPackage_DeprecatingLicenseUrl" xml:space="preserve">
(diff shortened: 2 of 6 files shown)


src/NuGetGallery/Services/IconUrlDeprecationValidationMessage.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// IconUrl element deprecation message.
```
scottbommarito: nit: Not a very useful comment, as it just restates the name of the class. I would prefer either getting rid of this comment or saying more about IconUrl and its deprecation.

tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
```diff
@@ -1141,6 +1141,53 @@ public async Task AcceptsPackagesWithEmbeddedIconForFlightedUsers()
...
+                var result = await _target.ValidateBeforeGeneratePackageAsync(
+                    _nuGetPackage.Object,
+                    GetPackageMetadata(_nuGetPackage),
+                    _currentUser);
+
+                Assert.Equal(PackageValidationResultType.Accepted, result.Type);
+                Assert.Null(result.Message);
+                var warning = Assert.Single(result.Warnings);
+                Assert.IsType<IconUrlDeprecationValidationMessage>(warning);
+                Assert.StartsWith("<iconUrl> element will be deprecated, please consider switching to specifying the icon in the package.", warning.PlainTextMessage);
```
zhhyu: It looks easy to break this test if we use the hard coded string. Maybe directly using the "Strings.UploadPackage_IconUrlDeprecated" is better for future maintenances. 
I find some other examples in this facts.

agr: That's the point of this check. To break if text changes. Using `Strings.UploadPackage_IconUrlDeprecated` directly will hide such changes.

karann-msft:
Can I see a mock with all possible warnings? I am afraid we may have one link for every warning and then another read more at the end which might be few too many warnings. We can choose to not fix this right now in which case the warning message looks fine. I'd just make sure it aligns with what I'd see on nuget pack.

agr:
@karann-msft , this is the second *warning* message with the link (the other one is about license URL deprecation). All other warnings don't have links. And all other messages with link are errors (which we don't display at the same time as warnings).

src/NuGetGallery/Strings.resx
```diff
@@ -1150,4 +1150,7 @@ The {1} Team</value>
   <data name="SiteAdminNotLoggedInWithRequiredTenant" xml:space="preserve">
     <value>The site admins are required to sign in with the '{0}' tenant only.</value>
   </data>
+  <data name="UploadPackage_IconUrlDeprecated" xml:space="preserve">
+    <value>&lt;iconUrl&gt; element will be deprecated, please consider switching to specifying the icon in the package.</value>
```
loic-sharma: Considering that we're warning on icon URLs already, isn't it already deprecated? What do you think of this messaging?

```suggestion
    <value>The &lt;iconUrl&gt; element is deprecated, please use the icon element instead.</value>
```

This is message doesn't mirror the client's message, but I would prefer a clearer message if possible.

karann-msft:
Let's update both messages to say "is deprecated". Rest looks good for now.

src/NuGetGallery/Strings.resx
agr: The message that client currently produces is:
> WARNING: NU5048: The 'PackageIconUrl'/'iconUrl' element is deprecated. Consider using the 'PackageIcon'/'icon' element instead. Learn more at https://aka.ms/deprecateIconUrl

And for `licenseUrl`:
> WARNING: NU5125: The 'licenseUrl' element will be deprecated. Consider using the 'license' element instead.

agr:
Like that?
image

tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
dominoFire: > Like that?
> image
> 
> Saying that element is deprecated when we didn't release a "stable" version of the client supporting it sounds a bit weird though.

@agr Do you need to update your tests?

loic-sharma:
> Saying that element is deprecated when we didn't release a "stable" version of the client supporting it sounds a bit weird though.

Does the client in .NET Core 3 warn? If so, that ships on the 23rd

PR #7561, opened 2019-09-16
Rewrite http images to https.

ryuyu:
Rewrites all http images to https at display time.
Does not affect text in the actual md file; this is only a display time rewrite.

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
@@ -54,6 +54,10 @@ namespace NuGetGallery
                     {
                         returnUri = uri.ToHttps();
                     }
+                    else
+                    {
+                        returnUri = uri;
+                    }
                 }
                 else if (uri.IsHttpsProtocol() || uri.IsHttpProtocol())
                 {
--- a/src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
@@ -41,16 +41,19 @@ namespace NuGetGallery
         /// If the input uri is https => leave as is
         /// If the input uri is not a valid uri or not http/https => return false
         /// </summary>
-        public static bool TryPrepareUrlForRendering(string uriString, out string readyUriString)
+        public static bool TryPrepareUrlForRendering(string uriString, out string readyUriString, bool rewriteAllHttp = false)
         {
             Uri returnUri = null;
             readyUriString = null;
 
             if (Uri.TryCreate(uriString, UriKind.Absolute, out var uri))
             {
-                if (uri.IsHttpProtocol() && uri.IsDomainWithHttpsSupport())
+                if (uri.IsHttpProtocol())
                 {
-                    returnUri = uri.ToHttps();
+                    if (rewriteAllHttp || uri.IsDomainWithHttpsSupport())
+                    {
+                        returnUri = uri.ToHttps();
+                    }
                 }
                 else if (uri.IsHttpsProtocol() || uri.IsHttpProtocol())
                 {
--- a/src/NuGetGallery/Services/ReadMeService.cs
+++ b/src/NuGetGallery/Services/ReadMeService.cs
@@ -210,16 +210,31 @@ namespace NuGetGallery
                     }
 
                     var inline = node.Inline;
-                    if (inline != null && inline.Tag == InlineTag.Link)
+                    if (inline != null)
                     {
-                        // Allow only http or https links in markdown. Transform link to https for known domains.
-                        if (!PackageHelper.TryPrepareUrlForRendering(inline.TargetUrl, out string readyUriString))
+                        if (inline.Tag == InlineTag.Link)
                         {
-                            inline.TargetUrl = string.Empty;
+                            // Allow only http or https links in markdown. Transform link to https for known domains.
+                            if (!PackageHelper.TryPrepareUrlForRendering(inline.TargetUrl, out string readyUriString))
+                            {
+                                inline.TargetUrl = string.Empty;
+                            }
+                            else
+                            {
+                                inline.TargetUrl = readyUriString;
+                            }
                         }
-                        else
+
+                        else if (inline.Tag == InlineTag.Image)
                         {
-                            inline.TargetUrl = readyUriString;
+                            if (!PackageHelper.TryPrepareUrlForRendering(inline.TargetUrl, out string readyUriString, rewriteAllHttp: true))
+                            {
+                                inline.TargetUrl = string.Empty;
+                            }
+                            else
+                            {
+                                inline.TargetUrl = readyUriString;
+                            }
                         }
                     }
                 }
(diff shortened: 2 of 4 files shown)


src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -210,16 +210,31 @@ internal static string GetReadMeHtml(string readMeMd)
...
-                    if (inline != null && inline.Tag == InlineTag.Link)
+                    if (inline != null)
                     {
-                        // Allow only http or https links in markdown. Transform link to https for known domains.
-                        if (!PackageHelper.TryPrepareUrlForRendering(inline.TargetUrl, out string readyUriString))
+                        if (inline.Tag == InlineTag.Link)
                         {
-                            inline.TargetUrl = string.Empty;
+                            // Allow only http or https links in markdown. Transform link to https for known domains.
+                            if (!PackageHelper.TryPrepareUrlForRendering(inline.TargetUrl, out string readyUriString))
```
joelverhagen: i think it may be clearer with no default and explicit `rewriteAllHttp: false` here

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -210,16 +210,31 @@ internal static string GetReadMeHtml(string readMeMd)
                     }
 
                     var inline = node.Inline;
-                    if (inline != null && inline.Tag == InlineTag.Link)
+                    if (inline != null)
```
joelverhagen: Are there unit tests for this path? This definitely seems like something we should test

src/NuGetGallery/Services/ReadMeService.cs
ryuyu: I wanted to leave the old behaviour as the default in order to avoid impacting other places that used this extension method, but we can adjust it if you think this is worth it.

src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
```diff
@@ -41,16 +41,19 @@ public static bool ShouldRenderUrl(string url, bool secureOnly = false)
...
-                if (uri.IsHttpProtocol() && uri.IsDomainWithHttpsSupport())
+                if (uri.IsHttpProtocol())
                 {
-                    returnUri = uri.ToHttps();
+                    if (rewriteAllHttp || uri.IsDomainWithHttpsSupport())
+                    {
+                        returnUri = uri.ToHttps();
+                    }
                 }
                 else if (uri.IsHttpsProtocol() || uri.IsHttpProtocol())
```
skofman1: there should be UTs for this..

skofman1: the logic is broken here: if uri is http but from a domain that doesn't support https, we will return false.

UTs are important :)

scottbommarito: `uri.IsHttpProtocol()` will always be false if the `if` statement above is not executed. You can safely remove it.

src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
```diff
@@ -41,16 +41,19 @@ public static bool ShouldRenderUrl(string url, bool secureOnly = false)
...
-        public static bool TryPrepareUrlForRendering(string uriString, out string readyUriString)
+        public static bool TryPrepareUrlForRendering(string uriString, out string readyUriString, bool rewriteAllHttp = false)
         {
             Uri returnUri = null;
             readyUriString = null;
 
             if (Uri.TryCreate(uriString, UriKind.Absolute, out var uri))
             {
-                if (uri.IsHttpProtocol() && uri.IsDomainWithHttpsSupport())
+                if (uri.IsHttpProtocol())
```
scottbommarito: I assume what you actually want here (to fix the bug mentioned by @skofman1) is:
```
if (uri.IsHttpProtocol() && (rewriteAllHttp || uri.IsDomainWithHttpsSupport())
{
    returnUri = uri.ToHttps();
}
else if (uri.IsHttpsProtocol() || uri.IsHttpProtocol())
{
    returnUri = uri;
}
```

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -210,16 +210,31 @@ internal static string GetReadMeHtml(string readMeMd)
...
+                            {
+                                inline.TargetUrl = string.Empty;
+                            }
+                            else
+                            {
+                                inline.TargetUrl = readyUriString;
+                            }
                         }
-                        else
+
```
scottbommarito: extra line

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -210,16 +210,31 @@ internal static string GetReadMeHtml(string readMeMd)
...
+                        else if (inline.Tag == InlineTag.Image)
                         {
-                            inline.TargetUrl = readyUriString;
+                            if (!PackageHelper.TryPrepareUrlForRendering(inline.TargetUrl, out string readyUriString, rewriteAllHttp: true))
```
scottbommarito: The code between `InlineTag.Link` and `InlineTag.Image` are identical, except `InlineTag.Image` passes `rewriteAllHttp: true`. I would suggest the following rewrite:
```
if (inline != null && (inline.Tag == InlineTag.Link || inline.Tag == InlineTag.Image))
{
    var shouldRewriteAllHttp = inline.Tag == InlineTag.Image;
    inline.TargetUrl = PackageHelper.TryPrepareUrlForRendering(inline.TargetUrl, out var readyUriString, shouldRewriteAllHttp) 
        ? readyUriString : string.Empty;
}
```

Also, you can replace `out string readyUriString` with `out var readyUriString`.

src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
scottbommarito: (upon reading the rest of the code, this does not seem like what you want here)

src/NuGetGallery/Services/ReadMeService.cs
ryuyu: This is much harder to read than keeping the two cases complete separate. I understand not wanting duplication, but I think it's warranted in this case for readability sake.

src/NuGetGallery.Services/PackageManagement/PackageHelper.cs
ryuyu: There were UTs for this. They caught this! It has since been fixed and i have added more cases around this specifically.

src/NuGetGallery/Services/ReadMeService.cs
scottbommarito: I disagree, but it's up to you.

skofman1:
>                                 block.Heading = new HeadingData(level);

the tests are here: [https://github.com/NuGet/NuGetGallery/blob/995709555c8cbcba05925756b27eefbcff0c8242/tests/NuGetGallery.Facts/Services/ReadMeServiceFacts.cs#L253](https://github.com/NuGet/NuGetGallery/blob/995709555c8cbcba05925756b27eefbcff0c8242/tests/NuGetGallery.Facts/Services/ReadMeServiceFacts.cs#L253)

---
Refers to: src/NuGetGallery/Services/ReadMeService.cs:195 in 1e4eb20. [](commit_id = 1e4eb20d7a8115359a4b17b216b96c2dd7367dab, deletion_comment = False)

PR #7566, opened 2019-09-17
Do not throw if trying to list an unlisted package that is marked latest

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/7536

Relisting a package properly resets the latest state of the package, so any packages that are unlisted and marked latest will be fixed. By throwing here, we are making it so that users cannot fix the data issue they have.

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
@@ -54,11 +54,6 @@ namespace NuGetGallery
                 throw new InvalidOperationException("A package that failed validation should never be listed!");
             }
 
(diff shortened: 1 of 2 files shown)


src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
```diff
@@ -172,10 +167,5 @@ private async Task UpdatePackagesInBulkAsync(IReadOnlyList<int> packageKeys)
                     $"Updated {result} packages instead of {expectedResult}.");
             }
         }
-
-        private bool ShouldUpdateIsLatestForPackageWhenUnlisting(Package package)
```
scottbommarito: Noticed this method was only used here so I've removed it. I think it's a relic from my refactoring of this service from the deprecation work.

PR #7580, opened 2019-09-25
Warn owners when ReadMe contains images that would be rewritten

ryuyu:
We introduced a change that causes http images in readme files to be rewritten using https.

This change alerts an owner when their documentation contains an image that causes this to happen, both at edit time, and on package details page.

Introduces compound return type from readme service.

### DIFF
--- a/src/NuGetGallery/Services/ReadMeService.cs
+++ b/src/NuGetGallery/Services/ReadMeService.cs
@@ -248,8 +248,8 @@ namespace NuGetGallery
                             }
                             else
                             {
+                                output.ImagesRewritten = output.ImagesRewritten || (inline.TargetUrl != readyUriString);
                                 inline.TargetUrl = readyUriString;
-                                output.ImagesRewritten = true;
                             }
                         }
                     }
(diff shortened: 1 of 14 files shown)


src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -233,6 +248,7 @@ internal static string GetReadMeHtml(string readMeMd)
                             }
                             else
                             {
+                                output.ImagesRewritten = output.ImagesRewritten || (inline.TargetUrl != readyUriString);
```
skofman1: I don't see UTs for this

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -222,6 +236,7 @@ internal static string GetReadMeHtml(string readMeMd)
                             else
                             {
                                 inline.TargetUrl = readyUriString;
+                                output.LinksRewritten = true;
```
skofman1: Why is this needed?

ryuyu: It isn't strictly necessary. Mostly was added as so we had information about all the rewrites we do on the readmes in case we wanted to surface other information.

ryuyu: After thinking about this for a bit, i have removed it.

src/NuGetGallery/Services/ReadMeService.cs
ryuyu: Good call. Updated some of the existing tests to check this part of the result as well.

PR #7595, opened 2019-10-01
Add GalleryTools command to delete unconfirmed accounts

scottbommarito:
https://github.com/nuget/engineering/issues/2707

I felt writing a tool to use the `DeleteAccountService` to delete these accounts was safer than writing a large `DELETE FROM Users` query, as the `DeleteAccountService` properly removes all credentials and other data and doesn't rely on foreign key deletion cascades.


src/GalleryTools/Commands/DeleteUnconfirmedAccountsCommand.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        private static async Task<int> ExecuteAsync()
+        {
+            var builder = new ContainerBuilder();
+            builder
+                .RegisterAssemblyModules(
+                    typeof(DefaultDependenciesModule).Assembly);
+
+            builder
+                .RegisterType<NullAuditingService>()
```
joelverhagen: Shouldn't we audit this?

joelverhagen: What does the AccountDeleter do in this regard?

joelverhagen:
Just curious, can we extend the AccountDeleter to do this?

src/GalleryTools/Commands/DeleteUnconfirmedAccountsCommand.cs
ryuyu: I'm fairly certain that deletion of unconfirmed account is still audited. AccountDeleter does whatever gallery did in the past. That being said, I'm not 100% sure of the value of taking the effort to have auditing here if this is going to be a one time pass and only affect unconfirmed accounts (which have basically no permissions anyway and can no longer be created).

drewgillies:
This looks to have stalled. @scottbommarito and @joelverhagen --has this reached an impasse and should it be closed? Or is there still something we can do in this space?

joelverhagen:
Let's close it. We'll take this later if it needs attention.

Issue #7601, opened 2019-10-02
Extensibility: How to load a custom storage implementation


babakontheweb:
### Is your feature request related to a problem? Please describe.
There are a number of open issues for users trying to deploy NuGet Gallery on AWS with S3 storage. The hardest part of keeping custom implementations up to date with primary repo is that we would need to add to enum StorageType (in this case, adding AwsS3 as a value).

### Describe the solution you'd like
My proposal is to replace the enum with a static class of constants by the same name. This would allow others to define their own constant values for different storage types while keeping the primary code base intact.

zhhyu:
Thank you for your suggestions!

joelverhagen:
I've merged this into our `dev` branch. I will close the issue next time we merge to `master`.

babakontheweb:
@joelverhagen I have a follow up question. I'm trying to refactor my AWS/S3 related implementations of `IFileReference` and `IFileStorageService` into a standalone Nuget package. Is there a package I can reference to be able to reference these interfaces instead of having to add them to a fork of this repo?

joelverhagen:
>  Is there a package I can reference to be able to reference these interfaces instead of having to add them to a fork of this repo?

So you want to have something like "babakontheweb.NuGetGallery.S3" package that gets loaded by gallery for it's storage implementation? How would you reference this package? That would itself require modifications to NuGetGallery.csproj right?

We have a "shims" infrastructure that is used for loading DLLs at runtime via MEF. nuget.org uses this for private implementations of `IAuditingService` and `ICookieComplianceService`. With some refactoring you could perhaps use this infrastructure to load `IFileStorageService` from there as well. See:
https://github.com/NuGet/NuGetGallery/blob/5c21a41506ea6e088eb4742b8b58047e5b94b819/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L1232-L1280

joelverhagen:
@babakontheweb, this is merged to master. I would close this issue but it seems like you have more questions related to extensibility of NuGetGallery. I will leave this issue open for a couple more days regarding my response https://github.com/NuGet/NuGetGallery/issues/7601#issuecomment-539725169.

loic-sharma:
@babakontheweb Out of curiosity, have you looked at alternative NuGet server implementations? There's a few that support AWS S3. Were they missing features you needed?

babakontheweb:
Thank you @loic-sharma and @joelverhagen I will review your comments but we can consider this issue closed.

PR #7602, opened 2019-10-02
Replacing enum StorageType with static class of constants by the same…

babakontheweb:
* Replacing enum StorageType with static class of constants by the same name.

Addresses https://github.com/NuGet/NuGetGallery/issues/7601


joelverhagen:
Hey @babakontheweb, thanks for the contribution!

Can you help me understand how this works end-to-end? My understanding is that you would need to update the default dependencies module's `switch` on this configuration value anyways to introduce another storage implementation so you would need to modify existing files anyhow. Is this what you are doing or have you figured out another extensibility point?

https://github.com/NuGet/NuGetGallery/blob/d5699f170ca2c28110fae4b82420e995ca974ea7/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L427-L438

If this makes hacking on top of NuGetGallery easier I'm all for it! I'm running the change through the CI but I'd like to understand how you have things working today.

Also, could you change your base branch to `dev`? We do all of our non-hotfix development on that branch and later merge it to `master` when we are ready to ship.

babakontheweb:
Thank you for the feedback @joelverhagen; I have pulled the `dev` branch into my feature branch and verified all tests still passing. 

We are currently running NuGet Gallery internally hosted in AWS using S3 as package storage. Our dependencies are injected following the same pattern in the code base but we had to modify `enum StorageType` to make it work.

```
            switch (configuration.Current.StorageType)
            {
                case StorageType.FileSystem:
                case StorageType.NotSpecified:
                    ConfigureForLocalFileSystem(builder, configuration);
                    defaultAuditingService = GetAuditingServiceForLocalFileSystem(configuration);
                    break;
                case StorageType.AzureStorage:
                    ConfigureForAzureStorage(builder, configuration, telemetryService);
                    defaultAuditingService = GetAuditingServiceForAzureStorage(builder, configuration);
                    break;
                case StorageType.S3:
                    ConfigureForAwsS3Storage(builder, configuration);
                    break;
            }
```
The constants would allow any implementing team to use whatever new identifier that makes sense for them and reduces the changes of any down stream changes in Nuget/NugetGallery from causing conflicts with their changes.

Issue #7324, opened 2019-07-09
Cannot edit invalid pre-filled e-mail address on Register Microsoft account screen


tombogle:
### Description
Attempting to log in using Microsoft account. After entering credentials, it goes to https://www.nuget.org/account/register, where it asks me to provide a username. The email field is already filled in and is not editable, but since it is my Microsoft account name and is not a valid e-mail address, I'm stuck. Attempting to go forward gives the error message: This doesn't appear to be a valid email address.

### To Reproduce
1. Go to this URL: https://www.nuget.org/users/account/LogOn?returnUrl=%2F
2. Click Sign in with Microsoft button
3. Enter Skype id (I've also tried logging out of skype.com, and coming back to nuget using the e-mail address associated with the Skype/Microsoft account - same results.)
4. When prompted, I enter my password. That opens this URL: https://www.nuget.org/users/account/authenticate/return?ReturnUrl=%2f&error=&errorDescription=
5. As shown in the screen shot, there is no way to edit the e-mail address. When I click Register there, I get the error: This doesn't appear to be a valid email address.

### Expected behavior
I either need nuget to correctly get the e-mail address associated with my Skype/Microsoft account or have it allow me to edit the e-mail address field manually.

### Screenshots
![image](https://user-images.githubusercontent.com/4229598/60890390-2df68400-a229-11e9-9448-123f5460e769.png)

shishirx34:
Hi @tombogle - it looks like the email address claim for your account doesn't actually return an email id but your skype id(?). Can you verify you have an email address linked to your Microsoft account(https://account.microsoft.com/profile/)? Perhaps, would it be possible for you to update your microsoft account(https://account.microsoft.com/profile/) and add email address as login mechanism?

agr:
OK, I managed to reproduce the issue and work around it.
Workaround:
1. Go to https://account.microsoft.com
2. Click on your profile picture in the top right corner and select "Sign out";
  * In my case, clicking that profile icon opened the account card that had my Skype name specified there, so I expect it would be the same for you;
3. Page reloads and your profile picture gets replaced with the generic "Sign in" icon;
4. Click the "sign in" icon;
5. Make sure you use your email (and not your Skype name) in the login process;
6. You should get back to your account page and clicking on the profile picture in the top right corner will open a card where it would have your email specified.
7. Go to https://www.nuget.org, click the "Sign in" button, followed by "Sign in with Microsoft";
[...]

PR #7606, opened 2019-10-06
[Hotfix] Fix AAD login issues for some tenants.

shishirx34:
The fix for the https://github.com/NuGet/NuGetGallery/issues/7324 regressed the behavior for certain AAD logins. For some AAD tenants, the `EmailClaim` is null. In such scenarios `preferred_username` claim seems to have been giving us the correct data, which is what we were using previously.

### DIFF
--- a/src/NuGetGallery.Services/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
+++ b/src/NuGetGallery.Services/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
@@ -188,7 +188,7 @@ namespace NuGetGallery.Authentication.Providers.AzureActiveDirectoryV2
             emailClaim = emailClaim ?? preferredUsernameClaim;
             if (emailClaim == null)
             {
-                throw new ArgumentException($"External Authentication is missing required claim: '{V2Claims.EmailAddress}'");
+                throw new ArgumentException($"External Authentication is missing required claims: '{V2Claims.EmailAddress}' and '{V2Claims.PreferredUsername}");
             }
 
             var acrClaim = claimsIdentity.FindFirst(V2Claims.ACR);
--- a/src/NuGetGallery.Services/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
+++ b/src/NuGetGallery.Services/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
@@ -184,6 +184,8 @@ namespace NuGetGallery.Authentication.Providers.AzureActiveDirectoryV2
 
             var nameClaim = claimsIdentity.FindFirst(V2Claims.Name);
             var emailClaim = claimsIdentity.FindFirst(V2Claims.EmailAddress);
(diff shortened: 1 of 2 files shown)


src/NuGetGallery.Services/Authentication/Providers/AzureActiveDirectoryV2/AzureActiveDirectoryV2Authenticator.cs
```diff
@@ -184,6 +184,8 @@ public override IdentityInformation GetIdentityInformation(ClaimsIdentity claims
 
             var nameClaim = claimsIdentity.FindFirst(V2Claims.Name);
             var emailClaim = claimsIdentity.FindFirst(V2Claims.EmailAddress);
+            var preferredUsernameClaim = claimsIdentity.FindFirst(V2Claims.PreferredUsername);
+            emailClaim = emailClaim ?? preferredUsernameClaim;
             if (emailClaim == null)
             {
                 throw new ArgumentException($"External Authentication is missing required claim: '{V2Claims.EmailAddress}'");
```
joelverhagen: Does the user see this `ArgumentException` message? Either way the message should mention both usable claim types.

shishirx34: user doesn't see these exceptions, it transpires as "Unknown error". I have made a note of improving this flow(to better reflect error messages to user). It will need bigger change. will add the claim in the exception message for now.

tests/NuGetGallery.Facts/Authentication/Providers/CommonAuth/AzureActiveDirectoryV2AuthenticatorFacts.cs
```diff
@@ -165,6 +165,26 @@ public void ThrowsForMissingEmailClaim()
...
+                    TestData.Name,
+                    TestData.PreferredUsername
+                });
+
+                // Act
+                var result = authenticator.GetIdentityInformation(claimsIdentity);
+
+                // Assert
+                Assert.NotNull(result);
+            }
```
joelverhagen: We should assert the email returned is the one from the PreferredUsername claim

tests/NuGetGallery.Facts/Authentication/Providers/CommonAuth/AzureActiveDirectoryV2AuthenticatorFacts.cs
```diff
@@ -220,6 +240,7 @@ private static class TestData
             public static Claim Name = new Claim(AzureActiveDirectoryV2Authenticator.V2Claims.Name, "bloog", ClaimValueTypes.String, Authority);
             public static Claim TenantId = new Claim(AzureActiveDirectoryV2Authenticator.V2Claims.TenantId, TEST_TENANT_ID, ClaimValueTypes.String, Authority);
             public static Claim Email = new Claim(AzureActiveDirectoryV2Authenticator.V2Claims.EmailAddress, "blarg@bloog.test", ClaimValueTypes.String, Authority);
+            public static Claim PreferredUsername = new Claim(AzureActiveDirectoryV2Authenticator.V2Claims.PreferredUsername, "blarg@bloog.test", ClaimValueTypes.String, Authority);
```
joelverhagen: There should be a test where this is a different value to assert the precedence (email is preferred over preferred_username)

Issue #7605, opened 2019-10-05
iconUrl warning appears for packages with no icon at all


joelverhagen:
### Describe the bug

For a package with no `<iconUrl>`, I get an `<iconUrl>` warning.

### To Reproduce

1. Make a package with no `<iconUrl>`.
2. Drop it in the gallery upload UI.

### Expected behavior
No warning.

### Screenshots

![image](https://user-images.githubusercontent.com/94054/66249103-a616fd80-e6e3-11e9-8d9b-b853e2bcbacf.png)

![image](https://user-images.githubusercontent.com/94054/66249107-b333ec80-e6e3-11e9-992e-4d3468426416.png)

/cc @agr

loic-sharma:
A customer has reported this issue on Twitter: https://twitter.com/peymanr34/status/1180636214954795008

PR #7608, opened 2019-10-07
Fix for iconUrl deprecation message appearing for packages without any icon

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/7605.

### DIFF
--- a/src/NuGetGallery/Services/PackageUploadService.cs
+++ b/src/NuGetGallery/Services/PackageUploadService.cs
@@ -350,7 +350,7 @@ namespace NuGetGallery
 
             if (iconElement == null)
             {
-                if (embeddedIconsEnabled && nuspecReader.GetIconUrl() != null)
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -350,7 +350,7 @@ private async Task<PackageValidationResult> CheckIconMetadataAsync(PackageArchiv
 
             if (iconElement == null)
             {
-                if (embeddedIconsEnabled && nuspecReader.GetIconUrl() != null)
+                if (embeddedIconsEnabled && !string.IsNullOrWhiteSpace(nuspecReader.GetIconUrl()))
```
joelverhagen: What was being returned? Empty string?

agr: Yes.

PR #7614, opened 2019-10-10
Display SHA-256 instead of SHA-1 for user certificates

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/2758.

In general we are avoiding the use of SHA-1 anyway as a cleanliness practice. SHA-256 or greater is the hash algorithm recommendation.

# Old

image

# New

image

### DIFF
--- a/src/Bootstrap/dist/css/bootstrap-theme.css
+++ b/src/Bootstrap/dist/css/bootstrap-theme.css
@@ -324,6 +324,10 @@ img.reserved-indicator-icon {
 .breadcrumb-menu h1 {
   margin-top: 0px;
 }
+.page-account-settings .small-fingerprint,
+.page-manage-organizations .small-fingerprint {
+  font-size: 75%;
+}
 .modal-backdrop.in {
   opacity: 0.8 !important;
   position: fixed;
--- a/src/NuGet.Services.Entities/Certificate.cs
+++ b/src/NuGet.Services.Entities/Certificate.cs
@@ -18,6 +18,7 @@ namespace NuGet.Services.Entities
         /// <summary>
         /// Gets or sets the SHA-1 thumbprint of the certificate.
         /// </summary>
+        [Obsolete("This property should not be used since SHA-1 usage is avoided in NuGetGallery.")]
         public string Sha1Thumbprint { get; set; }
 
         /// <summary>
--- a/src/NuGetGallery/Services/CertificateService.cs
+++ b/src/NuGetGallery/Services/CertificateService.cs
@@ -59,7 +59,9 @@ namespace NuGetGallery
 
                     certificate = new Certificate()
                     {
+#pragma warning disable CS0618 // Only set the SHA1 thumbprint, for backwards compatibility. Never read it.
                         Sha1Thumbprint = certificateFile.Sha1Thumbprint,
+#pragma warning restore CS0618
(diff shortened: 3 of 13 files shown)


src/Bootstrap/dist/css/bootstrap-theme.css
```diff
@@ -324,6 +324,10 @@ img.reserved-indicator-icon {
 .breadcrumb-menu h1 {
   margin-top: 0px;
 }
+.page-account-settings .small-fingerprint,
```
dtivel: Per your screenshot, this results in a red fingerprint.  Use of a different color is to grab attention, and I don't think it's needed here.  Besides, red is already used elsewhere to mean expired.

I think the fingerprint should use the default font color (e.g.:  black, in your screenshot).

src/NuGetGallery/Views/Shared/_AccountCertificates.cshtml
```diff
@@ -109,6 +109,16 @@
                             </tr>
                         </tbody>
                     </table>
+
+                    <p>
+                        The SHA-256 fingerprint can be found by calculating the SHA-256 hash of the DER encoded
```
dtivel: ...by calculating the **hex-encoded** SHA-256 hash of the DER-encoded...

src/Bootstrap/dist/css/bootstrap-theme.css
joelverhagen: Good point. Fixed.

src/NuGetGallery/Views/Shared/_AccountCertificates.cshtml
joelverhagen: Done.

src/NuGet.Services.Entities/Certificate.cs
```diff
@@ -18,6 +18,7 @@ public class Certificate : IEntity
         /// <summary>
         /// Gets or sets the SHA-1 thumbprint of the certificate.
         /// </summary>
+        [Obsolete("This property should not be used since SHA-1 usage is avoided in NuGetGallery.")]
         public string Sha1Thumbprint { get; set; }
```
cristinamanum: Why not removing it?

src/NuGetGallery/Services/CertificateService.cs
```diff
@@ -59,7 +59,9 @@ public async Task<Certificate> AddCertificateAsync(HttpPostedFileBase file)
 
                     certificate = new Certificate()
                     {
+#pragma warning disable CS0618 // Only set the SHA1 thumbprint, for backwards compatibility. Never read it.
                         Sha1Thumbprint = certificateFile.Sha1Thumbprint,
```
cristinamanum: If not used why not removing it? What are the backwards compat concerns?

joelverhagen: Rollback for the short term.

Spoke to @skofman1 about this. This is a scoped fix for SDL purposes. We can consider clean up later.

src/NuGet.Services.Entities/Certificate.cs
joelverhagen: See https://github.com/NuGet/NuGetGallery/pull/7614#discussion_r333665880.

PR #7619, opened 2019-10-11
[DB Migration] Use new EF6.3 and remove the fix. 

zhhyu:
Entity Framework 6.3 has released, and provide another way to pass DbContext to the Migrator;
Depend on EF6.3 and remove the hacky fix.

Ref:
https://github.com/aspnet/EntityFramework6/commit/9690d3c21e98cd85a6a5e8312af53de286173b22


src/DatabaseMigrationTools/SupportRequestDbMigrationContext.cs
```diff
@@ -24,7 +24,7 @@ public SupportRequestDbMigrationContext(SqlConnection sqlConnection)
             };
 
             var migrationsConfiguration = new SupportRequestMigrationsConfiguration();
-            GetDbMigrator = () => new DbMigrator(migrationsConfiguration);
+            GetDbMigrator = () => new DbMigrator(migrationsConfiguration, new SupportRequestDbContext(SqlConnection));
```
cristinamanum: Why the readOnly difference with Gallery db ?

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -46,8 +46,7 @@ public override async Task Run()
             var migrationContext = await _migrationContextFactory.CreateMigrationContextAsync(_migrationTargetDatabase, _serviceProvider);
 
             ExecuteDatabaseMigration(migrationContext.GetDbMigrator,
-                migrationContext.SqlConnection,
-                migrationContext.SqlConnectionAccessToken);
+                migrationContext.SqlConnection);
 
             migrationContext.SqlConnection.Dispose();
```
cristinamanum: On exception in ExecuteDatabaseMigration this will not be invoked .  We can do try / finally pattern or make the migrationContext disposable and use the using(...) {} pattern.

src/NuGet.Services.DatabaseMigration/Job.cs
```diff
@@ -96,14 +95,11 @@ public void CheckIsValidMigration(List<string> databaseMigrations, List<string>
             }
         }
 
-        private void ExecuteDatabaseMigration(Func<DbMigrator> getMigrator, SqlConnection sqlConnection, string accessToken)
+        private void ExecuteDatabaseMigration(Func<DbMigrator> getMigrator, SqlConnection sqlConnection)
```
cristinamanum: Why there is no need of access token anymore?

src/DatabaseMigrationTools/SupportRequestDbMigrationContext.cs
zhhyu: It seems that Support and Validation DbContext don't have the readonly mode.

src/NuGet.Services.DatabaseMigration/Job.cs
zhhyu: The access token is used to reset SQL connection because of the EF's issue.
Now we upgrade to EF6.3 and the access token is not needed for the fix any more.

zhhyu:
Yeah, I have verified against my personal DB.

zhhyu:
Upgrade Gallery to EF6.3 will break the PM console migration here: 
https://github.com/aspnet/EntityFramework6/issues/1290
There is a new preview version of EF6.4, but we may not want to use preview version in our production system.
EF6.2 works fine. @joelverhagen

PR #7622, opened 2019-10-11
Add configuration support for heartbeat interval to NuGetGallery

xavierdecoster:
NuGetGallery AI heartbeats currently emit every 15 minutes (default since AI upgrade to v2.10). This PR makes the interval configurable.

### DIFF
--- a/src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
+++ b/src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
@@ -73,13 +73,10 @@
       <Version>1.1.2</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Jobs.Common">
-      <Version>4.1.0-master-2758883</Version>
-    </PackageReference>
-    <PackageReference Include="NuGet.Services.Configuration">
-      <Version>2.55.0</Version>
+      <Version>4.2.0</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Validation">
-      <Version>2.55.0</Version>
+      <Version>2.57.0</Version>
     </PackageReference>
   </ItemGroup>
   <ItemGroup>
--- a/src/NuGet.Services.DatabaseMigration/NuGet.Services.DatabaseMigration.csproj
+++ b/src/NuGet.Services.DatabaseMigration/NuGet.Services.DatabaseMigration.csproj
@@ -89,10 +89,7 @@
       <Version>4.8.0</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Jobs.Common">
-      <Version>4.1.0-master-2602271</Version>
-    </PackageReference>
-    <PackageReference Include="NuGet.Services.Configuration">
-      <Version>2.55.0</Version>
+      <Version>4.2.0</Version>
     </PackageReference>
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
--- a/src/NuGetGallery.Core/NuGetGallery.Core.csproj
+++ b/src/NuGetGallery.Core/NuGetGallery.Core.csproj
@@ -238,20 +238,17 @@
     <EmbeddedResource Include="Infrastructure\MigrateUserToOrganization.sql" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="NuGet.Services.Entities">
-      <Version>2.55.0</Version>
-    </PackageReference>
     <PackageReference Include="NuGet.Services.FeatureFlags">
-      <Version>2.55.0</Version>
+      <Version>2.57.0</Version>
     </PackageReference>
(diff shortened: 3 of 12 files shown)


src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
```diff
@@ -73,13 +73,10 @@
       <Version>1.1.2</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Jobs.Common">
-      <Version>4.1.0-master-2758883</Version>
-    </PackageReference>
-    <PackageReference Include="NuGet.Services.Configuration">
-      <Version>2.55.0</Version>
```
xavierdecoster: This is another transitive dependency of `NuGet.Jobs.Common` that is upgraded to version 2.57.0 as part of this change.

src/NuGet.Services.DatabaseMigration/NuGet.Services.DatabaseMigration.csproj
```diff
@@ -89,10 +89,7 @@
       <Version>4.8.0</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Jobs.Common">
-      <Version>4.1.0-master-2602271</Version>
-    </PackageReference>
-    <PackageReference Include="NuGet.Services.Configuration">
-      <Version>2.55.0</Version>
```
xavierdecoster: Transitive dependency of `NuGet.Jobs.Common` that is upgraded to version 2.57.0 as part of this change.

src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -238,20 +238,17 @@
     <EmbeddedResource Include="Infrastructure\MigrateUserToOrganization.sql" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="NuGet.Services.Entities">
-      <Version>2.55.0</Version>
-    </PackageReference>
```
xavierdecoster: No idea why this was a package reference...

The project is part of the solution and a project reference already existed (see [these lines](https://github.com/NuGet/NuGetGallery/pull/7622/files#diff-41376359b1a4bde68aef3e4df1ad3045R288-R291)), so I removed the package reference.

src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
```diff
@@ -73,13 +73,10 @@
...
+      <Version>4.2.0</Version>
```
joelverhagen: I don't remember us ever having a "stable" jobs version.

Seems like a good practice (it's what we do with ServerCommon) but we would just wanna be sure assembly version changes don't cause us even more pain (i.e. more binding redirects)

skofman1:
What't the benefit of adding heartbeat to a service? We monitor it with Pingdom/AI..

scottbommarito:
@skofman1 - One thought (that @joelverhagen brought up during an offline discussion): we could potentially use this heartbeat to determine whether or not AI data is flowing. Our outside-in monitoring can't determine that already.

skofman1:
@scottbommarito ..  yeah... but our ping requests generate logs anyways...

src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
xavierdecoster: We really must change our process to deliver _stable_ versions every now and then.
Btw, version 4.2.0 is the one I'm rolling out everywhere, which introduces the 2.57.0 version of ServerCommon everywhere too. It's an alignment across repositories, and it helps us remove top-level dependencies and turn them into transitive ones, making it easier to manage them going forward: we manage _stable top-level packages_ instead.

Will double-check the binding redirects, but I've been able to remove some elsewhere simply by aligning dependencies.

xavierdecoster:
> What't the benefit of adding heartbeat to a service? We monitor it with Pingdom/AI..

Note: this PR doesn't _add_ heartbeats. It allows us to configure the interval.

The heartbeats were added by simply upgrading AppInsights to v2.10 in a previous PR.
So, given we have heartbeats anyway, we can as well align the configuration with our other server components.

src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
xavierdecoster: I've removed all `assemblyBinding` elements from the gallery's `web.config`, rebuild the solution, double-clicked the warning to fix binding redirects: resulting bindings are exactly the same, so I think we're good.

`DatabaseMigrationTools` is the only other project with a binding redirect, which remains unchanged, and which will likely even be obsolete once [this ServerCommon PR](https://github.com/NuGet/ServerCommon/pull/317) makes it in (needs 👀 ).

xavierdecoster:
Ping!

Issue #7638, opened 2019-10-21
[Azure Search] Update side-by-side page's messaging when the page is disabled


loic-sharma:
Old search is now the current search service. New search is now the preview search service.

joelverhagen:
Let's make this a graceful 404 page with clear message "this page is closed right now, keep an eye on twitter". For now we will 404 it since it's very confusing.

karann-msft:
I would say just redirect to the blog post that says the new search is live.

joelverhagen:
@karann-msft I think I blind redirect on a page that once worked is confusing. I am thinking something like this:

![image](https://user-images.githubusercontent.com/94054/67226878-def4e900-f3ea-11e9-93f7-bab606471d84.png)

Landing on a blog post when a user had a specific action in mind I think would be disorienting. A message like this is disappointing perhaps to the user but it's at least obvious what is going on.

PR #7639, opened 2019-10-21
Add better messaging when SxS page is disabled

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7638

When the side-by-side page is turned off, anyone hitting the URL will see:

image


joelverhagen:
Updated the messaging per @karann-msft's feedback.

PR #7644, opened 2019-10-23
[Token scanning-Phase 1-1] Admin page to verify the token

zhhyu:
[Token scanning-Phase 1-1] Admin page to verify the token.

1. Verify multiple tokens:
image

2. Invalid Input:
image

3. Not found tokens:
image

### DIFF
--- a/src/NuGetGallery.Services/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery.Services/Authentication/AuthenticationService.cs
@@ -158,7 +158,7 @@ namespace NuGetGallery.Authentication
                 new Credential { Type = CredentialTypes.ApiKey.Prefix, Value = apiKey });
         }
 
-        public Credential Verify(string apiKey)
+        public Credential GetApiKeyCredential(string apiKey)
         {
             var credential = new Credential { Type = CredentialTypes.ApiKey.Prefix, Value = apiKey };
             return FindMatchingApiKey(credential);
--- a/src/NuGetGallery.Services/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery.Services/Authentication/AuthenticationService.cs
@@ -158,6 +158,12 @@ namespace NuGetGallery.Authentication
                 new Credential { Type = CredentialTypes.ApiKey.Prefix, Value = apiKey });
         }
 
+        public Credential Verify(string apiKey)
+        {
+            var credential = new Credential { Type = CredentialTypes.ApiKey.Prefix, Value = apiKey };
+            return FindMatchingApiKey(credential);
+        }
+
         public virtual async Task<AuthenticatedUser> Authenticate(Credential credential)
         {
             return await AuthenticateInternal(FindMatchingCredential, credential);
--- a/src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
+++ b/src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
@@ -18,17 +18,15 @@ namespace NuGetGallery.Authentication
         Task RemoveCredential(User user, Credential cred, bool commitChanges = true);
 
         /// <summary>
-        /// Authenticate the Api Key
-        /// </summary>
-        /// <param name="apiKey">Api Key</param>
-        /// <returns>Returns a task that will complete with the authenticated user or null if there is no matched user.</returns>
-        Task<AuthenticatedUser> Authenticate(string apiKey);
-
-        /// <summary>
         /// Create the CredentialViewModel given the credential
         /// </summary>
-        /// <param name="credential">credential needed to generate the ViewModel</param>
         /// <returns>Returns the CredentialViewModel given the credential.</returns>
         CredentialViewModel DescribeCredential(Credential credential);
+
+        /// <summary>
+        /// Verify the ApiKey
+        /// </summary>
+        /// <returns>Return a credential if there is a matched apiKey or null if there is not.</returns>
+        Credential Verify(string apiKey);
     }
 }
--- a/src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
+++ b/src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
@@ -24,7 +24,7 @@ namespace NuGetGallery.Authentication
         CredentialViewModel DescribeCredential(Credential credential);
 
         /// <summary>
-        /// Verify the ApiKey
+        /// Get the ApiKey credential given the apiKey
         /// </summary>
         /// <returns>Return a credential if there is a matched apiKey or null if there is not.</returns>
         Credential GetApiKeyCredential(string apiKey);
--- a/src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
+++ b/src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
@@ -27,6 +27,6 @@ namespace NuGetGallery.Authentication
         /// Verify the ApiKey
         /// </summary>
         /// <returns>Return a credential if there is a matched apiKey or null if there is not.</returns>
-        Credential Verify(string apiKey);
+        Credential GetApiKeyCredential(string apiKey);
     }
 }
--- a/src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
+++ b/src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
@@ -16,5 +16,19 @@ namespace NuGetGallery.Authentication
         /// <param name="commitChanges">Default true. Commits changes immediately if true.</param>
         /// <returns>Returns a task that will complete when the credential has succesfully been removed.</returns>
         Task RemoveCredential(User user, Credential cred, bool commitChanges = true);
+
+        /// <summary>
+        /// Authenticate the Api Key
+        /// </summary>
+        /// <param name="apiKey">Api Key</param>
+        /// <returns>Returns a task that will complete with the authenticated user or null if there is no matched user.</returns>
+        Task<AuthenticatedUser> Authenticate(string apiKey);
+
+        /// <summary>
+        /// Create the CredentialViewModel given the credential
+        /// </summary>
+        /// <param name="credential">credential needed to generate the ViewModel</param>
+        /// <returns>Returns the CredentialViewModel given the credential.</returns>
+        CredentialViewModel DescribeCredential(Credential credential);
     }
 }
--- a/src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
+++ b/src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
@@ -2,22 +2,24 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Net;
 using System.Web.Mvc;
-using System.Threading.Tasks;
 using System.Collections.Generic;
 using Newtonsoft.Json;
-using NuGetGallery.Areas.Admin.ViewModels;
 using NuGetGallery.Authentication;
+using NuGetGallery.Areas.Admin.ViewModels;
 
 namespace NuGetGallery.Areas.Admin.Controllers
 {
     public class ApiKeysController : AdminControllerBase
     {
         private readonly IAuthenticationService _authenticationService;
+        private readonly ITelemetryService _telemetryService;
 
-        public ApiKeysController(IAuthenticationService authenticationService)
+        public ApiKeysController(IAuthenticationService authenticationService, ITelemetryService telemetryService)
         {
             _authenticationService = authenticationService ?? throw new ArgumentNullException(nameof(authenticationService));
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(authenticationService));
         }
 
         [HttpGet]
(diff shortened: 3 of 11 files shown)


src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -0,0 +1,72 @@
...
+            return View();
+        }
+
+        [HttpGet]
+        public async virtual Task<ActionResult> Verify(string verifyQuery)
+        {
+            var queries = verifyQuery.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+
+            var results = new List<ApiKeyRevokeViewModel>();
+            var verifiedApiKey = new HashSet<string>();
```
joelverhagen: maybe call this `uniqueApiKeys`. Looks like the only purpose of this is for deduping.

Are API keys case-insensitive? I think they are. Consider a case insensitive comparer.

src/NuGetGallery/Areas/Admin/ViewModels/ApiKeyRevokeViewModel.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace NuGetGallery.Areas.Admin.ViewModels
+{
+    public class ApiKeyRevokeViewModel
+    {
+        public ApiKeyViewModel ApiKeyViewModel { get; }
+        public string ApiKey { get; }
+        public string LeakedURL { get; }
+
+        public ApiKeyRevokeViewModel(ApiKeyViewModel apiKeyViewModel, string apiKey, string leakedURL)
```
joelverhagen: Nit: lower `Url`

src/NuGetGallery/Areas/Admin/Views/ApiKeys/Index.cshtml
```diff
@@ -0,0 +1,153 @@
...
+                            <!-- /ko -->
+                            <!-- ko ifnot: ApiKeyViewModel.Expires -->
+                            Never expires<br />
+                            <!-- /ko -->
+                            <!-- /ko -->
+                        </span>
+                        <span data-bind="visible: ApiKeyViewModel.Description">
+                            <code>Description:</code> <span data-bind="text: ApiKeyViewModel.Description"></span><br />
+                        </span>
+                        <code>Leaked URL:</code> <a data-bind="text: LeakedURL, attr: {href: LeakedURL}"></a><br />
```
joelverhagen: Using `<code>` for the property name is a little strange. It's readable but typically people use the monospace elements for verbatim _values_ rather than _names_. Follow your heart though. It's still quite readable.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -0,0 +1,72 @@
...
+            _authenticationService = authenticationService ?? throw new ArgumentNullException(nameof(authenticationService));
+        }
+
+        [HttpGet]
+        public ActionResult Index()
+        {
+            return View();
+        }
+
+        [HttpGet]
```
joelverhagen: Consider making this a POST. GET requests get saved in browser history which is probably not ideal for API keys in general. This is just the "verify" step so it's the key is still valid at this point (and may not even be acted on depending on what the admin does next)

src/NuGetGallery/Areas/Admin/Views/ApiKeys/Index.cshtml
```diff
@@ -0,0 +1,153 @@
...
+                this.validateInput = function () {
+                    var input = $self.verifyQuery();
+                    if (!input || input.trim().length === 0) {
+                        $self.isNotValidInput(true);
+                        $('#invalidInputAlert').html('<span>Invalid empty input!</span>');
+                        return false;
+                    }
+
+                    var queries = input.match(/[^\r\n]+/g);
+                    for (var i = 0; i < queries.length; i++) {
```
joelverhagen: Is it necessary to have client side validation? This seems like duplicate code that can be just done on the server side

src/NuGetGallery/Areas/Admin/Views/ApiKeys/Index.cshtml
```diff
@@ -0,0 +1,153 @@
...
+    <form>
+        <textarea placeholder="Verify Api keys to revoke" autocomplete="off" autofocus style="width: 100%" rows="5" data-bind="value: verifyQuery" required></textarea><br />
+        <input type="button" value="Verify" data-bind="click: verify" /><br />
+    </form>
+    <div class="alert alert-warning" id="invalidInputAlert" role="alert" style="display: none" data-bind="visible: isNotValidInput()"></div>
+    <table class="table" style="display: none" data-bind="visible: verifyResults().length > 0">
+        <thead>
+            <tr>
+                <th style="width: 2%; vertical-align: middle"><input type="checkbox" data-bind="visible: findAnyValidApiKeys"/></th>
+                <th style="width: 40%; vertical-align: middle">Api Key</th>
```
joelverhagen: nit: we call it "API key" in our display text typically, not "Api key" -- there is a divergence of display text and names in code.

src/NuGetGallery/Areas/Admin/Views/ApiKeys/Index.cshtml
```diff
@@ -0,0 +1,153 @@
...
+</section>
+
+@section BottomScripts {
+    <script>
+        $(document).ready(function () {
+            var viewModel = function () {
+                var $self = this;
+
+                this.verifyQuery = ko.observable('');
+                this.verifyResults = ko.observableArray([]);
```
joelverhagen: `verifyResults` sounds like a method name. `verifiedResults` may be clearer

src/NuGetGallery/Areas/Admin/Views/ApiKeys/Index.cshtml
```diff
@@ -0,0 +1,153 @@
...
+    </p>
+    <form>
+        <textarea placeholder="Verify Api keys to revoke" autocomplete="off" autofocus style="width: 100%" rows="5" data-bind="value: verifyQuery" required></textarea><br />
+        <input type="button" value="Verify" data-bind="click: verify" /><br />
+    </form>
+    <div class="alert alert-warning" id="invalidInputAlert" role="alert" style="display: none" data-bind="visible: isNotValidInput()"></div>
+    <table class="table" style="display: none" data-bind="visible: verifyResults().length > 0">
+        <thead>
+            <tr>
+                <th style="width: 2%; vertical-align: middle"><input type="checkbox" data-bind="visible: findAnyValidApiKeys"/></th>
```
joelverhagen: What's the purpose of the checkbox at this point?

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -0,0 +1,72 @@
...
+        public async virtual Task<ActionResult> Verify(string verifyQuery)
```
skofman1: string can be null

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                var queryObject = JsonConvert.DeserializeObject<ApiKeyAndLeakedURL>(query);
+                var apiKey = queryObject.ApiKey;
+                if (!verifiedApiKey.Add(apiKey))
+                {
+                    continue;
+                }
+                var leakedURL = queryObject.LeakedURL;
+
+                var authenticationResult = await _authenticationService.Authenticate(apiKey);
```
skofman1: This method is used for actual authentication. This means: 1. It audits the authentication
2. It updates "LastUsed" column for the credential
3. It expires a V3 API Key if it wasn't used for more then a year.

This is not the behavior you are looking for. Something along those lines will be simpler: [https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery.Services/Authentication/AuthenticationService.cs#L811](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery.Services/Authentication/AuthenticationService.cs#L811)

src/NuGetGallery/Areas/Admin/ViewModels/ApiKeyRevokeViewModel.cs
skofman1: null checks

src/NuGetGallery/Areas/Admin/Views/ApiKeys/Index.cshtml
zhhyu: The checkbox is set in this PR and will be used to achieve the true selection and revoking in the next PR.

src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
```diff
@@ -16,5 +16,19 @@ public interface IAuthenticationService
         /// <param name="commitChanges">Default true. Commits changes immediately if true.</param>
         /// <returns>Returns a task that will complete when the credential has succesfully been removed.</returns>
         Task RemoveCredential(User user, Credential cred, bool commitChanges = true);
+
+        /// <summary>
+        /// Authenticate the Api Key
+        /// </summary>
+        /// <param name="apiKey">Api Key</param>
```
scottbommarito: This comment doesn't provide any more information than the parameter name itself.

I typically exclude the `<param>` from the doc if its purpose is fairly obvious.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
scottbommarito: I feel like POST is the wrong verb though, because we're not modifying the API keys here. I understand the caching/history concern but is there not another way to tackle it? Could we tell the browser not to cache it without changing the verb though?

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        [HttpGet]
+        public async virtual Task<ActionResult> Verify(string verifyQuery)
+        {
+            var queries = verifyQuery.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+
+            var results = new List<ApiKeyRevokeViewModel>();
+            var verifiedApiKey = new HashSet<string>();
+            foreach (var query in queries)
```
scottbommarito: You shouldn't need to check the set in this array. If you create an `IEqualityComparer` for `ApiKeyAndLeakedURL` you can use the constructor of `HashSet`:

```
private class ApiKeyAndLeakedUrlEqualityComparer : IEqualityComparer<ApiKeyAndLeakedURL> { ... }
```

In `Verify`:
```
var uniqueApiKeys = new HashSet(queries.Select(q => JsonConvert.DeserializeObject<ApiKeyAndLeakedUrl>(q)), new ApiKeyAndLeakedUrlEqualityComparer());

foreach (var apiKey in uniqueApiKeys)
{
    ...do the stuff you already have in the loop...
}
```

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
zhhyu: Oops, sorry, I don't realize this problem.

src/NuGetGallery/Areas/Admin/Views/ApiKeys/Index.cshtml
scottbommarito: A couple thoughts here...
- The browser already automatically handles table column widths. In general I suggest not messing with them because it's simpler to just let the browser do its thing.

- If you really want to specify a width, I would convert the table to divs and use Bootstrap's grid classes, e.g.
```
<div class="row">
    <div class="col-sm-1>
        ....checkbox....
    </div>
    <div class="col-sm-5>
        ....API key...
    </div>
    <div class="col-sm-6>
        ....information...
    </div>
</div>
```

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
joelverhagen: If it is in the GET query parameters you need to obfuscate telemetry. Just doesn't seem worth the effort.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
zhhyu: The API keys V4 are normalized and always lower-case. And I also feel that maybe the "comparator" to be case-sensitive is better because we prefer to verify any possible cases.

src/NuGetGallery.Services/Authentication/AuthenticationService.cs
```diff
@@ -158,6 +158,12 @@ public virtual async Task<AuthenticatedUser> Authenticate(string apiKey)
                 new Credential { Type = CredentialTypes.ApiKey.Prefix, Value = apiKey });
         }
 
+        public Credential Verify(string apiKey)
```
shishirx34: may be `GetApiKeyCredential` be an appropriate name? You aren't really verifying the key.

src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
```diff
@@ -16,5 +16,17 @@ public interface IAuthenticationService
...
+        /// Create the CredentialViewModel given the credential
+        /// </summary>
+        /// <returns>Returns the CredentialViewModel given the credential.</returns>
+        CredentialViewModel DescribeCredential(Credential credential);
+
+        /// <summary>
+        /// Verify the ApiKey
+        /// </summary>
+        /// <returns>Return a credential if there is a matched apiKey or null if there is not.</returns>
+        Credential Verify(string apiKey);
```
shishirx34: see my comment above to rename this function.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -0,0 +1,98 @@
...
+                    var leakedUrl = queryObject.LeakedUrl;
+
+                    var credential = _authenticationService.Verify(apiKey);
+                    var apiKeyViewModel = credential == null ? null : new ApiKeyViewModel(_authenticationService.DescribeCredential(credential));
+
+                    results.Add(new ApiKeyRevokeViewModel(apiKeyViewModel, apiKey, leakedUrl, isRevocable: IsRevocable(apiKeyViewModel)));
+                }
+                catch (JsonException)
+                {
+                    return Json(HttpStatusCode.BadRequest, $"Invalid input! {query} is not using the valid JSON format.");
```
shishirx34: not super critical, but you could aggregate the check on all the queries to validate the json format and then return `BadRequest` with all the malformed queries.

src/NuGetGallery/Areas/Admin/ViewModels/ApiKeyRevokeViewModel.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace NuGetGallery.Areas.Admin.ViewModels
+{
+    public class ApiKeyRevokeViewModel
+    {
+        public ApiKeyViewModel ApiKeyViewModel { get; }
+        public string ApiKey { get; }
+        public string LeakedUrl { get; }
+        public bool IsRevocable { get; }
+
+        public ApiKeyRevokeViewModel(ApiKeyViewModel apiKeyViewModel, string apiKey, string leakedUrl, bool isRevocable)
```
skofman1: apiKeyViewModel can be null in some cases. This is usually not an expected behavior. Consider making it nullable ( ApiKeyViewModel? apiKeyViewModel) for readability

src/NuGetGallery/ViewModels/ApiKeyViewModel.cs
```diff
@@ -20,6 +20,11 @@ public ApiKeyViewModel(CredentialViewModel cred)
                 throw new ArgumentNullException(nameof(cred));
             }
 
+            if (cred.Scopes == null)
```
skofman1: V3 API Keys don't have scopes. I'm not sure if for this case scopes is an empty list or null. This might cause a bug.

zhhyu: That's a very important point. I am looking into it.

zhhyu: 1. The Scopes of Credential entity will always be an empty list, as the default constructor of Credential entity initializes it:
https://github.com/NuGet/NuGetGallery/blob/a7600e9fb744a2ad2c6bbfafdfa15d42123b37e3/src/NuGet.Services.Entities/Credential.cs#L20
2. The API Key profile page actually uses the same logic to prepare the ApiKeyViewModel, and the API Key profile page supports V3 no-scoped API Keys. If something goes wrong, it will pop up earlier;
3. I verify it locally using a V1 no-scoped API key, and the Scopes is an empty list rather null.  

So the check here is not a must if the credentialViewModel is flowed from the Credential entity. And it will still throw the exception if cred.Scopes is null without the null check here.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
zhhyu: Move to POST as GitHub actually sends us the POST message with the exposed API key .

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
zhhyu: That's a much more elegant way to achieve the goal. I wonder it will scan through the API keys twice? 
I prefer to make it run in fully parallel later, which will be faster when the number of API keys needed to be verified is big, because in my understanding, the hash calculation needs some time.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
zhhyu: I prefer to stop earlier when the number of input API keys is big.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
scottbommarito: Yup, it will scan the API keys twice--once to set up the `HashSet`, and then again to do the actions for each distinct key. Even though it is technically enumerating the API keys twice with this approach, however, it's doing the same amount of work it was doing before, because the previous loop hashed as well.

I think you would want to make `GetApiKeyCredential` accept multiple API keys so that it can make a single query to the database rather than parallelize this method.

zhhyu: I am thinking to make this for-loop to be in parallel. The "GetApiKeyCredential" locates in the "authenticationService", and I prefer to leave it as it is now. The authenticationService dependencies for each verification targets one apiKey.

src/NuGetGallery/Areas/Admin/ViewModels/ApiKeyRevokeViewModel.cs
zhhyu: Nullable reference type is not supported now until C# 8.0:
https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/nullable-reference-types

src/NuGetGallery/Areas/Admin/Views/ApiKeys/Index.cshtml
```diff
@@ -0,0 +1,127 @@
...
+        {"ApiKey":"oy2apikey2","LeakedUrl":"https://leakedUrl2"}
+    </p>
+    <form>
+        <textarea class="form-control" placeholder="Verify API keys to revoke" autocomplete="off" autofocus rows="5" data-bind="value: verifyQuery"></textarea><br />
+        <input type="button" value="Verify" data-bind="click: verify" /><br />
+    </form>
+    <!-- ko if: Error -->
+    @ViewHelpers.AlertDanger(@<text><span data-bind="html: Error"></span></text>)
+    <!-- /ko -->
+    <table class="table" style="word-break: break-word; display: none" data-bind="visible: verifiedResults().length > 0">
```
scottbommarito: `display: none` is redundant--the `data-bind="visible: ..."` will overwrite it.

`break-word` is deprecated: see https://developer.mozilla.org/en-US/docs/Web/CSS/word-break. What are you using it here for?

zhhyu: Move to Bootstrap style.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
zhhyu: Will follow-up this problem later.

PR #7645, opened 2019-10-24
[MSA] Audit the external credential for attempted login in Gallery

shishirx34:
We need to audit the credential which was used to attempt to login in the gallery before it was identified and authorized(from DB). This is necessary for retrospective analysis. Also, log the `Tenant Id` and `OID` used with the external credential.

This depends on the shims changes.

Addresses: https://github.com/NuGet/Engineering/issues/2652

### DIFF
--- a/src/NuGetGallery.Core/Auditing/AuditedUserAction.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditedUserAction.cs
@@ -24,6 +24,6 @@ namespace NuGetGallery.Auditing
         UpdateOrganizationMember,
         EnabledMultiFactorAuthentication,
         DisabledMultiFactorAuthentication,
-        ExternalLoginAttempt
+        ExternalLoginAttempt,
     }
 }
--- a/src/NuGetGallery.Core/Auditing/AuditedUserAction.cs
+++ b/src/NuGetGallery.Core/Auditing/AuditedUserAction.cs
@@ -23,6 +23,7 @@ namespace NuGetGallery.Auditing
         RemoveOrganizationMember,
         UpdateOrganizationMember,
         EnabledMultiFactorAuthentication,
-        DisabledMultiFactorAuthentication
+        DisabledMultiFactorAuthentication,
+        ExternalLoginAttempt
     }
 }
--- a/src/NuGetGallery.Services/Authentication/AuthenticationService.cs
+++ b/src/NuGetGallery.Services/Authentication/AuthenticationService.cs
@@ -675,6 +675,9 @@ namespace NuGetGallery.Authentication
             // Authenticate!
             if (result.Credential != null)
             {
+                // We want to audit the received credential from external authentication. We use the UserAuditRecord
+                // for easier logging, since it actually needs a `User`, instead we use a dummy user because at this point
+                // we do not have the actual user context since this request has not yet been authenticated.
                 await Auditing.SaveAuditRecordAsync(new UserAuditRecord(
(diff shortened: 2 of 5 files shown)


src/NuGetGallery.Core/Auditing/AuditedUserAction.cs
```diff
@@ -23,6 +23,7 @@ public enum AuditedUserAction
         RemoveOrganizationMember,
         UpdateOrganizationMember,
         EnabledMultiFactorAuthentication,
-        DisabledMultiFactorAuthentication
+        DisabledMultiFactorAuthentication,
+        ExternalLoginAttempt
```
joelverhagen: nit trailing comma to avoid diff noise

src/NuGetGallery.Services/Authentication/AuthenticationService.cs
```diff
@@ -675,6 +675,9 @@ public virtual async Task<AuthenticateExternalLoginResult> AuthenticateExternalL
             // Authenticate!
             if (result.Credential != null)
             {
+                await Auditing.SaveAuditRecordAsync(new UserAuditRecord(
```
joelverhagen: Why is this a dummy audit user?

shishirx34: At this stage we do not have the actual user, this is the credential that we receive from the azure identity side after authentication but before we authorize on our side(from the DB). I kept the changes to minimum(for shims) and reused the `UserAuditRecord`, which does need a "user"(otherwise it throws). Hence the "dummy audit" user.

We want to audit this information for retrospective analysis. Ideally we'd be using `CredentialAuditRecord` Actions for directly logging this event to Ifx, however, it'd be a much larger change on the Shims side and I don't see anymore such use cases explicitly for credentials audit, credentials are always associated with the `UserAuditRecord`.

shishirx34: I can add a comment for this.

tests/NuGetGallery.Core.Facts/Auditing/CredentialAuditRecordTests.cs
```diff
@@ -51,6 +51,17 @@ public void Constructor_NonRemovalOfNonPasswordDoesNotSetsValue()
             Assert.Null(record.Value);
         }
 
+        [Theory]
+        [InlineData(CredentialTypes.External.MicrosoftAccount)]
+        [InlineData(CredentialTypes.External.AzureActiveDirectoryAccount)]
+        public void Constructor_ExternalCredentialSetsValue(string externalType)
+        {
+            var credential = new Credential(type: externalType, value: "b");
+            var record = new CredentialAuditRecord(credential, removed: false);
```
zhhyu: I have seen the logic at "CredentialAuditRecord". Maybe it's better that we have some additional test coverage like:
1. the credential is external but removed is also true;
Because the logic there seems easily broken in the future.

src/NuGetGallery.Services/Authentication/AuthenticationService.cs
joelverhagen: just to be safe, could we use a username that contains unavailable characters such as `"(before login)"` or something?

tests/NuGetGallery.Core.Facts/Auditing/CredentialAuditRecordTests.cs
shishirx34: I have this covered in the shims, but ok to be more thorough here.

shishirx34:
verified this on dev, looks good.

Issue #7648, opened 2019-10-25
Implement job to fetch vulnerabilities from GitHub


scottbommarito:

skofman1:
Work left: deploy to all environments, verify data integrity

scottbommarito:
We will need to add GitHub PATs to the KeyVaults of all environments before deploying. These PATs should be created with @nugetlurker. The DEV PAT should be recreated with @nugetlurker as well, as it is currently using a PAT associated with my GitHub account.

See:
https://nuget.visualstudio.com/NuGetMicrosoft/_git/NuGetDeployment?path=%2Fsrc%2FJobs%2FNuGet.Jobs.Cloud%2FJobs%2FGitHubVulnerabilities2Db%2FDEV%2Fnorthcentralus%2Fappsettings.json&version=GBmaster
https://nuget.visualstudio.com/NuGetMicrosoft/_git/NuGetDeployment?path=%2Fsrc%2FJobs%2FNuGet.Jobs.Cloud%2FJobs%2FGitHubVulnerabilities2Db%2FINT%2Fnorthcentralus%2Fappsettings.json&version=GBmaster
https://nuget.visualstudio.com/NuGetMicrosoft/_git/NuGetDeployment?path=%2Fsrc%2FJobs%2FNuGet.Jobs.Cloud%2FJobs%2FGitHubVulnerabilities2Db%2FPROD%2Fnorthcentralus%2Fappsettings.json&version=GBmaster

skofman1:
Deployed to dev and running. We put the deployment to prod on hold to avoid reflowing. Should be deployed after V3 pipeline is ready.

drewgillies:
Closing this as data quality work has been completed and tested on PROD.

PR #7660, opened 2019-10-29
Add vulnerabilities entities

scottbommarito:
Part of https://github.com/nuget/nugetgallery/issues/7648

This PR adds 3 new tables:
1. `Vulnerabilities`, to store metadata about security vulnerabilities in packages.
2. `VulnerablePackageRanges`, to store package ID and version ranges that vulnerabilities apply to.
3. `VulnerablePackageRangePackages`, to map `Packages` to `VulnerablePackageRanges` that they are a part of.

# Script

```
CREATE TABLE [dbo].[VulnerablePackageVersionRanges] (
    [Key] [int] NOT NULL IDENTITY,
    [VulnerabilityKey] [int] NOT NULL,
    [PackageId] [nvarchar](128) NOT NULL,
    [PackageVersionRange] [nvarchar](132) NOT NULL,
    CONSTRAINT [PK_dbo.VulnerablePackageVersionRanges] PRIMARY KEY ([Key])
)
CREATE UNIQUE INDEX [IX_VulnerabilityKey_PackageId_PackageVersionRange] ON [dbo].[VulnerablePackageVersionRanges]([VulnerabilityKey], [PackageId], [PackageVersionRange])
CREATE TABLE [dbo].[PackageVulnerabilities] (
    [Key] [int] NOT NULL IDENTITY,
    [GitHubDatabaseKey] [int] NOT NULL,
    [ReferenceUrl] [nvarchar](max),
    [Severity] [int] NOT NULL,
    CONSTRAINT [PK_dbo.PackageVulnerabilities] PRIMARY KEY ([Key])
)
CREATE UNIQUE INDEX [IX_GitHubDatabaseKey] ON [dbo].[PackageVulnerabilities]([GitHubDatabaseKey])
CREATE TABLE [dbo].[VulnerablePackageVersionRangePackages] (
    [VulnerablePackageVersionRange_Key] [int] NOT NULL,
    [Package_Key] [int] NOT NULL,
    CONSTRAINT [PK_dbo.VulnerablePackageVersionRangePackages] PRIMARY KEY ([VulnerablePackageVersionRange_Key], [Package_Key])
)
CREATE INDEX [IX_VulnerablePackageVersionRange_Key] ON [dbo].[VulnerablePackageVersionRangePackages]([VulnerablePackageVersionRange_Key])
CREATE INDEX [IX_Package_Key] ON [dbo].[VulnerablePackageVersionRangePackages]([Package_Key])
ALTER TABLE [dbo].[VulnerablePackageVersionRanges] ADD CONSTRAINT [FK_dbo.VulnerablePackageVersionRanges_dbo.PackageVulnerabilities_VulnerabilityKey] FOREIGN KEY ([VulnerabilityKey]) REFERENCES [dbo].[PackageVulnerabilities] ([Key]) ON DELETE CASCADE
ALTER TABLE [dbo].[VulnerablePackageVersionRangePackages] ADD CONSTRAINT [FK_dbo.VulnerablePackageVersionRangePackages_dbo.VulnerablePackageVersionRanges_VulnerablePackageVersionRange_Key] FOREIGN KEY ([VulnerablePackageVersionRange_Key]) REFERENCES [dbo].[VulnerablePackageVersionRanges] ([Key]) ON DELETE CASCADE
ALTER TABLE [dbo].[VulnerablePackageVersionRangePackages] ADD CONSTRAINT [FK_dbo.VulnerablePackageVersionRangePackages_dbo.Packages_Package_Key] FOREIGN KEY ([Package_Key]) REFERENCES [dbo].[Packages] ([Key]) ON DELETE CASCADE
VALUES (N'201910291956170_AddPackageVulnerabilities', N'NuGetGallery.Migrations.MigrationsConfiguration',  ..., N'6.2.0-61023')
```


src/NuGet.Services.Entities/Package.cs
```diff
@@ -275,6 +276,11 @@ public bool HasReadMe
         /// </summary>
         public virtual ICollection<PackageDeprecation> AlternativeOf { get; set; }
 
+        /// <summary>
+        /// Gets and sets the list of vulnerabilites that this package has.
```
loic-sharma:
```suggestion
        /// Gets or sets the list of vulnerabilites that this package has.
```

src/NuGet.Services.Entities/PackageVulnerability.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGet.Services.Entities
+{
+    public class PackageVulnerability : IEntity
```
loic-sharma: Please add a quick comment on this new entity. Same with `VulnerablePackageVersionRange`

src/NuGet.Services.Entities/PackageVulnerability.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        public int Key { get; set; }
+
+        /// <summary>
+        /// The database key from GitHub.
+        /// </summary>
+        public int GitHubDatabaseKey { get; set; }
+
+        public string ReferenceUrl { get; set; }
```
loic-sharma: Please add a quick comment

src/NuGet.Services.Entities/Package.cs
```diff
@@ -275,6 +276,11 @@ public bool HasReadMe
...
+        /// </summary>
+        public ICollection<VulnerablePackageVersionRange> Vulnerabilities { get; set; }
```
skofman1: virtual

scottbommarito: That would enable lazy-loading, which has caused a number of perf bugs before. I don't think we should make new relationships load lazily, and instead ensure that we properly load the relationships when we query the DB.

src/NuGet.Services.Entities/VulnerablePackageVersionRange.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        public VulnerablePackageVersionRange()
+        {
+            Packages = new HashSet<Package>();
+        }
+
+        public int Key { get; set; }
+
+        public int VulnerabilityKey { get; set; }
+        public PackageVulnerability Vulnerability { get; set; }
```
skofman1: why do we need mapping from VulnerablePackageVersionRange to PackageVulnerability? Wouldn't the querying be done from the other direction?

src/NuGetGallery/Migrations/201910291956170_AddPackageVulnerabilities.cs
```diff
@@ -0,0 +1,64 @@
...
+                .PrimaryKey(t => t.Key)
+                .ForeignKey("dbo.PackageVulnerabilities", t => t.VulnerabilityKey, cascadeDelete: true)
+                .Index(t => new { t.VulnerabilityKey, t.PackageId, t.PackageVersionRange }, unique: true);
+            
+            CreateTable(
+                "dbo.PackageVulnerabilities",
+                c => new
+                    {
+                        Key = c.Int(nullable: false, identity: true),
+                        GitHubDatabaseKey = c.Int(nullable: false),
```
skofman1: if in the future we get vulnerabilities from a different source, GitHubDataBaseKey will be empty

src/NuGet.Services.Entities/VulnerablePackageVersionRange.cs
scottbommarito: There is both--`PackageVulnerability.AffectedRanges` and `VunerableVersionRange.Vulnerability`.

We need both, because if we know a package is vulnerable to this range, we can't get the vulnerability that it is vulnerable to without this field.

src/NuGetGallery.Core/Entities/IEntitiesContext.cs
```diff
@@ -19,6 +19,8 @@ public interface IEntitiesContext : IReadOnlyEntitiesContext
         DbSet<ReservedNamespace> ReservedNamespaces { get; set; }
         DbSet<UserCertificate> UserCertificates { get; set; }
         DbSet<SymbolPackage> SymbolPackages { get; set; }
+        DbSet<PackageVulnerability> Vulnerabilities { get; set; }
+        DbSet<VulnerablePackageVersionRange> VulnerableRanges { get; set; }
```
skofman1: I'm not sure u need it in the context, since u never need direct access to this table, only through Vulnerabilities table.

src/NuGetGallery/Migrations/201910291956170_AddPackageVulnerabilities.cs
scottbommarito: Yes, the plan for adding an additional source is:
1. Add the additional key field (e.g. `XDatabaseKey`)
2. Make GitHubDatabaseKey nullable
3. Remove the unique index on GitHubDatabaseKey and replace it with a unique index on both GitHubDatabaseKey and XDatabaseKey
4. Modify PackageVulnerabilityService to query based on both keys

src/NuGetGallery.Core/Entities/IEntitiesContext.cs
scottbommarito: The service uses it to add `VulnerablePackageVersionRange`s to the database, but does not query on it.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -449,6 +451,25 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
...
                 .WillCascadeOnDelete(false);
+
+            modelBuilder.Entity<PackageVulnerability>()
+                .HasKey(v => v.Key)
+                .HasMany(v => v.AffectedRanges)
+                .WithRequired(pv => pv.Vulnerability)
+                .HasForeignKey(pv => pv.VulnerabilityKey);
+
+            modelBuilder.Entity<PackageVulnerability>()
+                .HasIndex(v => v.GitHubDatabaseKey)
```
skofman1: GitHubDatabaseKey might be empty

scottbommarito: It will never be empty today. When we add additional data sources, we will update this schema and make it nullable.

src/NuGet.Services.Entities/VulnerablePackageVersionRange.cs
scottbommarito: On another note, to properly define the relationship in both directions, we need this navigation property.

src/NuGetGallery/Migrations/201910291956170_AddPackageVulnerabilities.cs
loic-sharma: This seems to be a common question, maybe add a comment about this on the entity?

src/NuGetGallery.Core/Entities/IEntitiesContext.cs
scottbommarito: Note that I would need to import an `IEntityRepository<VulnerablePackageVersionRange>` in the service in order to do this otherwise, and I felt using the single dependency is much cleaner.

src/NuGetGallery/Migrations/201910291956170_AddPackageVulnerabilities.cs
scottbommarito: Good call Loic! Added!

PR #7661, opened 2019-10-29
Add PackageVulnerabilityService

scottbommarito:
Part of https://github.com/nuget/nugetgallery/issues/7648

This PR adds the PackageVulnerabilityService, which will be used directly by the GitHub ingestion job (GitHubVulnerabilities2Db) to update the vulnerabilities in the DB.

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/IPackageVulnerabilityService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/IPackageVulnerabilityService.cs
@@ -8,6 +8,6 @@ namespace NuGetGallery
 {
     public interface IPackageVulnerabilityService
     {
-        Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wasWithdrawn);
+        Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool withdrawn);
     }
 }
--- a/src/NuGetGallery.Services/PackageManagement/IPackageVulnerabilityService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/IPackageVulnerabilityService.cs
@@ -1,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public interface IPackageVulnerabilityService
+    {
+        Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wasWithdrawn);
+    }
+}
--- a/src/NuGetGallery.Services/PackageManagement/IPackageVulnerabilityService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/IPackageVulnerabilityService.cs
@@ -8,6 +8,11 @@ namespace NuGetGallery
 {
     public interface IPackageVulnerabilityService
     {
+        /// <summary>
+        /// If we don't currently have <paramref name="vulnerability"/> in our database, adds it and its <see cref="PackageVulnerability.AffectedRanges"/>.
+        /// If we do, updates the existing entity.
+        /// </summary>
+        /// <param name="withdrawn">Whether or not the vulnerability was withdrawn.</param>
         Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool withdrawn);
     }
 }
--- a/src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
@@ -124,8 +124,12 @@ namespace NuGetGallery
 
             if (updateIndex)
             {
-                // The indexing service will find the latest version of the package to index--it doesn't matter what package we pass in.
-                _indexingService.UpdatePackage(packages.First());
+                // The indexing service will find the latest version of a package to index--it doesn't matter what package we pass in.
+                // We do, however, need to pass in a single package for each registration to ensure that each package is indexed.
+                foreach (var package in packages.GroupBy(p => p.PackageRegistration).Select(g => g.First()))
+                {
+                    _indexingService.UpdatePackage(package);
+                }
             }
         }
 
--- a/src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
@@ -120,11 +120,6 @@ namespace NuGetGallery
                 throw new ArgumentException(nameof(packages));
             }
 
-            if (packages.Select(p => p.PackageRegistrationKey).Distinct().Count() > 1)
-            {
-                throw new ArgumentException("All packages to update must have the same ID.", nameof(packages));
-            }
-
             await UpdatePackagesInBulkAsync(packages.Select(p => p.Key).ToList());
 
             if (updateIndex)
--- a/src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
@@ -1,0 +1,200 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Data.Entity;
+using System.Linq;
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+using NuGet.Versioning;
+
+namespace NuGetGallery
+{
+    public class PackageVulnerabilityService : IPackageVulnerabilityService
+    {
+        public PackageVulnerabilityService(
+            IEntitiesContext entitiesContext,
+            IPackageUpdateService packageUpdateService)
+        {
+            _entitiesContext = entitiesContext ?? throw new ArgumentNullException(nameof(entitiesContext));
+            _packageUpdateService = packageUpdateService ?? throw new ArgumentNullException(nameof(packageUpdateService));
+        }
+
+        private readonly IEntitiesContext _entitiesContext;
+        private readonly IPackageUpdateService _packageUpdateService;
+
+        public async Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wasWithdrawn)
+        {
+            if (vulnerability == null)
+            {
+                throw new ArgumentNullException(nameof(vulnerability));
+            }
+
+            using (var strategy = new SuspendDbExecutionStrategy())
+            using (var transaction = _entitiesContext.GetDatabase().BeginTransaction())
+            {
+                var packagesToUpdate = new HashSet<Package>();
+                UpdateVulnerabilityInternal(vulnerability, wasWithdrawn, packagesToUpdate);
+                await _entitiesContext.SaveChangesAsync();
+
+                if (packagesToUpdate.Any())
+                {
+                    await _packageUpdateService.UpdatePackagesAsync(packagesToUpdate.ToList());
+                }
+
+                transaction.Commit();
+            }
+        }
+
+        /// <summary>
+        /// Updates the database with <paramref name="vulnerability"/>.
+        /// </summary>
+        /// <param name="vulnerability">The <see cref="PackageVulnerability"/> to persist in the database.</param>
+        /// <param name="wasWithdrawn">Whether or not this vulnerability has been withdrawn.</param>
+        /// <param name="packagesToUpdate">The set of <see cref="Package"/>s affected by this operation that should be marked as updated.</param>
+        /// <returns>
+        /// If the vulnerability already exists, returns the existing vulnerability.
+        /// If the vulnerability does not already exist, returns <paramref name="vulnerability"/>.
+        /// </returns>
+        private void UpdateVulnerabilityInternal(PackageVulnerability vulnerability, bool wasWithdrawn, HashSet<Package> packagesToUpdate)
+        {
+            if (vulnerability == null)
+            {
+                throw new ArgumentNullException(nameof(vulnerability));
+            }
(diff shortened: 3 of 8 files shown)


src/NuGetGallery.Services/PackageManagement/IPackageVulnerabilityService.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public interface IPackageVulnerabilityService
+    {
+        Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wasWithdrawn);
```
loic-sharma: Please add a quick comment

loic-sharma: Side note, what do you think of renaming `wasWithdrawn` to just `withdraw`? I don't feel strongly about this.

scottbommarito: I am equally ambivalent, lol. `withdrawn` is shorter, however, so probably best to align on it.

scottbommarito: I am equally ambivalent, lol. `withdrawn` is shorter, however, so probably best to align on it.

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
```diff
@@ -120,11 +120,6 @@ public async Task UpdatePackagesAsync(IReadOnlyList<Package> packages, bool upda
                 throw new ArgumentException(nameof(packages));
             }
 
-            if (packages.Select(p => p.PackageRegistrationKey).Distinct().Count() > 1)
```
loic-sharma: Isn't this still necessary because of the `_indexingService.UpdatePackage(packages.First());` on line 133? See:

https://github.com/NuGet/NuGetGallery/pull/7661/files#diff-87dc49cd51c8b38e70e8ed822334d68dL133

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,200 @@
...
+    {
+        public PackageVulnerabilityService(
+            IEntitiesContext entitiesContext,
+            IPackageUpdateService packageUpdateService)
+        {
+            _entitiesContext = entitiesContext ?? throw new ArgumentNullException(nameof(entitiesContext));
+            _packageUpdateService = packageUpdateService ?? throw new ArgumentNullException(nameof(packageUpdateService));
+        }
+
+        private readonly IEntitiesContext _entitiesContext;
```
loic-sharma: Please move fields to before the constructor.

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,200 @@
...
+        /// <summary>
+        /// Updates the database with <paramref name="vulnerability"/>.
+        /// </summary>
+        /// <param name="vulnerability">The <see cref="PackageVulnerability"/> to persist in the database.</param>
+        /// <param name="wasWithdrawn">Whether or not this vulnerability has been withdrawn.</param>
+        /// <param name="packagesToUpdate">The set of <see cref="Package"/>s affected by this operation that should be marked as updated.</param>
+        /// <returns>
+        /// If the vulnerability already exists, returns the existing vulnerability.
+        /// If the vulnerability does not already exist, returns <paramref name="vulnerability"/>.
+        /// </returns>
```
loic-sharma: Looks like this returns comment is no longer up-to-date

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,200 @@
...
+            if (vulnerability == null)
+            {
+                throw new ArgumentNullException(nameof(vulnerability));
+            }
+
+            using (var strategy = new SuspendDbExecutionStrategy())
+            using (var transaction = _entitiesContext.GetDatabase().BeginTransaction())
+            {
+                var packagesToUpdate = new HashSet<Package>();
+                UpdateVulnerabilityInternal(vulnerability, wasWithdrawn, packagesToUpdate);
```
loic-sharma: Should `UpdateVulnerabilityInternal` just return `packagesToUpdate`? What's the benefit of passing around as a parameter?

scottbommarito: Yup, probably can do that. This code has been refactored several times and the nature of `packagesToUpdate` is a relic of that.

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,200 @@
...
+                throw new ArgumentNullException(nameof(vulnerability));
+            }
+
+            // Determine if we already have this vulnerability.
+            var existingVulnerability = _entitiesContext.Vulnerabilities
+                .Include(v => v.AffectedRanges)
+                .Include(v => v.AffectedRanges.Select(pv => pv.Packages))
+                .SingleOrDefault(v => v.GitHubDatabaseKey == vulnerability.GitHubDatabaseKey);
+
+            var rangeComparer = new RangeForSameVulnerabilityEqualityComparer();
```
loic-sharma: Should this be moved into the `else` since it's only used there?

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,200 @@
...
+            if (existingVulnerability == null)
```
loic-sharma: It might be nice to split this method up into bite size chunks. Something like:

```csharp
if (existingVulnerability == null)
{
  AddNewVulnerability();
  return;
}

if (wasWithdrawn || !vulnerability.AffectedRanges.Any())
{
  RemoveVulnerability();
  return;
}

UpdateVulnerability();
```

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,200 @@
...
+        /// Iterates through the <see cref="Package"/>s that could be vulnerable to <paramref name="range"/>.
+        /// If any of these packages have not been marked vulnerable and should be, mark them vulnerable and add them to <paramref name="packagesToUpdate"/>.
+        /// </summary>
+        /// <remarks>
+        /// It is not possible to query the database to only return packages satisfying the version range, so we must iterate through all <see cref="Package"/>s in the <see cref="PackageRegistration"/>.
+        /// </remarks>
+        private void ProcessNewVulnerabilityRange(VulnerablePackageVersionRange range, HashSet<Package> packagesToUpdate)
+        {
+            var versionRange = VersionRange.Parse(range.PackageVersionRange);
+            var packages = _entitiesContext.PackageRegistrations
```
loic-sharma: Do we need to do any batching logic for packages with thousands of versions?

scottbommarito: Technically, it is possible that we could ingest so many vulnerabilities simultaneously that it could cause some server stress, and that we would need to throttle the job. I don't think this is necessary right now, however--we have so few vulnerabilities right now that we could ingest all of them simultaneously without any issue. Furthermore, in the several weeks I've been working on the feature, there's only been a single new vulnerability added. I would consider this a task for the backlog in case we notice it is causing issues.

loic-sharma: Hm, we grab all packages on the display package page: https://github.com/NuGet/NuGetGallery/blob/1accbd935db4142d20c7d703c0cb0adabb68daea/src/NuGetGallery.Core/Services/CorePackageService.cs#L288

The display package is much, much more frequent than this code path. Feel free to ignore my comment.

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,200 @@
...
+            if (existingVulnerability == null)
+            {
+                // We do not yet have this vulnerability. We should add it if it is not withdrawn and could affect at least one package.
+                // If the vulnerability does not have any vulnerable ranges, it cannot affect any packages.
+                // Even if no packages are currently vulnerable to the vulnerability, as long as it has a vulnerable range, 
+                // there is at least one package that could be uploaded that would be vulnerable to it.
+                if (!wasWithdrawn && vulnerability.AffectedRanges.Any())
+                {
+                    _entitiesContext.Vulnerabilities.Add(vulnerability);
+                    _entitiesContext.VulnerableRanges.AddRange(vulnerability.AffectedRanges);
```
skofman1: if EF populates VulnerableRanges automatically, we should drop this line

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
scottbommarito: Good catch!

`_indexingService` no-ops unless the service is local, but we need to call it for the first package of each registration.

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
scottbommarito: At one time it was used in both lol...moved it back!

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
scottbommarito: I believe it does, but I don't want to rely too much on the EF magic here. It is very unclear how it manages these relationships, and we need to be very careful here to properly diff the new and old vulnerabilities.

src/NuGetGallery.Services/PackageManagement/IPackageVulnerabilityService.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public interface IPackageVulnerabilityService
+    {
+        Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool withdrawn);
```
loic-sharma: Don't forget to add comments 😊

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,224 @@
...
+        /// <summary>
+        /// Updates the database with <paramref name="vulnerability"/>.
+        /// </summary>
+        /// <param name="vulnerability">The <see cref="PackageVulnerability"/> to persist in the database.</param>
+        /// <param name="withdrawn">Whether or not this vulnerability has been withdrawn.</param>
+        /// <param name="packagesToUpdate">The set of <see cref="Package"/>s affected by this operation that should be marked as updated.</param>
+        /// <returns>
+        /// The set of packages that were updated as part of this operation.
+        /// </returns>
+        private HashSet<Package> UpdateVulnerabilityInternal(PackageVulnerability vulnerability, bool withdrawn)
```
loic-sharma: Considering how infrequent this code path is, log volume isn't a concern. Can we add detailed logging?

scottbommarito: I knew there was something I forgot to add to the job...will do.

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
scottbommarito: it has been chunked

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
```diff
@@ -120,17 +120,16 @@ public async Task UpdatePackagesAsync(IReadOnlyList<Package> packages, bool upda
...
-
             await UpdatePackagesInBulkAsync(packages.Select(p => p.Key).ToList());
 
             if (updateIndex)
             {
-                // The indexing service will find the latest version of the package to index--it doesn't matter what package we pass in.
-                _indexingService.UpdatePackage(packages.First());
+                // The indexing service will find the latest version of a package to index--it doesn't matter what package we pass in.
+                // We do, however, need to pass in a single package for each registration to ensure that each package is indexed.
+                foreach (var package in packages.GroupBy(p => p.PackageRegistration).Select(g => g.First()))
```
loic-sharma: Please add a unit test for this bad boy

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,275 @@
...
+                    _logger.LogInformation("Vulnerability with GitHub key {GitHubDatabaseKey} had its metadata updated", vulnerability.GitHubDatabaseKey);
+                    packagesToUpdate.UnionWith(vulnerablePackages);
+                }
+
+                UpdateRangesOfPackageVulnerability(vulnerability, existingVulnerability, packagesToUpdate);
+            }
+        }
+
+        /// <returns>
+        /// Whether not the metadata of the existing vulnerability was changed.
```
xavierdecoster:
```suggestion
        /// <c>True</c> when the metadata of the existing vulnerability was changed; otherwise <c>false</c>.
```

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
scottbommarito: Test added!

PR #7667, opened 2019-10-31
Removes the height differences when switching to the organization page

robertmuehsig:
I was browsing through the NuGet UI and notices a UI glitch on the organization page:

before

The "divider" is using the "ms-font-xxl", which enlarges the complete navigation bread crumb area. With a smaller version it keeps the same size as similar pages. 

after

Be aware:
There are (at least) two other pages with this xxl dividers, but they have some other controls on the same line, so... it would be more complicated to fix those.

image

image

Hope this helps :)


robertmuehsig:
Oh - that was fast :) Thanks and happy Halloween 👍

PR #7670, opened 2019-11-03
Add storage abstractions for list and snapshots

joelverhagen:
Add Azure Blob Storage enumeration and snapshot APIs.

This allows applications using the gallery-based storage abstractions to take snapshots of blobs if they one does not exist, which is a behavior supported by the V3 (NuGet.Services.Metadata) storage abstractions.

In particular, I am working on rewriting Catalog2Registration and I am aligning on the gallery storage abstractions. Current Catalog2Registration uses the V3 storage abstractions and ensures every blob has at least one snapshot to mitigate accidental deletion. In general the NuGetGallery.Core storage abstractions (`ISimpleCloudBlob` and friends) are better for mocking because they more closely match the shape of the underlying Azure Storage APIs. Many non-gallery jobs already use the gallery storage abstractions: Azure Search, validation pipeline.

Until we have aligned all of the code on a single set of storage abstractions we're going to have this situation where capabilities are added lazily/piecemeal.

### DIFF
--- a/src/NuGetGallery.Core/Services/BlobResultSegmentWrapper.cs
+++ b/src/NuGetGallery.Core/Services/BlobResultSegmentWrapper.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.WindowsAzure.Storage.Blob;
+
+namespace NuGetGallery
+{
+    public class BlobResultSegmentWrapper : ISimpleBlobResultSegment
+    {
+        public BlobResultSegmentWrapper(BlobResultSegment segment)
+        {
+            // For now, assume all of the blobs are block blobs. This library's storage abstraction only allows
+            // creation of block blobs so it's good enough for now. If another caller created a non-block blob, this
+            // cast will fail at runtime.
+            Results = segment.Results.Cast<CloudBlockBlob>().Select(x => new CloudBlobWrapper(x)).ToList();
+            ContinuationToken = segment.ContinuationToken;
+        }
+
+        public IReadOnlyList<ISimpleCloudBlob> Results { get; }
+        public BlobContinuationToken ContinuationToken { get; }
+    }
+}
--- a/src/NuGetGallery.Core/Services/ISimpleBlobResultSegment.cs
+++ b/src/NuGetGallery.Core/Services/ISimpleBlobResultSegment.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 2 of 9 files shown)


loic-sharma:
> In general the NuGetGallery.Core storage abstractions (ISimpleCloudBlob and friends) are better for mocking because they more closely match the shape of the underlying Azure Storage APIs. Many non-gallery jobs already use the gallery storage abstractions: Azure Search, validation pipeline.
>
> Until we have aligned all of the code on a single set of storage abstractions we're going to have this situation where capabilities are added lazily/piecemeal.

I wonder if we should align our code to Azure.Storage.Blobs SDK once that's released, as these are mockable.

joelverhagen:
> I wonder if we should align our code to Azure.Storage.Blobs SDK once that's released, as these are mockable.

Yeah, I think that would be a good north star. I haven't looked at the API delta to understand what it would take to move. Like I don't know if we'd still want a concept of a storage abstraction at the `ISimpleCloudBlob` level. The file storage stuff in gallery probably would stay as-is but is too high level for many cases like V3/Azure Search.

I guess it's something to keep in the back of our mind.

src/NuGetGallery.Core/Services/ISimpleBlobResultSegment.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.WindowsAzure.Storage.Blob;
+
+namespace NuGetGallery
+{
+    public interface ISimpleBlobResultSegment
```
cristinamanum: Why Simple in the name? ICloudBlobResultSegment maybe?

joelverhagen: To mimic the interface name of the contained item `ISimpleCloudBlob`.

src/NuGetGallery.Core/Services/BlobResultSegmentWrapper.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace NuGetGallery
+{
+    public class BlobResultSegmentWrapper : ISimpleBlobResultSegment
+    {
+        public BlobResultSegmentWrapper(BlobResultSegment segment)
+        {
+            // For now, assume all of the blobs are block blobs. This library's storage abstraction only allows
+            // creation of block blobs so it's good enough for now. If another caller created a non-block blob, this
+            // cast will fail at runtime.
+            Results = segment.Results.Cast<CloudBlockBlob>().Select(x => new CloudBlobWrapper(x)).ToList();
```
cristinamanum: Suggestion: if this does a network call I would move it to the Results below.

joelverhagen: It does not. The HTTP call is made prior to this constructor (in an `async` path) and the implementation actually passes a `List` to this constructor. I don't know why they made the constructor accept `IEnumerable`.

PR #7676, opened 2019-11-05
[Token scanning-Phase 1-2] Select, revoke and block.

zhhyu:
Enable the Admin to revoke the selected API keys:
1. Expire the API key;
2. Set the revokedby flag;

Revoke:
image
Done:
image

Migration SQL:
`ALTER TABLE [dbo].[Credentials] ADD [RevocationSourceKey] [int]`


src/NuGet.Services.Entities/Credential.cs
```diff
@@ -82,6 +82,8 @@ public Credential(string type, string value, TimeSpan? expiration)
 
         public DateTime? LastUsed { get; set; }
 
+        public CredentialRevokedByType? RevokedBy { get; set; }
```
joelverhagen: nit: suffix this column name with "Key" as we did with `SemVerLevelKey` and `PackageStatusKey` (all enums) so that if we want to normalize the data in the future the name will match.

src/NuGet.Services.Entities/CredentialRevokedByType.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.Entities
+{
+    /// <summary>
+    /// Specify the type about where the ApiKey is leaked
+    /// </summary>
+    public enum CredentialRevokedByType
```
joelverhagen: What about something more noun-like such as `CredentialRevocationSource`?

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -2,22 +2,29 @@
...
 using NuGetGallery.Areas.Admin.ViewModels;
+using NuGetGallery.Areas.Admin.Models;
 
 namespace NuGetGallery.Areas.Admin.Controllers
 {
     public class ApiKeysController : AdminControllerBase
     {
         private readonly IAuthenticationService _authenticationService;
+        private readonly ITelemetryService _telemetryService;
+        internal const int MaxAllowedVerifyQueryLength = 3000;
```
joelverhagen: If we want a max, please provide data on how you picked it. I think we should just remove this max though. It's an admin tool and IIS already has a max request body length.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -51,11 +63,29 @@ public JsonResult Verify(string verifyQuery)
...
+                        results.Add(new ApiKeyRevokeViewModel(apiKeyViewModel, apiKey, leakedUrl: null, revokedBy: null, isRevocable: false));
+                        continue;
+                    }
+                    if (!IsRevocable(apiKeyViewModel))
+                    {
+                        results.Add(new ApiKeyRevokeViewModel(apiKeyViewModel, apiKey, leakedUrl: null, revokedBy: apiKeyViewModel.RevokedBy, isRevocable: false));
+                        continue;
+                    }
+
+                    var supportedRevokedByType = Enum.GetNames(typeof(CredentialRevokedByType));
```
joelverhagen: >Enum [](start = 49, length = 4)

You can use `Enum.TryParse`

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -66,17 +96,61 @@ public JsonResult Verify(string verifyQuery)
...
+                catch(Exception e)
+                {
+                    failedApiKeys.Add($"{apiKeyInfo.ApiKey}");
+                    _telemetryService.TraceException(e);
+                }
+            }
+
+            if (failedApiKeys.Any())
+            {
+                TempData["ErrorMessage"] = $"Failed to revoke the API key: { string.Join(", ", failedApiKeys.ToArray()) }." +
```
joelverhagen: >key [](start = 70, length = 3)

`API key(s)`

src/NuGetGallery/Areas/Admin/Models/RevokeApiKeysRequest.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+
+namespace NuGetGallery.Areas.Admin.Models
+{
+    public class RevokeApiKeysRequest
+    {
+        public List<string> SelectedApiKeyRevokeViewModelsInJSON { get; set; }
```
joelverhagen: What about this is JSON? The content-encoding of the request body is unrelated to the model. A simpler property name like "SelectedApiKeys" might be better.

src/NuGetGallery/Areas/Admin/Models/RevokeApiKeysRequest.cs
```diff
@@ -0,0 +1,18 @@
...
+
+        [Required(ErrorMessage = "Please sign using your name.")]
+        [StringLength(1000)]
+        [Display(Name = "Signature")]
+        public string Signature { get; set; }
```
joelverhagen: How is the signature used? It should be audited right?

joelverhagen:
Could you provide the SQL that the migration generates?

Also make sure the migration is generated on top of latest dev branch since Scott recently added a migration and these things need to be added one on top of another.

src/NuGetGallery/Areas/Admin/Models/RevokeApiKeysRequest.cs
zhhyu: It will be audited as others.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -66,31 +89,78 @@ public JsonResult Verify(string verifyQuery)
...
+        public async Task<ActionResult> Revoke(RevokeApiKeysRequest revokeApiKeysRequest)
+        {
+            if (revokeApiKeysRequest == null)
+            {
+                TempData["ErrorMessage"] = "The API keys revoking request can not be null.";
+                return View(nameof(Index));
+            }
+
+            var failedApiKeys = new List<string>();
+            foreach (var selectedApiKey in revokeApiKeysRequest.SelectedApiKeys)
```
skofman1: Can SelectedApiKeys be null?

zhhyu: Consider the front end side, this value in this request will never be null, because the form can be only submitted when it's not null. But yeah, maybe a null check is better to be added here.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -66,31 +89,78 @@ public JsonResult Verify(string verifyQuery)
...
+            {
+                var apiKeyInfo = JsonConvert.DeserializeObject<ApiKeyRevokeViewModel>(selectedApiKey);
+                try
+                {
```
skofman1: You should check "IsRevocable" again... don't trust the client.

src/NuGetGallery.Services/Authentication/AuthenticationService.cs
```diff
@@ -164,6 +164,18 @@ public Credential GetApiKeyCredential(string apiKey)
...
+        public async Task RevokeCredential(Credential credential, CredentialRevokedByType revokedBy)
+        {
+            if (credential == null)
+            {
+                throw new ArgumentNullException(nameof(credential));
+            }
+
+            credential.Expires = _dateTimeProvider.UtcNow;
+            credential.RevokedBy = revokedBy;
+            await Entities.SaveChangesAsync();
```
skofman1: if we are revoking 50 API Keys, u will call SaveChangesAsync 50 times. This will cause unnecessary perf impact. The recommended pattern here is to provide a bool param to the method that will determine if it needs to save to DB or not. It's a common pattern in the code base. [https://github.com/NuGet/NuGetGallery/blob/e3ae678fd8bc9f1065a17ccfb0ab3be82328bf55/src/NuGetGallery.Services/PackageManagement/PackageOwnershipManagementService.cs#L35](https://github.com/NuGet/NuGetGallery/blob/e3ae678fd8bc9f1065a17ccfb0ab3be82328bf55/src/NuGetGallery.Services/PackageManagement/PackageOwnershipManagementService.cs#L35)

src/NuGetGallery.Services/Authentication/AuthenticationService.cs
```diff
@@ -164,6 +164,18 @@ public Credential GetApiKeyCredential(string apiKey)
...
         }
 
+        public async Task RevokeCredential(Credential credential, CredentialRevokedByType revokedBy)
+        {
+            if (credential == null)
+            {
+                throw new ArgumentNullException(nameof(credential));
+            }
+
+            credential.Expires = _dateTimeProvider.UtcNow;
```
skofman1: This method shouldn't trust the caller (it's public, all things public shouldn't trust callers). Before revoking, it should check that the Credential is not already expired, is acutally an API Key, etc.

src/NuGetGallery.Services/Authentication/AuthenticationService.cs
```diff
@@ -593,7 +605,8 @@ public virtual CredentialViewModel DescribeCredential(Credential credential)
                         s.Subject,
                         NuGetScopes.Describe(s.AllowedAction)))
                     .ToList(),
-                ExpirationDuration = credential.ExpirationTicks != null ? new TimeSpan?(new TimeSpan(credential.ExpirationTicks.Value)) : null
+                ExpirationDuration = credential.ExpirationTicks != null ? new TimeSpan?(new TimeSpan(credential.ExpirationTicks.Value)) : null,
+                RevokedBy = credential.RevokedBy != null ? Enum.GetName(typeof(CredentialRevokedByType), credential.RevokedBy) : null
```
agr: Add trailing comma, please. This will make subsequent PRs a bit cleaner.

src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
```diff
@@ -28,5 +28,11 @@ public interface IAuthenticationService
         /// </summary>
         /// <returns>Return a credential if there is a matched apiKey or null if there is not.</returns>
         Credential GetApiKeyCredential(string apiKey);
+
+        /// <summary>
+        /// Revoke the credential
+        /// </summary>
+        /// <returns>Revoke and expire the credential.</returns>
```
agr: That's not the return value, is it? Maybe drop the line completely?

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -66,31 +89,78 @@ public JsonResult Verify(string verifyQuery)
...
+            if (revokeApiKeysRequest == null)
+            {
+                TempData["ErrorMessage"] = "The API keys revoking request can not be null.";
+                return View(nameof(Index));
+            }
+
+            var failedApiKeys = new List<string>();
+            foreach (var selectedApiKey in revokeApiKeysRequest.SelectedApiKeys)
+            {
+                var apiKeyInfo = JsonConvert.DeserializeObject<ApiKeyRevokeViewModel>(selectedApiKey);
```
agr: Why are we parsing JSON here? Why can't the front end make a request that would be parsed by the ASP.NET?

src/NuGetGallery.Services/Authentication/IAuthenticationService.cs
zhhyu: Sorry, that's my bad.

src/NuGetGallery.Services/Authentication/AuthenticationService.cs
zhhyu: Added

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
zhhyu: Sink the check to the called method, and share the same code/logic to check revocable or not in order to avoid divergence.

zhhyu:
Migrations will be updated after we finalize the solution towards the unsupported Index problem from EF6.2.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -66,31 +90,64 @@ public JsonResult Verify(string verifyQuery)
...
+            var failedApiKeys = new List<string>();
+            foreach (var selectedApiKey in revokeApiKeysRequest.SelectedApiKeys)
             {
-                return false;
+                var apiKeyInfo = JsonConvert.DeserializeObject<ApiKeyRevokeViewModel>(selectedApiKey);
+                try
+                {
+                    var apiKeyCredential = _authenticationService.GetApiKeyCredential(apiKeyInfo.ApiKey);
+                    var revocationSourceKey = (CredentialRevocationSource) Enum.Parse(typeof(CredentialRevocationSource), apiKeyInfo.RevocationSource);
+                    await _authenticationService.RevokeApiKeyCredential(apiKeyCredential, revocationSourceKey);
```
joelverhagen: This should call `commitChanges: false` and only commit at the end of the HTTP request.

src/NuGetGallery.Services/Authentication/AuthenticationService.cs
```diff
@@ -164,6 +164,59 @@ public Credential GetApiKeyCredential(string apiKey)
...
+                throw new InvalidOperationException(string.Format(
+                    CultureInfo.CurrentCulture,
+                    ServicesStrings.RevokeCredential_UnrevocableApiKeyCredential,
+                    apiKeyCredentialViewModel.Key));
+            }
+
+            await RevokeCredential(apiKeyCredential, revocationSourceKey, commitChanges);
+        }
+
+        public bool IsRevocableApiKeyCredential(CredentialViewModel credentialViewModel)
```
skofman1: A service (unlike a controller) is pure business logic, and shouldn't have dependencies on ViewModels. 

IsRevocableApiKeyCredential should accepts a Credential as input.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -66,31 +91,55 @@ public JsonResult Verify(string verifyQuery)
...
+            try
             {
-                return false;
+                foreach (var selectedApiKey in revokeApiKeysRequest.SelectedApiKeys)
+                {
+                    var apiKeyInfo = JsonConvert.DeserializeObject<ApiKeyRevokeViewModel>(selectedApiKey);
+
+                    var apiKeyCredential = _authenticationService.GetApiKeyCredential(apiKeyInfo.ApiKey);
+                    var revocationSourceKey = (CredentialRevocationSource)Enum.Parse(typeof(CredentialRevocationSource), apiKeyInfo.RevocationSource);
+                    await _authenticationService.RevokeApiKeyCredential(apiKeyCredential, revocationSourceKey, false);
```
joelverhagen: name boolean params

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
```diff
@@ -66,31 +91,55 @@ public JsonResult Verify(string verifyQuery)
...
+                }
+
+                await _entitiesContext.SaveChangesAsync();
```
skofman1: you should either pass commitChanges=true to RevokeApiKeyCredential for the last API Key instead of calling SaveChangesAsync here, or delete that param since you always use false...

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
zhhyu: The issue is that the form data here is fetched from AJAX and binded using knockout.js , rather than through Razor or ASP.NET. So the server side may not be able to bind the from data with the correct model as expected.

src/NuGetGallery/Areas/Admin/Controllers/ApiKeysController.cs
zhhyu: Talk offline.

Issue #7567, opened 2019-09-18
[UX] Unlisting package redirects to package details page


loic-sharma:
### Is your feature request related to a problem? Please describe.
When I unlist a package on nuget.org, I am redirected to my package's display package page. This makes it difficult to bulk unlist/relist packages (see https://github.com/NuGet/NuGetGallery/issues/6959).

### Describe the solution you'd like
Don't redirect to the Package Details page, instead, stay on the Manage Package page.

/cc @anangaur

skofman1:
Related issue: https://github.com/NuGet/NuGetGallery/issues/7495

Issue #7495, opened 2019-08-29
Deprecating package on nuget.org jumps to the package details page after completing deprecation


anangaur:
Related Epic issue: #6697 

### Describe the bug
Deprecating package on nuget.org jumps to the package details page after completing deprecation

### To Reproduce
1. Go to package `Manage` \ `Deprecation`
2. Deprecate one or more version of the package
3. Click on `Save`

### Expected behavior
Package deprecation should finish. Have a banner stating the versions that were deprecated on the `Manage` page.

### Actual behavior
Upon `Save`, I am redirected to the package details page.

PR #7678, opened 2019-11-05
Manage Package page forms should link back to the Manage page

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7567
https://github.com/nuget/nugetgallery/issues/7495

Currently, when you submit a form on the Manage Package page (Deprecation, Listing, Documentation) you are linked back to the package details page. This is annoying when you are trying to do other management operations at the same time.

This changes it so that submitting the forms stays on the Manage Package page. Additionally, I added "TempData" messages so that it's clearer what operation was performed.

### DIFF
--- a/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
+++ b/src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
@@ -53,6 +53,13 @@ namespace NuGetGallery
                 return Json(error.Status, new { error = error.Message });
             }
 
+            var packagePluralString = request.Versions.Count() > 1 ? "packages have" : "package has";
+            var deprecatedString = request.IsLegacy || request.HasCriticalBugs || request.IsOther
+                ? "deprecated" : "undeprecated";
+            TempData["Message"] = 
+                $"Your {packagePluralString} been {deprecatedString}. " +
+                $"It may take several hours for this change to propagate through our system.";
+
             return Json(HttpStatusCode.OK);
         }
     }
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -1905,8 +1905,39 @@ namespace NuGetGallery
         [ValidateAntiForgeryToken]
         public virtual async Task<ActionResult> UpdateListed(string id, string version, bool? listed)
         {
-            // Edit does exactly the same thing that Delete used to do... REUSE ALL THE CODE!
-            return await Edit(id, version, listed, Url.Package);
+            var package = _packageService.FindPackageByIdAndVersionStrict(id, version);
+            if (package == null)
+            {
+                return HttpNotFound();
+            }
+
+            if (ActionsRequiringPermissions.EditPackage.CheckPermissionsOnBehalfOfAnyAccount(GetCurrentUser(), package) != PermissionsCheckResult.Allowed)
+            {
+                return HttpForbidden();
+            }
+
+            if (package.PackageRegistration.IsLocked)
+            {
+                return new HttpStatusCodeResult(403, string.Format(CultureInfo.CurrentCulture, Strings.PackageIsLocked, package.PackageRegistration.Id));
+            }
+
+            string action;
+            if (!(listed ?? false))
+            {
+                action = "unlisted";
+                await _packageUpdateService.MarkPackageUnlistedAsync(package);
+            }
+            else
+            {
+                action = "listed";
+                await _packageUpdateService.MarkPackageListedAsync(package);
+            }
+            TempData["Message"] = string.Format(
+                CultureInfo.CurrentCulture,
+                "The package has been {0}. It may take several hours for this change to propagate through our system.",
+                action);
+
+            return Redirect(Url.ManagePackage(new TrivialPackageVersionModel(package)));
         }
 
         [UIAuthorize]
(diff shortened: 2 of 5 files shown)


tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -3627,15 +3627,14 @@ public async Task UpdatesUnlistedIfSelected(User currentUser, User owner)
...
 
                 // Act
-                var result = await controller.Edit("Foo", "1.0", listed: false, urlFactory: (pkg, relativeUrl) => @"~\Bar.cshtml");
+                var result = await controller.UpdateListed("Foo", "1.0", false);
 
                 // Assert
                 packageService.Verify();
                 Assert.IsType<RedirectResult>(result);
-                Assert.Equal(@"~\Bar.cshtml", ((RedirectResult)result).Url);
             }
```
zhhyu: Is that possible to verify the redirect Url here?

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -3647,6 +3646,7 @@ public async Task UpdatesUnlistedIfNotSelected(User currentUser, User owner)
                 {
                     PackageRegistration = new PackageRegistration { Id = "Foo" },
                     Version = "1.0",
+                    NormalizedVersion = "1.0.0",
```
zhhyu: What is the purpose to introduce this normalizedVersion here?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1969,9 +2000,11 @@ public virtual async Task<JsonResult> Edit(string id, string version, VerifyPack
                 }
             }
 
+            TempData["Message"] = "Your package's documentation has been updated.";
+
             return Json(new
             {
-                location = returnUrl ?? Url.Package(id, version)
+                location = returnUrl ?? Url.ManagePackage(new TrivialPackageVersionModel(id, version))
```
zhhyu: Regarding this change on the Edit method, do we have any test update to reflect this change?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1905,8 +1905,39 @@ public virtual ActionResult Edit(string id, string version)
         [ValidateAntiForgeryToken]
         public virtual async Task<ActionResult> UpdateListed(string id, string version, bool? listed)
         {
-            // Edit does exactly the same thing that Delete used to do... REUSE ALL THE CODE!
-            return await Edit(id, version, listed, Url.Package);
+            var package = _packageService.FindPackageByIdAndVersionStrict(id, version);
+            if (package == null)
+            {
```
zhhyu: Maybe it's better that we have the test coverage on this case.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1905,8 +1905,39 @@ public virtual ActionResult Edit(string id, string version)
...
+            }
+            else
+            {
+                action = "listed";
+                await _packageUpdateService.MarkPackageListedAsync(package);
+            }
+            TempData["Message"] = string.Format(
+                CultureInfo.CurrentCulture,
+                "The package has been {0}. It may take several hours for this change to propagate through our system.",
+                action);
```
zhhyu: Maybe it's better that we also verify this TempData in the tests.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
scottbommarito: `TrivialPackageViewModel(Package)` uses `Package.NormalizedVersion`, so we need to specify it or else the URL generation in the test throws a null-ref.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: This is already covered by tests. `UpdateListed` used to be a one-liner that called a function that was used nowhere else, so I moved all the code in the function up into `UpdateListed`.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
scottbommarito: We typically don't for `UrlHelper.*` redirects, because the way the URLs are generated in the tests is sometimes unexpected or broken.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: No, as mentioned above, because `UrlHelper.*` calls are difficult to verify.

src/NuGetGallery/Controllers/PackagesController.cs
scottbommarito: Actually, good call, we don't have any tests for missing package here. I can add one.

src/NuGetGallery/Controllers/ManageDeprecationJsonApiController.cs
```diff
@@ -53,6 +53,13 @@ public virtual JsonResult GetAlternatePackageVersions(string id)
                 return Json(error.Status, new { error = error.Message });
             }
 
+            var packagePluralString = request.Versions.Count() > 1 ? "packages have" : "package has";
+            var deprecatedString = request.IsLegacy || request.HasCriticalBugs || request.IsOther
+                ? "deprecated" : "undeprecated";
+            TempData["Message"] =
```
zhhyu: Maybe here also verify the TempData in the test? :smile:

scottbommarito: Already did haha: https://github.com/NuGet/NuGetGallery/pull/7678/files#diff-d8fa5513996103b3cf41923124b0defbR109

Issue #7649, opened 2019-10-25
Newly pushed packages that are vulnerable should be marked vulnerable on upload/push


scottbommarito:

scottbommarito:
Verified that DB is not affected without job running.

Needs to be re-verified once job is running and emitting vulnerabilities.

drewgillies:
Having consulted GitHub engineering, this is a highly unlikely scenario and verification can be addressed if we have a specific case to pursue.

drewgillies:
I've reviewed code per discussions and I think this is good to close. Verification in PROD will need to take place when this situation occurs for a customer.

drewgillies:
For peace of mind I ran some test scenarios (both for adding a package within a vulnerable range and adding one just outside it) and this works fine. FYI @skofman1

PR #7679, opened 2019-11-05
Newly pushed packages that are vulnerable should be marked vulnerable on upload/push

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7649


src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -28,6 +28,29 @@ public class PackageVulnerabilityService : IPackageVulnerabilityService
...
+        public void ApplyExistingVulnerabilitiesToPackage(Package package)
+        {
+            if (package == null)
+            {
+                throw new ArgumentNullException(nameof(package));
+            }
+
+            var version = NuGetVersion.Parse(package.NormalizedVersion);
+            var possibleRanges = _entitiesContext.VulnerableRanges
+                .Where(r => r.PackageId == package.Id)
```
skofman1: case sensitivity issues?

skofman1: do we have an index on this column? Full table scan isn't a big deal right now, but it will be good to be ready for the future.

scottbommarito: The collation in the DB causes this check to be done case-insensitively.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -794,6 +795,8 @@ private PackageValidationResult CheckRepositoryMetadata(PackageMetadata packageM
                 }
             }
 
+            _vulnerabilityService.ApplyExistingVulnerabilitiesToPackage(package);
```
joelverhagen: Why did you choose here? If we put it in a shared path with reflow then we get an easy fix-up tool for free.

scottbommarito: A couple thoughts:
- I considered vulnerability metadata to be more similar to reserved namespace metadata, so I put it where reserved namespace metadata is initialized.
- Reflow only shares `PackageService.EnrichPackageFromNuGetPackage` with the upload flow, where updating vulnerability metadata doesn't make sense to do.
- Reflow does not currently refresh reserved namespace metadata.
- The current vulnerability update path, as proposed by this PR, does not remove any existing vulnerabilities that no longer apply--I would argue doing this is a required step for reflow.

Given all this, I believe reflowing vulnerability information is outside of the scope of this PR, although it's something we should consider in the future.

joelverhagen: Okay, sounds good. We should plan the administrative scenarios as a separate thing then:

1. Package has vulnerability it shouldn't
1. Package doesn't have vulnerability it should

At the very least a playbook with SQL or whatever.

skofman1: I think we should add this to the reflow... playbook with SQL is pain... Scott, can u track an issue?

scottbommarito: I don't think it should be a part of reflow--I think there should be some sort of admin dashboard.  This dashboard should allow us to add, edit, and delete vulnerabilities and the packages they apply to. I think the likelihood of a package's vulnerabilities getting out of sync is low--I think it's much more likely that we will want to hide a vulnerability from being visible, or update its metadata, which reflow would provide us anyway. Either way, out of scope of this PR.

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
scottbommarito: I don't believe there is one right now. Definitely a good call though.

src/NuGetGallery/Services/PackageUploadService.cs
scottbommarito: https://github.com/nuget/nugetgallery/issues/7683

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
scottbommarito: Added:
```
CREATE INDEX [IX_PackageId] ON [dbo].[VulnerablePackageVersionRanges]([PackageId])
```

src/NuGetGallery/Migrations/201911072143480_AddVulnerablePackageVersionRangeIdIndex.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddVulnerablePackageVersionRangeIdIndex : DbMigration
```
joelverhagen: Please coordinate with @zhhyu to make sure migrations are generated and applied sequence. He has a PR out with a migration.

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
```diff
@@ -28,6 +28,29 @@ public class PackageVulnerabilityService : IPackageVulnerabilityService
...
         }
 
+        public void ApplyExistingVulnerabilitiesToPackage(Package package)
+        {
+            if (package == null)
+            {
+                throw new ArgumentNullException(nameof(package));
+            }
+
+            var version = NuGetVersion.Parse(package.NormalizedVersion);
```
joelverhagen: nit: introduce locals as late as possible

src/NuGetGallery/Migrations/201911072143480_AddVulnerablePackageVersionRangeIdIndex.cs
scottbommarito: Good call! I think as long as we don't apply migrations to dev before all of these migration PRs are merged, we should be good. Note that my other PR out now also has a migration.

joelverhagen: they need to be generated in sequence. like if you merge first @zhhyu should regenerate his on top of yours

src/NuGetGallery.Services/PackageManagement/PackageVulnerabilityService.cs
scottbommarito: I declare it here so I don't have to call `NuGetVersion.Parse` in every iteration of the `for` loop.

joelverhagen: you can declare it outside the loop

the motivation is that a) it reduces cognitive load and b) theoretically memory can be held on to for less time

PR #7686, opened 2019-11-08
Ingest first patched version of security vulnerabilities from the GitHub API

scottbommarito:
We have decided we would like to ingest this field from GitHub as well.

Unfortunately, this means that we also need to introduce logic to compare the existing vulnerable range metadata to the updated vulnerable range metadata, because previously these could not change. If we find that a vulnerable range has been updated, we must mark all packages it applies to as updated as well, so the new first patched version will flow through V3.


src/NuGet.Services.Entities/VulnerablePackageVersionRange.cs
```diff
@@ -42,6 +42,9 @@ public VulnerablePackageVersionRange()
         [Required]
         public string PackageVersionRange { get; set; }
 
+        [StringLength(Constants.MaxPackageVersionLength)]
+        public string FirstPatchedPackageVersion { get; set; }
```
joelverhagen: How did we find this max? This string is coming from GitHub right?

scottbommarito: I assumed this maximum based on our existing limits. I guess it's technically true that GitHub hasn't given us an official max length on their version strings, but I think it's reasonable to expect that their version strings aren't longer than ours as they are explicitly referring to our packages. Furthermore, GitHub does not include any prerelease versions in their API, which even further reduces the possibility of a version that is too long.

joelverhagen: Okay seems fine.

Is it documented to exclude pre-release or there are no pre-release right now?

scottbommarito: The latter. Unfortunately, in general their docs are not very precise. I recently asked them about their behavior towards prerelease, specifically regarding their version ranges, so hopefully that will shed some light on this topic. I think we can update the logic later if necessary.

Issue #7689, opened 2019-11-11
Use MVC routes to point Curated feed routes to V2 controller


joelverhagen:
Curated feeds have been removed for quite some time now. To reduce code bloat and to share the server-side (in-memory) endpoint caches between V2 controller and Curated V2 controller, we can try to make all of the curated V2 routes point to the main V2 controller.

This would:

a) reduce memory consumption because the endpoint cache is per Controller/action.
b) make non-existent curated feeds appear the same as the microsoftdotnet framework, i.e. a curated feed containing all packages

We could reduce the code footprint of curated feeds down to just the route names!

Use this query to fetch distinct route patterns to test after the change:
```
requests
| where timestamp > ago(90d)
| where name startswith "GET OData"
| extend parsedUrl = parse_url(url)
| extend host = tostring(parsedUrl["Host"])
| extend path = tostring(parsedUrl["Path"])
| extend afterHost = substring(url, indexof(url, host) + strlen(host))
| extend afterPath = substring(afterHost, indexof(afterHost, path) + strlen(path))
| extend normalizedPath = replace(@"\(Id='?[^',]*'?(,Version='?[^'\)]*'?)?\)", "(Id='ID',Version='VERSION')", path)
| summarize sum(itemCount), any(afterHost) by name, normalizedPath
| order by sum_itemCount desc
```

PR #7690, opened 2019-11-11
Make server-side OData cache configurable

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/2773#issuecomment-551338259.

This introduces a feature flag and JSON configuration file to manipulating cache duration on OData endpoints.

My plan is to ship this to PROD and take the following steps:

1. Bring FindPackagesById/$count caching up to 60 seconds (to match with the non-count endpoint).
   - Wait 1 full business day
1. Bring Increase FindPackagesById and FindPackageById/$count caching up to 120 seconds
   - Wait 1 full business day
1. ... repeat until duration is 5 minutes
1. Wait 7 days and closely observe memory and search service impact (i.e. cache "misses").
   - Memory should not go out of control and cache misses should decrease

This is the state of cache durations today:

Endpoint | Controller | Cache Duration (seconds)
-- | -- | --
FindPackagesById | V1 | 60
FindPackagesById | V2 | 60
FindPackagesById | V2 Curated | 60
FindPackagesById/$count | V1 | 60
FindPackagesById/$count | V2 | 0
FindPackagesById/$count | V2 Curated | 0
Search | V1 | 45
Search | V2 | 45
Search | V2 Curated | 45
Search/$count | V1 | 45
Search/$count | V2 | 45
Search/$count | V2 Curated  | 45
Specific Package | V1 | 60
Specific Package | V2 | 60
Specific Package | V2 Curated | 60

Yes, today the FindPackagesById/$count endpoint is cached on V1 but not on V2 or V2 curated (???). Another oddity to note is that Search/$count is cached everywhere but FindPackagesById/$count is not (??????).

Another improvement that can be done is to update the curated feed routes to point to the main V2 controller so that the endpoint caches are shared. This is tracked here: https://github.com/NuGet/NuGetGallery/issues/7689.

Without the feature flag being turned on, the only impact of this change is to bring V1 $count endpoint into alignment with the other two controllers (i.e. disable server-side caching) The volume of this endpoint is very low so the impact will be unnoticable.

Client-side cache durations are unaffected by this PR. In general client-side cache durations are not important for us since official client does not observe them.

### DIFF
--- a/src/NuGetGallery/App_Data/Files/Content/OData-Cache-Configuration.json
+++ b/src/NuGetGallery/App_Data/Files/Content/OData-Cache-Configuration.json
@@ -1,0 +1,6 @@
+{
+  "FindPackagesByIdCacheTimeInSeconds": 60,
+  "FindPackagesByIdCountCacheTimeInSeconds": 0,
+  "GetSpecificPackageCacheTimeInSeconds": 60,
+  "SearchCacheTimeInSeconds": 45
+}
--- a/src/NuGetGallery/Controllers/ODataV1FeedController.cs
+++ b/src/NuGetGallery/Controllers/ODataV1FeedController.cs
@@ -12,6 +12,7 @@ using NuGet.Services.Entities;
 using NuGetGallery.Configuration;
 using NuGetGallery.OData;
 using NuGetGallery.OData.QueryFilter;
+using NuGetGallery.Services;
 using NuGetGallery.WebApi;
 using WebApi.OutputCache.V2;
 
@@ -76,7 +77,11 @@ namespace NuGetGallery.Controllers
 
         // /api/v1/Packages(Id=,Version=)
         [HttpGet]
-        [CacheOutput(ServerTimeSpan = NuGetODataConfig.GetByIdAndVersionCacheTimeInSeconds, Private = true, ClientTimeSpan = NuGetODataConfig.GetByIdAndVersionCacheTimeInSeconds)]
+        [ODataCacheOutput(
+            ODataCachedEndpoint.GetSpecificPackage,
+            serverTimeSpan: ODataCacheConfiguration.DefaultGetByIdAndVersionCacheTimeInSeconds,
+            Private = true,
+            ClientTimeSpan = ODataCacheConfiguration.DefaultGetByIdAndVersionCacheTimeInSeconds)]
         public async Task<IHttpActionResult> Get(ODataQueryOptions<V1FeedPackage> options, string id, string version)
         {
             var result = await GetCore(options, id, version, return404NotFoundWhenNoResults: true);
@@ -86,7 +91,11 @@ namespace NuGetGallery.Controllers
         // /api/v1/FindPackagesById()?id=
         [HttpGet]
         [HttpPost]
-        [CacheOutput(ServerTimeSpan = NuGetODataConfig.GetByIdAndVersionCacheTimeInSeconds, Private = true, ClientTimeSpan = NuGetODataConfig.GetByIdAndVersionCacheTimeInSeconds)]
+        [ODataCacheOutput(
+            ODataCachedEndpoint.FindPackagesById,
+            serverTimeSpan: ODataCacheConfiguration.DefaultGetByIdAndVersionCacheTimeInSeconds,
+            Private = true,
+            ClientTimeSpan = ODataCacheConfiguration.DefaultGetByIdAndVersionCacheTimeInSeconds)]
         public async Task<IHttpActionResult> FindPackagesById(ODataQueryOptions<V1FeedPackage> options, [FromODataUri]string id)
         {
             return await GetCore(options, id, version: null, return404NotFoundWhenNoResults: false);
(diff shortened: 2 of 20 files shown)


src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -86,15 +91,22 @@ public async Task<IHttpActionResult> Get(ODataQueryOptions<V1FeedPackage> option
         // /api/v1/FindPackagesById()?id=
         [HttpGet]
         [HttpPost]
-        [CacheOutput(ServerTimeSpan = NuGetODataConfig.GetByIdAndVersionCacheTimeInSeconds, Private = true, ClientTimeSpan = NuGetODataConfig.GetByIdAndVersionCacheTimeInSeconds)]
+        [ODataCacheOutput(
+            ODataCachedEndpoint.FindPackagesById,
+            ServerTimeSpan = ODataCacheConfiguration.DefaultGetByIdAndVersionCacheTimeInSeconds,
```
skofman1: Shouldn't it be FindPackagesByIdCacheTimeInSeconds config?

src/NuGetGallery/Infrastructure/ODataCacheOutputAttribute.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public TimeSpan ReloadDuration { get; set; } = TimeSpan.FromSeconds(60);
+
+        public override void OnActionExecuting(HttpActionContext actionContext)
+        {
+            // This is a very hot path, so cache the configuration and only reload every so often.
+            var now = DateTimeOffset.Now;
+            if (now - _configuredAt >= ReloadDuration)
+            {
+                if (_defaultServerTimeSpan == null)
```
cristinamanum: Any telemetry metrics needed?

src/NuGetGallery/Infrastructure/ODataCacheOutputAttribute.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// The <paramref name="configurationName"/> refers to an integer property name on
+        /// <see cref="IODataCacheConfiguration"/>. This property will be uses to override the
+        /// <see cref="CacheOutputAttribute.ServerTimeSpan"/>.
+        /// </summary>
+        public ODataCacheOutputAttribute(ODataCachedEndpoint endpoint)
+        {
+            _endpoint = endpoint;
+        }
+
+        public TimeSpan ReloadDuration { get; set; } = TimeSpan.FromSeconds(60);
```
cristinamanum: Is 60 related with any other configurations? Maybe add comments?

joelverhagen: Unrelated. I'll add a comment.

src/NuGetGallery/Infrastructure/ODataCacheOutputAttribute.cs
joelverhagen: Good idea. I'll add a trace to the reload case.

src/NuGetGallery/App_Data/Files/Content/OData-Cache-Configuration.json
```diff
@@ -0,0 +1,6 @@
+{
+    "GetSpecificPackageCacheTimeInSeconds": 60,
+    "FindPackagesByIdCacheTimeInSeconds": 60,
```
loic-sharma: >FindPackagesByIdCacheTimeInSeconds [](start = 5, length = 34)

What do you think of ordering by lex order of keys here? I don't feel strongly about this.

joelverhagen: I don't care about the order, i will sort them lex-wise

skofman1:
@joelverhagen , what's the evidence that increased caching to 5 mins will help search service load? It makes sense (of course) that it will help, but do you have numbers to prove this.

joelverhagen:
> @joelverhagen , what's the evidence that increased caching to 5 mins will help search service load? It makes sense (of course) that it will help, but do you have numbers to prove this.

See the linked comment in the original description. I looked at request frequency per query string on the gallery side and ran some projections.

src/NuGetGallery/Infrastructure/ODataCacheOutputAttribute.cs
```diff
@@ -0,0 +1,142 @@
...
+                            "For OData caching, the server-side cache time of {Seconds} seconds will now be used for endpoint {Endpoint}.",
+                            secondsOrNull,
+                            _endpoint);
+                    }
+
+                    ServerTimeSpan = newServerTimeSpan;
+                    ResetCacheTimeQuery();
+                }
+
+                _configuredAt = now;
```
loic-sharma: I believe this is susceptible to struct tearing as DateTime is greater than a CPU word. If so, you will need synchronization to use a DateTime...

Is it possible to use a "push" model to update `ServerTimeSpan`?

loic-sharma: If you stick to this current model, I believe you will need memory barriers around `ServerTimeSpan` and `_configuredAt`.

joelverhagen: Yeah that's true. I thought about concurrency but figured it would just result in some sporadic duplicate reloads which I wasn't too concerned about. Note that all the dependencies read to evaluate the new time span are cached/in-memory things. I should just able about to make a reference to contain the `_configuredAt` and call it a day, I think.

loic-sharma: I'm not sure the reference approach would work as you need to encapsulate both `_configuredAt` and `ServerTimeSpan`. Remember that the CPU can and will reorder instructions, meaning that `_configuredAt` and `ServerTimeSpan` may unexpected states relative to each other. If you made `_configuredAt` a reference, you still need to add memory barriers...

joelverhagen: I'm not sure if I follow regarding the particular details. If the assignment is out of order what is the negative impact?

loic-sharma: Latest changes look good to me 👍

PR #7692, opened 2019-11-12
Allow admins to see users' MFA setting on profile page

scottbommarito:
As requested by @karann-msft 

image


src/NuGetGallery/Views/Users/Profiles.cshtml
```diff
@@ -39,6 +39,8 @@
                             <dt>Unconfirmed Email Address:</dt>
                             <dd>@Model.UnconfirmedEmailAddress</dd>
                         }
+                        <dt>Multi-factor Authentication</dt>
```
karann-msft:
```suggestion
                        <dt>Multi-factor Authentication:</dt>
```

PR #7700, opened 2019-11-18
[Token scanning-Phase 1-3] API keys profile page

zhhyu:
- API keys profile page shows the revoked API keys info.
image
Or
image

- The page to verify API keys shows "Link" rather than the original URL. 
image

### DIFF
--- a/src/Bootstrap/less/theme/page-api-keys.less
+++ b/src/Bootstrap/less/theme/page-api-keys.less
@@ -93,4 +93,8 @@
   {
     border-top: @section-border;
   }
+
+  .revoke-api-key {
+    color: red
+  }
 }
--- a/src/Bootstrap/less/theme/page-api-keys.less
+++ b/src/Bootstrap/less/theme/page-api-keys.less
@@ -95,6 +95,6 @@
   }
 
   .revoke-api-key {
-    color: red
+    color: @brand-danger
   }
 }
--- a/src/NuGetGallery/Views/Users/ApiKeys.cshtml
+++ b/src/NuGetGallery/Views/Users/ApiKeys.cshtml
@@ -86,7 +86,7 @@
         have
         <!-- /ko -->
         been revoked.<br />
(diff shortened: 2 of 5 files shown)


src/NuGetGallery/Views/Users/ApiKeys.cshtml
```diff
@@ -68,6 +68,35 @@
...
+        <!-- /ko -->
+        <!-- ko text: window.nuget.commaJoin(RevocationDescriptions().map(function(d) { return "'" + d + "'"; } )) --><!-- /ko -->
+        <!-- ko if: RevocationDescriptions().length === 1 -->
+        has
+        <!-- /ko -->
+        <!-- ko ifnot: RevocationDescriptions().length === 1 -->
+        have
+        <!-- /ko -->
+        been revoked.<br />
+        Please check the email from <<span>support@nuget.org</span>> for details, and regenerate or create
```
joelverhagen: The markup for this email is pretty weird. `&lt;support@nuget.org&gt;` is probably what you want. `&lt;` and `&gt;` are called HTML entities. By the way do we normally enclose our support email in angle brackets like this when display it? Seems a bit odd.

Is there a variable/config for `support@nuget.org`?

src/Bootstrap/less/theme/page-api-keys.less
```diff
@@ -93,4 +93,8 @@
   {
     border-top: @section-border;
   }
+
+  .revoke-api-key {
+    color: red
```
joelverhagen: Can you check if this is the error color we use in other contexts? Consider the color we use for `AlertError`.

zhhyu: Use "brand-danger" now.

src/NuGetGallery/Views/Users/ApiKeys.cshtml
zhhyu: Use Config variable and remove the angle brackets.

Issue #7685, opened 2019-11-07
Write verification script to verify vulnerability data in DB matches data in the GitHub API


scottbommarito:

drewgillies:
This will need to be re-inserted as a JobRunner job to cover the new credential requirements.

PR #7706, opened 2019-11-19
Add command to compare PackageVulnerabilities to GitHub API

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7685

### DIFF
--- a/src/GalleryTools/Commands/VerifyGitHubVulnerabilitiesCommand.cs
+++ b/src/GalleryTools/Commands/VerifyGitHubVulnerabilitiesCommand.cs
@@ -114,7 +114,7 @@ namespace GalleryTools.Commands
             string connectionString,
             IReadOnlyList<SecurityAdvisory> advisories)
         {
-            Console.Write("Fetching vulnerabilities from DB...");
+            Console.WriteLine("Fetching vulnerabilities from DB...");
 
             using (var sqlConnection = new SqlConnection(connectionString))
             {
@@ -161,7 +161,9 @@ namespace GalleryTools.Commands
                 {
                     if (existingVulnerability != null)
                     {
-                        Console.Error.WriteLine($"Vulnerability was withdrawn or affects no packages and should not be in DB!");
+                        Console.Error.WriteLine(withdrawn ? 
+                            $@"Vulnerability advisory {vulnerability.GitHubDatabaseKey} was withdrawn and should not be in DB!" :
+                            $@"Vulnerability advisory {vulnerability.GitHubDatabaseKey} affects no packages and should not be in DB!");
                         HasErrors = true;
                     }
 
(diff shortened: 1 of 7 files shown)


src/GalleryTools/Commands/VerifyGitHubVulnerabilitiesCommand.cs
```diff
@@ -0,0 +1,233 @@
...
+                    if (existingRange.FirstPatchedPackageVersion != range.FirstPatchedPackageVersion)
+                    {
+                        Console.Error.WriteLine($"First patched version does not match!");
+                        HasErrors = true;
+                    }
+
+                    var packages = _entitiesContext.Packages
+                        .Where(p => p.PackageRegistration.Id == range.PackageId)
+                        .Include(p => p.Vulnerabilities)
+                        .Include(p => p.Vulnerabilities)
```
agr: Why this line appear twice?

scottbommarito: Thought I would need to include another field but realized I didn't. Will remove.

joelverhagen:
When will this be run? Currently (AFAIK) the GalleryTools job is not run on a regular cadence. It's mostly used for ops tools.

Should this be run from the runner infrastructure (which is more of a monitoring thing) instead of from a gallery entry point (which is more for production code)?

xavierdecoster:
> When will this be run? Currently (AFAIK) the GalleryTools job is not run on a regular cadence. It's mostly used for ops tools.

I've asked @scottbommarito the same question... 

> Should this be run from the runner infrastructure (which is more of a monitoring thing) instead of from a gallery entry point (which is more for production code)?

Well, this particular tool is doing a full data integrity check AFAIK. Seems like a little overkill to run continuously. The monitoring job should likely focus on checking the delta between the latest advisory available from the GH API and the job's cursor.

xavierdecoster:
@joelverhagen I believe @scottbommarito's intention was to run this script post-deployment to verify the `GitHubVulnerabilities2Db` is working properly. This has already run on DEV, but needs to run against INT and PROD after deploying the job there.

drewgillies:
@joelverhagen @agr @loic-sharma @xavierdecoster @shishirx34 @ryuyu @cristinamanum @skofman1 @dannyjdev @zhhyu @lyndaidaii 
I have this up-and-running now. Would one or two of you please take a final look over. You can see my minor adjustments to @scottbommarito's code in https://github.com/NuGet/NuGetGallery/commit/02bc74ca648d2f173fca98b903a178e9cbdb23b9, but essentially it's the same--stderr just gives us more informative errors now (and the advisory type has apparently changed name since this branch was created).
The intent of this utility is to run on an ad hoc basis, and it functions fine with an AAD Auth connection string and GitHub token.

src/GalleryTools/Commands/VerifyGitHubVulnerabilitiesCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+            return advisories;
+        }
+
+        private static async Task VerifyPackageVulnerabilities(
+            string connectionString,
+            IReadOnlyList<SecurityAdvisory> advisories)
+        {
+            Console.WriteLine("Fetching vulnerabilities from DB...");
+
+            using (var sqlConnection = new SqlConnection(connectionString))
```
joelverhagen, line 119: I'm not sure where this will be run, but if we use the SQL connection factory that our product code uses, then we can use a KeyVault certificate + AAD to auth with SQL Server and a managed identity to auth with KeyVault, thus allowing us to run this on a job VM manually with a connection string exactly like the one gallery uses. If you have another plan for what connection string to use, don't worry switching the implementation. You could do something like this to avoid DI:
https://github.com/NuGet/NuGetGallery/compare/jver-slowquery#diff-32214f1a3dbc78bf52f5f5b31c65d139R39-R45

I took this approach for another one-off tool recently (not checked in to a main branch).

src/GalleryTools/Commands/VerifyGitHubVulnerabilitiesCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+            };
+
+            var queryService = new QueryService(
+                config,
+                new HttpClient());
+
+            var advisoryQueryService = new AdvisoryQueryService(
+                queryService,
+                new AdvisoryQueryBuilder(),
+                NullLogger<AdvisoryQueryService>.Instance);
```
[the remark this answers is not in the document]
drewgillies: It's exactly that--this tool is run once in a while for checking purposes. It won't be a monitoring job.

src/GalleryTools/Commands/VerifyGitHubVulnerabilitiesCommand.cs
drewgillies: I think I'd go down this track if it were something we would use a lot, but it's really only a sanity check utility. I'll add some docs in our onenote for how to run it using a connection string. I'll cycle back to this thought though if some issues arise with that approach. Thanks!

PR #7709, opened 2019-11-20
Dropped the `DiagnosticMonitorTraceListener` usage.

agr:
Progress on https://github.com/NuGet/Engineering/issues/1912

Removed some dead code. `DiagnosticMonitorTraceListener` needs to be configured, we don't have any configuration code in the project (or in web.config).

### DIFF
--- a/src/NuGetGallery/App_Start/AppActivator.cs
+++ b/src/NuGetGallery/App_Start/AppActivator.cs
@@ -15,7 +15,6 @@ using System.Web.Optimization;
 using System.Web.Routing;
 using System.Web.UI;
 using Elmah;
-using Microsoft.WindowsAzure.Diagnostics;
 using Microsoft.WindowsAzure.ServiceRuntime;
 using NuGetGallery;
 using NuGetGallery.Configuration;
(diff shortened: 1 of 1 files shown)


joelverhagen:
Are we sure this is unused? I think we should do some smoke tests to verify they is no regression.

src/NuGetGallery/App_Start/AppActivator.cs
```diff
@@ -110,11 +97,6 @@ private static RazorViewEngine CreateViewEngine()
             return ret;
         }
 
-        private static void CloudPreStart()
-        {
-            Trace.Listeners.Add(new DiagnosticMonitorTraceListener());
```
joelverhagen: This is a `Trace.WriteLine` (etc) sink, right? Where were the lines showing up?

agr: This is supposed to log to a blob storage table `WADLogsTable`, but in order to do that, it has to be [configured](https://docs.microsoft.com/en-us/azure/cloud-services/cloud-services-dotnet-diagnostics) either in the [code](https://stackoverflow.com/questions/4494040/azure-how-to-create-the-wadlogstable-for-capturing-diagnostics-code) (which we don't do) or on an [individual VMs](https://docs.microsoft.com/en-us/powershell/module/servicemanagement/azure/set-azureservicediagnosticsextension?view=azuresmps-4.0.0) (we would be aware if we are doing that).

Issue #7699, opened 2019-11-15
Setup instructions cannot create database(s)


festivegoose:
When I clone the project and go through the setup instructions, specifically running `.\tools\Setup-DevEnvironment.ps1` I get errors on the database creation steps:

```
[BEGIN] Running Migrations
Updating Nuget Gallery database...
ERROR: Method not found: 'System.Data.Entity.ModelConfiguration.Configuration.IndexConfiguration System.Data.Entity.ModelConfiguration.EntityTypeConfiguration`1.HasIndex(System.Linq.Expressions.Expression`1<System.Func`2<!0,!!0>>)'.
Updating Nuget Gallery Support request database...
ERROR: Method not found: 'System.Data.Common.DbConnection System.Data.Entity.Core.Common.DbProviderServices.CloneDbConnection(System.Data.Common.DbConnection)'.
Update Complete!
[DONE] Running Migrations
```

### To Reproduce
1. Clone project
2. Run `.\build`
3. Run `.\tools\Setup-DevEnvironment.ps1`

### Expected behavior
Expect database creation to occur so that project can run successfully.

scottbommarito:
Hey @festivegoose, I looked into the issue you're running into. It seems like, in that script (that we don't run very often), [we hardcoded a version of EntityFramework](https://github.com/NuGet/NuGetGallery/blob/master/tools/Update-Databases.ps1#L8) we were on for a long time. Recently we've updated our EF version a couple times, and are using a feature that was introduced in EF 6.2.0.

If you change the version that's hardcoded there to a version of EF that's 6.2 or higher, the script should work as expected. We're planning on upgrading to EF 6.4 soon (when it releases) so we'll update the script officially when it does.

festivegoose:
That worked @scottbommarito , thanks a bunch!

scottbommarito:
Glad that worked for you, @festivegoose!

I'm going to reopen this, though, so we'll remember to update the script when we stabilize the version.

joelverhagen:
Turns out the latest versions of EntityFramework 6, starting in 6.3 time, no longer have `migrate.exe`. They now have `ef6.exe`. The script will need to be changed to use this new tool, which doesn't seem to have docs yet:
https://github.com/aspnet/EntityFramework.Docs/issues/1740

joelverhagen:
I've updated the script to:

1. Automatically discovery the current version
2. Use `ef6.exe` instead of `migrate.exe`, the latter of which no longer exists in the package.

PR #7711, opened 2019-11-21
Use ef6.exe instead of migrate.exe

joelverhagen:
This addresses a breaking change introduced in 6.3.0 which eliminates migrate.exe and replaces it with ef6.exe.
Address https://github.com/NuGet/NuGetGallery/issues/7699
Migration details available here: https://github.com/aspnet/EntityFramework.Docs/issues/1740#issuecomment-557204757

I also removed some sad "Nuget" instances with "NuGet". You're welcome, marketing!

### DIFF
--- a/tools/Update-Databases.ps1
+++ b/tools/Update-Databases.ps1
@@ -1,65 +1,76 @@
 param(
     [parameter(Mandatory=$true)]
     [string[]] $MigrationTargets,
-    [string] $NugetGallerySitePath)
+    [string] $NuGetGallerySitePath)
 
-function Initialize-MigrateExe() {
+function Initialize-EF6Exe() {
     [string] $migrateDirectory = [System.IO.Path]::Combine($PSScriptRoot, '__temp_migrate_directory_' + [guid]::NewGuid().ToString("N") )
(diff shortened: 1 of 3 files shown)


tools/Update-Databases.ps1
```diff
@@ -1,65 +1,76 @@
...
     }
 
+    if (!$efDirectory) {
+        # Read the current version of EntityFramework from NuGetGallery.csproj so that we can find the tools.
+        $csprojPath = Join-Path $PSScriptRoot "..\src\NuGetGallery\NuGetGallery.csproj"
+        [xml]$csproj = Get-Content $csprojPath
+        $efPackageReference = Select-Xml -Xml $csproj -XPath "//*[local-name()='PackageReference']" `
+            | Where-Object { $_.Node.Attributes["Include"].Value -eq "EntityFramework" }
+        $efVersion = $efPackageReference.Node.Version
+        Write-Host "Using EntityFramework version $efVersion."
```
loic-sharma: Should we add something to `test.ps1` to verify that this trick works as expected? Otherwise, one may break the `Update-Databases.ps1` script by moving our EntityFramework dependency to some subproject.

joelverhagen: Sure, I can try that.

PR #7713, opened 2019-11-25
Deployment label log enrichment

agr:
Progress on https://github.com/NuGet/Engineering/issues/1912

Need it to differentiate between the deployments for app service conversion since `DeploymentIdTelemetryEnricher` will not work there. Unfortunately, this change will only work with App services, so for transition period we would need to look at both. I didn't find a way to pass this configuration bit to the cloud service configuration.


src/NuGetGallery/Telemetry/DeploymentLabelEnricher.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace NuGetGallery
+{
+    public class DeploymentLabelEnricher : ITelemetryInitializer
+    {
+        private const string DeploymentLabel = "DeploymentLabel";
+        private readonly string _deploymentLabel;
+
+        public DeploymentLabelEnricher(IAppConfiguration appConfiguration)
+        {
+            _deploymentLabel = appConfiguration.DeploymentLabel;
```
loic-sharma: Does this need a null check?

joelverhagen:
You wanna add this to Azure Search? :)

src/NuGetGallery/Telemetry/DeploymentLabelEnricher.cs
```diff
@@ -0,0 +1,47 @@
...
+using NuGetGallery.Configuration;
+
+namespace NuGetGallery
+{
+    public class DeploymentLabelEnricher : ITelemetryInitializer
+    {
+        private const string DeploymentLabel = "DeploymentLabel";
+        private readonly string _deploymentLabel;
+
+        public DeploymentLabelEnricher(IAppConfiguration appConfiguration)
```
xavierdecoster: Can this be updated so that it simply receives the `string deploymentLabel` instead of the whole `IAppConfiguration`?
This sounds like a useful type to add to `NuGet.Services.Logging` instead, so we can easily share it with non-gallery code.

agr: Good call. https://github.com/NuGet/ServerCommon/pull/322

Issue #7720, opened 2019-12-03
Rename PackageVulnerability's ReferenceUrl to AdvisoryUrl and make it required


xavierdecoster:
Current db model assumes this is optional. It is not. The term _reference_ also doesn't match the target object type for this URL, which is an _advisory_.

PR #7721, opened 2019-12-03
Rename PackageVulnerability's ReferenceUrl to AdvisoryUrl and make it required

xavierdecoster:
The `PackageVulnerability.ReferenceUrl` field is mandatory and should be renamed to `PackageVulnerability.AdvisoryUrl`. The current db model assumes it is optional. We should enforce this restriction in the db model.

In addition, the [protocol](https://github.com/NuGet/Home/wiki/Flag-vulnerable-packages#protocol) and ubiqituous language for this feature have been updated:
* `ReferenceUrl` is actually an `AdvisoryUrl` (which points to a _resource to learn more about the security advisory_, which in turn may have _references_ to e.g. NIST database)
* As we don't use the _SecurityAdvisoryReference_ data anywhere, we don't need to query these from GH's GraphQL API
* The GH GraphQL API will be updated to include a new `permalink` field on the `SecurityAdvisory` object, but that's still a few months out. In the mean time, we can work-around it by using the permalink-format provided to us, which purposely was put into an extension method with a comment.

Addresses https://github.com/NuGet/NuGetGallery/issues/7720

Rollout plan:
* [ ] Disable `GitHubVulnerabilities2Db` job on DEV and INT
* [ ] Apply EF migrations to NuGetGallery db on DEV and INT
* [ ] Deploy patched `GitHubVulnerabilties2Db` job to DEV and INT


src/GitHubVulnerabilities2Db/Ingest/AdvisoryIngestor.cs
```diff
@@ -28,6 +28,13 @@ public async Task IngestAsync(IReadOnlyList<SecurityAdvisory> advisories)
         {
             foreach (var advisory in advisories)
             {
+                if (!advisory.References.Any())
+                {
+                    // This should not happen, but we should guard against it anyway to avoid data integrity issues.
+                    // Ignore advisories that don't have any reference.
```
joelverhagen: Should we halt consumption in this case? Or raise an alert? Ignoring the data seems scary without some other detection.

skofman1: When Scott was writing the code we discussed how to handle input that doesn't match out assumptions. The decision was that the job should get "stuck" so that monitoring will catch this.

xavierdecoster: I can remove this specific check again and an exception would be thrown on [line 51](https://github.com/NuGet/NuGetGallery/pull/7721/files#diff-b57708f58ab2fe1e3e43ca96856f0a7aR51), which would make sure the job gets stuck.

In terms of monitoring, though, I think it'd be beneficial if we did catch this scenario here and send out particular telemetry right before getting stuck, so we can alert sooner. 
The monitoring for this particular job is based on calculating the delta between its cursor and the GH API latest updated, which would cause significant delay before we notice the issue (there can be a gap of multiple days between two vulnerability updates on GH side).

xavierdecoster: Actually, there's no need for these _SecurityAdvisoryReference_ objects at all... See updated PR description regarding protocol updates.

src/GitHubVulnerabilities2Db/GraphQL/SecurityAdvisoryExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+            if (string.IsNullOrEmpty(advisory.GhsaId))
+            {
+                throw new ArgumentException("Cannot create a permalink for security advisory without GHSA ID.", nameof(advisory.GhsaId));
+            }
+
+            // This is the permalink format used by GitHub for security advisory pages.
+            // Note that the GHSA ID part of the URL is case-sensitive, so we pass in the GHSA ID as-is.
+
+            // Todo: remove this hard-coded work-around when the "permalink" field becomes available on the GH API.
+            // See: https://developer.github.com/v4/object/securityadvisory/
```
ryuyu: Should we track this work with Github issue?

src/GitHubVulnerabilities2Db/Ingest/AdvisoryIngestor.cs
```diff
@@ -41,7 +41,7 @@ public async Task IngestAsync(IReadOnlyList<SecurityAdvisory> advisories)
             {
                 GitHubDatabaseKey = advisory.DatabaseId,
                 Severity = (PackageVulnerabilitySeverity)Enum.Parse(typeof(PackageVulnerabilitySeverity), advisory.Severity, ignoreCase: true),
-                ReferenceUrl = advisory.References.FirstOrDefault()?.Url
+                AdvisoryUrl = advisory.GetPermalink()
             };
```
joelverhagen: Why do we persist the whole URL? Isn't this a view concern? Seems more flexible in the future to just persist what we know (`GhsaId`) and allow the caller to choose what to do with it (i.e. call another API, build a URL, etc).

xavierdecoster: We could do both, if desired. Though the entity is meant to be GitHub-agnostic...

xavierdecoster: Note that both `GhsaId` and `Permalink` will be fields retrieved from the GH GraphQL API. I think we are only interested in the `AdvisoryUrl`, which is what we'll be surfacing in our own protocol and UI.

joelverhagen: Okay, I thought we already non-generic with `GitHubDatabaseKey`. Should the be "ExternalKey" instead?

xavierdecoster: Yeah... I'm kinda wondering about the usefulness of that `GitHubDatabaseKey`...
It's the one thing GH guaranteed that wouldn't change. But the GraphQL API is using the `GhsaId` to query for a specific advisory.

See also [this comment](https://github.com/NuGet/NuGetGallery/pull/7721/files/d2e356cbd0fd4b542055f5e6d1c3c6fb823326cb#diff-9e9a4b0f4c0edd86442f8c478e7d3390L28) added by Scott when he modeled the entity.

src/GitHubVulnerabilities2Db/GraphQL/SecurityAdvisoryExtensions.cs
xavierdecoster: Created tracking issue: https://github.com/nuget/engineering/issues/2890

src/GitHubVulnerabilities2Db/Ingest/AdvisoryIngestor.cs
xavierdecoster: I guess it would come down to having some `Provider` column (in this case `GitHub` would be the value) so we can identify the source of the data, plus a generic external key. Or alternatively have the approach that Scott described of using a key per provider (see his comment).

PR #7731, opened 2019-12-05
[Hotfix] Align NuGet.Jobs and ServerCommon dependencies to 2.60.0

joelverhagen:
There was a mismatch in the graph. Some dependencies used 2.58.0 ServerCommon, others used 2.59.0. This resolved to 2.59.0. A breaking change was introduced in 2.59.0. This caused a runtime issue in entry points that use NuGet.Jobs dependencies.

Essentially this change moved to the latest version of NuGet.Jobs and aligns with 2.60.0.

### DIFF
--- a/src/NuGet.Services.DatabaseMigration/NuGet.Services.DatabaseMigration.csproj
+++ b/src/NuGet.Services.DatabaseMigration/NuGet.Services.DatabaseMigration.csproj
@@ -77,10 +77,7 @@
       <Version>4.8.0</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Jobs.Common">
-      <Version>4.3.0-dev-3159955</Version>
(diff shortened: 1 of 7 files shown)


src/NuGet.Services.DatabaseMigration/NuGet.Services.DatabaseMigration.csproj
```diff
@@ -77,10 +77,7 @@
       <Version>4.8.0</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Jobs.Common">
-      <Version>4.3.0-dev-3159955</Version>
-    </PackageReference>
-    <PackageReference Include="NuGet.Services.Configuration">
```
joelverhagen: This dependency is pulled in by NuGet.Jobs.Common.

PR #7745, opened 2019-12-06
No longer depend on Obsolete AppInsights APIs

xavierdecoster:
This PR updates the Gallery (and the jobs within the solution) to use the new Application Insights initialization logic introduced in ServerCommon 2.60.0 and NuGet.Jobs (current dev branch).

Goal is to move off obsolete AppInsights APIs, such as `TelemetryConfiguration.Active`.

In doing so, some legacy code and static classes required refactoring. Work in progress, but opening the PR already to collect feedback early on.

Also needs to consume a to-be-released update of `NuGet.Services.Logging` ([PR pending](https://github.com/NuGet/ServerCommon/pull/324))

### DIFF
--- a/src/AccountDeleter/AccountDeleter.csproj
+++ b/src/AccountDeleter/AccountDeleter.csproj
@@ -107,7 +107,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Validation.Common.Job">
-      <Version>4.3.0-dev-3159955</Version>
+      <Version>4.3.0-dev-3309631</Version>
     </PackageReference>
   </ItemGroup>
   <ItemGroup>
--- a/src/AccountDeleter/Job.cs
+++ b/src/AccountDeleter/Job.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.ComponentModel.Design;
 using System.IO;
 using System.Linq;
-using System.ServiceModel.Configuration;
 using Autofac;
 using Autofac.Core;
 using Microsoft.Extensions.Configuration;
@@ -101,14 +100,14 @@ namespace NuGetGallery.AccountDeleter
                 services.AddTransient<IMessageServiceConfiguration, CoreMessageServiceConfiguration>();
             }
 
-
             services.AddScoped<IEmailBuilderFactory, EmailBuilderFactory>();
-            services.AddScoped<ITelemetryClient, TelemetryClientWrapper>();
+            services.AddScoped<ITelemetryClient, TelemetryClientWrapper>(
+                sp => TelemetryClientWrapper.UseTelemetryConfiguration(ApplicationInsightsConfiguration.TelemetryConfiguration));
 
-            ConfigureGalleryServices(services, configurationRoot);
+            ConfigureGalleryServices(services);
         }
 
-        protected void ConfigureGalleryServices(IServiceCollection services, IConfigurationRoot configurationRoot)
+        protected void ConfigureGalleryServices(IServiceCollection services)
         {
             if (IsDebugMode)
             {
@@ -127,7 +126,6 @@ namespace NuGetGallery.AccountDeleter
                 services.AddScoped<IDeleteAccountService, DeleteAccountService>();
 
                 services.AddScoped<IUserService, AccountDeleteUserService>();
-                services.AddScoped<ITelemetryClient>(sp => { return TelemetryClientWrapper.Instance; });
                 services.AddScoped<IDiagnosticsService, DiagnosticsService>();
 
                 services.AddScoped<IPackageService, PackageService>();
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -20,6 +20,7 @@ using Autofac;
 using Autofac.Core;
 using Autofac.Extensions.DependencyInjection;
 using Elmah;
+using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Http;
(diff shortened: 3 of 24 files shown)


src/AccountDeleter/Job.cs
```diff
@@ -127,7 +126,6 @@ protected void ConfigureGalleryServices(IServiceCollection services, IConfigurat
                 services.AddScoped<IDeleteAccountService, DeleteAccountService>();
 
                 services.AddScoped<IUserService, AccountDeleteUserService>();
-                services.AddScoped<ITelemetryClient>(sp => { return TelemetryClientWrapper.Instance; });
```
xavierdecoster: This was already configured [here](https://github.com/NuGet/NuGetGallery/pull/7745/files?utf8=%E2%9C%93&diff=unified&w=1#diff-a55e62854ff973dd124a5bf076c676edR104).

src/AccountDeleter/AccountDeleter.csproj
```diff
@@ -107,7 +107,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Validation.Common.Job">
-      <Version>4.3.0-dev-3159955</Version>
+      <Version>4.3.0-dev-3297551</Version>
```
joelverhagen: Ah! I need to merge master back into dev.

src/NuGetGallery/App_Start/OwinStartup.cs
```diff
@@ -77,63 +77,65 @@ public static void Configuration(IAppBuilder app)
 
             // Setup telemetry
             var instrumentationKey = config.Current.AppInsightsInstrumentationKey;
-            if (!string.IsNullOrEmpty(instrumentationKey))
+            var heartbeatIntervalSeconds = config.Current.AppInsightsHeartbeatIntervalSeconds;
+
+            ApplicationInsightsConfiguration applicationInsightsConfiguration;
+
+            if (heartbeatIntervalSeconds > 0)
```
skofman1: I see that ApplicationInsightsConfiguration is initialized twice. Here and in DefaulyDependenciesModule. Why is this needed?

src/NuGetGallery/OData/NuGetODataController.cs
```diff
@@ -27,8 +28,12 @@ public abstract class NuGetODataController
         {
             _configurationService = configurationService ?? throw new ArgumentNullException(nameof(configurationService));
             _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+
+            ODataQueryVerifier = ODataQueryVerifier.Build(telemetryService);
```
skofman1: not your change.. please pull latest dev

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -93,30 +93,18 @@ protected override void Load(ContainerBuilder builder)
...
-                }
-                else
-                {
-                    // Configure instrumentation key,
-                    // and register NuGet.Services.Logging.TelemetryContextInitializer.
-                    ApplicationInsights.Initialize(instrumentationKey);
-                }
-            }
+            // Register the ILoggerFactory and configure AppInsights.
+            var applicationInsightsConfiguration = ConfigureApplicationInsights(configuration);
```
skofman1: >applicationInsightsConfiguration [](start = 16, length = 32)

Does this override the configuration we have in a applicationinsights.config? One important thing we configured is not to sample: [https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling](https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling)

Please verify that we don't get sampling back with this change. I think looking at the runtime config with debugger should be sufficient.

src/NuGetGallery/OData/NuGetODataController.cs
xavierdecoster: This is my change? 😕 

`ODataQueryVerifier` used to be a static class. Not anymore. It needs a `TelemetryService` using the configured `TelemetryConfiguration`.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
xavierdecoster: The first thing that `ApplicationInsights.Initialize` does, is reading the `applicationinsights.config` file.

I explicitly asked about this before making the change in ServerCommon.
See the code comment [here](https://github.com/NuGet/ServerCommon/pull/321/files#diff-c0eac733d736223afbbf9061c7200bb1R59) and my question on AI issue tracker [here](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1298#issuecomment-555505295).

I will verify sampling at runtime to be sure.

src/NuGetGallery/App_Start/OwinStartup.cs
xavierdecoster: I was wondering the same thing, but didn't get to digging into it more until now.

The `OwinStartup` class loads the `DefaultDependenciesModule` through a call to `app.UseAutofacInjection` [here](https://github.com/NuGet/NuGetGallery/blob/542a95582a7e306ca38fbc2e8449207e0cbb01be/src/NuGetGallery/App_Start/OwinStartup.cs#L54), which in turn calls `builder.RegisterAssemblyModules(currentAssembly);` [here](https://github.com/NuGet/NuGetGallery/blob/542a95582a7e306ca38fbc2e8449207e0cbb01be/src/NuGetGallery/App_Start/AutofacConfig.cs#L37).

My *thinking* is that we should consolidate everything within `DefaultDependenciesModule` and remove additional telemetry configuration from `OwinStartup`.

src/NuGetGallery/OData/NuGetODataController.cs
skofman1: oops! Looked similar to a different change 😿

Issue #7747, opened 2019-12-06
Application Insights is broken in NuGetGallery


joelverhagen:
This is as of ServerCommon 2.60.0. It's possible 2.59.0 is also affected. 2.58.0 is last known good.

Checked the following tables: 

- traces
- requests
- dependencies
- customMetrics

Checked the following apps:

- NuGetGallery
- DatabaseMigration.Gallery (DatabaseMigrationsTools)

PR #7750, opened 2019-12-09
[Hotfix] Revert to ServerCommon 2.58.0

joelverhagen:
Reverting to 2.58.0 of ServerCommon libraries until NuGet.Jobs has pulled in the latest version (2.60.0 or later).

Address https://github.com/NuGet/NuGetGallery/issues/7747.


src/NuGetGallery/App_Start/OwinStartup.cs
```diff
@@ -96,10 +96,6 @@ public static void Configuration(IAppBuilder app)
 
                 // Add enrichers
                 TelemetryConfiguration.Active.TelemetryInitializers.Add(new DeploymentIdTelemetryEnricher());
-                if (config.Current.DeploymentLabel != null)
```
joelverhagen: This API was added in 2.59.0 along with the breaking changes incompatible with 2.58.0.

Issue #7385, opened 2019-07-29
Add NPS survey to NuGet.org


karann-msft:
# Resources

* [Usabilla homepage](https://usabilla.com/)
* Examples: [Visual Studio website](https://visualstudio.microsoft.com/#) and [.NET Blog](https://devblogs.microsoft.com/dotnet/). Text reads `What do you think of this page?` Selecting a value (1-5) opens a text area that reads `How was your experience?`. This is placed on all pages.
* [What is Net Promoter Score (NPS)?](https://usabilla.com/net-promoter-score-nps/)
* [JavaScript API](https://support.usabilla.com/hc/en-us/articles/208019369-How-to-use-our-JavaScript-API-Command-List-), can be used to export data

# Steps

- [x] Karan+Joel: Make sure we're good from a GDPR standpoint on Usabilla
- [x] Joel: How can we/must we obfuscate URLs sent in feedback?
- [x] Karan+Joel: Add an SVG version of the feedback button
- [x] Joel: Deploy to DEV with a test button ID so we can play
- [x] Karan: submit customization of the form so we have NuGet logo or whatever
  - [x] Submitted
  - [x] Verified
- [x] Karan: provide Joel with 3 button IDs for DEV, INT, and PROD.
- [x] Joel: Check in button IDs to config
- [x] Joel: Merge change to dev branch
- [x] Joel: Deploy all the way to PROD
- [x] Joel: Turn feature flag on

PR #7768, opened 2019-12-13
Add support for a Usabilla feedback button on every page

joelverhagen:
This is behind feature flag `NuGetGallery.UsabillaEveryPage`.

Progress on https://github.com/NuGet/NuGetGallery/issues/7385.

image

The survey questions can be changed from the Usabilla portal.

image

Note that when the modal opens and the page content does not fill the viewport (i.e. there is no vertical scroll bar, the footer jumps up). I have not found a workaround and there is similar behavior on other websites:

image


src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -351,6 +351,58 @@
...
+{
+    var config = DependencyResolver.Current.GetService<IGalleryConfigurationService>();
+    if (config != null)
+    {
+        var buttonId = config.Current.UsabillaFeedbackButtonId;
+        if (!string.IsNullOrWhiteSpace(buttonId))
+        {
+            var featureFlagService = DependencyResolver.Current.GetService<IFeatureFlagService>();
+            if (featureFlagService.IsUsabillaButtonEnabledOnEveryPage())
+            {
```
zhhyu: Maybe a shared method to do the check?

src/NuGetGallery.Services/Configuration/FeatureFlagService.cs
```diff
@@ -33,6 +33,7 @@ public class FeatureFlagService : IFeatureFlagService
         private const string ODataCacheDurationsFeatureName = GalleryPrefix + "ODataCacheDurations";
         private const string ShowEnable2FADialog = GalleryPrefix + "ShowEnable2FADialog";
         private const string Get2FADismissFeedback = GalleryPrefix + "Get2FADismissFeedback";
+        private const string UsabillaEveryPage = GalleryPrefix + "UsabillaEveryPage";
```
zhhyu: UsabillaOnEveryPageFeatureName?

src/NuGetGallery/App_Code/ViewHelpers.cshtml
joelverhagen: Good point! I'll do that.

src/NuGetGallery.Services/Configuration/FeatureFlagService.cs
joelverhagen: Sure, sounds fine.

src/NuGetGallery/App_Data/Files/Content/flags.json
```diff
@@ -7,7 +7,8 @@
     "NuGetGallery.GravatarProxyEnSubdomain": "Enabled",
     "NuGetGallery.ODataCacheDurations": "Enabled",
     "NuGetGallery.ShowEnable2FADialog": "Enabled",
-    "NuGetGallery.Get2FADismissFeedback": "Disabled"
+    "NuGetGallery.Get2FADismissFeedback": "Disabled",
+    "NuGetGallery.UsabillaEveryPage": "Enabled"
```
zhhyu: Should this match the new name?

joelverhagen: The name is the same. I just changed the constant name.

PR #7770, opened 2019-12-14
Build extensions targets file for build pipelines

agr:
Progress on https://github.com/NuGet/Engineering/issues/2883

This adds an optional point to insert targets file during the build.


zhhyu:
What is the purpose of this change?

agr:
@zhhyu, it is needed to provide a point of injection of MSBuild targets so we could customize the list of item that get packaged when building an app service package.

PR #7788, opened 2019-12-23
Upgrade telemetry

xavierdecoster:
Consume NuGet.Jobs and ServerCommon updates for telemetry and heartbeats.


src/AccountDeleter/AccountDeleter.csproj
```diff
@@ -107,7 +107,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Validation.Common.Job">
-      <Version>4.3.0-dev-3159955</Version>
+      <Version>4.3.0-dev-aitraces-3341370</Version>
```
xavierdecoster: Will upgrade as soon as we have a stable release of NuGet.Jobs consuming the required changes from https://github.com/NuGet/NuGet.Jobs/pull/845

xavierdecoster: Done in https://github.com/NuGet/NuGetGallery/pull/7788/commits/ce3a3b047b818a8af2dcd1f67b45948c11bf5a45

xavierdecoster:
Verified `GitHubVulnerabilities2Db` job on DEV: traces are now traces again instead of customEvents.

xavierdecoster:
Been verifying this for the gallery: Request telemetry is being tracked, but for some reason dependencies are not. Investigating...

xavierdecoster:
The last commit https://github.com/NuGet/NuGetGallery/pull/7788/commits/4b3a30d3777b12328b44232ab15ebe410e2847a3 seems to fix dependency telemetry... 💥 

Verified on DEV for USE2 deployment (artifact `3347045`, release `NuGet.Server - NuGetGallery Release 0794-3347045-dev-serilog`)

AI query (same for `requests` and other telemetry types):

```
dependencies
| extend cloudDeploymentId = tostring(customDimensions.CloudDeploymentId)
| where isnotempty(cloudDeploymentId)
| where timestamp > ago(6h)
| order by timestamp desc 
| summarize count() by bin(timestamp, 10m), cloudDeploymentId
| render columnchart 
```

loic-sharma:
@xavierdecoster I would also check performance counters.

Also, considering how painful it's been to update App Insights, do you think it'd be worth while to have a OneNote page on "Things to Verify If You Update Application Insights"?

xavierdecoster:
> Just curious, what is this `Remove` for? I'm a little concerned that we may accidentally lose some configuration by registering the same module type twice.

The whole point is to work around this boolean here: https://github.com/microsoft/ApplicationInsights-dotnet/blob/d94ee824280afdd96dd70bc5fcc1a0c1b1b5b3d6/WEB/Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs#L116

Initializing modules twice doesn't work (for some, including `DependencyTrackingTelemetryModule`). That's why our own config wasn't picked up.

I'm calling Remove to remove any instance that may have been auto-added with the wrong config (assuming the `TelemetryConfiguration.Active` static one, which we don't want to use). Then I new up a new instance of the module, register our own config, and add it to the modules collection. This fixed it.

xavierdecoster:
> @xavierdecoster I would also check performance counters.
> 
> Also, considering how painful it's been to update App Insights, do you think it'd be worth while to have a OneNote page on "Things to Verify If You Update Application Insights"?

It surely could. But I'll add unit tests asserting that the required modules have been added and configured properly ;)
Doesn't guarantee runtime behavior, still, but at least we assert we did what we could 😄 
Btw, the code reflects the modules as they were configured previously using `applicationinsights.config`.

PR #7791, opened 2019-12-27
Remove an unnecessary SQL query

loic-sharma:
The display package page gets data through lazy loading. One such lazily loaded queries was a repeat:

1. Eagerly load all versions of a package: https://github.com/NuGet/NuGetGallery/blob/90f8fbea00217db2e5d918e140621c0aea0ca12d/src/NuGetGallery/Controllers/PackagesController.cs#L800
1. Lazily load all versions of a package: https://github.com/NuGet/NuGetGallery/blob/90f8fbea00217db2e5d918e140621c0aea0ca12d/src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs#L63

These queries are particularly expensive when a package many versions. Given that the second query is redundant, I piped the data from the first query down to the view model factory. 

Test build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3350351
Test release: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=541799

Part of https://github.com/NuGet/Engineering/issues/2910

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -827,8 +827,9 @@ namespace NuGetGallery
                 return HttpNotFound();
             }
 
+            var readme = await _readMeService.GetReadMeHtmlAsync(package);
             var deprecation = _deprecationService.GetDeprecationByPackage(package);
-            var model = _displayPackageViewModelFactory.Create(package, currentUser, deprecation, await _readMeService.GetReadMeHtmlAsync(package));
+            var model = _displayPackageViewModelFactory.Create(package, packages, currentUser, deprecation, readme);
 
             model.ValidatingTooLong = _validationService.IsValidatingTooLong(package);
             model.PackageValidationIssues = _validationService.GetLatestPackageValidationIssues(package);
@@ -1637,7 +1638,7 @@ namespace NuGetGallery
                 return HttpForbidden();
             }
 
-            var model = _deletePackageViewModelFactory.Create(package, currentUser, DeleteReasons);
+            var model = _deletePackageViewModelFactory.Create(package, packages, currentUser, DeleteReasons);
 
             // Fetch all versions of the package with symbols.
             var versionsWithSymbols = packages
--- a/src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
+++ b/src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
@@ -20,7 +20,7 @@ namespace NuGetGallery
 
         public DisplayPackageViewModel Create(
             Package package,
-            IReadOnlyCollection<Package> packageRegistration,
+            IReadOnlyCollection<Package> allVersions,
             User currentUser,
             PackageDeprecation deprecation,
             RenderedReadMeResult readmeResult)
@@ -29,7 +29,7 @@ namespace NuGetGallery
             return Setup(
                 viewModel,
                 package,
-                packageRegistration,
+                allVersions,
                 currentUser,
                 deprecation,
                 readmeResult);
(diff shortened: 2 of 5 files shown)


joelverhagen:
Can we verify this behavior via the `dependencies` AI table?

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
```diff
@@ -20,6 +20,7 @@ public DisplayPackageViewModelFactory(IIconUrlProvider iconUrlProvider)
 
         public DisplayPackageViewModel Create(
             Package package,
+            IReadOnlyCollection<Package> packageRegistration,
```
joelverhagen: IMO this should be `allVersions` or something like that. `packageRegistration` parameter name implies a `PackageRegistration` type.

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
```diff
@@ -60,9 +64,7 @@ public DisplayPackageViewModelFactory(IIconUrlProvider iconUrlProvider)
 
             viewModel.Dependencies = new DependencySetsViewModel(package.Dependencies);
 
-            var packageHistory = package
-                .PackageRegistration
-                .Packages
+            var packageHistory = packageRegistration
                 .OrderByDescending(p => new NuGetVersion(p.Version))
```
joelverhagen: This `OrderByDescending` must have already been evaluated in the app code right? Interesting that this was just happening silently since `PackageRegistration.Packages` is not a materialized sequence.

loic-sharma: Yes, I can't imagine that was translated to SQL. Entity Framework magic can be scary 😥

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
```diff
@@ -20,6 +20,7 @@ public DisplayPackageViewModelFactory(IIconUrlProvider iconUrlProvider)
 
         public DisplayPackageViewModel Create(
             Package package,
+            IReadOnlyCollection<Package> allVersions,
```
zhhyu: Maybe "packageWithAllVersions" is clearer?  "allVersions" looks general.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -827,8 +827,9 @@ public virtual async Task<ActionResult> DisplayPackage(string id, string version
                 return HttpNotFound();
             }
 
+            var readme = await _readMeService.GetReadMeHtmlAsync(package);
             var deprecation = _deprecationService.GetDeprecationByPackage(package);
-            var model = _displayPackageViewModelFactory.Create(package, currentUser, deprecation, await _readMeService.GetReadMeHtmlAsync(package));
+            var model = _displayPackageViewModelFactory.Create(package, packages, currentUser, deprecation, readme);
```
zhhyu: >packages [](start = 72, length = 8)

Maybe use the same name convention as the called method.

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
loic-sharma: Fixed

loic-sharma:
> Can we verify this behavior via the dependencies AI table?

Yup, I did this on my dev machine and made sure I had one less query.

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
loic-sharma: Personally I prefer `allVersions` because it's shorter. Let me know if you think a comment is needed on this method to clarify

zhhyu: It's fine.

src/NuGetGallery/Controllers/PackagesController.cs
loic-sharma: I feel a little torn on this. I like:

```csharp
 var packages = _packageService.FindPackagesById(...);
```

This doesn't seem as nice to me:

```csharp
 var allVersions = _packageService.FindPackagesById(...);
```

That being said, I agree it'd be nice to be consistent with the method's parameter name. Let me know if you feel strongly one over the other 😊

zhhyu: It's fine.

loic-sharma:
Replaced by https://github.com/NuGet/NuGetGallery/pull/7795

PR #7798, opened 2020-01-07
HTTP to HTTPS redirect differences between App and Cloud services

agr:
Progress on https://github.com/NuGet/Engineering/issues/2863

Cloud and app services handle the HTTP to HTTPS redirect differently (we can enable "Always HTTPS" for an app service, while we can't do that for the cloud service), so the test update is required to accommodate for that difference.
This change also requires the test configuration to be updated.

### DIFF
--- a/tests/NuGetGallery.FunctionalTests/Security/HttpToHttpsRedirectTests.cs
+++ b/tests/NuGetGallery.FunctionalTests/Security/HttpToHttpsRedirectTests.cs
@@ -43,7 +43,7 @@ namespace NuGetGallery.FunctionalTests.Security
         [Theory]
         [MemberData(nameof(GetForAllUrls))]
         [Priority(0)]
-        [Category("P0Tests")]
+        [Category("CloudServiceTests")]
         public async Task HttpToHttpsRedirectHappensForSupportedMethods(HttpMethod method, string url)
         {
             // Ideally, we should test both GET and HEAD methods for all the URLs, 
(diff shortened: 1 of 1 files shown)


tests/NuGetGallery.FunctionalTests/Security/HttpToHttpsRedirectTests.cs
```diff
@@ -91,6 +91,30 @@ public async Task ExcludedUrlsDontRedirect(string url)
                 HttpMethod.Get, uri);
         }
 
+        public static IEnumerable<object[]> RemainingUrlsAndMethodsForAppService =>
+                from url in UrlsToTest.Concat(UrlsExcludedFromRedirect).SelectMany(x => x)
```
loic-sharma: `UrlsExcludedFromRedirect` is only the proper name for cloud services right? It seems that now these URLs do redirect for App services? If so, maybe add a comment or rename this field to clarify

zhhyu: >UrlsExcludedFromRedirect [](start = 46, length = 24)

I wonder we can do a little refactoring here to extend the "HttpMethodsAndUrlsToTest" at the bottom in order to accept the testing Urls argument so we can easily form and feed the Urls and Methods to the test cases here.

agr: We are not going to be in the state where we have to test both cloud and app services for long. Instead of building a framework around the need to test both for some short time, we'll drop the cloud service specific tests in the near future.

PR #7799, opened 2020-01-07
Functional test script update to wait for the service to start up after the deployment

agr:
Progress on https://github.com/NuGet/Engineering/issues/2863

App service deployment reports success way before the service initialization completes successfully and is able to serve requests. This breaks our tests that encounter 502 responses and fail.
This change adds a delay and a response test to the test run script to make sure we run tests against the working service.

### DIFF
--- a/tests/Scripts/RunTestScripts.ps1
+++ b/tests/Scripts/RunTestScripts.ps1
@@ -13,6 +13,56 @@ Write-Host $fullDivider
 
 $failedTests = New-Object System.Collections.ArrayList
 
+Function Wait-ForServiceStart($MaxWaitSeconds) {
+    $configurationFile = $env:ConfigurationFilePath
+    if ($null -eq $configurationFile) {
+        Write-Error "Configuration file path environment variable is not specified"
+        return $false;
+    }
+    if (-not (Test-Path $configurationFile)) {
+        Write-Error "Missing configuration file: $configurationFile"
+        return $false
+    }
+    $configuration = Get-Content $configurationFile | ConvertFrom-Json;
+    if ($null -eq $configuration.Slot) {
+        Write-Error "`"Slot`" property was not found in the test configuration object: $configurationFile"
+        return $false
+    }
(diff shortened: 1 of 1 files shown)


tests/Scripts/RunTestScripts.ps1
```diff
@@ -13,6 +13,56 @@ Write-Host $fullDivider
...
+    $baseUrlPropertyName = if ($configuration.Slot -eq "staging") { "StagingBaseUrl" } else { "ProductionBaseUrl" }
+    if ($null -eq $configuration.$baseUrlPropertyName) {
+        Write-Error "`"$($baseUrlPropertyName)`" property was not found in the test configuration object: $configurationFile"
+        return $false
+    }
+
+    $galleryUrl = $configuration.$baseUrlPropertyName
+    $response = $null
+    Write-Host "$(Get-Date -Format s) Sleeping before querying the service";
+    Start-Sleep -Seconds 120
```
shishirx34: why wait for 2 mins if you are going to wait until non-502 below. Can't we simply wait on non-502? i.e. if the service wakes up before 2 mins we are good, likely won't, but future improvements might(?).

agr: I've seen situations when by the time script runs old version is still being served for the URLs being tested. Azure support claims it shouldn't happen, but I've definitely run into the check passing, followed by tests getting 502's. Delay is an attempt to mitigate that.

joelverhagen: Can we hit an endpoint and poll until we see a release label (or some other stamp indicating the new deployment is live)?

zhhyu: Maybe it's better that we also pass the configuration to set up this parameter? I wonder this time may not be a stable one, and we may need to tune this parameter later, or use other better approaches.

agr: It would need a mechanism to pass the deployment label to Cloud services, which we don't currently have. And inventing one sounds unreasonable given that we move off them. We can optimize that later, when we don't have to support cloud service.

PR #7803, opened 2020-01-09
Minor app service related changes

agr:
Renamed a variable and a method to emphasize their relation to cloud service infrastructure (not the cloud hosting in general).
Removed duplicate code that figured out machine name.
Added `<applicationInitialization>` section to the web.config.


src/NuGetGallery.Services/Configuration/ConfigurationService.cs
```diff
@@ -23,7 +24,7 @@ public class ConfigurationService : IGalleryConfigurationService
         protected const string ServiceBusPrefix = "AzureServiceBus.";
         protected const string PackageDeletePrefix = "PackageDelete.";
 
-        private bool _notInCloud;
+        private bool _notInCloudService;
```
joelverhagen: Hahahaha, this is an awesome delta.

agr: Well, it's the actual meaning of the variable: if cloud service API is available. We would be able to drop it completely once we are off cloud services.

src/NuGetGallery/Web.config
```diff
@@ -505,6 +505,10 @@
         </rule>
       </rules>
     </rewrite>
+    <applicationInitialization>
+      <add initializationPage="/"/>
+      <add initializationPage="/api/health-probe"/>
+    </applicationInitialization>
```
shishirx34: who uses these?

agr: First one is added mostly to align with the search service, and the second one is watched at least by the internal load balancer, yes.

Issue #7714, opened 2019-11-26
Gallery should reject packages with duplicate dependencies


joelverhagen:
These packages can't be opened by some tools (NuGet Package Explorer) and the semantics of a duplicate dependency are undefined.

For example:
https://www.nuget.org/packages/Paket.Core/5.237.1

![image](https://user-images.githubusercontent.com/94054/69655106-bf289480-102a-11ea-9368-ee717e540b9b.png)

These duplicate dependencies became an array with two items for the dependency range, which is invalid and caused Catalog2Lucene to get blocked.

The following query shows cases of this. There are very few.

```sql
SELECT pr.Id, p.NormalizedVersion, p.Created, p.PackageStatusKey, pd.Id AS DependencyId, pd.TargetFramework, COUNT(*) AS [Count]
FROM PackageRegistrations pr 
INNER JOIN Packages p ON pr.[Key] = p.PackageRegistrationKey
INNER JOIN PackageDependencies pd ON p.[Key] = pd.PackageKey
GROUP BY pr.Id, p.NormalizedVersion, p.Created, p.PackageStatusKey, pd.Id, pd.TargetFramework
HAVING COUNT(DISTINCT pd.VersionSpec) > 1
ORDER BY p.Created
```

Id | NormalizedVersion | Created | PackageStatusKey | DependencyId | TargetFramework | Count
-- | -- | -- | -- | -- | -- | --
Paket.Core | 5.237.1 | 11/26/2019 | 0 | Newtonsoft.Json | netstandard2.0 | 2
Paket.Core | 5.237.1 | 11/26/2019 | 0 | FSharp.Core | netstandard2.0 | 2
Paket.Core | 5.237.1 | 11/26/2019 | 0 | System.Net.Http.WinHttpHandler | netstandard2.0 | 2
Paket.Core | 5.237.1 | 11/26/2019 | 0 | System.Security.Cryptography.ProtectedData | netstandard2.0 | 2
Paket.Core | 5.237.2 | 11/26/2019 | 0 | FSharp.Core | netstandard2.0 | 2
Paket.Core | 5.237.2 | 11/26/2019 | 0 | Newtonsoft.Json | netstandard2.0 | 2
Paket.Core | 5.237.2 | 11/26/2019 | 0 | System.Security.Cryptography.ProtectedData | netstandard2.0 | 2
[...]

PR #7814, opened 2020-01-18
Reject packages with duplicate dependency groups

skofman1:
Addresses https://github.com/NuGet/NuGetGallery/issues/7714

### DIFF
--- a/src/NuGetGallery.Services/ServicesStrings.resx
+++ b/src/NuGetGallery.Services/ServicesStrings.resx
@@ -1131,6 +1131,6 @@ The {1} Team</value>
     <value>The API key credential with Key '{0}' is not revocable.</value>
   </data>
   <data name="NuGetPackageDuplicateDependencyGroup" xml:space="preserve">
-    <value>A nuget package may not contain multiple dependency groups for the same target framework.</value>
+    <value>A nuget package may not contain multiple dependency groups with the same target framework.</value>
   </data>
 </root>
--- a/src/NuGetGallery.Services/ServicesStrings.resx
+++ b/src/NuGetGallery.Services/ServicesStrings.resx
@@ -1130,4 +1130,7 @@ The {1} Team</value>
   <data name="RevokeCredential_UnrevocableApiKeyCredential" xml:space="preserve">
     <value>The API key credential with Key '{0}' is not revocable.</value>
   </data>
+  <data name="NuGetPackageDuplicateDependencyGroup" xml:space="preserve">
+    <value>A nuget package may not contain multiple dependency groups for the same target framework.</value>
+  </data>
 </root>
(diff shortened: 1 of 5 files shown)


src/NuGetGallery.Services/ServicesStrings.resx
```diff
@@ -1130,4 +1130,7 @@ The {1} Team</value>
   <data name="RevokeCredential_UnrevocableApiKeyCredential" xml:space="preserve">
     <value>The API key credential with Key '{0}' is not revocable.</value>
   </data>
+  <data name="NuGetPackageDuplicateDependencyGroup" xml:space="preserve">
+    <value>A nuget package may not contain multiple dependency groups for the same target framework.</value>
```
loic-sharma: Suggestion:

```suggestion
    <value>A NuGet package may not contain multiple dependency groups with the same target framework.</value>
```

We could also say "The package may not..."

skofman1: I went for similar language as the other error message in that area. They all had 'nuget' and not 'NuGet'. Will use the rest of the suggestion :)

loic-sharma: That seems inconsistent with our branding. Would it be difficult to fix up all the other error messages?

skofman1: I'm sure there's plenty of lowercase nuget in the strings. However, it's out of scope of this PR..

joelverhagen:
Have you checked if a duplicate dependency with a different version but in the same group has the same problem. I think it will and should be blocked similarly.

tests/NuGetGallery.Facts/Helpers/PackageHelperTests.cs
```diff
@@ -167,6 +169,49 @@ public void ChecksIdVersionCombinedLength()
                 var ex = Assert.Throws<EntityException>(() => PackageHelper.ValidateNuGetPackageMetadata(metadata));
                 Assert.Contains("ID and version", ex.Message);
             }
+
+            [Fact]
+            public void ChecksDupliacteDependencyGroups()
+            {
+                var duplicateDependencyGroup = new PackageDependencyGroup(
```
joelverhagen: Nit I would test this with a real nuspec/nupkg made in memory. We have some other tests that do this. Your test depends on groups of the same TFM not being clubbed together by the caller.

joelverhagen: If we have a test that verified this expectation I mentioned then this comment is less important. But afaik there's not a clear contract in this regard.

skofman1:
> Have you checked if a duplicate dependency with a different version but in the same group has the same problem. I think it will and should be blocked similarly.

Tested the scenario 'a duplicate dependency with a different version but in the same group'. We get an error message:
image

If the dependency has the same version, they seem to be merged.

PR #7826, opened 2020-01-29
[ReleasePrep] RI dev to master

agr:


src/NuGetGallery/Web.config
```diff
@@ -556,6 +556,10 @@
   </system.diagnostics>
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+			<dependentAssembly>
```
xavierdecoster: nit: formatting seems off

PR #7831, opened 2020-01-31
Format package URLs in descriptions

pakrym:
Just a suggestion. 

Currently package links in descriptions look like this:

image

What if we did something similar to what github does with issue links and formatted them with a shorter text:

image

Current regex is relatively strict to avoid any other links masquerading as nuget ones.

### DIFF
--- a/src/NuGetGallery/Helpers/HtmlExtensions.cs
+++ b/src/NuGetGallery/Helpers/HtmlExtensions.cs
@@ -82,7 +82,16 @@ namespace NuGetGallery.Helpers
 
                 if (PackageHelper.TryPrepareUrlForRendering(trimmedAnchorValue, out string formattedUri))
                 {
-                    return $"<a href=\"{formattedUri}\" rel=\"nofollow\">{formattedUri}</a>" + trimmedEntityValue;
+                    string anchorText = formattedUri;
+
+                    // Format links to NuGet packages
+                    Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, @"((http|https):\/\/www.nuget.org\/packages\/(?<name>[\w.-]+)\/?$)", RegexOptions.IgnoreCase);
+                    if (packageMatch != null && packageMatch.Groups["name"].Success)
+                    {
+                        anchorText = packageMatch.Groups["name"].Value;
+                    }
+
+                    return $"<a href=\"{formattedUri}\" rel=\"nofollow\">{anchorText}</a>" + trimmedEntityValue;
                 }
 
                 return match.Value;
--- a/src/NuGetGallery/Helpers/HtmlExtensions.cs
+++ b/src/NuGetGallery/Helpers/HtmlExtensions.cs
@@ -86,7 +86,7 @@ namespace NuGetGallery.Helpers
                     string anchorText = formattedUri;
                     string siteRoot = configurationService.GetSiteRoot(useHttps: true);
                     // Format links to NuGet packages
-                    Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, $@"({Regex.Escape(siteRoot)}\/packages\/(?<name>[^\/]+(\/[0-9a-zA-Z-.]+)?)\/?$)", RegexOptions.IgnoreCase);
+                    Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, $@"({Regex.Escape(siteRoot)}\/packages\/(?<name>\w+([_.-]\w+)*(\/[0-9a-zA-Z-.]+)?)\/?$)", RegexOptions.IgnoreCase);
                     if (packageMatch != null && packageMatch.Groups["name"].Success)
                     {
                         anchorText = packageMatch.Groups["name"].Value;
--- a/src/NuGetGallery/Helpers/HtmlExtensions.cs
+++ b/src/NuGetGallery/Helpers/HtmlExtensions.cs
@@ -85,7 +85,7 @@ namespace NuGetGallery.Helpers
                     string anchorText = formattedUri;
 
                     // Format links to NuGet packages
-                    Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, @"((http|https):\/\/www.nuget.org\/packages\/(?<name>[\w.-]+)\/?$)", RegexOptions.IgnoreCase);
+                    Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, @"((http|https):\/\/www.nuget.org\/packages\/(?<name>[^\/]+(\/[0-9a-zA-Z-.]+)?)\/?$)", RegexOptions.IgnoreCase);
                     if (packageMatch != null && packageMatch.Groups["name"].Success)
                     {
                         anchorText = packageMatch.Groups["name"].Value;
--- a/src/NuGetGallery/Helpers/HtmlExtensions.cs
+++ b/src/NuGetGallery/Helpers/HtmlExtensions.cs
@@ -58,7 +58,7 @@ namespace NuGetGallery.Helpers
                 .Replace(".", ".<wbr>"));
         }
 
-        public static IHtmlString PreFormattedText(this HtmlHelper self, string text)
+        public static IHtmlString PreFormattedText(this HtmlHelper self, string text, IGalleryConfigurationService configurationService)
         {
             // Encode HTML entities. Important! Security!
             var encodedText = HttpUtility.HtmlEncode(text);
@@ -84,17 +84,12 @@ namespace NuGetGallery.Helpers
                 if (PackageHelper.TryPrepareUrlForRendering(trimmedAnchorValue, out string formattedUri))
                 {
                     string anchorText = formattedUri;
-
-                    var configurationService = self.ViewContext.HttpContext.GetService<ConfigurationService>();
-                    if (configurationService != null)
+                    string siteRoot = configurationService.GetSiteRoot(useHttps: true);
+                    // Format links to NuGet packages
+                    Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, $@"({Regex.Escape(siteRoot)}packages\/(?<name>[^\/]+(\/[0-9a-zA-Z-.]+)?)\/?$)", RegexOptions.IgnoreCase);
+                    if (packageMatch != null && packageMatch.Groups["name"].Success)
                     {
-                        string siteRoot = configurationService.GetSiteRoot(useHttps: true);
-                        // Format links to NuGet packages
-                        Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, $@"({Regex.Escape(siteRoot)}\/packages\/(?<name>[^\/]+(\/[0-9a-zA-Z-.]+)?)\/?$)", RegexOptions.IgnoreCase);
-                        if (packageMatch != null && packageMatch.Groups["name"].Success)
-                        {
-                            anchorText = packageMatch.Groups["name"].Value;
-                        }
+                        anchorText = packageMatch.Groups["name"].Value;
                     }
 
                     return $"<a href=\"{formattedUri}\" rel=\"nofollow\">{anchorText}</a>" + trimmedEntityValue;
--- a/src/NuGetGallery/Helpers/HtmlExtensions.cs
+++ b/src/NuGetGallery/Helpers/HtmlExtensions.cs
@@ -86,7 +86,7 @@ namespace NuGetGallery.Helpers
                     string anchorText = formattedUri;
                     string siteRoot = configurationService.GetSiteRoot(useHttps: true);
                     // Format links to NuGet packages
-                    Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, $@"({Regex.Escape(siteRoot)}packages\/(?<name>[^\/]+(\/[0-9a-zA-Z-.]+)?)\/?$)", RegexOptions.IgnoreCase);
+                    Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, $@"({Regex.Escape(siteRoot)}\/packages\/(?<name>[^\/]+(\/[0-9a-zA-Z-.]+)?)\/?$)", RegexOptions.IgnoreCase);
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Helpers/HtmlExtensions.cs
```diff
@@ -82,7 +82,16 @@ string anchorEvaluator(Match match)
 
                 if (PackageHelper.TryPrepareUrlForRendering(trimmedAnchorValue, out string formattedUri))
                 {
-                    return $"<a href=\"{formattedUri}\" rel=\"nofollow\">{formattedUri}</a>" + trimmedEntityValue;
+                    string anchorText = formattedUri;
+
+                    // Format links to NuGet packages
+                    Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, @"((http|https):\/\/www.nuget.org\/packages\/(?<name>[\w.-]+)\/?$)", RegexOptions.IgnoreCase);
```
joelverhagen: This is hard-coded to nuget.org. There is a configuration value that has the site base URL which we should use instead.

Also, how do you think this work when the URL has the version in it, e.g. https://www.nuget.org/packages/Newtonsoft.Json/12.0.3 -- maybe Newtonsoft.Json 12.0.3?

pakrym: > This is hard-coded to nuget.org. There is a configuration value that has the site base URL which we should use instead.

Sounds good!.

> Also, how do you think this work when the URL has the version in it, e.g. https://www.nuget.org/packages/Newtonsoft.Json/12.0.3 -- maybe Newtonsoft.Json 12.0.3?

I specifically started with the simplest implementation but can add a version specific one.

joelverhagen:
I love the idea!

Let's pull in some PMs: @karann-msft, @anangaur - what do you think? This further diverges the behavior from Visual Studio but I think that's not a huge deal. In general we try to align the experience but this is a pretty improvement.

@pakrym, we should base this off of `dev` branch.

src/NuGetGallery/Helpers/HtmlExtensions.cs
joelverhagen: > I specifically started with the simplest implementation but can add a version specific one.

Sure, makes sense. It's probably worth looking at all of the package descriptions to see how many packages this will help, how many have the version component, how many don't have "www" etc. I can do this investigation and it will help us pick the most impactful regex. I'll put the data as a top level comment on this PR when I have it.

tests/NuGetGallery.Facts/Helpers/HtmlExtensionsFacts.cs
```diff
@@ -93,6 +93,34 @@ public void ConvertsUrlsToLinks(string input, string expected)
...
+            [InlineData("My package is https://www.nuget.org/packages/WindowsAzure.Storage/!", "My package is <a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/\" rel=\"nofollow\">WindowsAzure.Storage</a>!")]
+            [InlineData("My package is http://www.nuget.org/packages/WindowsAzure.Storage/", "My package is <a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/\" rel=\"nofollow\">WindowsAzure.Storage</a>")]
+            [InlineData("My package is http://www.nuget.org/packages/WindowsAzure.Storage/sub/path/", "My package is <a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/sub/path/\" rel=\"nofollow\">https://www.nuget.org/packages/WindowsAzure.Storage/sub/path/</a>")]
+            [InlineData("My package is https://www.nuget.org/packages/WindowsAzure.Storage/packages.", "My package is <a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/packages\" rel=\"nofollow\">https://www.nuget.org/packages/WindowsAzure.Storage/packages</a>.")]
```
joelverhagen: There should be tests explicitly for the version case since it's the most common sub-path.

Also, cases covering the case insensitivity option you passed.

pakrym: Done.

src/NuGetGallery/Helpers/HtmlExtensions.cs
pakrym: Added version support, format as `PackageName/Version`. I'm having trouble finding the base url. Other URL formatting code seems to hardcode nuget.org too:  https://github.com/pakrym/NuGetGallery/blob/b49bd7f09b4854999cbe34f7fa356b9393e5b8f4/src/NuGetGallery.Services/Extensions/UriExtensions.cs#L75

pakrym:
Rebased on `dev`

anangaur:
LGTM. I guess the need of the hour is Readme.md support (markdown) 😊

joelverhagen:
Note that `PreFormattedText` is called on the release notes and custom deprecation message as well as the description field. I think this is fine but it should be called out here.

src/NuGetGallery/Helpers/HtmlExtensions.cs
joelverhagen: https://github.com/NuGet/NuGetGallery/blob/08758f6582a1de098cfe725d5c0944f86252486b/src/NuGetGallery.Core/IAppConfiguration.cs#L211

joelverhagen:
Looks like `WindowsAzure.Storage` is the top package to be impacted. The 5 runners up are:

- Microsoft.Data.OData 5.8.4
- NLog 4.6.8
- NLog.Extensions.Logging 1.6.1
- iTextSharp 5.5.13.1
- Microsoft.Azure.ActiveDirectory.GraphClient 2.1.1

It's worth verifying that these cases look good too. @pakrym, could you small screenshots of the description sections for each of these? We are changing how the description is displayed potentially impacting some customer's expectation. It seems like a great change, but let's get eyeballs on how some of the top cases are looking after the change. Does this seem reasonable?

Regarding numbers (just listed packages, by distinct ID, exclude packages with null field):

Field | With package version URL | With package URL | Without package  URL
-- | -- | -- | --
Description | 13 | 348 | 186,590
ReleaseNotes | 41 | 161 | 64,551
Deprecation CustomMessage | 0 | 16 | 724

Not a high percentage but as shown above some very popular packages utilize this feature so I think it will improve a lot of page views.

tests/NuGetGallery.Facts/Helpers/HtmlExtensionsFacts.cs
```diff
@@ -93,6 +93,36 @@ public void ConvertsUrlsToLinks(string input, string expected)
...
+            [InlineData("My package is http://www.nuget.org/packages/WindowsAzure.Storage/?foo[]=a", "My package is <a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/?foo\" rel=\"nofollow\">https://www.nuget.org/packages/WindowsAzure.Storage/?foo</a>[]=a")]
+            [InlineData("http://www.nuget.org/packages/WindowsAzure.Storage is my package.", "<a href=\"https://www.nuget.org/packages/WindowsAzure.Storage\" rel=\"nofollow\">WindowsAzure.Storage</a> is my package.")]
+            [InlineData("\"http://www.nuget.org/packages/WindowsAzure.Storage/\" is my package.", "&quot;<a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/\" rel=\"nofollow\">WindowsAzure.Storage</a>&quot; is my package.")]
+            [InlineData("\'http://www.nuget.org/packages/WindowsAzure.Storage/\' is my package.", "&#39;<a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/\" rel=\"nofollow\">WindowsAzure.Storage</a>&#39; is my package.")]
+            [InlineData("http://www.nuget.org/packages/WindowsAzure.Storage/, is my package.", "<a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/\" rel=\"nofollow\">WindowsAzure.Storage</a>, is my package.")]
```
joelverhagen: What about missing `www`? (nuget.org redirects naked domain to `www`)

joelverhagen: (feel free to say "not worth it" from the implementation side, but this case should be asserted in UT depending on the selecting impl.)

src/NuGetGallery/Helpers/HtmlExtensions.cs
joelverhagen: If it's too painful to introduce the configuration, don't worry about it. Given the precedent you stated, we don't need to be "pure" here since we're not consistent already. So if it's easy use config, otherwise hard-code nuget.org and nugettest.org as we do elsewhere.

pakrym: I may have found a reasonable way to pull it through.

tests/NuGetGallery.Facts/Helpers/HtmlExtensionsFacts.cs
pakrym: Adding.

pakrym: I think it's either SiteRoot or supporting domains without www.

src/NuGetGallery/Helpers/HtmlExtensions.cs
pakrym: To double check what is SiteRoot  set to on the production site?

because description says `/// Gets the protocol-independent site root` but default value in `web.config` is `    <add key="Gallery.SiteRoot" value="http://localhost"/>`

joelverhagen: `SiteRoot` values:

- PROD: https://www.nuget.org
- INT: https://int.nugettest.org
- DEV: https://dev.nugettest.org

pakrym:
>could you small screenshots of the description sections for each of these?

# Microsoft.Data.OData 5.8.4
image

# NLog 4.6.8
image

# NLog.Extensions.Logging 1.6.1
image

# iTextSharp 5.5.13.1
image


# Microsoft.Azure.ActiveDirectory.GraphClient

image

src/NuGetGallery/Helpers/HtmlExtensions.cs
pakrym: Done

src/NuGetGallery/Helpers/HtmlExtensions.cs
```diff
@@ -82,7 +83,16 @@ string anchorEvaluator(Match match)
 
                 if (PackageHelper.TryPrepareUrlForRendering(trimmedAnchorValue, out string formattedUri))
                 {
-                    return $"<a href=\"{formattedUri}\" rel=\"nofollow\">{formattedUri}</a>" + trimmedEntityValue;
+                    string anchorText = formattedUri;
+                    string siteRoot = configurationService.GetSiteRoot(useHttps: true);
+                    // Format links to NuGet packages
+                    Match packageMatch = RegexEx.MatchWithTimeout(formattedUri, $@"({Regex.Escape(siteRoot)}\/packages\/(?<name>[^\/]+(\/[0-9a-zA-Z-.]+)?)\/?$)", RegexOptions.IgnoreCase);
```
joelverhagen: The regex for package IDs is well known here:
https://github.com/NuGet/NuGet.Client/blob/715edd1291b089d42cfef65566146f45a8cd6310/src/NuGet.Core/NuGet.Packaging/PackageCreation/Utility/PackageIdValidator.cs#L14

Any particular reason to you moved away from `\w` in https://github.com/NuGet/NuGetGallery/pull/7831/commits/d533d7b7e1dcb86e10693f137c46bd98eae9b51f?

heaths: Why escape each time, and not use compiled regex? Seems like a lot of work one very render that could probably be optimized.

tests/NuGetGallery.Facts/Helpers/HtmlExtensionsFacts.cs
```diff
@@ -84,10 +88,43 @@ public void EncodesHtml()
...
+            [InlineData("(http://www.nuget.org/packages/WindowsAzure.Storage/) is my package.", "(<a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/\" rel=\"nofollow\">WindowsAzure.Storage</a>) is my package.")]
+            [InlineData("http://www.nuget.org/packages/WindowsAzure.Storage/; is my package.", "<a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/\" rel=\"nofollow\">WindowsAzure.Storage</a>; is my package.")]
+            [InlineData("http://www.nuget.org/packages/WindowsAzure.Storage/- is my package.", "<a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/\" rel=\"nofollow\">WindowsAzure.Storage</a>- is my package.")]
+            [InlineData("My package is https://www.nuget.org/packages/WindowsAzure.Storage/9.3.1.", "My package is <a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/9.3.1\" rel=\"nofollow\">WindowsAzure.Storage/9.3.1</a>.")]
+            [InlineData("My package is https://www.nuget.org/packages/WindowsAzure.Storage/9.3.1/.", "My package is <a href=\"https://www.nuget.org/packages/WindowsAzure.Storage/9.3.1/\" rel=\"nofollow\">WindowsAzure.Storage/9.3.1</a>.")]
```
heaths: Probably should add a preview version test case as well.

src/NuGetGallery/Helpers/HtmlExtensions.cs
pakrym: I don't see a clear way to cache this regex considering how the application is factored.

pakrym: Nope, changed to using the regex you've linked to.

pakrym:
Thanks for the fast review! Also, the local dev environment was very easy to setup, kudos for that.

joelverhagen:
This is on PROD now.
https://www.nuget.org/packages/WindowsAzure.Storage/

PR #7000, opened 2019-03-16
[PROPOSAL] Tool to determine if a .nupkg has valid Microsoft metadata

joelverhagen:
This is minimal effort to make a tool that will allow teams to check the metadata of their package before pushing it to nuget.org. This is optional but will help teams catch metadata problems before shipping.

We can formalize the tool later, perhaps with `nuget.exe verify` and some kind of rules abstraction. But this is the cheap first step.

It's a old style tools package and requires .NET Framework 4.6.2. The .exe is here inside the package:

```
tools\NuGet.VerifyMicrosoftPackage.exe
```

In other words, it can be used like `NuGet.CommandLine` or `xunit.runner.console` where the user just hard codes the path inside the package of where the .exe is.

Teams with non-standard rules can specify a path to their custom rule set using `--rule-set`. We can communicate in a doc what owners have non-standard rule sets and what they are. There are 4 non-standard 4 rule sets in the PROD DB today.

We should also document that this does not cover author signature verification. This can easily be done with `nuget.exe verify -Signatures -CertificateFingerprint`.

## Open Questions

- How to know if the rules compiled into the tool are out of date?
   - We would release a new package to nuget.org after every master merge. Customers would just take latest package version
- Wouldn't a verify endpoint be easier? 🐱 

## Example output - invalid

Exit code is 1 (the number of bad packages).

```
Using the following package path argument:
C:\Users\jver\AppData\Local\Temp\NuGetTestFolder\5fda8093-4158-463b-b41e-de3e3b1fa2a1\inner\testA.nupkg

INVALID.
C:\Users\jver\AppData\Local\Temp\NuGetTestFolder\5fda8093-4158-463b-b41e-de3e3b1fa2a1\inner\testA.nupkg
The package TestPackage-65ed825e-dd81-4964-9599-f6a28d67171d 1.0.0 is not compliant.
There is 1 problem.
  - The package metadata defines 'Not Microsoft' as one of the authors which is not allowed by policy.

Valid package count: 0
Invalid package count: 1

The metadata validation used the following property names and JSON ruleset.

Readable .NET Name      | JSON Name
----------------------- | ----------
AllowedAuthors          | authors
AllowedCopyrightNotices | copy
ErrorMessageFormat      | error
IsLicenseUrlRequired    | licUrlReq
IsProjectUrlRequired    | projUrlReq
RequiredCoOwnerUsername | u

If question marks ('?') or weird characters appear below, consider using --write-default-rule-set.

{
  "u": "Microsoft",
[...]

PR #7834, opened 2020-02-04
Tool to determine if a .nupkg has valid Microsoft metadata

joelverhagen:
This is a new version of https://github.com/NuGet/NuGetGallery/pull/7000. I did this so that I can leave `jver-verify` as-is, which is hard-coded into Arcade:
https://github.com/dotnet/arcade/blob/a68b528ad1d0305f55c4fbdc977ebb0c02cfcc19/eng/common/post-build/nuget-validation.ps1#L12

Given that Arcade is using this in their official pipeline, it's time to merge this and get them to switch to a build that is properly signed and published by the NuGet team. The current dependency is precarious since the binary they are using is built by me and not published by the NuGet team.
https://www.nuget.org/packages/Jver.VerifyMicrosoftPackage/0.0.1-beta

We will be providing the tool as-is (no behavior changes).


src/VerifyMicrosoftPackage/verify.ps1
```diff
@@ -0,0 +1,119 @@
...
+   [string]$DownloadPath,
+   [Parameter(ValueFromRemainingArguments = $true)]
+   [string[]]$args
+)
+
+# The URL to download nuget.exe.
+$nugetExeUrl = "https://dist.nuget.org/win-x86-commandline/v4.9.4/nuget.exe"
+
+# The package ID of the verification tool.
+$packageId = "Jver.VerifyMicrosoftPackage"
```
loic-sharma: Is this the right package ID? Or should this be `NuGet.VerifyMicrosoftPackage`?

src/VerifyMicrosoftPackage/verify.ps1
```diff
@@ -0,0 +1,119 @@
...
+.PARAMETER args
+    Arguments that will be passed to the verification tool.
+.EXAMPLE
+    PS> .\verify.ps1 *.nupkg
+    Verifies the metadata of all .nupkg files in the currect working directory.
+.EXAMPLE
+    PS> .\verify.ps1 --help
+    Displays the help text of the downloaded verifiction tool.
+.LINK
+    https://github.com/NuGet/NuGetGallery/blob/jver-verify/src/VerifyMicrosoftPackage/README.md
```
loic-sharma: Should this point to the `master` branch to align with the readme?

src/VerifyMicrosoftPackage/verify.ps1
joelverhagen: You're right, `NuGet.VerifyMicrosoftPackage` is the correct one.

src/VerifyMicrosoftPackage/verify.ps1
joelverhagen: Yes.

PR #7838, opened 2020-02-06
Block version ranges that NuGet pack does not allow

joelverhagen:
When you try to pack a .nuspec with a floating version in it, NuGet pack prevents this by taking out the floating part and substituting in min version. "12.*" turns into "12.0.0". However, someone can hand-craft a .nuspec to get around it.

When we moved from 2.x APIs to 3.x APIs in gallery, we started supporting this without realizing it. Our .nuspec validation looks at each dependency version and parses it as a `VersionRange` which supports floating (since it's the same model that is used for `<PackageReference>` version).

These are the only packages that have utilized this:

Id | Version | Created
-- | -- | --
PInvokeCompiler  |  1.0.0-rc1 |  2017-01-09
Scissorhands.Helpers |   1.0.0-alpha-17  |  2016-01-13
Scissorhands.Helpers |   1.0.0-alpha-16  | 2016-01-13

Thanks @zivkan for the data.

We should block these since it has unpredictable impact on the restore graph and is a totally undocumented case for transitive package dependencies.

If we want to open the flood gates later we can but it would be coupled with official client support and a well thought through plan of how this will impact package consumers as a whole.

### DIFF
--- a/src/NuGetGallery.Core/Packaging/ManifestValidator.cs
+++ b/src/NuGetGallery.Core/Packaging/ManifestValidator.cs
@@ -145,6 +145,14 @@ namespace NuGetGallery.Packaging
                         }
 
                         // Versions
+                        if (dependency.VersionRange.IsFloating)
+                        {
+                            yield return new ValidationResult(String.Format(
+                                CultureInfo.CurrentCulture,
+                                CoreStrings.Manifest_InvalidDependencyVersionRange,
+                                dependency.VersionRange.OriginalString));
+                        }
+
                         if (dependency.VersionRange.MinVersion != null)
                         {
                             var versionRangeValidationResult =
--- a/tests/NuGetGallery.Core.Facts/Packaging/ManifestValidatorFacts.cs
+++ b/tests/NuGetGallery.Core.Facts/Packaging/ManifestValidatorFacts.cs
@@ -350,6 +350,20 @@ namespace NuGetGallery.Packaging
                   </metadata>
                 </package>";
 
+        private const string NuspecDependencyWithVersionRangeFormat = @"<?xml version=""1.0""?>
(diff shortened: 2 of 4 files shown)


tests/NuGetGallery.Core.Facts/Packaging/ManifestValidatorFacts.cs
```diff
@@ -498,6 +512,19 @@ public void ReturnsErrorIfDependencySetContainsDuplicateDependency()
             Assert.Equal(new[] { String.Format(CoreStrings.Manifest_DuplicateDependency, "net40", "SomeDependency") }, GetErrors(nuspecStream));
         }
 
+        [Theory]
+        [InlineData("*")]
+        [InlineData("1.0.0-beta*")]
```
ryuyu: Can we add one more case for the one @nkolev92 showed us, the 12.* case?

nkolev92:
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

src/NuGetGallery.Core/Packaging/ManifestValidator.cs
```diff
@@ -145,6 +145,14 @@ private static IEnumerable<ValidationResult> ValidateCore(PackageMetadata packag
                         }
 
                         // Versions
+                        if (dependency.VersionRange.IsFloating)
```
nkolev92: 🔪 🔪 🔪

tests/NuGetGallery.Core.Facts/Packaging/ManifestValidatorFacts.cs
joelverhagen: Added.

PR #7840, opened 2020-02-06
Enable using managed identities for KeyVault

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/2970

### DIFF
--- a/src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
+++ b/src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
@@ -63,11 +63,8 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="NuGet.Jobs.Common">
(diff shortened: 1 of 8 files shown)


shishirx34:
Are there any validation jobs related changes that you need to update from NuGet.Jobs?

joelverhagen:
I plan on updating the NuGet.Jobs reference when you merge https://github.com/NuGet/NuGet.Jobs/pull/857 but this PR is only intended on enabling MSIs for gallery. The jobs that run here will be evaluated separately (we can work together on who should drive that)

src/DatabaseMigrationTools/DatabaseMigrationTools.csproj
```diff
@@ -63,11 +63,8 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="NuGet.Jobs.Common">
```
joelverhagen: This comes in transitively from NuGet.Services.DatabaseMigration.

PR #7859, opened 2020-03-05
Reject broken ZIPs on API push with better message

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/3046

This was a little tricky to repro locally, when reading a ZIP of disk because the behavior of `FileStream` and `HttpInputStream` (the stream from an ASP.NET file upload) is different. `FileStream` does not throw when you seek out of bounds. It only throws when you read. `HttpInputStream` throws when you seek out of bounds, thus the `ArgumentOutOfRangeException` we see on server side.


src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -549,7 +549,18 @@ private async Task<ActionResult> CreatePackageInternal()
                                 Strings.PackageEntryFromTheFuture,
                                 entryInTheFuture.Name));
                         }
+                    }
+                    catch (Exception ex)
```
skofman1: is there anything more specific we can catch here?

joelverhagen: I'm concerned with catching something more generic since we'd be depending on an internal notion of `HttpInputStream`. I mimicked this:
https://github.com/NuGet/NuGetGallery/blob/2a8ab4b70c7fdcae9ffd637192515b40e02d78f0/src/NuGetGallery/Controllers/PackagesController.cs#L399

joelverhagen: Added a comment, as discussed offline.

PR #7862, opened 2020-03-05
Expose domain and diagnostics source in MEF container

joelverhagen:
This enables synchronous initialization (via the constructor) of the cookie compliance shim.
Also, move the refresh to the background so that initial start-up is not slowed down.

Fully lights up with https://nuget.visualstudio.com/NuGetMicrosoft/_git/NuGet.Shims/pullrequest/1342.
Related to https://github.com/NuGet/Engineering/issues/2980.

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -11,7 +11,6 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Mail;
 using System.Security.Principal;
-using System.Threading;
 using System.Threading.Tasks;
 using System.Web;
 using System.Web.Hosting;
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -1419,26 +1426,27 @@ private static void RegisterAuditingServices(ContainerBuilder builder, IAuditing
...
+            CookieComplianceServiceBase service = null;
+            if (configuration.Current.IsHosted)
             {
-                service = new NullCookieComplianceService();
+                var siteName = configuration.GetSiteRoot(true);
+                service = GetAddInServices<ICookieComplianceService>(sp =>
+                {
+                    sp.ComposeExportedValue("Domain", siteName);
+                    sp.ComposeExportedValue<IDiagnosticsService>(diagnostics);
+                }).FirstOrDefault() as CookieComplianceServiceBase;
```
agr: Do we expect multiple services here? Should we use `SingleOrDefault` instead of `FirstOrDefault`?

joelverhagen: It is possible for there to be multiple, technically, given the way MEF works. Our particular add-in drop process will only have one but someone could have multiple theoretically. I don't feel strongly, I just took the existing implementation forward.

PR #7867, opened 2020-03-13
Add KnownOperation telemetry and add more configuration to DI

joelverhagen:
This change has two main purposes:

1. Add a `KnownOperation` custom dimension so that we can have a metric column with low cardinality that identifies the type of request. Current the `name` (or `operation_Name` in other tables) has extremely high cardinality because some request patterns end up putting the verbatim URL path there.
1. Add more configuration types to the DI container so that closed-source/internal modules can get configuration from DI.
   - Since Gallery does not use `IOptions` or `IConfigurationRoot`, we use the older abstraction in our codebase which is `IConfigurationFactory`. This has been around for a while and is based on `ComponentModel`.
   - Update to new ServerCommon since I moved `StringArrayConverter` there. This allows closed-source implementations to also use `StringArrayConverter` in their own private configuration types.

Progress on https://github.com/NuGet/Engineering/issues/3020


src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -223,7 +223,7 @@ public static void RegisterUIRoutes(RouteCollection routes)
                 "packages/{id}/{version}",
                 new
                 {
-                    controller = "packages",
+                    controller = "Packages",
```
joelverhagen: This is why the operation name was `GET packages/DisplayPackage` (lowercase `packages`)... very silly.

src/NuGetGallery.Services/Configuration/AppConfiguration.cs
```diff
@@ -5,6 +5,7 @@
 using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
 using System.Net.Mail;
+using NuGet.Services.Configuration;
```
loic-sharma: Is this needed?

joelverhagen: Yes, I moved the `StringArrayConverter` attribute to this namespace (via package upgrade)

src/NuGetGallery.Services/Configuration/ConfigurationService.cs
```diff
@@ -73,6 +73,21 @@ public string GetSiteRoot(bool useHttps)
             return useHttps ? _httpsSiteRootThunk.Value : _httpSiteRootThunk.Value;
         }
 
+        public Task<T> Get<T>() where T : NuGet.Services.Configuration.Configuration, new()
+        {
+            // Get the prefix specified by the ConfigurationKeyPrefixAttribute on the class if it exists.
+            var classPrefix = string.Empty;
+            var configNamePrefixProperty = (ConfigurationKeyPrefixAttribute)typeof(T)
+                .GetCustomAttributes(typeof(ConfigurationKeyPrefixAttribute), true)
```
loic-sharma: Super nit:

```suggestion
                .GetCustomAttributes(typeof(ConfigurationKeyPrefixAttribute), inherit: true)
```

PR #7870, opened 2020-03-14
Optional configurable thread settings

agr:
Enables to set up thread pool in app service. Cloud service was set up through startup.ps1


src/NuGetGallery/Services/StatusService.cs
```diff
@@ -50,13 +51,17 @@ public async Task<ActionResult> GetStatus()
...
+
             return new HttpStatusCodeWithBodyResult(AvailabilityStatusCode(galleryServiceAvailable),
                 String.Format(CultureInfo.InvariantCulture,
                     StatusMessageFormat,
                     AvailabilityMessage(galleryServiceAvailable),
                     AvailabilityMessage(sqlAzureAvailable),
                     AvailabilityMessage(storageAvailable),
-                    HostMachine.Name));
+                    HostMachine.Name,
+                    workerThreads, ioCompletionThreads));
```
agr: Should I have it removed?

joelverhagen: I think it's probably fine. Although perhaps it's a sniff test for our service being under a lot of load/close to a breaking point?

shishirx34: +1 for the sniff test point. I don't think we should expose this information if we can avoid it.

PR #7872, opened 2020-03-16
Lazy DbContext initialization in OData controllers

agr:
Since we do the hijacking for great deal of OData requests we get, having DB contexts injected into OData controllers creates unnecessary connections that are likely to be unused.

This change postpones the context initialization until the moment we know for sure we are going to query the DB.

We have some issues with reporting the DB related perf counters from the app service, so verification was done by collecting profiler traces from different versions of the service.


skofman1:
how was this tested?

loic-sharma:
How did we choose this approach? It seems that our connection factory eagerly opens connections: https://github.com/NuGet/ServerCommon/blob/d919d67f87a49f30dcdf9465139a270e80e32d28/src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs#L67

Could we instead let EF open connections as necessary? Also, did we verify we are properly pooling our SQL connections? I wonder if setting `.AccessToken` may be causing pooling issues: https://github.com/NuGet/ServerCommon/blob/d919d67f87a49f30dcdf9465139a270e80e32d28/src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs#L82

zhhyu:
I think we don't provide an already opened DB connection, when we build the DbContext:
https://github.com/NuGet/NuGetGallery/blob/91729012653b2625f5aaf54401251effa4e99715/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L834
I think it's EF who decides when to open the connection, especially when we use Linq to query the data.

loic-sharma:
> I think we don't provide an already opened DB connection, when we build the DbContext:
> https://github.com/NuGet/NuGetGallery/blob/91729012653b2625f5aaf54401251effa4e99715/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L834
> 
> 
> I think it's EF who decides when to open the connection, especially when we use Linq to query the data.

That is the default behavior for most EF apps, however, the Gallery opens the connection manually:

1. We create the DBContext with a connection by calling `CreateDbConnection`: https://github.com/NuGet/NuGetGallery/blob/91729012653b2625f5aaf54401251effa4e99715/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L180
1. `CreateDbConnection` calls `AzureSqlConnectionFactory.CreateAsync`: https://github.com/NuGet/NuGetGallery/blob/91729012653b2625f5aaf54401251effa4e99715/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L834
1. `AzureSqlConnectionFactory.CreateAsync` calls `ConnectAsync`: https://github.com/NuGet/ServerCommon/blob/d919d67f87a49f30dcdf9465139a270e80e32d28/src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs#L60
1. `OpenAsync` calls `OpenConnectionAsync`: https://github.com/NuGet/ServerCommon/blob/d919d67f87a49f30dcdf9465139a270e80e32d28/src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs#L67
1. `OpenConnectionAsync` calls `SqlConnection.OpenAsync`: https://github.com/NuGet/ServerCommon/blob/d919d67f87a49f30dcdf9465139a270e80e32d28/src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs#L91

loic-sharma:
⚠ By the way the `CreateDbConnection` looks like it may be susceptible to thread pool starvation due to sync-over-async. That may be causing the slow context creation especially since this is on a hot path (OData). Notice the `Task.Run(() => ...).Result;` here: 

https://github.com/NuGet/NuGetGallery/blob/91729012653b2625f5aaf54401251effa4e99715/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L834

For more information on sync-over-async see: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#warning-deadlocks

agr:
@skofman1, deployed to dev, passed functional/E2E tests, added logging for DB connection creation, added logging for hijacked/non-hijacked requests, spammed the DEV service with various requests (both hijacked and non-hijacked) on /Packages and /FindPackagesById endpoints. Checked the logs that no DB connection was created for hijacked requests.

agr:
It was the `CreateDbConnection` calls that were slow. It very well may be that it's a sync over async issue.

agr:
Regarding connection pooling: as far as I understand, the "pool key" is the connection string, so as long as connection strings are same, they should be pooled. To verify that we need to get SQL client perf counters from app service that seem to unable to get.

agr:
In order to completely get rid of that `Task.Run().Result`, it would be required to do quite a rewrite of the way we obtain a DB connection instance. This change should help us to verify that it makes sense start figuring out how to do that.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -829,8 +829,10 @@ private static void RegisterAsynchronousEmailMessagingService(ContainerBuilder b
             return new AzureSqlConnectionFactory(connectionString, secretInjector, logger);
         }
 
-        private static DbConnection CreateDbConnection(ISqlConnectionFactory connectionFactory)
+        private static DbConnection CreateDbConnection(ISqlConnectionFactory connectionFactory, ITelemetryClient telemetryClient)
         {
+            var tds = new TraceDiagnosticsSource(nameof(OwinStartup), telemetryClient);
```
zhhyu: Why do we create a new tds here? Why not directly use the already registered "diagnosticsService"?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -829,8 +829,10 @@ private static void RegisterAsynchronousEmailMessagingService(ContainerBuilder b
...
+            tds.Information($"Creating DB connection to {connectionFactory.DataSource}, DB: {connectionFactory.InitialCatalog}");
```
zhhyu: Maybe it's better that we also log the time cost of this Task.Run().Result, so we can have on-line statistics of this critical metric for us to improve later?

zhhyu: Maybe the log information mentions like "Obtain SqlConnection to ...", which is clearer? The "DB connection" sounds like a physical connection.

tests/NuGetGallery.Facts/Controllers/ODataV1FeedControllerFacts.cs
```diff
@@ -9,6 +9,7 @@
 using NuGetGallery.OData;
 using Moq;
 using Xunit;
+using System;
```
zhhyu: System is the top one.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
agr: All the added logging is going to be removed.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
agr: Added instrumentation of the execution time. Removed logs.

agr: Removed log

PR #7925, opened 2020-03-30
Add link to search for pending validations

loic-sharma:
Today the on-call engineer has to connect to the database to find the pending validations. This adds a link to the admin panel to fetch the currently pending validations:

image

Addresses https://github.com/NuGet/NuGetGallery/issues/7926

Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3595685
Release: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=628330

### DIFF
--- a/src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
+++ b/src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
@@ -32,9 +32,9 @@ namespace NuGetGallery.Areas.Admin.Controllers
         {
             var validationSets = _validationAdminService.GetPending();
             var validatedPackages = ToValidatedPackages(validationSets);
-            var validationSetIds = validationSets
-                .Select(s => s.ValidationTrackingId)
-                .Distinct();
+            var validationSetIds = validatedPackages
+                .SelectMany(p => p.ValidationSets)
+                .Select(s => s.ValidationTrackingId);
             var query = string.Join("\r\n", validationSetIds);
 
             return View(nameof(Index), new ValidationPageViewModel(query, validatedPackages));
--- a/src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
+++ b/src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
@@ -30,12 +30,12 @@ namespace NuGetGallery.Areas.Admin.Controllers
         [HttpGet]
         public virtual ActionResult Pending()
         {
-            var packageValidationSets = _validationAdminService.Pending();
-            var validatedPackages = ToValidatedPackages(packageValidationSets);
-            var packageIdentities = packageValidationSets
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
```diff
@@ -27,15 +27,35 @@ public virtual ActionResult Index()
...
+        public virtual ActionResult Pending()
+        {
+            var validationSets = _validationAdminService.GetPending();
+            var validatedPackages = ToValidatedPackages(validationSets);
+            var validationSetIds = validationSets
+                .Select(s => s.ValidationTrackingId)
+                .Distinct();
+            var query = string.Join("\r\n", validationSetIds);
+
+            return View(nameof(Index), new ValidationPageViewModel(query, validatedPackages));
```
joelverhagen: Could we redirect the "Search" page with the `q` parameter filled in so the page is refreshable and the data doesn't just disappear?

src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
```diff
@@ -27,15 +27,35 @@ public virtual ActionResult Index()
             return View(nameof(Index), new ValidationPageViewModel());
         }
 
+        [HttpGet]
+        public virtual ActionResult Pending()
+        {
+            var validationSets = _validationAdminService.GetPending();
+            var validatedPackages = ToValidatedPackages(validationSets);
+            var validationSetIds = validationSets
```
joelverhagen: nit sort by lex order for consistent behavior

src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
loic-sharma: That was my first thought too, but I feel iffy redirecting to a potentially very long query string if there are tons of pending validations. Friday had over 100 pending validations. I’m not sure what our query string limits are, but I’d like this page to work well even if we have a really bad incident with several hundred pending validations. I don’t feel super strongly one way or another, please let me know if I missed something!

Also, the search box has the validation set IDs. One can press”Search” to refresh the current validations.

I do

joelverhagen: If you can press Search to refresh the current set that's perfect.

src/NuGetGallery/Areas/Admin/Controllers/ValidationController.cs
loic-sharma: The view's ordering is a little more complicated - it orders by package identity, package creation, validation set key, and then package type (package vs symbols). Rather than copy that logic, I updated the query generation to use the view model's `validatedPackages` as that is already sorted.

Issue #7889, opened 2020-03-20
[A11y]Narrator is reading improper role for Expand/Collapse button


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1069656
Repro Steps:​
1.Start Screen Reader. Hit the URL "https://www.nuget.org" and sign in with valid credentials​
2.Tab till the user menu button and press enter to expand it. Select "Manage Packages" and press enter​
3.Tab till "Published Packages", expand it and select any existing package.(If no packages are available, add a new package)​
4.Tab till "Manage Package" link under "Info" and press enter​
5.Tab till "Deprecation" section, enter required details, tab till "Save" button and press enter​
6.Tab till the package name in breadcrumb navigation and press enter​
7.Navigate to https://www.nuget.org/packages/DeltaX.fabel/​
8.Tab till the expand button present for the alert message that is displayed below the page heading​
9.Verify if the screen reader is reading the proper role for Expand/Collapse button​
​
Actual Result:​
Narrator is reading the Expand/Collapse button as "Group"
​
NOTE: NVDA is reading the role as button but it is not announcing the state(expanded/collapsed)

Expected Result:​
Narrator should read the role as button. Also, Narrator should read the state of the expand/collapse button, e.g.: button Expanded

Notes:
Giving this element a role="button" attribute seems to fix this issue.

Issue #7885, opened 2020-03-19
[A11y]Focus is not visible on the expand/collapse button


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1069655
Repro Steps:​
1.Hit the URL "https://www.nuget.org" and sign in with valid credentials​
2.Tab till the user menu button and press enter to expand it. Select "Manage Packages" and press enter​
3.Tab till "Published Packages", expand it and select any existing package.(If no packages are available, add a new package)​
4.Tab till "Manage Package" link under "Info" and press enter​
5.Tab till "Deprecation" section, enter required details, tab till "Save" button and press enter​
6.Tab till the package name in breadcrumb navigation and press enter​
7.Navigate to https://www.nuget.org/packages/DeltaX.fabel/​
8.Tab till the expand button present for the alert message that is displayed below the page heading​
9.Verify if the focus is visible on the Expand/Collapse button​
​
Actual Result:​
Focus is not visible on the expand/collapse button​
​
Expected Result:​
Focus should be clearly visible on the expand/collapse button​

Notes:
We will need to add some stuff in our CSS for the deprecation container focus outline. The current appears to be overridden to be non-existent.

ryuyu:
Looks like we can set outline-style for this element (focused) to "auto" and this fixes the issue.

ryuyu:
This was actually introduced by https://github.com/NuGet/NuGetGallery/pull/7326 when trying to address https://github.com/NuGet/NuGetGallery/issues/7305

Issue #7882, opened 2020-03-19
[A11y]'MIT' Link is not descriptive.


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1063680
Steps to Reproduce:
1. Hit the URL: https://www.nuget.org/
2. Navigate to "Sign in" control and hit enter.
3. Navigate to "Sign in with Microsoft" control and hit enter.
4. Search for a key term (say “json”) from search box.
5. Click on "Newtonsoft.json" control and hit enter.
6. Navigate to "Info" and "Statistics" section present on right of the screen.
7. Start the Narrator.
8. Navigate to "MIT" link under info.
9. Observe the Narrator announces the "MIT" link descriptively or not.
 
Actual:
Narrator is not announcing the proper description of 'MIT' link.

Similar issue has been observed with Chrome+NVDA

Expected:
The narrator should announce the description of the link.

Notes:
Context may be lost here that 'MIT' is actually descriptive of the license type. We should add an aria-label indicating "License type: MIT" or something similar.

Issue #7879, opened 2020-03-19
[A11y]"Learn more" link is not descriptive.


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1062807
Steps to Reproduce:
1. Hit the #URL:https://www.nuget.org/
2. Navigate to "Sign in" control and hit enter.
3. Navigate to "Sign in with Microsoft" control and hit enter.
4. Select “Manage Organizations” from the Profile Dropdown and hit enter.
5. Navigate to "Add New" control and hit enter.
6. Turn on narrator, navigate to "Learn" more link using Tab key.
7. Observe Narrator announcement.

Actual:
Narrator not announces "Learn More" link in descriptive. 
When "Learn More" link receives Narrator focus it does not describe what the learn more link means.

Same issue is Observed in "Chrome+NVDA" also.

Same issue is Observed for controls "Learn More" and "Read More" in  Scenario "UploadPackage_Verify". Below is the screenshot.
Image
Expected:
The narrator should announce a short description about the learn more link.

Notes:
Not exactly sure what is expected to be announced here since "Learn more" is already in the context of the organizations page. We could update the "Learn More" link to include an aria-label with the text "Learn more about organizations on NuGet.Org"

PR #7927, opened 2020-03-30
[A11y]Accessibility fixes 1

ryuyu:
Addresses the following issues:
[A11y]'MIT' Link is not descriptive. #7882
 - Updated link description to read "License {LicenseType}" instead of "{LicenseType}"

[A11y]"Learn more" link is not descriptive. #7879
 - Update link description to read "Learn more about Organizations on NuGet.Org" from "Learn more"

[A11y]Focus is not visible on the expand/collapse button #7885
 - This was actually introduced by #7326 when trying to address #7305
 - Undo this change, reintroduces #7305 on click.

[A11y]Narrator is reading improper role for Expand/Collapse button #7889
 - Gave element attribute role="button"

[A11y]Narrator does not announce the error message "At least one scope must be selected" after unchecking the 'Push' button. #7876
 - Added aria-live region and role="alert"
 - Delayed update of text for readme alerts so element can become visible before being updated (updating invisible elements are not read out).

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-edit-readme.js
+++ b/src/NuGetGallery/Scripts/gallery/page-edit-readme.js
@@ -345,7 +345,7 @@ var BindReadMeDataManager = (function () {
             // the message must change AFTER the element becomes visible. When we run all these commands
             // synchronously, it appears that the message content is changed BEFORE the element is finished rendering.
             // Delay it using a timeout so it will show after the message box is visible.
-            setTimeout(() => { $("#readme-error-content").text(errorMsg); }, 100);
+            setTimeout(() => { $("#readme-error-content").text(errorMsg); }, 0);
             
         }
 
--- a/src/NuGetGallery/Scripts/gallery/page-edit-readme.js
+++ b/src/NuGetGallery/Scripts/gallery/page-edit-readme.js
@@ -340,7 +340,13 @@ var BindReadMeDataManager = (function () {
         function displayReadMeError(errorMsg) {
             $("#readme-errors").removeClass("hidden");
             $("#preview-readme-button").attr("disabled", "disabled");
-            $("#readme-error-content").text(errorMsg);
+
+            // This is required because in order for the screen reader to correctly read the message and alert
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/Scripts/gallery/page-edit-readme.js
```diff
@@ -340,7 +340,13 @@ var BindReadMeDataManager = (function () {
         function displayReadMeError(errorMsg) {
             $("#readme-errors").removeClass("hidden");
             $("#preview-readme-button").attr("disabled", "disabled");
-            $("#readme-error-content").text(errorMsg);
+
+            // This is required because in order for the screen reader to correctly read the message and alert
+            // the message must change AFTER the element becomes visible. When we run all these commands
+            // synchronously, it appears that the message content is changed BEFORE the element is finished rendering.
+            // Delay it using a timeout so it will show after the message box is visible.
+            setTimeout(() => { $("#readme-error-content").text(errorMsg); }, 100);
```
joelverhagen: How do we know 100ms is enough? If we do 0ms will it work?

loic-sharma: It looks like from the jQuery documentation that the `removeClass` has a `complete` parameter that accepts a callback: https://api.jqueryui.com/removeClass/

Should we use that `complete` parameter?

ryuyu: I'll try these options.

ryuyu: Looks like complete parameter does not appear to work for this case, but setting 0ms does still seem to work. I'll update to 0 ms.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -159,7 +159,7 @@
 
         @switch (packageManager.AlertLevel)
         {
-            case AlertLevel.Info:
+            case AlertLevel.Info:
```
agr: nit: remove extra space?

Issue #7937, opened 2020-04-07
[2020-04-06] Deployment


agr:
### AccountDeleter

**Note:** this must be deployed to all regions before gallery

#### Changes
* Allow deletion of unconfirmed users and orgs (#7948) 

#### Build/release.
Migration release: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=664942
Release: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=665105

#### Progress
 - [x] Merge PR: https://github.com/NuGet/NuGetGallery/pull/7938
 - [x] DEV
 - [x] INT
 - [x] PROD

#### Changes
* Configuration update with new auto-scale changes.
* Updated MaxSupportedMinClientVersion to match the client version (#7899)
* Add link to search for pending validations (#7925)
* [A11y]Accessibility fixes 1 (#7927)
* [Rename] DB schema change (#7923)
* Update to latest ServerCommon and NuGet.Jobs version (#7929)
* Remove "-staging" suffix from cloud role name (#7935)
* Add CustomerResourceId to enrich push metrics with tenant ID (#7936)
* [A11y]Accessibility fixes 2 (#7939) 
*Allow deletion of unconfirmed users and orgs (#7948)
* Add simulated error for exceptions in view and error page (#7958)
* Remove redirects from error handling and clean Elmah (#7960)
* Use ToLowerInvariant to avoid locale issues. (#7967)
* Fix minification failures and delete unused assets (#7963)
* Add support for alternate instrumentation system in browser (#7971)
* Improved link functionality for nuget.org (#7942)
* [Rename] Display rename information on package detail page (#7964)

#### Build/release.
Migration release: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=664942

Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3700769&view=results
Release: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=666291

#### Progress
 - [x] Merge PR: https://github.com/NuGet/NuGetGallery/pull/7938
 - [x] Run migration for DEV
 - [x] DEV
 - [x] Run migration for INT
 - [x] INT
 - [x] Run migration for PROD
 - [x] PROD Cloud service
 - [x] PROD App service
[...]

PR #7938, opened 2020-04-07
[ReleasePrep][2020-04-07] RI dev to master

agr:
Deployment work item: https://github.com/nuget/nugetgallery/issues/7937


joelverhagen:
⚠️ we should have a deployment item. There is a migration that must be run first.

zhhyu:
We have to postpone this until the frozen time passes.

PR #7948, opened 2020-04-14
Allow deletion of unconfirmed users and orgs

joelverhagen:
Fix minor styling issue on the delete account page when an exception bubbles out.
Fix bug in the account delete handler which would not delete users when notifications were turned off.
Address https://github.com/NuGet/Engineering/issues/3099

Consider [hiding whitespace in the diff](https://github.com/NuGet/NuGetGallery/pull/7948/files?w=1).


src/AccountDeleter/AccountDeleteMessageHandler.cs
```diff
@@ -113,7 +118,7 @@ public async Task<bool> HandleAsync(AccountDeleteMessage command)
             catch (Exception e)
             {
                 _logger.LogError(0, e, "An unknown exception occured: {ExceptionMessage}");
```
loic-sharma, line 120: Hey if you have extra time could you fix this log message? It’s missing a parameter.

Issue #7922, opened 2020-03-27
Packages with duplicate entries in zip files are not blocked


skofman1:
This created several issues: 
https://github.com/NuGet/NuGetGallery/issues/7921
https://github.com/NuGet/NuGetGallery/issues/7744

PR #7954, opened 2020-04-17
Packages with duplicated entries should be block on upload.

dannyjdev:
A new validation was added at the ValidateBeforeGeneratePackageAsync method.

https://github.com/NuGet/NuGetGallery/issues/7922


src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -599,6 +606,18 @@ public UserContentEnabledNuspecReader(Stream stream)
             return null;
         }
 
+        private PackageValidationResult CheckPackageDuplicatedEntriesAsync(PackageArchiveReader nuGetPackage)
```
loic-sharma, line 609:
```suggestion
        private PackageValidationResult CheckPackageDuplicatedEntries(PackageArchiveReader nuGetPackage)
```

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -599,6 +606,18 @@ public UserContentEnabledNuspecReader(Stream stream)
...
+        {
+            var packageFiles = nuGetPackage.GetFiles();
+
+            if(packageFiles.Count() != packageFiles.Distinct().Count())
```
loic-sharma, line 613: Super nit:

```suggestion
            if (packageFiles.Count() != packageFiles.Distinct().Count())
```

tests/NuGetGallery.Core.Facts/TestUtils/PackageServiceUtility.cs
```diff
@@ -189,7 +191,16 @@ public static Mock<TestPackageReader> CreateNuGetPackage(Stream testPackage)
                             writer.Write("Fake signature file.");
                         }
                     }
-                }, desiredTotalEntryCount: desiredTotalEntryCount,
+
+                    if(entryNames != null)
```
loic-sharma, line 195: Super nit:

```suggestion
                    if (entryNames != null)
```

src/NuGetGallery/Services/PackageUploadService.cs
loic-sharma, line 613: Should this be case insensitive? What if I have entries `a.txt` and `A.TXT`? Consider adding a test to cover this too

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -599,6 +606,18 @@ public UserContentEnabledNuspecReader(Stream stream)
             return null;
         }
 
+        private PackageValidationResult CheckPackageDuplicatedEntriesAsync(PackageArchiveReader nuGetPackage)
+        {
+            var packageFiles = nuGetPackage.GetFiles();
```
joelverhagen, line 611: How does this handle normalization of paths, directory seperators, case, etc?

joelverhagen:
What kind of packages are published today that fail against this new validation?

src/NuGetGallery/Services/PackageUploadService.cs
dannyjdev: Updated.

src/NuGetGallery/Services/PackageUploadService.cs
dannyjdev: Updated.

tests/NuGetGallery.Core.Facts/TestUtils/PackageServiceUtility.cs
dannyjdev: Updated.

src/NuGetGallery/Services/PackageUploadService.cs
dannyjdev: I didn't cover case sensitivity at first, I think that could not trigger the issue I was using as an example for this fix, but it makes sense to have that, updated the code to consider that.

src/NuGetGallery/Services/PackageUploadService.cs
dannyjdev: I just added a Path.GetFullPath to cover the normalization, and also a minor modification to consider case sensitivity, I wasn't considering that at first because the issue I saw was related to duplicated files that seems to be only for files having the exact same route.

dannyjdev:
> What kind of packages are published today that fail against this new validation?

joelverhagen:
> I wanted to first have this PR completed to add this validation to the analyzer, it would be a minor addition to that since I have updated almost all the code.

I don't think we should check this in until we know it won't unexpectedly reject a lot of packages.

joelverhagen:
After we have analyzed the currently available packages that fail this validation and they all look legitimately bad/broken, :shipit:. The goal of analyzing existing packages is to act as a sanity check of the new validation to ensure we are not actually blocking a legitimate scenario.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -599,6 +606,19 @@ public UserContentEnabledNuspecReader(Stream stream)
             return null;
         }
 
+        private PackageValidationResult CheckPackageDuplicatedEntries(PackageArchiveReader nuGetPackage)
+        {
+            // Normalize paths and ensures case sensitivity is also considered
+            var packageFiles = nuGetPackage.GetFiles().Select(packageFile => Path.GetFullPath(packageFile.ToLower()));
```
joelverhagen, line 612: It's recommend `ToLowerInvariant` or using an `OrdinalIgnoreCase` comparer.

joelverhagen: Note that `Distinct` has an overload that accepts a comparer:
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.distinct?view=netframework-4.8#System_Linq_Enumerable_Distinct__1_System_Collections_Generic_IEnumerable___0__System_Collections_Generic_IEqualityComparer___0__

tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
```diff
@@ -509,6 +509,48 @@ public async Task WithNotTooManyPackageEntries_WhenRejectPackagesWithTooManyPack
...
+                    _currentUser);
+
+                // Assert
+                Assert.Equal(PackageValidationResultType.Invalid, result.Type);
+                Assert.Equal("The package contains one or more duplicated files in a same folder.", result.Message.PlainTextMessage);
+                Assert.Empty(result.Warnings);
+            }
+
+            [Theory]
+            [InlineData("noDuplicatedFile.txt", "./temp/noDuplicatedFile.txt")]
```
joelverhagen, line 534: what about varying slashes? `/` and `\` in the same package

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -599,6 +606,19 @@ public UserContentEnabledNuspecReader(Stream stream)
...
+
+            if (packageFiles.Count() != packageFiles.Distinct().Count())
```
shishirx34, line 614: so... then I am assuming `lib\f1\binary.dll` and `lib\f2\binary.dll` is allowed?

joelverhagen: this is DEFINITELY okay and very common in packages. Consider a package with both `net45` assets and `netstandard1.0` assets:

- `lib\net45\binary.dll`
- `lib\netstandard1.0\binary.dll`

src/NuGetGallery/Strings.resx
```diff
@@ -957,6 +957,9 @@ Policy violations: {0}</value>
   <data name="UploadPackage_PackageContainsTooManyEntries" xml:space="preserve">
     <value>The package contains too many files and/or folders.</value>
   </data>
+  <data name="UploadPackage_PackageContainsDuplicatedEntries" xml:space="preserve">
+    <value>The package contains one or more duplicated files in a same folder.</value>
```
loic-sharma, line 961:
```suggestion
    <value>The package contains one or more duplicated files in the same folder.</value>

```

dannyjdev: Updated.

src/NuGetGallery/Services/PackageUploadService.cs
dannyjdev: Updated.

tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
dannyjdev: Added test cases that cover that.

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -599,6 +606,19 @@ public UserContentEnabledNuspecReader(Stream stream)
             return null;
         }
 
+        private PackageValidationResult CheckPackageDuplicatedEntries(PackageArchiveReader nuGetPackage)
+        {
+            // Normalize paths and ensures case sensitivity is also considered
+            var packageFiles = nuGetPackage.GetFiles().Select(packageFile => Path.GetFullPath(packageFile.ToLowerInvariant()));
+
+            if (packageFiles.Count() != packageFiles.Distinct(StringComparer.OrdinalIgnoreCase).Count())
```
joelverhagen, line 614: You only need either the case insensitive comparer or `ToLowerInvariant` but not both. `OrdinalIgnoreCase` is preferable since it does not allocate lowercased strings.

dannyjdev: Updated.

tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
```diff
@@ -509,6 +509,51 @@ public async Task WithNotTooManyPackageEntries_WhenRejectPackagesWithTooManyPack
                 Assert.Empty(result.Warnings);
             }
 
+            [Theory]
+            [InlineData("duplicatedFile.txt", "duplicatedFile.txt")]
+            [InlineData("./temp/duplicatedFile.txt", "./temp/duplicatedFile.txt")]
+            [InlineData("./temp/duplicatedFile.txt", "./temp\\duplicatedFile.txt")]
+            [InlineData("./temp\\duplicatedFile.txt", "./temp\\duplicatedFile.txt")]
+            [InlineData("duplicatedFile.txt", "duplicatedFile.TXT")]
```
agr, line 517: How about the following cases:
```
[InlineData("./duplicatedFile.txt", "duplicatedFile.txt")]
[InlineData("./duplicatedFile.txt", "/duplicatedFile.txt")]
[InlineData("/duplicatedFile.txt", "duplicatedFile.txt")]
[InlineData(".\\duplicatedFile.txt", "./duplicatedFile.txt")]
```
Should we consider those?

src/NuGetGallery/Services/PackageUploadService.cs
agr, line 612: I am a bit worried about using `GetFullPath` here. It _might or might not_ produce different output for `.\file` and for `\file` depending on the call context (i.e. whether the current directory is in the root of some drive or not). I don't have any better suggestions though. Could we at least verify that the current directory is not root?

tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
dannyjdev: For the files that are added with the prefix "/", full path for those are "C:\duplicatedfile.txt" instead of "<folder>\duplicatedfile.txt".

agr: Question is, are those paths really considered different by zip extractors? And whether all that is relevant to the underlying issue.

dannyjdev: I'm checking that with the client team.

src/NuGetGallery/Services/PackageUploadService.cs
dominoFire: In NuGet.Client, we rely on `PathUtils.StripLeadingDirectorySeparators(string)` to normalize paths. Some usage examples are on the [tests](https://github.com/NuGet/NuGet.Client/blob/4dbf16e68eabf507b9f156c801b1b7c3b0586e43/test/NuGet.Core.Tests/NuGet.Common.Test/PathUtilityTests.cs#L23-L30).

tests/NuGetGallery.Facts/Services/PackageUploadServiceFacts.cs
dominoFire: Sometimes, those paths are normalized `PathUtils.StripLeadingDirectorySeparators(string)` 
 but, as far as I know, we don't normalize content that it is written on the nuspec. For example in the <packageIcon/> entry, that information is packaged as-is and then, it is normalized when the icon is read in the PM UI.

Issue #7868, opened 2020-03-13
Some exceptions thrown in gallery cause 302 redirect instead of 500


joelverhagen:
This means that in Application Insights that request is considered `success = true` which messes up analytics. Really, this should be considered a failed request. Because we redirect to shared error page, we lose the context at this high level that an error occurred. You can tease it apart on a case-by-case basis by finding the subsequent URL that the customer hit and see that it was the error page. Also, you can see an exception thrown for the 302 request, but this is not ideal.

I think we should change the code to just show the HTTP 500 error page inline, without a redirect.

It looks like our error handling functional tests don't cover this path. Additional cases that need to be covered:
- Exception thrown in `ViewHelpers`
- Nasty cookies like `__Controller::TempData=<script>alert(1)</script>`
- Nasty requests like ones with pipes in path (can't repro, but visible in AI)

PR #7958, opened 2020-04-21
Add simulated error for exceptions in view and error page

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7868.
Also switched from diagnostics service to logger factory to aid in debugging.

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -11,6 +11,7 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Mail;
 using System.Security.Principal;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Web;
 using System.Web.Hosting;
@@ -171,7 +172,7 @@ namespace NuGetGallery
                 .InstancePerLifetimeScope();
 
             var galleryDbConnectionFactory = CreateDbConnectionFactory(
-                diagnosticsService,
+                loggerFactory,
                 nameof(EntitiesContext),
                 configuration.Current.SqlConnectionString,
                 secretInjector);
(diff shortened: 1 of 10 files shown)


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -829,10 +830,13 @@ private static void RegisterAsynchronousEmailMessagingService(ContainerBuilder b
...
-        private static ISqlConnectionFactory CreateDbConnectionFactory(IDiagnosticsService diagnostics, string name,
-            string connectionString, ISecretInjector secretInjector)
+        private static ISqlConnectionFactory CreateDbConnectionFactory(
+            ILoggerFactory loggerFactory,
+            string name,
+            string connectionString,
+            ISecretInjector secretInjector)
         {
-            var logger = diagnostics.SafeGetSource($"AzureSqlConnectionFactory-{name}");
+            var logger = loggerFactory.CreateLogger($"AzureSqlConnectionFactory-{name}");
```
joelverhagen, line 839: This is much better because now any exception logged by `AzureSqlConnectionFactory` will be logged fully, instead of just the log message which is the `DiagnosticsService` behavior.

tests/NuGetGallery.FunctionalTests/ErrorHandling/ErrorHandlingTests.cs
```diff
@@ -93,6 +106,48 @@ public async Task PageThatDoesNotExist(string relativePath)
...
+        public async Task ErrorInErrorPageWithoutPath()
+        {
+            // Arrange
+            var cookies = new SimulatedErrorRequest(EndpointType.Pages, SimulatedErrorType.ExceptionInDedicatedErrorPage).GetCookies();
+
+            // Act
+            var response = await GetTestResponseAsync("/Errors/500", cookies);
+
+            // Assert
+            Validator.Redirect("500")(response);
```
joelverhagen, line 124: This is why we sometimes see 500s turn to redirects. It's when the error page itself throws and exception.

Issue #7961, opened 2020-04-23
NuGet.org - Api key is generated/copied with the local culture that contains 'ı'


JnRMnT:
### Describe the bug
I am using Nuget.Org Api Key generation to connect and publish my NuGet packages from Azure DevOps. I had noticed earlier that i needed to generate the api key multiple times in order to get it to work and now when i pasted the api key somewhere it was containing the character 'ı' which is the Turkish variant of 'i'. I believe your key generation/copy logic is not working culture invariant and those keys do not work.

### To Reproduce
1. Use a PC that is using Turkish culture.
2. Go to 'https://www.nuget.org/account/apikeys'
3. Generate a new Api Key.
4. Copy/Paste somewhere the generated key and see if it contains the character 'ı'. If not, generate a new key until you receive one key with 'ı'.
5. Try the generated key and see that it is not working.

### Expected behavior
Generated api keys should work irrespective of client culture.

### Screenshots
None

### Additional context
None

ryuyu:
Thanks for this report!
We were able to confirm that this is an issue and are working on a fix right now.

In the mean time, I was able to work around this (at least in Chrome) by moving the language preference to have English at the top. You can try this to see if it mitigates the issue for you until we can get a fix deployed.

1. Go to chrome://settings/languages
2. Expand Language section
3. Click the triple dot next to "English" and select "Move to the Top"
![image](https://user-images.githubusercontent.com/11051729/80258360-eade9380-8637-11ea-8a85-b3a29a1f5e9d.png)
4. Reload the page

JnRMnT:
Thanks for the prompt investigation and fix! I hope this gets to live environment soon :)

joelverhagen:
Hey @JnRMnT, we've go the fix merged into our DEV environment but we won't be able to deploy this change for at least a week. There is a deployment lock-down for our service in order to reduce risk of outage for those who are responding to the COVID-19 pandemic. We will be deploying this change as soon as this lock-down is over.

PR #7967, opened 2020-04-24
Fix locale sensitive API key copy

ryuyu:
Addressing https://github.com/NuGet/NuGetGallery/issues/7961

Issue was being caused by local sensitive lower casing.

### DIFF
--- a/src/NuGetGallery.Services/Authentication/ApiKeyV4.cs
+++ b/src/NuGetGallery.Services/Authentication/ApiKeyV4.cs
@@ -143,7 +143,7 @@ namespace NuGetGallery.Infrastructure.Authentication
 
         private string Normalize(string input)
         {
-            return input.ToLower();
+            return input.ToLowerInvariant();
         }
     }
 }


skofman1:
UTs?

skofman1:
Does this mean we have Turkish I API keys in the DB? I'm confused..

ryuyu:
No no.
The API keys in DB have an English lowercase 'i'
The issue is that when the key is displayed for copying, the "I" used to generate the key is then ToLower'ed, making it locale sensitive.
When the key is then used, a Turkish 'i' is submitted and the comparison fails since 'ı' is not 'i'

This fixes the display so users are actually copying the right key.

ryuyu:
Turns out that I was wrong and the Turkish i is actually making it in to the DB
Verification is also perfectly happy verifying these keys.
The issue appears to be further upstream. We should still go forward with this fix as it will prevent generation of keys with Turkish 'i' in the first place.

PR #7971, opened 2020-04-28
Add support for alternate instrumentation system in browser 

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/3020

We will use this on nuget.org to swap out the instrumentation system with the Microsoft internal one that allows sending data to both Application Insights and to the internal endpoint which uses the "tenant key" configuration that I added.

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-home.js
+++ b/src/NuGetGallery/Scripts/gallery/page-home.js
@@ -49,7 +49,7 @@ $(function () {
                 data: obj,
                 success: function (data) {
                     if (data.success) {
-                        emitAIMetric("Enable2FAModalProvidedFeedback");
+                        emitMetric("Enable2FAModalProvidedFeedback");
                         viewModel.dismissModalOrGetFeedback(false);
                     } else {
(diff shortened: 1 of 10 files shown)


src/NuGetGallery/Scripts/gallery/page-home.js
```diff
@@ -116,8 +116,8 @@ $(function () {
         });
     }
 
-    function emitAIMetric(metricName) {
-        window.nuget.sendAiMetric(metricName, 1, {});
+    function emitMetric(metricName) {
```
loic-sharma, line 119: Was renaming this function intentional? There's several references to `emitAIMetric` in this file

joelverhagen: Yes, intentional. Good catch.

zhhyu:
I am looking into it now, but it has been merged...

joelverhagen:
> I am looking into it now, but it has been merged...

Apologies. Feel free to leave comments and I'll address them in a subsequent PR.

PR #7979, opened 2020-05-05
Packages with duplicated entries in zip file should be blocked on upload.

dannyjdev:
A new validation was added at the ValidateBeforeGeneratePackageAsync method.

#7922


src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -118,6 +118,13 @@ public class PackageUploadService : IPackageUploadService
                 return result;
             }
 
+            result = CheckPackageDuplicatedEntries(nuGetPackage);
```
joelverhagen, line 121: Spoke offline. Let's get this validation for .snupkg files as well.

PR #7982, opened 2020-05-07
Update max supported nuget client version

loic-sharma:
Brings parity with: https://github.com/NuGet/NuGet.Services.Index/pull/166

Addresses https://github.com/nuget/nugetgallery/issues/7984

### DIFF
--- a/src/NuGetGallery/GalleryConstants.cs
+++ b/src/NuGetGallery/GalleryConstants.cs
@@ -27,7 +27,7 @@ namespace NuGetGallery
         public const int GravatarCacheDurationSeconds = 300;
 
         public const int MaxEmailSubjectLength = 255;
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/GalleryConstants.cs
```diff
@@ -27,7 +27,7 @@ public static class GalleryConstants
         public const int GravatarCacheDurationSeconds = 300;
 
         public const int MaxEmailSubjectLength = 255;
-        internal static readonly NuGetVersion MaxSupportedMinClientVersion = new NuGetVersion("5.5.0.0");
+        internal static readonly NuGetVersion MaxSupportedMinClientVersion = new NuGetVersion("5.6.0.0");
```
joelverhagen, line 30: Seems like this would be very easy to determine from:
https://dist.nuget.org/tools.json

PR #7985, opened 2020-05-08
ListPackages (default packages action) now responds to delete requests with a 405

drewgillies:
Addresses https://github.com/nuget/engineering/issues/2866

The default action for packages is method agnostic and can be scoped to Get to fix this.

Now returning a 405 for Http deletes.


loic-sharma:
> Note that this will cause a 404 to be returned ...

What does the nuget client output if the `nuget delete` gets a 404 response?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1061,6 +1061,7 @@ public virtual async Task<ActionResult> License(string id, string version)
             return View(model);
         }
 
+        [HttpGet]
```
loic-sharma, line 1064: Can this be tested?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1061,6 +1061,10 @@ public virtual async Task<ActionResult> License(string id, string version)
             return View(model);
         }
 
+        [HttpDelete]
+        public HttpStatusCodeResult ListPackages() => new HttpStatusCodeResult(HttpStatusCode.MethodNotAllowed);
```
shishirx34, line 1065: not super critical, but consider adding `Allow` headers if we are being specific about 405 implementation.

```
Response.Headers.Add("Allow", "GET");
```

Also, unit test for this would be a good idea to familiarize with testing logic.

shishirx34:
@drewgillies - The CI seems to be failing for unit tests on this change... You can click on the NuGetGallery CI Details link to see more details.

```
##[error]    NuGetGallery.Controllers.ControllerTests.AllActionsHaveAntiForgeryTokenIfNotGet [FAIL]
```

src/NuGetGallery/Controllers/PackagesController.cs
drewgillies: Done

src/NuGetGallery/Controllers/PackagesController.cs
drewgillies: I'm running nuget.exe tests on this.

drewgillies:
@loic-sharma I'm now returning a 405 for http deletes. The CLI indicates a 405 like this:
`Response status code does not indicate success: 405 (Method Not Allowed).`

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -789,6 +790,11 @@ private static async Task<byte[]> ReadPackageFile(PackageArchiveReader packageAr
             }
         }
 
+        [HttpDelete]
```
skofman1, line 793: Please add a comment with a link to the issue so we know why we need this special method.

drewgillies: Done

tests/NuGetGallery.Facts/TestUtils/ResultAssert.cs
```diff
@@ -127,6 +128,19 @@ public static HttpStatusCodeResult IsStatusCode(ActionResult result, int statusC
             return statusCodeResult;
         }
 
+        public static HttpStatusCodeWithHeadersResult IsStatusCodeWithHeaders(ActionResult result, HttpStatusCode statusCode, NameValueCollection headers)
+        {
+            var statusCodeResult = Assert.IsAssignableFrom<HttpStatusCodeWithHeadersResult>(result);
+            Assert.Equal((int)statusCode, statusCodeResult.StatusCode);
+
+            foreach(var key in headers.AllKeys)
```
loic-sharma, line 136: Style nitpick:

```suggestion
            foreach (var key in headers.AllKeys)
```

PR #7993, opened 2020-05-14
Append FAQ address to AAD unmanaged tenant authentication error

drewgillies:
Appending a FAQ link to the AAD authentication error when user account is in an unmanaged tenant state. Also added aka.ms to support this.

Addresses https://github.com/NuGet/NuGetGallery/issues/7544

Resultant error message on page (this change adds the last sentence):

image

### DIFF
--- a/src/NuGetGallery/Controllers/AuthenticationController.cs
+++ b/src/NuGetGallery/Controllers/AuthenticationController.cs
@@ -811,9 +811,20 @@ namespace NuGetGallery
 
         private ActionResult AuthenticationFailureOrExternalLinkExpired(string errorMessage = null)
         {
-            // User got here without an external login cookie (or an expired one)
-            // Send them to the logon action with a message
-            TempData["ErrorMessage"] = string.IsNullOrEmpty(errorMessage) ? Strings.ExternalAccountLinkExpired : errorMessage;
+            // We need a special case here because of https://github.com/NuGet/NuGetGallery/issues/7544. An unmanaged tenant scenario
+            // needs the FAQ URI appended to the AAD error, and we do that here so it appears in the header.
+            if ((errorMessage?.IndexOf("AADSTS65005", StringComparison.OrdinalIgnoreCase) ?? -1) > -1)
+            {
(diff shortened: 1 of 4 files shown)


src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -811,9 +811,20 @@ private string GetEmailAddressFromExternalLoginResult(AuthenticateExternalLoginR
 
         private ActionResult AuthenticationFailureOrExternalLinkExpired(string errorMessage = null)
         {
-            // User got here without an external login cookie (or an expired one)
-            // Send them to the logon action with a message
-            TempData["ErrorMessage"] = string.IsNullOrEmpty(errorMessage) ? Strings.ExternalAccountLinkExpired : errorMessage;
+            // We need a special case here because of https://github.com/NuGet/NuGetGallery/issues/7544. An unmanaged tenant scenario
+            // needs the FAQ URI appended to the AAD error, and we do that here so it appears in the header.
+            if ((errorMessage?.IndexOf("AADSTS65005", StringComparison.OrdinalIgnoreCase) ?? -1) > -1)
```
shishirx34, line 816: It seems like the error code `AADSTS65005` is for `MisconfiguredApplication`. It could possibly be likely that we might see the same error code for something other than `unmanaged state` issue i.e with a different error description. Pointing the user in such a case to the FAQ would be confusing. I recommend, also checking if `unmanaged` substring is present in the error message as well along with the error code.

See [error codes](https://docs.microsoft.com/en-us/azure/active-directory/develop/reference-aadsts-error-codes#aadsts-error-codes) and [misconfigured application](https://docs.microsoft.com/en-us/azure/active-directory/manage-apps/application-sign-in-problem-federated-sso-gallery?%2F%3FWT.mc_id=DMC_AAD_Manage_Apps_Troubleshooting_Nav#misconfigured-application) for more details.

drewgillies: Terrific--thanks. Added.

PR #7994, opened 2020-05-14
Update versions in readme

drewgillies:


README.md
```diff
@@ -9,11 +9,11 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 ## Build and Run the Gallery in (arbitrary number) easy steps
 
 1. Prerequisites:
-    1. Visual Studio 2017 - Install the following `Workloads`:
+    1. Visual Studio 2019 - Install the following `Workloads`:
```
joelverhagen, line 12: Are these all hard requirements now? Or would "2017 or newer" suffice?

loic-sharma: I believe we need VS 2017 (or is it 2015?) to run the functional tests. That might be worth calling out somewhere.

drewgillies: Ah, that's interesting. Perhaps this isn't a change worth doing. I'll leave well enough alone for now. Thanks!

PR #7998, opened 2020-05-19
Suppress ValidateAntiForgeryToken message in FxCop 

drewgillies:
nuget.exe doesn't provide an anti-forgery token, so we require an endpoint without one here, and should suppress the FxCop message.

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


drewgillies:
Note--with a ValidateAntiForgeryToken attribute, nuget.exe does this:
image

Without (which is what we need, hence FxCop message suppression):
image

joelverhagen:
Have you verified with an official build? You can queue one here: https://devdiv.visualstudio.com/DevDiv/_build?definitionId=9868&_a=summary. Make sure to set the NuGetGallerySubmoduleBranch build variable to be the branch name of this PR. This will check if the FX cop issue is still happening.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -799,6 +800,7 @@ private static async Task<byte[]> ReadPackageFile(PackageArchiveReader packageAr
 
         // This additional delete action addresses issue https://github.com/NuGet/Engineering/issues/2866 - we need to error out.
         [HttpDelete]
+        [SuppressMessage("Microsoft.Security", "CA3147:MarkVerbHandlersWithValidateAntiForgeryToken", Justification = "nuget.exe will not proviude a token")]
```
loic-sharma, line 803:
```suggestion
        [SuppressMessage("Microsoft.Security", "CA3147:MarkVerbHandlersWithValidateAntiForgeryToken", Justification = "nuget.exe will not provide a token")]
```

PR #8003, opened 2020-05-20
[A11y]Fix input group read out from table to nothing.

ryuyu:
Updates some css so the input row on upload page doesn't get read out as a table.

### DIFF
--- a/src/Bootstrap/dist/css/bootstrap.css
+++ b/src/Bootstrap/dist/css/bootstrap.css
@@ -3075,7 +3075,7 @@ tbody.collapse.in {
 }
 .input-group {
   position: relative;
-  display: table;
+  display: table-row;
   border-collapse: separate;
 }
 .input-group[class*="col-"] {
--- a/src/Bootstrap/less/input-groups.less
+++ b/src/Bootstrap/less/input-groups.less
@@ -8,7 +8,7 @@
 // -------------------------
 .input-group {
   position: relative; // For dropdowns
-  display: table;
+  display: table-row;
   border-collapse: separate; // prevent input groups from inheriting border styles from table cells when placed within a table
 
   // Undo padding and float of grid classes


joelverhagen:
Which forms does this affect?

agr:
That input row:
https://github.com/NuGet/NuGetGallery/blob/765917957b830837af97818e24a7a7be78f440ec/src/NuGetGallery/Views/Packages/UploadPackage.cshtml#L37-L44
? Why does it have anything to do with tables in the first place (I mean, why `display:block` is not good enough there)?

ryuyu:
Display:block makes it display like this:
image

Also note that this input-group is used in the search box as well (display block does this:)
image

Use Table-row was the only way I found to get it to display correctly, but not have narrator see a table there.

dannyjdev:
> Display:block makes it display like this:
> image
> 
> Also note that this input-group is used in the search box as well (display block does this:)
> image
> 
> Use Table-row was the only way I found to get it to display correctly, but not have narrator see a table there.

block gets all the possible width and basically pushes all other elements, I think it could work with inline-block, did you try it to see if it has the same results?

ryuyu:
Inline-block does this:
image

agr:
Making the div.input-group inline is not what we want. However, all the elements nested into it are already inline by default. I have a feeling we are fighting the default vertical form layout of bootstrap. How about adding a [`form-inline`](https://getbootstrap.com/docs/4.0/components/forms/#inline-forms) class to the `<form>` element:
https://github.com/NuGet/NuGetGallery/blob/765917957b830837af97818e24a7a7be78f440ec/src/NuGetGallery/Views/Packages/UploadPackage.cshtml#L34
?

ryuyu:
Sadly, it seems to affect the search box if we use Andrei's suggestion.

PR #8012, opened 2020-05-27
Banner about the upcoming TLS 1.2 enforcement

agr:
image

Added variable with the background color and put it in all places the verbatim value was used.


skofman1:
https://devblogs.microsoft.com/nuget/nuget-org-will-permanently-remove-support-for-tls-1-0-and-1-1-on-june-15th/

Let's align with the blog: "NuGet.org will permanently remove support for TLS 1.0 and 1.1 on June 15th. Please ensure that your systems use TLS 1.2."

agr:
@skofman1, done.

PR #8015, opened 2020-05-29
Fix for statistics not refreshing

agr:
Addresses https://github.com/NuGet/Engineering/issues/3201.


src/NuGetGallery/Services/JsonStatisticsService.cs
```diff
@@ -143,7 +145,7 @@ private bool ShouldRefresh()
                 return true;
             }
 
-            return (_lastRefresh - DateTime.UtcNow) >= _refreshInterval;
+            return (_systemTime.UtcNow.UtcDateTime - _lastRefresh) >= _refreshInterval;
```
agr, line 148: The meat of the fix.

src/NuGetGallery/Services/JsonStatisticsService.cs
```diff
@@ -31,6 +31,7 @@ public class JsonStatisticsService : IStatisticsService
         /// The service used to load reports in the form of JSON blobs.
         /// </summary>
         private readonly IReportService _reportService;
+        private readonly ISystemTime _systemTime;
```
zhhyu, line 34: Nit: comment this as others?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -441,6 +441,10 @@ protected override void Load(ContainerBuilder builder)
                 .As<IPackageVulnerabilityService>()
                 .InstancePerLifetimeScope();
 
+            builder.RegisterType<SystemTime>()
+                .As<ISystemTime>()
+                .InstancePerLifetimeScope();
```
zhhyu, line 446: I am not very sure whether it's a good choice to make it a singleton and introduce some risks here actually.
Is there any consideration as an instance scope here?

src/NuGetGallery/Services/JsonStatisticsService.cs
skofman1: how did it ever work?!

---
In reply to: [432213107](https://github.com/NuGet/NuGetGallery/pull/8015#discussion_r432213107) [](ancestors = 432213107)

src/NuGetGallery.Core/ISystemTime.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGetGallery
+{
+    /// <summary>
+    /// Provides a mockable way to access system time related functionality
+    /// </summary>
+    public interface ISystemTime
```
skofman1, line 11: we already have IDateTimeProvider. Can we reuse/extend it?

src/NuGetGallery/Services/JsonStatisticsService.cs
agr: @loic-sharma suggests that cloud service "works" because it restarts once a day, so it is forced to refresh those after every restart.

skofman1: 🤦

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
agr: I prefer not to make objects singletons unless it is a requirement (for example it is a shared cache of some sort). Making objects singletons by default imposes certain requirements (thread safety, for example) that we might not need and it might be in the way of enhancing the object in the future. The fact that the current implementation satisfies those requirements, should not get it "promoted" to singletons automatically.
Anyway, I got that class removed as we already have the same functionality elsewhere.

PR #8017, opened 2020-06-01
Added UI to reverse dependencies

dcc7497:
* Implemented the Used By UI in the package details page, includes the reverse dependencies

Progress on https://github.com/NuGet/Engineering/issues/3143

Before:
Before(Closed)

Before(Open)

After:

After(Closed)

After(Open)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -538,62 +538,113 @@
...
+                                        </tr>
+                                    }
+                                </tbody>
+                            </table>
+                        }
+
+                        else
+                        {
+                            <p>
+                                This package has no dependencies.
```
loic-sharma, line 595: Is this the right message? Shouldn't this be something like `This package has no dependents.`?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -538,62 +538,113 @@
...
-                    </div>
-                }
-                else
-                {
-                    <p class="github-usage panel-collapse collapse" aria-expanded="true" id="github-usage">
-                        This package is not used by any popular GitHub repositories.
-                    </p>
-                }
+                <div class="used-by panel-collapse collapse" aria-expanded="true" id="used-by">
+                    @if (Model.IsPackageDependentsEnabled && Model.PackageDependents != null)
```
joelverhagen, line 552: Why does this have the null check but the `GitHubDependenciesInformation` not have the null check? Seems like we should take a consistent approach here.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -538,62 +538,113 @@
...
+                                        <th></th>
+                                        <th class="used-by-adjust-table-head">Downloads</th>
+                                    </tr>
+                                </thead>
+                                <tbody class="no-border">
+                                    @foreach (var item in Model.PackageDependents.TopPackages)
+                                    {
+                                        <tr>
+                                            <td>
+                                                <a class="text-left ngp-link" href="@Url.Package(item.Id)" target="_blank">
```
joelverhagen, line 575: I don't think we want `target="_blank"` here. It should open in the same tab just like dependencies

loic-sharma: Yup, the pattern is to open in a new tab only if it's a link to different website. It's ok for the dependents section to not be consistent with the GitHub usage section here

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -538,62 +538,113 @@
...
-                else
-                {
-                    <p class="github-usage panel-collapse collapse" aria-expanded="true" id="github-usage">
-                        This package is not used by any popular GitHub repositories.
-                    </p>
-                }
+                <div class="used-by panel-collapse collapse" aria-expanded="true" id="used-by">
+                    @if (Model.IsPackageDependentsEnabled && Model.PackageDependents != null)
+                    {
+                        <h3 class="used-by-w400">
```
loic-sharma, line 554: Can you apply this styling to any `h3` element under the `.used-by` class and get rid of the `used-by-w400` class?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -538,62 +538,113 @@
                 }
             }
 
-            @if(Model.IsGitHubUsageEnabled && !Model.IsDotnetToolPackageType)
+            @if (!Model.IsDotnetToolPackageType && (Model.IsGitHubUsageEnabled || (Model.IsPackageDependentsEnabled && Model.PackageDependents != null)))
```
joelverhagen, line 541: When would `IsPackageDependentsEnabled` be true and `PackageDependents` be null at the same time? Is this possible? Could the check be simplified to just looking at one factor or the other?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
joelverhagen: Popular is not a qualifier here (there is not 100+ stars analog). But otherwise, I agree. @chgill-MSFT  can help with the text.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -538,62 +538,113 @@
...
+                        if (Model.PackageDependents.TotalPackageCount > 0)
+                        {
+                            <p>
+                                Showing the top @(Model.PackageDependents.TopPackages.Count) NuGet packages that depend on @(Model.Id):
+                            </p>
+                            <table class="table borderless">
+                                <thead>
+                                    <tr>
+                                        <th class="used-by-adjust-table-head">Package</th>
+                                        <th></th>
```
joelverhagen, line 566: What's the point of this empty column?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -538,62 +538,113 @@
...
-                        </p>
-                    </div>
-                }
-                else
-                {
-                    <p class="github-usage panel-collapse collapse" aria-expanded="true" id="github-usage">
-                        This package is not used by any popular GitHub repositories.
-                    </p>
-                }
+                <div class="used-by panel-collapse collapse" aria-expanded="true" id="used-by">
```
joelverhagen, line 551: /cc @ryuyu - should `aria-expanded` be here? And should it be true when the section is collapsed by default?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -538,62 +538,113 @@
...
+                                            <td></td>
+                                            <td>
+                                                <i class="ms-Icon ms-Icon--Download used-by-download-icon" aria-hidden="true"></i> <label class="used-by-count">@(item.DownloadCount.ToKiloFormat())</label>
+                                            </td>
+                                        </tr>
+                                    }
+                                </tbody>
+                            </table>
+                        }
+
```
joelverhagen, line 591: nit no empty line

joelverhagen:
Please try this UI on a varying window sizes (you can incrementally make the window smaller and smaller from full screen to see the transitions.

src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -161,6 +161,7 @@ public PackageDependents GetPackageDependents(string id)
 
         private IReadOnlyCollection<PackageDependent> GetListOfDependents(string id)
         {
+            int packagesDisplayed = 5;
```
zhhyu, line 164: It's better to make it a private const field under this class.

zhhyu:
Is that "1.8K+" or "24.5K+" supposed to be added here?
What does that "1.8K+" mean for GitHub usage? Maybe more than 1800 GitHub repositories use this package? As I remember, we only get the repositories with more stars. I feel that there should be many more repositories which depend on "Newtonsoft.Json". @loic-sharma @chgill-MSFT

chgill-MSFT:
@joelverhagen the single subheading under "Used By" during the AB test is fine with me 👌

src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -169,9 +170,9 @@ private IReadOnlyCollection<PackageDependent> GetListOfDependents(string id)
                                 group 1 by new { pr.Id, pr.DownloadCount, p.Description } into ng
                                 orderby ng.Key.DownloadCount descending
                                 select new { ng.Key.Id, ng.Key.DownloadCount, ng.Key.Description }
```
agr, line 172: You probably would be able to get `PackageDependent` objects directly from this query:
```
select new PackageDependent { Description = ng.Key.Id, DownloadCount = ng.key.DownloadCount, Id = ng.Key.Id }
```
then you wouldn't need an additional copy to another list.

Issue #7779, opened 2019-12-17
Some tests fails due to culture and line endings


volanavlad:
### To Reproduce
1. Using this git commit of NuGetGallery '705020bef78e677b41b3303d2eb02c79e165a652'
2. Run tests when current culture is Russian (i.e. ru-...)
3. See:
NuGetGallery.Security.MicrosoftTeamSubscriptionFacts.Policies_ReturnsMicrosoftTeamSubscriptionPolicies:
 Assert.Equal() Failure
                                     ↓ (pos 958)
    Expected: ···r more information.\r\nPolicy violations: {0}"}
    Actual:   ···r more information.\nPolicy violations: {0}"}
                                     ↑ (pos 958)

NuGetGallery.Packaging.PackageMetadataFacts.ThrowsForEmptyAndNonEmptyDuplicatesWhenDuplicateMetadataElementsDetectedAndParsingIsNotStrict
 
    Assert.Equal() Failure
              ↓ (pos 0)
    Expected: An item with the same key has already bee···
    Actual:   Элемент с тем же ключом уже был добавлен.
              ↑ (pos 0)

NuGetGallery.Packaging.PackageMetadataFacts.ThrowsWhenInvalidMinClientVersion(strict: True)
    Assert.Equal() Failure
                                       ↓ (pos 38)
    Expected: ···id version string.\r\nParameter name: value
    Actual:   ···id version string.\r\nИмя параметра: value
                                       ↑ (pos 38)

and so on

### Expected behavior
Tests work OK

### Additional context
It would be better to have localized strings in test asserts where it causes test failures.

skofman1:
@agr analyzed the issue and recommended a fix:

I installed the Russian language pack in Windows and managed to reproduce the test failures due to exception text localization. All but two failures are due to, in my opinion, a bit too strict Assert.Equals(expectedMessage, ex.Message), which can be replaced with Assert.StartsWith(expectedStart, ex.Message) which would not include the new line characters and the "Parameter name" tail of the message (which gets localized).

Out of the remaining two, ThrowsForEmptyAndNonEmptyDuplicatesWhenDuplicateMetadataElementsDetectedAndParsingIsNotStrict should be fixed by setting current thread's UI culture carefully, since if we are expecting certain exception text, we have to eliminate the possible variability. Alternatively, we can file an issue against corefx to introduce a special type of exception for duplicate keys in a dictionary, but I have a feeling that is unlikely to get addressed.

[...]

Issue #123, opened 2011-10-12
Delete Package view has some bad code/HTML transitions


half-ogre:
We have some code peeking through:

![screen capture](https://github.com/NuGet/IssueFiles/raw/master/ScreenCaptures/nugetgallery_123.png)

PR #8021, opened 2020-06-02
Fix unit test failed due to localized culture and line ending (#7779)

lyndaidaii:
Summary of the changes:

* Change to Assert.StartWith instead of Assert.Equal, avoid checking localized 'parameter'  
* Add new class called 'UseInvariantCultureAttribute', change thread UI culture before test, and reset 
   after test
* Addressed those unit test:  1.NuGetGallery.Packaging.PackageMetadataFacts.ThrowsForEmptyAndNonEmptyDuplicatesWhenDuplicateMetadataElementsDetectedAndParsingIsNotStrict
2. NuGetGallery.Packaging.PackageMetadataFacts.ThrowsWhenInvalidMinClientVersion
3.NuGetGallery.PackagesControllerFacts+TheUploadFileActionForPostRequests.WillRejectBrokenZipFiles  NuGetGallery.PackagesControllerFacts+TheUploadFileActionForPostRequests.WillShowViewWithErrorsIfPackageIdIsBreaksParsing
4.NuGetGallery.Authentication.AuthenticationServiceFacts+TheAuthenticateMethod.GivenPasswordCredential_ItThrowsArgumentException
5.NuGetGallery.Diagnostics.DiagnosticsServiceFacts+TheGetSourceMethod.RequiresNonNullOrEmptyName
6.NuGetGallery.Services.ContentServiceFacts+TheGetContentItemMethod.GivenANullOrEmptyName_ItShouldThrow
 

Addresses https://github.com/NuGet/NuGetGallery/issues/123

### DIFF
--- a/tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
+++ b/tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
@@ -3,12 +3,15 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
+using System.Threading;
 using NuGet.Packaging;
 using NuGet.Packaging.Core;
 using NuGet.Versioning;
+using NuGetGallery.TestUtils;
 using Xunit;
 
 namespace NuGetGallery.Packaging
@@ -215,7 +218,7 @@ namespace NuGetGallery.Packaging
 
             // Act & Assert
             var ex = Assert.Throws<ArgumentException>(() => PackageMetadata.FromNuspecReader(nuspec, strict));
-            Assert.Equal("'bad' is not a valid version string.\r\nParameter name: value", ex.Message);
+            Assert.StartsWith("'bad' is not a valid version string.", ex.Message);
         }
 
         [Fact]
@@ -252,17 +255,23 @@ namespace NuGetGallery.Packaging
         public void ThrowsForEmptyAndNonEmptyDuplicatesWhenDuplicateMetadataElementsDetectedAndParsingIsNotStrict()
         {
             // Arrange
+            var useInvariantCultureAttr = new UseInvariantCultureAttribute();
             var packageStream = CreateTestPackageStreamWithDuplicateEmptyAndNonEmptyMetadataElements();
             var nupkg = new PackageArchiveReader(packageStream, leaveStreamOpen: false);
             var nuspec = nupkg.GetNuspecReader();
+
+            useInvariantCultureAttr.Before(null);
 
             // Act & Assert
             var ex = Assert.Throws<ArgumentException>(() => PackageMetadata.FromNuspecReader(
                 nuspec,
                 strict: false));
+
+            useInvariantCultureAttr.After(null);
+
             Assert.Equal(
-                "An item with the same key has already been added.",
-                ex.Message);
+               "An item with the same key has already been added.",
+               ex.Message);
         }
 
         [Fact]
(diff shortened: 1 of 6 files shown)


tests/NuGetGallery.Core.Facts/TestUtils/UseInvariantCultureAttribute.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Globalization;
+using System.Reflection;
+using System.Threading;
+using Xunit.Sdk;
+
+namespace NuGetGallery.TestUtils
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public class UseInvariantCultureAttribute : BeforeAfterTestAttribute
+    {
```
cristinamanum, line 14: This blog uses a similar idea https://hamidmosalla.com/2018/08/30/xunit-beforeaftertestattribute-how-to-run-code-before-and-after-test. Lazy initialization I think is good choice.

tests/NuGetGallery.Core.Facts/Packaging/PackageMetadataFacts.cs
```diff
@@ -252,17 +255,23 @@ public void ThrowsPackagingExceptionWhenEmptyAndNonEmptyDuplicateMetadataElement
         public void ThrowsForEmptyAndNonEmptyDuplicatesWhenDuplicateMetadataElementsDetectedAndParsingIsNotStrict()
         {
             // Arrange
+            var useInvariantCultureAttr = new UseInvariantCultureAttribute();
             var packageStream = CreateTestPackageStreamWithDuplicateEmptyAndNonEmptyMetadataElements();
             var nupkg = new PackageArchiveReader(packageStream, leaveStreamOpen: false);
             var nuspec = nupkg.GetNuspecReader();
 
+            useInvariantCultureAttr.Before(null);
```
cristinamanum, line 263: Can use the attribute on the test method like here: https://hamidmosalla.com/2018/08/30/xunit-beforeaftertestattribute-how-to-run-code-before-and-after-test/ ?

agr: We definitely shouldn't instantiate the attribute class directly.

tests/NuGetGallery.Core.Facts/TestUtils/UseInvariantCultureAttribute.cs
agr: I'm not sure what benefits lazy initialization would bring in this particular case: we are not allocating anything at any point, just use static instance and keeping track of the original value.

lyndaidaii: I thought of those two ways Lazy initialization vs Static instance before, I think static instance works fine in this case.  I think Lazy initialization might be good if we are going to do more about culture in the future in terms of extensibility. Right now, I might prefer static instance? what do you guys think about those two approaches?

cristinamanum: I see, so use only InvariantCulture for now. It makes sense in this case to not have Lazy.

tests/NuGetGallery.Core.Facts/TestUtils/UseInvariantCultureAttribute.cs
```diff
@@ -0,0 +1,30 @@
...
+        private CultureInfo originalUICulture;
+
+        public CultureInfo OriginalUICulture { get; set; }
```
cristinamanum, line 17: I am not sure if this is used anywhere, If  not can it be removed?

PR #8043, opened 2020-06-12
Added configurable cache 

dcc7497:
Added files to be able to manually configure the cache time for the dependent packages in the package details page. Time can be configured using a json file.

Progress on: https://github.com/NuGet/NuGetGallery/issues/8038


src/NuGetGallery.Services/Configuration/CacheConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Services
+{
+    public class CacheConfiguration : ICacheConfiguration
+    {
+        public const int DefaultGetPackageDependentsCacheTimeInSeconds = 3600;
+
+        public int PackageDependentsCacheTimeInSeconds { get; set; } = DefaultGetPackageDependentsCacheTimeInSeconds;
```
joelverhagen, line 10: Consider changing the unit to minutes. seems like all the values we batted around were minute-specific not second-specific.

joelverhagen: Nevermind. That makes unit testing harder 😨

joelverhagen, line 10: If this value is zero or negative, what happens? Do things blow up?

src/NuGetGallery.Services/Configuration/CacheConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGetGallery.Services
+{
+    public class CacheConfiguration : ICacheConfiguration
+    {
+        public const int DefaultGetPackageDependentsCacheTimeInSeconds = 3600;
```
loic-sharma, line 8: Super nit: consider removing the "Get" verb here since it doesn't add too much value.

```suggestion
        public const int DefaultPackageDependentsCacheTimeInSeconds = 3600;
```

I don't feel strongly about this, feel free to keep it as-is if you prefer it :)

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1926,6 +1952,76 @@ public async Task CheckThatCacheKeyIsNotCaseSensitive()
                 Assert.Same(pd, _cache.Get(cacheKey));
             }
 
+            [Fact]
+            public async Task OneSecondCacheTime()
+            {
+                string id = "foo";
+                string cacheKey = "CacheDependents_" + id.ToLowerInvariant();
```
loic-sharma, line 1959: Please use `var`, especially when the type is obvious. I think we can update lines 1958, 1959, and 1964

zhhyu:
Please place your base branch from "master" to "dev" branch.
Unless it's a hotfix, we normally require that the change is firstly merged to "dev" branch.

src/NuGetGallery.Services/Configuration/CacheConfiguration.cs
dcc7497: Based off my own research and testing, maybe someone else can chime in, the method that uses this value returns a DateTime object type (line 963 of the PackagesController file). Nothing really crashes, but by giving it a negative number a cache is not "set" per say since you're giving a DateTime that has already passed so when we access the same page again, the cache has expired since the cache time was set to expire sometime in the past. While not ideal, setting the value to be a negative number would be the equivalent of not setting a cache time at all assuming we are running on normal time

joelverhagen: Thanks, this behavior for a weird case is acceptable.

PR #8052, opened 2020-06-17
[HOTFIX] Increase migration's database timeout

loic-sharma:
We added a migration to add an index on the `PackageDependencies` table. This migration takes longer than 30 seconds to run on PROD and is timing out. This change increases the timeout to 30 minutes which we hope is enough to create the new index.

⚠TODO: verify on DEV

Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3820287&view=results
Release: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=712036


zhhyu:
Maybe we want to make this a configured value later. But it should be good enough now.

loic-sharma:
@zhhyu Good idea, I created a new issue here: https://github.com/NuGet/NuGetGallery/issues/8053

Issue #8022, opened 2020-06-03
Detect disconnected client when package upload fails and return 4XX


joelverhagen:
### Bug description

When the client disconnects during package push (perhaps Ctrl + C on nuget.exe) we get an exception and return 500. Even though the client has disconnected, this gets recorded as a 500, affecting SLO/success metrics.

### Reproduction steps

1. Run latest gallery locally.
1. Push a very large package to your local gallery.
1. While gallery is reading the upload, terminate the upload on the client side.

### Expected behavior

A 400-level status code should be recorded since the client caused the error.

### Actual behavior

A `PackagePushFailure` metric is recorded and HTTP 500 is returned.

For recent instances in PROD (500 such requests in past 7 days) look for problem ID: `System.Runtime.InteropServices.COMException at NuGetGallery.ApiController.ReadPackageFromRequest`.

This 500 shows up in our failure reporting even though it was not a failure witnessed by the user and is not caused by our app or our dependencies.

### Additional context

It seems like we can detect this case with `Response.IsClientConnected`.
https://stackoverflow.com/a/12386676

PR #8060, opened 2020-06-18
Don't track the package push failure if the client has disconnected

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/8022

I was unable to reproduce the issue with many wild attempts (local devbox, app service directly, APIM, custom HTTP client implementation, raising firewall in the middle of request, killing connection via external tools, etc) but from reading online this seems like the recommended implementation. See https://stackoverflow.com/questions/4772597 for some recommendations. We can't use the cancellation token approach because the API that is throwing does not accept a token. It's this line that throws:
https://github.com/NuGet/NuGetGallery/blob/84ea5890b77e79b1f218521f2c031f8ee6c311f5/src/NuGetGallery/Controllers/ApiController.cs#L984

### DIFF
--- a/src/NuGetGallery/Strings.resx
+++ b/src/NuGetGallery/Strings.resx
@@ -1159,4 +1159,7 @@ The {1} Team</value>
   <data name="TwoFAFeedback_Error" xml:space="preserve">
     <value>There was an unexpected error when submitting feedback. Please contact NuGet support.</value>
   </data>
+  <data name="PackageUploadCancelled" xml:space="preserve">
+    <value>The package upload failed due to the client disconnecting.</value>
+  </data>
 </root>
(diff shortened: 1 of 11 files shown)


src/NuGetGallery/Strings.resx
```diff
@@ -1159,4 +1159,7 @@ The {1} Team</value>
   <data name="TwoFAFeedback_Error" xml:space="preserve">
     <value>There was an unexpected error when submitting feedback. Please contact NuGet support.</value>
   </data>
+  <data name="PackageUploadCancelled" xml:space="preserve">
```
loic-sharma, line 1162: Yay you also spelled "cancelled" the one true way 🥳

joelverhagen: `TaskCanceledException` 😡

PR #8061, opened 2020-06-18
Update `websocket-extensions` dependency

loic-sharma:
Bump `websocket-extensions` to the latest version. This is pulled in transitively by the `grunt-contrib-watch` development dependency. I verified this change by ensuring `grunt` still ran fine.

Addresses https://github.com/NuGet/NuGetGallery/network/alert/src/Bootstrap/package-lock.json/websocket-extensions/open
Addresses https://github.com/nuget/engineering/issues/3231

### DIFF
--- a/src/Bootstrap/package-lock.json
+++ b/src/Bootstrap/package-lock.json
@@ -2334,9 +2334,9 @@
       }
     },
     "websocket-extensions": {
-      "version": "0.1.3",
-      "resolved": "https://registry.npmjs.org/websocket-extensions/-/websocket-extensions-0.1.3.tgz",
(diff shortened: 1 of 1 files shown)


joelverhagen:
I think in the past we've lifted the dependency to the package.json so that it's more explicit and reproducible (e.g. if someone deletes the lock file 😓).

loic-sharma:
The approach taken is effectively what running `npm audit fix` would've done.

> I think in the past we've lifted the dependency to the package.json...

We don't have good processes around tracking what's a real direct dependency vs a transitive dependency that was lifted due to a vulnerability. If we lift the dependency, ideally, we'd later "unlift" it when our direct dependency `grunt-contrib-watch` is updated. We don't have good processes around doing this.

> ... e.g. if someone deletes the lock file 😓

I asked @zkat, they said:

> ... if you blow away your pkglock, it'll pull in the latest version anyway so that usually fixes vulns by default

I verified this behavior on my machine.

joelverhagen:
Sounds good, thanks for the context!

PR #8065, opened 2020-06-19
UI Polish

dcc7497:
* Removed "+" signs from total number of dependents.
* Removed number if there are 0 dependents 
* Added "popular" to text for GitHub dependents, i.e. "Showing the top 5 popular GitHub repositories that depend on Newtonsoft.Json:"

Before:
Before(Zero)
After:
After(zero)
Before:
Before(Plus and popular)
After:
After(Plus and popular)
Progress on: https://github.com/NuGet/NuGetGallery/issues/8063

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -554,7 +554,7 @@
                         if (Model.PackageDependents.TotalPackageCount > 0)
                         {
                             <h3>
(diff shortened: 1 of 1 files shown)


loic-sharma:
Copying my message from Teams chat just so it's public:

> Just curious, why are we removing the pluses?
​> I personally prefer them, even for the NuGet packages section. For example, Newtonsoft.Json may have 100 dependents on nuget.org, but there's likely more dependents on MyGet, Azure Artifacts, GitHub Registry, etc... The + is a slight indication to that
> Similarly for GitHub repositories, our repository count is more of a lowerbound estimate since we exclude GitHub repos with fewer than 100 stars

PR #8088, opened 2020-07-07
Replace permalink workaround with GitHub advisory permalink field

drewgillies:
Field is now available in GitHub GraphQL API, so using it now and removing extension method. We no longer need GhsaId so I'm removing it from our types.

Addresses https://github.com/NuGet/Engineering/issues/2890


tests/GitHubVulnerabilities2Db.Facts/AdvisoryIngestorFacts.cs
```diff
@@ -40,7 +40,7 @@ public async Task IngestsAdvisoryWithoutVulnerability(bool withdrawn)
                 var advisory = new SecurityAdvisory
                 {
                     DatabaseId = 1,
-                    GhsaId = "ghsa",
+                    Permalink = "GHSA-3456-abcd-7890",
```
loic-sharma, line 43: Should these permalink values be in a format like `https://github.com/advisories/GHSA-...`?

drewgillies: Yeah--I could change this for more correctness, I guess. The test is just for equality so I could have put `Foo` here. 👍

PR #8089, opened 2020-07-07
Add sorting and filtering in gallery  (without UI)

t-mog-msft:
This PR adds support for Sorting and Filtering (Advanced Search) on the Gallery. This PR is the first in a set of 2. This adds the backend backbone to communicate with the search service. The second PR will add the UI.


src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1111,6 +1112,8 @@ public virtual async Task<ActionResult> ListPackages(PackageListSearchViewModel
             var page = searchAndListModel.Page;
             var q = searchAndListModel.Q;
             var includePrerelease = searchAndListModel.Prerel ?? true;
+            var sortBy = searchAndListModel.SortBy;
+            var packageType = searchAndListModel.PackageType;
```
loic-sharma, line 1116: Please add the flight to ignore the sort by and package type filter. This should be a part of this change so that we can disable the feature and not block gallery deployments

src/NuGetGallery/GalleryConstants.cs
```diff
@@ -115,5 +115,26 @@ public static class FAQLinks
             public const string TransformToOrganization = "https://aka.ms/nuget-faq-transform-org";
             public const string AccountBelongsToUnmanagedTenant = "https://aka.ms/nuget-faq-unmanaged-tenant";
         }
+
+        public static class SearchSortNames
```
loic-sharma, line 119: I'd add a quick comment to explain these constants

src/NuGetGallery/GalleryConstants.cs
```diff
@@ -115,5 +115,26 @@ public static class FAQLinks
...
+            public const string TitleDesc = "title-desc";
+            public const string CreatedAsc = "created-asc";
+            public const string CreatedDesc = "created-desc";
+            public const string TotalDownloadsAsc = "totalDownloads-asc";
+            public const string TotalDownloadsDesc = "totalDownloads-desc";
+        }
+
+        public static class PackageTypeFilterNames
+        {
+            public const string AllTypes = "";
```
loic-sharma, line 134: What do you think of simplifying this name to just `All`? `PackageTypeFilterNames.AllTypes` feels a little redundant to me, but I don't feel strongly about this.

loic-sharma: Same comment for `PackageTypeFilter.AllTypes`

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -97,7 +97,9 @@
                 IncludePrerelease: '@Model.IncludePrerelease',
                 PageIndex: @Model.PageIndex,
                 TotalCount: @Model.TotalCount,
-                IsPreviewSearch: '@Model.IsPreviewSearch'
+                IsPreviewSearch: '@Model.IsPreviewSearch',
+                PackageType: '@Model.PackageType',
+                SortBy: '@Model.SortBy'
```
loic-sharma, line 102: The `PackageType` and `SortBy` values are user inputs. A user can craft a malicious `PackageType` that tries to escape the string parameter and inject javascript here. We should escape them like the search term:

```
SearchTerm: @Html.Raw(Json.Encode(Model.SearchTerm)),
```

Same comment for lines 125-126

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1111,12 +1112,21 @@ public virtual async Task<ActionResult> ListPackages(PackageListSearchViewModel
...
 
             if (page < 1)
             {
                 page = 1;
             }
 
+            // If advanced search is disabled, use the default experience
+            if (!(searchAndListModel.IsAdvancedSearchEnabled = _featureFlagService.IsAdvancedSearchEnabled(GetCurrentUser())))
```
loic-sharma, line 1124: This line is a bit busy. Consider splitting it into two lines: one to assign `IsAdvancedSearchEnabled` on the view model, and another to use the default experience if the flight is disabled.

src/NuGetGallery/OData/SearchService/SearchAdaptor.cs
```diff
@@ -26,28 +26,68 @@ public static class SearchAdaptor
...
                 Take = GalleryConstants.DefaultPackageListPageSize,
                 IncludePrerelease = includePrerelease,
                 SemVerLevel = semVerLevel
             };
 
+            switch (packageType)
+            {
+                case GalleryConstants.PackageTypeFilterNames.Dependency:
+                    searchFilter.PackageType = PackageTypeFilter.Dependency;
+                        break;
```
loic-sharma, line 46: Spacing looks off here

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -97,7 +97,9 @@
                 IncludePrerelease: '@Model.IncludePrerelease',
                 PageIndex: @Model.PageIndex,
                 TotalCount: @Model.TotalCount,
-                IsPreviewSearch: '@Model.IsPreviewSearch'
+                IsPreviewSearch: '@Model.IsPreviewSearch',
+                PackageType: '@Html.Raw(Json.Encode(Model.PackageType))',
+                SortBy: '@Html.Raw(Json.Encode(Model.SortBy))'
```
loic-sharma, line 102: Are these supposed to be wrapped in quotes? Line 96 for the `SearchTerm` isn't wrapped in quotes:

```SearchTerm: @Html.Raw(Json.Encode(Model.SearchTerm)),```

Comment (loic-sharma) [reply src/NuGetGallery/Views/Shared/ListPackages.cshtml:102]: Same comment for lines 125-126

File: tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs [hunk 8]
```diff
@@ -4842,6 +4842,49 @@ public async Task UsesPreviewSearchServiceForEmptyQuery(string q)
...
+                    httpContext: httpContext,
+                    searchService: searchService);
+                controller.SetCurrentUser(TestUtility.FakeUser);
+
+                var result = await controller.ListPackages(new PackageListSearchViewModel { Q = string.Empty, SortBy = sortBy, PackageType = packageType});
+                if(expectCached)
+                {
+                    Assert.NotNull(_cache.Get("DefaultSearchResults"));
+                }else
+                {
```
loic-sharma, line 4881:
```suggestion
                }
                else
                {
```

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -4842,6 +4842,49 @@ public async Task UsesPreviewSearchServiceForEmptyQuery(string q)
...
+            [InlineData(GalleryConstants.SearchSortNames.Published, null, false)]
+            [InlineData(GalleryConstants.SearchSortNames.TitleAsc, null, false)]
+            [InlineData(GalleryConstants.SearchSortNames.TitleDesc, null, false)]
+            [InlineData(GalleryConstants.SearchSortNames.TotalDownloadsAsc, null, false)]
+            [InlineData(GalleryConstants.SearchSortNames.TotalDownloadsDesc, null, false)]
+            [InlineData(null, GalleryConstants.PackageTypeFilterNames.Dependency, false)]
+            [InlineData(null, GalleryConstants.PackageTypeFilterNames.DotNetTool, false)]
+            [InlineData(null, GalleryConstants.PackageTypeFilterNames.Template, false)]
+            [InlineData(GalleryConstants.SearchSortNames.CreatedAsc, GalleryConstants.PackageTypeFilterNames.Dependency, false)]
+            public async Task DoesNotCacheAdvancedSearch(string sortBy, string packageType, bool expectCached)
```
loic-sharma, line 4861: We should also check whether we cache if the flight is disabled but the user provides advanced search filters

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -4842,6 +4842,49 @@ public async Task UsesPreviewSearchServiceForEmptyQuery(string q)
...
+            {
+                var httpContext = new Mock<HttpContextBase>();
+                httpContext.Setup(c => c.Cache).Returns(_cache);
+
+                var searchService = new Mock<ISearchService>();
+                searchService.Setup(s => s.Search(It.IsAny<SearchFilter>())).Returns(
```
loic-sharma, line 4867: I believe there's a `ReturnsAsync` so that you don't need to wrap your result with `Task.FromResult(...)` :)

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -4842,6 +4842,49 @@ public async Task UsesPreviewSearchServiceForEmptyQuery(string q)
...
+                var result = await controller.ListPackages(new PackageListSearchViewModel { Q = string.Empty, SortBy = sortBy, PackageType = packageType});
+                if(expectCached)
+                {
+                    Assert.NotNull(_cache.Get("DefaultSearchResults"));
+                }else
+                {
+                    Assert.Null(_cache.Get("DefaultSearchResults"));
+                }
+
+                searchService.Verify(x => x.Search(It.IsAny<SearchFilter>()), Times.Once);
```
loic-sharma, line 4885: We should check that the properties on the search filter are passed (or not!) as expected.

src/NuGetGallery/Services/SearchFilter.cs
```diff
@@ -24,6 +24,8 @@ public class SearchFilter
 
         public string SemVerLevel { get; set; }
 
+        public PackageTypeFilter PackageType { get; set; }
+
```
loic-sharma, line 28: This also affects the `LuceneSearchService`, please add tests for that.

src/NuGetGallery/GalleryConstants.cs
```diff
@@ -115,5 +115,32 @@ public static class FAQLinks
...
+            public const string CreatedDesc = "created-desc";
+            public const string TotalDownloadsAsc = "totalDownloads-asc";
+            public const string TotalDownloadsDesc = "totalDownloads-desc";
+        }
+
+        /// <summary>
+        /// These filter names represent the possible options that the search service supports for filtering packages.
+        /// </summary>
+        public static class PackageTypeFilterNames
+        {
```
joelverhagen, line 139: Why have a specific string here? Per nuspec schema and search API shape it is a freeform string.

Seems like these options are just a view concern.

src/NuGetGallery/Infrastructure/Lucene/Models/SortOrder.cs
```diff
@@ -11,5 +11,7 @@ public enum SortOrder
         TitleDescending,
         CreatedAscending,
         CreatedDescending,
+        TotalDownloadsAscending,
+        TotalDownloadsDescending
```
joelverhagen, line 15: nit trailing comma

src/NuGetGallery/OData/SearchService/SearchAdaptor.cs
```diff
@@ -26,28 +26,68 @@ public static class SearchAdaptor
...
             {
                 SearchTerm = q,
-                Skip = (page - 1) * GalleryConstants.DefaultPackageListPageSize, // pages are 1-based. 
+                Skip = (page - 1) * GalleryConstants.DefaultPackageListPageSize, 
                 Take = GalleryConstants.DefaultPackageListPageSize,
                 IncludePrerelease = includePrerelease,
                 SemVerLevel = semVerLevel
             };
 
+            switch (packageType)
```
joelverhagen, line 42: Why have mapping at all?

For sort-by there is a schema (fixed number of columns you can sort by). For package type name, it is a freeform string (with some minor restrictions that can be enforced in the search service)

src/NuGetGallery/Views/Shared/ListPackages.cshtml
joelverhagen: `@Html.ToJson()` is used in more places

src/NuGetGallery/GalleryConstants.cs
t-mog-msft: Yeah, but we don't support user input as a freeform string in the UI currently, so those are the supported options. In the future, if we design an experience with an input text field with autocompletion for example, we may remove this

joelverhagen: These constants can simply live in the UI (cshtml). Having service layer code around this seems unnecessarily stringent.

joelverhagen: Adding and removing package types is really only a UI concern. As the code stands today, there are multiple places to update for add/remove.

t-mog-msft: Yeah but we also want to limit the input in the Controller, if the users give an input that isn't supported, it reverts to default values. To do that, we need access to those constants

joelverhagen: > Yeah but we also want to limit the input in the Controller,

Why? Isn't validation that it's an acceptable package type enough? You can use package ID validator for that, per package [type spec](https://docs.microsoft.com/en-us/nuget/create-packages/set-package-type)

> Custom type packages use an arbitrary type identifier that conforms to the same format rules as package IDs. Any type other than `Dependency` and `DotnetTool`, however, are not recognized by the NuGet Package Manager in Visual Studio.

Perhaps this is a misunderstanding from the spec. I remember someone mentioning that we would allow people to fiddle with the URL parameters as an undocumented "hack" to filter by any package type.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1134,8 +1144,19 @@ public virtual async Task<ActionResult> ListPackages(PackageListSearchViewModel
             var isPreviewSearchEnabled = _abTestService.IsPreviewSearchEnabled(GetCurrentUser());
             var searchService = isPreviewSearchEnabled ? _searchServiceFactory.GetPreviewService() : _searchServiceFactory.GetService();
 
+            if(!isSupportedSortBy(searchAndListModel.SortBy))
```
joelverhagen, line 1147: nit: extra space after `if`

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -2967,5 +2996,13 @@ private static HttpStatusCodeResult HttpForbidden()
         {
             return new HttpStatusCodeResult(HttpStatusCode.Forbidden, Strings.Unauthorized);
         }
+
+        private static bool isSupportedSortBy(string sortBy)
```
loic-sharma, line 3000: Nit: C# style is to uppercase all the things

```suggestion
        private static bool IsSupportedSortBy(string sortBy)
```

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1134,8 +1144,19 @@ public virtual async Task<ActionResult> ListPackages(PackageListSearchViewModel
             var isPreviewSearchEnabled = _abTestService.IsPreviewSearchEnabled(GetCurrentUser());
             var searchService = isPreviewSearchEnabled ? _searchServiceFactory.GetPreviewService() : _searchServiceFactory.GetService();
 
+            if(!isSupportedSortBy(searchAndListModel.SortBy))
+            {
+                searchAndListModel.SortBy = GalleryConstants.SearchSortNames.Relevance;
+            }
+
+            var isDefaultSortBy = string.Equals(searchAndListModel.SortBy, GalleryConstants.SearchSortNames.Relevance, StringComparison.OrdinalIgnoreCase);
```
loic-sharma, line 1152: The fact that `searchAndListModel.SortBy` is guaranteed to not be `null` at this point is subtle given that `IsSupportedSortBy` is a separate method. This code would break if one updates `IsSupportedSortBy` to return `true` if `SortBy` is `null`.

What do you think of simply handling the `null` case here? If you make this change, you'll need to update `IsSupportedSortBy` to return `true` if `SortBy` is `null`.

src/NuGetGallery/ViewModels/PackageListViewModel.cs
```diff
@@ -60,5 +64,13 @@ public class PackageListViewModel
         public bool IncludePrerelease { get; }
 
         public bool IsPreviewSearch { get; }
+
+        public string PackageType { get; set; }
+
+        public string SortBy { get; set; }
+
+        public bool shouldDisplayAdvancedSearchPanel { get; set; }
```
loic-sharma, line 72: Nit: C# style is to uppercase all the things

```suggestion
        public bool ShouldDisplayAdvancedSearchPanel { get; set; }
```

src/NuGetGallery/Infrastructure/Lucene/GallerySearchClient.cs
```diff
@@ -61,6 +64,7 @@ public async Task<ServiceResponse<JObject>> GetDiagnostics()
             nameValue.Add("q", query);
             nameValue.Add("skip", skip.ToString());
             nameValue.Add("take", take.ToString());
+            nameValue.Add("packageType", packageType);
```
loic-sharma, line 67: This will pass empty/null values to the search service. I think it'd be better to follow the `semVerLevel` model were we only append the parameter if the value is not null nor empty.

tests/NuGetGallery.Facts/Infrastructure/Lucene/GallerySearchServiceFacts.cs
```diff
@@ -64,15 +64,16 @@ public async Task MapsAllSortOrders(SortOrder sortOrder)
...
+        [InlineData("query", "projectTypeFilter", true, "dependency", SortOrder.TitleAscending, 1, 10, true, true, true, true, "supportedFramework", "semVerLevel", "q=query&skip=1&take=10&packageType=dependency&sortBy=title-asc&semVerLevel=semVerLevel&supportedFramework=supportedFramework&projectType=projectTypeFilter&prerelease=true&explanation=true&ignoreFilter=true&countOnly=true")]
+        [InlineData("query", "projectTypeFilter", true, "dependency", SortOrder.TitleDescending, 1, 10, true, true, true, true, "supportedFramework", "semVerLevel", "q=query&skip=1&take=10&packageType=dependency&sortBy=title-desc&semVerLevel=semVerLevel&supportedFramework=supportedFramework&projectType=projectTypeFilter&prerelease=true&explanation=true&ignoreFilter=true&countOnly=true")]
 
         public async Task SearchArgumentsAreCorrectSet(string query,
             string projectTypeFilter,
             bool includePrerelease,
+            string packageType,
```
loic-sharma, line 76: Let's add a test to check that the `packageType` query parameter is skipped if the value is empty.

src/NuGetGallery/Views/Shared/ListPackages.cshtml
loic-sharma: @t-mog-msft it looks like this was missed

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -4842,6 +4843,99 @@ public async Task UsesPreviewSearchServiceForEmptyQuery(string q)
...
+
+                var controller = CreateController(
+                    GetConfigurationService(),
+                    httpContext: httpContext,
+                    searchService: searchService,
+                    featureFlagService: featureFlagService);
+                controller.SetCurrentUser(TestUtility.FakeUser);
+
+                var result = await controller.ListPackages(new PackageListSearchViewModel { Q = string.Empty, SortBy = sortBy, PackageType = packageType});
+                if(expectCached)
```
loic-sharma, line 4890:
```suggestion
                if (expectCached)
```

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -1134,8 +1144,19 @@ public virtual async Task<ActionResult> ListPackages(PackageListSearchViewModel
...
+            var isDefaultPackageType = searchAndListModel.PackageType == null || !searchAndListModel.PackageType.Any();
```
loic-sharma, line 1153: You can simplify this as:

```suggestion
            var isDefaultPackageType = string.IsNullOrEmpty(searchAndListModel.PackageType);
```

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -4842,6 +4843,99 @@ public async Task UsesPreviewSearchServiceForEmptyQuery(string q)
...
+            [InlineData(GalleryConstants.SearchSortNames.TitleAsc, SortOrder.Relevance, null, "")]
+            [InlineData(GalleryConstants.SearchSortNames.TitleDesc, SortOrder.Relevance, null, "")]
+            [InlineData(GalleryConstants.SearchSortNames.TotalDownloadsAsc, SortOrder.Relevance, null, "")]
+            [InlineData(GalleryConstants.SearchSortNames.CreatedDesc, SortOrder.CreatedDescending, null, "")]
+            [InlineData(GalleryConstants.SearchSortNames.TotalDownloadsDesc, SortOrder.TotalDownloadsDescending, null, "")]
+            [InlineData(GalleryConstants.SearchSortNames.Relevance, SortOrder.Relevance, null, "")]
+            [InlineData(null, SortOrder.Relevance, "Dependency", "Dependency")]
+            [InlineData(null, SortOrder.Relevance, "DotNetTool", "DotNetTool")]
+            [InlineData(null, SortOrder.Relevance, "Template", "Template")]
+            public async Task RedirectsToDefaultWhenInvalidAdvancedSearch(string sortBy, SortOrder expectedSortBy, string packageType, string expectedPackageType)
```
loic-sharma, line 4916: I test that an unknown package type like `ThisIsNotARealPackageType` is still passed to the search service.

PR #8104, opened 2020-07-14
3268 add column EmbeddedReadmeFileType to Package tables

lyndaidaii:

* add one column of enum EmbeddedReadmeFileType to package table

raw sql would be like here: 

            Sql("ALTER TABLE [dbo].[Packages] ADD EmbeddedReadmeFileType INT NOT NULL  DEFAULT 0");

Addresses https://github.com/NuGet/NuGetGallery/issues/123


src/NuGetGallery/Migrations/202006011927336_AddIndexToPackageDependencies.cs
```diff
@@ -28,4 +28,4 @@ public override void Down()
             DropIndex(table: "PackageDependencies", name: "IX_PackageDependencies_Id");
         }
     }
-}
\ No newline at end of file
+}
```
zhhyu, line 31: Undo this change. It should be not included in this PR.

src/NuGet.Services.Entities/EmbeddedReadmeFileType.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
```
zhhyu, line 3: Leave 1 blank line.

zhhyu:
There should be a change on the Package entity. Is it?

joelverhagen:
Could you provide the full SQL for this migration? This can be gotten locally via `Update-Database ... -Script`.

joelverhagen:
I see that the column is not nullable. Does this mean that there is a default value set on the field which is... 0 = Absent for all existing rows? This is why I am asking for the raw SQL (`ALTER TABLE...`) since this will tell use the exact behavior from a SQL perspective.

lyndaidaii:
> I see that the column is not nullable. Does this mean that there is a default value set on the field which is... 0 = Absent for all existing rows? This is why I am asking for the raw SQL (`ALTER TABLE...`) since this will tell use the exact behavior from a SQL perspective.

@joelverhagen What's other intention to have raw sql?

joelverhagen:
> @joelverhagen What's other intention to have raw sql?  

To know exactly what's going to happen on our production database.

src/NuGet.Services.Entities/EmbeddedReadmeFileType.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace NuGet.Services.Entities
+{
+    /// <summary>
+    /// Specifies the type of the readme file used in the package
+    /// </summary>
+    public enum EmbeddedReadmeFileType
+    {
+        /// <summary>
+        /// Indicates that package has no embedded readme file.
```
loic-sharma, line 13:
```suggestion
        /// Indicates that the package has no embedded readme file.
```

src/NuGet.Services.Entities/EmbeddedReadmeFileType.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// </summary>
+        Absent = 0,
+
+        /// <summary>
+        /// Indicates that embedded readme file is markdown.
```
loic-sharma, line 18:
```suggestion
        /// Indicates that the embedded readme file is markdown.
```

src/NuGetGallery/Migrations/202007141855149_ReadmeChanges.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace NuGetGallery.Migrations
+{
+    using System.Data.Entity.Migrations;
```
zhhyu, line 3: I recommend to use "AddEmbeddedReadmeFileTypeColumn" as the migration file's name. The "ReadmeChanges" is too general.

src/NuGetGallery/Migrations/202007150051382_AddEmbeddedReadmeFileTypeColumn.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace NuGetGallery.Migrations
+{
+    using System.Data.Entity.Migrations;
+
+    public partial class ReadmeChanges : DbMigration
+    {
+        public override void Up()
+        {
+            Sql("ALTER TABLE [dbo].[Packages] ADD EmbeddedReadmeFileType INT NOT NULL CONSTRAINT DF_Doc_Exz_EmbeddedReadmeFileType DEFAULT 0");
```
zhhyu, line 9: Could you please share why and how you add this constraint with "DF_Doc_Exz_EmbeddedReadmeFileType"?

joelverhagen, line 9: Hey Lynn, I must have communicated poorly here. I was just wanting to see the raw SQL *as a comment on the PR*. Using the EF helper methods inside the migration (what you had before) is 100% okay and the proper implementation. Apologies for the extra work and misunderstanding.

PR #8105, opened 2020-07-14
Add sorting and filtering UI to Gallery

t-mog-msft:
This PR adds the UI for Advanced Search to the Gallery. It also changes a the "There are X packages" component to use Bootstrap

image

### Phone screen preview
image


src/Bootstrap/less/theme/page-list-packages.less
```diff
@@ -27,4 +27,82 @@
...
+  background-color: #f4f4f4;
+  border-color: #f4f4f4;
+  margin-right: 0px;
+  margin-left: 0;
+  margin-bottom: 15px;
+}
+
+.advanced-search-panel fieldset {
+    margin: 20px;
+}
```
loic-sharma, line 82: You can nest stuff in .less. For example:

```less
.advanced-search-panel {
  ...

  fieldset {
    ...
    label {
      ..
    }
  }
}
```

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -36,11 +90,71 @@
                 }
             </h1>
         </div>
-        <div class="cell-controls">
-            <a href="?q=@Model.SearchTerm@(Model.IncludePrerelease ? "&prerel=false" : "")">@(Model.IncludePrerelease ? "Hide" : "Show") prerelease</a>
+        <div class="col-md-2 col-xs-3 col-sm-1 text-right no-padding">
+            @if(Model.IsAdvancedSearchFlightEnabled)
```
loic-sharma, line 94:
```suggestion
            @ if(Model.IsAdvancedSearchFlightEnabled)
```

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -36,11 +90,71 @@
...
+                        @foreach(var packageType in UI_SUPPORTED_PACKAGETYPES)
+                        {
+                            @AddPackageFilterOption(packageType.Value, packageType.Key, canBeNull: packageType.Key == DEFAULT_PACKAGETYPE)    
+                        }
+                    </fieldset>
+                </div>
+                <div class="col-md-2">
+                    <fieldset id="advanced-search-sortby">
+                        <legend>Sort by</legend>
+                        @AddSortByOption("Relevance", GalleryConstants.SearchSortNames.Relevance, canBeNull: true)
```
loic-sharma, line 132: What do you think of renaming `canBeNull` to `isDefault`?

src/Bootstrap/less/theme/page-list-packages.less
t-mog-msft: Wait this is so nice!

src/NuGetGallery/Views/Shared/ListPackages.cshtml
t-mog-msft: I named it that way because this is the condition that uses that parameter
image

loic-sharma: The `canBeNull` name leaks implementation details that's not obvious from the `@AddBySortByOption` call. I think at a glance someone won't understand why one of the options has `canBeNull` but the others don't. On the other hand, the parameter `isDefault` makes the parameter's purpose a little more clear in my mind.

src/NuGetGallery/Views/Shared/ListPackages.cshtml
t-mog-msft: I changed it then reverted it, this can't have a space between the @ and the if or it doesn't compile

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -6,9 +6,63 @@
...
+}
+
+@helper AddSortByOption(string optionName, string optionValue, bool canBeNull = false)
+{
+    @AddRadioGroupOption("advanced-search-sortby", optionName, optionValue, Model.SortBy, canBeNull);
+}
+
+@helper AddRadioGroupOption(string radioGroupName, string optionName, string optionValue, string optionCompareValue, bool canBeNull = false)
+{
+    var elemGuid = Guid.NewGuid();
```
joelverhagen, line 21: Could you just use the `@optionName` for the `id` instead of some GUID?

t-mog-msft: optionName may have spaces and I can't chose the optionValue since one of them is en empty string

src/Bootstrap/less/theme/page-list-packages.less
```diff
@@ -27,4 +27,82 @@
...
+.btn-command {
+  margin-top: 33px;
+  margin-bottom: 33px;
+  width: 90px;
+  height: 45px;
+  background-color: transparent;
+  border: none;
+}
+
+@media (max-width: 992px) {
```
loic-sharma, line 41: You can use the `@screen-md` variable instead of `992px`: https://github.com/NuGet/NuGetGallery/blob/5dde8cb3c552c24a9165767d3a408131edbd96d1/src/Bootstrap/less/variables.less#L307

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -6,9 +6,63 @@
...
+<script>
+    function ApplyAdvancedSearch() {
+        // string
+        var packageType = document.querySelector('input[name="advanced-search-packagetype"]:checked').value;
+        // string
+        var sortBy = document.querySelector('input[name="advanced-search-sortby"]:checked').value;
+        // bool
+        var includePrerelease = document.getElementById('advanced-search-includeprerelease').checked;
+
+        var searchTermStr = `?q=${'@(Model.SearchTerm)'}`
```
joelverhagen, line 48: This should be build using some API that knows about query string encoding.

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -6,9 +6,63 @@
...
+        { "dependency", "Dependency" },
+        { "dotnettool", ".NET tool" },
+        { "template" , "Template" },
+    };
+}
+
+<script>
+    function ApplyAdvancedSearch() {
+        // string
+        var packageType = document.querySelector('input[name="advanced-search-packagetype"]:checked').value;
```
joelverhagen, line 42: Consider using jQuery, it's used in many other places in our code

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -6,9 +6,63 @@
...
+    {
+        { DEFAULT_PACKAGETYPE, "All types" },
+        { "dependency", "Dependency" },
+        { "dotnettool", ".NET tool" },
+        { "template" , "Template" },
+    };
+}
+
+<script>
+    function ApplyAdvancedSearch() {
```
joelverhagen, line 40: Can this be moved to an external JS file at the footer of the page, similar to our other pages? I think the amount of JS in this page is getting too big.

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -36,11 +90,71 @@
...
+                        <div style="display: flex;">
+                            <input id="advanced-search-includeprerelease" type="checkbox" @(Model.IncludePrerelease ? "checked=\"checked\"" : "")>
+                            <label for="advanced-search-includeprerelease">Include prerelease</label>
+                        </div>
+                    </fieldset>
+                </div>
+            </div>
+            <div class="row">
+                <div class="col-xs-4 col-sm-3 col-md-2 col-lg-1">
+                    <input class="btn btn-primary form-control" type="button" value="Apply" onclick="ApplyAdvancedSearch()">
```
joelverhagen, line 149: We can just use a form with `method=get` to do this. You don't have to worry about building the URL (browser does this for you). It's the same approach as the search box itself.

src/NuGetGallery/Views/Shared/ListPackages.cshtml
joelverhagen: We can default the empty string to "all". 

Also, you don't necessarily need the `for` if you put the `input` inside the `label`. That might make it even simpler.

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -6,9 +6,63 @@
...
+    var elemGuid = Guid.NewGuid();
+    <div style="display: flex;">
+        <input id="@elemGuid" type="radio" name="@radioGroupName" @(optionCompareValue == optionValue || (canBeNull && optionCompareValue == null) ? "checked=\"checked\"" : "") value="@optionValue">
+        <label for="@elemGuid">@optionName</label>
+    </div>
+}
+
+@functions{
+    private const string DEFAULT_PACKAGETYPE = "";
+    private static readonly Dictionary<string, string> UI_SUPPORTED_PACKAGETYPES = new Dictionary<string, string> (StringComparer.OrdinalIgnoreCase)
```
joelverhagen, line 30: nit `IReadOnlyDictionary`

joelverhagen:
Could you post a small screen size (phone-ish) screenshot?

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -36,11 +90,71 @@
                 }
             </h1>
         </div>
-        <div class="cell-controls">
-            <a href="?q=@Model.SearchTerm@(Model.IncludePrerelease ? "&prerel=false" : "")">@(Model.IncludePrerelease ? "Hide" : "Show") prerelease</a>
+        <div class="col-md-2 col-xs-3 col-sm-1 text-right no-padding">
```
joelverhagen, line 93: Can you work with @ryuyu to do an accessibility pass on your own to get any low handing fruit?

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -36,11 +90,71 @@
...
+            @ if(Model.IsAdvancedSearchFlightEnabled)
```
joelverhagen, line 94:
```suggestion
            @if (Model.IsAdvancedSearchFlightEnabled)
```

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -36,11 +90,71 @@
...
-
+    @if (Model.IsAdvancedSearchFlightEnabled)
+    {
+        <div class="row clearfix collapse advanced-search-panel@(Model.ShouldDisplayAdvancedSearchPanel ? " in" : "")" id="advancedSearchPanel">
+            @if(Model.PackageType != null && !UI_SUPPORTED_PACKAGETYPES.ContainsKey(Model.PackageType))
+            {
+                <div class="row warning-panel">
+                    <div class="col-md-12 icon-text alert alert-warning">
+                        <i class="ms-Icon ms-Icon--Info" aria-hidden="true"></i>
+                        Filtering on unrecognized package type: @Html.ToJson(Model.PackageType)
```
joelverhagen, line 115: wondering why we need `ToJson` here. We can just `"@Model.PackageType"` right?

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -36,11 +90,71 @@
...
+                        @AddSortByOption("Relevance", GalleryConstants.SearchSortNames.Relevance, canBeNull: true)
+                        @AddSortByOption("Downloads", GalleryConstants.SearchSortNames.TotalDownloadsDesc)
+                        @AddSortByOption("Recently updated", GalleryConstants.SearchSortNames.CreatedDesc)
+                    </fieldset>
+                </div>
+                <div class="col-md-2">
+                    <fieldset>
+                        <legend>Options</legend>
+                        <div style="display: flex;">
+                            <input id="advanced-search-includeprerelease" type="checkbox" @(Model.IncludePrerelease ? "checked=\"checked\"" : "")>
```
joelverhagen, line 141: doesn't `checked="@Model.IncludePrerelease"` work?

src/Bootstrap/less/theme/page-list-packages.less
loic-sharma: Same comment for line 53

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -36,11 +90,71 @@
...
+                        </div>
+                    </fieldset>
+                </div>
+            </div>
+            <div class="row">
+                <div class="col-xs-4 col-sm-3 col-md-2 col-lg-1">
+                    <input class="btn btn-primary form-control" type="button" value="Apply" onclick="ApplyAdvancedSearch()">
+                </div>
+                <div class="col-xs-4 col-sm-3 col-md-2 col-lg-1">
+                    <input class="btn btn-link form-control" type="button" value="Reset" onclick="ResetAdvancedSearch()">
```
joelverhagen, line 152: It's more standard in our codebase to associate JavaScript events via jQuery rather than `onclick`

src/NuGetGallery/Views/Shared/ListPackages.cshtml
loic-sharma: > Also, you don't necessarily need the `for` if you put the `input` inside the `label`. That might make it even simpler.

I would double-check the accessibility guidelines. /cc @ryuyu

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -6,9 +6,63 @@
...
+        var sortBy = document.querySelector('input[name="advanced-search-sortby"]:checked').value;
+        // bool
+        var includePrerelease = document.getElementById('advanced-search-includeprerelease').checked;
+
+        var searchTermStr = `?q=${'@(Model.SearchTerm)'}`
+        var includePreReleaseStr = "&prerel=" + includePrerelease;
+        var packageTypeStr = !packageType ? "" : `&packageType=${packageType}`;
+        var sortByStr = !sortBy ? "" : `&sortBy=${sortBy}`;
+
+        location.href = searchTermStr + includePreReleaseStr + packageTypeStr + sortByStr;
```
loic-sharma, line 53: Could it be possible to implement this using a form with `method="get"`?

src/NuGetGallery/Views/Shared/ListPackages.cshtml
loic-sharma, line 94:
```suggestion
            @if (Model.IsAdvancedSearchFlightEnabled)
```

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -36,11 +90,71 @@
...
+            @ if(Model.IsAdvancedSearchFlightEnabled)
+            {
+                <button class="btn-command" data-toggle="collapse" data-target="#advancedSearchPanel" aria-expanded="false" aria-controls="advancedSearchPanel">
+                    <i class="ms-Icon @(Model.ShouldDisplayAdvancedSearchPanel ? "ms-Icon--FilterSolid" : "ms-Icon--Filter") ms-fontColor-blue" aria-hidden="true"></i>
+                    Filter
+                </button>
+            }
+            else
+            {
+                <a class="prerel-filter" href="?q=@Model.SearchTerm@(Model.IncludePrerelease ? "&prerel=false" : "")">@(Model.IncludePrerelease ? "Hide" : "Show") prerelease</a>
```
loic-sharma, line 103: Hm I just realized that this URL generation is bad:

1. [Search for `test&foo=bar`](https://www.nuget.org/packages?q=test%26foo%3Dbar)
2. Press the `Hide prelease` button

You'll be navigated to `https://www.nuget.org/packages?q=test&foo=bar&prerel=false`. This is a bug on nuget.org today. We should example the `searchTerm` here.

src/NuGetGallery/Views/Shared/ListPackages.cshtml
loic-sharma: I dun goof'd my suggestion. I meant to say:

```suggestion
            @if (Model.IsAdvancedSearchFlightEnabled)
```

src/NuGetGallery/Views/Shared/ListPackages.cshtml
t-mog-msft: Yes, just a wrong copy/paste ^^

src/NuGetGallery/Views/Shared/ListPackages.cshtml
t-mog-msft: no, checked="false" will still check it

joelverhagen: If `IncludePrerelease` is a boolean, Razor acts properly with `checked`, i.e. it does not render the attribute at all if the property is false. Did you try it? I'd be surprised if it does not work.

src/Bootstrap/less/theme/page-list-packages.less
```diff
@@ -27,4 +27,82 @@
...
+    margin-top: 54px;
+  }
+}
+
+.btn-command:hover {
+  background-color: #f4f4f4;
+}
+
+.no-padding {
+    padding: 0 !important;
```
loic-sharma, line 65: Spacing seems inconsistent here. Same comment for line 69

src/Bootstrap/less/theme/page-list-packages.less
```diff
@@ -27,4 +27,82 @@
...
+}
+
+.no-margin {
+    margin: 0 !important;
+}
```
loic-sharma, line 70: Hey is there a reason why this stuff isn't in the `.page-list-packages` selector?

t-mog-msft: Yes, it's not used by page-list-packages. It is used by the "there are X packages" row only

loic-sharma: Selectors should be as specific as possible. Otherwise, you can easily clobber other styling if you use global selectors. If possible, anything that can be nested within the `.package-list-packages` selected should be.

It looks like that "there are X packages" row is in the [`.package-list-packages` div](https://github.com/NuGet/NuGetGallery/pull/8105/files#diff-0adfb38d49494f516a945df8aca64717R63):

image

If so, I'd recommend nesting the the `.no-margin` and `.no-padding` selectors within `.page-list-packages`.

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -6,9 +6,63 @@
...
+{
+    var elemGuid = Guid.NewGuid();
+    <div style="display: flex;">
+        <input id="@elemGuid" type="radio" name="@radioGroupName" @(optionCompareValue == optionValue || (canBeNull && optionCompareValue == null) ? "checked=\"checked\"" : "") value="@optionValue">
+        <label for="@elemGuid">@optionName</label>
+    </div>
+}
+
+@functions{
+    private const string DEFAULT_PACKAGETYPE = "";
```
loic-sharma, line 29: Nit: C# style for constants is `FooBarQux`

```suggestion
    private const string DefaultPackageType = "";
```

loic-sharma: It looks like there's still two more constants that need to be updated

src/Bootstrap/less/theme/page-list-packages.less
t-mog-msft: Oh okay, I thought you meant why is the "margin: 0" styling not in page-list-packages selector

src/NuGetGallery/UrlHelperExtensions.cs
```diff
@@ -167,6 +169,16 @@ public static string StatisticsPackageDownloadsDetailReport(this UrlHelper url,
                 routeValues["prerel"] = "false";
             }
 
+            if(!string.IsNullOrWhiteSpace(packageType))
```
loic-sharma, line 172:
```suggestion
            if (!string.IsNullOrWhiteSpace(packageType))
```

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -6,9 +6,40 @@
...
+{
+    <div style="display: flex;">
+        <label>
+            <input type="radio" name="@radioGroupName" @(string.Equals(optionValue, optionCompareValue, StringComparison.OrdinalIgnoreCase) || (isDefault && optionCompareValue == null) ? "checked" : "") value="@optionValue">
+            @optionName
+        </label>
+    </div>
+}
+
+@functions{
```
loic-sharma, line 29: Should the `{` be on a new line?

src/NuGetGallery/OData/SearchService/SearchAdaptor.cs
```diff
@@ -41,43 +54,13 @@ public static SearchFilter GetSearchFilter(string q, int page, bool includePrere
                 PackageType = packageType,
             };
 
-            switch (sortOrder)
+            if(sortOrder == null || !SortOrders.TryGetValue(sortOrder, out var sortOrderValue))
```
loic-sharma, line 57:
```suggestion
            if (sortOrder == null || !SortOrders.TryGetValue(sortOrder, out var sortOrderValue))
```

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -4937,6 +4943,59 @@ public async Task RedirectsToDefaultWhenInvalidAdvancedSearch(string sortBy, Sor
...
+                if (!expectedSupport)
+                {
+                    Assert.NotNull(_cache.Get("DefaultSearchResults"));
+                }
+                else
+                {
+                    Assert.Null(_cache.Get("DefaultSearchResults"));
+                }
+
+                searchService.Verify(x => x.Search(It.IsAny<SearchFilter>()), Times.Once);
```
loic-sharma, line 4996: Let's check that the search filter is as expected. That is, we should pass sort by and package type values only if `expectedSupport` is true.

src/NuGetGallery/Infrastructure/Lucene/LuceneSearchService.cs
```diff
@@ -27,6 +27,8 @@ public class LuceneSearchService : ISearchService
 
         public bool ContainsAllVersions { get { return false; } }
 
+        public bool SupportsAdvancedSearch { get { return false; } }
```
loic-sharma, line 30: You can shorten this to:

```suggestion
        public bool SupportsAdvancedSearch => false;
```

The arrow syntax is relatively new, hence why `ContainsAllVersions` doesn't use it. You can update that property too :)

src/NuGetGallery/OData/SearchService/SearchAdaptor.cs
```diff
@@ -25,6 +25,19 @@ public static class SearchAdaptor
         ///     Determines the maximum number of packages returned in a single page of an OData result.
         /// </summary>
         internal const int MaxPageSize = 100;
+        private static IReadOnlyDictionary<string, SortOrder> SortOrders = new Dictionary<string, SortOrder>(StringComparer.OrdinalIgnoreCase)
```
loic-sharma, line 28: Can this be `readonly`?

```suggestion
        private static readonly IReadOnlyDictionary<string, SortOrder> SortOrders = new Dictionary<string, SortOrder>(StringComparer.OrdinalIgnoreCase)
```

src/NuGetGallery/Views/Shared/ListPackages.cshtml
loic-sharma: I think you need an `Html.Encode` on the search term here too

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -36,11 +68,75 @@
...
+            {
+                <a class="prerel-filter" href="?q=@Model.SearchTerm@(Model.IncludePrerelease ? "&prerel=false" : "")">@(Model.IncludePrerelease ? "Hide" : "Show") prerelease</a>
+            }
         </div>
     </div>
-
+    @if (Model.IsAdvancedSearchFlightEnabled)
+    {
+        <form method="get" class="row clearfix collapse advanced-search-panel@(Model.ShouldDisplayAdvancedSearchPanel ? " in" : "")" id="advancedSearchPanel">
+            @if(Model.PackageType != null && !UiSupportedPackageTypes.ContainsKey(Model.PackageType))
```
loic-sharma, line 88:
```suggestion
            @if (Model.PackageType != null && !UiSupportedPackageTypes.ContainsKey(Model.PackageType))
```

src/NuGetGallery/Views/Shared/ListPackages.cshtml
```diff
@@ -140,5 +236,14 @@
                 });
             });
         });
+
+        $('#prerel-checkbox').on('change', function (event) {
+            $('#prerel').val($(event.target).is(':checked'));
+        });
+
+        $('#reset-advanced-search').on('click', function () {
+            var searchTermStr = `?q=${'@Html.Encode(Model.SearchTerm)'}`
```
loic-sharma, line 245: You should be able to remove this js string interpolation

PR #8109, opened 2020-07-16
Add new column name "EmbeddedReadmeType" in Package table 

lyndaidaii:

* Add new column name "EmbeddedReadmeFileType" to Package table

raw sql
 Sql("ALTER TABLE [dbo].[Packages] ADD EmbeddedReadmeFileType INT NOT NULL");

Addresses https://github.com/nuget/engineering/issues/3268


src/NuGet.Services.Entities/EmbeddedReadmeFileType.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    public enum EmbeddedReadmeFileType
+    {
+        /// <summary>
+        /// Indicates that the package has no embedded readme file.
+        /// </summary>
+        Absent = 0,
+
+        /// <summary>
+        /// Indicates that tne embedded readme file is markdown.
```
joelverhagen, line 17:
```suggestion
        /// Indicates that the embedded readme file is markdown.
```

zhhyu:
I think there is a need to understand the behavior on how SQL database handles the already existed data in the table, when we add a not null-able column.

src/NuGet.Services.Entities/Package.cs
```diff
@@ -285,5 +285,7 @@ public bool HasReadMe
         /// A flag that indicates that the package metadata had an embedded icon specified.
         /// </summary>
         public bool HasEmbeddedIcon { get; set; }
+
+        public EmbeddedReadmeFileType EmbeddedReadmeFileType { get; set; }
```
zhhyu, line 289: EmbeddedLicenseType for Licenses?
Maybe we also use "EmbeddedReadmeType"? It's not a problem at all. I just highlight this here, and maybe others will have some opinions here.

src/NuGetGallery/Migrations/202007171757502_AddEmbeddedReadmeFileTypeColumn.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace NuGetGallery.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddEmbeddedReadmeFileTypeColumn : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.Packages", "EmbeddedReadmeFileType", c => c.Int(nullable: false));
```
agr, line 10: Does it automatically default to the value of `0` for the new column?

loic-sharma: You should be able to be explicit here by adding `defaultValue: 0`. See: https://docs.microsoft.com/en-us/dotnet/api/system.data.entity.migrations.builders.columnbuilder.int?view=entity-framework-6.2.0#System_Data_Entity_Migrations_Builders_ColumnBuilder_Int_System_Nullable_System_Boolean__System_Boolean_System_Nullable_System_Int32__System_String_System_String_System_String_System_Collections_Generic_IDictionary_System_String_System_Data_Entity_Infrastructure_Annotations_AnnotationValues__

src/NuGet.Services.Entities/Package.cs
loic-sharma: I also slightly prefer `EmbeddedReadmeType`, but I don't feel strongly about it either

lyndaidaii: Yes, EmbeddedLicenseType for Licenses. I will change to 

EmbeddedReadmeType
```suggestion
        public EmbeddedReadmeType EmbeddedReadmeType { get; set; }
```

 to be consistent

PR #8119, opened 2020-07-22
Add feature flag enable embedded readme file 

lyndaidaii:

* add feature flag to enable embedded readme file during development

Addresses https://github.com/nuget/engineering/issues/2342


src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -662,6 +662,7 @@ private Package CreatePackageFromNuGetPackage(PackageRegistration packageRegistr
             package.EmbeddedLicenseType = GetEmbeddedLicenseType(packageMetadata);
             package.LicenseExpression = GetLicenseExpression(packageMetadata);
             package.HasEmbeddedIcon = !string.IsNullOrWhiteSpace(packageMetadata.IconFile);
+            package.EmbeddedReadmeType = GetEmbeddedReadmeType(packageMetadata);
```
loic-sharma, line 665: Can we unit test this?

src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -711,6 +712,24 @@ private static EmbeddedLicenseFileType GetEmbeddedLicenseType(string licenseFile
             throw new ArgumentException($"Invalid file name: {licenseFileName}");
         }
 
+        private static EmbeddedReadmeFileType GetEmbeddedReadmeType(PackageMetadata packageMetadata)
+        {
+            const string MarkdownFileExtension = ".md";
```
cristinamanum, line 717: There is another const string (const string MarkdownFileExtension = ".md";) at the private static EmbeddedLicenseFileType GetEmbeddedLicenseType(string licenseFileName) level .

Why not consolidate it in a global readonly string var?

lyndaidaii: Probably not make it global readonly string. it's const. it will be global const?

src/NuGetGallery.Services/Configuration/IFeatureFlagService.cs
```diff
@@ -138,5 +138,10 @@ public interface IFeatureFlagService
         /// Whether the user is able to see or manage the package renames information.
         /// </summary>
         bool IsPackageRenamesEnabled(User user);
+
+        /// <summary>
+        /// Whether the user is able to upload package with an embedded readme file.
```
zhhyu, line 143: upload the package

lyndaidaii:
I committed other two approved PR to this change. 
 (1. https://github.com/NuGet/NuGetGallery/pull/8120
 2. https://github.com/NuGet/NuGetGallery/pull/8123)
Also changed base branch from my local feature branch to dev. 

Need review for pull request for dev

src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -711,6 +712,25 @@ private static EmbeddedLicenseFileType GetEmbeddedLicenseType(string licenseFile
...
+            }
+
+            var extension = Path.GetExtension(packageMetadata.ReadmeFile);
+
+            if (MarkdownFileExtension.Equals(extension, StringComparison.OrdinalIgnoreCase))
+            {
+                return EmbeddedReadmeFileType.Markdown;
+            }
+
+            throw new ArgumentException($"Invalid file name: {packageMetadata.ReadmeFile}");
```
joelverhagen, line 731: Can this error message be a little more specific? For example, `The file name for the package readme must have the "md" file extension`. Consider looking for other places in the code where we validate file extension and ensure the wording is consistent. I think the embedded icon code may validate extension.

src/NuGetGallery/Services/PackageMetadataValidationService.cs
```diff
@@ -0,0 +1,854 @@
...
+                return PackageValidationResult.Invalid(Strings.UploadPackage_UnsupportedIconImageFormat);
+            }
+
+            return null;
+        }
+
+        private async Task<PackageValidationResult> CheckReadmeMetadataAsync(PackageArchiveReader nuGetPackage, List<IValidationMessage> warnings, User user)
+        {
+            var nuspecReader = GetNuspecReader(nuGetPackage);
+            var readmeElement = nuspecReader.ReadmeElement;
```
joelverhagen, line 421: Is this case sensitive? For example, if someone does `<README>`, will it work? It should not, at least for new element. V3 behaves weirdly when the wrong casing is provided and in general XML is not case insensitive in element names.

tests/NuGetGallery.Facts/Services/PackageMetadataValidationServiceFacts.cs
```diff
@@ -0,0 +1,2106 @@
...
+                }
+                Assert.Empty(result.Warnings);
+            }
+
+            [Fact]
+            public async Task RejectsLongReadme()
+            {
+                const int ExpectedMaxReadmeLength = 1024 * 1024;
+
+                var readmeText = new String('a', ExpectedMaxReadmeLength + 100);
```
joelverhagen, line 1487: Consider N - 1, N, and N + 1 tests. N + 100 is arbitrary

tests/NuGetGallery.Facts/Services/PackageMetadataValidationServiceFacts.cs
```diff
@@ -0,0 +1,2106 @@
...
+                    GetPackageMetadata(_nuGetPackage),
+                    _currentUser);
+
+                Assert.Equal(PackageValidationResultType.Accepted, result.Type);
+                Assert.Null(result.Message);
+                Assert.Empty(result.Warnings);
+            }
+
+            [Theory]
+            [InlineData("<readme><something/></readme>")]
```
joelverhagen, line 1398: What if the element has only whitespace?
What if there is only a comment?
What if there are attributes?

zhhyu:
> using System.Collections.Generic;

It seems that there are more libraries which you can remove because of the code refactoring. Including: System.Collections.Generic, System.Linq, System.Threading.

---
Refers to: src/NuGetGallery/Services/PackageUploadService.cs:5 in 9ecfffe. [](commit_id = 9ecfffe307a8f4bffc9ebd2295dc60e42f00d682, deletion_comment = False)

zhhyu:
> using System.Collections.Generic;

It seems that more libraries can be removed because of the code refactoring: "System.Collections.Generic", "System.Linq" and "System.Threading".

---
Refers to: src/NuGetGallery/Services/PackageUploadService.cs:5 in 9ecfffe. [](commit_id = 9ecfffe307a8f4bffc9ebd2295dc60e42f00d682, deletion_comment = False)

src/NuGetGallery.Services/Configuration/IFeatureFlagService.cs
```diff
@@ -138,5 +138,10 @@ public interface IFeatureFlagService
         /// Whether the user is able to see or manage the package renames information.
         /// </summary>
         bool IsPackageRenamesEnabled(User user);
+
+        /// <summary>
+        /// Whether the user is able to the upload package with an embedded readme file.
+        /// </summary>
+        bool IsUploadEmbeddedReadmeEnabled(User user);
```
zhhyu, line 145: I prefer to keep the consistency with icons like: AreEmbeddedReadmeEnabled. Because you may also need this feature flag for package detail page, package management page, and package preview page.

src/NuGetGallery.Services/Configuration/FeatureFlagService.cs
```diff
@@ -37,6 +37,7 @@ public class FeatureFlagService : IFeatureFlagService
         private const string Get2FADismissFeedback = GalleryPrefix + "Get2FADismissFeedback";
         private const string UsabillaOnEveryPageFeatureName = GalleryPrefix + "UsabillaEveryPage";
         private const string PackageRenamesFeatureName = GalleryPrefix + "PackageRenames";
+        private const string UploadEmbeddedReadmeFeatureName = GalleryPrefix + "UploadEmbeddedReadmeFeatureName";
```
zhhyu, line 40: For this place, maybe it's better to use "EmbeddedReadmeFlightName = GalleryPrefix + "EmbeddedReadmes"" and keep consistency with icons. The flight one is to determine who is able to use this feature, while the feature one is more like whether we want to turn on this feature for everyone.

src/NuGetGallery.Services/PackageManagement/PackageService.cs
zhhyu: I agree with @cristinamanum that you should put this const value as a field of this class rather than a local variable, following DRY principle.

src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -711,6 +712,25 @@ private static EmbeddedLicenseFileType GetEmbeddedLicenseType(string licenseFile
             throw new ArgumentException($"Invalid file name: {licenseFileName}");
         }
 
+        private static EmbeddedReadmeFileType GetEmbeddedReadmeType(PackageMetadata packageMetadata)
+        {
+            const string MarkdownFileExtension = ".md";
+
+            if (packageMetadata.ReadmeFile == null)
```
zhhyu, line 719: Is that possible that "packageMetadata.ReadmeFile" is an empty string? What action we should take for this case?

src/NuGetGallery/Services/PackageUploadService.cs
```diff
@@ -9,68 +9,26 @@
...
-        private const string AllowedLicenseVersion = "1.0.0";
-        private const string Unlicensed = "UNLICENSED";
-
         private readonly IPackageService _packageService;
         private readonly IPackageFileService _packageFileService;
         private readonly IEntitiesContext _entitiesContext;
         private readonly IReservedNamespaceService _reservedNamespaceService;
         private readonly IValidationService _validationService;
         private readonly IAppConfiguration _config;
         private readonly ITyposquattingService _typosquattingService;
```
zhhyu, line 28: Do we still need to inject "ITyposquattingService" and "_config" in the PackageUploadService?

src/NuGetGallery/Services/PackageMetadataValidationService.cs
```diff
@@ -0,0 +1,854 @@
...
+
+            if (!isPng && !isJpg)
+            {
+                return PackageValidationResult.Invalid(Strings.UploadPackage_UnsupportedIconImageFormat);
+            }
+
+            return null;
+        }
+
+        private async Task<PackageValidationResult> CheckReadmeMetadataAsync(PackageArchiveReader nuGetPackage, List<IValidationMessage> warnings, User user)
```
zhhyu, line 418: >Readme [](start = 57, length = 6)

Do we use the "warnings" here? What is the purpose of introducing this?

zhhyu:
>     <comment>{0} is the type of file ('license', 'icon', etc.), {1} is the file name specified in the .nuspec</comment>

Should the comment here also include the readme?

---
Refers to: src/NuGetGallery/Strings.resx:1033 in 9ecfffe. [](commit_id = 9ecfffe307a8f4bffc9ebd2295dc60e42f00d682, deletion_comment = False)

src/NuGetGallery/Services/PackageMetadataValidationService.cs
```diff
@@ -0,0 +1,854 @@
...
+        /// <remarks>
+        /// This limit is chosen fairly arbitrarily, it has to be large enough to fit any sensible license
+        /// in plain text and small enough to not cause issues with scanning through such file a few times
+        /// during the package validation.
+        /// </remarks>
+        private const long MaxAllowedLicenseLengthForUploading = 1024 * 1024; // 1 MB
+        private const long MaxAllowedIconLengthForUploading = 1024 * 1024; // 1 MB
+        private const int MaxAllowedLicenseNodeValueLength = 500;
+        // 1 MB Keep consistent with icon, license for now, change value later once we define the size
+        private const long MaxAllowedReadmeLengthForUploading = 1024 * 1024;
```
zhhyu, line 61: What is the current size limit for readme files? We should align with the client to determine this number. It may be fine to keep this at the current stage if we only use it for testing.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -345,6 +345,11 @@ protected override void Load(ContainerBuilder builder)
                 .As<ISymbolPackageService>()
                 .InstancePerLifetimeScope();
 
+            builder.RegisterType<PackageMetadataValidationService>()
+                .AsSelf()
```
zhhyu, line 349: If we don't need to register the class itself, then DI should not register it.

src/NuGetGallery/Services/PackageMetadataValidationService.cs
```diff
@@ -0,0 +1,854 @@
...
+        private const string LicenseNodeName = "license";
+        private const string IconNodeName = "icon";
+        private const string AllowedLicenseVersion = "1.0.0";
+        private const string ReadmeNode = "readme";
```
zhhyu, line 65: It's better to keep consistency with License and Icon to use the name like "ReadmeNodeName" here.

src/NuGetGallery/Services/PackageMetadataValidationService.cs
zhhyu: @lyndaidaii We should verify these cases locally before checking in.

src/NuGetGallery/Strings.resx
```diff
@@ -1162,4 +1162,17 @@ The {1} Team</value>
...
+  </data>
+  <data name="UploadPackage_InvalidReadmeFileExtension" xml:space="preserve">
+    <value>The readme file has an invalid extension '{0}'. Extension must be one of the following: {1}.</value>
+    <comment>{0} is the readme file extension specified in the .nuspec, {1} is the list of allowed extensions</comment>
+  </data>
+  <data name="UploadPackage_ReadmeFileType" xml:space="preserve">
+    <value>readme</value>
+  </data>
+  <data name="UploadPackage_ReadmeMustBePlainText" xml:space="preserve">
+    <value>The readme file must be plain text using UTF-8 encoding.</value>
```
zhhyu, line 1176: This message seems a little ambiguous for me considering the check we do. Please talk to @agr on what is the suitable message for this check to communicate with customers.

src/NuGetGallery/Services/PackageMetadataValidationService.cs
zhhyu, line 418: >Readme [](start = 57, length = 6)

I think one thing we need to finalize and make the decision with the PM here is about whether we'd like to validate the content of the markdown file (like inline HTML). I notice that for the current readme implementation, we don't validate it when customers edit it, but take some tricks to avoid inline HTML and achieve no follow-up. But we should double check it with PM and even with the client. Do we expect to see warnings/errors or fail the pushing? Or we allow it but when we render it, we just force to edit it on our side. In addition, please also take a look at the current readme service to see how we can benefit from that place.

zhhyu: If we'd like to validate the markdown content here, it will need much more efforts, and we can handle it in another PR rather than together with this one.

lyndaidaii: yes, we would like to validate the content of markdown file, I created ticket to track this issue, it will come into picture later. it definitely needs more efforts.

src/NuGetGallery/Services/PackageMetadataValidationService.cs
lyndaidaii: It's only for testing purpose now, I also created a ticket for tracking the file size. we will adjust the file size once we decide the proper value.

src/NuGetGallery/Services/PackageMetadataValidationService.cs
lyndaidaii: warnings might be needed in later. however, I will remove it for now

src/NuGetGallery/Strings.resx
lyndaidaii: This message is follow the pattern of license.

```
  <data name="UploadPackage_LicenseMustBePlainText" xml:space="preserve">
    <value>The license file must be plain text using UTF-8 encoding.</value>
```

src/NuGetGallery/Services/PackageMetadataValidationService.cs
lyndaidaii: It is case sensitive.  if someone does <README>, it will ignore and treat there is no readme property in nuspec, followed same pattern in License and Icon 

      ```
  private const string ReadmeNodeName = "readme";
       public XElement ReadmeElement => MetadataNode.Element(MetadataNode.Name.Namespace + ReadmeNodeName);
```

File: src/NuGetGallery.Services/PackageManagement/PackageService.cs [hunk 17]
```diff
@@ -711,6 +713,23 @@ private static EmbeddedLicenseFileType GetEmbeddedLicenseType(string licenseFile
...
+            }
+
+            var extension = Path.GetExtension(packageMetadata.ReadmeFile);
+
+            if (MarkdownFileExtension.Equals(extension, StringComparison.OrdinalIgnoreCase) || string.Empty == extension)
+            {
+                return EmbeddedReadmeFileType.Markdown;
+            }
+
+            throw new ArgumentException($"The file name for the package readme must have the \"md\" file extentsion: {packageMetadata.ReadmeFile}");
```
loic-sharma, line 730:
```suggestion
            throw new ArgumentException($"The file name for the package readme must have the \"md\" file extension: {packageMetadata.ReadmeFile}");
```

Issue #8138, opened 2020-08-04
Dependency Injection broken on GitHubVulnerabilities2Db


drewgillies:
This change: https://github.com/NuGet/NuGetGallery/pull/8112, specifically [here](https://github.com/NuGet/NuGetGallery/pull/8112/files#diff-b744cb446049fcb4c87628c3d31a964fR39) has introduced a new parameter into the PackageService constructor, breaking dependency injection in GitHubVulnerabilities2Db.

Salient part of exception:
None of the constructors found with 'Autofac.Core.Activators.Reflection.DefaultConstructorFinder' on type 'NuGetGallery.PackageService' can be invoked with the available services and parameters:\r\nCannot resolve parameter 'NuGetGallery.IContentObjectService contentObjectService' of constructor 
**'Void .ctor(
NuGetGallery.IEntityRepository`1[NuGet.Services.Entities.PackageRegistration], 
NuGetGallery.IEntityRepository`1[NuGet.Services.Entities.Package], 
NuGetGallery.IEntityRepository`1[NuGet.Services.Entities.Certificate], 
NuGetGallery.Auditing.IAuditingService, 
NuGetGallery.ITelemetryService, 
NuGetGallery.Security.ISecurityPolicyService, 
NuGetGallery.IEntitiesContext, 
NuGetGallery.IContentObjectService)'. (See inner exception for details.) (See inner exception for details.) (See inner exception for details.) (See inner exception for details.)"}**

Full trace:
[...]

PR #8139, opened 2020-08-04
Unbreak GitHubVulnerabilities2Db dependency injection

drewgillies:
Addresses https://github.com/NuGet/NuGetGallery/issues/8138

If the ContentObjectService parameter becomes optional in the PackageService constructor, dependency injection in GitHubVulnerabilities2Db won't inflate the type (which is what caused the failure--no DI registration). Inflating the type properly wouldn't be the right fix here as its dependencies aren't lightweight and it's unused in this utility. So we need an optional parameter and failsafes for nulls in PackageService.

Some formatting fixes resulted from a copy/paste operation. Why not.

### DIFF
--- a/src/GitHubVulnerabilities2Db/Job.cs
+++ b/src/GitHubVulnerabilities2Db/Job.cs
@@ -5,12 +5,14 @@ using System;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Web;
 using Autofac;
 using GitHubVulnerabilities2Db.Collector;
 using GitHubVulnerabilities2Db.Configuration;
 using GitHubVulnerabilities2Db.Gallery;
 using GitHubVulnerabilities2Db.GraphQL;
 using GitHubVulnerabilities2Db.Ingest;
+using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
@@ -21,6 +23,8 @@ using NuGet.Services.Cursor;
 using NuGet.Services.Storage;
 using NuGetGallery;
 using NuGetGallery.Auditing;
+using NuGetGallery.Configuration;
+using NuGetGallery.Diagnostics;
 using NuGetGallery.Security;
 
 namespace GitHubVulnerabilities2Db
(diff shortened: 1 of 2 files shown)


drewgillies:
@loic-sharma 
> What are alternative solutions?
Another option here is to pass a "fake" ContentObjectService instead of altering PackageService. I might see if I can do that for the sake of this implementation, so that this change  doesn't pollute the services. A more rigorous change like you describe may be worthwhile, so I'll think about that too.

drewgillies:
@loic I've added registrations for all dependencies in the code and reverted the service changes. I haven't refactored services as it was becoming a yak-shaving exercise and actually made the DI code no simpler.

src/GitHubVulnerabilities2Db/Job.cs
```diff
@@ -152,6 +155,32 @@ protected void ConfigureCollectorServices(ContainerBuilder containerBuilder)
                 .Register(ctx => CreateCursor(ctx, config => config.AdvisoryCursorBlobName))
                 .As<ReadWriteCursor<DateTimeOffset>>();
 
+            containerBuilder.RegisterType<AppConfiguration>()
```
zhhyu, line 158: Should we register these interfaces within the "ConfigureGalleryServices"? https://github.com/NuGet/NuGetGallery/blob/de753f5eed01d61723e69331d9274d7565aa1fcf/src/GitHubVulnerabilities2Db/Job.cs#L75

src/GitHubVulnerabilities2Db/GitHubVulnerabilities2Db.csproj
```diff
@@ -107,6 +107,10 @@
       <Project>{c7d5e850-33fa-4ec5-8d7f-b1c8dd5d48f9}</Project>
       <Name>NuGetGallery.Services</Name>
     </ProjectReference>
+    <ProjectReference Include="..\NuGetGallery\NuGetGallery.csproj">
```
xavierdecoster, line 110: _shivers_ 👀 

Should the types referenced from NuGetGallery be put into their own project so you don't need to reference the entire gallery for this?

drewgillies: It might be an appropriate change. What I'm really trying to do is unbreak the break introduced by https://github.com/NuGet/NuGetGallery/issues/8138 so I can get on with more pressing work. Maybe a refactoring operation like this and the one @loic-sharma is describing can be tracked as separate issues. /cc @joelverhagen

drewgillies: Okay, I've removed the need for this reference by creating a fake ContentService. If ever the service is required by anything we do it GitHubVulnerabilities2Db (doubtful), we can look at removing the fake.

src/GitHubVulnerabilities2Db/Job.cs
drewgillies: Done.

PR #8141, opened 2020-08-06
Fix for: Malicious OData requests trigger exception monitoring

skofman1:
Addresses https://github.com/NuGet/Engineering/issues/3303

No need to log exceptions for user input errors. Those are not "exceptional".

### DIFF
--- a/src/NuGetGallery/Controllers/ODataV2FeedController.cs
+++ b/src/NuGetGallery/Controllers/ODataV2FeedController.cs
@@ -124,7 +124,7 @@ namespace NuGetGallery.Controllers
             }
             catch (ODataException ex) when (ex.InnerException != null && ex.InnerException is FormatException)
             {
-                // Sometimes users make invalid requests. It's not excetpional behavior, don't trace.
+                // Sometimes users make invalid requests. It's not exceptional behavior, don't trace.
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Controllers/ODataV2FeedController.cs
```diff
@@ -121,6 +122,10 @@ public class ODataV2FeedController
                     customQuery = true;
                 }
             }
+            catch (ODataException ex) when (ex.InnerException != null && ex.InnerException is FormatException)
+            {
+                // Sometimes users make invalid requests. It's not excetpional behavior, don't trace.
```
xavierdecoster, line 127:
```suggestion
                // Sometimes users make invalid requests. It's not exceptional behavior, don't trace.
```

loic-sharma:
Will this still return an HTTP 400 response status code? Can we add a functional test for these cases?

skofman1:
@loic-sharma , yes, this still returns a 400. I manually tested this. The change only prevents an exception from getting logged. 
I think adding a functional test here is out of scope and low pri. I will manually test during rollout that the 400 behavior is not impacted and exceptions are no longer logged for it.

PR #8144, opened 2020-08-07
When V1 Odata requests OrderBy is invalid, return 400

skofman1:
Fixes: https://github.com/NuGet/Engineering/issues/3308

This causes monitoring noise.


src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -58,11 +59,23 @@ public IHttpActionResult Get(ODataQueryOptions<V1FeedPackage> options)
             {
                 return BadRequest(ODataQueryVerifier.GetValidationFailedMessage(options));
             }
+
+            bool shouldIgnoreOrderById;
+
```
zhhyu, line 64: Nit: blank line

zhhyu:
Do we need the similar protection for "ODataV2FeedController"?

src/NuGetGallery/Controllers/ODataV1FeedController.cs
skofman1: the blank line separates the var declaration from the try{} statement.. you think they should be together?

skofman1:
@zhhyu , yes! Will add similar logic to V2 OData

src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -58,11 +59,23 @@ public IHttpActionResult Get(ODataQueryOptions<V1FeedPackage> options)
...
+            try
+            {
+                shouldIgnoreOrderById = ShouldIgnoreOrderById(options);
+            }
+            catch(ODataException)
```
loic-sharma, line 69:
```suggestion
            catch (ODataException)
```

src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -58,11 +59,23 @@ public IHttpActionResult Get(ODataQueryOptions<V1FeedPackage> options)
             {
                 return BadRequest(ODataQueryVerifier.GetValidationFailedMessage(options));
             }
+
+            bool shouldIgnoreOrderById;
+
+            try
+            {
+                shouldIgnoreOrderById = ShouldIgnoreOrderById(options);
```
loic-sharma, line 67: Since `ShouldIgnoreOrderById` has an "expected" error case, could we update the method such that it doesn't throw in this "expected" case? Maybe make the method `bool TryShouldIgnoreOrderById(options, out bool shouldIgnoreOrderById)`, and have it return `false` if an `ODataException` is thrown. This would ensure that future uses of this method would properly handle this "expected" error case.

src/NuGetGallery/Controllers/ODataV1FeedController.cs
zhhyu: It looks fine.

src/NuGetGallery/Controllers/ODataV1FeedController.cs
skofman1: fixed :)

PR #8146, opened 2020-08-07
Move GitHub vulnerabilities verifier to its own Job

drewgillies:
This continues to address https://github.com/NuGet/NuGetGallery/issues/7685, as we need the job to authorize using key vault secrets, so we can run it on a PROD VM. Moving it into existing job/config infrastructure achieves this, even though it won't need to be run as an ongoing job (hence Job.RunOnce).

@ryuyu and I discussed this at length as we were trying to get this to connect to the PROD gallery db as a GalleryTools command with a command line connection string authenticating via integrated AAD, but this simply didn't work except on DEV.


drewgillies:
@ryuyu I've included the signing props and targets in the project file and would like your opinion on whether I've done the right thing here. Everything works locally though, and when I have a signed build I can test it for real.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilityServiceVerifier.cs
```diff
@@ -0,0 +1,121 @@
...
+                            } is not properly marked vulnerable to vulnerability!");
+                        HasErrors = true;
+                    }
+                }
+            }
+
+            return Task.CompletedTask;
+        }
+
+        public bool HasErrors { get; private set; }
```
zhhyu, line 119: It's better to put this property at the top rather than below methods.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilityServiceVerifier.cs
```diff
@@ -0,0 +1,121 @@
...
+            _entitiesContext = entitiesContext ?? throw new ArgumentNullException(nameof(entitiesContext));
+        }
+
+        public void ApplyExistingVulnerabilitiesToPackage(Package package)
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool withdrawn)
+        {
```
zhhyu, line 30: Maybe null check on "vulnerability"?

src/VerifyGitHubVulnerabilities/Job.cs
```diff
@@ -0,0 +1,115 @@
...
+            ConfigureQueryServices(containerBuilder);
+            ConfigureIngestionServices(containerBuilder);
+        }
+
+        protected void ConfigureIngestionServices(ContainerBuilder containerBuilder)
+        {
+            ConfigureGalleryServices(containerBuilder);
+
+            containerBuilder
+                .RegisterType<PackageVulnerabilityService>()
```
zhhyu, line 65: Should we register "PackageVulnerabilityServiceVerifier" here for the interface? Sorry if I misunderstand something here.

src/VerifyGitHubVulnerabilities/Job.cs
```diff
@@ -0,0 +1,115 @@
...
+        public override async Task Run()
+        {
+            Console.Write("Fetching vulnerabilities from GitHub...");
+            var advisoryQueryService = _serviceProvider.GetRequiredService<IAdvisoryQueryService>();
+            var advisories = await advisoryQueryService.GetAdvisoriesSinceAsync(DateTimeOffset.MinValue, CancellationToken.None);
+            Console.WriteLine($" FOUND {advisories.Count} advisories.");
+
+            Console.WriteLine("Fetching vulnerabilities from DB...");
+            var verifier = new PackageVulnerabilityServiceVerifier(_serviceProvider.GetRequiredService<IEntitiesContext>());
+            var ingestor = new AdvisoryIngestor(verifier, new GitHubVersionRangeParser());
```
zhhyu, line 38: Could we depend on DI to create these dependencies here? I see we register some of them in the DI container below. Maybe I misunderstand something here.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilityServiceVerifier.cs
drewgillies: It's just moved code, but I agree.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilityServiceVerifier.cs
drewgillies: Yep :)

src/VerifyGitHubVulnerabilities/Job.cs
drewgillies: Good catch--that was just left from copied code and is irrelevant.

src/VerifyGitHubVulnerabilities/Job.cs
drewgillies: I could do that--it's not as straightforward as it looks as we need the verified type to access the HasErrors property, but I can cast it for that.

drewgillies: It's more practical to create these instances directly, as we need the `verifier` reference to test the `HasErrors` property after execution (which isn't contained in the interface, and it's not worth adding it for the sake of a utility like this).

zhhyu:
Maybe get another approval before merging it.

src/VerifyGitHubVulnerabilities/Job.cs
zhhyu: I see. That makes sense.

Issue #7975, opened 2020-05-03
Embedded licenses files in MD are not encoded as HTML for display.


JimBobSquarePants:
### Describe the bug

According to the specification embedded license files are accepted in either TXT or MD format.
https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg#license

However, MD files are not converted to HTML for display on the site and are rendered as raw Markdown.

### To Reproduce
1. On this URL 'https://www.nuget.org/packages/SixLabors.ImageSharp/1.0.0-rc0001/License'

### Expected behavior

I would expect the markdown to be encoded as HTML.

rocklan:
@skofman1 I have created a pull request for this. I'm sure it will take a while for it to be tested properly. It works on my machine :)

skofman1:
Thanks for the contribution @rocklan ! Very cool! 
@zhhyu will work with you next week on reviewing and testing the PR and eventually merging it in. Can't wait to see it live!

rocklan:
thanks and appreciate your enthusiam @skofman1 :)

zhhyu:
Hi, all! We will enable this feature after finish following issues: 1. https://github.com/NuGet/NuGetGallery/issues/8277; 2. https://github.com/NuGet/NuGetGallery/issues/8278 Thanks! :relaxed:

PR #8153, opened 2020-08-12
Display markdown licence files as HTML 

rocklan:
This adds support for displaying licence MD files as HTML. A few notes:

- I have refactored the `ReadmeService.GetReadMeHtml()` method into it's own service, named MarkdownService. I did this in one commit so that's it's a bit more obvious what's going on. 

- The CommonMark library has a weird bug that if a string begins with a BOM, it doesn't convert the first line properly. As a result I stripped it out and then it works ok. I've added a test for this. (It looks like the version of CommonMark used is old, and the package is also no longer supported, but that's another matter)

- A few of the unit tests eg `PackageControllerFacts.TheDisplayPackageMethod.WhenHasReadMeAndMarkdownExists_ReturnsContent` test that the markdown -> html is working properly, but that's arguably not the concern of the PackageControllerFacts and are tested anyway, inside the MarkdownServiceFacts. However, in the interests of keeping things working I've changed the `PackageControllerFacts` to inject an instantiated `MarkdownService` instead of a Mock. However if someone wiser than me wishes to change it to test against a mock and not test for the MD -> HTML logic then that seems to make sense to me, but it depends on how pure we want the unit tests to be. 

Addresses #7975

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/PackageService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageService.cs
@@ -755,7 +755,7 @@ namespace NuGetGallery
                 return EmbeddedReadmeFileType.Markdown;
             }
 
-            throw new ArgumentException($"The file name for the package readme must have the \"md\" file extentsion: {packageMetadata.ReadmeFile}");
+            throw new ArgumentException($"The file name for the package readme must have the \"md\" file extension: {packageMetadata.ReadmeFile}");
         }
 
         private static void ValidateSupportedFrameworks(string[] supportedFrameworks)


tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
```diff
@@ -0,0 +1,56 @@
...
+            [InlineData("<a href=\"javascript:alert('test');\">", "<p>&lt;a href=&quot;javascript:alert('test');&quot;&gt;</p>")]
+            public void EncodesHtmlInMarkdown(string originalMd, string expectedHtml)
+            {
+                Assert.Equal(expectedHtml, StripNewLines(
+                    _markdownService.GetHtmlFromMarkdown(originalMd).Content));
+            }
+
+            [Theory]
+            [InlineData("# Heading", "<h2>Heading</h2>", false)]
+            [InlineData("\ufeff# Heading with BOM", "<h2>Heading with BOM</h2>", false)]
```
rocklan, line 33: This is the extra test that has been added that was previously failing. It wasn't noticeable before as the only time MD text was parsed was sent through was from from ajax code, which doesn't include BOM for text.

Issue #8136, opened 2020-07-31
Breadcrumb link to package details from Report Abuse page doesn't for packages with additional build metadata in version string


chgill-MSFT:
### Describe the bug
**Applies to packages with additional metadata in their version string.**

After clicking on "Report" link on package details page, breadcrumb to get back to the package details page doesn't work if the package version string has additional build metadata. This is likely because the additional build metadata is appended to the URL when it shouldn't be.

Example of version string with additional build metadata:
![image](https://user-images.githubusercontent.com/15097183/89084748-dcde9400-d345-11ea-9f24-fb3e1dad28e2.png)

### To Reproduce
1. Go to https://www.nuget.org/packages/NuGet.Packaging/5.6.0/ReportAbuse
2. Click breadcrumb back to the package details page in the top left corner
![image](https://user-images.githubusercontent.com/15097183/89083260-4f9b3f80-d345-11ea-8a96-95bb030f8667.png)

3. Get Error 404
4. See that it works if you remove the additional build metadata from the URL

URL with metadata (wrong): https://www.nuget.org/packages/NuGet.Packaging/5.6.0%2b636570e68732c1f718ede9ca07802d7b1cc69aa0

Correct URL: https://www.nuget.org/packages/NuGet.Packaging/5.6.0

### Expected behavior
Breadcrumb link back to package details page should work.

### GIF of bug repro with NuGet.Packaging
![Link back from Report page doesn't work with package versions with build metadata](https://user-images.githubusercontent.com/15097183/89083178-fe8b4b80-d344-11ea-9390-c97dedb8883a.gif)

### Additional context
Additional examples for repro:
- https://www.nuget.org/packages/Lib.Build/5.0.0-preview.2.1%2b5cb4e1e
- https://www.nuget.org/packages/ConfigHandler/1.2.1%2b42c1ae0

rocklan:
@chgill-MSFT I have submitted #8157 pull request to fix this issue.

joelverhagen:
This change is deployed to nuget.org. Thanks again for your contribution!

PR #8157, opened 2020-08-13
Fixed breadcrumb links for packages with build metadata

rocklan:
Fixes breadcrumb links for packages with build metadata:

image

The only question now is if we want to continue to display the build metadata in the link (the +6...) part. I think it's ok, but that's me. 

Addresses #8136

### DIFF
--- a/src/NuGetGallery/UrlHelperExtensions.cs
+++ b/src/NuGetGallery/UrlHelperExtensions.cs
@@ -228,7 +228,7 @@ namespace NuGetGallery
             string version,
             bool relativeUrl = true)
         {
-            string normalized = NuGetVersionFormatter.Normalize(version);
+            string normalized = (version != null) ? NuGetVersionFormatter.Normalize(version) : version;
 
             string result = GetRouteLink(
                 url,
--- a/src/NuGetGallery/UrlHelperExtensions.cs
+++ b/src/NuGetGallery/UrlHelperExtensions.cs
@@ -228,6 +228,8 @@ namespace NuGetGallery
             string version,
             bool relativeUrl = true)
         {
+            string normalized = NuGetVersionFormatter.Normalize(version);
+
             string result = GetRouteLink(
                 url,
                 RouteName.DisplayPackage,
@@ -235,7 +237,7 @@ namespace NuGetGallery
                 routeValues: new RouteValueDictionary
                 {
                     { "id", id },
-                    { "version", version }
+                    { "version", normalized }
                 });
 
             // Ensure trailing slashes for versionless package URLs, as a fix for package filenames that look like known file extensions
--- a/tests/NuGetGallery.Facts/UrlHelperExtensionsFacts.cs
+++ b/tests/NuGetGallery.Facts/UrlHelperExtensionsFacts.cs
@@ -31,6 +31,30 @@ namespace NuGetGallery
             }
         }
 
+    
+        public class ThePackageBaseHelperMethod
+            : TestContainer
+        {
+            [Fact]
+            public void UsesNormalizedVersionInUrls()
+            {
+                var package = new Package
+                {
+                    PackageRegistration = new PackageRegistration
+                    {
+                        Id = "TestPackageId"
+                    },
+                    NormalizedVersion = "1.0.0-alpha.1",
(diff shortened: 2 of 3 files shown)


src/NuGetGallery/UrlHelperExtensions.cs
```diff
@@ -228,14 +228,16 @@ public static string Package(this UrlHelper url, string id, bool relativeUrl = t
             string version,
             bool relativeUrl = true)
         {
+            string normalized = NuGetVersionFormatter.Normalize(version);
```
joelverhagen, line 231: I think `version` can be `null` in some contexts, based on usage, e.g. https://github.com/NuGet/NuGetGallery/blob/b3c174ecc7affadf411d88270a50ed4b5eb1f2a4/src/NuGetGallery/Controllers/JsonApiController.cs#L135

I'd recommend either putting this in an `if` to protect from this case or to add a unit test to verify and "lock-in" the behavior of `Normalize` with a `null` value passed in.

Alternatively, you can update the particular call site to ensure `null` is never passed in but I think this is a desirable, centralized behavior.

joelverhagen:
Would it be possible to get a unit test to cover this case?

src/NuGetGallery/UrlHelperExtensions.cs
rocklan: Null reference checks? They'll never catch on 😄 . Sure, my mistake, done.

rocklan:
> Would it be possible to get a unit test to cover this case?

I've added a unit test to cover this in the `ThePackageBaseHelperMethod` class. And having done so, I can see how this bug snuck in. There is already a test to cover this exact scenario, inside the `ThePackageHelperMethod` class, but the problem is it calls one of the overloads of `Package`, where a `Package` object is passed through, and this function passes through the `package.NormalizedVersion`. However the code that renders the breadcrumbs doesn't call this method, it calls the one that accepts two strings, and it passes though the raw version instead of the normalised one. I don't think calling normalize() twice will cause any issues.

tests/NuGetGallery.Facts/UrlHelperExtensionsFacts.cs
```diff
@@ -53,6 +77,7 @@ public void UsesNormalizedVersionInUrls()
                 Assert.EndsWith(package.NormalizedVersion, fixedUrl);
             }
 
+
```
joelverhagen, line 80: nit: don't need the extra line

rocklan: ok done, fixed

joelverhagen:
Hey, it looks like this test is failing on your branch (per CI):
````
##[error]    NuGetGallery.ApiControllerFacts+TheGetStatsDownloadsAction.VerifyRecentPopularityStatsDownloads [FAIL]

      unexpected content result[3].Gallery
      Expected: True
      Actual:   False
      Stack Trace:
        tests\NuGetGallery.Facts\Controllers\ApiControllerFacts.cs(0,0): at NuGetGallery.ApiControllerFacts.TheGetStatsDownloadsAction.<VerifyRecentPopularityStatsDownloads>d__0.MoveNext()
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

rocklan:
> Hey, it looks like this test is failing on your branch (per CI): NuGetGallery.ApiControllerFacts+TheGetStatsDownloadsAction.VerifyRecentPopularityStatsDownloads [FAIL]

I've committed a fix for this, it was previously checking that the version number on the end of a link was `1.1` but now as they are normalized it should be checking for `1.1.0`

joelverhagen:
Awesome, thanks for this contribution @rocklan! 🥇

Issue #7806, opened 2020-01-15
Allow sorting of Manage Package page by download count


ryuyu:
Suggestion that came through support.

joelverhagen:
This is deployed to PROD. Thanks again, @rocklan!

PR #8158, opened 2020-08-13
package list can be sorted by ID, owners, downloads and version

rocklan:
This adds support for ordering the "Published Packages" and "Unlisted Packages" tables on the "Manage my packages" screen, eg:

sorting

Clicking on any of the column headers sorts by that column. I am by no means a JS guru but hopefully this does what we want and works for most people.

Addresses https://github.com/NuGet/NuGetGallery/issues/7806

### DIFF
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -585,9 +585,11 @@ namespace NuGetGallery
                 .ToList();
 
             for (int i = 0; i < listedPackages.Count; i++)
+            {
                 listedPackages[i].VersionSortOrder = i;
+            }
 
-            listedPackages.Sort((x, y) => x.Id.CompareTo(y.Id));
+            listedPackages.Sort((x, y) => string.Compare(x.Id,y.Id, StringComparison.OrdinalIgnoreCase));
 
             return listedPackages;
         }
--- a/src/NuGetGallery/Controllers/UsersController.cs
+++ b/src/NuGetGallery/Controllers/UsersController.cs
@@ -12,6 +12,7 @@ using System.Web;
 using System.Web.Mvc;
 using NuGet.Services.Entities;
 using NuGet.Services.Messaging.Email;
+using NuGet.Versioning;
 using NuGetGallery.Areas.Admin;
 using NuGetGallery.Areas.Admin.ViewModels;
 using NuGetGallery.Authentication;
@@ -521,16 +522,12 @@ namespace NuGetGallery
             var wasAADLoginOrMultiFactorAuthenticated = User.WasMultiFactorAuthenticated() || User.WasAzureActiveDirectoryAccountUsedForSignin();
 
             var packages = PackageService.FindPackagesByAnyMatchingOwner(currentUser, includeUnlisted: true);
-            var listedPackages = packages
-                .Where(p => p.Listed && p.PackageStatusKey == PackageStatus.Available)
-                .Select(p => _listPackageItemRequiredSignerViewModelFactory.Create(p, currentUser, wasAADLoginOrMultiFactorAuthenticated))
-                .OrderBy(p => p.Id)
-                .ToList();
-            var unlistedPackages = packages
-                .Where(p => !p.Listed || p.PackageStatusKey != PackageStatus.Available)
-                .Select(p => _listPackageItemRequiredSignerViewModelFactory.Create(p, currentUser, wasAADLoginOrMultiFactorAuthenticated))
-                .OrderBy(p => p.Id)
-                .ToList();
+
+            var listedPackages = GetPackages(packages, currentUser, wasAADLoginOrMultiFactorAuthenticated,
+                p => p.Listed && p.PackageStatusKey == PackageStatus.Available);
+            
+            var unlistedPackages = GetPackages(packages, currentUser, wasAADLoginOrMultiFactorAuthenticated,
+                p => !p.Listed || p.PackageStatusKey != PackageStatus.Available);
 
             // find all received ownership requests
             var userReceived = _packageOwnerRequestService.GetPackageOwnershipRequests(newOwner: currentUser);
@@ -565,6 +562,34 @@ namespace NuGetGallery
                 IsCertificatesUIEnabled = ContentObjectService.CertificatesConfiguration?.IsUIEnabledForUser(currentUser) ?? false
             };
             return View(model);
+        }
+
+        /// <summary>
+        /// Returns all packages based on the predicate, with the VersionSortOrder populated
+        /// </summary>
+        /// <param name="packages"></param>
+        /// <param name="currentUser"></param>
+        /// <param name="wasAADLoginOrMultiFactorAuthenticated"></param>
+        /// <param name="predicate"></param>
+        /// <returns></returns>
+        private List<ListPackageItemRequiredSignerViewModel> GetPackages(
+            IEnumerable<Package> packages,
+            User currentUser,
+            bool wasAADLoginOrMultiFactorAuthenticated,
+            Func<Package, bool> predicate)
+        {
+            var listedPackages = packages
+                .Where(p => predicate(p))
+                .Select(p => _listPackageItemRequiredSignerViewModelFactory.Create(p, currentUser, wasAADLoginOrMultiFactorAuthenticated))
+                .OrderBy(p => NuGetVersion.Parse(p.FullVersion))
+                .ToList();
+
+            for (int i = 0; i < listedPackages.Count; i++)
+                listedPackages[i].VersionSortOrder = i;
+
+            listedPackages.Sort((x, y) => x.Id.CompareTo(y.Id));
+
+            return listedPackages;
         }
 
         [HttpGet]
--- a/src/NuGetGallery/Scripts/gallery/page-manage-packages.js
+++ b/src/NuGetGallery/Scripts/gallery/page-manage-packages.js
@@ -355,6 +355,39 @@
             this.RequestsSent = new OwnerRequestsListViewModel(this, initialData.RequestsSent, false, true);
         }
 
+        function setupColumnSorting() {
+            $('.sortable').click(function () {
+
+                var table = $(this).parents('table').eq(0)
+                var rows = table.find('tr:gt(0)').toArray().sort(comparer($(this).index()))
+                this.asc = !this.asc
+                if (!this.asc) { rows = rows.reverse() }
+                for (var i = 0; i < rows.length; i++) { table.append(rows[i]) }
+
+                table.find('.sortable').each(function () {
+                    var currentText = $(this).text();
+                    var newText = currentText.replace(' ▲', '').replace(' ▼', '');
+                    $(this).text(newText);
+                });
+
+                var columnText = $(this).text();
+                $(this).text(columnText + " " + (this.asc ? "▼" : "▲"));
+
+            })
+            function comparer(index) {
+                return function (a, b) {
+                    var valA = getCellValue(a, index), valB = getCellValue(b, index)
+                    return $.isNumeric(valA) && $.isNumeric(valB) ? valA - valB : valA.toString().localeCompare(valB)
+                }
+            }
+            function getCellValue(row, index) {
+                var v = $(row).children('td').eq(index).text();
+                if (v)
+                    v = v.replace(',', '');
+                return v;
+            }
+        }
+
         // Immediately load initial expander data
         showInitialPackagesData("#listed-data", initialData.ListedPackages);
         showInitialPackagesData("#unlisted-data", initialData.UnlistedPackages);
@@ -365,6 +398,11 @@
         // Set up the data binding.
         var managePackagesViewModel = new ManagePackagesViewModel(initialData);
         ko.applyBindings(managePackagesViewModel, document.body);
+
+        setupColumnSorting();
+
+        
+
     });
 
 })();
--- a/src/NuGetGallery/Scripts/gallery/page-manage-packages.js
+++ b/src/NuGetGallery/Scripts/gallery/page-manage-packages.js
@@ -38,6 +38,7 @@
             this.CanEdit = packageItem.CanEdit;
             this.CanManageOwners = packageItem.CanManageOwners;
             this.CanDelete = packageItem.CanDelete;
+            this.VersionSortOrder = packageItem.VersionSortOrder;
 
             this.FormattedDownloadCount = ko.pureComputed(function () {
                 return ko.unwrap(this.DownloadCount).toLocaleString();
@@ -381,9 +382,16 @@
                 }
             }
             function getCellValue(row, index) {
-                var v = $(row).children('td').eq(index).text();
+                var td = $(row).children('td').eq(index);
+                var sortby = td.data('sortby');
+
+                if (sortby)
+                    return sortby;
+
+                var v = td.text();
                 if (v)
                     v = v.replace(',', '');
+
                 return v;
             }
         }
@@ -400,8 +408,6 @@
         ko.applyBindings(managePackagesViewModel, document.body);
 
         setupColumnSorting();
-
-        
 
     });
 
--- a/src/NuGetGallery/Scripts/gallery/page-manage-packages.js
+++ b/src/NuGetGallery/Scripts/gallery/page-manage-packages.js
@@ -377,7 +377,7 @@
             function comparer(index) {
                 return function (a, b) {
                     var valA = getCellValue(a, index), valB = getCellValue(b, index)
-                    return $.isNumeric(valA) && $.isNumeric(valB) ? valA - valB : valA.toString().localeCompare(valB)
+                    return $.isNumeric(valA) && $.isNumeric(valB) ? valB - valA : valA.toString().localeCompare(valB)
                 }
             }
             function getCellValue(row, index) {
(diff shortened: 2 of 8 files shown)


src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -4,7 +4,11 @@
     ViewBag.Title = "Manage My Package";
     ViewBag.Tab = "Packages";
 }
-
+<style type="text/css">
+    .sortable {
+        cursor: pointer;
+    }
+</style>
```
rocklan, line 11: I couldn't find a place where to put this CSS style other than creating a new file, as the only CSS files that are included are bootstrap and fabric. Not sure what you want to do here.

joelverhagen: This can be added to the following less file:
https://github.com/NuGet/NuGetGallery/blob/master/src/Bootstrap/less/theme/base.less

You must run `npm install` and `grunt` in the /src/Bootstrap directory after making this change so that the CSS is compiled and copied into the /src/NuGetGallery directory.

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -113,14 +117,14 @@
...
+                        <th class="sortable">Owners</th>
                     @if (Model.IsCertificatesUIEnabled)
                     {
-                        <th>Signing Owner</th>
+                        <th  class="sortable">Signing Owner</th>
                     }
-                        <th>Downloads</th>
-                        <th>Latest Version</th>
+                        <th  class="sortable">Downloads</th>
+                        <th  class="sortable">Latest Version</th>
```
joelverhagen, line 127: I'm concerned about this sort since it does not sort in the standard SemVer aware way seen all over the NuGet ecosystem. From what I can tell, this is a lex sort for non-numeric values and an numeric sort for numeric values.

joelverhagen: Perhaps we can limit the sorting behavior to package ID and downloads for now?

Alternatively, we could update the server-side code (which is SemVer aware via `NuGetVersion`) to calculate a "list position" for each cell w.r.t. other cell values in its column and write this to a data field in the markup. Then the JS could key off of that instead of doing `text()` manipulation.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -355,6 +355,39 @@
...
+                });
+
+                var columnText = $(this).text();
+                $(this).text(columnText + " " + (this.asc ? "▼" : "▲"));
+
+            })
+            function comparer(index) {
+                return function (a, b) {
+                    var valA = getCellValue(a, index), valB = getCellValue(b, index)
+                    return $.isNumeric(valA) && $.isNumeric(valB) ? valB - valA : valA.toString().localeCompare(valB)
```
joelverhagen, line 380: it's not clear to me if `localeCompare` is case insensitive. All of the string values in this table are considered case insensitive and should be sorted accordingly.

joelverhagen:
OMG this is amazing. /cc @jcjiang @skofman1 

This may be a tall order, but would it be possible to have a gif recording of the behavior in action? Like initial state -> sort by downloads -> sort by downloads -> sort by package ID? If not I may be able to cook it up myself (i'm not a GIF recording pro)

One concern I have is the performance on owners with many package IDs. For example, the Microsoft org has nearly 5000 packages. Would it be possible to do some rough tests of how it "feels" as the number of package IDs goes up? If we run into trouble we can say "well, it's still okay for small pages" or we can disable the behavior for when there are too many rows (based on some hard coded safe threshold).

joelverhagen:
Is there an associated issue for this PR? I see #123 in the description (perhaps from the template) but this doesn't seem right. If an issue does not exist, could you open one and start a discussion about the pain you have, your proposed solution, and other thoughts?

rocklan:
Yes it's issue #7806. I've updated the description.

src/NuGetGallery/Views/Users/Packages.cshtml
rocklan: Thanks, gotcha, done

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
rocklan: I am pretty sure it's case insensitive by default - you can make it case sensitive by adding ` { sensitivity: 'case'}` as a parameter - but I don't really understand what the ecma standard means, I think it just says "use the host default locale" - http://www.ecma-international.org/ecma-262/5.1/#sec-15.5.4.9

src/NuGetGallery/Views/Users/Packages.cshtml
rocklan: Ok I have put in sorting via a "List position" variable that I've named `VersionSortOrder`. There's probably a better way of doing what I've done, see my comments under the checkin

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -381,9 +382,16 @@
                 }
             }
             function getCellValue(row, index) {
-                var v = $(row).children('td').eq(index).text();
+                var td = $(row).children('td').eq(index);
+                var sortby = td.data('sortby');
```
rocklan, line 386: If we want to sort columns by a method that's calculated server side, we can just add the "sortby" data attribute to those `<td>`'s in the future and it will pick them up.

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -168,7 +168,7 @@
                         <td class="align-middle text-nowrap">
                             <span data-bind="text: FormattedDownloadCount"></span>
                         </td>
-                        <td class="align-middle text-nowrap">
+                        <td class="align-middle text-nowrap" data-bind="attr: { 'data-sortby': VersionSortOrder }">
```
rocklan, line 171: and I thought all of my years of using Knockout was useless knowledge :)

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -567,6 +564,34 @@ public virtual ActionResult Packages()
...
+                .Where(p => predicate(p))
+                .Select(p => _listPackageItemRequiredSignerViewModelFactory.Create(p, currentUser, wasAADLoginOrMultiFactorAuthenticated))
+                .OrderBy(p => NuGetVersion.Parse(p.FullVersion))
+                .ToList();
+
+            for (int i = 0; i < listedPackages.Count; i++)
+                listedPackages[i].VersionSortOrder = i;
+
+            listedPackages.Sort((x, y) => x.Id.CompareTo(y.Id));
+
```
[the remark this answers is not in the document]
joelverhagen: This is fine. We go from one server-side sort to two. I check and it looks like the `OrderBy` occurs after `ToList` so even before the sort was not happening in SQL.

Only comment is to switch the `CompareTo` to use `OrdinalIgnoreCare` comparison (which it didn't do before, but should have). All ID operations should be performed in a case-insensitive manner.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -567,6 +564,34 @@ public virtual ActionResult Packages()
...
+            Func<Package, bool> predicate)
+        {
+            var listedPackages = packages
+                .Where(p => predicate(p))
+                .Select(p => _listPackageItemRequiredSignerViewModelFactory.Create(p, currentUser, wasAADLoginOrMultiFactorAuthenticated))
+                .OrderBy(p => NuGetVersion.Parse(p.FullVersion))
+                .ToList();
+
+            for (int i = 0; i < listedPackages.Count; i++)
+                listedPackages[i].VersionSortOrder = i;
```
joelverhagen, line 588: Our coding style in both C# and JavaScript is explicit curly braces even if the block is a single line.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -355,6 +356,46 @@
...
+                    var valA = getCellValue(a, index), valB = getCellValue(b, index)
+                    return $.isNumeric(valA) && $.isNumeric(valB) ? valB - valA : valA.toString().localeCompare(valB)
+                }
+            }
+            function getCellValue(row, index) {
+                var td = $(row).children('td').eq(index);
+                var sortby = td.data('sortby');
+
+                if (sortby)
+                    return sortby;
```
joelverhagen, line 389: Explicit curly braces.

joelverhagen:
Hey @jcjiang, could you take a look and provide some feedback from the PM perspective?

src/NuGetGallery/Views/Users/Packages.cshtml
joelverhagen: I think the implementation is fine. What sort of improvements were you thinking?

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
joelverhagen: Okay, I'll test it on DEV environment and make sure everything's working okay.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
joelverhagen: Sounds good! Could you put a short code comment at this line describing what you've said here?

joelverhagen:
Hey, it looks like this test is failing on your branch (per CI):
````
##[error]    NuGetGallery.ApiControllerFacts+TheGetStatsDownloadsAction.VerifyRecentPopularityStatsDownloads [FAIL]

      unexpected content result[3].Gallery
      Expected: True
      Actual:   False
      Stack Trace:
        tests\NuGetGallery.Facts\Controllers\ApiControllerFacts.cs(0,0): at NuGetGallery.ApiControllerFacts.TheGetStatsDownloadsAction.<VerifyRecentPopularityStatsDownloads>d__0.MoveNext()
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

src/NuGetGallery/Controllers/UsersController.cs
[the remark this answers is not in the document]
rocklan: done

src/NuGetGallery/Controllers/UsersController.cs
rocklan: fixed

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
rocklan: fixed

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
rocklan: done

rocklan:
> VerifyRecentPopularityStatsDownloads

This test and all other tests pass on my machine, not sure why this one failed. It doesn't seem related to the changes that I've made. Maybe the build can just be kicked off again?

src/NuGetGallery/Views/Users/Packages.cshtml
rocklan: I was just referring to the sorting of the data on the server-side to calculate the VersionSortOrder. That's cool all good.

joelverhagen:
Already on the second build. Let me try locally and I'll get back to you. This line was suspicious to me because 1.1 is a non-normalized version:
https://github.com/NuGet/NuGetGallery/blob/b3c174ecc7affadf411d88270a50ed4b5eb1f2a4/tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs#L2859

joelverhagen:
Okay, I am crazy. I crossed my wires with your other PR https://github.com/NuGet/NuGetGallery/pull/8157. Sorry about the confusion. I'll copy my comment over there for clarity.

rocklan:
I've changed the test so that when failing it's a bit more obvious why. It's now:

```
Assert.Equal("/packages/B/1.1", (string)result[3]["Gallery"]);
Assert.Equal(5, (int)result[2]["Downloads"]);
```

And when running the test we get:

image

So I can see what's going on now. The link has now got a normalized version number. Instead of 1.1 it's normalising it to 1.1.0. I'm not sure if that's a genuine problem or if the test can just be changed to:

```
Assert.Equal("/packages/B/1.1.0", (string)result[3]["Gallery"]);
```

joelverhagen:
The test should be changed to be `1.1.0` but of course on the other PR (https://github.com/NuGet/NuGetGallery/pull/8157).

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -355,6 +356,54 @@
...
+                var sortby = td.data('sortby');
+
+                if (sortby) {
+                    return sortby;
+                }
+
+                // grab the text inside the td. Remove commas in case it's a number like 43,000
+                var v = td.text();
+                if (v) {
+                    v = v.replace(',', '');
```
agr, line 401: This will break with locales that use something other than `,` for thousand separators (see `de-DE`, `de-CH` and `ru-RU` locales for example)

rocklan: I have commited a fix for this, so that it uses the unformatted number to sort by. It should work for all locales. I couldn't find the code that is forcing the numbers to US locale, I could only find calls to `toLocaleString()` which I think does use the browser's locale.

agr: I see now, it is rendered by javascript, I thought it came straight from Razor. I tested by bumping Russian to the top of accepted language list in the browser settings, that causes using Russian locale on server side (and I saw numbers rendered appropriately on other pages), but browser itself was still defaulting to US locale, so `toLocaleString` still used commas for thousand separator. Thanks for the insight and for the fix!

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -567,6 +564,36 @@ public virtual ActionResult Packages()
...
+                .Select(p => _listPackageItemRequiredSignerViewModelFactory.Create(p, currentUser, wasAADLoginOrMultiFactorAuthenticated))
+                .OrderBy(p => NuGetVersion.Parse(p.FullVersion))
+                .ToList();
+
+            for (int i = 0; i < listedPackages.Count; i++)
+            {
+                listedPackages[i].VersionSortOrder = i;
+            }
+
+            listedPackages.Sort((x, y) => string.Compare(x.Id,y.Id, StringComparison.OrdinalIgnoreCase));
```
loic-sharma, line 592: Nit:

```suggestion
            listedPackages.Sort((x, y) => string.Compare(x.Id, y.Id, StringComparison.OrdinalIgnoreCase));
```

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -567,6 +564,36 @@ public virtual ActionResult Packages()
...
+                .Select(p => _listPackageItemRequiredSignerViewModelFactory.Create(p, currentUser, wasAADLoginOrMultiFactorAuthenticated))
+                .OrderBy(p => NuGetVersion.Parse(p.FullVersion))
+                .ToList();
+
+            for (int i = 0; i < listedPackages.Count; i++)
+            {
+                listedPackages[i].VersionSortOrder = i;
+            }
+
+            listedPackages.Sort((x, y) => string.Compare(x.Id, y.Id, StringComparison.OrdinalIgnoreCase));
```
joelverhagen, line 592: Could we get a unit test asserting this behavior? We have two sort orders encoded: ID lex order and Version SemVer order.

joelverhagen:
I am going to deploy this to our test environment and to give it another spin. After the UT and this smoke test, we should be good to go 😄

loic-sharma (approved):

joelverhagen:
I'm seeing some strange behavior with packages with some package versions being out of order. For example:
image

Could you take a look and see what's going on with this bug?
[Packages](https://github.com/NuGet/NuGetGallery/files/5170476/Downloads.zip)

rocklan:
> I'm seeing some strange behavior with packages with some package versions being out of order

Ahha, found it. I was using the following javascript code to check if the "SortBy" attribute+value existed:

```
var sortby = td.data('sortby');

if (sortby) {
    return sortby;
}
```

so it turns out if the value of sortby is zero, it will return false, any other value it will return true :) as a result, the first item in the list will always appear last. 

An easy fix:

```
if (typeof sortby !== 'undefined') {
    return sortby;
}
```

which is the correct way of checking if the attribute has a defined value or not.

joelverhagen:
@rocklan, this should not be closed, right?

joelverhagen:
Okay great! Once we get that UT mentioned in my comment https://github.com/NuGet/NuGetGallery/pull/8158#discussion_r483102515 above, we'll be good to go! We've got PM sign off and several devs have looked it over.

src/NuGetGallery/Controllers/UsersController.cs
rocklan: done, see `PackagesAreSortedById` and `PackagesVersionSortOrderIsSetBySemVer` tests

joelverhagen (approved):
This PR is ready to rock(lan) 😉!
Thanks!

PR #8160, opened 2020-08-13
Preview readme 

lyndaidaii:
Summary of the changes:

* UI to display preview of readme content
* Disable documentation section at manage package page when embeddedReadme feature is enabled 

- EmbeddedReadme flag feature enable

    - 1. Uploading packages with embedded readme
 NuGet Gallery _ Upload Package (1)
     - 2. Uploading packages without embedded readme
legacyreadme

  

- EmbeddedReadme flag feature disable:

    - 1. Uploading packages with embedded readme: 
readme

    - 2.  Uploading packages without embedded readme:
legacyreadme

Addresses :https://github.com/NuGet/Engineering/issues/3267

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/PackageService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageService.cs
@@ -692,6 +692,7 @@ namespace NuGetGallery
             package.EmbeddedLicenseType = GetEmbeddedLicenseType(packageMetadata);
             package.LicenseExpression = GetLicenseExpression(packageMetadata);
             package.HasEmbeddedIcon = !string.IsNullOrWhiteSpace(packageMetadata.IconFile);
+            package.HasReadMe = !string.IsNullOrWhiteSpace(packageMetadata.ReadmeFile);
             package.EmbeddedReadmeType = GetEmbeddedReadmeType(packageMetadata);
 
             return package;
@@ -750,7 +751,7 @@ namespace NuGetGallery
 
             var extension = Path.GetExtension(packageMetadata.ReadmeFile);
 
-            if (MarkdownFileExtension.Equals(extension, StringComparison.OrdinalIgnoreCase) || string.Empty == extension)
+            if (MarkdownFileExtension.Equals(extension, StringComparison.OrdinalIgnoreCase))
             {
                 return EmbeddedReadmeFileType.Markdown;
             }
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -17,6 +17,7 @@ using System.Threading.Tasks;
 using System.Web;
 using System.Web.Caching;
 using System.Web.Mvc;
+using System.Windows.Forms;
 using NuGet.Packaging;
 using NuGet.Services.Entities;
 using NuGet.Services.Licenses;
@@ -343,6 +344,8 @@ namespace NuGetGallery
             verifyRequest.IsSymbolsPackage = false;
             verifyRequest.LicenseFileContents = await GetLicenseFileContentsOrNullAsync(packageMetadata, packageArchiveReader);
             verifyRequest.LicenseExpressionSegments = GetLicenseExpressionSegmentsOrNull(packageMetadata.LicenseMetadata);
+            verifyRequest.ReadmeFileContents = await GetReadmeFileContentsOrNullAsync(packageMetadata, packageArchiveReader);
+
             model.InProgressUpload = verifyRequest;
             return View(model);
         }
@@ -613,6 +616,7 @@ namespace NuGetGallery
             model.Warnings.AddRange(packageContentData.Warnings.Select(w => new JsonValidationMessage(w)));
             model.LicenseFileContents = packageContentData.LicenseFileContents;
             model.LicenseExpressionSegments = packageContentData.LicenseExpressionSegments;
+            model.ReadmeFileContents = packageContentData.ReadmeFileContents;
 
             if (packageContentData.EmbeddedIconInformation != null)
             {
@@ -635,13 +639,15 @@ namespace NuGetGallery
                 IReadOnlyList<IValidationMessage> warnings,
                 string licenseFileContents,
                 IReadOnlyCollection<CompositeLicenseExpressionSegmentViewModel> licenseExpressionSegments,
-                EmbeddedIconInformation embeddedIconInformation)
+                EmbeddedIconInformation embeddedIconInformation,
+                string readmeFileContents)
             {
                 PackageMetadata = packageMetadata;
                 Warnings = warnings;
                 LicenseFileContents = licenseFileContents;
                 LicenseExpressionSegments = licenseExpressionSegments;
                 EmbeddedIconInformation = embeddedIconInformation;
+                ReadmeFileContents = readmeFileContents;
             }
 
             public JsonResult ErrorResult { get; }
@@ -650,6 +656,7 @@ namespace NuGetGallery
             public string LicenseFileContents { get; }
             public IReadOnlyCollection<CompositeLicenseExpressionSegmentViewModel> LicenseExpressionSegments { get; }
             public EmbeddedIconInformation EmbeddedIconInformation { get; }
+            public string ReadmeFileContents { get; }
         }
 
         private class EmbeddedIconInformation
@@ -673,6 +680,7 @@ namespace NuGetGallery
             IReadOnlyCollection<CompositeLicenseExpressionSegmentViewModel> licenseExpressionSegments = null;
             PackageMetadata packageMetadata = null;
             EmbeddedIconInformation embeddedIconInformation = null;
+            string readmeFileContents = null;
 
             using (Stream uploadedFile = await _uploadFileService.GetUploadFileAsync(currentUser.Key))
             {
@@ -720,6 +728,7 @@ namespace NuGetGallery
                     licenseFileContents = await GetLicenseFileContentsOrNullAsync(packageMetadata, packageArchiveReader);
                     licenseExpressionSegments = GetLicenseExpressionSegmentsOrNull(packageMetadata.LicenseMetadata);
                     embeddedIconInformation = await GetEmbeddedIconOrNullAsync(packageMetadata, packageArchiveReader);
+                    readmeFileContents = await GetReadmeFileContentsOrNullAsync(packageMetadata, packageArchiveReader);          
                 }
                 catch (Exception ex)
                 {
@@ -730,7 +739,7 @@ namespace NuGetGallery
                 }
             }
 
-            return new PackageContentData(packageMetadata, warnings, licenseFileContents, licenseExpressionSegments, embeddedIconInformation);
+            return new PackageContentData(packageMetadata, warnings, licenseFileContents, licenseExpressionSegments, embeddedIconInformation, readmeFileContents);
         }
 
         private IReadOnlyCollection<CompositeLicenseExpressionSegmentViewModel> GetLicenseExpressionSegmentsOrNull(LicenseMetadata licenseMetadata)
@@ -786,6 +795,18 @@ namespace NuGetGallery
             }
 
             return new EmbeddedIconInformation(imageContentType, imageData);
+        }
+
+        private async Task<string> GetReadmeFileContentsOrNullAsync(PackageMetadata packageMetadata, PackageArchiveReader packageArchiveReader)
+        {
+            if (string.IsNullOrWhiteSpace(packageMetadata.ReadmeFile))
+            {
+                return null;
+            }
+
+            var readmeFilename = FileNameHelper.GetZipEntryPath(packageMetadata.ReadmeFile);
+            var readmeResult = await _readMeService.GetReadMeHtmlAsync(readmeFilename, packageArchiveReader, Encoding.UTF8);
+            return readmeResult?.Content;
         }
 
         private static async Task<byte[]> ReadPackageFile(PackageArchiveReader packageArchiveReader, string filename)
@@ -2600,9 +2621,14 @@ namespace NuGetGallery
                 {
                     return afterValidationJsonResult;
                 }
-
+                
                 if (formData.Edit != null)
                 {
+                    if (package.HasReadMe && package.EmbeddedReadmeType != EmbeddedReadmeFileType.Absent)
+                    {
+                        return Json(HttpStatusCode.BadRequest, new[] { new JsonValidationMessage(Strings.ReadmeNotEditableWithEmbeddedReadme) });
+                    }
+
                     try
                     {
                         if (await _readMeService.SaveReadMeMdIfChanged(
--- a/src/NuGetGallery/Scripts/gallery/async-file-upload.js
+++ b/src/NuGetGallery/Scripts/gallery/async-file-upload.js
@@ -255,6 +255,11 @@
                 $(reportContainerElement).attr("data-bind", "template: { name: 'verify-metadata-template', data: data }");
                 $("#verify-package-container").append(reportContainerElement);
                 ko.applyBindings({ data: model }, reportContainerElement);
+                if (model.ReadmeFileContents) {
+                    $('#import-readme-container').addClass('hidden');
+                } else {
+                    $('#import-readme-container').removeClass('hidden');
+                }
 
                 var submitContainerElement = document.createElement("div");
                 $(submitContainerElement).attr("id", "submit-block");
--- a/src/NuGetGallery/Services/IReadMeService.cs
+++ b/src/NuGetGallery/Services/IReadMeService.cs
@@ -3,6 +3,7 @@
 
 using System.Text;
 using System.Threading.Tasks;
+using NuGet.Packaging;
 using NuGet.Services.Entities;
 
 namespace NuGetGallery
@@ -29,6 +30,16 @@ namespace NuGetGallery
         /// <param name="package">Package entity associated with the ReadMe.</param>
         /// <returns>ReadMe converted to HTML.</returns>
         Task<RenderedReadMeResult> GetReadMeHtmlAsync(Package package);
+
+        /// <summary>
+        /// Get the converted HTML from the package with Readme markdown.
+        /// </summary>
+        /// <param name="readmeFileName">The path of Readme markdown.</param>
+        /// <param name="packageArchiveReader">
+        /// The <see cref="PackageArchiveReader"/> instance providing the package metadata.
+        /// </param>
+        /// <returns>ReadMe converted to HTML.</returns>
+        Task<RenderedReadMeResult> GetReadMeHtmlAsync(string readmeFileName, PackageArchiveReader packageArchiveReader, Encoding encoding);
 
         /// <summary>
         /// Get package ReadMe markdown from storage.
--- a/src/NuGetGallery/Services/ReadMeService.cs
+++ b/src/NuGetGallery/Services/ReadMeService.cs
@@ -11,6 +11,7 @@ using System.Threading.Tasks;
 using System.Web;
 using CommonMark;
 using CommonMark.Syntax;
+using NuGet.Packaging;
 using NuGet.Services.Entities;
 
 namespace NuGetGallery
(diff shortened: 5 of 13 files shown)


src/NuGetGallery/Services/IReadMeService.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Text;
 using System.Threading.Tasks;
+using NuGet.Packaging;
```
lyndaidaii, line 6: out of order

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -343,6 +343,9 @@ public virtual async Task<ActionResult> UploadPackage()
             verifyRequest.IsSymbolsPackage = false;
             verifyRequest.LicenseFileContents = await GetLicenseFileContentsOrNullAsync(packageMetadata, packageArchiveReader);
             verifyRequest.LicenseExpressionSegments = GetLicenseExpressionSegmentsOrNull(packageMetadata.LicenseMetadata);
+            verifyRequest.ReadmeFileContents = await GetReadmeFileContentsOrNullAsync(packageMetadata, packageArchiveReader);
+
+            model.AreEmbeddedReadmesUpload = verifyRequest.ReadmeFileContents != null;
```
joelverhagen, line 348: It's not clear to me what this variable name means. Is it "Are embedded readmes **uploaded**"?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -788,6 +797,19 @@ private static async Task<EmbeddedIconInformation> GetEmbeddedIconOrNullAsync(Pa
             return new EmbeddedIconInformation(imageContentType, imageData);
         }
 
+        public async Task<string> GetReadmeFileContentsOrNullAsync(PackageMetadata packageMetadata, PackageArchiveReader packageArchiveReader)
+        {
+
```
joelverhagen, line 802: nit: extra line

src/NuGetGallery/Views/Packages/_ImportReadMe.cshtml
```diff
@@ -23,8 +23,8 @@
     </div>
 }
 
-<script type="text/html" id="import-readme-template">
-
+@helper ImportReadmeTemplate()
```
joelverhagen, line 26: What's the significance of this change? It seems like we introduced a helper used in exactly one place (unless I am missing) something. It may be simpler just to leave the contents of the template inline without the helper.

src/NuGetGallery/Controllers/PackagesController.cs
lyndaidaii: Yes, it indicates whether embedded readmes within package

src/NuGetGallery/Views/Packages/_ImportReadMe.cshtml
lyndaidaii: I will revert this change

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -255,6 +255,11 @@
                 $(reportContainerElement).attr("data-bind", "template: { name: 'verify-metadata-template', data: data }");
                 $("#verify-package-container").append(reportContainerElement);
                 ko.applyBindings({ data: model }, reportContainerElement);
+                if (model.ReadmeFileContents) {
```
joelverhagen, line 258: This can be fiddled with by a clever client. What happens when the feature flag is enabled, embedded readme is in the uploaded package, but the confirm POST message also includes something from the `#import-readme-container`? _My assumption would be that either the embedded readme is preferred or the request is entirely rejected since it's invalid._

In short hiding something on the client side does not strictly prevent a malicious or buggy client from submitting the hidden form/inputs anyway.

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: Perhaps `AreEmbeddedReadmesUploaded` is a clearer name then

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -788,6 +797,19 @@ private static async Task<EmbeddedIconInformation> GetEmbeddedIconOrNullAsync(Pa
             return new EmbeddedIconInformation(imageContentType, imageData);
         }
 
+        public async Task<string> GetReadmeFileContentsOrNullAsync(PackageMetadata packageMetadata, PackageArchiveReader packageArchiveReader)
```
cristinamanum, line 800: Does this method need to be public?

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -98,6 +99,28 @@ public async Task<RenderedReadMeResult> GetReadMeHtmlAsync(Package package)
...
+        /// </summary>
+        /// <param name="readmeFileName">The path of Readme markdown.</param>
+        /// <param name="packageArchiveReader">
+        /// The <see cref="PackageArchiveReader"/> instance providing the package metadata.
+        /// </param>
+        /// <returns>ReadMe converted to HTML.</returns>
+        public async Task<RenderedReadMeResult> GetReadMeHtmlAsync(string readmeFileName, PackageArchiveReader packageArchiveReader)
+        {
+            var readmeMd = await GetReadMeMdAsync(readmeFileName, packageArchiveReader);
+            var result = new RenderedReadMeResult
```
cristinamanum, line 113: From the logic below, the result does not need to be created if not needed.

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -113,6 +136,23 @@ public async Task<string> GetReadMeMdAsync(Package package)
...
 
+        /// <summary>
+        /// Get content of Readme markdown.
+        /// </summary>
+        /// <param name="readmeFileName">Package entity associated with the ReadMe.</param>
+        /// <param name="packageArchiveReader">
+        /// The <see cref="PackageArchiveReader"/> instance providing the package metadata.
+        /// </param>
+        /// <returns>ReadMe markdown from storage.</returns>
+        public async Task<string> GetReadMeMdAsync(string readmeFileName, PackageArchiveReader packageArchiveReader)
```
cristinamanum, line 147: Is public needed?

src/NuGetGallery/Services/ReadMeService.cs
lyndaidaii: Followed the same logic of License and Icon

src/NuGetGallery/Scripts/gallery/async-file-upload.js
lyndaidaii: If this is case, second solution will be move entire  "_importReadMe.cshtml" to "_verifyMetadata.cshtml",  if readmecontents are null, skip the script of "import-readme-template". 
Reason why we need move things around is the data model of "_importReadMe.cshtml" is different than "_verifyMetadata.cshtml".  In "_imoportReadMe.cshtml", we don't have readme content information from Client side, which cause the an issue that need refresh the page to update the UI page  if we introduce variable from backend.

What do you think of second approach? It will be more secure that block the readme request

lyndaidaii: I could commit the change for second solution if my explanation is not clear

joelverhagen: We should block on the server side (i.e. in a controller action, C#) when unexpected data is provided.

joelverhagen: It's okay to have convenient code like this on the client side (in JavaScript) since it's just easier to write like this, but we should not use it as a guarantee and should always validate untrusted (customer) inputs on the server-side.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -720,6 +728,7 @@ private async Task<PackageContentData> ValidateAndProcessPackageContents(User cu
                     licenseFileContents = await GetLicenseFileContentsOrNullAsync(packageMetadata, packageArchiveReader);
                     licenseExpressionSegments = GetLicenseExpressionSegmentsOrNull(packageMetadata.LicenseMetadata);
                     embeddedIconInformation = await GetEmbeddedIconOrNullAsync(packageMetadata, packageArchiveReader);
+                    embeddedReadmeFileContents = await GetReadmeFileContentsOrNullAsync(packageMetadata, packageArchiveReader);
```
zhhyu, line 732: >GetReadmeFileContentsOrNullAsync [](start = 55, length = 32)

nit: maybe it's better to align with the license file as "readmeFileContents". The icon one is an object, while this is a string.

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -113,6 +136,23 @@ public async Task<string> GetReadMeMdAsync(Package package)
...
+        /// </summary>
+        /// <param name="readmeFileName">Package entity associated with the ReadMe.</param>
+        /// <param name="packageArchiveReader">
+        /// The <see cref="PackageArchiveReader"/> instance providing the package metadata.
+        /// </param>
+        /// <returns>ReadMe markdown from storage.</returns>
+        private async Task<string> GetReadMeMdAsync(string readmeFileName, PackageArchiveReader packageArchiveReader)
+        {
+            using (var readmeFileStream = packageArchiveReader.GetStream(readmeFileName))
+            using (var streamReader = new StreamReader(readmeFileStream, Encoding.UTF8))
```
zhhyu, line 150: This one should not be hard coded here as "Encoding.UTF8". You should provide the flexibility for the caller.

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -98,6 +99,28 @@ public async Task<RenderedReadMeResult> GetReadMeHtmlAsync(Package package)
                 GetReadMeHtml(readMeMd);
         }
 
+        /// <summary>
+        /// Get the converted HTML from the stored Readme markdown.
```
zhhyu, line 103: The comment here is misleading. Is this from the stored location?

src/NuGetGallery/Services/IReadMeService.cs
```diff
@@ -30,6 +31,16 @@ public interface IReadMeService
         /// <returns>ReadMe converted to HTML.</returns>
         Task<RenderedReadMeResult> GetReadMeHtmlAsync(Package package);
 
+        /// <summary>
+        /// Get the converted HTML from the stored Readme markdown.
```
zhhyu, line 35: The comment here is misleading. Is this from the stored location?

src/NuGetGallery/Services/IReadMeService.cs
```diff
@@ -30,6 +31,16 @@ public interface IReadMeService
...
+        /// </summary>
+        /// <param name="readmeFileName">The path of Readme markdown.</param>
```
zhhyu, line 37: Does this mean the name of the readme file or the path which points to the location?

src/NuGetGallery/Services/ReadMeService.cs
zhhyu: Could you please give more explanations here? I am not very clear about the same logic here.

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -788,6 +797,18 @@ private static async Task<EmbeddedIconInformation> GetEmbeddedIconOrNullAsync(Pa
...
+        private async Task<string> GetReadmeFileContentsOrNullAsync(PackageMetadata packageMetadata, PackageArchiveReader packageArchiveReader)
+        {
+            if (string.IsNullOrWhiteSpace(packageMetadata.ReadmeFile))
+            {
+                return null;
+            }
+
+            var readmeFilename = FileNameHelper.GetZipEntryPath(packageMetadata.ReadmeFile);
+            var readmeResult = await _readMeService.GetReadMeHtmlAsync(readmeFilename, packageArchiveReader);
+            return readmeResult?.Content;
```
zhhyu, line 810: Here you only return the content of "readmeResult", so where will we use the "ImagesRewritten"? How does this "ImagesRewritten" affect the front-end part like others?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -2594,9 +2616,14 @@ public virtual async Task<JsonResult> VerifyPackage(VerifyPackageRequest formDat
                 {
                     return afterValidationJsonResult;
                 }
-
+                
                 if (formData.Edit != null)
                 {
+                    if (_featureFlagService.AreEmbeddedReadmesEnabled(currentUser) && formData.ReadmeFileContents != null)
```
zhhyu, line 2622: I don't think this is correct. 
Even if we allow the user to upload the embedded readme file, we should still allow him to edit the readme for a uploading package without the embedded readme file. Feature flag should not block this usage here.

lyndaidaii: This is just extra validation to prevent the case where people upload embedded readmefile, meanwhile, try to use the legacy readme to add readme file, which it not allowed.

lyndaidaii: The front, we hide the edit documentation session, here just double validation if there is case where user are using legacy and new embedded file at same time

src/NuGetGallery/Services/IReadMeService.cs
lyndaidaii: relative path of readme file

src/NuGetGallery/Controllers/PackagesController.cs
lyndaidaii: return content of "readmeResult" will be sufficient for preview.  In preview, it will display the readme file content, ImagesRewritten comes into picture when readme file store in storage and display readme from stored storage.

src/NuGetGallery/Controllers/PackagesController.cs
joelverhagen: Don't we also need to check here whether or not the package has an embedded readme? For example, if the feature flag is enabled for the current user but they have no embedded readme in the package, then they are still allows to specify a legacy readme in the upload page?

Perhaps it's worth thinking through all combinations of the boolean variables.

1. Is feature flag enabled for the user?
1. Is there an embedded readme in the package?
1. Did the user provide a legacy readme on the verify upload page?

For example, if the feature flag is **enabled**, there **no embedded readme**, and **a legacy readme is provided**, then the package should be accepted, right?

lyndaidaii: Yes, if the feature flag is enabled, there no embedded readme, and a legacy readme is provided, then the package should be accepted. 

This condition should include those combinations.

1. if the feature flag is enabled for the current user but they have embedded readme in the package, they also specify a legacy readme in the upload page, we will reject package and have error message
2. if the feature flag is enabled, there no embedded readme, and a legacy readme is provided, then the package will accepted
3. if feature flag is disable, it will accept legacy readme

src/NuGetGallery/Services/ReadMeService.cs
lyndaidaii: I think I misunderstood here, I think Cristina said if only use the Content of readmeMd, not user ImagesRewritten, probably don't create the var result?

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -2594,9 +2616,14 @@ public virtual async Task<JsonResult> VerifyPackage(VerifyPackageRequest formDat
                 {
                     return afterValidationJsonResult;
                 }
-
+                
                 if (formData.Edit != null)
                 {
+                    if (package.HasReadMe && package.EmbeddedReadmeType != EmbeddedReadmeFileType.Absent)
```
lyndaidaii, line 2622: formData.Edit!= null means there is legacy readme provided. 

We will check if package has embedded readme uploaded will be sufficient enough with no matter feature flag is enable or disable, for example if package has embedded readme, and also legacy readme provided, we will reject the package with error message. if package no embedded readme, we will accept add legacy readme whenever feature flag is enable or disable.  for the case like feature flag is disable, there is embedded readme, it should be caught  by validation in early steps. I think there is no necessary to check feature flag here.

src/NuGetGallery/Controllers/PackagesController.cs
lyndaidaii: Created  a ticket to add more future work for the images behavior

src/NuGetGallery/Strings.resx
```diff
@@ -1178,4 +1178,7 @@ The {1} Team</value>
   <data name="UploadPackage_InvalidReadmeName" xml:space="preserve">
     <value>the name of &lt;readme&gt; element is case sensitive, must use the &lt;readme&gt;</value>
   </data>
+  <data name="ReadMeInvalidWithEmbeddedReadme" xml:space="preserve">
+    <value>Documentation is not allowed with the package has embedded readme </value>
+  </data>
```
zhhyu, line 1183: with the package which has the embedded readme file.

src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -692,6 +692,7 @@ private Package CreatePackageFromNuGetPackage(PackageRegistration packageRegistr
             package.EmbeddedLicenseType = GetEmbeddedLicenseType(packageMetadata);
             package.LicenseExpression = GetLicenseExpression(packageMetadata);
             package.HasEmbeddedIcon = !string.IsNullOrWhiteSpace(packageMetadata.IconFile);
+            package.HasReadMe = !string.IsNullOrWhiteSpace(packageMetadata.ReadmeFile);
             package.EmbeddedReadmeType = GetEmbeddedReadmeType(packageMetadata);
```
zhhyu, line 696: Maybe add some test coverage like icons.

zhhyu:
>                 return EmbeddedReadmeFileType.Markdown;

Do we allow the extension to be empty?

---
Refers to: src/NuGetGallery.Services/PackageManagement/PackageService.cs:756 in 94d7484. [](commit_id = 94d7484bebab46130e95f934c4da39b708537fd9, deletion_comment = False)

src/NuGetGallery/Services/ReadMeService.cs
```diff
@@ -113,6 +136,23 @@ public async Task<string> GetReadMeMdAsync(Package package)
...
+        /// <param name="packageArchiveReader">
+        /// The <see cref="PackageArchiveReader"/> instance providing the package metadata.
+        /// </param>
+        /// <returns>ReadMe markdown from package metadata.</returns>
+        private async Task<string> GetReadMeMdAsync(string readmeFileName, PackageArchiveReader packageArchiveReader, Encoding encoding)
+        {
+            using (var readmeFileStream = packageArchiveReader.GetStream(readmeFileName))
+            using (var streamReader = new StreamReader(readmeFileStream, encoding))
+            {
+                 return await streamReader.ReadToEndAsync();
```
zhhyu, line 152: I think it's fine to use this method to read the content as we have validated the size of the embedded readme file before reaching here.
We also have a "ReadMaxAsync" method with the limitation of "MaxMdLengthBytes". We may prefer to depend on that, but there will be a discrepancy as the size limitation during the validation is bigger. 
We should create a task to track and realize this issue, and as we will finally deprecate the edited readme, this "MaxMdLengthBytes" will go away finally.

src/NuGetGallery/Strings.resx
```diff
@@ -1178,4 +1178,7 @@ The {1} Team</value>
   <data name="UploadPackage_InvalidReadmeName" xml:space="preserve">
     <value>the name of &lt;readme&gt; element is case sensitive, must use the &lt;readme&gt;</value>
   </data>
+  <data name="ReadMeInvalidWithEmbeddedReadme" xml:space="preserve">
```
zhhyu, line 1181: Maybe this is better: ReadmeNotEditableWithEmbeddedReadme

zhhyu:
Consider adding some test coverage such as the "PackagesController", if possible.

lyndaidaii:
> > ```
> >             return EmbeddedReadmeFileType.Markdown;
> > ```
> 
> Do we allow the extension to be empty? How does the client verify this?
> 
> Refers to: src/NuGetGallery.Services/PackageManagement/PackageService.cs:756 in [94d7484](https://github.com/NuGet/NuGetGallery/commit/94d7484bebab46130e95f934c4da39b708537fd9). [](commit_id = [94d7484](https://github.com/NuGet/NuGetGallery/commit/94d7484bebab46130e95f934c4da39b708537fd9), deletion_comment = False)

I think here probably better not allow extension to be empty for markdown file. I will check with Client and keep consistent

src/NuGetGallery/Services/ReadMeService.cs
lyndaidaii: Ticket is created to track this issue

PR #8161, opened 2020-08-17
Add hijack=false parameter to OData get specific package

joelverhagen:
This will facilitate functional tests and end-to-end tests that require querying for package availability in DB instead of availability in V3 (search hijack). See the functional tests containing `hijack=true` for context. There is also one such case in the end-to-end tests. In short, there are many integration tests today that depend on checking for a package's availability prior to V3 indexing.

Essentially the difference when the new parameter is specified as `hijack=false`, you can see a package even before it has made its way the search service but _after_ validation is complete. This was previously achieved in various monitoring and testing places in our code via opaque hacks like adding an `$orderby` that can't be hijacked or adding `1 eq 1` to a `$filter`. 

Given the excellent performance of an ID+version lookup, it is acceptable to allow this non-hijacked query from a performance perspective. By the way, the package details page does this SQL query and much more.

An alternative approach could have been to carefully scrutinize each place that uses `hijack=false` and convince ourselves that moving from a non-hijacked to a hijacked query is acceptable from both a test coverage and flakiness perspective. My feeling is that this is not worth the risk. Also, in general we would be giving up from a testing/monitoring perspective on observing the state before a package is fully indexed in search but after it is validated.

I think that this PR (simple test hook) is better than the alternative. I think it is reasonable to have a programmatic way of fetching the DB state of a package and this is the simplest approach given our current various codebases. This will be undocumented on our public API surface area and not intended for external consumption, much like the internal `/search/query` endpoint on the search service.

Progress on https://github.com/NuGet/Engineering/issues/2902.

### DIFF
--- a/src/NuGetGallery/Infrastructure/ODataCacheOutputAttribute.cs
+++ b/src/NuGetGallery/Infrastructure/ODataCacheOutputAttribute.cs
@@ -40,6 +40,22 @@ namespace NuGetGallery
             ServerTimeSpan = serverTimeSpan;
         }
 
+        protected override bool IsCachingAllowed(HttpActionContext actionContext, bool anonymousOnly)
+        {
+            if (_endpoint == ODataCachedEndpoint.GetSpecificPackage)
+            {
+                // Don't cache the non-hijacked ID+version lookup pattern.
+                if (actionContext.ActionArguments.TryGetValue("hijack", out var hijackObj)
+                    && hijackObj is bool hijack
+                    && !hijack)
+                {
+                    return false;
+                }
+            }
+
+            return base.IsCachingAllowed(actionContext, anonymousOnly);
+        }
+
         /// <summary>
         /// This setting determines how frequently the cache duration setting should be checked. Since the OData
         /// endpoints get a lot of traffic, we don't want to check every time. The 60 seconds here is picked to mimic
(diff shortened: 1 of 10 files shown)


src/NuGetGallery/Infrastructure/ODataCacheOutputAttribute.cs
```diff
@@ -40,6 +40,21 @@ public ODataCacheOutputAttribute(ODataCachedEndpoint endpoint, int serverTimeSpa
             ServerTimeSpan = serverTimeSpan;
         }
 
+        protected override bool IsCachingAllowed(HttpActionContext actionContext, bool anonymousOnly)
+        {
+            if (_endpoint == ODataCachedEndpoint.GetSpecificPackage)
+            {
+                // Don't cache the non-hijacked ID+version lookup pattern.
+                if (actionContext.ActionArguments.TryGetValue("hijack", out var hijackObj)
+                    && hijackObj is bool hijack)
```
loic-sharma, line 49: It seems the request can "force" caching by adding `hijack=true`. Should this include an `&& hijack == false`?

joelverhagen: That's a good point on deferring to the base implementation when hijack is true. The kind of queries we care about here should always have true returned from the base but I think it's good to led the code cleanly defer to the base as you described.

joelverhagen:
Tracked by https://github.com/NuGet/Engineering/issues/3329.

Issue #8152, opened 2020-08-11
[A11y]Reset button for advanced search filtering does not have sufficient contrast


ryuyu:
![image](https://user-images.githubusercontent.com/11051729/89946316-94da2f80-dbd7-11ea-82c5-21c0a815589a.png)
Title: WCAG 1.4.3: Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds (#reset-advanced-search)
Tags: Accessibility, WCAG 1.4.3, color-contrast

Issue: Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds (color-contrast - https://dequeuniversity.com/rules/axe/3.5/color-contrast?application=msftAI)

Target application: NuGet Gallery | Packages - https://www.nuget.org/packages

Element path: #reset-advanced-search

Snippet: <input id="reset-advanced-search" class="btn btn-link form-control" type="button" value="Reset">

How to fix: 
Fix any of the following:
  Element has insufficient color contrast of 4.14 (foreground color: #337ab7, background color: #f4f4f4, font size: 10.5pt (14px), font weight: normal). Expected contrast ratio of 4.5:1

Environment: Chrome version 84.0.4147.125

====

This accessibility issue was found using Accessibility Insights for Web 2.20.0 (axe-core 3.5.1), a tool that helps find and fix accessibility issues. Get more information & download this tool at http://aka.ms/AccessibilityInsights.

Issue #8091, opened 2020-07-07
[A11y] Color contrast issues on package details page, links on the selected version


joelverhagen:
Issue: Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds (color-contrast - https://dequeuniversity.com/rules/axe/3.5/color-contrast?application=msftAI)

Target application: `NuGet Gallery | Newtonsoft.Json 12.0.3 - https://www.nuget.org/packages/Newtonsoft.Json/`

Element paths: `a[title="\31 2\.0\.3"]`

Snippet: `<a href="/packages/Newtonsoft.Json/12.0.3" title="12.0.3"> 12.0.3 </a>`

Element path: `.bg-info > td:nth-child(4) > a`

Snippet: `<a href="/packages/Newtonsoft.Json/12.0.3/Manage">Listed</a>`

![image](https://user-images.githubusercontent.com/94054/86853070-af663980-c06a-11ea-9f94-4ab49b43c88a.png)

How to fix: 
Fix any of the following:
 Element has insufficient color contrast of 3.77 (foreground color: #337ab7, background color: #d9edf7, font size: 12.0pt (16px), font weight: normal). Expected contrast ratio of 4.5:1

Environment: Chrome version 83.0.4103.116

ryuyu:
It looks like changing the active row color for links to either not be a link (black) or to use the hovered value (#23527c) is sufficient contrast.

Issue #7847, opened 2020-02-15
Accessibility: Color-Contrast too low on package page


robertmuehsig:
### Describe the bug
I played around with [Accessibility Insights](https://accessibilityinsights.io/) and tried it on NuGet.org. The site seems (at least from the "Fast pass") quite good regarding accessibility, but the color on the package page for the currently released package is too low.

### To Reproduce
1. On this URL 'https://www.nuget.org/packages/Newtonsoft.Json/'
2. Perform this operation 'FastPass with https://accessibilityinsights.io/'
3. See error

### Expected behavior
"Congratulations!
No failed automated checks were found."

### Screenshots
![image](https://user-images.githubusercontent.com/756703/74595115-19d7f800-503e-11ea-8730-fb319a456680.png)

### Additional context
The link color and the background color (this blue-ish color) are problematic. I'm not sure if this combination is used on other pages as well. I would guess that the best fix would be to change the background color of the "bg-info" class. 

I could try to find a color combination, but I guess there is not "right" or "wrong", but if you would like that I should try something just give me a comment here. :)

skofman1:
Thanks for reporting! Adding this work item to the WCAG 2.1 epic.

ryuyu:
I believe this is now fixed!
Thanks for the report!

PR #8164, opened 2020-08-19
[A11y]Fix some minor text contrast issues.

ryuyu:
Addresses the following:
https://github.com/NuGet/NuGetGallery/issues/7847
https://github.com/NuGet/NuGetGallery/issues/8091
https://github.com/NuGet/NuGetGallery/issues/8152

### Old Package details:
image

### New Package details:
image

Selected row is slightly darkened.

### Old Advanced Search:
image

### New Advanced Search:
image

Reset Button text slightly darkened.

### DIFF
--- a/src/Bootstrap/less/theme/page-list-packages.less
+++ b/src/Bootstrap/less/theme/page-list-packages.less
@@ -83,6 +83,10 @@
       margin-left: 10px;
       margin-right: 10px;
     }
+
+    #reset-advanced-search {
+      color: darken(@link-color, 10%);
+    }
   }
 
   @media (min-width: @screen-sm) {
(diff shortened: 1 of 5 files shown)


src/Bootstrap/less/theme/page-list-packages.less
```diff
@@ -83,6 +83,10 @@
       margin-left: 10px;
       margin-right: 10px;
     }
+
+    #reset-advanced-search {
+      color: darken(@link-color, 10%);
```
joelverhagen, line 88: I wonder if we have this `@link-color` + `@alert-info-bg` interaction anywhere. For example, and link in an info alert would have the same problem right?

ryuyu: That's a good point. I did a quick look through for bg-info class, and it should be fixed for all of those cases, but I'll do a quick sweep for any other cases.

PR #8168, opened 2020-08-20
Move to Knockout 3.5.1 and fix admin JS errors

joelverhagen:
We have unused Knockout NuGet dependencies. We do not pull in Knockout in this way. They are just files checked in to source control.

The admin errors are because `error` is no longer the syntax for jQuery AJAX errors. It is `fail`. This was likely from the jQuery update we did a while back but went undetected.

Address https://github.com/NuGet/Engineering/issues/3311


joelverhagen:
> Nice fixes! Do you think it's worth doing a bug bash on DEV with the team?

Probably not worth randomizing people. Should be easy enough to exercise all of the pages that use `ko.*`. I've been working on that myself.

PR #8178, opened 2020-08-31
Add feature flags to all OData endpoints

joelverhagen:
This allows us to individually disable non-hijacked endpoints. These feature flags are enabled by default, which is not following the existing pattern but is desirable in this case. We're turning OFF a feature, not ON which is what feature flags have been used for thus far.  It's easy to achieve this with our framework.

Progress on https://github.com/NuGet/Engineering/issues/2902.


src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -185,13 +205,18 @@ private async Task<IHttpActionResult> GetCore(ODataQueryOptions<V1FeedPackage> o
...
             }
-            catch (Exception ex)
+            catch (Exception ex) when (isNonHijackEnabled)
             {
                 // Swallowing Exception intentionally. If *anything* goes wrong in search, just fall back to the database.
                 // We don't want to break package restores. We do want to know if this happens, so here goes:
                 QuietLog.LogHandledException(ex);
             }
 
+            if (!isNonHijackEnabled)
```
loic-sharma, line 215: At first glance I thought "Shouldn't we return a `BadRequest` immediately at the beginning of the method"? Consider adding a quick comment that we attempted to hijack, had to fallback to db, but fallback is disabled.

src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -185,13 +205,18 @@ private async Task<IHttpActionResult> GetCore(ODataQueryOptions<V1FeedPackage> o
                     customQuery = true;
                 }
             }
-            catch (Exception ex)
+            catch (Exception ex) when (isNonHijackEnabled)
```
loic-sharma, line 208: I'm not sure I understand this. Do we want to throw an exception if the flag is disabled? Who catches the exception? Maybe add a comment if this is intentional.

loic-sharma: Also, do you know if this exception is covered by a test?

joelverhagen: If our only dependency fails (the case when the FF is disabled), we should 500.

loic-sharma: Ahh I see, that makes sense. I would add a quick comment for folks like me that aren't very fast on the uptake 😂

src/NuGetGallery/Controllers/ODataV1FeedController.cs
joelverhagen: Yup, sounds good.

src/NuGetGallery/Controllers/ODataV1FeedController.cs
joelverhagen: Done. More comments are good :)

PR #8179, opened 2020-09-01
[SAS Token] CloudBlobClientWrapper update to support SAS Tokens.

dannyjdev:


src/NuGetGallery.Core/Services/CloudBlobClientWrapper.cs
```diff
@@ -2,30 +2,55 @@
...
-            _defaultRequestOptions = defaultRequestOptions;
+            _cloudBlobClient = CloudBlobClientFactory.CreateCloudBlobClient(storageConnectionString);
+
+            if (defaultRequestOptions != null)
+            {
+                _cloudBlobClient.DefaultRequestOptions = defaultRequestOptions;
+            }
+        }
+
+        public CloudBlobClientWrapper(string sasToken, string storageAccountName, bool readAccessGeoRedundant)
```
skofman1, line 36: >string sasToken, string storageAccountName [](start = 38, length = 42)

why do we need to have special ctors for SAS tokens? Can we continue using a connection string, just instead of storage key it will have a SAS token? I see on portal that it's possible (I created a SAS token for a storage account, and one of the options is to copy a connection string

joelverhagen: Yeah, ideally we keep the storage connection string as a single secret that includes the SAS token, storage account name, and storage suffix. My understanding is that we are essentially swapping out the shared key with a SAS token and changing one (or a couple) property names in the connection string accordingly.

Issue #8181, opened 2020-09-03
[A11y][UI]No Visual representation is provided for required field 'Username' 


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1184475
## Browser Details:
Firefox Version: 79.0 (64 bit)

## OS Details

Microsoft Windows 10 Enterprise  ​

Version 2004 (OS Build 19041.450)

## Repro Steps:

    Hit the URL and login with appropriate credentials
    Tab till "Open Profile drop down" and press enter
    Tab till "API Keys" and Press enter
    Tab till "Create button" and press enter
    Tab till "Key Name" editable field
    Check whether a visual representation like asterisk is provided at the required field "Key Name" or not 

## Actual Results:

No Visual representation is provided for required field 'Username'
## Expected Results:
Visual representation like asterisk should be provided at the required fields.
## User Impact:
Users may not understand if there is no visual representation at the required fields.

## Notes:
This appears to be about adding an * by required fields in forms.

PR #8207, opened 2020-09-14
adding an * by required fields in forms

lyndaidaii:
Adding an * by required fields for Key Name

Before changes:
UI-d

After changes: 
Ui_LI
Addresses https://github.com/NuGet/NuGetGallery/issues/8181

### DIFF
--- a/src/Bootstrap/less/theme/page-api-keys.less
+++ b/src/Bootstrap/less/theme/page-api-keys.less
@@ -53,6 +53,11 @@
         height: 282px;
       }
     }
+
+    .required:after {
+      color: red;
+      content: " *";
+    }
   }
 
   .add-api-key {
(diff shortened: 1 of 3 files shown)


src/Bootstrap/less/theme/page-api-keys.less
```diff
@@ -53,6 +53,11 @@
         height: 282px;
       }
     }
+
+    .required:after{
```
joelverhagen, line 57: image

joelverhagen:
Would it be possible to have a little more space before the asterisk?

Also, how does the screen reader behave for this case? /cc @ryuyu Is it correct?

ryuyu:
As far as I can tell, it looks like what is expected. Let's make sure that what screen reader reads makes sense, and I think we are good.

joelverhagen (approved):
Address comments then :shipit:

loic-sharma:
+1 for adding space, it looks a little compact to me currently.

loic-sharma (approved):

lyndaidaii:
screen reader reads 'the key name is required' when we click the input box of key name.

Issue #8121, opened 2020-07-22
Exceptions thrown when loading Gallery machine keys


loic-sharma:
Problem ID: `System.ObjectDisposedException at Autofac.Core.Lifetime.LifetimeScope.CheckNotDisposed`
Innermost message: `Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it has already been disposed.`

Here is affected line:

https://github.com/NuGet/NuGetGallery/blob/e73857484000509779c18ab4ea1bec93cc6131fa/src/NuGetGallery/App_Start/GalleryMachineKeyConfigurationProvider.cs#L25

This is the change that appears to be causing the exception: https://github.com/NuGet/NuGetGallery/pull/8098

[LINK](https://ms.portal.azure.com#@33e01921-4d64-4f8c-a055-5bdaffd5e33d/blade/Microsoft_Azure_Monitoring_Logs/LogsBlade/resourceId/%2Fsubscriptions%2F685c4662-53d8-40f9-ac51-926097ede041%2FresourceGroups%2Fnuget-prod-0-v2gallery%2Fproviders%2Fmicrosoft.insights%2Fcomponents%2Fnuget-prod-v2gallery/source/LogsBlade.AnalyticsShareLinkToQuery/q/H4sIAAAAAAAAA0WNQQrCMBRE94J3%252BPQAoR6gC6kuBLGLniD5mdJo4w%252FJFy14eMlC3Q3Mmzd4MZIGuZft5k3PGRmUsrgF8eSp66gZ16KIZnBXsB5CSVLgj98ZWaX9Q2WybHrJMOcwQUP8h5ElwfQz%252BHaRn6Cpd5I9MrmVKlfUxkQehWu1hBiUdm37AbD9qqehAAAA/timespan/P1D)

```
exceptions
| where problemId == "System.ObjectDisposedException at Autofac.Core.Lifetime.LifetimeScope.CheckNotDisposed"
| order by timestamp desc
| limit 100
```

Call stack:

```
[
   {
      "outerId":"0",
      "message":"Failed to decrypt using provider 'GalleryMachineKeyConfigurationProvider'. Error message from the provider: Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it has already been disposed. (D:\\home\\site\\wwwroot\\web.config line 364)",
      "parsedStack":[
         {
[...]

PR #8209, opened 2020-09-15
[Cookie Compliance] The new cookie compliance pipeline.

zhhyu:
**Call the new async cookie compliance API:**
image

**UI changes for customers in EU countries:**
Banner:
image
Footer:
image
Pop-up window to change the cookie preference:
image


src/NuGetGallery.Services/ServicesConstants.cs
```diff
@@ -46,6 +46,11 @@ public static class ServicesConstants
 
         public const string ApiKeyHeaderName = "X-NuGet-ApiKey";
 
+        /// <summary>
+        /// Parameter for passing the cookie compliance permission.
+        /// </summary>
+        public const string CookieComplianceCanWriteAnalyticsCookies = "CanWriteAnalyticsCookies";
```
joelverhagen, line 52: I see three categories in the interface: social media, analytics, and advertising. But I only see one const here. Why is that?

zhhyu: We don't hold cookies used for social media and advertising, so the implementation now doesn't actively handles these kinds of cookies. If we need them in the future, it can be easily extended.

src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
```diff
@@ -15,37 +11,16 @@ namespace NuGetGallery.Cookies
...
-            }
-
-            _domain = domain;
-            _diagnostics = diagnostics.GetSource(GetType().Name);
-
-            return Task.Delay(0);
-        }
-        
-        public abstract bool CanWriteNonEssentialCookies(HttpRequestBase request);
+        public abstract Task<bool> CanWriteAnalyticsCookies(HttpRequestBase request);
```
loic-sharma, line 14: Can we add `Async` suffixes to these methods?

tests/NuGetGallery.Facts/Controllers/AppControllerFacts.cs
```diff
@@ -57,6 +62,76 @@ public void DefaultsToDenyGet()
...
+            [Theory]
+            [InlineData(true)]
+            [InlineData(false)]
+            public void SetViewBagGivenHttpContextItemsWithCanWriteAnalyticsCookies(bool canWriteAnalyticsCookies)
+            {
+                // Arrange
+                var controller = GetController<TestableAppController>();
+                var cookieComplianceService = new Mock<ICookieComplianceService>();
+                cookieComplianceService.Setup(c => c.ExpireAnalyticsCookies(It.IsAny<HttpContextBase>()));
+                controller.SetCookieComplianceService(cookieComplianceService.Object);
```
loic-sharma, line 76: Can we remove the `SetCookieComplianceService` method if we use the [`TestContainer.GetMock`](https://github.com/NuGet/NuGetGallery/blob/765917957b830837af97818e24a7a7be78f440ec/tests/NuGetGallery.Facts/Framework/TestContainer.cs#L111-L139) helper?

Here's an example that does that: https://github.com/NuGet/NuGetGallery/blob/4c7cfff361ce13f7f6828362461b6f7c187860f1/tests/NuGetGallery.Facts/Controllers/ManageDeprecationJsonApiControllerFacts.cs#L73

tests/NuGetGallery.Facts/Controllers/AppControllerFacts.cs
```diff
@@ -57,6 +62,76 @@ public void DefaultsToDenyGet()
...
+
+                // Act
+                InvokeOnActionExecutedMethod(controller.ControllerContext, httpContext.Object, controller);
+
+                // Assert
+                Assert.Equal(canWriteAnalyticsCookies, controller.ViewBag.CanWriteAnalyticsCookies);
+                if (canWriteAnalyticsCookies)
+                {
+                    cookieComplianceService.Verify(c => c.ExpireAnalyticsCookies(It.IsAny<HttpContextBase>()), Times.Never);
+                } else
```
loic-sharma, line 93: Nit:
```suggestion
                }
                else
```

joelverhagen:
Can you show a screenshot of what happens when you click "Manage Cookies"?

src/NuGetGallery.Services/ServicesConstants.cs
joelverhagen: Will an option to turn them off be surfaced in the UI? That may be needlessly confusing/scary to users if we have no cookies in that category.

joelverhagen: An option _to turn off categories that do not apply for our site_ I mean.

zhhyu:
> Can you show a screenshot of what happens when you click "Manage Cookies"?

Add it. It's fully managed by the JS we load.

src/NuGetGallery.Services/ServicesConstants.cs
zhhyu: Yeah, that's a good point. The UI of changing cookie preference is not controlled by us. I will ask them about it whether there is an option.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -1469,12 +1469,6 @@ private static void RegisterCookieComplianceService(ContainerBuilder builder, Co
                     sp.ComposeExportedValue("Domain", siteName);
                     sp.ComposeExportedValue<IDiagnosticsService>(diagnostics);
                 }).FirstOrDefault() as CookieComplianceServiceBase;
-
-                if (service != null)
-                {
-                    // Initialize the service on App_Start to avoid any performance degradation during initial requests.
-                    HostingEnvironment.QueueBackgroundWorkItem(async cancellationToken => await service.InitializeAsync(siteName, diagnostics, cancellationToken));
```
joelverhagen, line 1476: 👏

src/NuGetGallery/Controllers/AppController.cs
```diff
@@ -96,5 +105,31 @@ protected override void OnActionExecuting(ActionExecutingContext filterContext)
...
+        }
+
+        private void SetCookieCompliance(ActionExecutedContext filterContext)
+        {
+            if (filterContext.HttpContext?.Items[ServicesConstants.CookieComplianceCanWriteAnalyticsCookies] == null
+                || (bool)filterContext.HttpContext.Items[ServicesConstants.CookieComplianceCanWriteAnalyticsCookies] == false)
+            {
+                ViewBag.CanWriteAnalyticsCookies = false;
+
+                _cookieComplianceService.ExpireAnalyticsCookies(filterContext.HttpContext);
```
joelverhagen, line 127: I'm curious about why you put the expiration logic behind the shim. The cookies are dropped by components in the open source part of NuGetGallery, namely: AI integration, GA integration, and the nugetab cookie.

src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -311,16 +311,36 @@
...
 }
 
-@helper AnalyticsScript()
+@helper CookieComplianceScript()
+{
+    <script src="https://consentdeliveryfd.azurefd.net/mscc/lib/v2/wcp-consent.js"></script>
+    <script>
+        'use strict';
+        var WcpSiteConsent;
+        WcpConsent.init("en-US", "cookie-banner", function (err, _siteConsent) {
```
joelverhagen, line 320: Could this be included in the bundle loaded on every page?

joelverhagen: This would reduce each page's size but putting in an a single cached asset.

src/NuGetGallery/App_Code/ViewHelpers.cshtml
```diff
@@ -311,16 +311,36 @@
...
+            if (err != undefined) {
```
joelverhagen, line 321:
```suggestion
            if (err !== undefined) {

```

src/NuGetGallery/Controllers/AppController.cs
joelverhagen: Ideally we put as little as possible in the shim since it's more difficult to iterate there.

src/NuGetGallery/Modules/AsyncSetCookieComplianceModule.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        internal ICookieComplianceService CookieComplianceService;
+
+        public void Dispose()
+        {
+        }
+
+        public void Init(HttpApplication context)
+        {
+            CookieComplianceService = DependencyResolver.Current.GetService<ICookieComplianceService>();
```
joelverhagen, line 22: Should we really use a single one of these services for the lifetime of the application? Most of our other (open source) services are one-per-request. Is it safe?

src/NuGetGallery/Modules/AsyncSetCookieComplianceModule.cs
```diff
@@ -0,0 +1,43 @@
...
+            var eventHandlerTaskAsyncHelper = new EventHandlerTaskAsyncHelper(SetCookieComplianceAsync);
+            context.AddOnBeginRequestAsync(eventHandlerTaskAsyncHelper.BeginEventHandler, eventHandlerTaskAsyncHelper.EndEventHandler);
+        }
+
+        internal async Task SetCookieComplianceAsync(object sender, EventArgs e)
+        {
+            var context = ((HttpApplication)sender).Context;
+
+            var request = new HttpRequestWrapper(context.Request);
+            if (await CookieComplianceService.CanWriteAnalyticsCookies(request))
```
joelverhagen, line 33: If this throws, every request in our application will fail. How reliable is this code? Would it be worth having a try/catch that logs and assumes that we _can't_ write analytics cookies? I'm just thinking, not sure at this point.

src/NuGetGallery/Modules/AsyncSetCookieComplianceModule.cs
joelverhagen, line 22: We have run into a problem where this throws `ObjectDisposedException` in a non-request context in the past but it was hard to repro and therefore we could not root cause it. Would it be possible to _not_ use `DependencyResolver.Current`? Perhaps we can activate it without DI?

joelverhagen: Or, with a dedicate DI container that we control the lifetime of (with the `Dispose` method).

joelverhagen: See https://github.com/NuGet/NuGetGallery/issues/8121.

src/NuGetGallery/Modules/AsyncSetCookieComplianceModule.cs
zhhyu: Good point! That's what we need to introduce.

src/NuGetGallery/Modules/AsyncSetCookieComplianceModule.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System;
+using System.Web;
+using System.Web.Mvc;
+using System.Threading.Tasks;
+using NuGetGallery.Cookies;
+
+namespace NuGetGallery.Modules
+{
+    public class AsyncSetCookieComplianceModule : IHttpModule
```
skofman1, line 12: >AsyncSetCookieComplianceModule [](start = 17, length = 30)

I don't think we need Async as part of the name of the class. Just CookieComlianceHttpModule...

src/NuGetGallery/Modules/AsyncSetCookieComplianceModule.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        internal async Task SetCookieComplianceAsync(object sender, EventArgs e)
+        {
+            var context = ((HttpApplication)sender).Context;
+
+            var request = new HttpRequestWrapper(context.Request);
+            if (await CookieComplianceService.CanWriteAnalyticsCookies(request))
+            {
+                context.Items.Add(ServicesConstants.CookieComplianceCanWriteAnalyticsCookies, true);
```
skofman1, line 35: >true [](start = 94, length = 4)

use named parameters please

src/NuGetGallery/Modules/AsyncSetCookieComplianceModule.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            CookieComplianceService = DependencyResolver.Current.GetService<ICookieComplianceService>();
+
+            var eventHandlerTaskAsyncHelper = new EventHandlerTaskAsyncHelper(SetCookieComplianceAsync);
+            context.AddOnBeginRequestAsync(eventHandlerTaskAsyncHelper.BeginEventHandler, eventHandlerTaskAsyncHelper.EndEventHandler);
+        }
+
+        internal async Task SetCookieComplianceAsync(object sender, EventArgs e)
+        {
+            var context = ((HttpApplication)sender).Context;
```
skofman1, line 30: >context [](start = 16, length = 7)

Please program more defensively here. For example, this line will throw if sender is not an HttpApplication. Context may be null, context.Request may ne null, etc.

src/NuGetGallery/Web.config
```diff
@@ -396,6 +397,7 @@
       <add name="ErrorMail" type="Elmah.ErrorMailModule, Elmah, Version=1.2.14706.0, Culture=neutral, PublicKeyToken=57eac04b2e0f138e, processorArchitecture=MSIL" preCondition="managedHandler"/>
       <add name="ErrorFilter" type="Elmah.ErrorFilterModule, Elmah, Version=1.2.14706.0, Culture=neutral, PublicKeyToken=57eac04b2e0f138e, processorArchitecture=MSIL" preCondition="managedHandler"/>
       <add name="AsyncFileUpload" type="NuGetGallery.AsyncFileUpload.AsyncFileUploadModule, NuGetGallery.Services" preCondition="managedHandler"/>
+      <add name="AsyncSetCookieCompliance" type="NuGetGallery.Modules.AsyncSetCookieComplianceModule, NuGetGallery" preCondition="managedHandler"/>
```
skofman1, line 400: >l [](start = 35, length = 1)

Seems like this module will be loading for all requests. The thing is, the vast (vast vast) majority of Gallery requests are V2 OData.. I'm concerned that this will impact perf and also CPU. How heavy is this load? If it's heavy, can we somehow exclude OData?

src/NuGetGallery.Core/Cookies/CookieComplianceServiceBase.cs
```diff
@@ -15,37 +11,16 @@ namespace NuGetGallery.Cookies
     /// </summary>
     public abstract class CookieComplianceServiceBase : ICookieComplianceService
```
skofman1, line 12: why do we need this class? The interface is not sufficient?

src/NuGetGallery/Controllers/AppController.cs
```diff
@@ -30,9 +32,16 @@ public void SetOwinContextOverride(IOwinContext owinContext)
             _overrideContext = owinContext;
         }
 
+        public void SetCookieComplianceService(ICookieComplianceService cookieComplianceService)
```
skofman1, line 35: >SetCookieComplianceService [](start = 20, length = 26)

is this method used only for testing? If so, please add a comment

src/NuGetGallery/Controllers/AppController.cs
```diff
@@ -96,5 +105,31 @@ protected override void OnActionExecuting(ActionExecutingContext filterContext)
...
+        {
+            SetCookieCompliance(filterContext);
+
+            base.OnActionExecuted(filterContext);
+        }
+
+        private void SetCookieCompliance(ActionExecutedContext filterContext)
+        {
+            if (filterContext.HttpContext?.Items[ServicesConstants.CookieComplianceCanWriteAnalyticsCookies] == null
+                || (bool)filterContext.HttpContext.Items[ServicesConstants.CookieComplianceCanWriteAnalyticsCookies] == false)
```
skofman1, line 123: >HttpContext [](start = 39, length = 11)

HttpsContext can be null

tests/NuGetGallery.Facts/Controllers/AppControllerFacts.cs
zhhyu: The key is here: https://github.com/NuGet/NuGetGallery/blob/8d571889795f62e241da2c4457e63a7d4b9330c4/src/NuGetGallery/Controllers/AppController.cs#L38-L41
It's hard to mock "DependencyResolver.Current.GetService<T>()" as "GetService" is a static method.

src/NuGetGallery/Modules/AsyncSetCookieComplianceModule.cs
zhhyu: The "context.Items" is a dictionary. I feel that the key's name provides the clue here.

src/NuGetGallery/Web.config
zhhyu: It works in a singleton way. The memory burden should be minimized now. The processing time is also minimal based on the experiment.
If we meet performance issues here, we can consider some ways to filter some routes.

src/NuGetGallery/Controllers/AppController.cs
zhhyu: The first check will make it equal to null when the HttpContext is null, so the second one will not be triggered. Let me know if I misunderstand here.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -1458,29 +1462,21 @@ private static void RegisterAuditingServices(ContainerBuilder builder, IAuditing
                 .SingleInstance();
         }
 
-        private static void RegisterCookieComplianceService(ContainerBuilder builder, ConfigurationService configuration, DiagnosticsService diagnostics)
+        private static void RegisterCookieComplianceService(ConfigurationService configuration, ILoggerFactory loggerFactory)
         {
-            CookieComplianceServiceBase service = null;
+            var logger = loggerFactory.CreateLogger("CookieComplianceService");
```
loic-sharma, line 1467: Please use `nameof` if possible so that this string doesn't get stale if we rename the type.

```suggestion
            var logger = loggerFactory.CreateLogger(nameof(CookieComplianceService));
```

src/NuGetGallery/Infrastructure/IABTestService.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Threading.Tasks;
```
loic-sharma, line 4: Is this change necessary?

src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -635,5 +635,17 @@
                 skippedToContent.focus();
             }
         });
+
+        window.WcpConsent && WcpConsent.init("en-US", "cookie-banner", function (err, _siteConsent) {
+            if (err !== undefined) {
+                console.log("Error initializing WcpConsent: " + err);
```
loic-sharma, line 641: Is `err` an object? If so, you can dump objects to the browser's console:

```suggestion
                console.log("Error initializing WcpConsent: ", err);
```

Please ignore this comment if `err` is a string.

src/NuGetGallery/Scripts/gallery/common.js
```diff
@@ -635,5 +635,17 @@
...
+            } else {
+                window.nuget.wcpSiteConsent = _siteConsent;  // wcpSiteConsent is used to get the current consent
+            }
+        });
+
+        if (window.nuget.wcpSiteConsent.isConsentRequired) {
```
loic-sharma, line 647: Just double-checking - will this work ok if the initialization fails and `wcpSiteConsent` isn't set?

src/NuGetGallery/Modules/CookieComplianceHttpModule.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                var requestWrapper = new HttpRequestWrapper(request);
+                canWriteAnalyticsCookies = await CookieComplianceService.Instance?.CanWriteAnalyticsCookiesAsync(requestWrapper);
+            }
+            catch (Exception exception)
+            {
+                CookieComplianceService.Logger?.LogError(0, exception, "Cookie compliance check failed in the module: {ModuleName}", nameof(CookieComplianceHttpModule));
+            }
+
+            if (canWriteAnalyticsCookies)
```
joelverhagen, line 55: this could just be a one-liner right, where you pass `canWriteAnalyticsCookies` as a second parameter?

joelverhagen (approved):
Looks good, please verify in DEV/INT for several days before moving to PROD as well making sure `git clone` F5 experience still works (i.e. with the null implementation).

src/NuGetGallery.Core/Cookies/CookieExpirationService.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        // Application Insights cookies
+        private static readonly IReadOnlyList<string> ApplicationInsightsCookies = new[]
+        {
+            "ai_user",
+            "ai_session",
+        };
+
+        private readonly string Domain;
+        private readonly string RootDomain;
```
loic-sharma, line 31: Nit, please use `_abc` style for private fields:

```suggestion
        private readonly string _domain;
        private readonly string _rootDomain;
```

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -442,6 +442,10 @@ protected override void Load(ContainerBuilder builder)
                 .As<IPackageVulnerabilityService>()
                 .InstancePerLifetimeScope();
 
+            builder.Register(c => new CookieExpirationService(domain: configuration.GetSiteRoot(true)))
```
loic-sharma, line 445: Can you use a named parameter here? It makes it a little clearer what the `true` is for:

```suggestion
            builder.Register(c => new CookieExpirationService(domain: configuration.GetSiteRoot(useHttps: true)))
```

loic-sharma (approved):

Issue #8189, opened 2020-09-03
[A11y][UI]Ensures every id attribute value is unique (form)


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1184527

On NuGetGallery | Manage Organization page
## Snippet
<form aria-hidden="true" id="AntiForgeryForm"> <input name="__RequestVerificationToken" type="hidden" value="..
## How to fix
Fix any of the following:
- Document has multiple static elements with the same id attribute: AntiForgeryForm

## Environment
Chrome version 84.0.4147.135 This accessibility issue was found using Accessibility Insights for Web 2.21.1 (axe-core 4.0.1), a tool that helps find and fix accessibility issues. Get more information & download this tool at http://aka.ms/AccessibilityInsights.

## Browser Details:
Firefox Version: 79.0 (64 bit)

OS Details

Microsoft Windows 10 Enterprise  ​

Version 2004 (OS Build 19041.450)

Issue #50, opened 2011-08-17
Add a link labelled "Where is the download?"


haacked:
The link should point to a page that briefly explains what NuGet is and how it works.

PR #8210, opened 2020-09-15
make attribute id value unique

lyndaidaii:
Remove additional call for @ViewHelpers.AjaxAntiForgeryToken(Html) in _AccountCertificates.cshtml to avoid duplicate attribute id value

Addresses https://github.com/NuGet/NuGetGallery/issues/8189

### DIFF
--- a/src/NuGetGallery/Views/Shared/_AccountCertificates.cshtml
+++ b/src/NuGetGallery/Views/Shared/_AccountCertificates.cshtml
@@ -47,7 +47,6 @@
         <span id="certificates-section-header"></span>
     </text>,
     @<text>
-        @ViewHelpers.AjaxAntiForgeryToken(Html)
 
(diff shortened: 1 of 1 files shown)


joelverhagen (approved):

joelverhagen:
Please ensure this partial is only used in one place. If it is not, ensure the page still has a token form.

src/NuGetGallery/Views/Shared/_AccountCertificates.cshtml
```diff
@@ -47,7 +47,6 @@
         <span id="certificates-section-header"></span>
     </text>,
     @<text>
-        @ViewHelpers.AjaxAntiForgeryToken(Html)
 
         <form id="addCertificateForm" enctype="multipart/form-data" aria-hidden="true">
```
lyndaidaii, line 51: There is token form called here. Delete #50 should be good. @joelverhagen

PR #8214, opened 2020-09-21
Add roslyn analyzer and fix security warning issue

lyndaidaii:

* Add roslyn analyzer prototype 
* Fix security warning generated by analayzer
* Chang use MSbuild with VS2019 in build script in Functional test

Addresses https://github.com/NuGet/Engineering/issues/3251


src/NuGet.Services.Entities/Properties/AssemblyInfo.cs
```diff
@@ -1,13 +1,16 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+﻿using System.Resources;
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Reflection;
 using System.Runtime.InteropServices;
 
+[assembly: AssemblyVersion("1.0.0")]
```
joelverhagen, line 8: Assembly version should be generated in the build.ps1. 
https://github.com/NuGet/NuGetGallery/blob/8d571889795f62e241da2c4457e63a7d4b9330c4/build.ps1#L84

Is the analyzer running before build.ps1?

src/NuGetGallery.Core/Certificates/CertificateFile.cs
```diff
@@ -63,10 +63,9 @@ private static MemoryStream CopyAsReadOnly(Stream source)
             }
         }
 
-        [SuppressMessage("Microsoft.Security.Cryptography", "CA5354:SHA1CannotBeUsed", Justification = @"SHA1 thumbprint is only used to distinguish certificates in the Gallery view.")]
         private static string GetSha1Thumbprint(MemoryStream stream)
         {
-            using (var hashAlgorithm = SHA1.Create())
+            using (var hashAlgorithm = SHA256.Create())
```
joelverhagen, line 68: Is this method even used? I would argue that changing the implementation to SHA256 is the wrong approach since the method name is SHA1 and it's possible the callers are expecting SHA1. Ideally, we don't even have a method like this.

src/NuGetGallery/Controllers/AccountsController.cs
```diff
@@ -380,6 +381,7 @@ public virtual async Task<ActionResult> Delete(DeleteAccountAsAdminViewModel mod
 
         protected abstract DeleteAccountViewModel GetDeleteAccountViewModel(TUser account);
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Security", "CA3147:Mark Verb Handlers With Validate Antiforgery Token", Justification = "<Pending>")]
```
joelverhagen, line 384: Would we provide a justification here?

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+#pragma warning disable CA3147 // Mark Verb Handlers With Validate Antiforgery Token
```
joelverhagen, line 4: Could you leave a more specific comment at the top here about why we can disable this?

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -489,6 +489,8 @@ public ActionResult ChallengeAuthentication(string returnUrl, string provider, A
         /// </summary>
         /// <param name="returnUrl">The url to return upon credential replacement</param>
         /// <returns><see cref="ActionResult"/> for returnUrl</returns>
+        [HttpPost]
+        [ValidateAntiForgeryToken]
```
joelverhagen, line 493: Is this really a `POST`?

tests/NuGetGallery.Core.Facts/Services/CloudBlobCoreFileStorageServiceIntegrationTests.cs
```diff
@@ -183,7 +183,7 @@ public async Task OpenWriteAsyncReturnsWritableStream()
             var expectedContent = "Hello, world.";
             var bytes = Encoding.UTF8.GetBytes(expectedContent);
             string expectedContentMD5;
-            using (var md5 = MD5.Create())
+            using (var md5 = SHA256.Create())
```
joelverhagen, line 186: This needs to remain as MD5 since it's comparing against the blob storage MD5 header. I recommend a suppression here since this is only test code.

lyndaidaii:
Build Failed with VS2017, Build successfully with VS 2019. Are we considering to update to VS2019 in CI machine?

src/NuGet.Services.Entities/Properties/AssemblyInfo.cs
lyndaidaii: CA1016 comes from compiler, Roslyn analyzer is included within the .NET SDK starting 5.0, that's why there is CA 1016 warning happen? is running during build 
here is reference: https://docs.microsoft.com/en-us/dotnet/fundamentals/productivity/code-analysis#code-style-analysis

src/NuGetGallery.Core/Certificates/CertificateFile.cs
lyndaidaii: Yes, this method is used. SHA1 is not secure hash function, I think we could suppress this warning

src/NuGetGallery/Controllers/AccountsController.cs
lyndaidaii: Not necessary, roslyn analyzer complains about the CA3147 warning here.

src/NuGetGallery/Controllers/ApiController.cs
lyndaidaii: The reason why disable here is Api call, it's not necessary need to Validate Antiforgery Token if I understand right here?

src/NuGetGallery/Controllers/AuthenticationController.cs
lyndaidaii: `GET`

.editorconfig
```diff
@@ -0,0 +1,33 @@
+[*.cs]
+
+# Suggest performance, usage, naming, maintainability, style diagnostics.
```
joelverhagen, line 3:
```suggestion
# Suppress performance, usage, naming, maintainability, style diagnostics.
```

src/NuGetGallery.Core/Certificates/CertificateFile.cs
loic-sharma: I believe this is used to display your signing certificate's SHA-1 thumbprint. This is only used for display purposes, NOT, for security purposes.

Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers">
+    <Version>3.3.0</Version>
```
loic-sharma, line 4: Please indent lines 4-6

src/NuGet.Services.Entities/Properties/AssemblyInfo.cs
```diff
@@ -1,4 +1,5 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+﻿using System.Resources;
```
loic-sharma, line 1: The copy right header should be the first thing in the file. Please move this down to line 6

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -2089,6 +2092,7 @@ public virtual async Task<ActionResult> UpdateListed(string id, string version,
         [ValidateInput(false)] // Security note: Disabling ASP.Net input validation which does things like disallow angle brackets in submissions. See http://go.microsoft.com/fwlink/?LinkID=212874
         [ValidateAntiForgeryToken]
         [RequiresAccountConfirmation("edit a package")]
+        [SuppressMessage("Security", "CA5363:Do Not Disable Request Validation", Justification = "<Pending>")]
```
loic-sharma, line 2095: Can we move the comment on line 2092 to the `Justification`? Same for line 2318

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -535,6 +536,7 @@ public virtual async Task<ActionResult> LinkOrChangeExternalCredential(string re
             return SafeRedirect(returnUrl);
         }
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Security", "CA3147:Mark Verb Handlers With Validate Antiforgery Token", Justification = "<Pending>")]
```
loic-sharma, line 539: Why don't we need an antiforgery token here? Is this a GET request?

lyndaidaii: This function is not clear as GET request or POST request.  I could be both depends on condition. That's why I suppressed

loic-sharma: Gotcha, that makes sense. Please update the `Justification` with your comment :)

src/NuGetGallery.Core/Certificates/CertificateFile.cs
```diff
@@ -63,10 +63,11 @@ private static MemoryStream CopyAsReadOnly(Stream source)
             }
         }
 
-        [SuppressMessage("Microsoft.Security.Cryptography", "CA5354:SHA1CannotBeUsed", Justification = @"SHA1 thumbprint is only used to distinguish certificates in the Gallery view.")]
         private static string GetSha1Thumbprint(MemoryStream stream)
```
joelverhagen, line 66: We don't use the SHA1 thumbprint in the view. The property that exposes it on the EF model is `[Obsolete]`.

src/NuGetGallery/Services/SqlAggregateStatsService.cs
```diff
@@ -26,6 +26,7 @@ public SqlAggregateStatsService(IAppConfiguration configuration, ISqlConnectionF
             _connectionFactory = galleryDbConnectionFactory;
         }
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Security", "CA2100:Review SQL queries for security vulnerabilities", Justification = "<Pending>")]
```
joelverhagen, line 29: Does making `GetStatisticsSql` into a `const` negate the need for this suppression?

src/NuGetGallery.Core/Certificates/CertificateFile.cs
lyndaidaii: SHA1 thumbprint is called Create certificate File. https://github.com/NuGet/NuGetGallery/blob/db74821c68bb39442d5c2185824cc75dab12fdeb/src/NuGetGallery.Core/Certificates/CertificateFile.cs#L22
you mean we could set Sha1Thumbprint with [ObsoleteAttribute], and don't use this function in create and also construct?

src/NuGetGallery/Services/SqlAggregateStatsService.cs
lyndaidaii: `const`  negate the need for this suppression.

src/NuGetGallery.Core/Certificates/CertificateFile.cs
joelverhagen: Sorry, I was providing some information to include in the justification. I should have been more clear.

src/NuGetGallery/Controllers/AuthenticationController.cs
```diff
@@ -535,6 +536,7 @@ public virtual async Task<ActionResult> LinkOrChangeExternalCredential(string re
             return SafeRedirect(returnUrl);
         }
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Security", "CA3147:Mark Verb Handlers With Validate Antiforgery Token", Justification = "Disable antiforgery token validation due to actions handle GET/POST requests")]
```
loic-sharma, line 539: Please add a `using System.Diagnostics.CodeAnalysis` to shorten this line

loic-sharma (approved):

Issue #8190, opened 2020-09-03
[A11y][Reflow]Some of the contents under the select packages is getting overlapped when reflow is applied


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1188114
## Browser Details:
Firefox Version: 79.0 (64 bit)

## OS Details

Microsoft Windows 10 Enterprise  ​

Version 2004 (OS Build 19041.450) 

## Repro Steps:

    Change the Display resolution to 1280*1024 in System display settings
    Hit the URL and login with appropriate credentials
    Tab till "Open Profile drop down" and press enter
    Tab till "API Keys" and Press enter
    Tab till "Create button" and press enter
    Apply reflow
    Check whether data is overlapping is observed or not  

## Actual Results:

Some of the contents under the select packages is getting overlapped when reflow is applied 
## Expected Results:
User should be able to view the data without overlapping and all the information should be visible clearly with one scrollbar after applying reflow
## User Impact:
Users will not be able to understand the information if it gets overlapped in reflow

## Notes
This is in reference to the section in Create Key that shows example glob patterns. You do not have to change display resolution to see this. It is sufficient to shrink the window to minimum width and zoom in until the overlap occurs
![image](https://user-images.githubusercontent.com/11051729/92165397-d8e1de00-edeb-11ea-9bd8-7ce5bde30820.png)

PR #8215, opened 2020-09-22
Fix for overflowing api key hint text on certain zoom levels.

agr:
Addresses #8190 

It seems that at some point, in order to align the bottoms of the package list and the API key hint container their respective heights were explicitly specified in CSS (see diff). At certain zoom levels (and browser window widths) the contents of the hint container gets longer than the container permits and overflows:
image

In order to get rid of hardcoding container heights a bunch of flex containers were introduced, hint container was wrapped with an additional `<div class="form-group">` in order to layout the contents in the same way as in the container on the left (i.e. to have the same margin-bottom), so bottom border end up on the same level.
At the same time, table inside the hint container was styled to not have a bottom margin to eliminate weird white space:
image

Resulting in the following look (at various widths and zoom levels):

image

image

image


src/Bootstrap/less/theme/page-api-keys.less
```diff
@@ -42,15 +42,36 @@
     .available-packages {
       .panel-body {
         overflow: auto;
-        height: 204px;
+        min-height: 180px;
```
agr, line 45: This is set to be a little bit less than the height of a hint container on any zoom level, so it only takes effect when all is laid out in a single column.
Should I add a media query around it to enforce the behavior?

agr: Added media queries.

joelverhagen (approved):
Wow, great improvement!

loic-sharma (approved):

dannyjdev (approved):

agr:
This solution misbehaves when the package list is large (it stretches the container instead of adding a scroll bar), looking for a fix.

src/NuGetGallery/Views/Users/ApiKeys.cshtml
```diff
@@ -415,59 +415,63 @@
...
         </div>
         <div class="row select-packages-form" data-bind="css: { disabled: !SelectPackagesEnabled() }">
-            <div class="col-sm-7">
-                <div class="form-group">
-                    <input type="text" class="form-control" data-val="true"
-                           data-val-regex="Invalid glob pattern. See the example glob patterns below for further details."
-                           data-val-regex-pattern="^[\*\w_.-]*$"
-                           data-bind="textInput: PendingGlobPattern,
+            <div class="col-sm-7 select-packages-inputs">
+                <div>
```
agr, line 419: New div added here to allow absolute positioning.

agr:
I think I found a [hacky] solution for long lists: added a wrapper around the left section with `position: absolute` to prevent the content of that div to contribute to the size of the container (but still positioned to occupy all available vertical space), so now the height of `.select-packages-inputs` element is determined by the only remaining item - API hint section.
All the container elements down to the list of packages were made flex-boxes, so they would stretch to occupy the available vertical space, an `overflow: auto` was moved one level above to reduce the amount of nested flexboxes.

Issue #8195, opened 2020-09-03
[A11y][Contrast]Color contrast for the "Copy" button's background with page background ratio is 2.5:1 which is less than 3:1


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1188278
## Browser Details:
Firefox Version: 79.0 (64 bit)

## OS Details

Microsoft Windows 10 Enterprise  ​

Version 2004 (OS Build 19041.450) 

## Note:
Same issue is observed at Search button and for the stars in the table
## Repro Steps:

    Hit the URL and login with appropriate credentials
    Tab till "Profile" button in the menu bar and press enter
    Tab till any Manage packages"  and press enter.
    Tab till search bar and search for 'Json' and press enter 
    Tab till the first result "Newtonsoft.Json" and press enter 
    Tab till "Copy" button and  Check whether the color contrast for the background to icon is greater than equal to 3:1 or not

## Actual Results:

Color contrast for the background to focus on the icons Facebook and Twitter is which is less than 3:1.

   
## Expected Results:
Color contrast for the "Copy" button's background with page background ratio is 2.5:1 which is less than 3:1
## User Impact:
Users with low vision would face difficulty to identify the elements if the non-text contrast is less than 3:1

## Notes
This looks similar to https://github.com/NuGet/NuGetGallery/issues/8192
A global fix for this button type should be made.

PR #8219, opened 2020-09-24
Darken the orange buttons for accessibility

loic-sharma:
For WCAG 2.1, web components must have a 3:1 contrast ratio. This darkens both the "search" button in the header as well as the "copy" button for installation instructions.

Before:
image

After:
image

Addresses https://github.com/NuGet/NuGetGallery/issues/8195

### DIFF
--- a/src/Bootstrap/dist/css/bootstrap.css
+++ b/src/Bootstrap/dist/css/bootstrap.css
@@ -2585,27 +2585,27 @@ fieldset[disabled] .btn-info.focus {
 }
 .btn-warning {
   color: #fff;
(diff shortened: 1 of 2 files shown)


joelverhagen:
Wow, the gray is actually a much better button now. The light gray is hardly visible on my laptop screen.

The orange does pop less, in my opinion but it's still good looking and now accessible now so 👍.

joelverhagen (approved):

loic-sharma:
I split the gray button out from this change as it will require a little more thought.

joelverhagen (approved):

Issue #8228, opened 2020-09-29
Fix Orchestrator runtime issue


loic-sharma:
Orchestrator fails with error:

```
Could not load file or assembly 'NuGet.Packaging.Core, Version=5.0.0.1, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.
```

Steps to be healthy:

* [x] Rollback orchestrator https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=829396
* [x] Update Gallery https://github.com/NuGet/NuGetGallery/pull/8229
* [x] Verify Gallery https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=833326
* [x] Publish NuGetGallery.Core https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=4116611&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=f03d4b0b-5c7f-5a3c-32ca-d79e6a87d689
* [x] Update Orchestrator's NuGetGallery.Core https://github.com/NuGet/NuGet.Jobs/pull/922
* [x] Verify Orchestrator DEV https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=833772

PR #8229, opened 2020-09-29
Update NuGet client libraries to 5.8.0 prerelease

loic-sharma:
The Gallery uses an older version of `NuGet.Packaging` that depends on `NuGet.Packaging.Core`. The Orchestrator was updated to use a newer version of `NuGet.Packaging` that no longer depends on `NuGet.Packaging.Core`. The Orchestrator uses assemblies produced from the Gallery, thereby causing runtime issues when these assemblies attempt to reference types from `NuGet.Packaging.Core.dll` which is unavailable.

Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4116398&view=results
Release: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=833326

Part of https://github.com/NuGet/NuGetGallery/issues/8228

### DIFF
--- a/src/NuGetGallery.Core/Packaging/PackageMetadata.cs
+++ b/src/NuGetGallery.Core/Packaging/PackageMetadata.cs
@@ -276,7 +276,7 @@ namespace NuGetGallery.Packaging
             return new PackageMetadata(
                 nuspecReader.GetMetadata().ToDictionary(kvp => kvp.Key, kvp => kvp.Value),
                 nuspecReader.GetDependencyGroups(useStrictVersionCheck: strict),
-                nuspecReader.GetFrameworkReferenceGroups(),
+                nuspecReader.GetFrameworkAssemblyGroups(),
                 nuspecReader.GetPackageTypes(),
                 nuspecReader.GetMinClientVersion(),
                 nuspecReader.GetRepositoryMetadata(),
--- a/tests/NuGetGallery.Facts/Services/PackageMetadataValidationServiceFacts.cs
+++ b/tests/NuGetGallery.Facts/Services/PackageMetadataValidationServiceFacts.cs
@@ -67,13 +67,13 @@ namespace NuGetGallery
             return packageMetadataValidationService.Object;
         }
 
-        public class TheValidateMetadatabeforeUploadAsyn : FactsBase
+        public class TheValidateMetadataBeforeUploadAsync : FactsBase
         {
             private Mock<TestPackageReader> _nuGetPackage;
             private PackageRegistration _packageRegistration;
             private User _currentUser;
 
-            public TheValidateMetadatabeforeUploadAsyn()
+            public TheValidateMetadataBeforeUploadAsync()
             {
                 _nuGetPackage = GeneratePackage(isSigned: true);
                 _packageRegistration = _package.PackageRegistration;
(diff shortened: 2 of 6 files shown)


src/NuGetGallery.Core/Packaging/PackageMetadata.cs
```diff
@@ -276,7 +276,7 @@ public static PackageMetadata FromNuspecReader(NuspecReader nuspecReader, bool s
             return new PackageMetadata(
                 nuspecReader.GetMetadata().ToDictionary(kvp => kvp.Key, kvp => kvp.Value),
                 nuspecReader.GetDependencyGroups(useStrictVersionCheck: strict),
-                nuspecReader.GetFrameworkReferenceGroups(),
+                nuspecReader.GetFrameworkAssemblyGroups(),
```
loic-sharma, line 279: The `GetFrameworkReferenceGroups` method has been deprecated. Its implementation simply calls its replacement, `GetFrameworkAssemblyGroups`. See: https://github.com/NuGet/NuGet.Client/blob/65abcc61f73ab3d87928495f789ec74011edde7f/src/NuGet.Core/NuGet.Packaging/NuspecReader.cs#L202-L206

joelverhagen (approved):
Verify in DEV then :shipit:

tests/NuGetGallery.Facts/Services/PackageMetadataValidationServiceFacts.cs
```diff
@@ -611,7 +611,7 @@ public async Task RejectsUnlicensedPackages()
             [InlineData("Vim WITH Font-exception-2.0", true)] // we are not checking if license expression make sense
             [InlineData("Vim with Font-exception-2.0", false)]
             [InlineData("Vim With Font-exception-2.0", false)]
-            [InlineData("(EUPL-1.1+ OR (SPL-1.0 WITH Nokia-Qt-exception-1.1) AND Sleepycat)", true)]
+            [InlineData("(EUPL-1.1+ OR (SPL-1.0 WITH Font-exception-2.0) AND Sleepycat)", true)]
```
loic-sharma, line 614: This test failed as the client now marks the `Nokia-Qt-exception-1.1` license exception as deprecated. I created a task to update licenses.nuget.org as this should reflect the licenses accepted by the NuGet client: https://github.com/NuGet/NuGetGallery/issues/8230

joelverhagen: 😴🐱

tests/NuGetGallery.Facts/Services/PackageMetadataValidationServiceFacts.cs
```diff
@@ -67,13 +67,13 @@ public class PackageMetadataValidationServiceFacts
             return packageMetadataValidationService.Object;
         }
 
-        public class TheValidateMetadatabeforeUploadAsyn : FactsBase
+        public class TheValidateMetadatabeforeUploadAsync : FactsBase
```
agr, line 70: Maybe fix the casing in `before` while at it, too?

agr (approved):

tests/NuGetGallery.Facts/Services/PackageMetadataValidationServiceFacts.cs
loic-sharma: Nice catch, fixed

PR #8241, opened 2020-10-06
Add feature flags to OData $count endpoints

joelverhagen:
Related to https://github.com/NuGet/Engineering/issues/2902
Address https://github.com/NuGet/Engineering/issues/3442

### DIFF
--- a/src/NuGetGallery/Controllers/ODataV1FeedController.cs
+++ b/src/NuGetGallery/Controllers/ODataV1FeedController.cs
@@ -9,7 +9,6 @@ using System.Threading.Tasks;
 using System.Web.Http;
 using System.Web.Http.OData;
 using System.Web.Http.OData.Query;
-using Microsoft.Data.OData;
 using NuGet.Services.Entities;
 using NuGetGallery.Configuration;
 using NuGetGallery.OData;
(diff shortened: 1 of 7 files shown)


src/NuGetGallery/Controllers/ODataV1FeedController.cs
```diff
@@ -253,6 +272,38 @@ public IHttpActionResult GetPropertyFromPackages(string propertyName, string id,
...
+        {
+            return (await Search(
+                options,
+                searchTerm,
+                targetFramework,
+                _featureFlagService.IsODataV1SearchCountNonHijackedEnabled()))
+                .FormattedAsCountResult<V1FeedPackage>();
+        }
+
+        private async Task<IHttpActionResult> Search(
```
loic-sharma, line 302: Nit: can this be renamed to `SearchAsync`?

joelverhagen: Yeah good call

src/NuGetGallery/Controllers/ODataV2FeedController.cs
```diff
@@ -60,6 +60,31 @@ public class ODataV2FeedController
...
+            [FromUri] string semVerLevel = null)
+        {
+            return (await Get(
+                options,
+                semVerLevel,
+                _featureFlagService.IsODataV2GetAllCountNonHijackedEnabled()))
+                .FormattedAsCountResult<V2FeedPackage>();
+        }
+
+        private async Task<IHttpActionResult> Get(
```
loic-sharma, line 84: Nit: can we also add the `Async` suffix here?

src/NuGetGallery/Controllers/ODataV1FeedController.cs
loic-sharma: Same comment for other async private methods you added

loic-sharma: Ah actually I realize that you didn't moved these methods, not added them. Never mind

loic-sharma (approved):

src/NuGetGallery/Controllers/ODataV1FeedController.cs
joelverhagen: The private methods are new and can be easily renamed.

Issue #8250, opened 2020-10-08
Remove A/B test for package dependents feature


joelverhagen:
This feature has been running successfully for several months. We no longer need the A/B test check. The existing feature flag can remain for emergency scenarios.

Found via tweet:
https://twitter.com/DonBeham/status/1313766887285764096

PR #8251, opened 2020-10-08
Remove A/B test check for the package dependents feature

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/8250

Also, did a little thinking about what the next A/B test would do. Either re-use the "pd" property in the cookie if you don't care much about independent treatment groups or add a new schema version and leave "pd" as vestigial.

### DIFF
--- a/src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
+++ b/src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
@@ -4,28 +4,32 @@
 using System;
 using System.Security.Cryptography;
 using System.Threading;
-using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 
 namespace NuGetGallery
 {
+    /// <summary>
+    /// This is the implementation that handles all current and past versions of the "nugetab" cookie. This, along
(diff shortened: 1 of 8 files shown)


src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
```diff
@@ -4,28 +4,32 @@
...
+    /// an A/B test feature is always on or off for a single user browser.
+    /// </summary>
     public class ABTestEnrollmentFactory : IABTestEnrollmentFactory
     {
-        private const int SchemaVersion1 = 1;
-        private const int SchemaVersion2 = 2;
+        private const int SchemaVersion1 = 1; // PreviewSearch: {"v":1,"ps":100}
+        private const int SchemaVersion2 = 2; // PreviewSearch + PackageDependent: {"v":2,"ps":100,"pd":100}
+
+        // Note that a new schema version could theoretically reused any currently unused cookie properties. However
```
loic-sharma, line 21:
```suggestion
        // Note that a new schema version could theoretically reuse any currently unused cookie properties. However
```

loic-sharma (approved):

Issue #8182, opened 2020-09-03
[A11y][ARIA]NVDA is not reading the label name for the 'Select Scopes' checkboxes


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1184483

## Browser Details:
Firefox Version: 79.0 (64 bit)
Edge Dev: Version 86.0.601.1 (Official build) dev (64-bit)
Screen Reader: NVDA, Narrator

## OS Details

Microsoft Windows 10 Enterprise  ​

Version 2004 (OS Build 19041.450)

## Repro Steps:

    Turn on NVDA.
    Hit the URL and login with appropriate credentials
    Tab till "Open Profile drop down" and press enter
    Tab till "API Keys" and Press enter
    Tab till "Create button" and press enter
    Tab till "Select scopes" section
    Check whether NVDA is reading the label name for the 'Select Scopes' 

## Actual Results:

NVDA is not reading the label name for the 'Select Scopes' Checkboxes

## Expected Results:
NVDA should read the label name for the "Select Scopes"  
## User Impact:
Users with visual disabilities may not understand the information clearly if screen reader doesn't read the label names

## Notes:
Check the aria-label or aria-labelled-by property on the scopes. Chances are, this isn't labelled correctly.

PR #8252, opened 2020-10-08
[A11y] Call out the 'select scopes' field label

loic-sharma:
Update the API key form to call out the `select scopes` label after tabbing to inputs within that field group.

Addresses https://github.com/NuGet/NuGetGallery/issues/8182

### DIFF
--- a/src/NuGetGallery/Views/Users/ApiKeys.cshtml
+++ b/src/NuGetGallery/Views/Users/ApiKeys.cshtml
@@ -353,11 +353,11 @@
                                 <div class="radio">
                                     <div class="label-sibling">
                                         <input name="PushScope" type="radio" value="@NuGetScopes.PackagePush"
-                                               aria-labelledby="select-scopes-package-push select-scopes"
+                                               aria-labelledby="select-scopes-push-package select-scopes"
                                                data-bind="checked: PushScope, enable: PushNewEnabled,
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Users/ApiKeys.cshtml
```diff
@@ -335,16 +335,16 @@
...
                 <span class="has-error">
                     <span class="help-block" data-bind="text: ScopesError" aria-live="polite" role="alert"></span>
                 </span>
                 <ul role="presentation">
                     <li>
                         <div class="checkbox">
-                            <label>
-                                <input type="checkbox" data-bind="checked: PushEnabled, enable: PushAnyEnabled" />
+                            <label id="select-scopes-push">
+                                <input type="checkbox" data-bind="checked: PushEnabled, enable: PushAnyEnabled" aria-labelledby="select-scopes-push select-scopes" />
```
loic-sharma, line 347: The narrator calls out the field, followed by the group, as per this guidance: https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/checkbox/checkbox-2.html

joelverhagen: TIL. Nice one.

joelverhagen (approved):

PR #8256, opened 2020-10-11
[Cookie Compliance] Stabilize the pipeline

zhhyu:
Just a Revert, nothing needs to be changed in this PR.
Will send another PR to expire GA's cookie.


loic-sharma (approved):

joelverhagen:
Why do we expire the cookie in 1 year? Looks like the default is 2, so what's the significant difference?

zhhyu:
> Why do we expire the cookie in 1 year? Looks like the default is 2, so what's the significant difference?

This is for compliance. The cookie should be expired in 12 months. I will send another PR to handle GA's cookie.

joelverhagen (approved):

PR #8263, opened 2020-10-13
[Cookies] Expire and set the domain of Google Analytics cookies

zhhyu:
- Expire Google Analytics cookies in a year;
- Set the domain as 'none', to avoid sharing Google Analytics cookies with the status page.
References:
https://developers.google.com/analytics/devguides/collection/analyticsjs/cookies-user-id
https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#cookieDomain
https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#cookieExpires

### DIFF
--- a/src/NuGetGallery.Core/Cookies/CookieExpirationService.cs
+++ b/src/NuGetGallery.Core/Cookies/CookieExpirationService.cs
@@ -28,7 +28,6 @@ namespace NuGetGallery.Cookies
         };
 
         private readonly string _domain;
-        private readonly string _rootDomain;
 
         public CookieExpirationService(string domain)
         {
@@ -38,7 +37,6 @@ namespace NuGetGallery.Cookies
             }
 
             _domain = domain;
-            _rootDomain = GetRootDomain(_domain);
         }
 
         public void ExpireAnalyticsCookies(HttpContextBase httpContext)
@@ -48,7 +46,7 @@ namespace NuGetGallery.Cookies
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            GoogleAnalyticsCookies.ToList().ForEach(cookieName => ExpireCookieByName(httpContext, cookieName));
+            GoogleAnalyticsCookies.ToList().ForEach(cookieName => ExpireCookieByName(httpContext, cookieName, _domain));
             ApplicationInsightsCookies.ToList().ForEach(cookieName => ExpireCookieByName(httpContext, cookieName));
         }
 
@@ -83,23 +81,6 @@ namespace NuGetGallery.Cookies
                     response.Cookies[cookieName].Domain = domain;
                 }
             }
-        }
-
-        private string GetRootDomain(string domain)
-        {
-            var index1 = domain.LastIndexOf('.');
-            if (index1 < 0)
-            {
-                return domain;
-            }
-
-            var index2 = domain.LastIndexOf('.', index1 - 1);
-            if (index2 < 0)
-            {
-                return domain;
-            }
-
-            return domain.Substring(index2 + 1);
         }
     }
 }
--- a/src/NuGetGallery.Core/Cookies/CookieExpirationService.cs
+++ b/src/NuGetGallery.Core/Cookies/CookieExpirationService.cs
@@ -13,7 +13,7 @@ namespace NuGetGallery.Cookies
         private static readonly DateTime CookieExpirationTime = new DateTime(2010, 1, 1);
 
         // Google Analytics cookies
-        private static readonly IReadOnlyList<string> GoogleAnalyticsCookies = new[]
+        private static readonly string[] GoogleAnalyticsCookies = new[]
         {
             "_ga",
             "_gid",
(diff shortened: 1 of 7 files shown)


src/NuGetGallery.Services/Configuration/ConfigurationService.cs
```diff
@@ -91,6 +91,20 @@ public string GetSiteRoot(bool useHttps)
             return useHttps ? _httpsSiteRootThunk.Value : _httpSiteRootThunk.Value;
         }
 
+        /// <summary>
+        /// Gets the site domain
+        /// </summary>
+        public string GetSiteDomain()
+        {
+            var siteDomain = _httpsSiteRootThunk.Value.Substring(8);
```
joelverhagen, line 99: "https://", right?
This 8 is very opaque. Consider `"https://".Length`

src/NuGetGallery.Services/Configuration/ConfigurationService.cs
```diff
@@ -91,6 +91,20 @@ public string GetSiteRoot(bool useHttps)
             return useHttps ? _httpsSiteRootThunk.Value : _httpSiteRootThunk.Value;
         }
 
+        /// <summary>
+        /// Gets the site domain
+        /// </summary>
+        public string GetSiteDomain()
+        {
+            var siteDomain = _httpsSiteRootThunk.Value.Substring("https://".Length);
```
agr, line 99: `GetSiteRoot` method above has the `useHttps` argument allowing use over http, this method seems to have https hardcoded to it. Will it work correctly on standalone Gallery working over plain HTTP? Can Gallery work over plain HTTP?

agr, line 99: can `_httpsSiteRootThunk` contain port number? Should it be returned from this method?

zhhyu: Good point! Add the check for it.

tests/NuGetGallery.Facts/App_Start/ConfigurationServiceFacts.cs
```diff
@@ -99,6 +100,25 @@ public void WillThrowIfConfiguredSiteRootIsNotHttpOrHttps()
...
+
+            [Theory]
+            [InlineData("http://theSiteRoot")]
+            [InlineData("https://theSiteRoot")]
+            [InlineData("http://theSiteRoot/")]
+            [InlineData("https://theSiteRoot/")]
+            [InlineData("http://theSiteRoot:80")]
+            [InlineData("https://theSiteRoot:443")]
+            [InlineData("http://theSiteRoot:80/")]
+            [InlineData("https://theSiteRoot:443/")]
```
joelverhagen, line 112: could you have some tests where the domain has multiple labels (e.g. `www.theSiteRoot.example`)

joelverhagen (approved):

src/NuGetGallery.Services/Configuration/ConfigurationService.cs
agr: Another question: could `_httpsSiteRootThunk` be an IPv6 address with many `:`?

zhhyu: Here I mainly handle the case like "http://localhost:80/". For us, we get the "SiteRoot" from the configuration and parse it. If it's not set in the configuration but from the request ("GetHttpSiteRoot()" method), we will not be able to get the correct domain anyway.
If it's an IP address (IPv4 or IPv6), I don't think the browser can handle the cookie set with an IP address rather than a domain. So if we can not get the correct domain, then cookie set will just fail. I don't feel strong that we should take the validation check on whether it's an IP address and throw the exception if it is. Maybe I am wrong.

src/NuGetGallery.Core/Cookies/CookieExpirationService.cs
```diff
@@ -27,28 +27,14 @@ public class CookieExpirationService : ICookieExpirationService
...
-
         public void ExpireAnalyticsCookies(HttpContextBase httpContext)
         {
             if (httpContext == null)
             {
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            GoogleAnalyticsCookies.ToList().ForEach(cookieName => ExpireCookieByName(httpContext, cookieName, _rootDomain));
+            GoogleAnalyticsCookies.ToList().ForEach(cookieName => ExpireCookieByName(httpContext, cookieName));
```
agr, line 37: Do you need that `ToList()` call here? `ForEach` should work fine directly with `GoogleAnalyticsCookies`.

agr (approved):

src/NuGetGallery.Core/Cookies/CookieExpirationService.cs
zhhyu: Yeah, the IReadOnlyList (GA cookies) or its extension methods don't expose "ForEach" loop method.

PR #8269, opened 2020-10-16
Change httpPost to httpGet

lyndaidaii:

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -227,9 +227,8 @@ namespace NuGetGallery
             return Json(progress, JsonRequestBehavior.AllowGet);
         }
 
-        [HttpPost]
+        [HttpGet]
         [UIAuthorize]
-        [ValidateAntiForgeryToken]
         [RequiresAccountConfirmation("upload a package")]
         public virtual async Task<ActionResult> UploadPackage()
         {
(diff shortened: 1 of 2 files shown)


joelverhagen (approved):
Yes, these are the two that I found as well. Could you deploy this private build to DEV environment and verify the following?

- E2E tests pass
- Gallery functional tests pass
- All scenarios around the changed endpoints still work (manual tests)

src/NuGetGallery/Controllers/PackagesController.cs
```diff
@@ -227,9 +227,8 @@ public virtual JsonResult UploadPackageProgress()
             return Json(progress, JsonRequestBehavior.AllowGet);
         }
 
-        [HttpPost]
+        [HttpGet]
```
skofman1, line 230: Why change this one to HttpGet? Doesn't it change DB state and hence should be a POST?

joelverhagen: This is a read-only endpoint. It checks for an already uploaded package (uploads container, not yet submitted) and returns the metadata view model. Otherwise, it returns an empty view model.

joelverhagen: The overloads are very confusing here TBH, but if you open it in VS and step through the code, this is the read-only endpoint that is the first step of the UI upload process. A subsequent "verify" endpoint is the POST and does the DB state changes.

zhhyu:
If it's related to the failure of functional tests, should we understand what changes make it failed recently but succeed before? Rather than suppress it?

lyndaidaii:
Deployed to dev,

E2E tests pass
Gallery functional tests pass
manual test the endpoint involved changes

Issue #8186, opened 2020-09-03
[A11y][Reflow]links like 'asp.netcore' and the other links are missing when we resize the page to 150% and 175 %


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1184498
## Browser Details:
Firefox Version: 79.0 (64 bit)

## OS Details

Microsoft Windows 10 Enterprise  ​

Version 2004 (OS Build 19041.450)

## Repro Steps:

    Zoom the page to 150%
    Hit the URL and login with appropriate credentials
    Tab till search box, type "JSON" and press enter
    Check whether all the links are clearly visible without any content loss or not

## Actual Results:

Links like 'asp.netcore' and the other links are missing when we resize the page to 150%
   
## Expected Results:
All the links in the page should be clearly visible without any content loss when the page is resized
## User Impact:
User with low vision who need to zoom the page, may face difficultly in understanding, if the content in the page is not clearly visible.

## Notes
The row that contains the download counts, last updated, and tags will cut off with ellipses when the page width is not sufficient to contain all the information. The recommendation here is probably to make it wrap. Note that this occurs even without resizing on packages that contain a lot of tags (see Microsoft.IdentityModel.JsonWebTokens)

agr:
![image](https://user-images.githubusercontent.com/102933/93525172-76f49e80-f8ea-11ea-8c4e-d60968d6da3b.png)

PR #8274, opened 2020-10-17
Enabling word wrapping for tags in all layouts

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/8186

Currently, unless we are in "smartphone" layout tags, package details (download count, last updated, latest version, tags) are forced into single line, which can lead to tags not showing at all (if the latest version is very long) or tag list cut at different places depending on zoom and window size.

The code changes the behavior to always showing the first 10 tags wrapping on the second (and third) lines if needed, making the amount of displayed content constant. "More tags" link is shown when there are more than 10 tags. The link points to a package details page. Dividers that we have between items were removed as they don't look good when items are wrapped onto the next line.

Screenshots:

image
image


ryuyu (approved):
Love it.

joelverhagen (approved):
Consider updating the description to match the final state for posterity. It mentions a chevron, but I don't see that in the code or screenshot.

PR #8275, opened 2020-10-18
[Cookies] Expire the AB test cookie in a year

zhhyu:
We should expire any cookie in 13 months, no matter whether it's in EU or not.
We upgrade the cookie to V3, in order to handle these already dropped cookies.

### DIFF
--- a/src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
+++ b/src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
@@ -17,6 +17,7 @@ namespace NuGetGallery
     {
         private const int SchemaVersion1 = 1; // PreviewSearch: {"v":1,"ps":100}
         private const int SchemaVersion2 = 2; // PreviewSearch + PackageDependent: {"v":2,"ps":100,"pd":100}
+        private const int SchemaVersion3 = 3; // PreviewSearch + PackageDependent: {"v":2,"ps":100,"pd":100}, and expired in a year.
 
         // Note that a new schema version could theoretically reuse any currently unused cookie properties. However
         // this does have questionable statistical correctness due treatment assignment of one A/B test being reused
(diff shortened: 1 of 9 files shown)


joelverhagen (approved):

joelverhagen:
What about existing cookies on people's devices with an expiration that's too long?

zhhyu:
> What about existing cookies on people's devices with an expiration that's too long?

Good question. I create a tracking issue for this: https://github.com/NuGet/Engineering/issues/3463
We should be able to reset the expiration time as one year by checking its expiration time.

joelverhagen:
> We should be able to reset the expiration time as one year by checking its expiration time.

AFAIK expiration time is not sent to the server.

zhhyu:
> > We should be able to reset the expiration time as one year by checking its expiration time.
> 
> AFAIK expiration time is not sent to the server.



> > We should be able to reset the expiration time as one year by checking its expiration time.
> 
> AFAIK expiration time is not sent to the server.

I have tested it, and I think you are right. We may need to confirm to see whether we are allowed to renew the cookie.

loic-sharma (approved):

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
```diff
@@ -140,11 +143,45 @@ private bool TryDeserializeStateVer2(string serialized, out ABTestEnrollment enr
...
+
+                return true;
+            }
+            catch (JsonException)
+            {
+                return false;
+            }
+        }
+
+        private bool TryDeserializeStateVer3(string serialized, out ABTestEnrollment enrollment)
```
joelverhagen, line 165: Schema version 2 and 3 are the same exception the version property. Would it be possible to share more code? Even the  `StateVersion2` model could be renamed to `StateVersion2And3`.

joelverhagen (approved):
Try to share more code, but it's pretty much fine as-is.

src/NuGetGallery/Infrastructure/ABTestEnrollmentFactory.cs
zhhyu: I have combined StateVersion2 and StateVersion3. I prefer to leave others as it is now. I thought about whether it's better to fully integrate version 2 and version3, but I feel that it's more readable and maintainable as the current one. Let me know if you have other thoughts here.

PR #8276, opened 2020-10-18
[Cookies] Comment for local debugging

zhhyu:
Add a comment for local debugging


src/NuGetGallery.Core/Cookies/CookieExpirationService.cs
```diff
@@ -78,6 +78,9 @@ public void ExpireCookieByName(HttpContextBase httpContext, string cookieName, s
             {
                 response.Cookies[cookieName].Expires = CookieExpirationTime;
 
+                // Uncomment the line below when you debug this service locally
+                // response.Cookies[cookieName].Secure = false;
```
joelverhagen, line 82: this this be toggled off of the force SSL configuration?

Issue #8196, opened 2020-09-03
[A11y][ARIA]NVDA/Narrator is reading about the data which is not present on the screen at the table "Version History"


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1188406
## Browser Details:
Firefox Version: 79.0 (64 bit)

## OS Details

Microsoft Windows 10 Enterprise  ​

Version 2004 (OS Build 19041.450) 

## Repro Steps:

    Turn on NVDA
    Hit the URL and login with appropriate credentials
    Tab till "Profile" button in the menu bar and press enter
    Tab till any Manage packages"  and press enter.
    Tab till search bar and search for 'Json' and press enter 
    Tab till the first result "Newtonsoft.Json" and press enter 
    Tab till "Version" of any result 
    Tab till "Version history" and press enter to expand
    Verify whether the NVDA is announcing the blank columns and allows navigation with CTRL + ALT + Arrow keys

## Actual Results:

NVDA/Narrator is reading about the data which is not present on the screen at the table "Version History"

   
## Expected Results:
Data should be visible on the screen.

Narrator/NVDA should read correct data which is present in the table and which is visible on the screen

Note: when data is not present in any column it would be better if screen reader announce it as empty cell.

## User Impact:

Screen reader users may not be understand about the columns in the table and also get confused with navigation 

## Notes
In a surprise twist, we are reading out too much information. This seems most likely connected to the fact that the table has columns that are not visible at all times (status, signature information, deprecation information). For cases where this information isn't present, it should not be in the table either (if it should never be presented). If it could possibly be presented, but it isn't filled in, we should indicate that the cell is empty.

PR #8287, opened 2020-10-21
Version history table rendering update for accessibility

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/8196

Previously, version history table always had 2 hardly visible columns: 1 for signature information and another for deprecation information. When user is not authenticated on the web site, we never show the package signature information, so the cells although exist are always empty. The deprecation information is present, but most of the time packages are not deprecated and empty cells do not have any narrator markup, so are not read (but headers are).

This change moves the `Model.IsCertificatesUIEnabled` check up to the lines that generate column header, causing the certificate column to be completely absent for unauthenticated users. Same is done for `Model.IsPackageDeprecationEnabled` but it is always `true` in our case. Also `aria-label` properties added for empty cells when those are rendered.

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -742,15 +742,15 @@
 
                                     @if (Model.IsCertificatesUIEnabled)
                                     {
-                                        if (!string.IsNullOrEmpty(packageVersion.SignatureInformation))
+                                        if (string.IsNullOrEmpty(packageVersion.SignatureInformation))
+                                        {
+                                            <td class="package-icon-cell" aria-label="No signature"></td>
+                                        }
+                                        else
                                         {
                                             <td class="package-icon-cell">
                                                     <i class="ms-Icon ms-Icon--Ribbon package-icon" title="@packageVersion.SignatureInformation"></i>
                                             </td>
-                                        }
-                                        else
-                                        {
-                                            <td class="package-icon-cell" aria-label="No signature"></td>
                                         }
                                     }
                                     @if (Model.IsPackageDeprecationEnabled)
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -723,7 +729,7 @@
 
                                         <td>
                                             @if (packageStatusSummary == PackageStatusSummary.Listed ||
-                                                 packageStatusSummary == PackageStatusSummary.Unlisted)
+     packageStatusSummary == PackageStatusSummary.Unlisted)
```
loic-sharma, line 732: Was this change intentional? I prefer the old styling :)

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -734,15 +740,22 @@
...
-                                    <td class="package-icon-cell">
-                                        @if (Model.IsPackageDeprecationEnabled && packageVersion.DeprecationStatus != PackageDeprecationStatus.NotDeprecated)
+                                        else
+                                        {
+                                            <td class="package-icon-cell" aria-label="No signature"></td>
+                                        }
+                                    }
+                                    @if (Model.IsPackageDeprecationEnabled)
+                                    {
+                                        if (packageVersion.DeprecationStatus != PackageDeprecationStatus.NotDeprecated)
```
loic-sharma, line 758: Nit: consider flipping this condition as the `else` block is much smaller. I don't feel strongly about this though

loic-sharma (approved):

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
cristinamanum: I agree it is easier to read packageVersion.DeprecationStatus == PackageDeprecationStatus.Deprecated than two negations.

cristinamanum (approved):

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
agr: VS messes up formatting in this file really bad when I edit it.

zhhyu (approved):

Issue #8286, opened 2020-10-20
[A11y][WCAG2.1][Contrast]The color contrast ratio for the stars under "Stars" column is < 3:1


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1069750

# Environment Details:
#URL:https://www.nuget.org/​
 ​
## OS Details:​
Windows 10 Version 1909 Build 18363.657​
​
## Browser and Version:​
Chromium Edge Version 80.0.361.51​
​
# Repro Steps:​
1.Hit the URL "https://www.nuget.org" 
2.Tab till "Packages" link in the top navigation and press enter​
3.Tab till any existing package and press enter​
4.Tab till "GitHub Usage" section and press enter​
5.Using color contrast analyzer, verify the color contrast ratio for the stars under "Stars" column header
​
## Actual Result:​
The color contrast ratio for the stars under "Stars" column is 2.51:1 which is < 3:1

## Expected Result:​
The color contrast ratio for non-text content should be >= 3:1
​
## User Impact:​
Users with visual impairments will find it difficult to access the content
​
# Notes:
Depending on how these stars are rendered, this will either be very simple, or very not simple.

PR #8288, opened 2020-10-21
[A11y] Darken the GitHub star icon

loic-sharma:
Darkens the orange GitHub star icon for 3:1 color contrast ratio.

image

Addresses https://github.com/NuGet/NuGetGallery/issues/8286

### DIFF
--- a/src/Bootstrap/less/theme/page-display-package.less
+++ b/src/Bootstrap/less/theme/page-display-package.less
@@ -171,7 +171,7 @@
   .gh-star {
     font-size: 14px;
     line-height: 12px;
-    color: @brand-warning;
+    color: darken(@brand-warning, 4.5%);
   }
 
   .used-by {
--- a/src/Bootstrap/less/theme/page-display-package.less
+++ b/src/Bootstrap/less/theme/page-display-package.less
@@ -171,6 +171,7 @@
   .gh-star {
     font-size: 14px;
     line-height: 12px;
+    /* Use a slightly darker orange color to increase contrast for accessibility */
     color: darken(@brand-warning, 4.5%);
   }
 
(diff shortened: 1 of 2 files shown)


src/Bootstrap/less/theme/page-display-package.less
```diff
@@ -171,7 +171,7 @@
   .gh-star {
     font-size: 14px;
     line-height: 12px;
-    color: @brand-warning;
+    color: darken(@brand-warning, 4.5%);
```
agr, line 174: Should we introduce some "standard" darkening values/constants? 4.5% seems rather random...

loic-sharma: 4.5% was chosen as it was the lowest value that let us reach the 3:1 ratio. I don't think adding a variable would give us that much value, but please let me know if you feel strongly about this

agr: Should we just hardcode some specific color then? Darkening `@brand-warning` by 4.5% specifically looks really arbitrary. At least, please add a comment explaining what's happening here.

agr (approved):

zhhyu (approved):

src/Bootstrap/less/theme/page-display-package.less
loic-sharma: Added a comment. FYI, we've heavily used the `darken` pattern to increase the contrast ratio in previous instances. Examples:

https://github.com/NuGet/NuGetGallery/blob/d7aaa8dd8aa213f77ca61d88eee1c4c7df2ed2d9/src/Bootstrap/less/variables.less#L38-L39

Or:

https://github.com/NuGet/NuGetGallery/blob/7e6f309c08e2f6316c3e6d6a40092a128f15d374/src/Bootstrap/less/theme/page-list-packages.less#L88

You can find lots more if you search for `darken` 🙂

PR #8290, opened 2020-10-22
Add restricted API for soft deleting packages 

joelverhagen:
This PR extends the existing `DELETE /api/v2/package/{ID}/{version}` API to allow the soft deletion of packages. Currently, this endpoint can be used by any unlist scoped API key to unlist a package. This endpoint is mainly used by the `nuget.exe delete` and `dotnet nuget delete` commands to unlist packages from the command line.

The soft delete behavior is triggered by including a request body in the `DELETE` that has the following shape:
```json
{
  "Type": "SoftDelete"
}
```

If no request body is included, the `"Type"` property defaults to `"Unlist"`. This maintains backwards compatibility. In other words, the default `DELETE` behavior for nuget.org remains to unlist but a client can opt-in to the soft delete behavior with a request body parameter.

The intent of this API is to enable automation for test data clean up, i.e. the packages published by NugetTestAccount during gallery functional tests and E2E tests. 

The soft delete gesture is gated on the following conditions:
1. *(existing, same as unlist)* The target package must exist. You can still operate on a soft deleted package.
1. *(existing, same as unlist)* The API key must have the "unlist" scope.
1. *(existing, same as unlist)* The API key must have a package ID scope matching the target package.
1. *(existing, same as unlist)* The package ID must not be locked.
1. **(new, only for soft delete)** The user must be in the `NuGetGallery.DeletePackageApi` flight.
1. **(new, only for soft delete)** The user must _not_ be an admin. I don't want the risk of deleting stuff unintentionally.
1. **(new, only for soft delete)** The package download count must not be greater than or equal to 1000.

I will present a full design before moving forward, but generally my plan is to roll this out and write a simple task in the `Gallery.Maintenance` job to call this delete API on packages owned by `NugetTestAccount`.

I chose to make this a REST API so that the complex, multistep delete logic is deployed only to a single location (gallery front-end) instead of multiple places (gallery front-end and some delete job). I have no intent of opening this API up to anyone but `NugetTestAccount`, but the code is written in such a way that it could be extended to future scenarios.


loic-sharma:
> The user must not be an admin. I don't want the risk of deleting stuff unintentionally.

Could you expand your reasoning here? From my understanding an admin would need to manually craft a request to call this API, which seems unlikely to happen by accident.

src/NuGetGallery.Services/Configuration/IFeatureFlagService.cs
```diff
@@ -225,5 +225,10 @@ public interface IFeatureFlagService
         /// Whether the /Search()/$count endpoint is enabled for non-hijacked queries for the V2 OData API.
         /// </summary>
         bool IsODataV2SearchCountNonHijackedEnabled();
+
+        /// <summary>
+        /// Whether or not the user delete a package through the API.
```
loic-sharma, line 230:
```suggestion
        /// Whether or not the user can delete a package through the API.
```

joelverhagen:
> Could you expand your reasoning here?

The plan for API is just for test data clean-up. There's no need for having more access. Personally, I like the current admin delete flow where you have to type your signature and click a confirmation pop-up before a delete goes through. These safeguards don't exist for an API. As far as I know, we have not needed an API for an administrative package deletion so however small the risk I don't see value in allowing it. If the need comes up in the future we can easily remove this check after we have considered the non-zero risk.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -884,7 +887,55 @@ public virtual async Task<ActionResult> DeletePackage(string id, string version)
...
+                        return new HttpStatusCodeWithBodyResult(
+                            HttpStatusCode.Forbidden,
+                            string.Format(CultureInfo.CurrentCulture, Strings.DeletePackage_TooManyDownloads, downloadCountLimit));
+                    }
+
+                    await PackageDeleteService.SoftDeletePackagesAsync(
+                        new[] { package },
+                        currentUser,
+                        PackageDeleteApiReason,
+                        currentUser.Username);
```
loic-sharma, line 932: Just double-checking is it ok to store PII here? Sorry for being pedantic :)

loic-sharma (approved):

src/NuGetGallery/Controllers/ApiController.cs
joelverhagen: It's a good question, not pedantic at all. I think it would be simplest to just put a constant value here. The delete record is linked to the user record so baking the username into the record doesn't add much.

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -884,7 +887,55 @@ public virtual async Task<ActionResult> DeletePackage(string id, string version)
...
 
-            await PackageUpdateService.MarkPackageUnlistedAsync(package);
+            var currentUser = GetCurrentUser();
+            DeletePackageAction action;
+            if (request?.Type == null)
+            {
+                action = DeletePackageAction.Unlist;
+            }
+            else if (!DeletePackageApiRequest.Actions.TryGetValue(request?.Type, out action))
+            {
```
skofman1, line 897: Can we move this logic to DeletePackageApiRequest class?

skofman1 (approved):
:shipit:

src/NuGetGallery/Controllers/ApiController.cs
joelverhagen: Sure, done.

Issue #8268, opened 2020-10-16
Dependencies not showing platform component of .NET 5 TFM


clairernovotny:
The gallery is only showing `NetCoreApp 5.0` under the dependencies node and not `net5.0-windows10.19041`. 

I have a package that multi-targets both `net5.0` and `net5.0-windows10.19041`. I'd expect both to show up on the website; they are distinct dependency groups in the NuSpec:

![image](https://user-images.githubusercontent.com/1427284/96260782-7acd1c80-0f8d-11eb-8394-678d2728b471.png)

joelverhagen:
The fix is now in PROD. Thanks for bringing this to our attention, @clairernovotny!

https://www.nuget.org/packages/System.Reactive/5.0.0-preview.220
![image](https://user-images.githubusercontent.com/94054/97733625-585ef700-1a95-11eb-98a0-e7c8219dafc2.png)

Issue #8235, opened 2020-10-04
.NET 5.0 not .NETFramework 5.0


f2calv:
### Describe the bug
When viewing dependencies on the web site for a package that targets .NET 5.0 the UI reads '.NETFramework 5.0' which I believe is incorrect.

### To Reproduce
Publish a package that targets .NET 5.0

### Expected behavior
Should display ".NET 5.0" instead of ".NETFramework 5.0"

### Screenshots
![image](https://user-images.githubusercontent.com/16097639/95006735-9f2d0e80-0607-11eb-8b15-d3be2f25a535.png)

joelverhagen:
Great catch, @f2calv! Essentially the problem here is that NuGet.org is using an older version of the framework definitions prior to the introduction of .NET 5.0.

We'll need to update the newest version of these definitions and reprocess packages that have .NET 5.0 dependencies.

/cc @zkat @nkolev92, is there a version of NuGet.Frameworks that we should be using for stable .NET 5.0 framework parsing?

Note to the server team: I think feed2catalog and perhaps even catalog2registration need to upgrade as well because they interact with dependency group metadata as well.

/cc @loic-sharma, who recently updated the package versions used on the server side. We may already have the version we need and just need to deploy it then reflow the packages.

My guess is other package feeds will run into similar problems. It may be worth an announcement about the introduction of the TFM and how it interacts with NuGet client and server implementations.

nkolev92:
@joelverhagen 
5.8 early previews have *all* the changes necessary.

Anything above 5.8.0 P2 should work, but prefer the latest.

zivkan:
Note to server team/whoever makes the fix:

`NuGetFramwork.DotNetFrameworkName` is designed explicitly to be compatible with `System.Runtime.Versioning.FrameworkName`. As per the [.NET 5 TFM spec](https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md), this is not the recommended string for display.  Please use `NuGetFramework.ToString()` to get the recommended display name.

joelverhagen:
Here are several variants and how they behave on nuget.org today.

### `net5.0`

Example package with `net5.0` in the dependency group:
https://www.nuget.org/packages/Microsoft.AspNetCore.Authorization/5.0.0-rc.1.20451.17
```xml
      <group targetFramework="net5.0">
[...]

PR #8291, opened 2020-10-22
Display net5.0 using ToString instead of custom logic

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/8235
Address https://github.com/NuGet/NuGetGallery/issues/8268

Instead of using the custom framework display logic that nuget.org currently has for framework names in dependency groups, use `NuGetFramework.ToString()`. This is the recommended approach from client team and aligns with the package management UI in Visual Studio. The VS UI is spec'd here which we are mimicking.
https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md

> - Package Manager UI in VS
>    - [Use short form](https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md#tfms-in-the-ui)

### Example without platform

image

### Example with platform

image

### DIFF
--- a/src/NuGetGallery/ExtensionMethods.cs
+++ b/src/NuGetGallery/ExtensionMethods.cs
@@ -243,6 +243,15 @@ namespace NuGetGallery
                 throw new ArgumentNullException(nameof(frameworkName));
             }
 
+            // Defer to the NuGet client logic for displaying .NET 5 frameworks. This aligns with Visual Studio package
+            // management UI.
+            var isNet5Era = frameworkName.Version.Major >= 5
+                && StringComparer.OrdinalIgnoreCase.Equals(FrameworkConstants.FrameworkIdentifiers.NetCoreApp, frameworkName.Framework);
+            if (isNet5Era)
+            {
+                return frameworkName.ToString();
+            }
+
             var sb = new StringBuilder();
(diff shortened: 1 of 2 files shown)


skofman1 (approved):
:shipit:

skofman1:
>         public static string ToFriendlyName(this NuGetFramework frameworkName, bool allowRecurseProfile = true)

Why do we need Gallery specific code if it's already covered by NuGetFramework.ToString()?

---
Refers to: src/NuGetGallery/ExtensionMethods.cs:239 in a338cdd. [](commit_id = a338cdd4a5e3479981239af2577876cc9e31dccd, deletion_comment = False)

joelverhagen:
> Why do we need Gallery specific code if it's already covered by NuGetFramework.ToString()?

This is existing behavior. Why we have it is lost to history. I don't see anything helpful in the code comments.

We can certainly investigate changing it, but if we use `ToString()` we will start seeing a heading
> ## .NETFramework,Version=v4.6.1
instead of today's 
> ## .NETFramework 4.6.1 

Again, this would align with VS PM but in my opinion this change is out of scope for the .NET 5 bug. Frankly, I personally think we should show the short folder name everywhere (e.g. `net461`) to be self-consistent but this is also different from PM UI and would use `GetShortFolderName()` instead of `ToString()`.

loic-sharma (approved):

PR #8292, opened 2020-10-23
Markdig version update

lyndaidaii:
Update markdig with latest version to have all new features

- Update markdig to latest version 0.22.0

- New extension added: 
   table, tasklist, autolink, emoji

- Feature flag was added

Below showed the table, autolink, emoji render difference with two libs
Old CommonMark render: 
NuGet Gallery _ Manage Package ReadmeWithURLTest33

New Markdig render: 
NuGet Gallery _ Manage Package xiaodaidaitest

Details comparison about Markdig and CommonMark, please check out OneNote: Markdig vs CommonMark 

Addresses https://github.com/NuGet/Engineering/issues/3429


loic-sharma (approved):
What's the update for? FYI, the issue link seems to be incorrect

lyndaidaii:
Tested with updated markdig version in serverCommon, need update package version with NuGet.Services.Messaging.Email, then  merge this after change in serverCommon

joelverhagen:
Please verify the account deletion emails in particular. The account deleter is in the strange position to refer to ServerCommon, NuGet.Jobs, and NuGetGallery libraries all at once. There can be some strange dependency issues because of that.

joelverhagen (approved):

lyndaidaii:
Would have separate PR for migrating functionality using markdig, I will revert the last commit "migrate to markdig" after test

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -110,5 +122,72 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+
+        public RenderedMarkdownResult GetHtmlFromMarkdownMarkdig(string markdownString, int incrementHeadersBy)
+        {
+            var output = new RenderedMarkdownResult()
+            {
+                ImagesRewritten = false,
+                Content = ""
+            };
+
+            var readmeWithoutBom = markdownString.StartsWith("\ufeff") ? markdownString.Replace("\ufeff", "") : markdownString;
```
joelverhagen, line 134: Why not just `TrimStart`? Replacing throughout the entire string seems like too much.

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
```diff
@@ -46,6 +46,35 @@ public void ConvertsMarkdownToHtml(string originalMd, string expectedHtml, bool
...
+            [InlineData("> <text>Blockquote</text>", "<blockquote><p>&lt;text&gt;Blockquote&lt;/text&gt;</p></blockquote>", false)]
+            [InlineData("[text](http://www.asp.net)", "<p><a href=\"https://www.asp.net/\" rel=\"nofollow\">text</a></p>", false)]
+            [InlineData("[text](badurl://www.asp.net)", "<p><a href=\"\" rel=\"nofollow\">text</a></p>", false)]
+            [InlineData("![image](http://www.asp.net/fake.jpg)", "<p><img src=\"https://www.asp.net/fake.jpg\" alt=\"image\" /></p>", true)]
+            [InlineData("![image](https://www.asp.net/fake.jpg)", "<p><img src=\"https://www.asp.net/fake.jpg\" alt=\"image\" /></p>", false)]
+            [InlineData("![image](http://www.otherurl.net/fake.jpg)", "<p><img src=\"https://www.otherurl.net/fake.jpg\" alt=\"image\" /></p>", true)]
+            public void ConvertsMarkdownToHtmlMarkdig(string originalMd, string expectedHtml, bool imageRewriteExpected)
+            {
+                var readMeResult = _markdownService.GetHtmlFromMarkdownMarkdig(originalMd);
+                Assert.Equal(expectedHtml, StripNewLines(readMeResult.Content));
```
joelverhagen, line 74: Why do we have `StripNewLines` to "clean" the readme result? Can't we just include the proper characters in the test data? In my opinion, unit tests should not hide behavior like this.

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -110,5 +122,72 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
                 return output;
             }
         }
+
+        public RenderedMarkdownResult GetHtmlFromMarkdownMarkdig(string markdownString, int incrementHeadersBy)
+        {
```
cristinamanum, line 127: Any argument validation needed?

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -110,5 +122,72 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
                 return output;
             }
         }
+
+        public RenderedMarkdownResult GetHtmlFromMarkdownMarkdig(string markdownString, int incrementHeadersBy)
```
cristinamanum, line 126: This method seems to be used only [here](https://github.com/NuGet/NuGetGallery/blob/0e925e775abdd5a8767170c9689b61d54c95a627/src/NuGetGallery/Services/MarkdownService.cs#L31). Is the parametrization of the "incrementHeadersBy" needed?

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -110,5 +122,72 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+
+            var document = Markdown.Parse(encodedMarkdown, pipeline);
+
+            foreach (var node in document.Descendants())
+            {
+                if (node is Markdig.Syntax.Block)
+                {
+                    if (node is HeadingBlock heading)
+                    {
+                        heading.Level = (byte)Math.Min(heading.Level + incrementHeadersBy, 6);
```
cristinamanum, line 149: Maybe add a comment why 6?

src/NuGetGallery/Services/IMarkdownService.cs
```diff
@@ -19,5 +19,20 @@ public interface IMarkdownService
...
+        /// <returns>HTML data</returns>
+        RenderedMarkdownResult GetHtmlFromMarkdownMarkdig(string markdownString);
+
+        /// <summary>
+        /// Returns HTML from the supplied markdown using Markdig
+        /// </summary>
+        /// <param name="markdownString">markdown input</param>
+        /// <param name="incrementHeadersBy">headers can be incremented by this value, eg if 2 supplied then h1 will become h3</param>
+        /// <returns>HTML data</returns>
+        RenderedMarkdownResult GetHtmlFromMarkdownMarkdig(string markdownString, int incrementHeadersBy);
```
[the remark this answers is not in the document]
lyndaidaii: We eventually would remove the  implementation of the GetHtmlFromMarkdown after we do the test comparing those two library

src/NuGetGallery/Services/MarkdownService.cs
lyndaidaii: Current we support adaptive header, in order to continue support the same functionality, I believe we still need "incrementHeaderBy"

src/NuGetGallery/Services/IMarkdownService.cs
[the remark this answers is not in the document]
cristinamanum: Did we choose the name `GetHtmlFromMarkdownMarkdig` due to the fact that Markdig is used underneath? Will this api be invalid if we move to CommonMark? If the answer is yes for the above maybe we can find a name that is not related with the library underneath.

zhhyu:
Will there be any rendering/translation difference between "Markdig" and "CommonMark"? (different grammar supports, different rendering results, etc.) How do we test it? Maybe we'd like to have a feature flag here, in case that we need to roll back easily? This change may potentially break the current rendering with customer impacts.

zhhyu:
>         private static readonly Regex CommonMarkLinkPattern = new Regex("<a href=([\"\']).*?\\1", RegexOptions.None, RegexTimeout);

Should this be also updated?

---
Refers to: src/NuGetGallery/Services/MarkdownService.cs:24 in 76109a0. [](commit_id = 76109a012dd9ce203a42a5f1f067162cba3f8f89, deletion_comment = False)

zhhyu:
>         private static readonly Regex CommonMarkLinkPattern = new Regex("<a href=([\"\']).*?\\1", RegexOptions.None, RegexTimeout);

Should this be also updated?

---
Refers to: src/NuGetGallery/Services/MarkdownService.cs:24 in 76109a0. [](commit_id = 76109a012dd9ce203a42a5f1f067162cba3f8f89, deletion_comment = False)

src/NuGetGallery/Services/IMarkdownService.cs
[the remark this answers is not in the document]
zhhyu: Personally, I feel that maybe it's better that we put a feature flag and switch between the new one and the old "CommonMark" rending. We can have another PR to remove "CommonMark" implementations, once we stabilize and have the confidence to use "Markdig".

zhhyu:
Talk offline, we need to compare the result against readmes files we currently hold to ensure that there will be no broken customer experience.

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -3,10 +3,17 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using System.Text.RegularExpressions;
+using System.Timers;
 using System.Web;
 using CommonMark;
 using CommonMark.Syntax;
```
loic-sharma, line 11: Can we remove CommonMark `using`s now?

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -29,86 +36,67 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
-            // CommonMark.Net does not support link attributes, so manually inject nofollow.
-            using (var htmlWriter = new StringWriter())
-            {
-                CommonMarkConverter.ProcessStage3(document, htmlWriter, settings);
-
-                output.Content = CommonMarkLinkPattern.Replace(htmlWriter.ToString(), "$0" + " rel=\"nofollow\"").Trim();
-                return output;
-            }
+            StringWriter htmlWriter = new StringWriter();
+            var renderer = new HtmlRenderer(htmlWriter);
```
loic-sharma, line 94: `StringWriter` is disposable, please wrap that in a `using`. Please do the same for `HtmlRenderer` if it is also disposable

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -29,86 +36,67 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+            renderer.Render(document);
+            htmlWriter.Flush();
+            //manually inject nofollow since markdig doesn't support inject nofollw in encode 
+            output.Content = CommonMarkLinkPattern.Replace(htmlWriter.ToString(), "$0" + " rel=\"nofollow\"").Trim();
```
loic-sharma, line 98: It looks like this is supported by Markdig by calling `UseReferralLinks("nofollow")` on `MarkdownPipelineBuilder`. See: https://github.com/lunet-io/markdig/blob/930eee383f97219b78bcdd1e92c9c7e2a52d68a8/src/Markdig/MarkdownExtensions.cs#L460-L464

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -3,25 +3,59 @@
...
+        {
+            if (_features.IsMarkdigMdRenderingEnabled())
+            {
+                return GetHtmlFromMarkdownMarkdig(markdownString, incrementHeadersBy);
+            }
+            else
+            {
+                return GetHtmlFromMarkdownCommonMark(markdownString, incrementHeadersBy);
+            }
+
```
lyndaidaii, line 55: delete this line

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -106,9 +140,102 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+                        LeafBlock codeBlock = (LeafBlock)node;
+                        var lines =codeBlock.Lines;
+                        for (int i = 0; i < lines.Count; i++) 
+                        {
+                            var content = lines.Lines[i].Slice.ToString();
+                            var unencodedContent = HttpUtility.HtmlDecode(content);
+                            lines.Lines[i].Slice = new Markdig.Helpers.StringSlice(unencodedContent);
+                        }
+                    }
+
```
lyndaidaii, line 201: delete this line

lyndaidaii:
To do: 

- pull latest changes, resolve conflicts 

- update the ServerCommon

joelverhagen (approved):

src/NuGetGallery.Services/Configuration/IFeatureFlagService.cs
```diff
@@ -227,6 +227,10 @@ public interface IFeatureFlagService
         bool IsODataV2SearchCountNonHijackedEnabled();
 
         /// <summary>
+        /// Whether rending Markdown content to HTML using Markdig is enabaled
```
loic-sharma, line 230:
```suggestion
        /// Whether rending Markdown content to HTML using Markdig is enabled.
```

src/NuGetGallery/Services/MarkdownService.cs
loic-sharma: Resolving this as there is now a feature flag to switch between markdig/commonmark

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -106,7 +139,100 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+                .UsePipeTables()
+                .UseListExtras()
+                .UseTaskLists()
+                .UseSoftlineBreakAsHardlineBreak()
+                .UseEmojiAndSmiley()
+                .UseReferralLinks("nofollow")
+                .UseAutoLinks()
+                .Build();
+
+            using(StringWriter htmlWriter = new StringWriter())
```
loic-sharma, line 171: Style nit:

```suggestion
            using (var htmlWriter = new StringWriter())
```

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -106,7 +139,100 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+                        {
+                            heading.Level = (byte)Math.Min(heading.Level + incrementHeadersBy, 6);
+                        }
+
+                        // Decode preformatted blocks to prevent double encoding.
+                        // Skip BlockTag.BlockQuote, which are partially decoded upfront.
+                        if (node is FencedCodeBlock || node is CodeBlock)
+                        {
+                            LeafBlock codeBlock = (LeafBlock)node;
+                            var lines =codeBlock.Lines;
```
loic-sharma, line 193:
```suggestion
                            var lines = codeBlock.Lines;
```

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -106,7 +139,100 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+            var pipeline = new MarkdownPipelineBuilder()
+                .UseGridTables()
+                .UsePipeTables()
+                .UseListExtras()
+                .UseTaskLists()
+                .UseSoftlineBreakAsHardlineBreak()
+                .UseEmojiAndSmiley()
+                .UseReferralLinks("nofollow")
+                .UseAutoLinks()
+                .Build();
```
loic-sharma, line 169: Do we need to build a new pipeline each time or could this be a dependency injection singleton? I did a quick search and found an article that mentioned you should cache this object: https://codemag.com/Article/1811071/Marking-up-the-Web-with-ASP.NET-Core-and-Markdown

> The code works, but it’s not very efficient as the pipeline is recreated for each parse operation. It’s much better to build a small abstraction layer around the Markdown parser, so the parser can be cached for better performance.

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -106,7 +139,100 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+            var output = new RenderedMarkdownResult()
+            {
+                ImagesRewritten = false,
+                Content = ""
+            };
+
+            var readmeWithoutBom = markdownString.TrimStart('\ufeff');
+
+            // HTML encode markdown, except for block quotes, to block inline html.
+            var encodedMarkdown = EncodedBlockQuotePattern.Replace(HttpUtility.HtmlEncode(readmeWithoutBom), "> ");
```
loic-sharma, line 158: It seems Markdig can block inline HTML if you use [`DisableHtml`](https://github.com/lunet-io/markdig/blob/7964bd0160d4c18e4155127a4c863d61ebd8944a/src/Markdig/MarkdownExtensions.cs#L306) on your pipeline builder. This may let you delete this line and remove the code block decoding later.

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
```diff
@@ -1,54 +1,168 @@
...
+                var originalMd = "This is a paragraph\r\n with a break inside";
+                var expectedHtml = "<p>This is a paragraph<br />\nwith a break inside</p>";
+                _featureFlagService.Setup(x => x.IsMarkdigMdRenderingEnabled()).Returns(isMarkdigMdRenderingEnabled);
+                var readMeResult = _markdownService.GetHtmlFromMarkdown(originalMd);
+                Assert.Equal(expectedHtml, readMeResult.Content);
+                Assert.Equal(false, readMeResult.ImagesRewritten);
+            }
+
+            [Fact]
+            public void TestToHtmlWithPipeTable() {
```
loic-sharma, line 83:
```suggestion
            public void TestToHtmlWithPipeTable()
            {
```

src/NuGetGallery/Services/MarkdownService.cs
lyndaidaii: This is good point. After consideration, I think probably keep it as it now. 
Markdig render is called only package has readme. It's really small amount of percent packages contain readme. 
Aslo from the performance report show now, it doesn't show big negative impact in terms of CPU and Response time. 
However, I will create ticket to track this issue, possibly is an enhancement.

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -3,25 +3,58 @@
...
+
+        private readonly IFeatureFlagService _features;
+
+        public MarkdownService(IFeatureFlagService features)
+        {
+            _features = features ?? throw new ArgumentNullException(nameof(features));
+        }
 
         public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString)
         {
```
zhhyu, line 34: Null check for a public method is needed.

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -3,25 +3,58 @@
...
+                return GetHtmlFromMarkdownMarkdig(markdownString, 1);
+            }
+            else
+            {
+                return GetHtmlFromMarkdownCommonMark(markdownString, 1);
+            }
         }
 
         public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int incrementHeadersBy)
+        {
```
zhhyu, line 46: Null check for a public method is needed.

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
```diff
@@ -1,54 +1,169 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Moq;
+using System.Data.Entity.Core.Objects;
```
zhhyu, line 5: nit: system library at the very top

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -106,7 +139,84 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+                .UseListExtras()
+                .UseTaskLists()
+                .UseSoftlineBreakAsHardlineBreak()
+                .UseEmojiAndSmiley()
+                .UseReferralLinks("nofollow")
+                .UseAutoLinks()
+                .DisableHtml() //block inline html
+                .Build();
+
+            using(var htmlWriter = new StringWriter())
```
loic-sharma, line 169: Nit:

```suggestion
            using (var htmlWriter = new StringWriter())
```

loic-sharma (approved):

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
```diff
@@ -1,54 +1,169 @@
...
 {
     public class MarkdownServiceFacts
     {
-       
+
         public class GetReadMeHtmlMethod
         {
             private readonly MarkdownService _markdownService;
+            private readonly Mock<IFeatureFlagService> _featureFlagService;
+            private bool isMarkdigMdRenderingEnabled = true;
```
zhhyu, line 17: What is the benefit to introduce this field. For Fact tests, you can directly pass true, is it?

lyndaidaii: Make readable instead of what true means

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -106,7 +139,84 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+            var readmeWithoutBom = markdownString.TrimStart('\ufeff');
+
+            var pipeline = new MarkdownPipelineBuilder()
+                .UseGridTables()
+                .UsePipeTables()
+                .UseListExtras()
+                .UseTaskLists()
+                .UseSoftlineBreakAsHardlineBreak()
+                .UseEmojiAndSmiley()
+                .UseReferralLinks("nofollow")
```
zhhyu, line 164: Maybe it's better to use "UseNoFollowLinks".

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -106,7 +139,84 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+
+                foreach (var node in document.Descendants())
+                {
+                    if (node is Markdig.Syntax.Block)
+                    {
+                        // Demote heading tags so they don't overpower expander headings.
+                        if (node is HeadingBlock heading)
+                        {
+                            heading.Level = (byte)Math.Min(heading.Level + incrementHeadersBy, 6);
+                        }
```
zhhyu, line 184: Is there any reason to use "byte" here? I see that "heading.Level" is int.

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
```diff
@@ -1,54 +1,169 @@
...
+                var readMeResult = _markdownService.GetHtmlFromMarkdown(originalMd);
+                Assert.Equal(expectedHtml, readMeResult.Content);
+                Assert.Equal(false, readMeResult.ImagesRewritten);
+            }
+
+            public void TestToHtmlWithAutoLinks()
+            {
+                var originalMd = "This is a http://www.google.com URL and https://www.google.com";
+
+                var expectedHtml = "<p>This is a <a href=\"https://www.google.com\">https://www.google.com</a> URL and <a href=\"https://www.google.com\">https://www.google.com</a>";
```
zhhyu, line 162: Should auto links also use "nofollow"?

src/NuGetGallery/Services/MarkdownService.cs
zhhyu, line 46: Could we have some test coverages on this method given any "incrementHeadersBy"?

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -106,7 +139,84 @@ public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int inc
...
+                    }
+                    else if (node is Markdig.Syntax.Inlines.Inline)
+                    {
+                        if (node is LinkInline linkInline)
+                        {
+                            if (linkInline.IsImage)
+                            {
+                                // Allow only http or https links in markdown. Transform link to https for known domains.
```
zhhyu, line 192: Should this comment be used for links? I see that we always apply HTTPS for images, even though I am not sure why we design it in this way.

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
```diff
@@ -1,54 +1,169 @@
...
+            [InlineData("![image](http://www.asp.net/fake.jpg)", "<p><img src=\"https://www.asp.net/fake.jpg\" alt=\"image\" /></p>", true, true)]
+            [InlineData("![image](http://www.asp.net/fake.jpg)", "<p><img src=\"https://www.asp.net/fake.jpg\" alt=\"image\" /></p>", true, false)]
+            [InlineData("![image](https://www.asp.net/fake.jpg)", "<p><img src=\"https://www.asp.net/fake.jpg\" alt=\"image\" /></p>", false, true)]
+            [InlineData("![image](https://www.asp.net/fake.jpg)", "<p><img src=\"https://www.asp.net/fake.jpg\" alt=\"image\" /></p>", false, false)]
+            [InlineData("![image](http://www.otherurl.net/fake.jpg)", "<p><img src=\"https://www.otherurl.net/fake.jpg\" alt=\"image\" /></p>", true, true)]
+            [InlineData("![image](http://www.otherurl.net/fake.jpg)", "<p><img src=\"https://www.otherurl.net/fake.jpg\" alt=\"image\" /></p>", true, false)]
+            [InlineData("## License\n\tLicensed under the Apache License, Version 2.0 (the \"License\");", "<h3>License</h3>\n<pre><code>Licensed under the Apache License, Version 2.0 (the &quot;License&quot;);\n</code></pre>", false, true)]
```
zhhyu, line 61: nice!

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
```diff
@@ -1,54 +1,169 @@
...
+            [InlineData("\ufeff# Heading with BOM", "<h2>Heading with BOM</h2>", false, true)]
+            [InlineData("\ufeff# Heading with BOM", "<h2>Heading with BOM</h2>", false, false)]
+            [InlineData("- List", "<ul>\n<li>List</li>\n</ul>", false, true)]
+            [InlineData("- List", "<ul>\r\n<li>List</li>\r\n</ul>", false, false)]
+            [InlineData("[text](http://www.test.com)", "<p><a href=\"http://www.test.com/\" rel=\"nofollow\">text</a></p>", false, true)]
+            [InlineData("[text](http://www.test.com)", "<p><a href=\"http://www.test.com/\" rel=\"nofollow\">text</a></p>", false, false)]
+            [InlineData("[text](javascript:alert('hi'))", "<p><a href=\"\" rel=\"nofollow\">text</a></p>", false, true)]
+            [InlineData("[text](javascript:alert('hi'))", "<p><a href=\"\" rel=\"nofollow\">text</a></p>", false, false)]
+            [InlineData("> <text>Blockquote</text>", "<blockquote>\n<p>&lt;text&gt;Blockquote&lt;/text&gt;</p>\n</blockquote>", false, true)]
+            [InlineData("> <text>Blockquote</text>", "<blockquote>\r\n<p>&lt;text&gt;Blockquote&lt;/text&gt;</p>\r\n</blockquote>", false, false)]
```
zhhyu, line 50: I am not very sure but it may be helpful to have a test case of "Nested Blockquotes".

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
zhhyu: Personally I feel that it's more readable to use "true", because the setup `Setup(x => x.IsMarkdigMdRenderingEnabled()) ` itself sounds like a question. If you prefer the current style, please follow the naming convention and have an underline at the beginning.

src/NuGetGallery/Services/MarkdownService.cs
zhhyu: We can easily make this service as a singleton, and build the pipeline when we register the DI module. I don't see any shared states here (feature flag service is a singleton), but we need to double check that it's safe to share the pipeline.

src/NuGetGallery/Services/MarkdownService.cs
lyndaidaii: ? what eles attribute to check here? I believe features is only property we need check if is NULL?

src/NuGetGallery/Services/MarkdownService.cs
lyndaidaii: No reason, I followed the way we used to implemented with CommonMark.  There is no impact if is int or byte

src/NuGetGallery/Services/MarkdownService.cs
lyndaidaii: The default is 1.
[InlineData("# Heading", "<h2>Heading</h2>", false, true)]
[InlineData("# Heading", "<h2>Heading</h2>", false, false)]

src/NuGetGallery/Services/MarkdownService.cs
lyndaidaii: Yes, It's also used to link. I could make same comments for links if you think it's not clear

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
lyndaidaii: I believe autolinks also use "nofollow", probably we need have our own markdig extension, require more work, we could remove autolink now, make as an enhance work in the furture.

src/NuGetGallery/Services/MarkdownService.cs
lyndaidaii: You mean make markdownService as singleton?

src/NuGetGallery/Services/MarkdownService.cs
zhhyu: We have checked "_features" in the constructor, so there is no need to check it again. But checking whether "markdownString" is null or not is needed. It's better to throw asap rather than wait until internal calls throw null exceptions.

src/NuGetGallery/Services/MarkdownService.cs
zhhyu: If there is no need to use "byte" here, let's just drop it.

src/NuGetGallery/Services/MarkdownService.cs
zhhyu: We need to have some tests given "incrementHeadersBy" with "+1", "-1", "0", and "heading.Level + incrementHeadersBy" > 6. What if people feed "incrementHeadersBy" which triggers "heading.Level + incrementHeadersBy" <= 0?

src/NuGetGallery/Services/MarkdownService.cs
zhhyu: I think this comment only applies to links (see "CommonMark" one). For images, we don't "allow http _or_ https", and we don't transform links to https only for _known domains_. We always force images to use "https", is it ("rewriteAllHttp: true")?

src/NuGetGallery/Services/MarkdownService.cs
zhhyu: Yeah, that's right. If there are no performance concerns, we can keep the current one. Introducing singleton needs more considerations.

lyndaidaii: I think I need do more investigation about markdig pipeline if this is thread safe to determine if we want to introduce singleton here

lyndaidaii: https://github.com/lunet-io/markdig/blob/8acf5a548edc200db81f805a78dec00552aa2051/src/Markdig/MarkdownPipelineBuilder.cs#L16 
Looks like markdownPipelinebuilder is not thread safe, probably not good idea to make singleton

src/NuGetGallery/Services/MarkdownService.cs
lyndaidaii: Thanks for advice of null check here for `markdownString `, I think it might be good to return if  `markdownString` is null.  It's not necassary to throw a null exception here. it's posible the markdownString is null based on the readme sourceType. for example, customer put url as readme source, but there is nothing is readme. I don't think we want to throw an exception in this case. 
What do you think?

src/NuGetGallery/Services/MarkdownService.cs
cristinamanum: I agree with @zhhyu. Is incrementHeadersBy allowed to be negative?  If not could argument validation be added?

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -3,25 +3,70 @@
...
+        }
 
         public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString)
         {
-            return GetHtmlFromMarkdown(markdownString, 1);
+            if (_features == null)
+            {
+                throw new ArgumentNullException(nameof(_features));
+
+            }
```
skofman1, line 39: you don't need to check if _features is null here. You already check it in the ctor

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -3,25 +3,70 @@
...
         }
 
         public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int incrementHeadersBy)
+        {
+            if (_features == null)
+            {
+                throw new ArgumentNullException(nameof(_features));
+
+            }
+
```
skofman1, line 58: same.

src/NuGetGallery/Services/MarkdownService.cs
zhhyu: Good finding! We should avoid sharing the pipeline, so keeping the current implementation is correct.

src/NuGetGallery/Services/MarkdownService.cs
zhhyu: Personally, I prefer to keep throwing the exception. One reason is that we should align with the previous behavior: for "CommonMark", if we feed a null string, it will throw the exception because of `markdownString.StartsWith("\ufeff")`. So a better choice here is to follow this and avoid breaking the current caller's logic. Another reason is that this service will be widely shared among multiple components: licenses, change logs, and etc. We should determine a unified behavior of this compact and reliable service for the future usage, rather than just considering the readme design. It's the caller's responsibility to determine how to handle the exception or validate the input before feeding into this service, and I think readme services properly handle different cases before feeding into this service.

lyndaidaii: Added validation for parameter, please check latest changes

src/NuGetGallery/Services/MarkdownService.cs
lyndaidaii: Added argument validation: 
https://github.com/NuGet/NuGetGallery/blob/bc7da900d78455b970ccf92f2d09537025c8cbf3/src/NuGetGallery/Services/MarkdownService.cs#L58

Test: 
https://github.com/NuGet/NuGetGallery/blob/bc7da900d78455b970ccf92f2d09537025c8cbf3/tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs#L62

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
lyndaidaii: Nested Blockquotes added with latest changes

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
```diff
@@ -1,54 +1,190 @@
...
 
             public GetReadMeHtmlMethod()
             {
-                _markdownService = new MarkdownService();
+                _featureFlagService = new Mock<IFeatureFlagService>();
+                _markdownService = new MarkdownService(_featureFlagService.Object);
+            }
+
+            [Fact]
+            public void ThrowsArgumentOutOfRangeExceptionForNegativeParameterValue()
```
zhhyu, line 26: nit: how about naming it like "ThrowsArgumentOutOfRangeExceptionForNegativeIncrementHeadersBy()"?

cristinamanum (approved):

src/NuGetGallery/Services/MarkdownService.cs
```diff
@@ -3,25 +3,74 @@
...
         public RenderedMarkdownResult GetHtmlFromMarkdown(string markdownString, int incrementHeadersBy)
+        {
+            if (markdownString == null)
+            {
+                throw new ArgumentNullException(nameof(markdownString));
+            }
+
+            if (incrementHeadersBy < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(incrementHeadersBy));
```
zhhyu, line 60: nit: maybe add a message about the accepted range, for example:
https://github.com/NuGet/NuGetGallery/blob/e53e62343775db2bb65bc8b14d6488d870a576c5/src/NuGetGallery/Helpers/StreamHelper.cs#L28-L31
and you can also verify this message in the test.

zhhyu (approved):

src/NuGetGallery/Services/MarkdownService.cs
lyndaidaii: Added changes with more details message for exception 
https://github.com/NuGet/NuGetGallery/blob/a0ef14ff4946f4f9b953f2e1a768563984ffc4ff/src/NuGetGallery/Services/MarkdownService.cs#L60

test: 
https://github.com/NuGet/NuGetGallery/blob/a0ef14ff4946f4f9b953f2e1a768563984ffc4ff/tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs#L27
```
            [Theory]
            [InlineData(-1)]
            public void ThrowsArgumentOutOfRangeExceptionForNegativeIncrementHeadersBy(int incrementHeadersBy)
            {
                var exception = Assert.Throws<ArgumentOutOfRangeException>(() => _markdownService.GetHtmlFromMarkdown("markdown file test", incrementHeadersBy));
                Assert.Equal(nameof(incrementHeadersBy), exception.ParamName);
                Assert.Contains("must be greater than or equal to 0", exception.Message);
            }
```

Issue #8185, opened 2020-09-03
[A11y][Contrast]Color contrast for the focus on icons Facebook and Twitter with background is 1.658 which is less than 3:1


ryuyu:
Copied from https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1184494
## Browser Details:
Firefox Version: 79.0 (64 bit)

## OS Details

Microsoft Windows 10 Enterprise  ​

Version 2004 (OS Build 19041.450)

## Repro Steps:

    Hit the URL and login with appropriate credentials
    Tab till "Packages tab" item and press enter
    Tab till any package and press enter.
    Tab till "Share" section
    Check whether the color contrast for the background to focus on the icons "Facebook" and "Twitter" greater than or equal to 3:1

## Actual Results:

Color contrast for the focus on icons Facebook and Twitter with background is 1.658 which is less than 3:1

   
## Expected Results:
Color contrast for the focus on icons Facebook and Twitter with background is 1.658 it should be greater than or equal to 3:1
## User Impact:
Users with low vision would face difficulty to identify the elements if the non-text contrast is less than 3:1

## Notes
We don't control these icons. I don't know if we are allowed to/should change them.

PR #8295, opened 2020-10-26
[A11y] Improve color contrast of focus outline on Firefox

loic-sharma:
Improve the color contrast of Firefox's outline for focused items. This effectively reverts https://github.com/NuGet/NuGetGallery/commit/c245f2caeedc052be830a791879f2d4c8e0cc943 as Firefox now properly outlines elements.

Chromium browsers (Chrome, Edgium, Opera) are all unaffected by this change. I also verified that Internet Explorer's focus outlines pass the 3:1 color contrast requirement.

# Screenshots

## Firefox (new)

image

Note that radio buttons and checkboxes are unaffected by this change.

## Firefox (old)

image

The color contrast for links on our white background is not 3:1

Addresses https://github.com/NuGet/NuGetGallery/issues/8185

### DIFF
--- a/src/Bootstrap/less/mixins/tab-focus.less
+++ b/src/Bootstrap/less/mixins/tab-focus.less
@@ -7,5 +7,6 @@
   // The next line came from: https://stackoverflow.com/a/38571103
   outline: 2px auto Highlight;
   outline: 5px auto -webkit-focus-ring-color;
+  outline: revert;
   outline-offset: -2px;
 }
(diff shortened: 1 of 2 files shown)


joelverhagen (approved):
Woah, nice digging into the history.

src/Bootstrap/less/mixins/tab-focus.less
```diff
@@ -4,8 +4,6 @@
   // WebKit-specific. Other browsers will keep their default outline style.
   // (Initially tried to also force default via `outline: initial`,
   // but that seems to erroneously remove the outline in Firefox altogether.)
-  // The next line came from: https://stackoverflow.com/a/38571103
```
joelverhagen, line 7: You can remove more of the comment, right?

loic-sharma: What's left is the "vanilla" Bootstrap content from this: https://github.com/twbs/bootstrap/blob/f17f882df292b29323f1e1da515bd16f326cee4a/less/mixins/tab-focus.less#L1

Please let me know if I missed something though!

Issue #8282, opened 2020-10-19
[A11y][WCAG2.1]Two scroll bars appears on the screen after changing the display orientation to portrait


ryuyu:
This is the same issue in both of these:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1221931
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1067651
​
Environment Details:
Application Name: NuGet Service
Application URL:https://www.nuget.org/
Windows Version - Windows10 
Edge Version-Microsoft Edge 44.18362.449.0
Chrome Version- Version 79.0.3945.130 (Official Build) (64-bit)

Pre-requisite:
Set the display orientation setting to "Portrait"
Start>Setting>System>Display Orientation.

Steps to Reproduce :
1.Launch application using provided URL.(https://www.nuget.org/)
2. Navigate to "Sign in" control and hit enter.
3. Navigate to "Sign in with Microsoft" control and hit enter.
4. Select “API Keys” from the Profile Dropdown and hit enter.
5. "API Keys" screen gets opened.
6. Observe that Two scroll bars appears on the screen after changing the display orientation to the portrait.

Actual:
Two scroll bars appear on the screen after changing the display orientation to the portrait.
Similar issue is also observed with the following screens:
API Keys-Create
Create-Select Packages
API Keys-Manage
Manage-Edit 
Upload Package
Manage Organizations
Manage Organizations-Add New
Manage Organizations -Organizations
Packages returned for json
.Json-InfoSection
.Json-OwnersSection
.Json-PackageReference
.Json-Paket CLI
.Json-Dependencies
.Json-GitHub Usage
.Json-Version History 
Expected:
Two scroll bars should not appear on the screen after changing the display orientation to the portrait.
 
User Impact: 
Users will find it difficult to view the data in a single layout .

Notes:
This is in reference to a horizontal scrollbar appearing:
![image](https://user-images.githubusercontent.com/11051729/96502880-4a90b280-1207-11eb-86b9-8faa167b3708.png)

PR #8300, opened 2020-10-27
[A11y][WCAG2.1]Shrink narrow width by a small amount

ryuyu:
Addresses https://github.com/NuGet/NuGetGallery/issues/8282
Narrowing the width at minimum width by a small amount keeps the horizontal scrollbar from appearing.

### DIFF
--- a/src/Bootstrap/less/theme/base.less
+++ b/src/Bootstrap/less/theme/base.less
@@ -248,7 +248,7 @@ body {
   .alert {
     margin: 0;
     padding: 10px;
-  } 
+  }
 }
 
 .alert-transient-jumbotron {
@@ -260,7 +260,7 @@ body {
 
   li {
     position: relative;
-    
+
     i.ms-Icon {
       position: absolute;
       left: -24px;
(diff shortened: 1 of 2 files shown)


loic-sharma (approved):
Code changes look good. Please add some info on how you plan to verify your changes haven't broken anything.

src/Bootstrap/less/theme/base.less
```diff
@@ -285,11 +285,12 @@ h2, h3 {
...
     // Bootstrap doesn't set the container width when the screen is xs.
     // This works fine when there are no elements in the container that are wider than the viewport,
-    // however, if there are elements in the container that are wider than the viewport, 
+    // however, if there are elements in the container that are wider than the viewport,
     // their overflow properties will not work properly because they are not bound by the size of the container.
     // To fix this, we need to set the container width to be the width of the viewport minus the row margin on xs screens.
     // This way, those elements that are larger than the viewport are properly bound by the size of the container and overflow properly.
-    width: calc(100vw - 15px);
+    // 2020.10.27: We actually need to set it slightly smaller than the row margin as a horizontal scrollbar appears at xs when exactly matched.
+    width: calc(100vw - 20px);
```
agr, line 293: Should we just use `width: 100%` here without any math?

ryuyu: I tried this and it doesn't seem to work. Not 100% sure why it doesn't work, but it doesn't seem too.

agr: The horizontal bar seems to appear due to `100vw` to include the width of the vertical scrollbar, so when vertical scrollbar appears, it eats few pixels but does not cause `100vw` to change, causing horizontal scrollbar to appear. I'm not sure if this behavior is according to spec.

agr (approved):

Issue #8302, opened 2020-10-29
Can't delete user certificate


agr:
### Describe the bug
Individual account management page have trouble when trying to delete added certificates: clicking the delete button does nothing.

### To Reproduce
Add a certificate to an individual account. Try to delete it. See that nothing happens when you click the "delete" button (actually an ajax request is sent that fails).

### Expected behavior
Delete button should delete certificate.

PR #8303, opened 2020-10-29
Fix for being unable to delete user account certificates

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/8302.

The page was missing anti-forgery token for ajax requests.

### DIFF
--- a/src/NuGetGallery/Views/Users/Account.cshtml
+++ b/src/NuGetGallery/Views/Users/Account.cshtml
@@ -4,6 +4,7 @@
     ViewBag.MdPageColumns = GalleryConstants.ColumnsFormMd;
     TempData["Parent"] = this;
 }
+@ViewHelpers.AjaxAntiForgeryToken(Html)
 
 <section role="main" class="container main-container page-account-settings">
     <div class="row">


joelverhagen (approved):
When did this regress? Are we sure the fix associated with the regression changeset is still "fixed"?

zhhyu (approved):

zhhyu:
> When did this regress? Are we sure the fix associated with the regression changeset is still "fixed"?

It seems that it's related to this: https://github.com/NuGet/NuGetGallery/commit/fae91fe66c8d8d294a92dc6cc43c7dd46bfe796b
We should understand why we introduce it.

lyndaidaii:
> > When did this regress? Are we sure the fix associated with the regression changeset is still "fixed"?
> 
> It seems that it's related to this: #8210
> We should understand why we introduced it at that time.

For #8210,  Fastpass complained about duplicated anti-forgery token,  one of anti-forgery token was called here at same page, that's why we remove one https://github.com/NuGet/NuGetGallery/blob/fae91fe66c8d8d294a92dc6cc43c7dd46bfe796b/src/NuGetGallery/Views/Shared/_AccountCertificates.cshtml#L52

lyndaidaii (approved):

agr:
That partial view was used from 2 views: Accounts.cshtml and ManageOrganization.cshtml. The latter had a call to `@ViewHelpers.AjaxAntiForgeryToken(Html)`, the former didn't. Having that call in ManageOrganization and in the partial caused that duplicate id issue. This PR finalizes the id fix by adding it back to the view which didn't have it on its own.

zhhyu (approved):

loic-sharma (approved):

PR #8310, opened 2020-11-05
[A11y] break word for a certificate thumbprint

zhhyu:
I decide to move to a quicker solution, and just break word for a certificate thumbprint:
image

Before:
image

@ryuyu

### DIFF
--- a/src/NuGetGallery/Views/Shared/_AccountCertificates.cshtml
+++ b/src/NuGetGallery/Views/Shared/_AccountCertificates.cshtml
@@ -90,7 +90,7 @@
                         </thead>
                         <tbody data-bind="foreach: $data.certificates">
                             <tr class="manage-certificate-listing" role="listitem">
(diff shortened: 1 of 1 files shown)


skofman1:
What will the new UX look like when the subject, expiration, issuer have values?

loic-sharma:
> What will the new UX look like when the subject, expiration, issuer have values?

FYI @zhhyu, this information is filled out only after you validate a package that was signed with your certificate. This is tricky to do on a local deployment, I'd recommend deploying your changes to DEV and looking at the Microsoft organization's certificates: https://dev.nugettest.org/organization/Microsoft

skofman1:
Faking data in local DB also works :)

zhhyu:
> What will the new UX look like when the subject, expiration, issuer have values?

image

It's just another line by default. The thumbprint is a long string, which we have to break the word, and it only applies to the thumbprint.

skofman1 (approved):

PR #8311, opened 2020-11-05
[UI] No need to use form-group for Account Change Email

zhhyu:
It's a better UI when reflowing to align Account Change Email with others, by dropping the form-group:
image

Before:
image

### DIFF
--- a/src/NuGetGallery/Views/Shared/_AccountChangeEmail.cshtml
+++ b/src/NuGetGallery/Views/Shared/_AccountChangeEmail.cshtml
@@ -15,7 +15,7 @@
                 {
                     @Html.AntiForgeryToken()
 
-                    <div class="row form-group">
+                    <div class="row">
                         <div class="col-sm-6">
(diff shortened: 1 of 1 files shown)


drewgillies (approved):

shishirx34:
I might be out of the loop, but, was this a UX team feedback to make this change? Current UX shows the buttons side by side, like below.

image

when did we switch to "row" based buttons? was there a UX review done on this?

shishirx34:
oh is that the overflow view of buttons when resizing the window? if yes, ignore my comments.

zhhyu:
> oh is that the overflow view of buttons when resizing the window? if yes, ignore my comments.

Yes, it's used when reflowing.

zhhyu:
> > oh is that the overflow view of buttons when resizing the window? if yes, ignore my comments.
> 
> Yes, it's used when reflowing. The only thing changed is that we remove the empty line between two buttons.

PR #8315, opened 2020-11-09
[SAS Token] Fallback when permissions cannot be use on container creation.

dannyjdev:
In some places we create a container and right after we set the permissions access, since we cannot modify permissions after a container has been created when using a sas token we need to modify this behavior to add the permissions at creation time.

**AuthorizationFailure** is the error we get when trying to perform an update of the permissions of an already created container.
See error on: https://docs.microsoft.com/en-us/rest/api/storageservices/sas-error-codes

Address: https://github.com/NuGet/Engineering/issues/3484


src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
```diff
@@ -69,9 +81,18 @@ public async Task<bool> DeleteIfExistsAsync()
...
 
-        public async Task CreateAsync()
+        public async Task CreateAsync(BlobContainerPermissions permissions)
         {
-            await _blobContainer.CreateAsync();
+            var publicAccess = permissions?.PublicAccess;
+
+            if (publicAccess.HasValue)
+            {
+                await _blobContainer.CreateAsync(publicAccess.Value, null, null);
```
joelverhagen, line 90: name params that are null, to improve readability.

src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
```diff
@@ -40,10 +40,22 @@ public CloudBlobContainerWrapper(CloudBlobContainer blobContainer)
...
+            var publicAccess = permissions?.PublicAccess;
+
+            ICancellableAsyncResult beginCreateIfNotExistsResult;
+            if (publicAccess.HasValue)
+            {
+                beginCreateIfNotExistsResult = _blobContainer.BeginCreateIfNotExists(publicAccess.Value, null, null, null, null);
+            }
+            else
+            {
+                beginCreateIfNotExistsResult = _blobContainer.BeginCreateIfNotExists(null, null);
```
joelverhagen, line 54: Could we just use `CreateIfNotExistsAsync`. The `Begin*` stuff is a very old async model and should be avoided.

joelverhagen (approved):

src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
```diff
@@ -40,10 +40,22 @@ public CloudBlobContainerWrapper(CloudBlobContainer blobContainer)
             return new BlobResultSegmentWrapper(segment);
         }
 
-        public Task CreateIfNotExistAsync()
+        public Task CreateIfNotExistAsync(BlobContainerPermissions permissions)
         {
+            var publicAccess = permissions?.PublicAccess;
+
+            ICancellableAsyncResult beginCreateIfNotExistsResult;
+            if (publicAccess.HasValue)
```
shishirx34, line 48: can passed in `permissions` be `null`? I am assuming it could be given the check above.

perhaps consider adding a null check for `publicAccess` as well?

```suggestion
            if (publicAccess != null && publicAccess.HasValue)
```

src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
```diff
@@ -40,10 +40,22 @@ public CloudBlobContainerWrapper(CloudBlobContainer blobContainer)
...
+            {
+                beginCreateIfNotExistsResult = _blobContainer.BeginCreateIfNotExists(publicAccess.Value, null, null, null, null);
```
shishirx34, line 50: nit: null param names.

shishirx34 (approved):

src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
dannyjdev: Updated.

src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
dannyjdev: Updated.

src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
dannyjdev: Updated.

PR #8320, opened 2020-11-11
Add netstandard2.1 support to NuGetGallery.Core and NuGet.Services.Entities

joelverhagen:
This is required because these two libraries are transitive dependencies of NuGet.Services.AzureSearch. It would be technically possible to run the Azure Search Service without this change, but it would require a significantly more complex changeset in NuGet.Jobs to split all of the projects and files depended on by the search service into two halves: "required by Azure Search Service" and "not required by Azure Search Service". For example, there is a shared DI wire-up step for all search components which would need to be cleverly split. See [DependencyInjectionExtensions.cs](https://github.com/NuGet/NuGet.Jobs/blob/bbd7a3c6ebf87e1f4c7b2e6e34e3607c5ce242da/src/NuGet.Services.AzureSearch/DependencyInjectionExtensions.cs#L25).

The simpler approach is to make all project dependencies of the search service compatible with a `netcoreapp3.1` (and eventually `net5.0`) entry point. In this cases, it's a `netstandard2.1` cross-target. We need `netstandard2.1` instead of the ServerCommon `netstandard2.0` since Entity Framework supports `netstandard2.1`. Since ServerCommon already has the required .NET Core support, these are the only projects that remain outside of NuGet.Jobs to have .NET Core support.

Changes:
- Move NuGetGallery.Core and NuGet.Services.Entities to SDK-based projects
- Introduce a shared props file for these two projects. We can't use Build.Directory.props since not all projects are converted.
- Add a `netstandard2.1` target to these two projects.
- Update ServerCommon version
- Conditionally compile a select few lines of code that are not required for the Azure Search Service.
- Note that this brings Markdig up to version 0.22.0 (/cc @lyndaidaii)

Progress on https://github.com/NuGet/Engineering/issues/3487.


src/NuGetGallery.Core/EmbeddedLicenseFileType.cs
```diff
@@ -1,26 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-namespace NuGetGallery
-{
-    /// <summary>
-    /// Specifies the type of the license file used in the package
-    /// </summary>
-    public enum EmbeddedLicenseFileType
```
skofman1, line 9: >EmbeddedLicenseFileType [](start = 16, length = 23)

Don't we need those files?

joelverhagen: They are duplicates. The reason this was found was because the new project style automatically includes all *.cs files in the directory. Previously, they were in the directory but not included in the .csproj, essentially hidden from us in VS:
image

src/NuGetGallery.Services/NuGetGallery.Services.csproj
```diff
@@ -305,10 +305,10 @@
       <Version>5.8.0-preview.3.6823</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Configuration">
-      <Version>2.74.0</Version>
+      <Version>2.79.0-jver-netstandard-4208332</Version>
```
loic-sharma, line 308: Should these two be `2.79.0` as well?

joelverhagen: Yup, good catch.

src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs
```diff
@@ -42,16 +42,12 @@ public CloudBlobContainerWrapper(CloudBlobContainer blobContainer)
 
         public Task CreateIfNotExistAsync()
         {
-            return Task.Factory.FromAsync<bool>(
-                _blobContainer.BeginCreateIfNotExists(null, null), 
-                _blobContainer.EndCreateIfNotExists);
+            return _blobContainer.CreateIfNotExistsAsync();
```
loic-sharma, line 45: Nit: consider adding `await`s here. I don't feel strongly though

Common.props
```diff
@@ -0,0 +1,40 @@
+<Project>
+  <PropertyGroup>
```
loic-sharma, line 2: Could you add a quick comment that these properties are for SDK style projects only. For properties shared by both SDK0style projects and non-SDK style projects, use the `Directory.Build.props` file instead

joelverhagen: Sure, sounds good.

loic-sharma (approved):

skofman1 (approved):
:shipit:

Issue #8313, opened 2020-11-06
[A11y] [WCAG2.1] Two scroll bars shown on status page, when zooming in.


zhhyu:
See this: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1221931/

agr:
Work with @chgill-MSFT and UI folks to determine expexted behavior for narrow screen layouts.

lyndaidaii:
We will continue to reach out to UX experts to find out if there is a better fix

PR #8330, opened 2020-11-23
Fix two scroll bar at status page 

lyndaidaii:

Required to be only one scroll. 

- set up min-width for table
- word break

After changes: 
NuGet _ Status

Addresses https://github.com/NuGet/NuGetGallery/issues/8313


skofman1 (approved):

lyndaidaii:
Merge current appoach for now so that Vendor could verify the changes in dev. We will reach out to UI expert see if there is a better solution

Issue #7650, opened 2019-10-25
Gallery authors and package consumers should see a message on the Display Package page if the package is vulnerable


scottbommarito:
Spec: https://github.com/NuGet/Home/blob/dev/designs/PackageVulnerability/NuGet.orgVulnerabilitiesPhase1.md

PR #8343, opened 2020-12-04
Surface vulnerabilities in package detail page

drewgillies:
First steps in addressing https://github.com/NuGet/NuGetGallery/issues/7650
In this PR we have:
- the feature flag
- initial changes to the view model
- initial changes to the view--we'll iterate on this in consultation with PM (for example, we don't yet have the distinction between a customer view and an owner view of the vulnerability expander, we will need to discuss tooltip and other language used, etc)--but we're under a feature flag in order to do this iteratively
- test coverage will need to be augmented (refactored, actually) and I want to do this in a different PR, as some class renames will affect vulnerability ingestion code--it will be noisy, so I'll separate it out for reviewing

Here is the current state of the view, highlights in red boxes:
image

Other notes:
- Because deprecations and vulnerabilities share the same bang icon, you'll see a combining of the logic in the view, and combination messages in the tooltip. These can be sweetened later as well, of course.
- I needed to separate out styles for deprecation and vulnerability for two reasons--we may wish to change the look of one or the other in future, and JS code that affects the expanders was having its operation polluted by reusing existing styles for the new expander.
- Colorizing severities can come in a future CSS tweak
- EF needed some attention as the join between Packages and VulnerablePackageVersionRanges wasn't functional (specifically the navigation properties in the m:m relationship)

### DIFF
--- a/src/NuGet.Services.Entities/Package.cs
+++ b/src/NuGet.Services.Entities/Package.cs
@@ -24,7 +24,7 @@ namespace NuGet.Services.Entities
             SymbolPackages = new HashSet<SymbolPackage>();
             Deprecations = new HashSet<PackageDeprecation>();
             AlternativeOf = new HashSet<PackageDeprecation>();
-            Vulnerabilities = new HashSet<VulnerablePackageVersionRange>();
+            VulnerableVersionRanges = new HashSet<VulnerablePackageVersionRange>();
             Listed = true;
         }
 #pragma warning restore 618
@@ -279,7 +279,7 @@ namespace NuGet.Services.Entities
         /// <summary>
         /// Gets or sets the list of vulnerabilites that this package has.
         /// </summary>
-        public ICollection<VulnerablePackageVersionRange> Vulnerabilities { get; set; }
+        public ICollection<VulnerablePackageVersionRange> VulnerableVersionRanges { get; set; }
 
         /// <summary>
         /// A flag that indicates that the package metadata had an embedded icon specified.
--- a/src/NuGetGallery.Core/Entities/EntitiesContext.cs
+++ b/src/NuGetGallery.Core/Entities/EntitiesContext.cs
@@ -482,7 +482,7 @@ namespace NuGetGallery
             modelBuilder.Entity<VulnerablePackageVersionRange>()
                 .HasKey(pv => pv.Key)
                 .HasMany(pv => pv.Packages)
-                .WithMany(p => p.Vulnerabilities);
+                .WithMany(p => p.VulnerableVersionRanges);
 
             modelBuilder.Entity<VulnerablePackageVersionRange>()
                 .HasIndex(pv => pv.PackageId);
@@ -490,6 +490,13 @@ namespace NuGetGallery
             modelBuilder.Entity<VulnerablePackageVersionRange>()
                 .HasIndex(pv => new { pv.VulnerabilityKey, pv.PackageId, pv.PackageVersionRange })
                 .IsUnique();
+
+            modelBuilder.Entity<VulnerablePackageVersionRange>()
+                .HasMany<Package>(vpvr => vpvr.Packages)
+                .WithMany(p => p.VulnerableVersionRanges)
+                .Map(c => c.ToTable("VulnerablePackageVersionRangePackages")
+                    .MapLeftKey("VulnerablePackageVersionRange_Key")
+                    .MapRightKey("Package_Key"));
 
             modelBuilder.Entity<PackageRename>()
                 .HasKey(r => r.Key)
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -5,6 +5,26 @@ $(function () {
     window.nuget.configureExpander("rename-content-container", "ChevronDown", null, "ChevronUp");
     configureExpanderWithEnterKeydown($('#show-rename-content-container'));
 
+    // Configure the vulnerability information container
+    var container = $('#show-vulnerabilities-content-container');
+    if ($('#vulnerabilities-content-container').children().length) {
+        // If the deprecation information container has content, configure it as an expander.
+        window.nuget.configureExpander("vulnerabilities-content-container", "ChevronDown", null, "ChevronUp");
+        configureExpanderWithEnterKeydown(container)
+    }
+    else {
+        // If the container does not have content, remove its expander attributes
+        var expanderAttributes = ['data-toggle', 'data-target', 'aria-expanded', 'aria-controls', 'tabindex'];
+        for (var i in expanderAttributes) {
+            container.removeAttr(expanderAttributes[i]);
+        }
+
+        // The expander should not be clickable when it doesn't have content
+        container.find('.vulnerabilities-expander').removeAttr('role');
+
+        $('#vulnerabilities-expander-icon-right').hide();
+    }
+
     // Configure the deprecation information container
     var container = $('#show-deprecation-content-container');
     if ($('#deprecation-content-container').children().length) {
--- a/src/NuGetGallery/Services/PackageVulnerabilitiesService.cs
+++ b/src/NuGetGallery/Services/PackageVulnerabilitiesService.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Data.Entity;
+using System.Linq;
+using System.Threading.Tasks;
+using NuGet.Services.Entities;
+using NuGetGallery.Auditing;
+
+namespace NuGetGallery
+{
+    public class PackageVulnerabilitiesService : IPackageVulnerabilitiesService
+    {
+        private readonly IEntitiesContext _entitiesContext;
+
+        public PackageVulnerabilitiesService(IEntitiesContext entitiesContext)
+        {
+            _entitiesContext = entitiesContext ?? throw new ArgumentNullException(nameof(entitiesContext));
+        }
+
+        public IReadOnlyDictionary<int, IReadOnlyList<PackageVulnerability>> GetVulnerabilitiesById(string id)
+        {
+            var result = new Dictionary<int, List<PackageVulnerability>>();
+            foreach (var package in _entitiesContext.Packages.Where(p => p.PackageRegistration != null && p.PackageRegistration.Id == id)
+                .Include(p => p.VulnerableVersionRanges).Include("VulnerableVersionRanges.Vulnerability"))
+            {
+                if (package.VulnerableVersionRanges == null)
+                {
+                    continue;
+                }
+
+                var packageVulnerabilities = (List<PackageVulnerability>)null;
+                foreach (var vulnerabilityVersionRange in package.VulnerableVersionRanges)
+                {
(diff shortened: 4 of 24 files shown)


drewgillies:
FYI @JonDouglas

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -684,9 +689,19 @@
                             {
                                 <th aria-hidden="true" abbr="Signature Information"></th>
                             }
-                            @if (Model.IsPackageDeprecationEnabled)
+                            @if (Model.IsPackageDeprecationEnabled || Model.IsPackageVulnerabilitiesEnabled)
```
skofman1, line 692: >IsPackageDeprecationEnabled [](start = 39, length = 27)

we should get rid of the deprecation feature flag, but in the meantime, for completeness, please handle the case that deprecation is disabled and only vulnerability is enabled

src/NuGet.Services.Entities/Package.cs
```diff
@@ -279,7 +279,7 @@ public bool HasReadMe
         /// <summary>
         /// Gets or sets the list of vulnerabilites that this package has.
         /// </summary>
-        public ICollection<VulnerablePackageVersionRange> Vulnerabilities { get; set; }
+        public ICollection<VulnerablePackageVersionRange> VulnerableVersionRanges { get; set; }
```
skofman1, line 282: >VulnerableVersionRanges [](start = 58, length = 23)

can u explain why this is needed?

skofman1:
@drewgillies , this PR is pretty big. Please split.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
```diff
@@ -491,6 +491,13 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
                 .HasIndex(pv => new { pv.VulnerabilityKey, pv.PackageId, pv.PackageVersionRange })
                 .IsUnique();
 
+            modelBuilder.Entity<VulnerablePackageVersionRange>()
```
cristinamanum, line 494: What scenario we try to enable with this change?

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -5,6 +5,26 @@ $(function () {
...
+    var container = $('#show-vulnerabilities-content-container');
+    if ($('#vulnerabilities-content-container').children().length) {
+        // If the deprecation information container has content, configure it as an expander.
+        window.nuget.configureExpander("vulnerabilities-content-container", "ChevronDown", null, "ChevronUp");
+        configureExpanderWithEnterKeydown(container)
+    }
+    else {
+        // If the container does not have content, remove its expander attributes
+        var expanderAttributes = ['data-toggle', 'data-target', 'aria-expanded', 'aria-controls', 'tabindex'];
+        for (var i in expanderAttributes) {
```
agr, line 18: `expanderAttributes.forEach(attribute => container.removeAttr(attribute))`?

src/NuGetGallery/Services/PackageVulnerabilitiesService.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public PackageVulnerabilitiesService(IEntitiesContext entitiesContext)
+        {
+            _entitiesContext = entitiesContext ?? throw new ArgumentNullException(nameof(entitiesContext));
+        }
+
+        public IReadOnlyDictionary<int, IReadOnlyList<PackageVulnerability>> GetVulnerabilitiesById(string id)
+        {
+            var result = new Dictionary<int, List<PackageVulnerability>>();
+            foreach (var package in _entitiesContext.Packages.Where(p => p.PackageRegistration != null && p.PackageRegistration.Id == id)
```
agr, line 26: For readability, please move everything after `in` into its own variable.

src/NuGetGallery/Services/PackageVulnerabilitiesService.cs
```diff
@@ -0,0 +1,51 @@
...
+                .Include(p => p.VulnerableVersionRanges).Include("VulnerableVersionRanges.Vulnerability"))
```
agr, line 27: The last `Include("VulnerableVersionRanges.Vulnerability")` should cause `VulnerableVersionRanges` to get included, too, so preceding `Include(p => p.VulnerableVersionRanges)` is not needed.

agr, line 27: Also, have you considered constructing the `"VulnerableVersionRanges.Vulnerability"` string from couple of `nameof()` calls? That would make sure the code won't break if we rename/remove any of those properties.

src/NuGetGallery/Services/PackageVulnerabilitiesService.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                if (package.VulnerableVersionRanges == null)
+                {
+                    continue;
+                }
+
+                var packageVulnerabilities = (List<PackageVulnerability>)null;
+                foreach (var vulnerabilityVersionRange in package.VulnerableVersionRanges)
```
agr, line 35: Is it the same as
```
if (package.VulnerableVersionRanges.Any())
{
    result.Add(package.Key, package.VulnerableVersionRanges.ToList());
}
```
?

src/NuGetGallery/Services/PackageVulnerabilitiesService.cs
```diff
@@ -0,0 +1,51 @@
...
+                    {
+                        packageVulnerabilities = new List<PackageVulnerability>();
+                        result.Add(package.Key, packageVulnerabilities);
+                    }
+
+                    packageVulnerabilities.Add(vulnerabilityVersionRange.Vulnerability);
+                }
+            }
+
+            return result.Count == 0 ? null :
```
agr, line 47: nit: consider using `!result.Any()` instead of comparing with zero.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -684,9 +689,19 @@
...
+                                if (Model.IsPackageDeprecationEnabled && Model.IsPackageVulnerabilitiesEnabled)
+                                {
+                                    abbreviation = "Deprecation and Vulnerability Information";
+                                }
+                                else if (Model.IsPackageDeprecationEnabled)
+                                {
+                                    abbreviation = "Deprecation Information";
+                                }
+
+                                <th aria-hidden="true" abbr =@abbreviation />
```
agr, line 704: extra space after `abbr`. Shouldn't `@abbreviation` be in quotes?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -784,8 +800,31 @@
...
                                             }
 
+                                            useAlso = true;
+                                        }
+
+                                        if (Model.IsPackageVulnerabilitiesEnabled && packageVersion.Vulnerabilities != null && packageVersion.Vulnerabilities.Any())
+                                        {
+                                            vulnerabilitiesTitle = useAlso ? "Also, " : "";
+                                            var severity = Enum.GetName(typeof(PackageVulnerabilitySeverity), packageVersion.MaxVulnerabilitySeverity).ToLowerInvariant();
+                                            vulnerabilitiesTitle += packageVersion.Version + " has at least one vulnerability with " + severity + " severity.";
```
agr, line 810: Please, use `$""` instead of chain of string concatenations.

src/NuGetGallery/Views/Packages/_DisplayPackageVulnerabilities.cshtml
```diff
@@ -0,0 +1,45 @@
...
+
+                <div class="vulnerabilities-expander-container">
+                    <i id="vulnerabilities-expander-icon-right" class="vulnerabilities-expander-icon vulnerabilities-expander-info-right ms-Icon ms-Icon--ChevronDown" aria-hidden="true"></i>
+                </div>
+            </div>
+        </div>
+
+        <div class="vulnerabilities-content-container collapse" id="vulnerabilities-content-container">
+            <b>Details</b>
+            <table style="width:100%; border:none; border-collapse:unset">
```
agr, line 26: We shouldn't use inline styles. Same on line 38.

src/NuGetGallery/Views/Packages/_DisplayPackageVulnerabilities.cshtml
```diff
@@ -0,0 +1,45 @@
...
+                @{
+                    foreach (var vulnerability in Model.Vulnerabilities)
+                    {
+                        var truncatedUrl = vulnerability.AdvisoryUrl;
+                        if (truncatedUrl.Length > 50)
+                        {
+                            truncatedUrl = truncatedUrl.Substring(0, 47) + "...";
```
agr, line 33: Where `50` and `47` are coming from? Can we just link the word "Advisory" itself (on line 37) instead of using truncated URL for that?

src/NuGet.Services.Entities/Package.cs
drewgillies: It's a poorly-named navigation property - it doesn't reference a collection of Vulnerability (i.e. advisories); it references a collection of VulnerablePackageRange (i.e. the records describing the version range for a given advisory). It's helpful to give this a more descriptive (and correct) name. Come to think of it, I should update the summary comment above it.

src/NuGetGallery.Core/Entities/EntitiesContext.cs
drewgillies: Walking the package version ranges for a given package (which is a m:m) - this enables us to build a dictionary of vulnerabilities for a given package record.

src/NuGetGallery/Scripts/gallery/page-display-package.js
drewgillies: copied code, but happy to make this change.

src/NuGetGallery/Services/PackageVulnerabilitiesService.cs
drewgillies: Yes, that should work.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
drewgillies: Will do.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
drewgillies: I think you're right--seems not to matter but will do this.

src/NuGetGallery/Views/Packages/_DisplayPackageVulnerabilities.cshtml
drewgillies: Didn't realize. Thanks. Do we have these coding guidelines documented?

src/NuGetGallery/Views/Packages/_DisplayPackageVulnerabilities.cshtml
drewgillies: Arbitrary value to match the display width--we could have an Advisory link but to show the URL is also interesting. I expect us to iterate on this design as people test it.

drewgillies:
Thanks for all of your feedback, everyone. I'll close this PR and create separate ones per Svetlana's request. I've answered the questions asked here.

Issue #8344, opened 2020-12-04
Cannot convert account to organization


losolio:
### Describe the bug
Error when converting account to organization

### To Reproduce
1. Log with my account `losol`, go to this URL 'https://www.nuget.org/account/transform.', add losolio as new admin
1. Log in as losolio, go to `https://www.nuget.org/account/Organizations`
2. Perform this operation: accept joining organization by clicking the V
3. See error: Oops! Error 500 - Internal Server Error. An error occurred while processing your request. We really messed up this time...

### Expected behavior
Joining losol organization

joelverhagen:
Hey @losolio and @waltiam (from https://github.com/NuGet/NuGetGallery/issues/8349), this is a legitimate bug on nuget.org right now. Thanks for bringing this to our attention. We are working quickly to get a fix out to unblock your experience.

waltiam:
As a work around ... individual packages can be reassigned successfully.

joelverhagen:
@losolio and @waltiam, a fix has been shipped to nuget.org with a fix. You should be able to convert your account into an organization now.

Please go ahead and give it a try again and let me know if you're still blocked.

losolio:
Thanks, @joelverhagen. Confirming that this bug is fixed, and that the organization conversion succeeded. cc @waltiam

PR #8350, opened 2020-12-10
[Hotfix] Add missing resource

joelverhagen:
This PR fixes the currently broken transform-user-to-organization flow: https://github.com/NuGet/NuGetGallery/issues/8344

This flow was broken by my move to .SDK-based projects. When converting NuGetGallery.Core.csproj to SDK-based, I missed an embedded resource item.

I have verified this on DEV. I'd like to take this to INT and PROD today to unblock our users.

### DIFF
--- a/src/NuGetGallery.Core/NuGetGallery.Core.csproj
+++ b/src/NuGetGallery.Core/NuGetGallery.Core.csproj
@@ -35,6 +35,14 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Remove="Infrastructure\MigrateUserToOrganization.sql" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Include="Infrastructure\MigrateUserToOrganization.sql" />
+  </ItemGroup>
+
+  <ItemGroup>
     <PackageReference Include="NuGet.Packaging">
       <Version>5.8.0</Version>
     </PackageReference>


skofman1 (approved):

src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -34,6 +34,14 @@
     <Compile Remove="Services\CorePackageFileService.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="Infrastructure\MigrateUserToOrganization.sql" />
```
cristinamanum, line 38: Why is this needed?

joelverhagen: This is added automatically by the IDE when changing the type to embedded.

By default, *.sql (and many other file types) are included as `<None>` items in SDK-based projects.

It's not strictly necessary AFAIK (`<None>` has no effect) but I figured it's best to follow the IDE defaults to avoid future unexpected .csproj changes.

cristinamanum (approved):

src/NuGetGallery.Core/NuGetGallery.Core.csproj
cristinamanum: It sounds good

PR #8353, opened 2020-12-15
[A11y]Fixing a missed horizontal scroll bar case

ryuyu:
Fixing https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1188115 being reopened.
image

image

### DIFF
--- a/src/NuGetGallery/Views/Users/_UserAccountChangeExternalCredential.cshtml
+++ b/src/NuGetGallery/Views/Users/_UserAccountChangeExternalCredential.cshtml
@@ -30,10 +30,10 @@
                 @Html.AntiForgeryToken()
 
                 <div class="login-account-row">
-                    <span>
+                    <span class="col-sm-9"  style="padding: 0;">
                         Microsoft account: <b>@cred.Identity</b>
                     </span>
-                    <span class="col-sm-3">
+                    <span class="col-sm-3" style="padding: 0;">
                         <button class="btn btn-default btn-block" type="submit"
                                 data-confirm="This action will replace all of your linked accounts and password login with the new login account. Are you sure you want to change your login account?">
                             Change Account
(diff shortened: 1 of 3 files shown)


loic-sharma:
Do we need to update these snippets as well?

https://github.com/NuGet/NuGetGallery/blob/41db291887bb0a08ab7e7c086624df0331c3079c/src/NuGetGallery/Views/Users/_UserAccountChangeExternalCredential.cshtml#L72-L80

https://github.com/NuGet/NuGetGallery/blob/41db291887bb0a08ab7e7c086624df0331c3079c/src/NuGetGallery/Views/Users/_UserAccountChangeExternalCredential.cshtml#L84-L89

src/NuGetGallery/Views/Users/_UserAccountChangeExternalCredential.cshtml
```diff
@@ -30,10 +30,10 @@
                 @Html.AntiForgeryToken()
 
                 <div class="login-account-row">
-                    <span>
+                    <span class="col-sm-9"  style="padding: 0;">
```
loic-sharma, line 33: What's with the `padding: 0`?  Do these spans need to be wrapped by a `<div class="row"> ... </div>` or something?

ryuyu: I tried that and it behaved very strangely. Currently, overriding this for this line was the most robust solution in practice in terms of reflow. btn-block appears to make the button behave oddly for reflowing.

ryuyu:
Good catch on those snippets @loic-sharma 
Updated.

src/NuGetGallery/Views/Users/_UserAccountChangeExternalCredential.cshtml
lyndaidaii: I am aslo not sure` padding: 0` here, I tried build your changes locally, it appears display improperly
Capture

ryuyu: Good eye @lyndaidaii ! I noticed when testing on dev that I forgot to push a CSS change that needs to accompany this. It is now pushed.

lyndaidaii (approved):

PR #8356, opened 2020-12-17
Rename vulnerability management service, add feature flag

drewgillies:
Starts to address: https://github.com/nuget/nugetgallery/issues/7650

This is a rename of the existing vulnerability service to a vulnerability management service, and the vulnerability service verifier to a vulnerability verifier. Quite a few files in this PR, but it's very repetitive.

Also, I've introduced a feature flag here for use in vulnerability messaging.

This is all groundwork for the PRs to follow.

PR #8357, opened 2020-12-17
Add vulnerabilities service

drewgillies:
Starts to address: https://github.com/nuget/nugetgallery/issues/7650.

PR https://github.com/NuGet/NuGetGallery/pull/8356 must be merged before this one.

Here I introduce the new package vulnerability service. This service's job is simple at this point--return a dictionary of package keys with a list of vulnerabilities for each. Some plumbing is also added to make those values available to view models.

There is also the renaming of an EF navigation property Vulnerabilities to VulnerablePackageRanges, which is more representative.

Again, a few files here, but there's quite a bit of repetition for the rename and plumbing.

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/PackageVulnerabilitiesManagementService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageVulnerabilitiesManagementService.cs
@@ -45,7 +45,7 @@ namespace NuGetGallery
                 var versionRange = VersionRange.Parse(possibleRange.PackageVersionRange);
                 if (versionRange.Satisfies(version))
                 {
-                    package.Vulnerabilities.Add(possibleRange);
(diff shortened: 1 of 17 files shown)


src/NuGetGallery.Services/PackageManagement/PackageVulnerabilitiesManagementService.cs
```diff
@@ -278,7 +278,7 @@ private void ProcessNewVulnerabilityRange(VulnerablePackageVersionRange range, H
                         package.NormalizedVersion,
                         range.PackageVersionRange);
 
-                    package.Vulnerabilities.Add(range);
```
skofman1, line 281: >Vulnerabilities [](start = 28, length = 15)

Please make sure we deploy the job after this merge (not only gallery)

skofman1 (approved):
:shipit:

PR #8358, opened 2020-12-17
Add vulnerability messaging to package details page

drewgillies:
Completes: https://github.com/nuget/nugetgallery/issues/7650

Here we change the view to include a vulnerabilities panel for an affected package version.

Non-owner view, not expanded:
image

Owner view, expanded:
image

- bangs and tooltips in the version list (combined for deprecations and vulnerabilities). Again, formats and icons may be tweaked going forward. 

Vulnerability only:
image

Vulnerability + deprecation:
image

### DIFF
--- a/src/Bootstrap/less/theme/page-display-package.less
+++ b/src/Bootstrap/less/theme/page-display-package.less
@@ -63,6 +63,50 @@
     }
   }
 
+  .vulnerabilities-container {
+    .vulnerabilities-expander {
+      display: flex;
+      justify-content: space-between;
+      vertical-align: middle;
+      width: 100%;
+
+      .vulnerabilities-expander-container {
+        display: flex;
+
+        .vulnerabilities-expander-icon {
+          position: unset;
+          top: unset;
+        }
+
+        .vulnerabilities-expander-info-right {
+          padding-left: 15px;
+        }
+      }
+    }
+
+    .vulnerabilities-content-container {
+      margin-top: 15px;
+      padding-top: 15px;
+      border-top: 1px solid lightgray;
+
+      .vulnerabilities-list {
+        border-collapse: unset;
+      }
+    }
+
+    .vulnerabilities-severity-critical {
+      color: red
+    }
+
+    .vulnerabilities-severity-high {
+      color: red
+    }
+
+    .vulnerabilities-severity-moderate {
+      color: blue
+    }
+  }
+
   .failed-validation-alert-list {
     margin-top: 15px;
     margin-bottom: 15px;
--- a/src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
+++ b/src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
@@ -183,7 +183,8 @@ namespace NuGetGallery
                 }
             }
 
-            if (packageKeyToDeprecation != null && packageKeyToDeprecation.TryGetValue(package.Key, out var deprecation))
+            PackageDeprecation deprecation = null;
+            if (packageKeyToDeprecation != null && packageKeyToDeprecation.TryGetValue(package.Key, out deprecation))
             {
                 viewModel.DeprecationStatus = deprecation.Status;
             }
@@ -192,20 +193,74 @@ namespace NuGetGallery
                 viewModel.DeprecationStatus = PackageDeprecationStatus.NotDeprecated;
             }
 
+            PackageVulnerabilitySeverity? maxSeverity = null;
             if (packageKeyToVulnerabilities != null 
                 && packageKeyToVulnerabilities.TryGetValue(package.Key, out var vulnerabilities)
                 && vulnerabilities != null && vulnerabilities.Any())
             {
                 viewModel.Vulnerabilities = vulnerabilities;
-                viewModel.MaxVulnerabilitySeverity = viewModel.Vulnerabilities.Max(v => v.Severity);
+                maxSeverity = viewModel.Vulnerabilities.Max(v => v.Severity); // cache for messaging
+                viewModel.MaxVulnerabilitySeverity = maxSeverity.Value;
             }
             else
             {
-                viewModel.Vulnerabilities = new List<PackageVulnerability>();
-                viewModel.MaxVulnerabilitySeverity = PackageVulnerabilitySeverity.Low;
+                viewModel.Vulnerabilities = null;
+                viewModel.MaxVulnerabilitySeverity = default;
             }
 
+            viewModel.PackageWarningIconTitle =
+                GetWarningIconTitle(viewModel.Version, deprecation, maxSeverity);
+
             return viewModel;
+        }
+
+        private static string GetWarningIconTitle(
+            string version, 
+            PackageDeprecation deprecation,
+            PackageVulnerabilitySeverity? maxSeverity)
+        {
+            // We want a tooltip title for the warning icon, which concatenates deprecation and vulnerability information cleanly
+
+            var deprecationTitle = "";
+            if (deprecation != null)
+            {
+                deprecationTitle = version;
+                var isLegacy = deprecation.Status.HasFlag(PackageDeprecationStatus.Legacy);
+                var hasCriticalBugs = deprecation.Status.HasFlag(PackageDeprecationStatus.CriticalBugs);
+                if (hasCriticalBugs)
+                {
+                    if (isLegacy)
+                    {
+                        deprecationTitle += " is deprecated because it's legacy and has critical bugs";
+                    }
+                    else
+                    {
+                        deprecationTitle += " is deprecated because it has critical bugs";
+                    }
+                }
+                else if (isLegacy)
+                {
+                    deprecationTitle += " is deprecated because it's legacy and no longer maintained";
+                }
+                else
+                {
+                    deprecationTitle += " is deprecated";
+                }
+            }
+
+            if (maxSeverity.HasValue)
+            {
+                var severity = Enum.GetName(typeof(PackageVulnerabilitySeverity), maxSeverity)?.ToLowerInvariant() ?? "unknown";
+                var vulnerabilitiesTitle = $"{version} has at least one vulnerability with {severity} severity.";
+
+                return string.IsNullOrEmpty(deprecationTitle)
+                    ? vulnerabilitiesTitle
+                    : $"{deprecationTitle}; {vulnerabilitiesTitle}";
+            }
+
+            return string.IsNullOrEmpty(deprecationTitle)
+                ? string.Empty
+                : $"{deprecationTitle}.";
         }
 
         private static string GetPushedBy(Package package, User currentUser, Dictionary<User, string> pushedByCache)
--- a/src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
+++ b/src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
@@ -192,10 +192,17 @@ namespace NuGetGallery
                 viewModel.DeprecationStatus = PackageDeprecationStatus.NotDeprecated;
             }
 
-            if (packageKeyToVulnerabilities != null && packageKeyToVulnerabilities.TryGetValue(package.Key, out var vulnerabilities))
+            if (packageKeyToVulnerabilities != null 
+                && packageKeyToVulnerabilities.TryGetValue(package.Key, out var vulnerabilities)
+                && vulnerabilities != null && vulnerabilities.Any())
             {
                 viewModel.Vulnerabilities = vulnerabilities;
-                viewModel.MaxVulnerabilitySeverity = vulnerabilities.Max(v => v.Severity);
+                viewModel.MaxVulnerabilitySeverity = viewModel.Vulnerabilities.Max(v => v.Severity);
+            }
+            else
+            {
+                viewModel.Vulnerabilities = new List<PackageVulnerability>();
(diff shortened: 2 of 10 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -784,8 +804,31 @@
                                                 deprecationTitle += " is deprecated.";
                                             }
 
+                                            useAlso = true;
+                                        }
+
+                                        if (Model.IsPackageVulnerabilitiesEnabled && packageVersion.Vulnerabilities != null && packageVersion.Vulnerabilities.Any())
```
skofman1, line 810: >packageVersion.Vulnerabilities != null && packageVersion.Vulnerabilities.Any() [](start = 85, length = 78)

This condition is repeated a couple of times. Can we move it to a shared location?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -784,8 +804,31 @@
...
+                                        {
+                                            vulnerabilitiesTitle = useAlso ? "Also, " : "";
+                                            var severity = Enum.GetName(typeof(PackageVulnerabilitySeverity), packageVersion.MaxVulnerabilitySeverity).ToLowerInvariant();
+                                            vulnerabilitiesTitle += packageVersion.Version + " has at least one vulnerability with " + severity + " severity.";
```
skofman1, line 814: >packageVersion.Version + " has at least one vulnerability with " + severity + " severity."; [](start = 68, length = 91)

consider string.Format for readability.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -753,15 +772,16 @@
                                             </td>
                                         }
                                     }
-                                    @if (Model.IsPackageDeprecationEnabled)
+
+                                    @if (Model.IsPackageDeprecationEnabled || Model.IsPackageVulnerabilitiesEnabled)
                                     {
```
skofman1, line 777: There's a bunch of logic here and it's not testable because it's in the cshtml. Consider moving it to the ViewModel so you can add UTs

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
drewgillies: Done--thanks. I wanted to do an interpolated string, but we're targeting a .NET version that doesn't support them in razor. string.Format is 👍 though.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
drewgillies: I've taken a simpler approach here and called the column "Package Warnings"--I don't think we need to customize a hidden column header based on feature flags.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -684,9 +689,9 @@
                             {
                                 <th aria-hidden="true" abbr="Signature Information"></th>
                             }
-                            @if (Model.IsPackageDeprecationEnabled)
+                            @if (Model.IsPackageDeprecationEnabled || Model.IsPackageVulnerabilitiesEnabled)
                             {
-                                <th aria-hidden="true" abbr="Deprecation Information"></th>
+                                <th aria-hidden="true" abbr ="Package Warnings" />
```
zhhyu, line 694: Should we still keep "</th>" here?

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
zhhyu: If I understand correctly, we have a code block here to obtain "iconTitle", and set up a warning icon with the associated title based on some checks. Maybe it's better that we add another property named "warningIconTitle" (and with another additional boolean property named "hasWarningIcon") in the view model, and when we create the view model in the factory, we can run these checks and set up this property. Then in the view, we can easily check this property to determine whether to show the icon and title for a specific version. 
In this way, we can move this logic to the factory, and some tests can be added to cover these checks.

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -5,22 +5,36 @@ $(function () {
     window.nuget.configureExpander("rename-content-container", "ChevronDown", null, "ChevronUp");
     configureExpanderWithEnterKeydown($('#show-rename-content-container'));
 
+    // Configure the vulnerability information container
+    var expanderAttributes = ['data-toggle', 'data-target', 'aria-expanded', 'aria-controls', 'tabindex'];
+    var vulnerabilitiesContainer = $('#show-vulnerabilities-content-container');
+    if ($('#vulnerabilities-content-container').children().length) {
+        // If the deprecation information container has content, configure it as an expander.
```
zhhyu, line 12: nit: "// If the vulnerabilities information container ..."

src/Bootstrap/less/theme/page-display-package.less
```diff
@@ -63,6 +63,50 @@
     }
   }
 
+  .vulnerabilities-container {
```
zhhyu, line 66: Maybe it's better that we share styles with the deprecation container? We can have a "warning-container" to define these shared styles, and it will be easier to maintain if we need more warning containers in the future.

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
```diff
@@ -186,10 +186,17 @@ public DisplayPackageViewModelFactory(IIconUrlProvider iconUrlProvider)
...
+                && vulnerabilities != null && vulnerabilities.Any())
             {
                 viewModel.Vulnerabilities = vulnerabilities;
-                viewModel.MaxVulnerabilitySeverity = vulnerabilities.Max(v => v.Severity);
+                viewModel.MaxVulnerabilitySeverity = viewModel.Vulnerabilities.Max(v => v.Severity);
+            }
+            else
+            {
+                viewModel.Vulnerabilities = new List<PackageVulnerability>();
+                viewModel.MaxVulnerabilitySeverity = PackageVulnerabilitySeverity.Low;
```
zhhyu, line 199: Are there any considerations to have an empty list and a low "MaxVulnerabilitySeverity" if there are no vulnerabilities? I see that we check "Vulnerabilities.Any()" in the view. Maybe it's better that we just leave "Vulnerabilities" as null, and take a null check of "Vulnerabilities" when there are no vulnerabilities? Setting "MaxVulnerabilitySeverity" as low seems a little confusing.

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -5,22 +5,36 @@ $(function () {
     window.nuget.configureExpander("rename-content-container", "ChevronDown", null, "ChevronUp");
     configureExpanderWithEnterKeydown($('#show-rename-content-container'));
 
+    // Configure the vulnerability information container
```
zhhyu, line 8: nit: maybe it's better to put this comment below the next line as "expanderAttributes" is also used for the deprecation container.

src/NuGetGallery/Views/Packages/_DisplayPackageVulnerabilities.cshtml
```diff
@@ -0,0 +1,54 @@
...
+            <b>Details</b>
+            <table width="100%" class="vulnerabilities-list">
+                @{
+                    foreach (var vulnerability in Model.Vulnerabilities)
+                    {
+                        var truncatedUrl = vulnerability.AdvisoryUrl;
+                        if (truncatedUrl.Length > 50)
+                        {
+                            truncatedUrl = truncatedUrl.Substring(0, 47) + "...";
+                        }
```
zhhyu, line 41: How do we pick up the number 50 or 47? This may lead to the accessibility issue. How will a reflow affect the appearance? Feel free to ignore this if we have already used the trick.

src/Bootstrap/less/theme/page-display-package.less
drewgillies: @zhhyu Looks like this change isn't practical--the js script that toggles chevron state uses the outer style name to identify the panel on which to rotate it when expanded. Messing with an outer style name has cascading effects. It could be feasible to change inner style names but this wouldn't achieve much as the names would then be duplicated in each outer style.

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
drewgillies: There's no consideration except that it defaults to low. Changing it now would be disruptive to the db and I'm not sure much would be gained. I have made the null change though.

src/NuGetGallery/Views/Packages/_DisplayPackageVulnerabilities.cshtml
drewgillies: I'm removing this truncation. It will simply wrap if it needs to, and the table caters for that nicely now.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
drewgillies: I've made this consistent--thanks.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
drewgillies: Good call--this is much more robust now.

drewgillies:
Note that I've also made calls to the deprecation and vulnerabilities services conditional on the respective feature flags (for perf and cleanness), so UTs needed updating to not verify calls against them that we no longer expect to happen.

skofman1 (approved):
:shipit:

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
```diff
@@ -192,15 +193,76 @@ public DisplayPackageViewModelFactory(IIconUrlProvider iconUrlProvider)
...
+
             return viewModel;
         }
 
+        private static string GetWarningIconTitle(
+            string version, 
+            PackageDeprecation deprecation,
+            PackageVulnerabilitySeverity? maxSeverity)
+        {
+            // We want a tooltip title for the warning icon, which concatenates deprecation and vulnerability information cleanly
```
zhhyu, line 222: nit: empty line

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
```diff
@@ -192,15 +193,76 @@ public DisplayPackageViewModelFactory(IIconUrlProvider iconUrlProvider)
...
+                viewModel.MaxVulnerabilitySeverity = maxSeverity.Value;
+            }
+            else
+            {
+                viewModel.Vulnerabilities = null;
+                viewModel.MaxVulnerabilitySeverity = default;
             }
 
+            viewModel.PackageWarningIconTitle =
+                GetWarningIconTitle(viewModel.Version, deprecation, maxSeverity);
```
zhhyu, line 212: nit: maybe it's more clear as:  maxSeverity -> maxVulnerabilitySeverity

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -753,41 +758,16 @@
...
-                                                deprecationTitle += " is deprecated.";
-                                            }
 
-                                            <td class="package-icon-cell">
-                                                <i class="ms-Icon ms-Icon--Warning package-icon" title="@deprecationTitle"></i>
-                                            </td>
-                                        }
+                                    @if (string.IsNullOrEmpty(packageVersion.PackageWarningIconTitle))
+                                    {
+                                        <td class="package-icon-cell" aria-hidden="true" />
```
zhhyu, line 764: Should we still keep "</td>" here?

drewgillies: I'm not sure what you mean.

drewgillies: I'm not sure of the advantage of empty tags over self-closing ones. I've changed this to an empty tag for consistency, but it just seems like a waste of chars to me. Self-closing tags are neater IMHO.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -1475,6 +1406,225 @@ public async Task ShowsFirstDeprecationPerPackage()
...
+                deprecationService.Verify();
+            }
+
+            [Theory]
+            [InlineData(false, false, "")]
+            [InlineData(true, false, "{0} is deprecated because it's legacy and no longer maintained.")]
+            [InlineData(false, true, "{0} has at least one vulnerability with {1} severity.")]
+            [InlineData(true, true, "{0} is deprecated because it's legacy and no longer maintained; {0} has at least one vulnerability with {1} severity.")]
+            public async Task ShowsCombinedDeprecationAndVulnerabilitiesIconTitle(
+                bool isDeprecationEnabled,
```
zhhyu, line 1510: Great test!

zhhyu (approved):

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
drewgillies: Thanks! :)

PR #8359, opened 2020-12-22
update term used for .NET tools

adamralph:
As per the [linked document](https://aka.ms/global-tools), these are now referred to as ".NET tools" —

> How to manage .NET tools
> 
> A .NET tool is a special NuGet package that contains a console application.

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -34,7 +34,7 @@
                 CommandPrefix = "> ",
                 InstallPackageCommand = string.Format("dotnet tool install --global {0} --version {1}", Model.Id, Model.Version),
                 AlertLevel = AlertLevel.Info,
-                AlertMessage = "This package contains a <a href='https://aka.ms/global-tools'>.NET Core Global Tool</a> you can call from the shell/command line.",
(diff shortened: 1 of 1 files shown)


drewgillies (approved):

joelverhagen:
@wli3 - can you check the wording here? The command _does_ install as a global tool but it seems like the "Core" team is being dropped. The message looks like this today:


Example: https://www.nuget.org/packages/dotnet-reportgenerator-globaltool/
image


@augustoproiete - it looks like your proposed UI (https://github.com/NuGet/NuGetGallery/issues/8373) covers this change, right? I think with your tab name enhancement to say "global" vs. "local", we don't need to say "global" in the subtext.

augustoproiete:
@joelverhagen Correct. The second bullet point on #8373 suggests the exact same wording that this PR is proposing (which LGTM). The tab enhancement would be to help distinguish global vs local.

wli3:
@kathleendollard has the call. But I think this change is good.

skofman1:
@adamralph , thanks for creating the PR and your proposal. Since @augustoproiete's suggestion in https://github.com/NuGet/NuGetGallery/issues/8373 addresses your concern and provides a more complete solution, we will proceed with that.

adamralph:
https://github.com/NuGet/NuGetGallery/issues/8373 is just an issue and is still being discussed. Why not merge and deploy this in the meantime? Sure, it doesn't have all the other stuff that https://github.com/NuGet/NuGetGallery/issues/8373 is proposing, but it's still a valuable change on it's own no?

Issue #8361, opened 2020-12-22
Gallery authors should see a message on the Manage Packages page if the package is vulnerable


drewgillies:
Spec: https://github.com/NuGet/Home/blob/dev/designs/PackageVulnerability/NuGet.orgVulnerabilitiesPhase1.md

drewgillies:
Metrics in DEV: flag off: 2846ms, flag on: 3699ms

drewgillies:
Metrics in INTL flag off: 2087ms, flag on: 2068ms   (1547 package ids)

drewgillies:
Metrics in PROD: flag off: 12542ms, flag on: 13443ms

drewgillies:
We can use caching to support this now.

PR #8362, opened 2020-12-22
Expand view models for vulnerabilities in manage packages page

drewgillies:
Addresses view model portion of https://github.com/NuGet/NuGetGallery/issues/8361 (I split the PR as best as is practical because it's large-ish). Full description of change to customer experience in PR (view changes) on this issue.

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/IPackageService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/IPackageService.cs
@@ -84,9 +84,11 @@ namespace NuGetGallery
         Package FilterLatestPackageBySuffix(IReadOnlyCollection<Package> packages,
             string version, bool prerelease);
 
-        IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted, bool includeVersions = false);
+        IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted, bool includeVersions = false, bool includeVulnerabilities = false);
 
-        IEnumerable<Package> FindPackagesByAnyMatchingOwner(User user, bool includeUnlisted, bool includeVersions = false);
+        IEnumerable<Package> FindPackagesByAnyMatchingOwner(User user, bool includeUnlisted, bool includeVersions = false, bool includeVulnerabilities = false);
+
+        IQueryable<Package> GetLatestVersions(IQueryable<Package> packages);
 
         IQueryable<PackageRegistration> FindPackageRegistrationsByOwner(User user);
 
--- a/src/NuGetGallery.Services/PackageManagement/PackageService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageService.cs
@@ -399,9 +399,12 @@ namespace NuGetGallery
             return package;
         }
 
-        public IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted, bool includeVersions = false)
+        public IEnumerable<Package> FindPackagesByOwner(User user, 
+            bool includeUnlisted, 
+            bool includeVersions = false,
+            bool includeVulnerabilities = false)
         {
-            return GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions);
+            return GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions, includeVulnerabilities);
         }
 
         /// <summary>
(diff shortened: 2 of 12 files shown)


src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -399,9 +399,11 @@ Package GetLatestStable()
             return package;
         }
 
-        public IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted, bool includeVersions = false)
+        public IEnumerable<Package> FindPackagesByOwner(User user, 
+            bool includeUnlisted, 
+            bool includeVersions = false)
         {
-            return GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions);
+            return GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions, includeVulnerabilities: false);
```
zhhyu, line 406: How about that we also provide the flexibility to callers of this method?
```
public IEnumerable<Package> FindPackagesByOwner(User user, 
            bool includeUnlisted, 
            bool includeVersions = false,
            bool includeVulnerabilities = false)
```

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -573,20 +580,24 @@ public virtual ActionResult Packages()
...
-            IEnumerable<Package> packages,
+            IQueryable<Package> packages,
             User currentUser,
             bool wasAADLoginOrMultiFactorAuthenticated,
             Func<Package, bool> predicate)
         {
-            var listedPackages = packages
-                .Where(p => predicate(p))
+            var filteredPackages = packages.Where(p => predicate(p));
+            var latestPackages = PackageService.GetLatestVersions(filteredPackages);
```
zhhyu, line 589: Could we have any hints about the reason to introduce the filter here and get latest packages, compared with the old approach?

src/NuGetGallery.Services/PackageManagement/PackageService.cs
drewgillies: It's unused, but a trivial change, so ...done.

src/NuGetGallery/Controllers/UsersController.cs
drewgillies: Yes, I need to cache *all* of the versions that clear the filter before filtering to latest versions only. It is this all-versions collection that I need to examine to determine if *any* version (not just the latest) has a vulnerability. Then we decorate the list item with a bang. We've not needed this distinction before now.

I split out latestPackages into a separate line for readability in the following statement, that it's the latest versions we're using to create our list.

src/NuGetGallery/Services/PackageVulnerabilitiesService.cs
```diff
@@ -41,5 +41,27 @@ public PackageVulnerabilitiesService(IEntitiesContext entitiesContext)
...
+
+            Package first = null;
+            foreach (var package in packages)
+            {
+                if (package.VulnerablePackageRanges != null
+                    && package.VulnerablePackageRanges.Any()
+                    && package.VulnerablePackageRanges.FirstOrDefault(vpr => vpr.Vulnerability != null) != null)
+                {
+                    first = package;
+                    break;
```
zhhyu, line 60: We can simplify this part: here we can return true and at the end, we can return false. Then there is no need to define "first".

src/NuGetGallery/Controllers/UsersController.cs
zhhyu: Thanks! That makes sense and is a workable solution.

src/NuGetGallery.Services/PackageManagement/IPackageService.cs
```diff
@@ -84,9 +84,11 @@ public interface IPackageService : ICorePackageService
         Package FilterLatestPackageBySuffix(IReadOnlyCollection<Package> packages,
             string version, bool prerelease);
 
-        IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted, bool includeVersions = false);
+        IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted, bool includeVersions = false, bool includeVulnerabilities = false);
 
-        IEnumerable<Package> FindPackagesByAnyMatchingOwner(User user, bool includeUnlisted, bool includeVersions = false);
+        IEnumerable<Package> FindPackagesByAnyMatchingOwner(User user, bool includeUnlisted, bool includeVersions = false, bool includeVulnerabilities = false);
+
+        IQueryable<Package> GetLatestVersions(IQueryable<Package> packages);
```
zhhyu, line 91: Maybe provide a summary for this new introduced method, because we have another method named "FilterLatestPackage".

zhhyu (approved):

tests/NuGetGallery.Facts/Services/PackageVulnerabilitiesServiceFacts.cs
```diff
@@ -132,12 +145,30 @@ private void SetUp()
...
+                Version = "0.1.0",
+                VulnerablePackageRanges = null
+            };
+            _packageNotVulnerable011 = new Package
+            {
+                Key = 5,
+                PackageRegistration = _registrationNotVulnerable,
+                Version = "0.1.1",
                 VulnerablePackageRanges = null
             };
```
zhhyu, line 161: Could we have some test coverage on other cases?

zhhyu:
Looks good to me. I suggest getting another approval before checking in this change.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -521,7 +524,9 @@ public virtual ActionResult Packages()
 
             var wasAADLoginOrMultiFactorAuthenticated = User.WasMultiFactorAuthenticated() || User.WasAzureActiveDirectoryAccountUsedForSignin();
 
-            var packages = PackageService.FindPackagesByAnyMatchingOwner(currentUser, includeUnlisted: true);
+            var packages = PackageService.FindPackagesByAnyMatchingOwner(
```
skofman1, line 527: >FindPackagesByAnyMatchingOwner [](start = 42, length = 30)

This page is super slow, especially for orgs with a big amount of packages (like MSFT). I recommend you test the perf on DEV before merging. Loading version information may be super expensive.

drewgillies:
Closing this PR for now (but keeping the branch) as this approach performs badly--in the order or a 3x or 4x degradation. Starting afresh and the new approach is tracked in the epic.

Issue #7850, opened 2020-02-20
Add fuget.org link to package details page


anangaur:
This is the issue tracking the community PR: https://github.com/NuGet/NuGetGallery/pull/7832 as it needs some UX and messaging work before going live.

Rationale for the feature: 
fuget.org shows API level data for the packages and lets you compare the APIs across versions. This has been super useful for developers and they have already been using this. It would make sense for developers to directly explore their package on fuget.org from nuget.org's package details page.

anangaur:
We would need to add a couple of icons and messaging to tell users that fuget.org is a third-party site not owned by Microsoft. Working on a proposal...

anangaur:
Here is the proposal (undergoing UX review):

![image](https://user-images.githubusercontent.com/14800916/75476419-68936500-594f-11ea-81ee-a8cd78cc9279.png)

Text for tooltip: **TBD**
Accessibility review: **TBD**

praeclarum:
Hello, I'm very excited for this feature!

chgill-MSFT:
Plans as of now:

**Icon (this is for show, svg will be used):** <img src="https://user-images.githubusercontent.com/15097183/84452554-f64b4580-ac0a-11ea-9f7c-7c0330d85766.png" alt="Image" height="30">

**Link tooltip:** “Explore additional package info on fuget.org”
**Info (i) icon tooltip:** “fuget.org is a 3rd party website, not controlled by Microsoft. This link is made available to you per the NuGet Terms of Use.”

**Link accessibility:** "Link, open in fuget.org explorer"
**Info (i) icon accessibility:** “fuget.org is a 3rd party website, not controlled by Microsoft. This link is made available to you per the NuGet Terms of Use.”

PR #7832, opened 2020-02-01
Add FuGet link

Krzysztof-Cieslak:
FuGet is a NuGet package browser combined with an API browser. It's a really useful tool for .Net developers and it would be good to expose it directly from the package page. It enables developers to inspect content of NuGet packages, the API of packages, and even API diffs between different versions of the package. Here's example page - https://www.fuget.org/packages/FSharp.Core 

The PR adds link to FuGet page for the current package in the Info section of the page.

PR #8364, opened 2020-12-23
Add fuget.org links to packages

drewgillies:
Addresses: https://github.com/NuGet/NuGetGallery/issues/7850

Packages will now have an additional link per specs outlined here: https://github.com/NuGet/NuGetGallery/pull/7832, linking to the package's fuget.org view.

image


drewgillies:
FYI @chgill-MSFT @anangaur 
Chris--I've put a fallback png for the icon which was simply scraped from the other PR (I've used the provided SVG as the main icon file). If you want to include a better fallback, let me know.

skofman1 (approved):
:shipit:

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
```diff
@@ -163,6 +163,12 @@ public DisplayPackageViewModelFactory(IIconUrlProvider iconUrlProvider)
                 viewModel.ProjectUrl = projectUrl;
             }
 
+            var fugetUrl = $"https://www.fuget.org/packages/{package.Id}/{package.Version}";
```
joelverhagen, line 166: is this the "full version" or the "normalized version"? We should use the normalized version because that includes all URL safe characters. The build metadata is included in the full and original, non-normalized version strings. So `1.0.0+foo` is a valid version but we only want `1.0.0` in the URL. To be safe, it's best to use URL encoding methods to build this URL so you don't have to think about it.

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
```diff
@@ -34,6 +34,7 @@ public class DisplayPackageViewModel : ListPackageItemViewModel
         public bool IsAtomFeedEnabled { get; set; }
         public bool IsPackageDeprecationEnabled { get; set; }
         public bool IsPackageVulnerabilitiesEnabled { get; set; }
+        public bool IsFuGetLinksEnabled { get; set; }
```
joelverhagen, line 37: Couldn't we just make the `FuGetUrl` == `null` when the feature flag is disabled? Might be simpler.

tests/NuGetGallery.Facts/ViewModels/DisplayPackageViewModelFacts.cs
```diff
@@ -104,6 +104,26 @@ public void ItInitializesProjectUrl(string projectUrl, string expected)
             Assert.Equal(expected, model.ProjectUrl);
         }
 
+        [Theory]
+        [InlineData("foo", "1.0.0", "https://www.fuget.org/packages/foo/1.0.0")]
+        [InlineData("foo", "1.1.0", "https://www.fuget.org/packages/foo/1.1.0")]
+        [InlineData("foo.bar", "1.1.0-beta", "https://www.fuget.org/packages/foo.bar/1.1.0-beta")]
```
joelverhagen, line 110: nit: add uppercase package IDs and versions as well as versions with build metadata to assert the respective behaviors

joelverhagen (approved):

loic-sharma (approved):

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
drewgillies: Turns out it's not so simple--we read the feature flag after calling this Create method in the controller, so we'd need to move the feature flag read ahead of it and pass either its value or the fuget url itself into the creator. It would be simpler in the view, yes, but it's more tangled in the model. Perf savings would be negligible as we'd testing a bool either way, and merely avoiding the construction of a string if it's false.

tests/NuGetGallery.Facts/ViewModels/DisplayPackageViewModelFacts.cs
drewgillies: This is at the model level, not ingestion. I don't see these transformations at this level (DisplayPackageViewModelFactory). Setting a version and not setting a normalized version will leave the normalized version blank after calling the endpoint. Setting an id or version with an uppercase character will result in an uppercase character in the url. Should there be some intelligence in this factory which strips metadata strings or normalizes versions and ids? I can add this if necessary. I expect it will be splitting on + and lowercasing.

drewgillies: Although normalizing versions also strips leading zeroes etc. Anyway, the first question is... which parts of this should be happening here?

src/NuGetGallery/ViewModels/DisplayPackageViewModel.cs
joelverhagen: Sounds good!

src/NuGetGallery/Helpers/ViewModelExtensions/DisplayPackageViewModelFactory.cs
drewgillies: Changed it to `NormalizedVersion`. Thanks!

PR #8375, opened 2021-01-12
Add F# interactive instructions

loic-sharma:
Screenshot:

image

Addresses https://github.com/NuGet/Engineering/issues/3527

/cc @cartermp @jonsequitur

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -91,7 +91,18 @@
                 Id = "paket-cli",
                 CommandPrefix = "> ",
                 InstallPackageCommand = string.Format("paket add {0} --version {1}", Model.Id, Model.Version),
-            }
+            },
+
+            new PackageManagerViewModel("F# Interactive")
+            {
+                Id = "fsharp-interactive",
+                CommandPrefix = "> ",
+                InstallPackageCommand = string.Format("#r \"nuget: {0}, {1}\"", Model.Id, Model.Version),
+                AlertLevel = AlertLevel.Info,
+                AlertMessage = string.Format(
+                    "For F# interactive projects that support <a href=\"{0}\">#r syntax</a>, copy this into the project file to reference the package.",
+                    "https://docs.microsoft.com/en-us/dotnet/fsharp/tools/fsharp-interactive/#referencing-packages-in-f-interactive")
+            },
         };
     }
 }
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -91,7 +91,18 @@
...
+            },
+
+            new PackageManagerViewModel("F# Interactive")
+            {
+                Id = "fsharp-interactive",
+                CommandPrefix = "> ",
+                InstallPackageCommand = string.Format("#r \"nuget: {0}, {1}\"", Model.Id, Model.Version),
+                AlertLevel = AlertLevel.Info,
+                AlertMessage = string.Format(
+                    "For F# interactive projects that support <a href=\"{0}\">#r syntax</a>, copy this into the project file to reference the package.",
```
loic-sharma, line 103: I believe that `project file` here is incorrect, I will confirm with the F# team:

```suggestion
                    "For F# interactive projects that support <a href=\"{0}\">#r syntax</a>, copy this into the source file to reference the package.",
```

joelverhagen:
Can you put a screenshot in the description?

loic-sharma:
@joelverhagen Added!

joelverhagen (approved):
Make sure it looks at least usable/reasonable at smaller resolutions or windows sizes then :shipit:

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
baronfel: (not F# team but active community member) This is correct terminology, project files aren't used in interactive contexts.

zhhyu (approved):

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
cartermp: Seconded @baronel's note here

cartermp: Slightly revised proposed wording:

> For F# scripts that support #r syntax, copy this into the source code to reference the package.

loic-sharma: Great, thanks everyone for the feedback! I've updated the text accordingly

loic-sharma:
@joelverhagen I verified this is reasonably usable in a mobile view. Sadly, the tabs wrap onto a new line:

image

This is existing behavior though on nuget.org:

image

I don't plan to fix this as part of this change.

joelverhagen:
I've noticed this wrapping before and frankly, it's not that bad. Perhaps we could use a bit of polish there but it's still quite usable. The command itself wraps nicely so I'm happy with that if you are :)

PR #8380, opened 2021-01-19
[Stats]Enable Alternate storage for downloads.v1.json

ryuyu:
Enables (behind a feature flag) support for allowing an alternate location for downloads.v1.json source.


skofman1 (approved):
:shipit:

joelverhagen:
I must have missed a design spec or something -- what's the motivation here? Are we moving the file and this allows side-by-side?

ryuyu:
The idea is to support sourcing downloads.v1.json from multiple locations.
This way, we can have multiple generators of downloads.v1.json and allow all to operate without clobbering each other.

PR #8405, opened 2021-01-29
improved the #r directive instructions

filipw:
This is a follow up to https://github.com/NuGet/NuGetGallery/pull/8375

Since `#r` is not exclusive to F#, it is a bit inaccurate to use the current header for the instructions - it would make sense to use a more generic name for it. 
Therefore I propose to rename `F# Interactive` to `.NET Interactive` as it then correctly captures both C# and F#.

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -93,14 +93,14 @@
                 InstallPackageCommand = string.Format("paket add {0} --version {1}", Model.Id, Model.Version),
             },
 
-            new PackageManagerViewModel("F# Interactive")
+            new PackageManagerViewModel(".NET Interactive")
             {
-                Id = "fsharp-interactive",
+                Id = "net-interactive",
                 CommandPrefix = "> ",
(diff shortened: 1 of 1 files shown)


cartermp:
Note that the tab is correct as-is too, since F# Interactive is orthogonal to .NET Interactive. The disjoint nature of package acquisition for C# in "interactive" scenarios when compared to F# likely makes it more challenging to represent.

filipw:
> Note that the tab is correct as-is too, since F# Interactive is orthogonal to .NET Interactive.

I meant `.NET Interactive` as a general class of interactive tools for .NET, not specifically the product `.NET Interactive`. I have renamed it to `Script & Interactive` which better conveys the meaning.

This captures both F# and C# *if* this syntax is supported in a given runner, as the explanation underneath states.

LadyNaggaga:
PM - .NET Interactive here.  @filipw recommend tab name Script & Interactive sounds like a fair description. I have chatted with @cartermp about this as well.  A note would need to be included along the lines of _#r directive can be used in F# Interactive,  C# scripting and .NET Interactive_  thoughts?

jcjiang:
That makes sense to me. I agree that we should add the additional context and proceed with the PR.

drewgillies:
Screenshot FYI
image

drewgillies (approved):
Thanks for your contribution, @filipw! Merging.

PR #8412, opened 2021-02-04
Extend vulnerabilities verification to metadata

drewgillies:
Addresses: https://github.com/NuGet/Engineering/issues/3623

The verification test already existed for doing a full set check of database vulnerabilities--this change adds the capability of downloading registration metadata blobs to determine that the pipeline has propagated incremental changes correctly.

The config file changes allow you to test either or both.

### DIFF
--- a/src/VerifyGitHubVulnerabilities/Configuration/VerifyGitHubVulnerabilitiesConfiguration.cs
+++ b/src/VerifyGitHubVulnerabilities/Configuration/VerifyGitHubVulnerabilitiesConfiguration.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.ServiceModel.Configuration;
 
 namespace VerifyGitHubVulnerabilities.Configuration
 {
@@ -16,5 +17,20 @@ namespace VerifyGitHubVulnerabilities.Configuration
         /// The personal access token to use to authenticate with GitHub.
         /// </summary>
         public string GitHubPersonalAccessToken { get; set; }
+
+        /// <summary>
+        /// The v3 index URI string for fetching registration metadata
+        /// </summary>
+        public string NuGetV3Index { get; set; }
+
+        /// <summary>
+        /// Whether to verify GitHubVulnerabilities in the gallery database
+        /// </summary>
+        public bool VerifyDatabase { get; set; } = true;
+
+        /// <summary>
+        /// Whether to verify GitHubVulnerabilities in the registration blobs
+        /// </summary>
+        public bool VerifyRegistrationMetadata { get; set; } = true;
     }
 }
--- a/src/VerifyGitHubVulnerabilities/Configuration/VerifyGitHubVulnerabilitiesConfiguration.cs
+++ b/src/VerifyGitHubVulnerabilities/Configuration/VerifyGitHubVulnerabilitiesConfiguration.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.ServiceModel.Configuration;
 
 namespace VerifyGitHubVulnerabilities.Configuration
 {
--- a/src/VerifyGitHubVulnerabilities/Job.cs
+++ b/src/VerifyGitHubVulnerabilities/Job.cs
@@ -34,13 +34,15 @@ namespace VerifyGitHubVulnerabilities
             Console.WriteLine($" FOUND {advisories.Count} advisories.");
 
             Console.WriteLine("Fetching vulnerabilities from DB...");
-            var verifier = new PackageVulnerabilitiesVerifier(_serviceProvider);
-            var ingestor = new AdvisoryIngestor(verifier, new GitHubVersionRangeParser());
+            var ingestor = _serviceProvider.GetRequiredService<IAdvisoryIngestor>();
             await ingestor.IngestAsync(advisories);
 
+            var verifier =
+                _serviceProvider.GetRequiredService<IPackageVulnerabilitiesManagementService>() as
+                    PackageVulnerabilitiesVerifier;
             Console.WriteLine(verifier.HasErrors ? 
                 "DB does not match GitHub API - see stderr output for details" :
-                "DB matches GitHub API!");
+                "DB/metadata matches GitHub API!");
         }
 
         protected override void ConfigureJobServices(IServiceCollection services, IConfigurationRoot configurationRoot)
@@ -68,6 +70,10 @@ namespace VerifyGitHubVulnerabilities
             containerBuilder
                 .RegisterType<AdvisoryIngestor>()
                 .As<IAdvisoryIngestor>();
+
+            containerBuilder
+                .RegisterType<PackageVulnerabilitiesVerifier>()
+                .As<IPackageVulnerabilitiesManagementService>();
         }
 
         protected void ConfigureGalleryServices(ContainerBuilder containerBuilder)
--- a/src/VerifyGitHubVulnerabilities/Job.cs
+++ b/src/VerifyGitHubVulnerabilities/Job.cs
@@ -34,7 +34,7 @@ namespace VerifyGitHubVulnerabilities
             Console.WriteLine($" FOUND {advisories.Count} advisories.");
 
             Console.WriteLine("Fetching vulnerabilities from DB...");
-            var verifier = new PackageVulnerabilitiesVerifier(_serviceProvider.GetRequiredService<IEntitiesContext>());
+            var verifier = new PackageVulnerabilitiesVerifier(_serviceProvider);
             var ingestor = new AdvisoryIngestor(verifier, new GitHubVersionRangeParser());
             await ingestor.IngestAsync(advisories);
 
--- a/src/VerifyGitHubVulnerabilities/Job.cs
+++ b/src/VerifyGitHubVulnerabilities/Job.cs
@@ -37,9 +37,7 @@ namespace VerifyGitHubVulnerabilities
             var ingestor = _serviceProvider.GetRequiredService<IAdvisoryIngestor>();
             await ingestor.IngestAsync(advisories);
 
-            var verifier =
-                _serviceProvider.GetRequiredService<IPackageVulnerabilitiesManagementService>() as
-                    PackageVulnerabilitiesVerifier;
+            var verifier = _serviceProvider.GetRequiredService<IPackageVulnerabilitiesVerifier>();
             Console.WriteLine(verifier.HasErrors ? 
                 "DB does not match GitHub API - see stderr output for details" :
                 "DB/metadata matches GitHub API!");
@@ -73,7 +71,9 @@ namespace VerifyGitHubVulnerabilities
 
             containerBuilder
                 .RegisterType<PackageVulnerabilitiesVerifier>()
-                .As<IPackageVulnerabilitiesManagementService>();
+                .As<IPackageVulnerabilitiesManagementService>()
+                .As<IPackageVulnerabilitiesVerifier>()
+                .SingleInstance();
         }
 
         protected void ConfigureGalleryServices(ContainerBuilder containerBuilder)
--- a/src/VerifyGitHubVulnerabilities/Verify/IPackageVulnerabilitiesVerifier.cs
+++ b/src/VerifyGitHubVulnerabilities/Verify/IPackageVulnerabilitiesVerifier.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.Entities;
+using NuGetGallery;
+
+namespace VerifyGitHubVulnerabilities.Verify
+{
+    /// <summary>
+    /// We need a <see cref=IPackageVulnerabilitiesManagementService" /> for inserting the verifier into the ingestion job,
+    /// so we extend for our extra reporting data.
+    /// </summary>
+    public interface IPackageVulnerabilitiesVerifier : IPackageVulnerabilitiesManagementService
+    {
+        /// <summary>
+        /// An error flag for verification
+        /// </summary>
+        bool HasErrors { get; }
+    }
+}
--- a/src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
+++ b/src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
@@ -7,7 +7,6 @@ using System.Data.Entity;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Extensions.DependencyInjection;
 using NuGet.Configuration;
 using NuGet.Protocol;
 using NuGet.Protocol.Core.Types;
@@ -23,19 +22,22 @@ namespace VerifyGitHubVulnerabilities.Verify
         private readonly VerifyGitHubVulnerabilitiesConfiguration _configuration;
         private readonly IEntitiesContext _entitiesContext;
 
-        private PackageMetadataResource _packageMetadataResource;
+        private Lazy<Task<PackageMetadataResource>> _packageMetadataResource;
         private Dictionary<string, IEnumerable<IPackageSearchMetadata>> _packageMetadata;
 
-        public PackageVulnerabilitiesVerifier(IServiceProvider serviceProvider)
+        static readonly SemaphoreSlim semaphoreSlim = new SemaphoreSlim(1);
+
+        public PackageVulnerabilitiesVerifier(VerifyGitHubVulnerabilitiesConfiguration configuration,
+            IEntitiesContext entitiesContext)
         {
-            _configuration = serviceProvider.GetRequiredService<VerifyGitHubVulnerabilitiesConfiguration>() ??
-                             throw new Exception(
-                                 $"{nameof(VerifyGitHubVulnerabilitiesConfiguration)} service cannot be null");
+            _configuration = configuration;
             if (_configuration.VerifyDatabase)
             {
-                _entitiesContext = serviceProvider.GetRequiredService<IEntitiesContext>() ??
-                                   throw new Exception($"{nameof(IEntitiesContext)} service cannot be null for database operations");
+                _entitiesContext = entitiesContext;
             }
+
+            _packageMetadata = new Dictionary<string, IEnumerable<IPackageSearchMetadata>>();
+            _packageMetadataResource = new Lazy<Task<PackageMetadataResource>>(InitializeMetadataResourceAsync);
         }
 
         public bool HasErrors { get; private set; }
@@ -63,7 +65,7 @@ namespace VerifyGitHubVulnerabilities.Verify
             // is no longer present. Covering withdrawn advisory processing in the database will be adequate.
             if (_configuration.VerifyRegistrationMetadata && !withdrawn)
             {
-                return VerifyVulnerabilityInMetadata(vulnerability);
+                VerifyVulnerabilityInMetadata(vulnerability);
             }
 
             return Task.CompletedTask;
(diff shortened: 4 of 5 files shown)


src/VerifyGitHubVulnerabilities/Job.cs
```diff
@@ -34,7 +34,7 @@ public override async Task Run()
             Console.WriteLine($" FOUND {advisories.Count} advisories.");
 
             Console.WriteLine("Fetching vulnerabilities from DB...");
-            var verifier = new PackageVulnerabilitiesVerifier(_serviceProvider.GetRequiredService<IEntitiesContext>());
+            var verifier = new PackageVulnerabilitiesVerifier(_serviceProvider);
```
loic-sharma, line 37: Passing the `IServiceProvider` to a type's constructor and letting it find its dependencies is called the "service locator" pattern - it is an anti-pattern that isn't recommended (see [this](https://docs.microsoft.com/en-us/dotnet/core/extensions/dependency-injection-guidelines#recommendations)). It makes it difficult to reuse or test `PackageVulnerabilitiesVerifier` - how do I know which types I need to register in the dependency injection container for the constructor to work? Also it's a slippery path that can lead you to storing the service provider and breaking dependency injection scopes.

Do we need to manually `new` up the `PackageVulnerabilitiesVerifier` type? Could we instead register it in the dependency injection container and create it using `GetRequiredService`?

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -2,23 +2,40 @@
...
 
-        public PackageVulnerabilitiesVerifier(
-            IEntitiesContext entitiesContext)
+        private PackageMetadataResource _packageMetadataResource;
+        private Dictionary<string, IEnumerable<IPackageSearchMetadata>> _packageMetadata;
+
+        public PackageVulnerabilitiesVerifier(IServiceProvider serviceProvider)
         {
-            _entitiesContext = entitiesContext ?? throw new ArgumentNullException(nameof(entitiesContext));
+            _configuration = serviceProvider.GetRequiredService<VerifyGitHubVulnerabilitiesConfiguration>() ??
```
loic-sharma, line 31: You don't need the null check with `GetRequiredService`

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -36,7 +53,26 @@ public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wi
...
+            {
+                VerifyVulnerabilityInDatabase(vulnerability, withdrawn);
+            }
+
+            // Note: testing a withdrawn advisory isn't practical in registration metadata. We can only download
+            // metadata for a package, and would need to download all package/version blobs to determine an advisory
+            // is no longer present. Covering withdrawn advisory processing in the database will be adequate.
+            if (_configuration.VerifyRegistrationMetadata && !withdrawn)
+            {
+                return VerifyVulnerabilityInMetadata(vulnerability);
```
loic-sharma, line 66: I recommend `await`ing the task. See: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -113,15 +149,150 @@ public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wi
...
                             }, range {range.PackageId} {range.PackageVersionRange}, package {package.NormalizedVersion
                             } is not properly marked vulnerable to vulnerability!");
                         HasErrors = true;
                     }
                 }
             }
+        }
 
-            return Task.CompletedTask;
+        private Task VerifyVulnerabilityInMetadata(PackageVulnerability gitHubAdvisory)
```
loic-sharma, line 161: Nit: please add an `Async` suffix on methods that should be `await`ed.

```suggestion
        private Task VerifyVulnerabilityInMetadataAsync(PackageVulnerability gitHubAdvisory)
```

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -113,15 +149,150 @@ public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wi
...
+
+        private async Task InitializeMetadataResourceAsync()
+        {
+            var providers = Repository.Provider.GetCoreV3();
+            var packageSource = new PackageSource(_configuration.NuGetV3Index, "NuGet Source", isEnabled: true);
+            var sourceRepository = Repository.CreateSource(providers, packageSource, FeedType.Undefined);
+            _packageMetadataResource =
+                await sourceRepository.GetResourceAsync<PackageMetadataResource>(CancellationToken.None);
+
+            _packageMetadata = new Dictionary<string, IEnumerable<IPackageSearchMetadata>>();
```
loic-sharma, line 295: Registering these two fields so late means that we could run into weird logical bugs where we try to use the fields before they are initialized. I would consider the following changes:

1. In the constructor, create `_packageMetadata`
2. In the constructor, create a `Lazy<Task<PackageMetadataResource>> _packageMetadata` in the constructor. The Lazy's factory would then create the resource.
3. Wherever you need to use `_packageMetadata` you would do something like `var packageMetadata = await _packageMetadata.Value;`

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -113,15 +149,150 @@ public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wi
...
+                    ranges: rangeById.Value, 
+                    gitHubAdvisory.AdvisoryUrl, 
+                    gitHubAdvisory.GitHubDatabaseKey, 
+                    gitHubAdvisory.Severity));
+            }
+
+            return Task.WhenAll(verificationTasks);
+        }
+
+        private async Task VerifyVulnerabilityForRangeAsync (
```
loic-sharma, line 199:
```suggestion
        private async Task VerifyVulnerabilityForRangeAsync(
```

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -113,15 +149,150 @@ public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wi
...
+            if (!_packageMetadata.TryGetValue(package, out IEnumerable<IPackageSearchMetadata> metadata))
+            {
+                metadata = await _packageMetadataResource.GetMetadataAsync(
+                    package,
+                    includePrerelease: true,
+                    includeUnlisted: false,
+                    sourceCacheContext: new SourceCacheContext(),
+                    log: NuGet.Common.NullLogger.Instance,
+                    token: CancellationToken.None);
+                _packageMetadata[package] = metadata;
```
loic-sharma, line 281: I would consider calling `metadata.ToList();` before storing this value. You can't tell what `IEnumerable` is doing, it may be expensive stuffs every time you enumerate it. (I may be overly paranoid here. I caused a live site once because I repeatedly enumerated an `IEnumerable` that did reflection)

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -113,15 +149,150 @@ public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wi
...
+            {
+                verificationTasks.Add(VerifyVulnerabilityForRangeAsync(
+                    packageId: rangeById.Key, 
+                    ranges: rangeById.Value, 
+                    gitHubAdvisory.AdvisoryUrl, 
+                    gitHubAdvisory.GitHubDatabaseKey, 
+                    gitHubAdvisory.Severity));
+            }
+
+            return Task.WhenAll(verificationTasks);
```
loic-sharma, line 196: This will run your `VerifyVulnerabilityForRangeAsync` method concurrently, but I don't think `VerifyVulnerabilityForRangeAsync` is thread-safe (it mutates the `_packageMetadata` dictionary).

Also, is there a limit to how many tasks this will kick off?

loic-sharma:
Would it be possible to add unit tests?

drewgillies:
@loic-sharma this is an awesome review--thanks for all of this. As for the unit tests, it was a question playing on my mind--whether to test the test. It didn't have unit tests on its earlier incarnation because it in itself is a test. If it fails (which it did more than once during development) we can check its logic against the live data to determine which is at fault--thankfully it was the logic each time. I opted to not add unit tests because it's tantamount to writing a bunch of faux data which is nowhere as effective as the real stuff in exposing logic flaws.

src/VerifyGitHubVulnerabilities/Configuration/VerifyGitHubVulnerabilitiesConfiguration.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.ServiceModel.Configuration;
```
joelverhagen, line 5: Useless import?

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
drewgillies: Some of the other changes resulted in this method no longer being awaitable, but point taken--thanks.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
drewgillies: Great feedback. Implemented.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
drewgillies: Good call--I've done this.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
drewgillies: This was a really good catch--the number of tasks is data-driven, and I've now added an upper limit of 20, which seems like a nominal value for this kind of data.

drewgillies: Also, the method which manipulates the _packageMetadata cache is now thread-safe.

src/VerifyGitHubVulnerabilities/Job.cs
```diff
@@ -34,13 +34,15 @@ public override async Task Run()
...
 
             Console.WriteLine("Fetching vulnerabilities from DB...");
-            var verifier = new PackageVulnerabilitiesVerifier(_serviceProvider.GetRequiredService<IEntitiesContext>());
-            var ingestor = new AdvisoryIngestor(verifier, new GitHubVersionRangeParser());
+            var ingestor = _serviceProvider.GetRequiredService<IAdvisoryIngestor>();
             await ingestor.IngestAsync(advisories);
 
+            var verifier =
+                _serviceProvider.GetRequiredService<IPackageVulnerabilitiesManagementService>() as
+                    PackageVulnerabilitiesVerifier;
```
loic-sharma, line 42: Can/should we add `HasErrors` to `IPackageVulnerabilitiesManagementService` to avoid the cast here?

zhhyu, line 42: I am thinking that this will give you a new verifier rather than the one which should be built by the ingestor.
Maybe registering it as a singleton will work?

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
zhhyu: If we see the context, it seems that each thread only owns its corresponding specific package key. I am still learning in this area, but does it mean that the thread-safe is not a big concern here? Or using the concurrent dictionary will be better?

drewgillies: ConcurrentDictionary is a good thought here, but it does introduce some complexity I'd rather avoid given that `GetMetadataAsync` is `async`. Separating the logic out using a semaphore is a more elegant solution.

src/VerifyGitHubVulnerabilities/Job.cs
drewgillies: It was a little clumsy, yes (but wasn't a change in this PR). Rather than pollute `IPackageVulnerabilitiesManagementService`, I've extended it with `IPackageVulnerabilitiesVerifier` and implementing that in `PackageVulnerabilitiesVerifier`. It's now registered as both the new and the old interface, so the ingestion code can still pick it up.

drewgillies: Good catch, @zhhyu --thanks. It was indeed spinning up a new instance and not reporting a failed state. I'm now using a single instance registration.

zhhyu:
Looking now...

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -113,15 +151,147 @@ public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wi
...
+        }
 
-            return Task.CompletedTask;
+        private void VerifyVulnerabilityInMetadata(PackageVulnerability gitHubAdvisory)
+        {
+            Console.WriteLine($"[Metadata] Verifying vulnerability {gitHubAdvisory.GitHubDatabaseKey}.");
+
+            if (gitHubAdvisory.AffectedRanges == null || !gitHubAdvisory.AffectedRanges.Any())
+            {
+                return;// Task.CompletedTask;
```
zhhyu, line 169: Should we remove "Task.CompletedTask;" now? It's working in a sync way now.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -2,23 +2,42 @@
...
+    public class PackageVulnerabilitiesVerifier : IPackageVulnerabilitiesVerifier
     {
+        private readonly VerifyGitHubVulnerabilitiesConfiguration _configuration;
         private readonly IEntitiesContext _entitiesContext;
 
-        public PackageVulnerabilitiesVerifier(
+        private Lazy<Task<PackageMetadataResource>> _packageMetadataResource;
+        private Dictionary<string, IEnumerable<IPackageSearchMetadata>> _packageMetadata;
+
+        static readonly SemaphoreSlim semaphoreSlim = new SemaphoreSlim(1);
```
zhhyu, line 28: nit: -> "private static readonly SemaphoreSlim _semaphoreSlim"

src/VerifyGitHubVulnerabilities/Job.cs
```diff
@@ -68,6 +68,12 @@ protected void ConfigureIngestionServices(ContainerBuilder containerBuilder)
             containerBuilder
                 .RegisterType<AdvisoryIngestor>()
                 .As<IAdvisoryIngestor>();
+
+            containerBuilder
+                .RegisterType<PackageVulnerabilitiesVerifier>()
+                .As<IPackageVulnerabilitiesManagementService>()
```
zhhyu, line 74: I am not very sure, but do we need to register the interface "IPackageVulnerabilitiesManagementService" here?

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -113,15 +151,147 @@ public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wi
...
+                        Console.Error.WriteLine(
+                            $@"[Metadata] Vulnerability advisory {advisoryDatabaseKey
+                                }, version {versionMetadata} of package {packageId} is marked vulnerable and is not in a vulnerable range!");
+                        HasErrors = true;
+                    }
+                }
+            }
+        }
+
+        private async Task<IEnumerable<IPackageSearchMetadata>> GetPackageMetadataAsync(string package)
```
zhhyu, line 262: How about naming the argument as "packageId"?

src/VerifyGitHubVulnerabilities/Verify/IPackageVulnerabilitiesVerifier.cs
```diff
@@ -0,0 +1,20 @@
...
+using NuGet.Services.Entities;
+using NuGetGallery;
+
+namespace VerifyGitHubVulnerabilities.Verify
+{
+    /// <summary>
+    /// We need a <see cref=IPackageVulnerabilitiesManagementService" /> for inserting the verifier into the ingestion job,
+    /// so we extend for our extra reporting data.
+    /// </summary>
+    public interface IPackageVulnerabilitiesVerifier : IPackageVulnerabilitiesManagementService
```
zhhyu, line 13: I like this!

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -113,15 +151,147 @@ public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wi
...
+
+        private async Task VerifyVulnerabilityForRangeAsync(
+            string packageId,
+            IList<string> ranges,
+            string advisoryUrl, 
+            int advisoryDatabaseKey, 
+            PackageVulnerabilitySeverity advisorySeverity)
+        {
+            // Fetch metadata from registration blobs for verification--a collection of all versions of the package Id
+            var metadata = await GetPackageMetadataAsync(packageId);
```
zhhyu, line 205: If we see the context, it seems that each thread will be responsible for its corresponding package Id. So is there a need to introduce the dictionary here? We only get it as a local variable and process it at a specific thread, but never shared or reused it again.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
```diff
@@ -113,15 +151,147 @@ public Task UpdateVulnerabilityAsync(PackageVulnerability vulnerability, bool wi
...
+                    packageVersionRangeForId.Add(range.PackageVersionRange);
+                }
+                else
+                {
+                    rangesById[id] = new List<string> {range.PackageVersionRange};
+                }
+            }
+
+            Parallel.ForEach(rangesById, new ParallelOptions {MaxDegreeOfParallelism = 20}, async rangeById =>
+                await VerifyVulnerabilityForRangeAsync(
```
zhhyu, line 188: I have to point out here that it's an anti-pattern here. It works in a fire-and-forget way, and it will not wait until all tasks finish. So this will lead to the issue that we check "HasErrors" before all checks have been done. Using "Task.WhenAll" seems a good option.
See: [https://stackoverflow.com/questions/23137393/parallel-foreach-and-async-await](https://stackoverflow.com/questions/23137393/parallel-foreach-and-async-await)
Please ignore this if it's by design.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
drewgillies: This was unfinished cleanup. Thanks for spotting it.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
drewgillies: The code is called once for each vulnerability advisory, and there could be many per packageId. So it is performant to cache and reuse the metadata.

src/VerifyGitHubVulnerabilities/Verify/PackageVulnerabilitiesVerifier.cs
drewgillies: On reflection I agree. I'd introduced this approach due to its thread throttling, but given the high unlikelihood of a high number of threads here (given the nature of advisory data), I'm returning it to a simple task grouping model. Thanks.

drewgillies:
ping @loic-sharma @zhhyu @joelverhagen

zhhyu (approved):

PR #8423, opened 2021-02-17
workaround for narrator announce the status changes of copy button

lyndaidaii:

We currently are using bootstrap popover in nuget gallery, tooltips uses aria-describedby which is dynamically added once the tooltip is triggered 
in aria-describedby, it doesn't have aria attribute also it won't focus on the tooltip, what’s why Narrator would not read status. Besides that Narrator doesn’t read data content attribute as well.  For other screen readers like NVDA, they would read data content. That’s why our current fix in gallery work with NVDA, not Narrator.
 
This is common issue that existing in bootstrap community.(https://github.com/twbs/bootstrap/issues/28446)

This is workaround for narrator announce the staus of copy button to statisfy accessibility requirement

expected behavior with current changes: Narrator will announce that content is copied button is click at first time, it won't announce the status if click is made within 1500milisecond, it will continue to announce after 1500milisecond

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -97,16 +97,7 @@ $(function () {
             window.nuget.copyTextToClipboard(text, copyButton);
             copyButton.popover('show');
             //This is workaround for Narrator announce the status changes of copy button to achieve accessibility.
-            switch (copyButton.attr('aria-pressed')) {
-                case 'false':
-                    copyButton.attr('aria-pressed', 'true');
-                    break;
-                case 'true':
-                    copyButton.attr('aria-pressed', 'true');
-                    break;
-                default:
-                    copyButton.attr('aria-pressed', 'false');
-            }
+            copyButton.attr('aria-pressed', 'true');
             setTimeout(function () {
                 copyButton.popover('destroy');
             }, 1000);
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -96,9 +96,23 @@ $(function () {
             var text = $('#' + id + '-text').text().trim();
             window.nuget.copyTextToClipboard(text, copyButton);
             copyButton.popover('show');
+            //This is workaround for Narrator announce the status changes of copy button to achieve accessibility.
+            switch (copyButton.attr('aria-pressed')) {
+                case 'false':
+                    copyButton.attr('aria-pressed', 'true');
+                    break;
+                case 'true':
+                    copyButton.attr('aria-pressed', 'true');
+                    break;
+                default:
+                    copyButton.attr('aria-pressed', 'false');
+            }
             setTimeout(function () {
                 copyButton.popover('destroy');
             }, 1000);
+            setTimeout(function () {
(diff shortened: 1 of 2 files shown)


src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -96,9 +96,23 @@ $(function () {
             var text = $('#' + id + '-text').text().trim();
             window.nuget.copyTextToClipboard(text, copyButton);
             copyButton.popover('show');
+            //This is workaround for Narrator announce the status changes of copy button to achieve accessibility.
+            switch (copyButton.attr('aria-pressed')) {
+                case 'false':
+                    copyButton.attr('aria-pressed', 'true');
```
drewgillies, line 102: Should this be true or false? I'm presuming you have it correct here, but it might be good to explain the reason it looks odd--someone might be tempted to "fix" it.

lyndaidaii: default will be false, means no button is not pressed, true, means button is pressed

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -96,9 +96,23 @@ $(function () {
...
+                    break;
+                case 'true':
+                    copyButton.attr('aria-pressed', 'true');
+                    break;
+                default:
+                    copyButton.attr('aria-pressed', 'false');
```
loic-sharma, line 108: Just curious, what's this default case for? Would it make sense to always set the `aria-pressed` attribute to `true` and let the timer reset it back to `false` if somehow the attribute was an unexpected value?

Also, could you explain how you chose the 1500ms timer value?

loic-sharma (approved):

src/NuGetGallery/Scripts/gallery/page-display-package.js
lyndaidaii: default case actually is not necessary, i think we could delete it to make it clear. we have tooltip it shows the data content"copied" when button is pressed, which is 1000ms, in order to let narrator we read status completely. I think additional 500 should be enough here

src/NuGetGallery/Scripts/gallery/page-display-package.js
drewgillies: Got it--thanks. I'd suggest adding a comment to clear this up, but won't block the PR on this.

drewgillies (approved):

src/NuGetGallery/Scripts/gallery/page-display-package.js
drewgillies: Wait, so -- the default case isn't necessary, and both of the other cases do the same thing. If you delete the default case, you'll have a switch where the 2 cases do the same thing. Will you remove the switch?

lyndaidaii: I think leave the switch so that it will be clear from logic sense? what do you think?

loic-sharma: Yup @drewgillies's comment is what I was getting at. If we don't need that default case, can we replace the entire `switch` with just `copyButton.attr('aria-pressed', 'true');`.

drewgillies:
Looks good! :shipit:

PR #8431, opened 2021-02-26
Use new TFM logic for package ingestion in backfill command

drewgillies:
Please address https://github.com/NuGet/NuGetGallery/pull/8432 first (this PR depends on it).

Addresses: https://github.com/NuGet/Engineering/issues/3605

This adds a new backfill command to update TFM details in packages, using the improved logic in the abope PR's branch. Some notes on this:
- In order to get a file list efficiently from the flat container blob, I've used @joelverhagen's Knapcode.MiniZip utility to extract from the stream the file list--this circumvents the needs to download the whole nupkg.
- Backfill commands all extend `BackfillCommand`, and while the structure of the type is appropriate for use here, I needed to make it more flexible while not compromising other backfill jobs. For example, specifying includes and limits in query sets, and providing the above file list for metadata extraction when needed.
- The SQL connections used by backfill jobs were not using managed identity, and this is corrected here.

Once the other PR is merged and operational in Gallery code, I will run the backfill on the PROD database to bring old TFMs up to speed. then we can address UX changes still under discussion here: https://github.com/NuGet/Home/pull/10573

/cc @anangaur @jcjiang @nkolev92

### DIFF
--- a/src/GalleryTools/Commands/BackfillCommand.cs
+++ b/src/GalleryTools/Commands/BackfillCommand.cs
@@ -5,13 +5,15 @@ using Autofac;
 using CsvHelper;
 using CsvHelper.Configuration;
 using CsvHelper.TypeConversion;
+using Knapcode.MiniZip;
 using Microsoft.Extensions.CommandLineUtils;
 using NuGet.Packaging;
 using NuGet.Services.Entities;
 using NuGetGallery;
-using NuGetGallery.Configuration;
 using System;
+using System.Collections.Generic;
 using System.Data.Entity;
+using System.Data.SqlClient;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
@@ -20,6 +22,7 @@ using System.Threading.Tasks;
 using System.Xml;
 using System.Xml.Linq;
 using GalleryTools.Utils;
+using NuGet.Services.Sql;
 
 namespace GalleryTools.Commands
 {
@@ -34,6 +37,16 @@ namespace GalleryTools.Commands
         protected virtual int CollectBatchSize => 10;
 
         protected virtual int UpdateBatchSize => 100;
+
+        protected virtual int LimitTo => 0;
+
+        protected virtual MetadataSourceType SourceType => MetadataSourceType.Nuspec;
+
+        protected virtual bool UpdateNeedsContext => false;
+
+        protected virtual string QueryIncludes => null;
+
+        protected IPackageService _packageService;
 
         public static void Configure<TCommand>(CommandLineApplication config) where TCommand : BackfillCommand<TMetadata>, new()
         {
@@ -53,10 +66,12 @@ namespace GalleryTools.Commands
                 builder.RegisterAssemblyModules(typeof(DefaultDependenciesModule).Assembly);
                 var container = builder.Build();
 
-                var connectionString = container.Resolve<IAppConfiguration>().SqlConnectionString;
+                var sqlConnectionFactory = container.Resolve<ISqlConnectionFactory>();
+                var sqlConnection = await sqlConnectionFactory.CreateAsync();
                 var serviceDiscoveryUriValue = new Uri(serviceDiscoveryUri.Value());
 
                 var command = new TCommand();
+                command._packageService = container.Resolve<IPackageService>();
 
                 var metadataFileName = fileName.HasValue() ? fileName.Value() : command.MetadataFileName;
                 
@@ -75,24 +90,26 @@ namespace GalleryTools.Commands
                         }
                     }
 
-                    await command.Collect(connectionString, serviceDiscoveryUriValue, lastCreateTime, metadataFileName);
+                    await command.Collect(sqlConnection, serviceDiscoveryUriValue, lastCreateTime, metadataFileName);
                 }
 
                 if (updateDB.HasValue())
                 {
-                    await command.Update(connectionString, metadataFileName);
+                    await command.Update(sqlConnection, metadataFileName);
                 }
 
                 return 0;
             });
         }
 
-        public async Task Collect(string connectionString, Uri serviceDiscoveryUri, DateTime? lastCreateTime, string fileName)
+        public async Task Collect(SqlConnection connection, Uri serviceDiscoveryUri, DateTime? lastCreateTime, string fileName)
         {
-            using (var context = new EntitiesContext(connectionString, readOnly: true))
+            using (var context = new EntitiesContext(connection, readOnly: true))
             using (var cursor = new FileCursor(CursorFileName))
             using (var logger = new Logger(ErrorsFileName))
             {
+                context.SetCommandTimeout(300); // large query
+
                 var startTime = await cursor.Read();
 
                 logger.Log($"Starting metadata collection - Cursor time: {startTime:u}");
@@ -100,10 +117,19 @@ namespace GalleryTools.Commands
                 var repository = new EntityRepository<Package>(context);
 
                 var packages = repository.GetAll()
-                    .Include(p => p.PackageRegistration)
+                    .Include(p => p.PackageRegistration);
+                if (QueryIncludes != null)
+                {
+                    packages = packages.Include(QueryIncludes);
+                }
+                
+                packages = packages
                     .Where(p => p.Created < lastCreateTime && p.Created > startTime)
-                    .Where(p => p.PackageStatusKey == PackageStatus.Available || p.PackageStatusKey == PackageStatus.Validating)
-                    .OrderBy(p => p.Created);
+                    .OrderBy(p => p.PackageRegistration.Id);
+                if (LimitTo > 0)
+                {
+                    packages = packages.Take(LimitTo);
+                }
 
                 var flatContainerUri = await GetFlatContainerUri(serviceDiscoveryUri);
 
@@ -117,12 +143,32 @@ namespace GalleryTools.Commands
                     {
                         var id = package.PackageRegistration.Id;
                         var version = package.NormalizedVersion;
-
-                        var nuspecUri = $"{flatContainerUri}/{id.ToLowerInvariant()}/{version.ToLowerInvariant()}/{id.ToLowerInvariant()}.nuspec";
+                        var idLowered = id.ToLowerInvariant();
+                        var versionLowered = version.ToLowerInvariant();
 
                         try
                         {
-                            var metadata = await FetchMetadata(http, nuspecUri);
+                            var metadata = default(TMetadata);
+
+                            var nuspecUri =
+                                $"{flatContainerUri}/{idLowered}/{versionLowered}/{idLowered}.nuspec";
+                            using (var nuspecStream = await http.GetStreamAsync(nuspecUri))
+                            {
+                                var document = LoadDocument(nuspecStream);
+
+                                var nuspecReader = new NuspecReader(document);
+
+                                if (SourceType == MetadataSourceType.Nuspec)
+                                {
+                                    metadata = ReadMetadata(nuspecReader);
+                                }
+                                else if (SourceType == MetadataSourceType.Entities)
+                                {
+                                    var nupkgUri =
+                                        $"{flatContainerUri}/{idLowered}/{versionLowered}/{idLowered}.{versionLowered}.nupkg";
+                                    metadata = await FetchMetadataAsync(http, nupkgUri, nuspecReader);
+                                }
+                            }
 
                             if (ShouldWriteMetadata(metadata))
                             {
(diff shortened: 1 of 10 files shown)


src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -251,15 +313,26 @@ private static async Task<string> GetFlatContainerUri(Uri serviceDiscoveryUri)
...
+                var files = zipDirectory
+                    .Entries
+                    .Select(x => x.GetName())
+                    .ToList();
+
+                return ReadMetadata(files, nuspecReader);
+            }
+            catch (Exception)
+            {
+                return default; // fail silently without a value
```
joelverhagen, line 335: We should understand these errors instead of failing silently.

src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -460,5 +533,11 @@ public void Dispose()
                 Writer.Dispose();
             }
         }
+
+        public enum MetadataSourceType
+        {
+            Nuspec,
```
joelverhagen, line 539: Please comment this enum and values

src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -251,15 +313,26 @@ private static async Task<string> GetFlatContainerUri(Uri serviceDiscoveryUri)
...
+            var httpZipProvider = new HttpZipProvider(httpClient);
+            httpZipProvider.RequireAcceptRanges = false;
+            httpZipProvider.RequireContentRange = false;
 
-                var reader = new NuspecReader(document);
-
-                return ReadMetadata(reader);
+            try
+            {
+                var zipDirectoryReader = await httpZipProvider.GetReaderAsync(new Uri(nupkgUri));
```
joelverhagen, line 324: This will result in downloads. Options to avoid this:
1. Override the user agent so we don't get included in stats: https://github.com/joelverhagen/ExplorePackages/blob/425079a0be53aadfef84fd6222c102a0c062a288/src/ExplorePackages.Logic/ServiceCollectionExtensions.cs#L172-L183
2. Hit our blob container directly, bypassing the CDN stats pipeline. For this option we could have an option to the backfill command that allows you to override the flat container base URL (command line param instead of read from service index)

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                return supportedTFMs;
+            }
+
+            var supportedFrameworks = _packageService.GetSupportedFrameworks(nuspecReader, files);
+            foreach (var tfm in supportedFrameworks)
+            {
+                supportedTFMs += supportedTFMs == string.Empty
+                    ? tfm.GetShortFolderName()
+                    : $";{tfm.GetShortFolderName()}";
```
joelverhagen, line 43: use `StringBuilder` or even easier `string.Join` with `;`

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+                ? Enumerable.Empty<string>() 
+                : metadata.Split(';').OrderBy(f => f);
+
+            if (Enumerable.SequenceEqual(existingTFMs, newTFMs))
+            {
+                return; // nothing to change
+            }
+
+            // clean out the old (which will be left unattached in table otherwise) before adding new
+            foreach (var supportedFramework in package.SupportedFrameworks.ToList())
```
joelverhagen, line 72: nit: I don't think we need a `ToList` here, right?

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        protected override void UpdatePackage(EntitiesContext context, Package package, string metadata)
+        {
+            var existingTFMs = package.SupportedFrameworks == null
+                ? Enumerable.Empty<string>()
+                : package.SupportedFrameworks.Select(f => f.FrameworkName.GetShortFolderName()).OrderBy(f => f);
+
+            var newTFMs = string.IsNullOrEmpty(metadata)
+                ? Enumerable.Empty<string>() 
+                : metadata.Split(';').OrderBy(f => f);
```
joelverhagen, line 64: Might be worth noting that `;` does not appear in any frameworks so it is a safe delimiter ... right now (which means it's fine for backfill).

To be defensive we should throw an exception in `ReadMetadata` if a framework contains a `;`. Or we can refactor to allow metadata to be a List as well but that seems like too much work for a backfill job.

src/GalleryTools/GalleryTools.csproj
```diff
@@ -84,6 +85,9 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Knapcode.MiniZip">
+      <Version>0.15.0</Version>
```
joelverhagen, line 89: There's a newer version that gives better error message if the HTTP requests fail.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    public sealed class BackfillTfmMetadataCommand : BackfillCommand<string>
+    {
+        protected override string MetadataFileName => "tfmMetadata.txt";
+        
+        protected override MetadataSourceType SourceType => MetadataSourceType.Entities;
+        
+        protected override string QueryIncludes => $"{nameof(Package.SupportedFrameworks)}";
+
+        protected override int LimitTo => 100000;
```
joelverhagen, line 21: This seems... pretty big. Might be worth taking a lower value due to the relatively high cost per package. No need to put so much in memory here or on SQL side.

joelverhagen: 5000 is my go-to for no reason at all.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
joelverhagen: Oh wait is see `BackfillCommand<string>`. Could we make this `BackfillCommand<List<string>>`  or `BackfillCommand<List<NuGetFramework>>`? It would save this strange string encoding stuff.

joelverhagen:
Just curious - any projects on how long it will take to run on NuGet.org?

drewgillies:
> Just curious - any projects on how long it will take to run on NuGet.org?
None at this stage. I will discuss scale with you when your week starts.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
drewgillies: Actually, I'll remove this. This is an artifact from creating a test subset and should be 0 or gone--it's not a batching metric. But it does raise the question of the Collect step creating a csv file covering every package/version in NuGet.org. That will be large.

src/GalleryTools/Commands/BackfillCommand.cs
drewgillies: I also updated the names, as my strategy for metadata extraction changed midstream from entities to blobs.

src/GalleryTools/Commands/BackfillCommand.cs
drewgillies: I'll reach out to you directly on this.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
drewgillies: We do. We can't enumerate the collection while we're changing it, so I need to `ToList` it to give the loop an unchanging value.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
drewgillies: This is much better. Thanks.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
drewgillies: `TMetadata` being `List<string>` (your suggestion below) removes this issue.

src/GalleryTools/Commands/BackfillCommand.cs
drewgillies: Logging now.

src/GalleryTools/GalleryTools.csproj
drewgillies: Now using `0.16.0`.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
joelverhagen: Missed that. Good point.

src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -35,6 +38,16 @@ public abstract class BackfillCommand<TMetadata>
 
         protected virtual int UpdateBatchSize => 100;
 
+        protected virtual int LimitTo => 0;
+
+        protected virtual MetadataSourceType SourceType => MetadataSourceType.NuspecOnly;
+
+        protected virtual bool UpdateNeedsContext => false;
+
+        protected virtual string QueryIncludes => null;
```
joelverhagen, line 47: I think this would be more extensible as a `Func<IQueryable<...>, IQueryable<...>>`. This would allow implementing classes to do whatever they want to the query.

src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -199,8 +245,14 @@ public async Task Update(string connectionString, string fileName)
 
                             if (package != null)
                             {
-
-                                UpdatePackage(package, metadata.Metadata);
+                                if (UpdateNeedsContext)
```
joelverhagen, line 248: I'm curious if these is really a helpful option. Wouldn't it be simpler to just always use the three parameter overload and implementations can choose to ignore the `context` parameter if they don't need it?

src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -251,15 +313,28 @@ private static async Task<string> GetFlatContainerUri(Uri serviceDiscoveryUri)
...
 
-        private async Task<TMetadata> FetchMetadata(HttpClient httpClient, string nuspecUri)
+        private async Task<TMetadata> FetchMetadataAsync(
+            HttpClient httpClient, string nupkgUri, NuspecReader nuspecReader, string id, string version, Logger logger)
         {
-            using (var nuspecStream = await httpClient.GetStreamAsync(nuspecUri))
-            {
-                var document = LoadDocument(nuspecStream);
+            var httpZipProvider = new HttpZipProvider(httpClient);
+            httpZipProvider.RequireAcceptRanges = false;
```
joelverhagen, line 320: I'm curious why `RequireAcceptRanges` and `RequireContentRange` are set. I set them in my Blazor app because that was internally using AJAX which is subjected to CORS header restrictions which today do not include these two referenced headers. In a normal .NET app it is better to require these headers since they indicate the underlying Range requests are working as expected.

src/GalleryTools/Commands/BackfillCommand.cs
joelverhagen: Additionally this would allow the `.Include(x => x.SupportedFrameworks)` overload which has better compile time guarantees over the string based one we are using right now.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public static void Configure(CommandLineApplication config)
+        {
+            Configure<BackfillTfmMetadataCommand>(config);
+        }
+
+        protected override List<string> ReadMetadata(IList<string> files, NuspecReader nuspecReader)
+        {
+            var supportedTFMs = new List<string>();
+            if (_packageService == null)
```
joelverhagen, line 31: Why would `_packageService` be null ever?

src/GalleryTools/Commands/BackfillCommand.cs
drewgillies: `Include` is fluent fun, so I've needed to go with `Expression<Func<Package, object>>`. However, this doesn't allow an override as separate includes requires separate `Include`s.

src/GalleryTools/Commands/BackfillCommand.cs
drewgillies: Removed and extended existing signature.

src/GalleryTools/Commands/BackfillCommand.cs
drewgillies: Good to know--thanks. Removed.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
drewgillies: Hypervigilance on null checks. Removed.

drewgillies:
ping @joelverhagen :)

joelverhagen (approved):

Issue #8386, opened 2021-01-25
Replace TFM logic in Gallery package service


drewgillies:
Replace the logic that populates this field with a nupkg lib folders-based approach. This will give us a mostly-accurate quick win on displaying TFMs supported by a package.

PR #8432, opened 2021-02-26
New TFM logic for package ingestion

drewgillies:
Addresses: https://github.com/NuGet/NuGetGallery/issues/8386

The existing logic in `PackageService` relies on `package.GetSupportedFrameworks`, which while giving (largely) correct data, doesn't give enough of it--~85% with this change increasing it to ~93%. The new method in `PackageService` utilizes the restore logic employed by restore, gathering it into a single location for easy and robust maintenance. This logic will give us the supported frameworks for packages whose anatomy follows our guidelines, e.g. https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks. Some notes on this:
- Pattern set matching is our heuristic here--the only time we use the nuspec (which has misled us in the prior approach) is to determine package type, specifically answering one question: is this a _tools_ package and nothing else? If it is a tools package _only_, we will look to match tools pattern sets _only_ in order to list supported TFMs. Other pattern sets don't exist. This matches client restore behavior.
- Packages of type _build_ will only match build pattern sets when a `{id}.props|targets` file exists in a directory named for the TFM (in an appropriate pattern sets for build)
- See unit tests for cases - I've tried to cover as much as is reasonable there.

I've split the file list from the nuspec reader in this method, so that it can be called by existing logic in the `PackageService`, as well as by the new backfill job `BackfillTfmMetadataCommand`. (see https://github.com/NuGet/NuGetGallery/pull/8431, which should be merged _after_ this PR).

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/IPackageService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/IPackageService.cs
@@ -1,9 +1,11 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using NuGet.Frameworks;
 using NuGet.Packaging;
 using NuGet.Services.Entities;
 using NuGet.Versioning;
@@ -111,6 +113,8 @@ namespace NuGetGallery
         Task<Package> CreatePackageAsync(PackageArchiveReader nugetPackage, PackageStreamMetadata packageStreamMetadata, User owner, User currentUser, bool isVerified);
 
         Package EnrichPackageFromNuGetPackage(Package package, PackageArchiveReader packageArchive, PackageMetadata packageMetadata, PackageStreamMetadata packageStreamMetadata, User user);
+
+        IEnumerable<NuGetFramework> GetSupportedFrameworks(NuspecReader nuspecReader, IList<string> packageFiles);
 
         Task PublishPackageAsync(string id, string version, bool commitChanges = true);
         Task PublishPackageAsync(Package package, bool commitChanges = true);
--- a/src/NuGetGallery.Services/PackageManagement/PackageService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageService.cs
@@ -708,8 +708,8 @@ namespace NuGetGallery
         /// <summary>
         /// This method combines the logic used in restore operations to make a determination about the TFM supported by the package.
         /// We have curated a set of compatibility requirements for our needs in NuGet.org. The client logic can be found here:
-        /// https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs#L252-L291
-        /// https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs#L439-L442
+        /// https://github.com/NuGet/NuGet.Client/blob/63255047fe7052cc33b763356ff995d9166f719e/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs#L252-L294
+        /// https://github.com/NuGet/NuGet.Client/blob/63255047fe7052cc33b763356ff995d9166f719e/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs#L439-L442
         /// ...and our combination of these elements is below.
         /// The logic is essentially this:
         /// - Determine whether we're looking at a tools package. In this case we will use tools "pattern sets" (collections of file patterns
--- a/src/NuGetGallery.Services/PackageManagement/PackageService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageService.cs
@@ -8,14 +8,18 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using NuGet.Client;
+using NuGet.ContentModel;
 using NuGet.Frameworks;
 using NuGet.Packaging;
 using NuGet.Packaging.Core;
+using NuGet.RuntimeModel;
 using NuGet.Services.Entities;
 using NuGet.Versioning;
 using NuGetGallery.Auditing;
 using NuGetGallery.Packaging;
 using NuGetGallery.Security;
+using PackageType = NuGet.Packaging.Core.PackageType;
 
 namespace NuGetGallery
 {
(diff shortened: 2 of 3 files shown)


src/NuGetGallery.Services/PackageManagement/IPackageService.cs
```diff
@@ -112,6 +114,8 @@ public interface IPackageService : ICorePackageService
 
         Package EnrichPackageFromNuGetPackage(Package package, PackageArchiveReader packageArchive, PackageMetadata packageMetadata, PackageStreamMetadata packageStreamMetadata, User user);
 
+        IEnumerable<NuGetFramework> GetSupportedFrameworks(NuspecReader nuspecReader, IList<string> packageFiles);
```
loic-sharma, line 117: What do you think of exposing the other `GetSupportedFrameworks` method signature instead? It seems a little bit more convenient for the caller.

```cs
public virtual IEnumerable<NuGetFramework> GetSupportedFrameworks(PackageArchiveReader package)
```

src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -698,9 +702,95 @@ private Package CreatePackageFromNuGetPackage(PackageRegistration packageRegistr
...
+                    }
+
+                    supportedTFMs = groups
+                        .SelectMany(p => p.Properties)
+                        .Where(pair => pair.Key == ManagedCodeConventions.PropertyNames.TargetFrameworkMoniker)
+                        .Select(pair => pair.Value)
+                        .Cast<NuGetFramework>()
+                        .Distinct();
+                }
+                catch (Exception)
```
loic-sharma, line 763: When would an exception be thrown? Should we log this? Consider explaining in the comment why the exception is ignored.

src/NuGetGallery.Services/PackageManagement/IPackageService.cs
loic-sharma: If you go with this change, can we make the `GetSupportedFrameworks` that accepts the `packageFiles` argument private?

src/NuGetGallery.Services/PackageManagement/PackageService.cs
joelverhagen: Agreed. If we can't analyze the assets of a package, it's quite likely that restore won't work on it either. And perhaps it should be rejected.

src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -698,9 +702,95 @@ private Package CreatePackageFromNuGetPackage(PackageRegistration packageRegistr
             return package;
         }
 
-        public virtual IEnumerable<NuGetFramework> GetSupportedFrameworks(PackageArchiveReader package)
+        public virtual IEnumerable<NuGetFramework> GetSupportedFrameworks(PackageArchiveReader package) =>
+            GetSupportedFrameworks(package.NuspecReader, package.GetFiles().ToList());
+
+        public virtual IEnumerable<NuGetFramework> GetSupportedFrameworks(NuspecReader nuspecReader, IList<string> packageFiles)
```
joelverhagen, line 708: Consider linking to the responding line(s) of code in NuGet/NuGet.Client.

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -970,6 +970,207 @@ private async Task WillNotSaveAnySupportedFrameworksWhenThereIsAnAnyTargetFramew
...
+            }
+
+            /// <summary>
+            /// These cases use the guidance laid out here:
+            /// https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package
+            /// https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks
+            /// https://docs.microsoft.com/en-us/nuget/reference/target-frameworks
+            /// https://docs.microsoft.com/en-us/dotnet/standard/frameworks
+            /// </summary>
+            public static IEnumerable<object[]> TargetFrameworkCases =>
```
joelverhagen, line 1015: Nice tests.

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
```diff
@@ -970,6 +970,207 @@ private async Task WillNotSaveAnySupportedFrameworksWhenThereIsAnAnyTargetFramew
...
+                            "runtimes/win10-x86/lib/net40/_._",
+                            "runtimes/win10-x86/lib/net471/_._",
+                            "ref/net5.0-watchos/_1._",
+                            "ref/net5.0-watchos/_2._",
+                            "build/netstandard21/Foo.props",
+                            "build/netstandard20/Foo.targets",
+                            "tools/netcoreapp3.1/win10-x86/tool1/_._",
+                            "tools/netcoreapp3.1/win10-x86/tool2/_._"
+                        }, 
+                        new List<string>{"net40", "net471", "net5.0-watchos", "netstandard2.0", "netstandard2.1"}
```
joelverhagen, line 1110: for simple compat purposes, `net471` could be implied by `net40`. Should we do some reduction so that the output is more terse? This is essentially a UI question since I don't know how the UI is supposed to look in this case.

src/NuGetGallery.Services/PackageManagement/PackageService.cs
skofman1, line 708: please add a lot of comments about the algorithm in header

src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -698,9 +702,95 @@ private Package CreatePackageFromNuGetPackage(PackageRegistration packageRegistr
...
         {
-            return package.GetSupportedFrameworks();
+            var supportedTFMs = Enumerable.Empty<NuGetFramework>();
```
loic-sharma, line 710: Are we concerned about this algorithm's performance? If it may be expensive, we should consider adding a duration metric here

src/NuGetGallery.Services/PackageManagement/IPackageService.cs
drewgillies: We need the new method to be public for performance in the backfill job--the backfill job fetches the file list separately (and we get this quickly using MiniZip) and doesn't provide an entire package archive like the package service does. So, this method is the integration point. We could make the old method private but that will exclude it from its tests, and I'd like to keep those old tests alive as well ad my new ones, as they test the surface area used by the service. I could make it internal and ensure `InternalsVisibleTo` is appropriately connecting, if you feel we need to remove it from public access.

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
drewgillies: It's a good question and I'd need to look at the data to see if this case even exists. I was randomly choosing net versions here to ensure the pattern matching works. If it's a non-case, I'll change it to something realistic. If it really occurs, I'd be interested in PM thoughts here. Of course, there's the corner case where we have net40 in lib nodes and net471 in content or build nodes, and we may wish to list separately there.

src/NuGetGallery.Services/PackageManagement/PackageService.cs
drewgillies: I'm happy to do that if it's necessary, although the logic itself is simply swapping some rudimentary pattern matching with something that includes more patterns. There are no loops here (except enumerations of constant size) so it's still O(n)--I can't imagine it adding much more processor usage.

src/NuGetGallery.Services/PackageManagement/PackageService.cs
drewgillies: I've revised this logic--it had exception handling as a carry-over from when I was using it for package analysis and we had no nupkg for packages in test dbs. There's no need for it here.

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
drewgillies: I'd be inclined to capture the supported TFMs as-is at this stage of the package's life cycle though. If we wanted terseness, we could determine that at display time. If display requirements were to change later, we still have the data that best represents the package's anatomy if we don't abridge here. Thoughts?

src/NuGetGallery.Services/PackageManagement/PackageService.cs
drewgillies: Done.

drewgillies: I've done this as well as put comments throughout the algorithm.

joelverhagen: Consider doing a "permalink" for the GitHub URLs that include the commit hash. This will mitigate line number changes in the future. You can switch to the commit-specific URL of any file in the UI by pressing "Y".

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs
becomes
https://github.com/NuGet/NuGet.Client/blob/01c198fafe132c664a995f21852300639fc034d4/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs

Or use:
image

src/NuGetGallery.Services/PackageManagement/PackageService.cs
joelverhagen: I don't anticipate a problem given the relatively low rate of package pushes and autoscale being on the resource that would likely be affected here (CPU). If we want to be cautious we can feature flag it in the calling service when it is more integrated.

tests/NuGetGallery.Facts/Services/PackageServiceFacts.cs
joelverhagen: Yes, I agree. This can be a display time consideration in the near term, if necessary. Work case scenario we just do another backfill later with a more terse version.

joelverhagen (approved):

src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -698,9 +702,107 @@ private Package CreatePackageFromNuGetPackage(PackageRegistration packageRegistr
...
         }
 
-        public virtual IEnumerable<NuGetFramework> GetSupportedFrameworks(PackageArchiveReader package)
+        public virtual IEnumerable<NuGetFramework> GetSupportedFrameworks(PackageArchiveReader package) =>
+            GetSupportedFrameworks(package.NuspecReader, package.GetFiles().ToList());
+
+        /// <summary>
+        /// This method combines the logic used in restore operations to make a determination about the TFM supported by the package.
+        /// We have curated a set of compatibility requirements for our needs in NuGet.org. The client logic can be found here:
+        /// https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs#L252-L291
```
agr, line 711: The `dev` branch links with line numbers will likely become invalid over time (at the very least, lines will move, at worst the file can be moved/deleted). Links to specific git snapshot are much more likely to stay valid (unless whole repo is destroyed).

src/NuGetGallery.Services/PackageManagement/PackageService.cs
agr: Current link point to `dev` branch, consider linking to specific snapshots, otherwise over time there is a good chance those links would become invalid as `dev` contents changes.

agr (approved):

src/NuGetGallery.Services/PackageManagement/PackageService.cs
drewgillies: Thanks--done.

Issue #8373, opened 2021-01-11
Add instructions to install .NET Tools as local


augustoproiete:
The current NuGet Gallery UI for .NET Tools shows instructions on how to install them as global tools, but doesn't have instructions for installing as a local tool.

![image](https://user-images.githubusercontent.com/177608/104142779-5d90ca00-5393-11eb-8113-10cf8bb1fa47.png)

---

### Proposed UI

I suggest adding a new tab for instructions on how to install it as a local tool, including creating the tool manifest, as that's currently required.

![2021-01-10_22-26-15 (1)](https://user-images.githubusercontent.com/177608/104142789-67b2c880-5393-11eb-9c55-bfd58d8526bd.gif)

- Rename the existing ".NET CLI" tab to ".NET CLI (Global)"
- Update the info text at the bottom to `This package contains a .NET Tool you can call from the shell/command line.`
- Add new tab ".NET CLI (Local)" with the following instructions:
```powershell
> dotnet new tool-manifest
> dotnet tool install --local <ToolName> --version <ToolVersion>
```

skofman1:
@augustoproiete , love your proposal! Would you be willing to work with us to on getting a polished UX spec and contributing a PR?
//cc @jcjiang

augustoproiete:
@skofman1 Absolutely. Let me know next steps.

joelverhagen:
@nkolev92 - do you know who on the .NET CLI side would have an opinion here?

joelverhagen:
@augustoproiete, let's wait for feedback from folks that own the .NET CLI tool experience. Once we get some consensus, I'd love to review a PR you make 😄.

wli3:
@KathleenDollard may have more opinion on this. Stating `dotnet new tool-manifest` all the time might be an issue. If there is an existing tool manifest in user's desired directory or directory above,  it is not required.

augustoproiete:
@wli3 Related/for reference, I also suggested that `dotnet tool install` could create the manifest automatically for the user (if one does not exist) which I think is a better experience: https://github.com/dotnet/sdk/issues/15254

KathleenDollard:
@wli3 Do we ignore further calls to `dotnet new tool-manifest`? 

If not, should we?

If so, I like this addition. It will be the right place for people to start understanding the difference between global and local tools (they are considering installing one at that time) and they might think "oh, what's a tool-manifest". 

[...]

PR #8433, opened 2021-02-28
Add dotnet local tool instructions tab

augustoproiete:
Implementation of the tab with installation instructions for .NET local tools:

- Rename the existing `.NET CLI` tab to `.NET CLI (Global)`

- Update the info text at the bottom:

```diff
- This package contains a .NET Core Global Tool you can call from the shell/command line
+ This package contains a .NET tool you can call from the shell/command line
```

- Add a new tab `.NET CLI (Local)` with the following instructions:

```
dotnet new tool-manifest # if you are setting up this repo
dotnet tool install --local <ToolName> --version <ToolVersion>
```

---

image

---

image

---

Addresses https://github.com/NuGet/NuGetGallery/issues/8373

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -28,13 +28,13 @@
     {
         packageManagers = new PackageManagerViewModel[]
         {
-            new PackageManagerViewModel(".NET CLI")
+            new PackageManagerViewModel(".NET CLI (Global)")
             {
                 Id = "dotnet-cli",
                 CommandPrefix = "> ",
                 InstallPackageCommand = string.Format("dotnet tool install --global {0} --version {1}", Model.Id, Model.Version),
                 AlertLevel = AlertLevel.Info,
-                AlertMessage = "This package contains a <a href='https://aka.ms/global-tools'>.NET Core Global Tool</a> you can call from the shell/command line.",
+                AlertMessage = "This package contains a <a href='https://aka.ms/global-tools'>.NET tool</a> you can call from the shell/command line.",
             },
 
             new ThirdPartyPackageManagerViewModel("Cake", "https://cakebuild.net/support/nuget")
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -30,9 +30,22 @@
         {
             new PackageManagerViewModel(".NET CLI (Global)")
             {
-                Id = "dotnet-cli",
+                Id = "dotnet-cli-global",
                 CommandPrefix = "> ",
                 InstallPackageCommand = string.Format("dotnet tool install --global {0} --version {1}", Model.Id, Model.Version),
+                AlertLevel = AlertLevel.Info,
+                AlertMessage = "This package contains a <a href='https://aka.ms/global-tools'>.NET tool</a> you can call from the shell/command line.",
+            },
+
+            new PackageManagerViewModel(".NET CLI (Local)")
+            {
+                Id = "dotnet-cli-local",
+                CommandPrefix = "> ",
+                InstallPackageCommand = string.Join
+                    (Environment.NewLine,
+                        "dotnet new tool-manifest # if you are setting up this repo",
+                        string.Format("  dotnet tool install --local {0} --version {1}", Model.Id, Model.Version)
+                    ),
                 AlertLevel = AlertLevel.Info,
                 AlertMessage = "This package contains a <a href='https://aka.ms/global-tools'>.NET tool</a> you can call from the shell/command line.",
             },
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -181,9 +181,12 @@
     <div role="tabpanel" class="tab-pane @(active ? "active" : string.Empty)" id="@packageManager.Id">
         <div>
             <div class="install-script-row">
+                @{
+                    var lastIndex = packageManager.InstallPackageCommands.Length - 1;
+                    var cs = packageManager.InstallPackageCommands.Select((c, i) => i < lastIndex ? c + Environment.NewLine : c);
+                }
                 @* Writing out the install command must be on a single line to avoid undesired whitespace in the <pre> tag. *@
-                @{ var cs = packageManager.InstallPackageCommands; }
-                <pre class="install-script" id="@packageManager.Id-text">@foreach (var c in cs) {<span class="install-command-row">@(c + Environment.NewLine)</span>}</pre>
+                <pre class="install-script" id="@packageManager.Id-text">@foreach (var c in cs) {<span class="install-command-row">@c</span>}</pre>
                 <div class="copy-button">
                     <!--In order to statisfy the requirement to announce button status both on NVDA/Narrator, other screen reader like NVDA will
                         announce the data-content "Copied" everytime when we press button, however, it won't work with Narrator when we use bootstrap popover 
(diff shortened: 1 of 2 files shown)


joelverhagen:
I think changing the model to a list of strings (one string per command line) is a good one assuming the copy button still copies the expected text. It's important that the copied text is actually two lines with nothing else added, so that pasting into a terminal behaves nicely.

Also, when you add multi-command support, please check what happens with a longer package ID and a narrow screen. Right now the text wrapping looks "good enough" for a single command. We'd want it to look similarly "good enough" 😄 for a command text.

joelverhagen:
Hey @augustoproiete, is something I can help out with here or clarify? This seems like a good change and I'd be happy to review the next iteration per my previous comment (https://github.com/NuGet/NuGetGallery/pull/8433#issuecomment-788213063) whenever you're ready 😃.

augustoproiete:
Hey @joelverhagen, thanks for the nudge. I think it's all clear and the ball is in my court now to try to implement the multi-line ​commands whilst preserving the current copy/paste behavior - though any ideas are welcome.

I'll set some time aside this coming weekend and see if I can complete this one.

augustoproiete:
@joelverhagen This is ready for round 2.

I've updated the view model to allow more than one install command and I'm rendering each command inside of a `span` in order to add the command prefix to each individual line.

The copy/paste continues to copy the expected text in all cases, preserving line-breaks (if any).

I tested all tabs (Package Manager, .NET CLI, PackageReference, Paket CLI, F# Interactive, Cake, and .NET CLI for Templates), and they continue to display the install commands as expected.

---

### Single command with one line

image

```diff
- dotnet tool install --global dotnet-ef --version 5.0.4
+ <span class="install-command-row">dotnet tool install --global dotnet-ef --version 5.0.4</span>
```

---

### Multiple commands :new:

image

```diff
+ <span class="install-command-row">dotnet new tool-manifest # if you are setting up this repo
+ </span><span class="install-command-row">dotnet tool install --local dotnet-ef --version 5.0.4</span>
```

---

### Single command with multiple lines (e.g. PackageReference for Development dependencies)

image

```diff
- &lt;PackageReference Include=&quot;GitVersion.MsBuild&quot; Version=&quot;5.6.6&quot;&gt;
-   &lt;PrivateAssets&gt;all&lt;/PrivateAssets&gt;
-   &lt;IncludeAssets&gt;runtime; build; native; contentfiles; analyzers&lt;/IncludeAssets&gt;
- &lt;/PackageReference&gt
+ <span class="install-command-row">&lt;PackageReference Include=&quot;GitVersion.MsBuild&quot; Version=&quot;5.6.6&quot;&gt;
+   &lt;PrivateAssets&gt;all&lt;/PrivateAssets&gt;
+   &lt;IncludeAssets&gt;runtime; build; native; contentfiles; analyzers&lt;/IncludeAssets&gt;
+ &lt;/PackageReference&gt;</span>
```

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -166,10 +179,16 @@
...
+
+    var htmlInstallPackageCommandSeparator = string.Format("{0}</span><span class=\"install-command-row\">", Environment.NewLine);
+    var htmlInstallPackageCommands = "<span class=\"install-command-row\">" +
+        string.Join(htmlInstallPackageCommandSeparator, encodedInstallPackageCommands) + "</span>";
+
     <div role="tabpanel" class="tab-pane @(active ? "active" : string.Empty)" id="@packageManager.Id">
         <div>
             <div class="install-script-row">
-                <pre class="install-script" id="@packageManager.Id-text">@packageManager.InstallPackageCommand</pre>
+                <pre class="install-script" id="@packageManager.Id-text">@Html.Raw(htmlInstallPackageCommands)</pre>
```
joelverhagen, line 191: In general, we try to avoid `@Html.Raw` as much as possible since it can be a door for XSS. Could we try a `foreach` and emit the `<span>` in actual Razor rather than in .NET strings?

joelverhagen: I see the `Html.Encode` -- but if we can avoid `@Html.Raw` it will be more obviously secure into the future as other things change.

augustoproiete: We definitely could, but it comes with a price (_which is why I didn't go that route_): Everything must be on a single line so that no extra whitespace is introduced in the expression when it's rendered, because we're writing content inside of the `pre` tag which is sensitive to whitespace.

We would end up with at least more one single-line helper method similar to [`MakeLicenseLink`](https://github.com/NuGet/NuGetGallery/blob/bd9271e82877b18e54ac1fe967b3bb3a9e568310/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml#L219) but prob. longer in length/horizontal scrolling.

Still think it's a better trade-off?

If not, we could add a Razor comment just above the `pre` tag explaining that the `@Html.Raw` is made secure by the `@Html.Encode` above.

joelverhagen: And then I went on a code golf adventure 😅. We've had our own pain with `Html.Raw` so maybe we're seeing it from different perspectives. I think a long line is preferable. From a security perspective, I would take a simpler code audit over horizontal scrolling.

What do you think about something like this? https://github.com/NuGet/NuGetGallery/commit/52a7b713ea159a69f9d623373fb6e119a6771602

It has an additional trailing CR LF but from what I could tell it does not affect the display or copied text, perhaps due to the `trim()` here: https://github.com/NuGet/NuGetGallery/blob/25d2d3b32b2d9f0b1ca6e0a105b0210c2c4820f4/src/NuGetGallery/Scripts/gallery/page-display-package.js#L96

augustoproiete: :smile: Works for me. I noticed the JS `trim()` before so my concern was only with the trailing CRLF affecting the display but, indeed, I tested [`52a7b71`](https://github.com/NuGet/NuGetGallery/commit/52a7b713ea159a69f9d623373fb6e119a6771602) and didn't notice any visual difference, so LGTM.

I squashed `52a7b71` into the PR if you're keen to merge as-is.

Just to have a comparison, I tried an alternative approach (admittedly more verbose) to ensure there's no trailing CRLF: [`82eac2f`](https://github.com/augustoproiete-forks/NuGet--NuGetGallery/commit/82eac2fb1256afad60b2e91558e7e80aad244243), which is not terrible... :thinking:, so happy to go either way.

joelverhagen: Your https://github.com/augustoproiete-forks/NuGet--NuGetGallery/commit/82eac2fb1256afad60b2e91558e7e80aad244243 is even better! Nice work!

augustoproiete: Great. Do you need me to update this PR to bring https://github.com/augustoproiete-forks/NuGet--NuGetGallery/commit/82eac2fb1256afad60b2e91558e7e80aad244243?

joelverhagen: Sure, that'd be great. After I see that show up I'll run it through the CI and get this thing merged.

augustoproiete: Great. Give me 5 minutes :)

augustoproiete: PR updated with https://github.com/NuGet/NuGetGallery/pull/8433/commits/f0972688425565fd817237741ea69f230cd93d29

joelverhagen (approved):

loic-sharma (approved):

joelverhagen:
Thanks @augustoproiete!

Issue #8381, opened 2021-01-20
Add Cake instructions for NuGet packages


augustoproiete:
## Background

**[Cake](https://cakebuild.net)** (C# Make) is a free and open source cross-platform build automation system with a C# DSL for tasks such as compiling code, copying files and folders, running unit tests, compressing files and building NuGet packages.

[![image](https://user-images.githubusercontent.com/177608/105109451-f067ed80-5a92-11eb-8dff-a6978c3d9205.png)](https://cakebuild.net/)

Cake is widely used by thousands of [open-source projects](https://github.com/cake-build/cake/network/dependents?package_id=UGFja2FnZS0xNTY3NDU3NzY%3D), is the [third most downloaded .NET tool](https://www.nuget.org/packages?packagetype=dotnettool&sortby=relevance&q=&prerel=True) at the moment with over 4 million downloads, has a [healthy community of contributors](https://cakebuild.net/community/thanks/), and is a [member of the .NET Foundation](https://dotnetfoundation.org/projects/cake).

One of the features of Cake is the ability to consume NuGet packages from any NuGet compatible feed, and leverage them in the build orchestration process. For example, within a Cake build, a user can reference the [Polly](https://www.nuget.org/packages/Polly/) NuGet package, and leverage all of its features, defining and executing different Polly policies.

NuGet packages can be used in Cake in a variety of ways:

- As a regular NuGet package (Package Reference, packages.config, NuGet Package Manager UI, etc.) on a [Console application](https://cakebuild.net/docs/running-builds/runners/cake-frosting).

- As a **[Cake Addin](https://cakebuild.net/docs/extending/addins/)** (`#addin`). In short, this means: Download the NuGet package, and load the assemblies within it into the host process executing the build process. Virtually any NuGet package containing .NET assembly(ies) can be loaded as a Cake Addin. e.g. `#addin nuget:?package=Polly&version=7.2.1`

- As a **[Cake Tool](https://cakebuild.net/docs/writing-builds/tools/installing-tools)** (`#tool`). In short, this means: Download the NuGet package, and extract its contents to a folder, so that its files become available to the build process. Virtually any NuGet package can be loaded as a Cake Tool. e.g. `#tool nuget:?package=7-Zip.CommandLine&version=18.1.0`

[...]

PR #8434, opened 2021-02-28
Implementation of Cake tab with instructions for NuGet packages

augustoproiete:
Add installation instructions of NuGet packages in Cake scripts.

Credits to Nils Andresen ( @nils-a ), member of the Cake team, who did all the hard work :muscle:

| :jigsaw: | Packages identified as Cake extension of type "Addin" (via tags) |
|:---: | --- |
| | image |
| | Packages of type `Dependency` that have the tag `cake-addin` |

---

| :jigsaw: | Packages identified as Cake extension of type "Module" (via tags) |
|:---: | --- |
| | image |
| | Packages of type `Dependency` that have the tag `cake-module` |

---

| :jigsaw: | Packages identified as Cake extension of type "Recipe" (via tags) |
|:---: | --- |
| | image |
| | Packages of type `Dependency` that have the tag `cake-recipe` |

---

| :package: | Packages identified as Dependency (via PackageType) |
|:---: | --- |
| | image |
| | Packages of type `Dependency` |

---

/cc @joelverhagen

Addresses https://github.com/NuGet/NuGetGallery/issues/8381

### DIFF
--- a/src/Bootstrap/dist/js/bootstrap.js
+++ b/src/Bootstrap/dist/js/bootstrap.js
@@ -1,6 +1,6 @@
 /*!
  * Bootstrap v3.4.1 (https://getbootstrap.com/)
- * Copyright 2011-2020 Twitter, Inc.
+ * Copyright 2011-2021 Twitter, Inc.
  * Licensed under the MIT license
  */
 
--- a/src/Bootstrap/less/theme/page-display-package.less
+++ b/src/Bootstrap/less/theme/page-display-package.less
@@ -300,9 +300,12 @@
             // in high contrast mode (since borders are shown).
             border-color: @panel-footer-bg;
             border-style: solid;
-            border-width: 1px 0 1px 1px;
+            border-width: 1px;
+            border-radius: 0;
             user-select: all;
             vertical-align: middle;
+            padding: 10.5px;
+            margin: 0 0 11px;
           }
 
           .copy-button {
--- a/src/Bootstrap/less/theme/page-display-package.less
+++ b/src/Bootstrap/less/theme/page-display-package.less
@@ -300,12 +300,10 @@
             // in high contrast mode (since borders are shown).
             border-color: @panel-footer-bg;
             border-style: solid;
-            border-width: 1px;
-            border-radius: 0;
+            border-width: 1px 0 1px 1px;
             user-select: all;
             vertical-align: middle;
-            padding: 10.5px;
-            margin: 0 0 11px;
+            word-break: break-word;
           }
 
           .copy-button {
--- a/src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
+++ b/src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Web;
+
+namespace NuGetGallery
+{
+    public static class CakeBuildManagerExtensions
+    {
+        public static string GetCakeInstallPackageCommand(this DisplayPackageViewModel model)
+        {
+            var reference = $"nuget:?package={model.Id}&version={model.Version}";
+
+            if (model.Prerelease)
+            {
+                reference += "&prerelease";
+            }
+
+            if (model.Tags.Contains("cake-addin", StringComparer.CurrentCultureIgnoreCase))
+            {
+                return $"#addin {reference}";
+            }
+
+            if (model.Tags.Contains("cake-module", StringComparer.CurrentCultureIgnoreCase))
+            {
+                return $"#module {reference}";
+            }
+
+            if (model.Tags.Contains("cake-recipe", StringComparer.CurrentCultureIgnoreCase))
+            {
+                return $"#load {reference}";
+            }
+
+            return string.Join(Environment.NewLine, new[]
+            {
+                $"// Install {model.Id} as a Cake Addin",
+                $"#addin {reference}",
+                "",
+                $"// Install {model.Id} as a Cake Tool",
+                $"#tool {reference}",
+            });
+
+        }
+    }
+}
--- a/src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
+++ b/src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
 using System.Linq;
 
 namespace NuGetGallery
--- a/src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
+++ b/src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
@@ -1,8 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.CompilerServices;
-using System.Web;
 
 namespace NuGetGallery
 {
@@ -10,11 +7,17 @@ namespace NuGetGallery
     {
         public static string GetCakeInstallPackageCommand(this DisplayPackageViewModel model)
         {
-            var reference = $"nuget:?package={model.Id}&version={model.Version}";
+            var scheme = model.IsDotnetToolPackageType ? "dotnet" : "nuget";
+            var reference = $"{scheme}:?package={model.Id}&version={model.Version}";
 
             if (model.Prerelease)
             {
                 reference += "&prerelease";
+            }
+
+            if (model.IsDotnetToolPackageType)
+            {
+                return $"#tool {reference}";
             }
 
             if (model.Tags.Contains("cake-addin", StringComparer.CurrentCultureIgnoreCase))
--- a/src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
+++ b/src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
@@ -8,6 +8,11 @@ namespace NuGetGallery
 {
     public static class CakeBuildManagerExtensions
     {
+        public static bool IsCakeExtension(this DisplayPackageViewModel model)
+        {
+            return IsCakeAddin(model) || IsCakeModule(model) || IsCakeRecipe(model);
+        }
+
         public static string GetCakeInstallPackageCommand(this DisplayPackageViewModel model)
         {
             var reference = $"nuget:?package={model.Id}&version={model.Version}";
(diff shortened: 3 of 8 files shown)


src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
```
joelverhagen, line 1: We have a common file header for all C#, please copy

src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                reference += "&prerelease";
+            }
+
+            if (model.IsDotnetToolPackageType)
+            {
+                return $"#tool {reference}";
+            }
+
+            if (model.Tags.Contains("cake-addin", StringComparer.CurrentCultureIgnoreCase))
```
joelverhagen, line 23: `StringComparer.OrdinalIgnoreCase`. bit better perf will avoid any strange culture-specific issues.

joelverhagen, line 23: `model.Tags` can be null

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -35,7 +35,13 @@
                 InstallPackageCommand = string.Format("dotnet tool install --global {0} --version {1}", Model.Id, Model.Version),
                 AlertLevel = AlertLevel.Info,
                 AlertMessage = "This package contains a <a href='https://aka.ms/global-tools'>.NET Core Global Tool</a> you can call from the shell/command line.",
-            }
+            },
+
+            new ThirdPartyPackageManagerViewModel("Cake", "https://cakebuild.net/support/nuget")
+            {
+                Id = "cake",
```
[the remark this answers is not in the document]
joelverhagen: Going one step further, what do you think about having a different ID for the non-tagged case as well? For example, maybe a user wants the Cake command for a `cake-*` package because there's a specific intent that it's a Cake package. But for other random dependency packages perhaps a better guess is the PackageReference tab? In other words, three IDs:
- cake-tool
- cake-generic
- cake-tag

Or something like that. We could even have a different ID per Cake tag. This will support more granular copy events that we record but maybe that granularity is too much.

src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Linq;
+
+namespace NuGetGallery
+{
+    public static class CakeBuildManagerExtensions
+    {
+        public static string GetCakeInstallPackageCommand(this DisplayPackageViewModel model)
+        {
+            var scheme = model.IsDotnetToolPackageType ? "dotnet" : "nuget";
+            var reference = $"{scheme}:?package={model.Id}&version={model.Version}";
```
joelverhagen, line 11: This is a very long string for some package ID/version combinations without whitepace breaking it. This results in some not-so-prettyness:
image

Would it be possible to do some CSS/`<wbr>` tricks to make the line break better? (perhaps on `&` or `=`)

tests/NuGetGallery.Facts/Extensions/CakeBuildManagerExtensionsFacts.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            [Fact]
+            public void ReturnsALoadDirectiveWithPrerelease()
+            {
+                // Arrange
+                var model = new DisplayPackageViewModel
+                {
+                    Id = "Cake.Recipe",
+                    Version = "1.0.0",
+                    Tags = new[] { "cake-recipe" },
```
joelverhagen, line 157: Need null tag tests for each of these cases

joelverhagen:
/cc @jcjiang and @JonDouglas, could you take a look?

I have a couple code level comment but I'd like your input on:

1. Should we add the "Cake" tab to .NET tools as well (see the last screenshot above). It was not spec'd but it seems like a reasonable addition especially given there is only a single tab there today. I'll let you make the call.
2. Should selecting the "Cake" tab on a dependency package make it stick to "Cake" on a tool package as well? My thought is **no**. We shouldn't assume the user selection on one usage pattern (dependency package) indicates anything about a different usage pattern (.net tool). If we decouple them the default selected tab can still both be Cake but the user will have to select it separately on the two page types.

augustoproiete:
@joelverhagen Thanks for the review!! I'll address the code-level comments later tonight, and will move the .NET Tools part to a separate PR so it doesn't block the other features from moving forward.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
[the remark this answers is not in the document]
augustoproiete: I confirm that because both Cake tabs (Dependency and .NET Tool) are using the same ID `cake` it becomes sticky on both. I'll make sure to use a different ID on the separate PR for .NET Tool.

Having a separate ID for cake-tagged dependency packages vs regular dependency packages make sense to me as we can't know for sure if the user is looking to use the package with Cake or not - we just know they _can_. `PackageReference` would be a better guess nowadays.

Keeping the same ID across any `cake-*`-tagged package I think leads to a better UX because if the user visited a `cake-addin`-tagged package before and clicked on the `Cake` tab, and now goes to a `cake-module`-tagged package, I'd say it's very likely they are looking for the `Cake` tab again. So unless you value the granular copy events, I'd suggest using the same ID for any (`cake-addin`, `cake-module`, `cake-recipe`)-tagged packages.

src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
augustoproiete: Added common file header to the two new C# files

src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
augustoproiete: Updated to `StringComparer.OrdinalIgnoreCase`

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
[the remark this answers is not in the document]
augustoproiete: I'm now using the ID `cake-extension` for packages that have a `cake-*` tag, and using the ID `cake` for regular dependency packages.

src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
augustoproiete: Added null check & tests

src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
augustoproiete: I was able to make it wrap on `-` with `white-space: pre-wrap;` on a `div`

image

tests/NuGetGallery.Facts/Extensions/CakeBuildManagerExtensionsFacts.cs
augustoproiete: Added tests for when tags are `null` or empty

augustoproiete:
@joelverhagen This is ready for round 2. All your comments have been addressed.

I removed the Cake tab from .NET Tool packages, and will send a separate PR with that.

tests/NuGetGallery.Facts/Extensions/CakeBuildManagerExtensionsFacts.cs
```diff
@@ -0,0 +1,216 @@
...
+
+                    var actual = model.GetCakeInstallPackageCommand();
+
+                    Assert.Contains("#addin nuget:?package=Polly&version=1.0.0&prerelease", actual);
+                    Assert.Contains("#tool nuget:?package=Polly&version=1.0.0&prerelease", actual);
+                }
+            }
+        }
+
+        public class PackageTagsKnownToBeCakeExtensions : IEnumerable<object[]>
```
joelverhagen, line 192: This can be a bit more terse with `[MemberData]` where you can return an `IEnumerable<object[]>` from a public static get-only property. I typically have an array initializer (perhaps with nested anonymous type initializers) with the parameters in natural form (precise types) then have a `.Select(x => new object[] { x.A, x.B, ... })` as the last line to make it meet the `IEnumerable<object[]>` contract.

joelverhagen: But you can leave it as-is if you want.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -157,7 +163,7 @@
     <div role="tabpanel" class="tab-pane @(active ? "active" : string.Empty)" id="@packageManager.Id">
         <div>
             <div class="install-script-row">
-                <pre class="install-script" id="@packageManager.Id-text">@packageManager.InstallPackageCommand</pre>
+                <div class="install-script" id="@packageManager.Id-text">@packageManager.InstallPackageCommand</div>
```
joelverhagen, line 166: Can you compare the CSS final rules applied to `pre` vs `div` in Chrome dev tools to ensure we're not losing anything here?

src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
joelverhagen: Wow, this is working a lot better. Nice work!

joelverhagen:
This is looking really good. @NuGet/gallery-team - could I get another pair of eyes on here?

src/Bootstrap/less/theme/page-display-package.less
```diff
@@ -300,9 +300,12 @@
             // in high contrast mode (since borders are shown).
             border-color: @panel-footer-bg;
             border-style: solid;
-            border-width: 1px 0 1px 1px;
+            border-width: 1px;
```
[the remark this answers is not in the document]
augustoproiete: Will do. I did have to run `grunt` in order to test it locally on my machine, but didn't commit the files on `dist` because I thought this was part of your build system.

ps: Maybe you guys need some Cake :cake:? :stuck_out_tongue_winking_eye:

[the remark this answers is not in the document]
joelverhagen: We accept PRs 😸

tests/NuGetGallery.Facts/Extensions/CakeBuildManagerExtensionsFacts.cs
augustoproiete: Np. I updated it to use `[MemberData]` to align with the other tests in the codebase

src/Bootstrap/dist/js/bootstrap.js
```diff
@@ -1,6 +1,6 @@
 /*!
  * Bootstrap v3.4.1 (https://getbootstrap.com/)
- * Copyright 2011-2020 Twitter, Inc.
+ * Copyright 2011-2021 Twitter, Inc.
```
augustoproiete, line 3: updated by the grunt build process

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
augustoproiete: Turned out it was easier to override the `word-break` of the `pre` to get the same result

src/Bootstrap/less/theme/page-display-package.less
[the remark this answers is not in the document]
augustoproiete: I committed the changes to the bootstrap `dist` generated by `grunt`

augustoproiete:
@joelverhagen This is ready for round 3. All your comments have been addressed.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
joelverhagen: Nice.

loic-sharma (approved):
Looks great, nice work! 🎉

lyndaidaii (approved):
Looks good to me, great work. 🥇

joelverhagen (approved):
CI is green. Great work!

JonDouglas:
🚢

PR #8435, opened 2021-03-01
Add IsVulnerable to Manage Packages view model

drewgillies:
Addresses view model portion of https://github.com/NuGet/NuGetGallery/issues/8361.

The intent is to add an `IsVulnerable` field to the packages listed in the Manage Packages view. The packages listed (in both the listed and unlisted lists) are the latest versions of those packages. If those latest versions are vulnerable, we want to display a warning, indicating that something needs to be done. It will be generic and will direct customers to the Package Details page for more information.

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/PackageService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageService.cs
@@ -403,9 +403,12 @@ namespace NuGetGallery
             return package;
         }
 
-        public IEnumerable<Package> FindPackagesByOwner(User user, bool includeUnlisted, bool includeVersions = false)
+        public IEnumerable<Package> FindPackagesByOwner(User user, 
+            bool includeUnlisted, 
+            bool includeVersions = false,
+            bool includeVulnerabilities = false)
         {
-            return GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions);
+            return GetPackagesForOwners(new[] { user.Key }, includeUnlisted, includeVersions, includeVulnerabilities);
         }
 
         /// <summary>
@@ -414,15 +417,17 @@ namespace NuGetGallery
         public IEnumerable<Package> FindPackagesByAnyMatchingOwner(
             User user,
             bool includeUnlisted,
-            bool includeVersions = false)
+            bool includeVersions = false,
+            bool includeVulnerabilities = false)
         {
             var ownerKeys = user.Organizations.Select(org => org.OrganizationKey).ToList();
             ownerKeys.Insert(0, user.Key);
 
-            return GetPackagesForOwners(ownerKeys, includeUnlisted, includeVersions);
+            return GetPackagesForOwners(ownerKeys, includeUnlisted, includeVersions, includeVulnerabilities);
         }
 
-        private IEnumerable<Package> GetPackagesForOwners(IEnumerable<int> ownerKeys, bool includeUnlisted, bool includeVersions)
+        private IEnumerable<Package> GetPackagesForOwners(IEnumerable<int> ownerKeys, bool includeUnlisted,
+            bool includeVersions, bool includeVulnerabilities)
         {
             IQueryable<Package> packages = _packageRepository.GetAll()
                 .Where(p => p.PackageRegistration.Owners.Any(o => ownerKeys.Contains(o.Key)));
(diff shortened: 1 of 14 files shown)


src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -446,11 +451,17 @@ private IEnumerable<Package> GetPackagesForOwners(IEnumerable<int> ownerKeys, bo
...
+            var result = packages
                 .Include(p => p.PackageRegistration)
                 .Include(p => p.PackageRegistration.Owners)
-                .Include(p => p.PackageRegistration.RequiredSigners)
-                .ToList();
+                .Include(p => p.PackageRegistration.RequiredSigners);
+
+            if (includeVulnerabilities)
+            {
+                result = result.Include($"{nameof(Package.VulnerablePackageRanges)}.{nameof(VulnerablePackageVersionRange.Vulnerability)}");
```
joelverhagen, line 461: Ideally this is done with the `Expression<Func<...>>` include style used above. Can that be done or is there some blocker?

src/NuGetGallery/Services/IPackageVulnerabilitiesService.cs
```diff
@@ -14,5 +13,11 @@ public interface IPackageVulnerabilitiesService
         /// </summary>
         /// <param name="id">id of the package for this query</param>
         IReadOnlyDictionary<int, IReadOnlyList<PackageVulnerability>> GetVulnerabilitiesById(string id);
+
+        /// <summary>
+        /// Returns true if the package has a vulnerability
+        /// </summary>
+        /// <param name="package">package to examine</param>
+        bool PackageIsVulnerable(Package package);
```
joelverhagen, line 21: nit `IsPackageVulnerable`

src/NuGetGallery/Services/PackageVulnerabilitiesService.cs
```diff
@@ -41,5 +41,10 @@ public PackageVulnerabilitiesService(IEntitiesContext entitiesContext)
             return !result.Any() ? null :
                 result.ToDictionary(kv => kv.Key, kv => kv.Value as IReadOnlyList<PackageVulnerability>);
         }
+
+        public bool PackageIsVulnerable(Package package) =>
+            package != null
```
joelverhagen, line 46: I think it'd be fine to throw `ArgumentNullException` if someone provides a null package. The caller should be more careful than that. I think it is nonsensical to say that a null package is not vulnerable.

joelverhagen:
Does this PR have performance issues? I vaguely remember you mentioning that this additional include tanks the performance for owners with many, many packages. Perhaps I am misremembering.

joelverhagen (changes requested):
Questions/comments.

src/NuGetGallery.Services/PackageManagement/PackageService.cs
drewgillies: The blocker is the nesting. A second level of nesting could be addressed in .NET Core with a `ThenInclude` but the way to do it here is via a string.

drewgillies:
> Does this PR have performance issues? I vaguely remember you mentioning that this additional include tanks the performance for owners with many, many packages. Perhaps I am misremembering.

No, that was the earlier implementation where I queried the entire set of a package's versions to determine whether it had any vulnerabilities, and I was displaying a warning if that were true. I'm now going with the far more performant (and useful) approach of just indicating if the version being displayed has one or more vulnerability.

drewgillies:
ping @joelverhagen :)

src/NuGetGallery/Services/PackageVulnerabilitiesService.cs
```diff
@@ -41,5 +41,15 @@ public PackageVulnerabilitiesService(IEntitiesContext entitiesContext)
             return !result.Any() ? null :
                 result.ToDictionary(kv => kv.Key, kv => kv.Value as IReadOnlyList<PackageVulnerability>);
         }
+
+        public bool IsPackageVulnerable(Package package)
+        {
+            if (package == null)
+            {
+                throw new ArgumentNullException(nameof(package));
```
joelverhagen, line 49: Should this throw if `VulnerablePackageRanges` is null? That would indicate it was not included in the EF query via `Include(...)`. It should be an empty list instead of null when there are no vulnerable ranges, right? In other words, I think it is strange to return `false` here if the user forgot to fetch VulnerablePackageRanges.

Or will this perform a lazy query if `Include` was not uses and `VulnerablePackageRanges` will never be null?

joelverhagen (approved):
Small question then :shipit:

PR #8436, opened 2021-03-01
Add vulnerable warnings to Manage Packages page

drewgillies:
Note: this PR must be merged first: https://github.com/NuGet/NuGetGallery/pull/8435

Addresses remainder of https://github.com/NuGet/NuGetGallery/issues/8361

This change decorates all packages, of which the latest version is vulnerable, in the manage packages view with a bang and tooltip indicating to go to the details page.

I've also included telemetry here so we can track any perf degradation from the additional querying.

### DIFF
--- a/src/NuGetGallery.Services/Telemetry/TelemetryService.cs
+++ b/src/NuGetGallery.Services/Telemetry/TelemetryService.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Security.Principal;
 using System.Web;
+using Microsoft.Owin.Security.MicrosoftAccount;
 using Newtonsoft.Json;
 using NuGet.Services.Entities;
 using NuGet.Services.FeatureFlags;
@@ -90,6 +91,7 @@ namespace NuGetGallery
             public const string ABTestEvaluated = "ABTestEvaluated";
             public const string PackagePushDisconnect = "PackagePushDisconnect";
             public const string SymbolPackagePushDisconnect = "SymbolPackagePushDisconnect";
+            public const string ManagePackagesQueryPerformance = "ManagePackagesQueryPerformance";
         }
 
         private readonly IDiagnosticsSource _diagnosticsSource;
@@ -225,6 +227,9 @@ namespace NuGetGallery
         public const string IsActive = "IsActive";
         public const string TestBucket = "TestBucket";
         public const string TestPercentage = "TestPercentage";
+
+        // Manage package query performance properties
+        public const string PackageIdCount = "PackageIdCount";
 
         public TelemetryService(IDiagnosticsSource diagnosticsSource, ITelemetryClient telemetryClient)
         {
@@ -1089,6 +1094,14 @@ namespace NuGetGallery
                 properties.Add(IsAuthenticated, isAuthenticated.ToString());
                 properties.Add(TestBucket, testBucket.ToString());
                 properties.Add(TestPercentage, testPercentage.ToString());
+            });
+        }
+
+        public void TrackManagePackagesQueryPerformance(long milliseconds, int packageIdCount)
+        {
+            TrackMetric(Events.ManagePackagesQueryPerformance, milliseconds, properties =>
+            {
+                properties.Add(PackageIdCount, packageIdCount.ToString());
             });
         }
 
--- a/src/NuGetGallery.Services/Telemetry/TelemetryService.cs
+++ b/src/NuGetGallery.Services/Telemetry/TelemetryService.cs
@@ -91,7 +91,6 @@ namespace NuGetGallery
             public const string ABTestEvaluated = "ABTestEvaluated";
             public const string PackagePushDisconnect = "PackagePushDisconnect";
             public const string SymbolPackagePushDisconnect = "SymbolPackagePushDisconnect";
-            public const string ManagePackagesQueryPerformance = "ManagePackagesQueryPerformance";
         }
 
         private readonly IDiagnosticsSource _diagnosticsSource;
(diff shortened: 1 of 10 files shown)


src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -513,6 +514,8 @@ public virtual ActionResult Thanks()
         [UIAuthorize]
         public virtual ActionResult Packages()
```
skofman1, line 515: we can get duration for this method in AI Requests table (like for all other requests). Don we need a special telemetry?

drewgillies: I think what I'll do is scope the telemetry in closer. @joelverhagen and I had discussed the value of specifically measuring perf for the query--I left the telemetry covering the whole method because I didn't know this about all endpoint methods getting duration metrics, and I expect the query is orders of magnitude costlier than other pieces here.

src/NuGetGallery/Controllers/UsersController.cs
```diff
@@ -568,6 +571,9 @@ public virtual ActionResult Packages()
                 IsPackageVulnerabilitiesEnabled = _featureFlagService.IsDisplayVulnerabilitiesEnabled()
             };
 
+            stopwatch.Stop();
+            TelemetryService.TrackManagePackagesQueryPerformance(stopwatch.ElapsedMilliseconds, (listedPackages?.Count ?? 0) + unlistedPackages?.Count ?? 0);
```
loic-sharma, line 575: Just curious, why do you use `listedPackages` and `unlistedPackages` here instead of `packages`?

src/NuGetGallery.Services/Telemetry/TelemetryService.cs
```diff
@@ -1092,6 +1097,14 @@ public void TrackMetricForSearchOnTimeout(string searchName, string correlationI
             });
         }
 
+        public void TrackManagePackagesQueryPerformance(long milliseconds, int packageIdCount)
```
joelverhagen, line 1100: Nit: provide a `TimeSpan` to this method and map to ms/seconds/whatever as an implementation detail.

src/NuGetGallery/Controllers/UsersController.cs
joelverhagen: An additionally telemetry event allows us to correlation duration to number of package IDs. The theory is that for owners with many packages this will be slower and slower. At least that's what I remember from our conversation 😄

joelverhagen (changes requested):
It's unclear to me what the rollout strategy is and whether this implementation (+ dependency PR) will scale to authors with many packages. Could you provide a link to the design in the PR description if one exists?

src/NuGetGallery/Controllers/UsersController.cs
drewgillies: You remember correctly, although the scenario was different. I was looping through entire sets of package versions to determine whether to show the vulnerability bang. This will slow down performance minimally. On reflection, with this in mind, I'll pull the telemetry for now.

src/NuGetGallery/Controllers/UsersController.cs
drewgillies: They were separated in an earlier approach (with no `packages`) and this was a piece of cleanup I missed. Removed now though.

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -171,6 +175,12 @@
                         <td class="align-middle text-nowrap" data-bind="attr: { 'data-sortby': VersionSortOrder }">
                             <span data-bind="text: LatestVersion"></span>
                         </td>
+                    @if (Model.IsPackageVulnerabilitiesEnabled)
+                    {
+                        <td class="package-icon-cell">
+                            <i class="ms-Icon ms-Icon--Warning package-icon" data-bind="visible: IsVulnerable" title="This version has at least one known vulnerability. Details on package details page."></i>
```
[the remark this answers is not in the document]
drewgillies: This is a tooltip, and there is a package link provided to the left. So, it's a good call, but probably needs just to be a change in wording for the tooltip, indicating to click on the package name, like this:
image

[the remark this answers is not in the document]
drewgillies: This is a tooltip, and there is a package link provided to the left. So, it's a good call, but probably needs just to be a change in wording for the tooltip, indicating to click on the package name, like this:
image

drewgillies:
> It's unclear to me what the rollout strategy is and whether this implementation (+ dependency PR) will scale to authors with many packages. Could you provide a link to the design in the PR description if one exists?

There's no reason to think there would be scaling issues. However, I'll do some deploy and measure testing. No specific engineering design document for this work exists as it's a single-field addition to a view model.

src/NuGetGallery.Services/Telemetry/TelemetryService.cs
```diff
@@ -90,6 +91,7 @@ public class Events
             public const string ABTestEvaluated = "ABTestEvaluated";
             public const string PackagePushDisconnect = "PackagePushDisconnect";
             public const string SymbolPackagePushDisconnect = "SymbolPackagePushDisconnect";
+            public const string ManagePackagesQueryPerformance = "ManagePackagesQueryPerformance";
```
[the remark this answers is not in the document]
drewgillies: Telemetry is removed now.

src/NuGetGallery.Services/Telemetry/TelemetryService.cs
drewgillies: Telemetry is removed now.

drewgillies:
ping @joelverhagen @loic-sharma :)

joelverhagen (approved):
It looks like there are changes from another PR in here and some merge conflicts, but the changes unique to this PR look good. Resolve branch issues then :shipit:

PR #8440, opened 2021-03-02
Implementation of Cake tab with instructions for .NET Tool packages

augustoproiete:
This PR adds the Cake tab for `PackageType=DotnetTool`, which was not included in the [initial proposal](https://github.com/NuGet/NuGetGallery/issues/8381) because at the time of the proposal support for .NET Tools in Cake was provided by a community package.

Since then this feature has been added to the Cake core codebase and now Cake supports installing .NET tools directly from within Cake scripts through the `#tool` directive & `dotnet` scheme as shown in the screenshot below.

| :hammer_and_wrench: | Packages identified as .NET tool (via PackageType) |
|:---: | --- |
| | image |
| | Packages of type `DotnetTool` |

---

/cc @joelverhagen

Addresses https://github.com/NuGet/NuGetGallery/issues/8381
Relates to https://github.com/NuGet/NuGetGallery/pull/8434

### DIFF
--- a/src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
+++ b/src/NuGetGallery/Extensions/CakeBuildManagerExtensions.cs
@@ -15,11 +15,17 @@ namespace NuGetGallery
 
         public static string GetCakeInstallPackageCommand(this DisplayPackageViewModel model)
         {
-            var reference = $"nuget:?package={model.Id}&version={model.Version}";
+            var scheme = model.IsDotnetToolPackageType ? "dotnet" : "nuget";
+            var reference = $"{scheme}:?package={model.Id}&version={model.Version}";
 
             if (model.Prerelease)
             {
                 reference += "&prerelease";
+            }
+
+            if (model.IsDotnetToolPackageType)
+            {
+                return $"#tool {reference}";
             }
 
             if (IsCakeAddin(model))
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -36,6 +36,12 @@
                 AlertLevel = AlertLevel.Info,
                 AlertMessage = "This package contains a <a href='https://aka.ms/global-tools'>.NET Core Global Tool</a> you can call from the shell/command line.",
             },
+
+            new ThirdPartyPackageManagerViewModel("Cake", "https://cakebuild.net/support/nuget")
+            {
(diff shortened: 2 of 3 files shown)


joelverhagen:
Thanks for splitting this out, @augustoproiete. We'll work internally to pull in the folks that need to review this. Please allow us some time since this review essentially unplanned work for a partner team.

pascalberger:
@JonDouglas Cake basically just calls the `dotnet` CLI internally. Therefore it provides a .NET Tool Provider for which documentation is available here: https://cakebuild.net/docs/writing-builds/preprocessor-directives/tool/dotnet-provider. The implementation of the provider which calls the CLI is here: https://github.com/cake-build/cake/blob/develop/src/Cake.DotNetTool.Module/DotNetToolPackageInstaller.cs#L91.

> Are they installed as global or local tools?

By default, tools will be installed to the configured Cake Tools folder. If the tool should be installed globally on the machine, the user can pass the global parameter.

> How will tools get uninstalled or updated?

There's some handling if the tool is already installed:

* If the Tool is already installed, with required version nothing is done
* If the Tool is already installed, and no specific version has been requested via pre-processor directive, we leave the current version installed
* If the Tool is already installed, but a different version has been requested, an uninstall/install will be performed

> What's the general user experience here compared to say installing a dotnet tool?

Recommended approach is that users declare the .NET tools which the build needs in the build script and pin to a specific version. This will make sure that the required tool is available to the build script. Updates to the tool will be managed by updating the version string in the pre-processor directive.

JonDouglas:
Thank you all kindly for the quick responses.

We've reviewed this with the right people & am happy to announce that we can move forward with this proposal. A big thanks to @pascalberger for a quick turnaround on the functional explanation of things. 

Please work with @joelverhagen & @jcjiang for the final implementation & merging of this PR.

Thanks again for your patience!

🚢

augustoproiete:
Great news @JonDouglas! :tada::cake:

I've just rebased the PR on top of the latest `dev` & resolved the conflicts, and it's ready for review.
/cc @joelverhagen & @jcjiang

joelverhagen (approved):
CI is green, nice simple change. Thanks @augustoproiete!

joelverhagen:
@augustoproiete, Cake friends, et al - this is now live on PROD! For example:
https://www.nuget.org/packages/Microsoft.dotnet-interactive/

image

augustoproiete:
Thanks a lot for all the help with getting this one though @joelverhagen and NuGet team!!! :package: :rocket:

PR #8406, opened 2021-01-29
Image allowlist 

lyndaidaii:
Summary of the changes (in less than 80 characters):

* check domain for most of image source
* check specific path for github badge, there are security vulnerabilities if we only check the github.com domain, we are doing same behavior what VS market doing. 
* allow raw.github.com/ raw.githubusercontent.com/ user-images.githubusercontent.com/ camo.githubusercontent.com
* replace nofollow  with noopener noreferrer nofollow
    * nooperner open new tab for links, prevent malicious attacker from using window.opener to change the content of originating page 
    * noreferrer similar to noopener, especially in older browsers, it make sense to use both

Todo: 

Documentation page about allowlist that customer could refer to
Warning need to be updated with reference above
upload page 
upload
manage package page
manage
display package page
display

Addresses https://github.com/NuGet/Engineering/issues/3507

PR #8443, opened 2021-03-05
Fix condition with readme container & remove dev.azure from allowlist

lyndaidaii:
This is fix for bug I found  with PR https://github.com/NuGet/NuGetGallery/pull/8406

Expect legacy readme work when uploading package without embedded readme

Without changes: legacy readme container invisable at upload package page if we upload package without embedded readme
wrongreadme

After changes:   legacy readme container display when package has legacy readme

uploadPage

No problem with embedded readme

Addresses https://github.com/NuGet/Engineering/issues/3507

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/async-file-upload.js
+++ b/src/NuGetGallery/Scripts/gallery/async-file-upload.js
@@ -255,10 +255,10 @@
                 $(reportContainerElement).attr("data-bind", "template: { name: 'verify-metadata-template', data: data }");
                 $("#verify-package-container").append(reportContainerElement);
                 ko.applyBindings({ data: model }, reportContainerElement);
-                if (model.ReadmeFileContents.Content) {
-                    $('#import-readme-container').addClass('hidden');
-                } else {
+                if (model.ReadmeFileContents == null) {
                     $('#import-readme-container').removeClass('hidden');
+                } else if (model.ReadmeFileContents.Content) {
+                    $('#import-readme-container').addClass('hidden');
                 }
 
                 var submitContainerElement = document.createElement("div");
--- a/src/NuGetGallery/Scripts/gallery/async-file-upload.js
+++ b/src/NuGetGallery/Scripts/gallery/async-file-upload.js
@@ -255,6 +255,9 @@
                 $(reportContainerElement).attr("data-bind", "template: { name: 'verify-metadata-template', data: data }");
                 $("#verify-package-container").append(reportContainerElement);
                 ko.applyBindings({ data: model }, reportContainerElement);
+                //Content of ReadmeFileContents indicates if embedded readme exists in the package.
+                //Support legacy readme by displaying readme container if ReadmeFileContents is null.
+                //Disable legacy readme by hiding readme container only if embedded readme content is not null.
                 if (model.ReadmeFileContents == null) {
                     $('#import-readme-container').removeClass('hidden');
                 } else if (model.ReadmeFileContents.Content) {
--- a/src/NuGetGallery/App_Data/Files/Content/Trusted-Image-Domains.json
+++ b/src/NuGetGallery/App_Data/Files/Content/Trusted-Image-Domains.json
@@ -17,7 +17,6 @@
         "codecov.io",
         "codefactor.io",
         "coveralls.io",
-        "dev.azure.com",
         "gitlab.com",
         "img.shields.io",
         "isitmaintained.com",


joelverhagen (approved):

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -255,10 +255,10 @@
                 $(reportContainerElement).attr("data-bind", "template: { name: 'verify-metadata-template', data: data }");
                 $("#verify-package-container").append(reportContainerElement);
                 ko.applyBindings({ data: model }, reportContainerElement);
-                if (model.ReadmeFileContents.Content) {
-                    $('#import-readme-container').addClass('hidden');
-                } else {
+                if (model.ReadmeFileContents == null) {
```
joelverhagen, line 258: You could just do `!model.ReadmeFileContents`

src/NuGetGallery/Scripts/gallery/async-file-upload.js
```diff
@@ -255,10 +255,10 @@
...
                     $('#import-readme-container').removeClass('hidden');
+                } else if (model.ReadmeFileContents.Content) {
+                    $('#import-readme-container').addClass('hidden');
                 }
```
drewgillies, line 262: Is there a case where `model.ReadmeFileContents` is not null and `model.ReadmeFileContents.Content` is false?

lyndaidaii: Yes, eg, you include the readme in nuspc, readme file is empty/ or no actual readme file exist.

zhhyu:
Could you provide more explanations about how it works and why it was broken? I don't very understand this fix, and how it deals with the "Submit" button. Thanks!

src/NuGetGallery/Scripts/gallery/async-file-upload.js
drewgillies: And in this case we don't change container state? i.e. if it's hidden it stays hidden, if it's visible it stays visible?

lyndaidaii:
> Could you please provide more explanations about how it works and why it was broken? I don't very understand this fix, and how it deals with the "Submit" button. Thanks!

In PR below, In order to preview image when we upload package with embedded readme, ReadmeFileContents was changed to type of RenderedMarkdownResult instead of string. 

https://github.com/NuGet/NuGetGallery/pull/8406/files#diff-289b3f52256d09316200396f1a20bfd6526bcb73b9bbc9e6a1501e3dae8454c8R258

`                if (model.ReadmeFileContents.Content) {
                    $('#import-readme-container').addClass('hidden');
                } else {
                    $('#import-readme-container').removeClass('hidden');`

Here, ReadmeFileContents are NULL, it will never go to if statement, which readme container is disapear. 


https://github.com/NuGet/NuGetGallery/pull/8406/files#diff-af1ecc85bc3f873d40ff9cc74d6335b402a60ed36fe472cf693d2ad8f47106c0R114

src/NuGetGallery/Scripts/gallery/async-file-upload.js
lyndaidaii: If this is case, we won't disable legacy readme container since customer wants to use embedded readme, legacy readme will be disabled

src/NuGetGallery/Scripts/gallery/async-file-upload.js
loic-sharma: Do you think it's worth adding a quick comment to describe the `null` vs `false` cases?

loic-sharma (approved):

src/NuGetGallery/Scripts/gallery/async-file-upload.js
drewgillies: I'm happy with the change you proposed in response to @joelverhagen's comment--I believe this case was being missed in code.

drewgillies:
I've hit the re-review button for Joel Verhagen--you can consider his approval mine too, as your response to his suggestion satisfies my concern too.

joelverhagen (approved):

zhhyu:
> > Could you please provide more explanations about how it works and why it was broken? I don't very understand this fix, and how it deals with the "Submit" button. Thanks!
> 
> @zhhyu, please let me know if my explanationis not clear.
> In PR below, In order to preview image when we upload package with embedded readme, ReadmeFileContents was changed to type of RenderedMarkdownResult instead of string.
> 
> https://github.com/NuGet/NuGetGallery/pull/8406/files#diff-289b3f52256d09316200396f1a20bfd6526bcb73b9bbc9e6a1501e3dae8454c8R258
> 
> ```
>                if (model.ReadmeFileContents.Content) {
>                     $('#import-readme-container').addClass('hidden');
>                 } else {
>                     $('#import-readme-container').removeClass('hidden');
> ```
> 
> Here, ReadmeFileContents are NULL, it will never go to if statement, which readme container disapear.
> 
> https://github.com/NuGet/NuGetGallery/pull/8406/files#diff-af1ecc85bc3f873d40ff9cc74d6335b402a60ed36fe472cf693d2ad8f47106c0R114

I am still curious about why the submit button is associated here.

lyndaidaii:
> > > Could you please provide more explanations about how it works and why it was broken? I don't very understand this fix, and how it deals with the "Submit" button. Thanks!
> > 
> > 
> > @zhhyu, please let me know if my explanationis not clear.
> > In PR below, In order to preview image when we upload package with embedded readme, ReadmeFileContents was changed to type of RenderedMarkdownResult instead of string.
> > https://github.com/NuGet/NuGetGallery/pull/8406/files#diff-289b3f52256d09316200396f1a20bfd6526bcb73b9bbc9e6a1501e3dae8454c8R258
> > ```
> >                if (model.ReadmeFileContents.Content) {
> >                     $('#import-readme-container').addClass('hidden');
> >                 } else {
> >                     $('#import-readme-container').removeClass('hidden');
> > ```
> > 
> > 
> > Here, ReadmeFileContents are NULL, it will never go to if statement, which readme container disapear.
> > https://github.com/NuGet/NuGetGallery/pull/8406/files#diff-af1ecc85bc3f873d40ff9cc74d6335b402a60ed36fe472cf693d2ad8f47106c0R114
> 
> I am still curious about why the submit button is associated here?

@zhhyu , there is exception threw for ReadmeFileContents is null, initialization for upload page will be hold, which means submit session won't be initailzed becuase of 404, please see screenshoot below
ex

PR #8446, opened 2021-03-06
Symbols perf improvement

skofman1:
Fixes issue: https://github.com/NuGet/Engineering/issues/3676

Also changed default behavior of Symbols upload & Icons upload in Gallery to improve on-prem experience.

### DIFF
--- a/src/NuGetGallery/App_Data/Files/Content/Symbols-Configuration.json
+++ b/src/NuGetGallery/App_Data/Files/Content/Symbols-Configuration.json
@@ -1,5 +1,5 @@
 {
-    "isSymbolsUploadEnabledForAll": false,
+    "isSymbolsUploadEnabledForAll": true,
     "alwaysEnabledForEmailAddresses": [
     ],
     "alwaysEnabledForDomains": [
--- a/src/NuGetGallery/App_Data/Files/Content/Symbols-Configuration.json
+++ b/src/NuGetGallery/App_Data/Files/Content/Symbols-Configuration.json
@@ -1,5 +1,5 @@
 {
-    "isSymbolsUploadEnabledForAll": true,
+    "isSymbolsUploadEnabledForAll": false,
     "alwaysEnabledForEmailAddresses": [
     ],
     "alwaysEnabledForDomains": [
--- a/src/NuGetGallery/Controllers/ApiController.cs
+++ b/src/NuGetGallery/Controllers/ApiController.cs
@@ -193,17 +193,42 @@ namespace NuGetGallery
         [ActionName("GetSymbolPackageApi")]
         public virtual async Task<ActionResult> GetSymbolPackage(string id, string version)
         {
-            return await GetPackageInternal(id, version, isSymbolPackage: true);
+            // some security paranoia about URL hacking somehow creating e.g. open redirects
+            // validate user input: explicit calls to the same validators used during Package Registrations
+            // Ideally shouldn't be necessary?
+            if (!PackageIdValidator.IsValidPackageId(id ?? string.Empty))
+            {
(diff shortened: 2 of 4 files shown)


src/NuGetGallery/App_Data/Files/Content/Symbols-Configuration.json
```diff
@@ -1,5 +1,5 @@
 {
-    "isSymbolsUploadEnabledForAll": false,
+    "isSymbolsUploadEnabledForAll": true,
```
loic-sharma, line 2: The on-prem Gallery can't be used a symbol server. This may be clearer to customers if we left this as disabled

loic-sharma (approved):
Nice!

src/NuGetGallery/Controllers/ApiController.cs
```diff
@@ -193,17 +193,42 @@ protected ApiController()
...
-            return await GetPackageInternal(id, version, isSymbolPackage: true);
+            // some security paranoia about URL hacking somehow creating e.g. open redirects
+            // validate user input: explicit calls to the same validators used during Package Registrations
+            // Ideally shouldn't be necessary?
+            if (!PackageIdValidator.IsValidPackageId(id ?? string.Empty))
+            {
+                return new HttpStatusCodeWithBodyResult(HttpStatusCode.BadRequest, "The format of the package id is invalid");
+            }
+
+            if (string.IsNullOrEmpty(version))
```
zhhyu, line 204: It seems that the old flow contains the logic to retrieve the latest version of the package if we don't feed "version", and now we lose this ability. Will it break the API definition?

tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
```diff
@@ -1968,134 +1968,126 @@ public async Task WillNotUnlistThePackageIfItIsLocked()
...
+                var actionResult = new RedirectResult("http://foo");
 
                 var controller = new TestableApiController(GetConfigurationService(), MockBehavior.Strict);
                 controller.MockPackageService
                     .Setup(x => x.FindPackageByIdAndVersionStrict(packageId, packageVersion))
-                    .Returns(package).Verifiable();
-                controller.MockSymbolPackageFileService
-                    .Setup(x => x.CreateDownloadSymbolPackageActionResultAsync(It.IsAny<Uri>(), packageId, packageVersion))
-                    .Returns(Task.FromResult<ActionResult>(new HttpStatusCodeWithBodyResult(HttpStatusCode.OK, "Test package")))
+                    .Returns((Package)null).Verifiable();
```
drewgillies, line 2072: I don't believe this needs to be verifiable if you're not verifying it in the asserts. Nor does `MockPackageFileService`.

drewgillies (approved):

src/NuGetGallery/Controllers/ApiController.cs
skofman1: I checked the logs and there is no usage of this API without a version. In addition, this is an undocumented API so we have the right to make breaking changes.

src/NuGetGallery/App_Data/Files/Content/Symbols-Configuration.json
skofman1: ok. Changed back.

tests/NuGetGallery.Facts/Controllers/ApiControllerFacts.cs
skofman1: Good catch! The test actually wasn't testing what it was supposed to be testing at all (not my code). Fixed.

PR #8448, opened 2021-03-10
Add feature flag for new TFM heuristics

drewgillies:
Adding a feature flag to this change: https://github.com/NuGet/NuGetGallery/pull/8432

I'd like a fallback codepath in case this triggers anything unforeseen.

@agr this is the change I'd like to get in before integrations.

### DIFF
--- a/src/NuGetGallery.Services/PackageManagement/PackageService.cs
+++ b/src/NuGetGallery.Services/PackageManagement/PackageService.cs
@@ -32,6 +32,7 @@ namespace NuGetGallery
         private readonly ISecurityPolicyService _securityPolicyService;
         private readonly IEntitiesContext _entitiesContext;
         private readonly IContentObjectService _contentObjectService;
+        private readonly IFeatureFlagService _featureFlagService;
         private const int packagesDisplayed = 5;
 
         public PackageService(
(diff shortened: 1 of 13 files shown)


src/NuGetGallery.Services/PackageManagement/PackageService.cs
```diff
@@ -702,8 +705,15 @@ private Package CreatePackageFromNuGetPackage(PackageRegistration packageRegistr
             return package;
         }
 
-        public virtual IEnumerable<NuGetFramework> GetSupportedFrameworks(PackageArchiveReader package) =>
-            GetSupportedFrameworks(package.NuspecReader, package.GetFiles().ToList());
+        public virtual IEnumerable<NuGetFramework> GetSupportedFrameworks(PackageArchiveReader package)
+        {
+            if (_featureFlagService.ArePatternSetTfmHeuristicsEnabled())
```
joelverhagen, line 710: There should be a UT verifying this behavior. You may need a test package that is known to behave differently in these two flows.

lyndaidaii (approved):
Address Joel's comments, LGTM

joelverhagen (approved):
Nice test 👍. Thanks!

PR #8450, opened 2021-03-12
Add Knapcode.MiniZip to 3rd party signing

drewgillies:
Build broken due to Knapcode.MiniZip.dll not being 3rd party signed. Added.


loic-sharma (approved):
From a technical standpoint this change looks good. Please make sure this package is acceptable for use by Microsoft

Issue #8460, opened 2021-03-19
[README] Empty embedded readme breaks the manage package > documentation section


joelverhagen:
### Describe the bug

I created a package with an empty README file. It seems to confuse the upload UI and the manage package Documentation section. It sort of looks like some of the code thinks it's a legacy readme package and other code thinks its an embedded readme.

### To Reproduce

Go here: https://dev.nugettest.org/packages/ICanHazReadme/1.0.2/Manage

Or

1. Create a nupkg with an embedded readme and make the readme empty (0 bytes)
2. Upload to Gallery
3. Notice the README preview on the upload package
4. After the package is uploaded, go to the Manage Package page and expand the Documentation section

### Expected behavior

Either the package should be rejected or the package should be considered to have an embedded readme and no Preview button should appear.

### Screenshots

No submit button:
![image](https://user-images.githubusercontent.com/94054/111712721-71c9cb00-880b-11eb-8ed0-5522618fd639.png)

lyndaidaii:
reject empty readme file to aglin with client pack validation behavior

PR #8464, opened 2021-03-24
reject empty readme file & fix error message for invalid extension

lyndaidaii:

reject empty readme file: 
emptyReadme

correct error message for invalid extension: 
invalidextension

Addresses https://github.com/NuGet/Engineering/issues/3707, https://github.com/NuGet/NuGetGallery/issues/8460

### DIFF
--- a/src/NuGetGallery/Services/PackageMetadataValidationService.cs
+++ b/src/NuGetGallery/Services/PackageMetadataValidationService.cs
@@ -458,6 +458,16 @@ namespace NuGetGallery
             }
 
             var readmeFileEntry = nuGetPackage.GetEntry(readmeFilePath);
+
+            if (readmeFileEntry.Length == 0)
+            {
+                return PackageValidationResult.Invalid(
+                    string.Format(
+                        Strings.ReadmeErrorEmpty,
+                        Strings.UploadPackage_ReadmeFileType,
+                        readmeFilePath));
+            }
+
             if (readmeFileEntry.Length > MaxAllowedReadmeLengthForUploading)
             {
                 return PackageValidationResult.Invalid(
--- a/src/NuGetGallery/Strings.resx
+++ b/src/NuGetGallery/Strings.resx
@@ -1166,7 +1166,7 @@ The {1} Team</value>
     <value>The &lt;readme&gt; element is not currently supported.</value>
   </data>
   <data name="UploadPackage_InvalidReadmeFileExtension" xml:space="preserve">
-    <value>The readme file has an invalid extension '{0}'. Extension must be one of the following: {1}.</value>
+    <value>The readme file has an invalid extension '{0}'. Extension must be end in: {1}.</value>
     <comment>{0} is the readme file extension specified in the .nuspec, {1} is the list of allowed extensions</comment>
   </data>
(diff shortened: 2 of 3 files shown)


zhhyu:
Should we also consider fixing the JS rather than just reject this kind of packages?

lyndaidaii:
> Should we also consider fixing the JS rather than just reject this kind of packages?

Main reason to reject empty pakcage becasue we want to be consistent with what client validation during pack.

src/NuGetGallery/Strings.resx
```diff
@@ -1166,7 +1166,7 @@ The {1} Team</value>
     <value>The &lt;readme&gt; element is not currently supported.</value>
   </data>
   <data name="UploadPackage_InvalidReadmeFileExtension" xml:space="preserve">
-    <value>The readme file has an invalid extension '{0}'. Extension must be one of the following: {1}.</value>
+    <value>The readme file has an invalid extension '{0}'. Extension must be end in: {1}.</value>
```
loic-sharma, line 1169: What do you think of this?

```suggestion
    <value>The readme file has an invalid extension '{0}'. The extension must be: {1}.</value>
```

src/NuGetGallery/Strings.resx
```diff
@@ -1215,4 +1215,7 @@ The {1} Team</value>
   <data name="UploadPackage_OwnerlessIdNamespaceConflictHtml" xml:space="preserve">
     <value>The package ID is reserved. You can upload your package with a different package ID. Reach out to &lt;a href="mailto:support@nuget.org"&gt;support@nuget.org&lt;/a&gt; if you have questions.</value>
   </data>
+  <data name="ReadmeErrorEmpty" xml:space="preserve">
+    <value>The readme file '{0}' cannot be empty</value>
```
loic-sharma, line 1219: Should this end with a period?

```suggestion
    <value>The readme file '{0}' cannot be empty.</value>

```

loic-sharma (approved):

src/NuGetGallery/Services/PackageMetadataValidationService.cs
```diff
@@ -458,6 +458,16 @@ private async Task<PackageValidationResult> CheckReadmeMetadataAsync(PackageArch
             }
 
             var readmeFileEntry = nuGetPackage.GetEntry(readmeFilePath);
+
+            if (readmeFileEntry.Length == 0)
```
joelverhagen, line 462: What about a readme with just whitespace? Would it be worth rejecting those too?

lyndaidaii: good question, probably not rejecting those? @chgill-MSFT  what do you think?

zhhyu: We should keep a consistent behavior on validations between the server and client.
If the client accepts a whitespace readme but the server rejects it, it's not a good customer experience.

joelverhagen: Our validations are already a superset of client, by design. I don't think an empty string README is useful. But also, it's the author's fault if they do this. My guess it is likely unintentional and not a helpful content to the user. But I don't feel strongly.

PR #8420, opened 2021-02-12
Update Service Bus libraries from ServerCommon

loic-sharma:
See: https://github.com/NuGet/ServerCommon/pull/359

Tracked by: https://github.com/NuGet/Engineering/issues/3650

DEV build: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=4585079&view=results
DEV test release: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=989991

NuGetGallery.Core build: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=4585250&view=results

PR #8466, opened 2021-03-25
Gallery storage secret refresh

agr:
Progress on https://github.com/NuGet/Engineering/issues/3522

This change makes Gallery refresh storage credentials. While the general mechanism to refresh secrets is introduced, I did *not* go through all DI container registrations to make sure we don't have other singletons holding on to old secrets. The goal was to make storage secrets rotate at this moment.

The problem with secret refresh in Gallery has several reasons:
* We have too many singletons that inject secrets once at their creation, then hold on to those injected values;
* The `AppConfiguration` object is expensive to create (and we need one for every request we get) as it might try to read secrets from KeyVault, so it was registered in DI container as a singleton as well, so there was no way to retrieve fresh set of credentials.

To address the latter concern a [change](https://github.com/NuGet/ServerCommon/pull/363) in ServerCommon was made that introduces the sync secret injection with refresh in background task. Now, `ConfigurationService` during startup enumerates all the secrets Gallery uses, feeds them into the sync secret reader to start the background refresh loop. Now, injecting secret is a sync operation that never does any I/O. The downside here is that if we don't discover all secrets during startup, the injection of those missed secrets will fail at run time. To mitigate that a self-healing mechanism was added to the sync reader that would add those new secrets to the refresh list, so while the initial attempts to inject those would fail, it should eventually start working. Sync reader emits a metric if it runs into such situation, we'll have an alert for that.

The process of enumerating all secrets got a bit complicated since Authentication-related settings had their own mechanism of reading secrets.

The next major change was making sure storage-related registrations are not singletons. There were a few issues with those:
* Object themselves once constructed hold on to resolved secrets forever;
* Object factories in `ConfigureForAzureStorage` used a single instance of `AppConfiguration` to create new instances;
* Some of storage-related objects have to be singletons.

`EditableFeatureFlagFileStorageService` and `ContentService` have some caching mechanisms inside, which requires them to be a singleton. One option to solve it would be to extract the caching functionality into a separate class that would not have anything to do with underlying storage, but that would be quite complicated change, instead, instances of those classes now get storage service factories, instantiate individual services as needed and discard them right after that. Some DI registrations had to be updated to make sure they receive properly keyed factories.

`GalleryMachineKeyConfigurationProvider` was updated to not to try to create configuration for itself if it is not available. Now it waits until `ConfigurationService` is properly initialized.

Since the new sync secret reader emits metrics, we now need to set up telemetry service before the configuration service is fully initialized, so some workarounds had to be added there (specifically, to prevent attempting to inject secrets to telemetry-related options, since those don't contain secrets).

`_configurationTypes` field was added to `ConfigurationService` to unify the way various types of configuration are crated and to make sure the secret enumeration for the new sync reader has a reliable source of data.

To test the change connection string hash logging was added to `CloudBlobClientWrapper` at the point of use (not instantiation),  to verify the refresh actually happens and secret refresh interval was reduced to 15 minutes. Since we are reading SAS tokens generated by KeyVault, they effectively change every second (issue time is precise to the second in it), which provides a good opportunity to verify they are not reused longer than expected. After running for couple of nights in dev and few bugfixes, logs indicate that secrets are indeed refreshed every 15 minutes.

Unfortunately, some of our unit tests assume that `ConfigurationService.Current` always returns the same object, so some workaround was added to `TestGalleryConfigurationService` to make them pass.


joelverhagen:
> The downside here is that if we don't discover all secrets during startup, the injection of those missed secrets will fail at run time. 

I'm not sure if I understand. When would there be a secret we can't discover at runtime? Shouldn't all secret names be referenced in our JSON configuration? Is there some other kind of dynamic secret name I am forgetting about?

agr:
@joelverhagen, I meant to say if the secret discovery code has bugs and misses some secrets.

src/NuGetGallery/Web.config
```diff
@@ -573,6 +573,14 @@
...
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+			<dependentAssembly>
+				<assemblyIdentity name="System.IdentityModel.Tokens.Jwt" publicKeyToken="31BF3856AD364E35" culture="neutral"/>
+				<bindingRedirect oldVersion="0.0.0.0-5.3.0.0" newVersion="5.3.0.0"/>
+			</dependentAssembly>
+			<dependentAssembly>
+				<assemblyIdentity name="Microsoft.Azure.Services.AppAuthentication" publicKeyToken="31BF3856AD364E35" culture="neutral"/>
+				<bindingRedirect oldVersion="0.0.0.0-1.3.1.0" newVersion="1.3.1.0"/>
+			</dependentAssembly>
```
agr, line 583: Need to fix spacing

src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -47,6 +47,7 @@
       <Version>5.9.0</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.FeatureFlags">
+      <Version>2.87.0-agr-secrets-4592564</Version>
```
agr, line 50: Will use release version once the other PR is merged.

src/NuGetGallery.Core/Services/CloudBlobClientWrapper.cs
```diff
@@ -16,20 +19,27 @@ public class CloudBlobClientWrapper : ICloudBlobClient
...
             _storageConnectionString = storageConnectionString;
             _defaultRequestOptions = defaultRequestOptions;
+            _createdAt = DateTime.UtcNow;
         }
 
         public ISimpleCloudBlob GetBlobFromUri(Uri uri)
-        {
+         {
+            // TODO: Remove
+            Trace.TraceInformation($"CloudBlobClientWrapper age: {DateTime.UtcNow - _createdAt} {GetHash(_storageConnectionString)}");
```
agr, line 41: Will remove

src/NuGetGallery/Telemetry/GlobalMetricPropertyProvider.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using NuGetGallery.Helpers;
+
+namespace NuGetGallery
+{
+    public class GlobalMetricPropertyProvider : IGlobalMetricPropertyProvider
```
agr, line 9: It seems, we can drop that. `cloud_RoleInstance` in AI contains the same information.

joelverhagen:
> @joelverhagen Joel Verhagen FTE, I meant to say if the secret discovery code has bugs and misses some secrets.
> 
> Or we have some non-obvious configuration access patterns. I'd say, the way authentication services access their configuration is was bit unexpected and I missed it in initial implementation.

Personally I would prefer a simpler implementation that just breaks/throws if something unexpected happens. I'm a little worried about have a "heavy" solution here since it's a lot of new code going on in the background, which is harder to debug/monitor.

src/NuGetGallery.Services/Telemetry/TelemetryService.cs
```diff
@@ -227,10 +235,20 @@ public class Events
         public const string TestBucket = "TestBucket";
         public const string TestPercentage = "TestPercentage";
 
-        public TelemetryService(IDiagnosticsSource diagnosticsSource, ITelemetryClient telemetryClient)
+        // secret reader related properties
+        public const string SecretName = "SecretName";
+
+        // common properties
+        public const string HostName = "SecretName";
```
agr, line 242: value is wrong.

src/NuGetGallery.Core/Infrastructure/TableErrorLog.cs
```diff
@@ -220,7 +220,7 @@ private void Obfuscate(Error error)
 
             error.ServerVariables["HTTP_X_NUGET_APIKEY"] = string.Empty;
 
-            var forwardedIps = error.ServerVariables["HTTP_X_FORWARDED_FOR"].Split(',');
+            var forwardedIps = (error.ServerVariables["HTTP_X_FORWARDED_FOR"] ?? "").Split(',');
```
agr, line 223: Will revert.

src/NuGetGallery/App_Start/AsyncSecretInjectorAdaptor.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
agr, line 1: delete

src/NuGetGallery/NuGetGallery.csproj
```diff
@@ -2293,6 +2294,9 @@
     <PackageReference Include="WindowsAzure.Caching">
       <Version>1.7.0</Version>
     </PackageReference>
+    <PackageReference Include="WindowsAzure.ServiceBus">
+      <Version>6.0.3</Version>
+    </PackageReference>
```
loic-sharma, line 2299: Is this necessary? The `WindowsAzure.ServiceBus` package reference should come from ServerCommon, this package reference was removed in this PR:

https://github.com/NuGet/NuGetGallery/pull/8420/files#diff-3679aeabd67b851f32395fba538913e2c57eeb62241f445a818e90352f618b0bL2296-L2298

src/NuGetGallery.Services/Configuration/ConfigurationService.cs
```diff
@@ -185,29 +253,129 @@ public string ReadRawSetting(string settingName)
...
+        private ICollection<string> GetSecretNames<T>(T instance, string prefix)
+        {
+            var secretNames = new HashSet<string>();
+            foreach (var property in GetConfigProperties<T>(instance))
+            {
+                string settingName = GetSettingName(property, prefix);
+                string settingRawValue = ReadRawSetting(settingName);
+                if (settingRawValue != null)
+                {
+                    secretNames.UnionWith(NuGet.Services.KeyVault.SecretUtilities.GetSecretNames(settingRawValue, NuGet.Services.KeyVault.SecretInjector.DefaultFrame));
```
loic-sharma, line 271: Would it be possible to remove the namespaces here?

src/NuGetGallery/App_Start/GalleryMachineKeyConfigurationProvider.cs
```diff
@@ -4,14 +4,29 @@
...
+using System.Threading;
 using System.Xml;
 using NuGetGallery.Configuration;
 
 namespace NuGetGallery
 {
     public class GalleryMachineKeyConfigurationProvider : ProtectedConfigurationProvider
     {
-        public static IGalleryConfigurationService Configuration { get; set; }
+        private static ManualResetEventSlim _configurationSet = new ManualResetEventSlim(false);
```
skofman1, line 15: please add comments

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -876,6 +876,12 @@ public static void RegisterApiV2Routes(RouteCollection routes)
                 "api/v2/query",
                 new { controller = "Api", action = "Query" });
 
+            // TODO: remove
+            routes.MapRoute(
+                "v2Benchmark",
+                "api/v2/benchmark",
+                new { controller = "Api", action = "Benchmark" });
+
```
drewgillies, line 884: You'll pull out all of these benchmark TODOs before merging?

src/NuGetGallery.Services/Configuration/SecretReader/EmptySecretReaderFactory.cs
```diff
@@ -5,16 +5,16 @@
 
 namespace NuGetGallery.Configuration.SecretReader
 {
-    public class EmptySecretReaderFactory : ISecretReaderFactory
```
drewgillies, line 8: Did you drop the implementation because it's unused?

zhhyu:
Looking now ...

src/NuGetGallery/App_Start/Routes.cs
agr: Yes

PR #8469, opened 2021-03-30
Using singletons for storage access

agr:
Progress on https://github.com/NuGet/Engineering/issues/3522

This change is the subset of https://github.com/NuGet/NuGetGallery/pull/8466, which includes only the part that makes storage access objects transient where possible.

Configuration here is still singleton, so no potential perf impact.


src/NuGetGallery/Web.config
```diff
@@ -573,6 +573,14 @@
   </system.diagnostics>
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="System.IdentityModel.Tokens.Jwt" publicKeyToken="31BF3856AD364E35" culture="neutral"/>
```
skofman1, line 577: why are those changes needed?

skofman1 (approved):
:shipit:

src/NuGetGallery.Core/Services/CloudBlobClientWrapper.cs
```diff
@@ -51,6 +59,9 @@ public ISimpleCloudBlob GetBlobFromUri(Uri uri)
 
         public ICloudBlobContainer GetContainerReference(string containerAddress)
         {
+            // TODO: Remove
```
skofman1, line 62: don't forget to remove those..

src/NuGetGallery.Core/NuGetGallery.Core.csproj
```diff
@@ -47,7 +47,7 @@
       <Version>5.9.0</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.FeatureFlags">
-      <Version>2.86.0</Version>
+      <Version>2.86.0-agr-ffs-4546672</Version>
```
skofman1, line 50: why is this needed?

agr: Will update that to release. The change in server common was to make feature flag cache use a factory to get instances of storage classes, instead of holding on to a single instance preventing secret rotation.

src/NuGetGallery/Web.config
agr: It was unrelated to the change, just addressing a build warning.

src/NuGetGallery.Core/Services/CloudBlobClientWrapper.cs
```diff
@@ -29,7 +30,7 @@ public CloudBlobClientWrapper(string storageConnectionString, BlobRequestOptions
         }
 
         public ISimpleCloudBlob GetBlobFromUri(Uri uri)
-        {
+         {
```
joelverhagen, line 33:
```suggestion
        {
```

src/NuGetGallery.Core/Services/CloudBlobClientWrapper.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
```
joelverhagen, line 5: Unnecessary?

joelverhagen (approved):

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -1380,14 +1380,13 @@ private static void ConfigureForAzureStorage(ContainerBuilder builder, IGalleryC
...
+            foreach (var dependent in dependents)
             {
                 if (completedBindingKeys.Add(dependent.BindingKey))
                 {
-                    builder.RegisterInstance(new CloudBlobClientWrapper(dependent.AzureStorageConnectionString, configuration.Current.AzureStorageReadAccessGeoRedundant))
-                       .AsSelf()
-                       .As<ICloudBlobClient>()
-                       .SingleInstance()
+                    builder.Register(_ => new CloudBlobClientWrapper(dependent.AzureStorageConnectionString, configuration.Current.AzureStorageReadAccessGeoRedundant))
+                       .InstancePerLifetimeScope()
```
zhhyu, line 1389: I am still very curious about the benefit to use "transient" here?

zhhyu (approved):

src/NuGetGallery.Services/Storage/ContentService.cs
```diff
@@ -20,10 +20,11 @@ public class ContentService : IContentService
         // Each entry should _always_ have a value. Values never expire, they just need to be updated. Updates use the existing data,
         // so we don't want data just vanishing from a cache.
         private ConcurrentDictionary<string, ContentItem> _contentCache = new ConcurrentDictionary<string, ContentItem>(StringComparer.OrdinalIgnoreCase);
+        private Func<IFileStorageService> _fileStorageFactory;
 
         private IDiagnosticsSource Trace { get; set; }
 
-        public IFileStorageService FileStorage { get; protected set; }
+        protected virtual IFileStorageService FileStorage => _fileStorageFactory();
```
zhhyu, line 27: My understanding about the reason to use Func here is because that "ContentService" is a singleton while the "FileStorageService" is transient now, so here it will always give us a new instance of "FileStorageService" each time when we need it. Otherwise, there will be only one "FileStorageService" associated with the singleton "ContentService" during the lifetime of the application. Am I correct?

zhhyu (approved):

src/NuGetGallery.Services/Storage/ContentService.cs
agr: Yup, that's the gist of it. `ContentService` has to be singleton to maintain cache, so I made underlying storage access object transient.

PR #8472, opened 2021-03-30
[Statistics]Adding Support for other reports from alternate container

ryuyu:
Adds support for sourcing other reports from the alternate stats container.
Currently, feature flagging is all-or-nothing.

### DIFF
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -707,18 +707,6 @@ namespace NuGetGallery
 
         private static void RegisterStatisticsServices(ContainerBuilder builder, IGalleryConfigurationService configuration, ITelemetryService telemetryService)
         {
-            // when running on Windows Azure, we use a back-end job to calculate stats totals and store in the blobs
-            builder.RegisterInstance(new JsonAggregateStatsService(configuration.Current.AzureStorage_Statistics_ConnectionString, configuration.Current.AzureStorageReadAccessGeoRedundant))
-                .AsSelf()
-                .As<IAggregateStatsService>()
-                .SingleInstance();
-
-            // when running on Windows Azure, pull the statistics from the warehouse via storage
-            builder.RegisterInstance(new CloudReportService(configuration.Current.AzureStorage_Statistics_ConnectionString, configuration.Current.AzureStorageReadAccessGeoRedundant))
-                .AsSelf()
-                .As<IReportService>()
-                .SingleInstance();
-
             // when running on Windows Azure, download counts come from the downloads.v1.json blob
             builder.Register(c => new SimpleBlobStorageConfiguration(configuration.Current.AzureStorage_Statistics_ConnectionString, configuration.Current.AzureStorageReadAccessGeoRedundant))
                 .SingleInstance()
@@ -727,6 +715,32 @@ namespace NuGetGallery
             builder.Register(c => new SimpleBlobStorageConfiguration(configuration.Current.AzureStorage_Statistics_ConnectionString_Alternate, configuration.Current.AzureStorageReadAccessGeoRedundant))
                 .SingleInstance()
                 .Keyed<IBlobStorageConfiguration>(BindingKeys.AlternateStatisticsKey);
+
+            // when running on Windows Azure, we use a back-end job to calculate stats totals and store in the blobs
+            builder.Register(c =>
+            {
+                var primaryConfiguration = c.ResolveKeyed<IBlobStorageConfiguration>(BindingKeys.PrimaryStatisticsKey);
+                var alternateConfiguration = c.ResolveKeyed<IBlobStorageConfiguration>(BindingKeys.AlternateStatisticsKey);
+                var featureFlagService = c.Resolve<IFeatureFlagService>();
+                var jsonAggregateStatsService = new JsonAggregateStatsService(featureFlagService, primaryConfiguration, alternateConfiguration);
+                return jsonAggregateStatsService;
+            })
+                .AsSelf()
+                .As<IAggregateStatsService>()
+                .SingleInstance();
+
+            // when running on Windows Azure, pull the statistics from the warehouse via storage
+            builder.Register(c =>
+            {
+                var primaryConfiguration = c.ResolveKeyed<IBlobStorageConfiguration>(BindingKeys.PrimaryStatisticsKey);
+                var alternateConfiguration = c.ResolveKeyed<IBlobStorageConfiguration>(BindingKeys.AlternateStatisticsKey);
+                var featureFlagService = c.Resolve<IFeatureFlagService>();
+                var cloudReportService = new CloudReportService(featureFlagService, primaryConfiguration, alternateConfiguration);
+                return cloudReportService;
+            })
+                .AsSelf()
+                .As<IReportService>()
+                .SingleInstance();
 
             builder.Register(c =>
             {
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -707,27 +707,41 @@ private static void RegisterDeleteAccountService(ContainerBuilder builder, Confi
 
         private static void RegisterStatisticsServices(ContainerBuilder builder, IGalleryConfigurationService configuration, ITelemetryService telemetryService)
         {
+            // when running on Windows Azure, download counts come from the downloads.v1.json blob
```
zhhyu, line 710: This comment looks a little confused, is it?

zhhyu (approved):

src/NuGetGallery/Services/CloudReportService.cs
```diff
@@ -1,23 +1,30 @@
...
+        public CloudReportService(
+            IFeatureFlagService featureFlagService,
+            IBlobStorageConfiguration primaryBlobStorageConfiguration,
+            IBlobStorageConfiguration alternateBlobStorageConfiguration)
         {
-            _connectionString = connectionString;
-            _readAccessGeoRedundant = readAccessGeoRedundant;
+            _featureFlagService = featureFlagService ?? throw new ArgumentNullException(nameof(featureFlagService));
+            _primaryStorageConfiguration = primaryBlobStorageConfiguration ?? throw new ArgumentNullException(nameof(primaryBlobStorageConfiguration));
+            _alternateBlobStorageConfiguration = alternateBlobStorageConfiguration;
```
zhhyu, line 27: Will it better if we also do the null check here?

src/NuGetGallery/Services/JsonAggregateStatsService.cs
```diff
@@ -1,31 +1,47 @@
...
+        public JsonAggregateStatsService(
+            IFeatureFlagService featureFlagService,
+            IBlobStorageConfiguration primaryBlobStorageConfiguration,
+            IBlobStorageConfiguration alternateBlobStorageConfiguration)
         {
-            _connectionString = connectionString;
-            _readAccessGeoRedundant = readAccessGeoRedundant;
+            _featureFlagService = featureFlagService ?? throw new ArgumentNullException(nameof(featureFlagService));
+            _primaryStorageConfiguration = primaryBlobStorageConfiguration ?? throw new ArgumentNullException(nameof(primaryBlobStorageConfiguration));
+            _alternateBlobStorageConfiguration = alternateBlobStorageConfiguration;
```
zhhyu, line 27: Will it better if we also do the null check here?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -707,27 +707,41 @@ private static void RegisterDeleteAccountService(ContainerBuilder builder, Confi
...
+
+            builder.Register(c => new SimpleBlobStorageConfiguration(configuration.Current.AzureStorage_Statistics_ConnectionString_Alternate, configuration.Current.AzureStorageReadAccessGeoRedundant))
+                .SingleInstance()
+                .Keyed<IBlobStorageConfiguration>(BindingKeys.AlternateStatisticsKey);
+
             // when running on Windows Azure, we use a back-end job to calculate stats totals and store in the blobs
-            builder.RegisterInstance(new JsonAggregateStatsService(configuration.Current.AzureStorage_Statistics_ConnectionString, configuration.Current.AzureStorageReadAccessGeoRedundant))
+            builder.Register(c =>
+            {
+                var primaryConfiguration = c.ResolveKeyed<IBlobStorageConfiguration>(BindingKeys.PrimaryStatisticsKey);
```
skofman1, line 722: >primaryConfiguration [](start = 20, length = 20)

What do you think about using attributes for resolving keyed dependencies: [https://autofaccn.readthedocs.io/en/latest/advanced/keyed-services.html#resolving-with-attributes](https://autofaccn.readthedocs.io/en/latest/advanced/keyed-services.html#resolving-with-attributes) ?
We might be able to eliminate the creation of JsonAggregateStatsService class here and simply register it.

src/NuGetGallery/Services/CloudReportService.cs
ryuyu: I did no null check intentionally to not force alternate blob storage configuration to have to be set so that it works in place with current config.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
ryuyu: This comment was around this block when i initially made code changes. I'm not sure where it went, but it's back here. I imagine this was originally here to note the difference from running locally, where stats come from DB.

src/NuGetGallery/Services/JsonAggregateStatsService.cs
ryuyu: I did no null check intentionally to not force alternate blob storage configuration to have to be set so that it works in place with current config.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
ryuyu: This looks like it could be good idea, and might make this block a bit easier to read. I'll experiment with it.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -707,27 +707,41 @@ private static void RegisterDeleteAccountService(ContainerBuilder builder, Confi
...
+                var alternateConfiguration = c.ResolveKeyed<IBlobStorageConfiguration>(BindingKeys.AlternateStatisticsKey);
+                var featureFlagService = c.Resolve<IFeatureFlagService>();
+                var jsonAggregateStatsService = new JsonAggregateStatsService(featureFlagService, primaryConfiguration, alternateConfiguration);
+                return jsonAggregateStatsService;
+            })
                 .AsSelf()
```
agr, line 728: Do we need all those `.AsSelf()`? Does any service directly depend on `JsonAggregateStatsService`?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
ryuyu: I'm having a little trouble getting this to work right now. I've filed https://github.com/NuGet/NuGetGallery/issues/8477 to make this change after things are in place (if I am understanding this correctly, we should be able to make this change independent of the overall functionality.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -707,27 +707,41 @@ private static void RegisterDeleteAccountService(ContainerBuilder builder, Confi
 
         private static void RegisterStatisticsServices(ContainerBuilder builder, IGalleryConfigurationService configuration, ITelemetryService telemetryService)
         {
+            // when running on Windows Azure, download counts come from the downloads.v1.json blob
+            builder.Register(c => new SimpleBlobStorageConfiguration(configuration.Current.AzureStorage_Statistics_ConnectionString, configuration.Current.AzureStorageReadAccessGeoRedundant))
```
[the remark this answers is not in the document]
agr: Disregard that. I'll address it with my changes.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
agr: Ah, we have two arguments of type `IBlobStorageConfiguration` here, so it won't work directly.

agr (approved):

PR #8478, opened 2021-04-01
[Stats]Faster stats refresh

ryuyu:
Small change to speed up the stats refresh since pipeline updates should allow the report to generate fast enough for this.

In theory, this will quadruple the hits on storage through this, and performance impact of this has not been analyzed.
However, I do not expect the impact of hitting 6 files 4 times instead of once every hour to be major.

This is not a critical change for functionality, and I'm happy to defer it if we so desire.

### DIFF
--- a/src/NuGetGallery/Services/JsonStatisticsService.cs
+++ b/src/NuGetGallery/Services/JsonStatisticsService.cs
@@ -20,7 +20,7 @@ namespace NuGetGallery
         /// <summary>
         /// How often statistics reports should be refreshed using the <see cref="_reportService"/>.
         /// </summary>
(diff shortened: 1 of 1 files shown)


joelverhagen:
How long does it take to reload the stats reports today?
Do we notice any increase in request durations during the reload?

If the answer to the second question is "no" I'm totally fine with this change. In general I like the idea.

skofman1:
Those are big files we are loading.. Like @joelverhagen  said, we should understand the perf implications first. For example, how long does it take to load a file?

joelverhagen:
Closed per discussion with @ryuyu offline.

PR #8484, opened 2021-04-02
Report services are not using Blob storage SDK classes anymore.

agr:
Report services were using Storage SDK classes directly instead of our wrappers around them.  To centralize blob storage access code this change makes them use `ICloudBlobClient`.

Duplicated code that switches between the primary and alternate source for stats was removed from individual services and moved into DI container. A registration of `ICloudBlobClient` keyed as `BindingKeys.FeatureFlaggedStatisticsKey` resolves to a blob client that points to the proper storage account. Services are supposed to request `Func<ICloudBlobClient>` from the DI container that will provide proper client on each individual call.

A `SimpleCloudBlobExtensions` class was introduced to close the gap between `CloudBlockBlobReference` class provided by the storage SDK and `CloudBlobWrapper` in our code (specifically, our wrapper lacks equivalent of `DownloadTextAsync`). 

Additionally, report services did a `.ExistAsync()` checks before accessing the blob contents, causing unnecessary network round trip. More helpers were added to `SimpleCloudBlobExtensions` to handle missing blobs gracefully on content retrieval instead, saving one HTTP request on each report content access.

Lastly, `CloudDownloadCountService` was made async to accommodate for async-only `ISimpleCloudBlob`.


src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -716,38 +717,54 @@ private static void RegisterStatisticsServices(ContainerBuilder builder, IGaller
...
+                    if (c.Resolve<IFeatureFlagService>().IsAlternateStatisticsSourceEnabled())
+                    {
+                        return c.ResolveKeyed<ICloudBlobClient>(BindingKeys.AlternateStatisticsKey);
+                    }
+                    return c.ResolveKeyed<ICloudBlobClient>(BindingKeys.PrimaryStatisticsKey);
+                })
+                .Keyed<ICloudBlobClient>(BindingKeys.FeatureFlaggedStatisticsKey);
+
+            // when running on Windows Azure, we use a back-end job to calculate stats totals and store in the blobs
+            builder.Register(c => new JsonAggregateStatsService(c.ResolveKeyed<Func<ICloudBlobClient>>(BindingKeys.FeatureFlaggedStatisticsKey)))
```
skofman1, line 753: >JsonAggregateStatsService [](start = 38, length = 25)

with CloudDownloadCountService you register AsSelf, but with JsonAggregateStatsService you don't. Probably doesn't make much of a difference, but I'm wondering about the reason for the discrepancy

skofman1 (approved):
:shipit:

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
agr: Oversight. I double checked and removed `.AsSelf()` from other places.

zhhyu:
looking now ...

src/NuGetGallery.Core/Services/SimpleCloudBlobExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace NuGetGallery
+{
+    public static class SimpleCloudBlobExtensions
+    {
+        /// <summary>
+        /// Retrieves the blob contents as a string assuming UTF8 encoding if the blob exists.
+        /// </summary>
+        /// <param name="blob">Blob reference.</param>
+        /// <returns>The text content of the blob or null if the blob does not exist.</returns>
+        public static async Task<string> DownloadTextIfExistAsync(this ISimpleCloudBlob blob)
```
loic-sharma, line 22: What do you think of this name? I don't feel strongly either ways


```suggestion
        public static async Task<string> DownloadTextIfExistsAsync(this ISimpleCloudBlob blob)
```

loic-sharma: Same comment for `OpenReadIfExistAsync`

loic-sharma, line 22: Could you add unit tests for these methods?

src/NuGetGallery/Infrastructure/Lucene/CloudDownloadCountServiceRefreshJob.cs
```diff
@@ -19,7 +19,7 @@ public CloudDownloadCountServiceRefreshJob(TimeSpan interval, CloudDownloadCount
 
         public override Task Execute()
         {
-            return new Task(_downloadCountService.Refresh);
+            return _downloadCountService.Refresh();
```
loic-sharma, line 22: Can we make sure that all async method names have an `Async` suffix?

Also, please add an `await` here. See: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -716,50 +717,63 @@ private static void RegisterStatisticsServices(ContainerBuilder builder, IGaller
...
+
+            var hasSecondaryStatisticsSource = !string.IsNullOrWhiteSpace(configuration.Current.AzureStorage_Statistics_ConnectionString_Alternate);
+
+            builder.Register(c =>
+                {
+                    if (!hasSecondaryStatisticsSource)
+                    {
+                        return c.ResolveKeyed<ICloudBlobClient>(BindingKeys.PrimaryStatisticsKey);
+                    }
+                    if (c.Resolve<IFeatureFlagService>().IsAlternateStatisticsSourceEnabled())
```
loic-sharma, line 744: Please make sure to test by hand that this works as I believe this is the first time we've used feature flags in DI. From what I can tell, this should work fine though! 😃

loic-sharma (approved):
This is awesome!! 🥇

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -716,50 +717,63 @@ private static void RegisterStatisticsServices(ContainerBuilder builder, IGaller
...
+                    {
+                        return c.ResolveKeyed<ICloudBlobClient>(BindingKeys.AlternateStatisticsKey);
+                    }
+                    return c.ResolveKeyed<ICloudBlobClient>(BindingKeys.PrimaryStatisticsKey);
+                })
```
zhhyu, line 749: Will this be better? Not very sure though.

```
builder.Register(c =>
{
  if (hasSecondaryStatisticsSource && c.Resolve<IFeatureFlagService>().IsAlternateStatisticsSourceEnabled())
  {
    return c.ResolveKeyed<ICloudBlobClient>(BindingKeys.AlternateStatisticsKey);
  } else {
    return c.ResolveKeyed<ICloudBlobClient>(BindingKeys.PrimaryStatisticsKey);
  }                    
})
.Keyed<ICloudBlobClient>(BindingKeys.FeatureFlaggedStatisticsKey);
```

src/NuGetGallery.Core/Services/SimpleCloudBlobExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.WindowsAzure.Storage;
+
+namespace NuGetGallery
+{
+    public static class SimpleCloudBlobExtensions
+    {
```
zhhyu, line 16: I see that we have an "Extensions" folder under "NuGetGallery.Core". Should we also move this extension to that folder?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
agr: I did test that.

src/NuGetGallery/Infrastructure/Lucene/CloudDownloadCountServiceRefreshJob.cs
agr: Unfortunately, our `NuGetJobCoordinator` complains if that `Execute` method returns a started task, so we have to wrap it in a `new Task()`, updated the code.
https://github.com/NuGet/NuGetGallery/blob/be38792c3d3d327765113a32d82cd5b9b68cd1b0/src/NuGetGallery/Infrastructure/Jobs/NuGetJobCoordinator.cs#L45-L60
(line 53)

src/NuGetGallery/Services/CloudDownloadCountService.cs
```diff
@@ -152,24 +148,19 @@ protected virtual int CalculateSum(ConcurrentDictionary<string, int> versions)
         /// This method is added for unit testing purposes. It can return a null stream if the blob does not exist
         /// and assumes the caller will properly dispose of the returned stream.
         /// </summary>
```
zhhyu, line 150: Should we remove this comment? This method is used now.

src/NuGetGallery.Core/Services/SimpleCloudBlobExtensions.cs
zhhyu, line 16: Are there any considerations to use "Extensions" here rather than provide more methods in the interface and wrapper?

zhhyu (approved):

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
agr: Updated the code.

src/NuGetGallery/Services/CloudDownloadCountService.cs
agr: It always had been used. The purpose of the method is to allow descendants to override it, providing mock streams for testing.
The only difference with the old code is that explicit blob existence check was removed from `GetBlobReference` and instead it relies on `OpenReadIfExistsAsync` to provide the same behavior.

src/NuGetGallery/Infrastructure/Lucene/CloudDownloadCountServiceRefreshJob.cs
```diff
@@ -19,7 +19,7 @@ public CloudDownloadCountServiceRefreshJob(TimeSpan interval, CloudDownloadCount
 
         public override Task Execute()
         {
-            return new Task(_downloadCountService.Refresh);
+            return new Task(async () => { await _downloadCountService.Refresh(); });
```
zhhyu, line 22: My shallow understanding here is that we will block a thread from the thread pool, and loose the benefit of async/await to release the current thread. Am I right?
But it should be fine because we only have 1 background thread to execute the refresh action, so scalability is not a concern here.

agr: It shouldn't block, it's just one task wrapped into another because at some point the task returned by the `Execute` method here is used to call `.Start()` and that call is only allowed on tasks explicitly created using the class constructor and that haven't been started by the time of that call.
My understanding is that we should be fine nesting tasks without introducing unnecessary thread blocking.

src/NuGetGallery.Core/Services/SimpleCloudBlobExtensions.cs
agr: Those felt like utility functions doing more than one thing and they could be implemented in terms of `ISimpleCloudBlob` interface, so making them extensions and not burdening implementers of the interface to implement those functions feels like a natural thing to do. After some thought, having them as extensions would complicate mocking the `ISimpleCloudBlob` interface, so I moved them into the interface. If we update to C# 8, we can implement those directly in the interface.

src/NuGetGallery/Infrastructure/Lucene/CloudDownloadCountServiceRefreshJob.cs
agr: At this particular line nothing happens. It creates a task, which, when started would call `_downloadCountService.Refresh()` and asynchronously wait for its completion. My change does not introduce any new thread blocking, just changed the way the function called.
Whatever code calls that `Execute()` method *can* block, but changes I made cannot affect this behavior (i.e. it would have been blocking before the change, too).

PR #8485, opened 2021-04-02
Preparing ErrorLog for secret refresh.

agr:
Using connection string factory in `AzureEntityList`  instead of explicit connection string. Allows singleton to refresh connection string.

### DIFF
--- a/src/NuGetGallery.Core/Infrastructure/AzureEntityList.cs
+++ b/src/NuGetGallery.Core/Infrastructure/AzureEntityList.cs
@@ -22,25 +22,25 @@ namespace NuGetGallery.Infrastructure
         private const string IndexPartitionKey = "INDEX";
         private const string IndexRowKey = "0";
 
-        private CloudTable _tableRef;
+        private readonly string _tableName;
+        private readonly bool _readAccessGeoRedundant;
+        private readonly Func<string> _connectionStringFactory;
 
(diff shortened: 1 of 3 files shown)


src/NuGetGallery.Core/Infrastructure/AzureEntityList.cs
```diff
@@ -333,6 +342,16 @@ System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator()
...
         }
 
+        private CloudTable GetTableReference()
+        {
+            var tableClient = CloudStorageAccount.Parse(_connectionStringFactory()).CreateCloudTableClient();
+            if (_readAccessGeoRedundant)
+            {
+                tableClient.DefaultRequestOptions.LocationMode = LocationMode.PrimaryThenSecondary;
+            }
+            return tableClient.GetTableReference(_tableName);
```
skofman1, line 352: As far as I understand, this makes a network call. So we are basically adding a network call to every error log. Do you have concerns about this? I'm sure there are ways to reduce it if needed.

joelverhagen: GetTableReference does not go to network. It just builds a model that takes the table client's credentials and the provided table name for future operations. You can see the code in ILSpy pretty easily:
image

joelverhagen (approved):

PR #8486, opened 2021-04-04
CloudAuditService preparing for secret refresh

agr:
Changed `CloudAuditService` to use `ICloudBlobClient` to access blob storage instead of using SDK classes directly, so it is ready for secret refresh changes. Also, DI registration was updated to resolve all dependencies within DI container instead of pulling in values from closures, which should help with secret refresh, too.

### DIFF
--- a/src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
+++ b/src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
@@ -21,17 +21,17 @@ namespace NuGetGallery.Auditing
     {
         public static readonly string DefaultContainerName = "auditing";
 
-        private CloudBlobContainer _auditContainer;
+        private Func<ICloudBlobContainer> _auditContainerFactory;
         private Func<Task<AuditActor>> _getOnBehalfOf;
 
-        public CloudAuditingService(string storageConnectionString, bool readAccessGeoRedundant, Func<Task<AuditActor>> getOnBehalfOf)
-            : this(GetContainer(storageConnectionString, readAccessGeoRedundant), getOnBehalfOf)
+        public CloudAuditingService(Func<ICloudBlobClient> cloudBlobClientFactory, Func<Task<AuditActor>> getOnBehalfOf)
+            : this(() => GetContainer(cloudBlobClientFactory), getOnBehalfOf)
         {
         }
 
-        public CloudAuditingService(CloudBlobContainer auditContainer, Func<Task<AuditActor>> getOnBehalfOf)
+        public CloudAuditingService(Func<ICloudBlobContainer> auditContainerFactory, Func<Task<AuditActor>> getOnBehalfOf)
         {
-            _auditContainer = auditContainer;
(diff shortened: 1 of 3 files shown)


src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
```diff
@@ -74,37 +75,24 @@ protected override async Task SaveAuditRecordAsync(string auditData, string reso
...
             return cloudBlobClient.GetContainerReference(DefaultContainerName);
         }
 
-        private static async Task WriteBlob(string auditData, string fullPath, CloudBlockBlob blob)
+        private static async Task WriteBlob(string auditData, string fullPath, ISimpleCloudBlob blob)
         {
             try
             {
-                var strm = await Task.Factory.FromAsync(
-                    (cb, s) => blob.BeginOpenWrite(
```
agr, line 100: This is the part I am most worried about. Not sure why it was written this way, the code is 8 years old. Best case, at the time of writing there were simply no appropriate async methods in the library.

Our wrappers don't have the `Begin*` and `End*` methods.

joelverhagen: These methods should be equivalent. In fact, if you look at the implementation of `OpenWriteAsync` in `WindowsAzure.Storage` it uses it's own APM wrapper to do a very similar thing. This _looks_ pretty safe.

src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
```diff
@@ -74,37 +75,24 @@ protected override async Task SaveAuditRecordAsync(string auditData, string reso
             {
                 // Create the container and try again,
                 // this time we let exceptions bubble out
-                await Task.Factory.FromAsync(
-                    (cb, s) => _auditContainer.BeginCreateIfNotExists(cb, s),
-                    ar => _auditContainer.EndCreateIfNotExists(ar),
-                    null);
+                var permissions = new BlobContainerPermissions { PublicAccess = BlobContainerPublicAccessType.Off };
```
joelverhagen, line 78: This should be the default behavior. Have you tested?

joelverhagen (approved):

src/NuGetGallery.Core/Auditing/CloudAuditingService.cs
agr: I didn't find an explicit statement in documentation about the default behavior, so went with specifying it explicitly myself. Looking at the decompiled storage SDK DLL, it seems you are correct. Will update.

agr: Well, our interface does not have a parameterless version:
https://github.com/NuGet/NuGetGallery/blob/be38792c3d3d327765113a32d82cd5b9b68cd1b0/src/NuGetGallery.Core/Services/ICloudBlobContainer.cs#L13

dannyjdev (approved):

Issue #8461, opened 2021-03-19
[README] Show more link jumps to the top of the page


joelverhagen:
### Describe the bug

Clicking the Show More link clears the URL fragment and goes to the top of the page.

### To Reproduce

1. Go to https://dev.nugettest.org/packages/ICanHazReadme/1.0.1#show-readme-container
2. Click Show More
3. Wait a second

### Expected behavior

The README should expand but the viewport should stay where it is.

Actual: it jumps up.

### Screenshots

![repro](https://user-images.githubusercontent.com/94054/111713555-21536d00-880d-11eb-91f1-90fa6225b28b.gif)

agr:
Package details page redesign should address it or make it unapplicable.

PR #8501, opened 2021-04-06
show more link jumps problem

lyndaidaii:

before changes: 
Clicking the Show More link clears the URL fragment and goes to the top of the page.

after changes: 
The README should expand but the viewport should stay where it is.

Addresses https://github.com/NuGet/NuGetGallery/issues/8461

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -67,6 +67,7 @@ $(function () {
         $("#show-readme-more").click(function (e) {
             showLess.collapse("toggle");
             e.preventDefault();
+            return false;
         });
         showLess.on('hide.bs.collapse', function (e) {
             e.stopPropagation();
@@ -74,7 +75,6 @@ $(function () {
         showLess.on('show.bs.collapse', function (e) {
             e.stopPropagation();
         });
-        return false;
     }
 
     // Configure expanders
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -74,6 +74,7 @@ $(function () {
         showLess.on('show.bs.collapse', function (e) {
             e.stopPropagation();
         });
+        return false;
     }
 
     // Configure expanders
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -517,7 +517,7 @@
                             @Html.Raw(Model.ReadMeHtml)
                         </div>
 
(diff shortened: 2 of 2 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -517,7 +517,7 @@
                             @Html.Raw(Model.ReadMeHtml)
                         </div>
 
-                        <a href="#" role="button" data-toggle="collapse" class="icon-link" data-target="#readme-more"
+                        <a href="#show-readme-container" role="button" data-toggle="collapse" class="icon-link" data-target="#readme-more"
```
joelverhagen, line 520: Hmm.. if this is the fix it seems that the click event is bubbling out and the default (navigation) behavior is remaining. 
https://stackoverflow.com/a/866587

This is just making the viewport jump to this link, right?

joelverhagen: I am confused by the GIF -- does the value in the address bar change?

lyndaidaii: when we click show-more, viewport jump to this link(`https://localhost/packages/embeddedReadmeTestApp/#show-readme-container` insteadof ` https://localhost/packages/embeddedReadmeTestApp/#`.

joelverhagen: This should not be necessary. This is a button (as evidenced by the `role` attribute not a traditional link. A navigation event should not even occur. See that Stack Overflow above. Is the event bubbling out of this click handler?
https://github.com/NuGet/NuGetGallery/blob/be38792c3d3d327765113a32d82cd5b9b68cd1b0/src/NuGetGallery/Scripts/gallery/page-display-package.js#L67-L70

I'm wondering if perhaps this behavior regressed when we updated jQuery...

Maybe we just need `return false` in that method? 

Ideally nothing happens in the address bar when you click this button. It's not really a link, per se.

lyndaidaii: Looks like `return false` works, nothing happens in the address bar when we click this button.

src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -74,6 +74,7 @@ $(function () {
         showLess.on('show.bs.collapse', function (e) {
             e.stopPropagation();
         });
+        return false;
```
joelverhagen, line 77: this is not where I would expect it. I would expect the `return false` in the event (click) handler. Maybe this very long SO answer will help clarify the situation for you?
https://stackoverflow.com/a/30473685
Let's sync offline if things are still confusing :)

loic-sharma (approved):

PR #8510, opened 2021-04-08
return backfill to order by date, batch size 1000

drewgillies:
Backfill job was ordered in such a way as to not work with cursoring. There was a filtering condition removed which also needed to be reinstated. I also upped the batching size to 1000 for the TFM backfill.

### DIFF
--- a/src/GalleryTools/Commands/BackfillCommand.cs
+++ b/src/GalleryTools/Commands/BackfillCommand.cs
@@ -125,7 +125,7 @@ namespace GalleryTools.Commands
 
                 packages = packages
                     .Where(p => p.Created < lastCreateTime && p.Created > startTime)
-                    .Where(p => p.PackageStatusKey == PackageStatus.Available || p.PackageStatusKey == PackageStatus.Validating)
+                    .Where(p => p.PackageStatusKey == PackageStatus.Available)
                     .OrderBy(p => p.Created);
                 if (LimitTo > 0)
                 {
--- a/src/GalleryTools/Commands/BackfillCommand.cs
+++ b/src/GalleryTools/Commands/BackfillCommand.cs
@@ -122,10 +122,11 @@ namespace GalleryTools.Commands
                 {
                     packages = packages.Include(QueryIncludes);
                 }
-                
+
                 packages = packages
                     .Where(p => p.Created < lastCreateTime && p.Created > startTime)
-                    .OrderBy(p => p.PackageRegistration.Id);
+                    .Where(p => p.PackageStatusKey == PackageStatus.Available || p.PackageStatusKey == PackageStatus.Validating)
+                    .OrderBy(p => p.Created);
                 if (LimitTo > 0)
                 {
                     packages = packages.Take(LimitTo);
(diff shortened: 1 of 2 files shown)


src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -122,10 +122,11 @@ public async Task Collect(SqlConnection connection, Uri serviceDiscoveryUri, Dat
                 {
                     packages = packages.Include(QueryIncludes);
                 }
-                
+
                 packages = packages
                     .Where(p => p.Created < lastCreateTime && p.Created > startTime)
-                    .OrderBy(p => p.PackageRegistration.Id);
+                    .Where(p => p.PackageStatusKey == PackageStatus.Available || p.PackageStatusKey == PackageStatus.Validating)
```
joelverhagen, line 128: `Validating` packages won't be available in v3-flatcontainer. Could we stick to `PackageStatus.Available`?

joelverhagen (changes requested):
Unsure about the validating packages addition.

src/GalleryTools/Commands/BackfillCommand.cs
drewgillies: @joelverhagen Great insight. Thank you. Note that this is for all backfill jobs--this is the parent class.

drewgillies:
>Unsure about the validating packages addition.
@joelverhagen note that this is removed code that I reintroduced. It could be an implementation detail (i.e. a flag) set in the subclasses if we like, defaulting to false.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
```diff
@@ -28,10 +31,34 @@ public static void Configure(CommandLineApplication config)
...
-            var supportedFrameworks = _packageService.GetSupportedFrameworks(nuspecReader, files);
+
+            // We wrap this int a try catch because fetching supported frameworks on existing packages can sometimes throw due to format errors
+            // (usually portable TFM formatting). In this case we'll clear the TFMs.
+            var supportedFrameworks = Enumerable.Empty<NuGetFramework>();
+            try
+            {
+                supportedFrameworks = _packageService.GetSupportedFrameworks(nuspecReader, files);
+            }
+            catch (Exception e)
```
joelverhagen, line 42: We can catch a more specific exception here, right? I think it's a specific one that is caused by the bad portables.

joelverhagen:
```suggestion
            catch
```
if we end up using a generic catch-all like the other places

joelverhagen (approved):
Try to catch a more specific exception if possible. But otherwise looks good.

src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs
drewgillies: It's a little generic--it's an `ArgumentException` which is why I opted for catching all. But I'm happy to make it explicit as exceptions not of this type will fall through and be logged, which could be useful.

Issue #8515, opened 2021-04-09
GitHubVulnerabilties2Db build break due to DI change


drewgillies:
Including the feature flag service in the constructor of the package service has broken dependency injection in GitHubVulnerabilities2Db.

drewgillies:
As this is a DI fix and currently deployed instances aren't affected, will close this without stepping through redeployments.

PR #8516, opened 2021-04-09
Fix break in githubvulnerabilities2db dependency injection

drewgillies:
Address: https://github.com/NuGet/NuGetGallery/issues/8515

Dependency injection was broken by the inclusion of the feature flag service in the package service constructor. Added a fake type for this and moved it and an additional fake into a fakes directory.

[this surfaced in BCDR analysis]

### DIFF
--- a/src/GitHubVulnerabilities2Db/Fakes/FakeContentService.cs
+++ b/src/GitHubVulnerabilities2Db/Fakes/FakeContentService.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using System.Web;
+using NuGetGallery;
+
+namespace GitHubVulnerabilities2Db.Fakes
+{
+    public class FakeContentService : IContentService
+    {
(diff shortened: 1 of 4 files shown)


joelverhagen (approved):

src/GitHubVulnerabilities2Db/Fakes/FakeContentService.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using System.Web;
+using NuGetGallery;
+
+namespace GitHubVulnerabilities2Db.Fakes
```
joelverhagen, line 9: Consider putting these fake implementations in one of the shared projects.

loic-sharma (approved):

lyndaidaii (approved):

src/GitHubVulnerabilities2Db/Fakes/FakeContentService.cs
drewgillies: Discussed offline. Will come back to that thought later if we see the need.

Issue #8490, opened 2021-04-05
[A11y]Tooltip provided for blue checkmark is not available for keyboard only user.


ryuyu:
Link: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1294859

## ​Environment Details:
Application Name: NuGet Service
#URL: https://www.nuget.org
Operating System: Windows 10
Browser Version: Chrome Version 85.0.4183.121 (Official Build) (64-bit),
Chromium Edge Version 87.0.637.0 (Official build) dev (64-bit)

## Repro Steps:​
 Hit the URL " https://www.nuget.org" and login with appropriate credentials.
Tab Navigate to "Sign in" control and hit enter.
 Tab Navigate to "Sign in with Microsoft" control and hit enter.
Search for a key term (say “json”) from search box.
"Packages returned for json" screen gets opened.
Tab navigate to Correct icon and observe its actionable using keyboard or not.
## Actual:​
Tooltip provided for icon (\/) besides Newtonsoft.Json link is not provided for keyboard user.
## Expected:​
Tooltip provided for icon (\/) besides Newtonsoft.Json link is should be provided for keyboard user.
## ​User Impact:​
If tooltip is not provided for keyboard user then keyboard user will miss some information that provided in the tooltip.

### Notes
The tooltip information that displays when hovering the blue checkmark on the search results page contains information that is not surface-able in any other way. We need to provide a text alternative to the hover text. UI will likely need to be discussed.

PR #8521, opened 2021-04-13
[a11y] Add popovers to verification checkmarks

drewgillies:
Addresses: https://github.com/NuGet/NuGetGallery/issues/8490

We need a tab stop on this checkmark (`tabindex='0'` means it won't mess with tab ordering), and an id that can be found by the popover code. This will respond to hover and focus events.

Screenreaders also read the text as written (tested on Narrator and NVDA).

Just realized how fast this popover flashes by in the below video. The text is the same as the current tooltip: 
```The ID prefix of this package has been reserved for one of the owners of this package by NuGet.org```

popoverwithouttitle

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-list-packages.js
+++ b/src/NuGetGallery/Scripts/gallery/page-list-packages.js
@@ -1,0 +1,22 @@
+$(function () {
+    'use strict';
+
+    for (var i = 0; i < listItemCount; i++) {
+        var id = "reserved-indicator-" + i;
+        configureCheckmarkImagePopover(id);
+    }
+
+    function configureCheckmarkImagePopover(id) {
+        var checkmarkImage = $('#' + id);
+        if (checkmarkImage.length == 1) {   // i.e. checkmark exists
+            checkmarkImage.popover({ trigger: 'hover focus' });
+            checkmarkImage.click(function() {
+                checkmarkImage.popover('show');
+                setTimeout(function() {
+                        checkmarkImage.popover('destroy');
+                    },
+                    1000);
+            });
+        }
+    }
+});
--- a/src/NuGetGallery/Scripts/gallery/page-list-packages.js
+++ b/src/NuGetGallery/Scripts/gallery/page-list-packages.js
@@ -1,14 +1,9 @@
-$(function () {
+$(function() {
     'use strict';
 
-    for (var i = 0; i < listItemCount; i++) {
-        var id = "reserved-indicator-" + i;
-        configureCheckmarkImagePopover(id);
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/NuGetGallery/Scripts/gallery/page-list-packages.js [hunk 1]
```diff
@@ -0,0 +1,22 @@
+$(function () {
+    'use strict';
+
+    for (var i = 0; i < listItemCount; i++) {
+        var id = "reserved-indicator-" + i;
+        configureCheckmarkImagePopover(id);
```
ryuyu, line 6: Instead of doing this here, we should leverage the jQuery selectors we have available. Something like this:
```
$(".reserved-indicator").each(
function() {  
	var checkmarkImage = $(this);
	checkmarkImage.popover({ trigger: 'hover focus' });
	checkmarkImage.click(function() {
		checkmarkImage.popover('show');
		setTimeout(function() {
			checkmarkImage.popover('destroy');
		},
		1000);
	});
})
````
This should remove the need to pass listItemCount as a global var and should also remove the need to set an id on each element.

The above should be able to replace the for loop and the configureCheckmarkImagePopover function below and I believe is functionally identical.

File: src/NuGetGallery/Views/Shared/ListPackages.cshtml [hunk 2]
```diff
@@ -244,5 +244,9 @@
         $('#reset-advanced-search').on('click', function () {
             location.href = '?q=@Uri.EscapeDataString(Model.SearchTerm)';
         });
+
+        var listItemCount = @Model.Items.Count;
```
ryuyu, line 248: We don't need this if we use jQuery class selectors instead. See other comment.

src/NuGetGallery/Views/Shared/_ListPackage.cshtml
```diff
@@ -50,7 +50,8 @@
                     <img class="reserved-indicator"
                          src="~/Content/gallery/img/reserved-indicator.svg"
                          @ViewHelpers.ImageFallback(Url.Absolute("~/Content/gallery/img/reserved-indicator-20x20.png"))
-                         title="@Strings.ReservedNamespace_ReservedIndicatorTooltip" />
+                         data-content="@Strings.ReservedNamespace_ReservedIndicatorTooltip"
+                         id=@("reserved-indicator-" + ViewData["ItemIndex"]) tabindex="0" />
```
ryuyu, line 54: id does not need to be set here if you use the class to enumerate the images that need the popover configured. See above.

ryuyu (approved):

Issue #8491, opened 2021-04-05
[A11y]Screenreader does not announce the column header keyword with cell values on Version History Table


ryuyu:
Link: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1294864

​## Environment Details:
Application Name: NuGet Service
#URL: https://www.nuget.org
Operating System: Windows 10
Browser Version: Chrome Version 85.0.4183.121 (Official Build) (64-bit),
Chromium Edge Version 87.0.637.0 (Official build) dev (64-bit)

## Repro Steps:​
 Hit the URL " https://www.nuget.org" and login with appropriate credentials.
Tab Navigate to "Sign in" control and hit enter.
 Tab Navigate to "Sign in with Microsoft" control and hit enter.
Search for a key term (say “json”) from search box.
Click on "Newtonsoft.json" control and hit enter.
Tab navigate to "Version History" control and hit enter.
Start screenreader and navigate using arrow key to the cell values and observe its announce the column header reference with cell values or not.
## Actual:​
Screenreader does not announce the column header keyword with cell values for the history version table.

This issue is observed throughout the application.

This issue is observed with both JAWS and NVDA.
## Expected:​
Screenreader should announce the column header keyword with cell values for the history version table.
## ​User Impact:​
Screenreader user will not get proper information regarding cell values belongs to which column header.

### Notes
It looks like the expectation is the for each row in the version history table, we read out "Version [Version] Downloads [Download Count] Last Updated [Last updated date]". In short, header must be included when reading out the table. The documentation for this standard appears to indicate that it is not applicable to websites, but is highly recommended. Let's evaluate how much work it is to fix this and address it if we can.

PR #8529, opened 2021-04-15
[A11y] Make announcers read the version history table data correctly.

dannyjdev:
Added `tabindex` to every `td` tag on the version history table so that the value can be announced using NVDA or other announcers.

Addresses https://github.com/NuGet/NuGetGallery/issues/8491

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -726,20 +726,20 @@
                 <table aria-label="Version History of @Model.Id" class="table borderless">
                     <thead>
                         <tr>
-                            <th>Version</th>
-                            <th>Downloads</th>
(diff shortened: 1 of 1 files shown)


loic-sharma:
Do we need to add `tabindex` on all our tables? For example, all our statistics tables here: https://www.nuget.org/stats

loic-sharma (approved):

lyndaidaii (approved):

ryuyu (approved):
As long as keyboard nav isn't wonky with this change, it looks good to me.

dannyjdev:
> Do we need to add `tabindex` on all our tables? For example, all our statistics tables here: https://www.nuget.org/stats

They already have the tabindex=0,

@ryuyu tabindex allows us to traverse the table cell by cell, before we could only traverse the Version and Status column because the values from those rows are links.

Issue #8474, opened 2021-03-31
Fetch withdrawn advisories from GitHub and include them in ingestion flows


drewgillies:
This should be as straightforward as adding the field here: https://github.com/NuGet/NuGetGallery/blob/main/src/GitHubVulnerabilities2Db/Collector/AdvisoryQueryBuilder.cs

and ensuring it's propagated to the model here: https://github.com/NuGet/NuGetGallery/blob/main/src/GitHubVulnerabilities2Db/GraphQL/SecurityAdvisory.cs#L17

We have code to process them here: https://github.com/NuGet/NuGetGallery/blob/main/src/NuGetGallery.Services/PackageManagement/PackageVulnerabilitiesManagementService.cs#L155-L162

There are already tests here, but behavior should be verified manually: https://github.com/NuGet/NuGetGallery/blob/main/tests/NuGetGallery.Facts/Services/PackageVulnerabilitiesManagementServiceFacts.cs#L82-L167

PR #8544, opened 2021-04-23
Ingest withdrawn data in GitHub GraphQL

drewgillies:
Addresses: https://github.com/NuGet/NuGetGallery/issues/8474

All of the logic exists in the service and UTs for processing the withdrawn field, but it needs to be added to the query.

Note that there are no withdrawn advisories for NuGet packages at this time, but we need this in place.

### DIFF
--- a/src/GitHubVulnerabilities2Db/Collector/AdvisoryQueryBuilder.cs
+++ b/src/GitHubVulnerabilities2Db/Collector/AdvisoryQueryBuilder.cs
@@ -23,6 +23,7 @@ namespace GitHubVulnerabilities2Db.Collector
         databaseId
         permalink
         severity
+        withdrawnAt
         updatedAt
         " + CreateVulnerabilitiesConnectionQuery() + @"
       }
--- a/src/GitHubVulnerabilities2Db/GraphQL/SecurityVulnerability.cs
+++ b/src/GitHubVulnerabilities2Db/GraphQL/SecurityVulnerability.cs
@@ -13,6 +13,7 @@ namespace GitHubVulnerabilities2Db.GraphQL
         public SecurityVulnerabilityPackage Package { get; set; }
         public string VulnerableVersionRange { get; set; }
         public SecurityVulnerabilityPackageVersion FirstPatchedVersion { get; set; }
+        public DateTimeOffset? WithdrawnAt { get; set; }
         public DateTimeOffset UpdatedAt { get; set; }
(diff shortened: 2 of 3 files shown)


joelverhagen (approved):

src/GitHubVulnerabilities2Db/Collector/AdvisoryQueryService.cs
```diff
@@ -77,7 +77,9 @@ private SecurityAdvisory MergeAdvisories(SecurityAdvisory advisory, SecurityAdvi
             nextAdvisory.Vulnerabilities.Edges = advisory.Vulnerabilities.Edges.Concat(
                 nextAdvisory.Vulnerabilities.Edges ?? Enumerable.Empty<Edge<SecurityVulnerability>>());
             // We are not querying the advisories feed at this time so we do not want to advance the advisory cursor past what it was originally.
+            // We also want to ensure a withdrawn advisory affects all vulnerabilities it comprises
             nextAdvisory.UpdatedAt = advisory.UpdatedAt;
+            nextAdvisory.WithdrawnAt = advisory.WithdrawnAt;
```
skofman1, line 82: Will this cause a vulnerability to stop showing up on PackageDetails page? V3 feed?

shishirx34 (approved):

src/GitHubVulnerabilities2Db/Collector/AdvisoryQueryService.cs
drewgillies: Yes, that's right. It's a deletion in the db.

skofman1: The deletion in the DB will flow to V3 only if the package edit time gets updated (as far as I know). Will it be?

agr (approved):

src/GitHubVulnerabilities2Db/Collector/AdvisoryQueryService.cs
drewgillies: Yes--withdrawn advisories (and I've just sanity-tested this) enter the _remove_ side of this block, and any package added to this collection will have its lastedited time updated: https://github.com/NuGet/NuGetGallery/blob/280f08350c9b995eb0cad14bd75ff6e5365c3564/src/NuGetGallery.Services/PackageManagement/PackageVulnerabilitiesManagementService.cs#L159

Issue #7662, opened 2019-10-29
In semver2, the metadata in the version after the "+" can be confusing


karann-msft:
for example:
![image](https://user-images.githubusercontent.com/16904420/67807605-5b727200-fa52-11e9-8f19-97924902e521.png)

Daniel Sterling said
>I noticed on nuget.org that the versioning for the Nuget.Packaging package that I need to use in our build system has a super strange versioning scheme.  Looks like a guid is getting squished in there as well.  Is that intentional? I was just confused as to the format I needed to use in my packageReference elements in my csproj.

@scottbommarito said
> It’s unclear if the typical usage of build metadata in SemVer2 is like this—information that is not very useful for consumers—or if people are using build metadata to add meaningful content to their version string. We never analyzed this so we display it by default. we don’t allow multiple versions with the same build metadata today. This is because of SemVer2’s spec, which specifically claims that two versions with the same build metadata have identical precedence: “Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.”

xavierdecoster:
NuGet does NOT allow multiple versions of the same package only differing in build metadata.
If you were to push a version without build metadata, you’d encounter a conflict when a same version without build metadata exists already, and vice versa.

The build metadata is not random, and in fact provided by the publisher, so **it has meaning to the publisher** (whereas the consumer typically only cares about the normalized version - unless that consumer is the publisher), but I agree we may do something about its presentation on the site.

Perhaps we should simply show the _normalized version_ in headers on the site, and display the full version somewhere else, as part of the other metadata (e.g. above or below description).

skofman1:
@karann-msft , @xavierdecoster , what is the action item here?

scottbommarito:
I'd like to add that I also analyzed the existing build metadatas on NuGet.org, and it seemed like almost every single one was either a guid or some other type of build number which has zero meaning to a consumer. I think that we should not display it so prominently because it seems to mostly provoke confusion, although we may still want to show it to authors.

xavierdecoster:
I agree with @scottbommarito: I think the build metadata is mainly useful to the publisher, less to the consumer.

The action item here would be to show it less prominently, but still make it discoverable.
I'd suggest to update the version in the title of the page to be normalized.
Perhaps we can show the full semantic version as part of the current SemVer warning?

[...]

PR #8554, opened 2021-04-27
Display version metadata only in an info box

rocklan:
Minor changes to the UI so that the version metadata only appears inside an info box on the view package page. Before:

image

after:

image

The issue has a screenshot of a warning but it seems to be that an info box might be more appropriate, as version metadata isn't really cause for an alert.

The build metadata is also removed from the search page. Before:

image

after:

image

Addresses https://github.com/NuGet/NuGetGallery/issues/7662

### DIFF
--- a/src/NuGetGallery/Helpers/ViewModelExtensions/PackageViewModelFactory.cs
+++ b/src/NuGetGallery/Helpers/ViewModelExtensions/PackageViewModelFactory.cs
@@ -34,6 +34,7 @@ namespace NuGetGallery
             }
 
             viewModel.FullVersion = NuGetVersionFormatter.ToFullString(package.Version);
+            viewModel.NormalizedVersion = NuGetVersionFormatter.Normalize(package.Version);
 
             viewModel.Id = package.Id;
             viewModel.Version = String.IsNullOrEmpty(package.NormalizedVersion) ?
--- a/src/NuGetGallery/Helpers/ViewModelExtensions/PackageViewModelFactory.cs
+++ b/src/NuGetGallery/Helpers/ViewModelExtensions/PackageViewModelFactory.cs
@@ -34,8 +34,6 @@ namespace NuGetGallery
             }
 
             viewModel.FullVersion = NuGetVersionFormatter.ToFullString(package.Version);
-            viewModel.NormalizedVersion = NuGetVersionFormatter.Normalize(package.Version);
-
             viewModel.Id = package.Id;
             viewModel.Version = String.IsNullOrEmpty(package.NormalizedVersion) ?
                 NuGetVersionFormatter.Normalize(package.Version) :
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -247,7 +247,7 @@
             <div class="package-title">
                 <h1>
                     @Html.BreakWord(Model.Id)
-                    <small class="text-nowrap">@Model.FullVersion</small>
+                    <small class="text-nowrap">@Model.NormalizedVersion</small>
 
                     @if (Model.IsVerified.HasValue && Model.IsVerified.Value)
                     {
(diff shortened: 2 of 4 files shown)


drewgillies:
@jcjiang I like this idea (thanks, @rocklan!) and it has come at a good time, as we have work going on atm about the package details page: https://twitter.com/jiachenjiang_/status/1385332921750360070?s=20. This may need to be considered in that context rather than taken as a stand-alone change, so it would be good to have your take on this.

src/NuGetGallery/Helpers/ViewModelExtensions/PackageViewModelFactory.cs
```diff
@@ -34,6 +34,7 @@ public PackageViewModel Setup(PackageViewModel viewModel, Package package)
             }
 
             viewModel.FullVersion = NuGetVersionFormatter.ToFullString(package.Version);
+            viewModel.NormalizedVersion = NuGetVersionFormatter.Normalize(package.Version);
```
joelverhagen, line 37: Does `viewModel.Version` not work, below (line 40-42)? From the code it seems to be normalized version.

src/NuGetGallery/Views/Shared/_ListPackage.cshtml
```diff
@@ -88,7 +88,7 @@
                 <li>
                     <span class="icon-text">
                         <i class="ms-Icon ms-Icon--Flag" aria-hidden="true"></i>
-                        Latest version: <span class="text-nowrap">@Model.FullVersion @(Model.Prerelease ? "(prerelease)" : "")</span>
+                        Latest version: <span class="text-nowrap">@Model.NormalizedVersion @(Model.Prerelease ? "(prerelease)" : "")</span>
```
joelverhagen, line 91: Interesting, it looks like search on PROD already hides build metadata:
image
https://www.nuget.org/packages?packagetype=&sortby=relevance&q=nuget&prerel=false

Perhaps this is related to the integration between gallery and our dedicated search service. Good for stand-alone deployments though.

joelverhagen:
I like this change, good polish prior to the redesign of the page. @jcjiang - could you take a look at this UX change?

jcjiang:
This looks great! Ship it!

src/NuGetGallery/Helpers/ViewModelExtensions/PackageViewModelFactory.cs
rocklan: You are correct, that does work. I've changed it over.

src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -760,7 +768,7 @@
                                 @VersionListDivider(rowCount, versionsExpanded)
                                 <tr class="@(packageVersion.IsCurrent(Model) ? "bg-info" : null)">
                                     <td tabindex="0">
-                                        <a href="@Url.Package(packageVersion)" title="@packageVersion.FullVersion">
+                                        <a href="@Url.Package(packageVersion)" title="@packageVersion.Version">
```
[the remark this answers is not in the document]
rocklan: Well I was just making it consistent so that the anchor text is the same as the title. I was also thinking having metadata just for screen readers would be very annoying. But also, hiding info like build metadata on a mouseover isn't a particularly good design as it's not discoverable by the user.. and I think the point of this PR is that build metadata isn't useful for 99% of users, so why have it there at all? 

In terms of accessibility, `title` is a strange choice, technically it should be aria-label (or even better, aria-labelledby).. but the contents of the anchor are the same as the title so there's no need for it at all.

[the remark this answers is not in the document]
joelverhagen: I've thought this over and I agree with your current impl.

joelverhagen (approved):

joelverhagen:
Thank you for the continued support in improving nuget.org, @rocklan!

Issue #8564, opened 2021-05-03
[Readme-bugbash] Display some content of readme with show more button


lyndaidaii:
Issue:  Without showing any of the Readme when we display 'show more' button.

reproduce: 
- go to page: https://www.nuget.org/packages/NuGetCleaner/#
- there is no readme content is showed with button 'show more'
![nugetcleaner.PNG](https://images.zenhubusercontent.com/5ea7662f8e9f163fb34aaf70/c25ed526-9ed5-4200-a9ca-d735bb531f72)
expected: 
show certain readme content with 'show more' button

PR #8570, opened 2021-05-05
show some content of readme if is too long

lyndaidaii:

we are using clamp.js to cut off the html element.  Here we are going to increase the number of lines to cut off. 

clamp(): this controls where and when to clamp the text of an element. 

useNativeClamp (Boolean). Enables or disables using the native -webkit-line-clamp in a supported browser (ie. Webkit). It defaults to true if you're using Webkit, but it can behave wonky sometimes so you can set it to false to use the JavaScript- based solution.

```
            var height = getMaxHeight(clampValue);
            if (height <= element.clientHeight) {
                clampedText = truncate(getLastChild(element), height);
            }
```

before: 

after : 
readme0

Addresses https://github.com/NuGet/NuGetGallery/issues/8564

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -62,7 +62,7 @@ $(function () {
             "Show more");
 
         var showLess = $("#readme-less");
-        $clamp(showLess[0], { clamp: 10, useNativeClamp: false });
+        $clamp(showLess[0], { clamp: 30, useNativeClamp: false});
 
         $("#show-readme-more").click(function (e) {
             showLess.collapse("toggle");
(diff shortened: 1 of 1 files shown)


joelverhagen:
In your screenshot it seems that the current behavior looks more like 0 or 1 lines. Does line count 10 not have the effect that we are expecting?

lyndaidaii:
> In your screenshot it seems that the current behavior looks more like 0 or 1 lines. Does line count 10 not have the effect that we are expecting?

increase the number from 10 to 30, it will increase the chunk height(number line we define * line height)
if chunk height is less then one html element height, it will show ...

loic-sharma (approved):

loic-sharma:
FYI our plan is to display the entire readme once we move to a tabbed design. In other words, it's OK if this fix isn't perfect for all readmes.

PR #8577, opened 2021-05-10
[non-Gallery] Add a non-locked monitoring cursor for inspection during backfill job

drewgillies:
This makes monitoring a large-scale collect/update job a lot easier to track--it writes a monitoring cursor (date only) file every time it updates the locked cursor file. Not being locked, the monitoring cursor can be inspected randomly to determine progress.

### DIFF
--- a/src/GalleryTools/Commands/BackfillCommand.cs
+++ b/src/GalleryTools/Commands/BackfillCommand.cs
@@ -36,6 +36,8 @@ namespace GalleryTools.Commands
 
         protected virtual string CursorFileName => "cursor.txt";
 
+        protected virtual string MonitoringCursorFileName => "monitoring_cursor.txt";
+
         protected virtual int CollectBatchSize => 10;
 
         protected virtual int UpdateBatchSize => 100;
@@ -204,6 +206,12 @@ namespace GalleryTools.Commands
                             logger.Log($"Writing {package.Created:u} to cursor...");
                             await cursor.Write(package.Created);
                             counter = 0;
+
+                            // Write a monitoring cursor (not locked) so for a large job we can inspect progress
+                            if (!string.IsNullOrEmpty(MonitoringCursorFileName))
+                            {
+                                File.WriteAllText(MonitoringCursorFileName, package.Created.ToString("G"));
+                            }
                         }
                     }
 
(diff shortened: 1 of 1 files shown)


src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -399,6 +407,12 @@ private async Task CommitBatch(EntitiesContext context, FileCursor cursor, Logge
             if (cursorTime.HasValue)
             {
                 await cursor.Write(cursorTime.Value);
+
+                // Write a monitoring cursor (not locked) so for a large job we can inspect progress
+                if (!string.IsNullOrEmpty(MonitoringCursorFileName))
+                {
+                    File.WriteAllText(MonitoringCursorFileName, cursorTime.Value.ToShortDateString());
```
joelverhagen, line 414: It'd probably be more future proof just to write the full cursor value. Who knows when we'll need that more precise value.

src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -399,6 +407,12 @@ private async Task CommitBatch(EntitiesContext context, FileCursor cursor, Logge
             if (cursorTime.HasValue)
             {
                 await cursor.Write(cursorTime.Value);
```
joelverhagen, line 409: Could we modify `FileCursor` to have this same behavior of not keeping the file handle open forever?

joelverhagen:
`Console.WriteLine` could be used to the similar affect, right?

src/GalleryTools/Commands/BackfillCommand.cs
drewgillies: I was wondering about this and thought this was a less intrusive fix. But I could take a look.

drewgillies:
> Console.WriteLine could be used to the similar affect, right?

@joelverhagen the problem there is when you no longer have access to the user session that initiated the process. You can only view the users tab in the task manager to see if it's still chewing CPU, and there's no available console.out.

joelverhagen (approved):
Looks good. If the file cursor change is too much work, don't worry about it.

src/GalleryTools/Commands/BackfillCommand.cs
```diff
@@ -275,6 +277,12 @@ public async Task Update(SqlConnection connection, string fileName)
                         {
                             await CommitBatch(context, cursor, logger, metadata.Created);
                             counter = 0;
+
+                            // Write a monitoring cursor (not locked) so for a large job we can inspect progress
+                            if (!string.IsNullOrEmpty(MonitoringCursorFileName))
```
zhhyu, line 282: I am a little curious about the reason to log the cursor time again here, because "CommitBatch" at line 278 has already logged it, is it?

drewgillies: Good catch, @zhhyu --this is a double-up. I've moved this piece of logic to the collect method.

Issue #8525, opened 2021-04-14
Prepare and launch survey banner on NuGet.org to launch Monday, April 16th 2021


chgill-MSFT:
We would like to begin the survey campaign on NuGet.org on Monday, April 19th. It should run for a full week until Monday, April 26th.

**Survey link**: https://aka.ms/NuGetSurvey
**Banner design:** Same golden banner at the top of the page as last time
**Campaign length:** April 19-26th (1 week)

The survey will be open for review and comments until Friday, April 16th. You can leave comments on the [survey preview](https://www.surveymonkey.com/r/Preview/?sm=7ErWSygoU5eU7aWLff3jWhSnO1hhHJ6MiL1vBObirMbUOzVFxjKRz4fUw_2BUuSIx3).

PR #8579, opened 2021-05-12
Add banner to all pages

lyndaidaii:

before: banner only show on the home page

after: banner show on all pages
package details page: 
displayp
Upload package page: 

Addresses https://github.com/NuGet/NuGetGallery/issues/8525

### DIFF
--- a/src/NuGetGallery/Controllers/AppController.cs
+++ b/src/NuGetGallery/Controllers/AppController.cs
@@ -149,7 +149,7 @@ namespace NuGetGallery
 
         private void SetBannerFeatureFlag()
         {
-            if (_featureFalgService.IsDisplayBannerEnabled())
+            if (_featureFlagService.IsDisplayBannerEnabled())
             {
                 ViewBag.DisplayBanner = true;
             }
--- a/src/NuGetGallery/Controllers/AppController.cs
+++ b/src/NuGetGallery/Controllers/AppController.cs
@@ -16,7 +16,7 @@ namespace NuGetGallery
         : Controller
     {
         private ICookieExpirationService _cookieExpirationService;
-        private IFeatureFlagService _featureFalgService;
+        private IFeatureFlagService _featureFlagService;
 
         private IOwinContext _overrideContext;
 
@@ -47,7 +47,7 @@ namespace NuGetGallery
         /// </summary>
         public void SetFeatureFlagsService(IFeatureFlagService featureFlagService)
         {
-            _featureFalgService = featureFlagService;
+            _featureFlagService = featureFlagService;
         }
 
         protected AppController()
@@ -55,7 +55,7 @@ namespace NuGetGallery
             NuGetContext = new NuGetContext(this);
 
             _cookieExpirationService = GetService<ICookieExpirationService>();
-            _featureFalgService = GetService<IFeatureFlagService>();
+            _featureFlagService = GetService<IFeatureFlagService>();
         }
 
         protected internal virtual T GetService<T>()
--- a/src/NuGetGallery/Controllers/AppController.cs
+++ b/src/NuGetGallery/Controllers/AppController.cs
@@ -42,6 +42,14 @@ namespace NuGetGallery
             _cookieExpirationService = cookieExpirationService;
         }
 
+        /// <summary>
+        /// This method is used for the unit test.
+        /// </summary>
+        public void SetFeatureFlagsService(IFeatureFlagService featureFlagService)
+        {
+            _featureFalgService = featureFlagService;
+        }
+
         protected AppController()
         {
             NuGetContext = new NuGetContext(this);
--- a/src/NuGetGallery/Controllers/AppController.cs
+++ b/src/NuGetGallery/Controllers/AppController.cs
@@ -16,6 +16,7 @@ namespace NuGetGallery
         : Controller
     {
         private ICookieExpirationService _cookieExpirationService;
+        private IFeatureFlagService _featureFalgService;
 
         private IOwinContext _overrideContext;
 
@@ -46,6 +47,7 @@ namespace NuGetGallery
             NuGetContext = new NuGetContext(this);
 
             _cookieExpirationService = GetService<ICookieExpirationService>();
+            _featureFalgService = GetService<IFeatureFlagService>();
         }
 
         protected internal virtual T GetService<T>()
@@ -117,6 +119,7 @@ namespace NuGetGallery
         protected override void OnActionExecuted(ActionExecutedContext filterContext)
         {
             SetCookieCompliance(filterContext);
+            SetBannerFeatureFlag();
 
             base.OnActionExecuted(filterContext);
         }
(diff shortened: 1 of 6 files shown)


joelverhagen:
Do we want it on all pages? How did we pick this specific set?

lyndaidaii:
> Do we want it on all pages? How did we pick this specific set?

yes, we want to it on all pages if this possible. the reason why pick those pages is it has high traffic to thoses pages, especially display package details pages. it has traffic from PMUI or google.

src/NuGetGallery/Views/Users/Packages.cshtml
```diff
@@ -3,6 +3,7 @@
 @{
     ViewBag.Title = "Manage My Package";
     ViewBag.Tab = "Packages";
+    ViewBag.DisplayBanner = Model.DisplayBannerFeatureEnabled;
```
joelverhagen, line 6: Can we do this in the Layout.cshtml + NuGetViewBase?

https://github.com/NuGet/NuGetGallery/blob/main/src/NuGetGallery/Views/Shared/Gallery/Layout.cshtml
https://github.com/NuGet/NuGetGallery/blob/main/src/NuGetGallery/Views/NuGetViewBase.cs

This would hit every page.

shishirx34:
Can we make this banner dismissable? It is very annoying to see it constantly for folks who don't care. It was tolerable when it was only on the home page but the user experience is terrible once we enable this on all pages. When asking for survey we should only ask it once when the page is opened after that allow it to be dismissable or it should auto-dismiss i.e. hide it after sometime if the user doesn't click on it, or say user logs in. Even if the user decides to respond to a survey why would we want to show the banner again.

shishirx34 (changes requested):
I would avoid putting this on all pages, or at least redesign the banner to be dismissible.

src/NuGetGallery/Views/Users/Packages.cshtml
lyndaidaii: I am not sure about this approach. if we include in NuGetViewBase, which means flag for banner will be in the configure file instead of feature flag json file. let me know if my thinking is in right track?

lyndaidaii:
> I would avoid putting this on all pages, or at least redesign the banner to be dismissible.

The reason why we want add to most of pages is because of we are getting small amount of customer response this time comparing last time.

jcjiang:
@shishirx34 Some more context to not letting the banner be dismissed - the main reason why we decided to sidestep that functionality (for now) is because 1) saving the dismissed state for future sessions from the users would require additional engineering work that given the 2) relatively short windows - usually a week or two - during which the survey is out, might not be worth it.

Completely agree though that as we have surveys that are out for longer or are more visually distracting, we should allow the banner to be dismissed and/or not show on all pages.

shishirx34:
@jcjiang - I disagree. Even the period of 1-2 weeks is too long of a period to showcase this anti-pattern. We get millions of views every week on the gallery. I've had personal experience with some users being annoyed when asked to give feedback for not using 2FA after login and this was a one time pop-up. I am certain lots of users will find this annoying. Being persistent and annoying at showing this banner to users just to get the feedback is not a good way to get the data, moreover, it disinclines the users to actually follow through. 

I remember that we had some form of survey (HATS?) where we had a "feedback" side rollout pop-up which was a much better way of getting feedback. This banner is pretty loud with a color scheme that doesn't really go with the site for continuous view and at the top of the page. Let's say a user does provide you with the feedback, we still continue to show this banner to that user which isn't a good user experience. I believe this banner was introduced for BLM support and now we are extending, probably misusing, it for survey purpose. 

We should either find a different way to get feedback, or at the very least invest engineering effort on making this dismissible so that it is reusable in the future. 

We should have a discussion offline if you still disagree.

/cc: @skofman1

lyndaidaii:
After discussion, we decide to display banner to all pages now due to reasons following:

- communication overhead with customers.  less confusion for customers since survey banner is always alive
- the reason why we are getting low response is not clear, we will investigate more, improve experience after getting more feedback from customers
- dismissble button requires some details info(how long banner will be dismiss), more time is required, which potentially make some annoying time for customers

We will add dismissible button for next NuGet.org survey to improve customer experience.

shishirx34 (approved):

src/NuGetGallery/Controllers/AppController.cs
```diff
@@ -16,6 +16,7 @@ public abstract partial class AppController
         : Controller
     {
         private ICookieExpirationService _cookieExpirationService;
+        private IFeatureFlagService _featureFalgService;
```
zhhyu, line 19: _featureFalgService -> _featureFlagService

tests/NuGetGallery.Facts/Controllers/AppControllerFacts.cs
```diff
@@ -71,7 +71,9 @@ public void SetViewBagGivenHttpContextItemsWithCanWriteAnalyticsCookies(bool can
             {
                 // Arrange
                 var cookieExpirationService = GetMock<ICookieExpirationService>();
+                var featureFlagsService = GetMock<IFeatureFlagService>();
                 cookieExpirationService.Setup(e => e.ExpireAnalyticsCookies(It.IsAny<HttpContextBase>()));
+                featureFlagsService.Setup(e => e.IsDisplayBannerEnabled()).Returns(false);
```
zhhyu, line 76: It's better to have a separate test for the banner flag.

tests/NuGetGallery.Facts/Controllers/AppControllerFacts.cs
```diff
@@ -120,6 +127,33 @@ public void SetViewBagGivenHttpContextItemsWithNullCanWriteAnalyticsCookies()
...
+
+                var controller = GetController<TestableAppController>();
+                controller.SetCookieExpirationService(cookieExpirationService.Object);
+                controller.SetFeatureFlagsService(featureFlagsService.Object);
+
+                // Act
+                InvokeOnActionExecutedMethod(controller.ControllerContext, httpContext.Object, controller);
+
+                // Assert
+                Assert.False(controller.ViewBag.CanWriteAnalyticsCookies);
```
joelverhagen, line 152: Why assert on analytics cookies? There is already a UT up there to do the same.

joelverhagen (approved):
Please limit UT assertions to only describe the UT name, i.e. the name of the "fact".
:shipit:

tests/NuGetGallery.Facts/Controllers/AppControllerFacts.cs
zhhyu, line 76: There is no need to verify the banner flag in these two tests for cookies, if we have another test to cover it.

tests/NuGetGallery.Facts/Controllers/AppControllerFacts.cs
```diff
@@ -120,6 +127,33 @@ public void SetViewBagGivenHttpContextItemsWithNullCanWriteAnalyticsCookies()
...
             }
 
+            [Theory]
+            [InlineData(true)]
+            [InlineData(false)]
+            public void SetViewBagWithBannerWhenIsDisplayBannerEnabled(bool isDisplayBannerEnabled) {
+                // Arrange
+                var cookieExpirationService = GetMock<ICookieExpirationService>();
+                var featureFlagsService = GetMock<IFeatureFlagService>();
+                cookieExpirationService.Setup(e => e.ExpireAnalyticsCookies(It.IsAny<HttpContextBase>()));
```
zhhyu, line 137: You can also remove this setup and verification because you test the banner flag here.

zhhyu (approved):

tests/NuGetGallery.Facts/Controllers/AppControllerFacts.cs
lyndaidaii: it's private method. I don't think we could remove the set up

Issue #8543, opened 2021-04-23
Add caching to vulnerabilitiesById dictionary in PackageService


drewgillies:
See Svetlana's comment here for additional details: https://github.com/NuGet/NuGetGallery/pull/8540#discussion_r618936844

PR #8580, opened 2021-05-13
Add caching to package details page vulnerabilities query

drewgillies:
Addresses: https://github.com/NuGet/NuGetGallery/issues/8543

Note that this change is against `dev` and will need to be rebased to `main` if we wish to deploy it as a hotfix.

This introduces the `PackageVulnerabilitiesCacheService` which is modelled on TypoSquatting caching. I've moved the query logic into the cache service, using an arbitrary 30-minute caching length per query result.

### DIFF
--- a/src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
+++ b/src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
@@ -409,6 +409,6 @@ namespace NuGetGallery
         /// Track how long it takes to populate the vulnerabilities cache
         /// </summary>
         /// <param name="milliseconds">Refresh duration for vulnerabilities cache</param>
-        void TrackVulnerabilitiesCacheRefreshDuration(long milliseconds);
+        void TrackVulnerabilitiesCacheRefreshDurationMs(long milliseconds);
     }
 }
--- a/src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
+++ b/src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
@@ -404,5 +404,11 @@ namespace NuGetGallery
             bool isAuthenticated,
             int testBucket,
             int testPercentage);
+
+        /// <summary>
+        /// Track how long it takes to populate the vulnerabilities cache
+        /// </summary>
+        /// <param name="milliseconds">Refresh duration for vulnerabilities cache</param>
+        void TrackVulnerabilitiesCacheRefreshDuration(long milliseconds);
     }
 }
--- a/src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
+++ b/src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
@@ -16,7 +16,7 @@ namespace NuGetGallery
 
         void TrackGetPackageRegistrationDownloadCountFailed(string packageId);
 
-        void TrackDownloadJsonRefreshDuration(long milliseconds);
+        void TrackDownloadJsonRefreshDuration(TimeSpan duration);
 
         void TrackDownloadCountDecreasedDuringRefresh(string packageId, string packageVersion, long oldCount, long newCount);
 
@@ -408,7 +408,7 @@ namespace NuGetGallery
         /// <summary>
         /// Track how long it takes to populate the vulnerabilities cache
         /// </summary>
-        /// <param name="milliseconds">Refresh duration for vulnerabilities cache</param>
-        void TrackVulnerabilitiesCacheRefreshDurationMs(long milliseconds);
+        /// <param name="duration">Refresh duration for vulnerabilities cache</param>
+        void TrackVulnerabilitiesCacheRefreshDuration(TimeSpan duration);
     }
 }
--- a/src/NuGetGallery.Services/Telemetry/TelemetryService.cs
+++ b/src/NuGetGallery.Services/Telemetry/TelemetryService.cs
@@ -261,9 +261,9 @@ namespace NuGetGallery
             });
         }
 
-        public void TrackDownloadJsonRefreshDuration(long milliseconds)
+        public void TrackDownloadJsonRefreshDuration(TimeSpan duration)
         {
-            TrackMetric(Events.DownloadJsonRefreshDuration, milliseconds, properties => { });
+            TrackMetric(Events.DownloadJsonRefreshDuration, duration.TotalMilliseconds, properties => { });
         }
 
         public void TrackDownloadCountDecreasedDuringRefresh(string packageId, string packageVersion, long oldCount, long newCount)
@@ -1104,9 +1104,9 @@ namespace NuGetGallery
             TrackMetric(Events.SymbolPackagePushDisconnect, 1, p => { });
         }
 
-        public void TrackVulnerabilitiesCacheRefreshDurationMs(long milliseconds)
+        public void TrackVulnerabilitiesCacheRefreshDuration(TimeSpan duration)
         {
-            TrackMetric(Events.VulnerabilitiesCacheRefreshDurationMs, milliseconds, properties => { });
+            TrackMetric(Events.VulnerabilitiesCacheRefreshDurationMs, duration.TotalMilliseconds, properties => { });
         }
 
         /// <summary>
--- a/src/NuGetGallery.Services/Telemetry/TelemetryService.cs
+++ b/src/NuGetGallery.Services/Telemetry/TelemetryService.cs
@@ -91,6 +91,7 @@ namespace NuGetGallery
             public const string ABTestEvaluated = "ABTestEvaluated";
             public const string PackagePushDisconnect = "PackagePushDisconnect";
             public const string SymbolPackagePushDisconnect = "SymbolPackagePushDisconnect";
+            public const string VulnerabilitiesCacheRefreshDuration = "VulnerabilitiesCacheRefreshDuration";
         }
 
         private readonly IDiagnosticsSource _diagnosticsSource;
@@ -1101,6 +1102,11 @@ namespace NuGetGallery
         public void TrackSymbolPackagePushDisconnectEvent()
         {
             TrackMetric(Events.SymbolPackagePushDisconnect, 1, p => { });
+        }
+
+        public void TrackVulnerabilitiesCacheRefreshDuration(long milliseconds)
+        {
+            TrackMetric(Events.VulnerabilitiesCacheRefreshDuration, milliseconds, properties => { });
         }
 
         /// <summary>
--- a/src/NuGetGallery.Services/Telemetry/TelemetryService.cs
+++ b/src/NuGetGallery.Services/Telemetry/TelemetryService.cs
@@ -91,7 +91,7 @@ namespace NuGetGallery
             public const string ABTestEvaluated = "ABTestEvaluated";
             public const string PackagePushDisconnect = "PackagePushDisconnect";
             public const string SymbolPackagePushDisconnect = "SymbolPackagePushDisconnect";
-            public const string VulnerabilitiesCacheRefreshDuration = "VulnerabilitiesCacheRefreshDuration";
+            public const string VulnerabilitiesCacheRefreshDurationMs = "VulnerabilitiesCacheRefreshDurationMs";
         }
 
         private readonly IDiagnosticsSource _diagnosticsSource;
@@ -1104,9 +1104,9 @@ namespace NuGetGallery
             TrackMetric(Events.SymbolPackagePushDisconnect, 1, p => { });
         }
 
-        public void TrackVulnerabilitiesCacheRefreshDuration(long milliseconds)
+        public void TrackVulnerabilitiesCacheRefreshDurationMs(long milliseconds)
         {
-            TrackMetric(Events.VulnerabilitiesCacheRefreshDuration, milliseconds, properties => { });
+            TrackMetric(Events.VulnerabilitiesCacheRefreshDurationMs, milliseconds, properties => { });
         }
 
         /// <summary>
--- a/src/NuGetGallery/App_Start/AppActivator.cs
+++ b/src/NuGetGallery/App_Start/AppActivator.cs
@@ -15,7 +15,7 @@ using System.Web.Optimization;
 using System.Web.Routing;
 using System.Web.UI;
 using Elmah;
-using Microsoft.WindowsAzure.ServiceRuntime;
+using Microsoft.Extensions.DependencyInjection;
 using NuGetGallery;
 using NuGetGallery.Configuration;
 using NuGetGallery.Diagnostics;
@@ -282,11 +282,12 @@ namespace NuGetGallery
             if (packageVulnerabilitiesCacheService != null)
             {
                 // Perform initial refresh + schedule new refreshes every 30 minutes
-                HostingEnvironment.QueueBackgroundWorkItem(_ => packageVulnerabilitiesCacheService.RefreshCache());
+                var serviceScopeFactory = DependencyResolver.Current.GetService<IServiceScopeFactory>();
+                HostingEnvironment.QueueBackgroundWorkItem(_ => packageVulnerabilitiesCacheService.RefreshCache(serviceScopeFactory));
                 if (configuration.StorageType == StorageType.AzureStorage)
                 {
                     jobs.Add(new PackageVulnerabilitiesCacheRefreshJob(TimeSpan.FromMinutes(30),
-                        packageVulnerabilitiesCacheService));
+                        packageVulnerabilitiesCacheService, serviceScopeFactory));
                 }
             }
 
--- a/src/NuGetGallery/App_Start/AppActivator.cs
+++ b/src/NuGetGallery/App_Start/AppActivator.cs
@@ -265,12 +265,28 @@ namespace NuGetGallery
 
             if (configuration.StorageType == StorageType.AzureStorage)
             {
-                var cloudDownloadCountService = DependencyResolver.Current.GetService<IDownloadCountService>() as CloudDownloadCountService;
+                var cloudDownloadCountService =
+                    DependencyResolver.Current.GetService<IDownloadCountService>() as CloudDownloadCountService;
                 if (cloudDownloadCountService != null)
                 {
                     // Perform initial refresh + schedule new refreshes every 15 minutes
                     HostingEnvironment.QueueBackgroundWorkItem(_ => cloudDownloadCountService.RefreshAsync());
-                    jobs.Add(new CloudDownloadCountServiceRefreshJob(TimeSpan.FromMinutes(15), cloudDownloadCountService));
+                    jobs.Add(new CloudDownloadCountServiceRefreshJob(TimeSpan.FromMinutes(15),
+                        cloudDownloadCountService));
+                }
+            }
+
+            var packageVulnerabilitiesCacheService =
+                    DependencyResolver.Current.GetService<IPackageVulnerabilitiesCacheService>() as
+                        PackageVulnerabilitiesCacheService;
+            if (packageVulnerabilitiesCacheService != null)
+            {
+                // Perform initial refresh + schedule new refreshes every 30 minutes
+                HostingEnvironment.QueueBackgroundWorkItem(_ => packageVulnerabilitiesCacheService.RefreshCache());
+                if (configuration.StorageType == StorageType.AzureStorage)
+                {
+                    jobs.Add(new PackageVulnerabilitiesCacheRefreshJob(TimeSpan.FromMinutes(30),
+                        packageVulnerabilitiesCacheService));
                 }
             }
 
--- a/src/NuGetGallery/App_Start/AppActivator.cs
+++ b/src/NuGetGallery/App_Start/AppActivator.cs
@@ -265,8 +265,7 @@ namespace NuGetGallery
 
             if (configuration.StorageType == StorageType.AzureStorage)
             {
-                var cloudDownloadCountService =
-                    DependencyResolver.Current.GetService<IDownloadCountService>() as CloudDownloadCountService;
+                var cloudDownloadCountService = DependencyResolver.Current.GetService<IDownloadCountService>() as CloudDownloadCountService;
                 if (cloudDownloadCountService != null)
                 {
                     // Perform initial refresh + schedule new refreshes every 15 minutes
@@ -276,20 +275,11 @@ namespace NuGetGallery
                 }
             }
 
-            var packageVulnerabilitiesCacheService =
-                    DependencyResolver.Current.GetService<IPackageVulnerabilitiesCacheService>() as
-                        PackageVulnerabilitiesCacheService;
-            if (packageVulnerabilitiesCacheService != null)
-            {
-                // Perform initial refresh + schedule new refreshes every 30 minutes
-                var serviceScopeFactory = DependencyResolver.Current.GetService<IServiceScopeFactory>();
-                HostingEnvironment.QueueBackgroundWorkItem(_ => packageVulnerabilitiesCacheService.RefreshCache(serviceScopeFactory));
-                if (configuration.StorageType == StorageType.AzureStorage)
-                {
-                    jobs.Add(new PackageVulnerabilitiesCacheRefreshJob(TimeSpan.FromMinutes(30),
-                        packageVulnerabilitiesCacheService, serviceScopeFactory));
-                }
-            }
+            // Perform initial refresh for vulnerabilities cache + schedule new refreshes every 30 minutes
+            var packageVulnerabilitiesCacheService = DependencyResolver.Current.GetService<IPackageVulnerabilitiesCacheService>();
+            var serviceScopeFactory = DependencyResolver.Current.GetService<IServiceScopeFactory>();
+            HostingEnvironment.QueueBackgroundWorkItem(_ => packageVulnerabilitiesCacheService.RefreshCache(serviceScopeFactory));
+            jobs.Add(new PackageVulnerabilitiesCacheRefreshJob(TimeSpan.FromMinutes(30), packageVulnerabilitiesCacheService, serviceScopeFactory));
 
             if (jobs.AnySafe())
             {
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -460,7 +460,7 @@ namespace NuGetGallery
                 .As<IPackageVulnerabilitiesManagementService>()
                 .InstancePerLifetimeScope();
 
-            builder.RegisterType<PackageVulnerabilitiesCacheService>()
+            builder.Register(c => new PackageVulnerabilitiesCacheService(c.Resolve<IPackageVulnerabilitiesManagementService>()))
                 .AsSelf()
                 .As<IPackageVulnerabilitiesCacheService>()
                 .SingleInstance();
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -460,6 +460,11 @@ namespace NuGetGallery
                 .As<IPackageVulnerabilitiesManagementService>()
                 .InstancePerLifetimeScope();
 
+            builder.RegisterType<PackageVulnerabilitiesCacheService>()
+                .AsSelf()
+                .As<IPackageVulnerabilitiesCacheService>()
+                .SingleInstance();
+
             services.AddHttpClient();
             services.AddScoped<IGravatarProxyService, GravatarProxyService>();
 
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -460,7 +460,7 @@ namespace NuGetGallery
                 .As<IPackageVulnerabilitiesManagementService>()
                 .InstancePerLifetimeScope();
 
-            builder.Register(c => new PackageVulnerabilitiesCacheService(c.Resolve<ITelemetryService>()))
+            builder.RegisterType<PackageVulnerabilitiesCacheService>()
                 .AsSelf()
                 .As<IPackageVulnerabilitiesCacheService>()
                 .SingleInstance();
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -460,9 +460,7 @@ namespace NuGetGallery
                 .As<IPackageVulnerabilitiesManagementService>()
                 .InstancePerLifetimeScope();
 
-            builder.Register(c =>
-                    new PackageVulnerabilitiesCacheService(c.Resolve<IPackageVulnerabilitiesManagementService>(),
-                        c.Resolve<ITelemetryService>()))
+            builder.Register(c => new PackageVulnerabilitiesCacheService(c.Resolve<ITelemetryService>()))
                 .AsSelf()
                 .As<IPackageVulnerabilitiesCacheService>()
                 .SingleInstance();
--- a/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
+++ b/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
@@ -460,7 +460,9 @@ namespace NuGetGallery
                 .As<IPackageVulnerabilitiesManagementService>()
                 .InstancePerLifetimeScope();
 
-            builder.Register(c => new PackageVulnerabilitiesCacheService(c.Resolve<IPackageVulnerabilitiesManagementService>()))
+            builder.Register(c =>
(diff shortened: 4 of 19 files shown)


joelverhagen:
Could we do caching at a higher level and use the same caching method used elsewhere?
`HttpContext.Cache`
e.g.
https://github.com/NuGet/NuGetGallery/blob/f248171208356b66499810ddd502276784157422/src/NuGetGallery/Controllers/PackagesController.cs#L1025-L1055

This has expiration logic built in. There are other example usages of this API.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -460,6 +460,11 @@ protected override void Load(ContainerBuilder builder)
                 .As<IPackageVulnerabilitiesManagementService>()
                 .InstancePerLifetimeScope();
 
+            builder.RegisterType<PackageVulnerabilitiesCacheService>()
+                .AsSelf()
```
zhhyu, line 464: If we don't inject the implementation directly, we don't need to register it as self.

drewgillies:
> Could we do caching at a higher level and use the same caching method used elsewhere?
HttpContext.Cache

@joelverhagen is this preferable to the typosquatting approach (i.e. https://github.com/NuGet/NuGetGallery/blob/main/src/NuGetGallery/Services/TyposquattingCheckListCacheService.cs)? What are the merits of each?

joelverhagen:
`HttpContext.Cache`: automatically handles concurrency and cache invalidation time. Smaller code. Perhaps more consistent with other caching scenarios, but it's hard to discover all instances of the alternative.
`TyposquattingCheckListCacheService`: not dependent on ASP.NET/HTTP types. Probably easier to UT.

joelverhagen:
What do you think about those merits? Perhaps you can take a look at pros/cons as well? I think it's possible that `HttpContext.Cache` automatically evicts the cache in the event of memory pressure but I'm not 100% sure.

zhhyu:
I think the use case of "Typosquatting" cache may be different compared with this. The "Typosquatting" cache only contains one entry, which is a list of package names and only used when a package is pushed. For vulnerabilities, it is a very similar scenario as "Used By" info. So I agree that maybe using the highly-optimized caching component provided by the framework is better.

skofman1:
Another thought: we have a perf problem with Vulnerabilities when it comes to the 'Manage Packages' page as well (https://github.com/nuget/nugetgallery/issues/8361). This might help to fix both. 
Btw @drewgillies , should we re-open #8361 ? It's currently disabled due to perf, right?

drewgillies:
@skofman1 yes, that's right--I closed because we needed an entirely new approach. I've reopened it just now so we can investigate this idea. I'd be (pleasantly) surprised if caching is going to get us over the line on that one, as my understanding is that any regression at all is unacceptable given how poor the page's perf is already. And introducing a new table (let alone 3) regresses perf whichever way we slice it.
@joelverhagen I like the notion of http caching on the strength of what you're saying--UTs will still cover the logic, and if I can make the change more elegantly in code, I'm up for that.

skofman1:
I made the comment before going over the code. My initial thinking was that we can cache in memory all the vulnerability information, and load it from DB on start-up. What do you think?

drewgillies:
That's an interesting idea, @skofman1. I'll look into it.

drewgillies:
Interested in thoughts on this approach @joelverhagen @skofman1 @zhhyu --based on @skofman's idea, I'm loading the cache on startup, and have chosen an arbitrary 1 day for cache invalidation. This approach could then be extended to the manage packages page if we wish to experiment. More testing required but there is UT coverage here and the cache loads nicely. I went for the complex linq statement to get the best perf out of EF, so commented heavily.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
drewgillies: I'm now registering a specific instance in order to get the cache loaded on startup.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,83 @@
...
+                    ikv => (cachedAt: DateTime.Now, 
+                            vulnerabilitiesById: ikv.GroupBy(kv => kv.PackageKey, kv => kv.Vulnerability)
+            // - build the inner dictionaries, all under the same <id>, each keyed by <package key>
+                            .ToDictionary(kv => kv.Key,
+                                kv => kv.ToList().AsReadOnly() as IReadOnlyList<PackageVulnerability>)));
+        }
+
+        private bool ShouldCachedValueBeUpdated(string id) => !vulnerabilitiesByIdCache.ContainsKey(id) ||
+                                                              vulnerabilitiesByIdCache[id].cachedAt
+                                                                  .AddMinutes(CachingLimitMinutes) < DateTime.Now;
```
joelverhagen, line 81: nit: `DateTimeOffset.UtcNow` instead of `DateTime.Now` for clarity

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,83 @@
...
+    public class PackageVulnerabilitiesCacheService : IPackageVulnerabilitiesCacheService
+    {
+        private const int CachingLimitMinutes = 1440; // We could make this 1 day instead (same value) but this is easier for spot testing the cache
+        private readonly object Locker = new object();
+        private IDictionary<string,
+            (DateTime cachedAt, Dictionary<int, IReadOnlyList<PackageVulnerability>> vulnerabilitiesById)> vulnerabilitiesByIdCache
+            = new Dictionary<string, (DateTime, Dictionary<int, IReadOnlyList<PackageVulnerability>>)>();
+
+        private readonly IPackageVulnerabilitiesManagementService _packageVulnerabilitiesManagementService;
+        public PackageVulnerabilitiesCacheService(IPackageVulnerabilitiesManagementService packageVulnerabilitiesManagementService)
```
joelverhagen, line 22: I'm wondering whether this dependency even makes much sense. The two new methods introduced on `IPackageVulnerabilitiesManagementService` are very small so it might be clearer to just depend on `IEntitiesContext` here.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            _packageVulnerabilitiesManagementService = packageVulnerabilitiesManagementService;
+            Initialize();
```
joelverhagen, line 25: We should not do SQL in the constructor

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,83 @@
...
+                            .ToDictionary(pv => pv.Key,
+                                pv => pv.ToList().AsReadOnly() as IReadOnlyList<PackageVulnerability>);
+
+                        vulnerabilitiesByIdCache[id] = (cachedAt: DateTime.Now, vulnerabilitiesById: packageKeyAndVulnerability);
+                    }
+                }
+            }
+
+            return vulnerabilitiesByIdCache[id].vulnerabilitiesById.Any()
+                ? new ReadOnlyDictionary<int, IReadOnlyList<PackageVulnerability>>(vulnerabilitiesByIdCache[id].vulnerabilitiesById)
```
joelverhagen, line 55: nit: you can just return a `Dictionary<TKey, TValue>` which itself implements `IReadOnlyDictionary<TKey, TValue>`. No need to construct a `ReadOnlyDictionary`.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,83 @@
...
+        public IReadOnlyDictionary<int, IReadOnlyList<PackageVulnerability>> GetVulnerabilitiesById(string id)
+        {
+            if (string.IsNullOrEmpty(id))
+            {
+                throw new ArgumentException("Must have a value.", nameof(id));
+            }
+
+            if (ShouldCachedValueBeUpdated(id))
+            {
+                lock (Locker)
```
joelverhagen, line 37: This is a scary lock. My understanding is that every package details page request to a single node will contend on this lock.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,83 @@
...
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Data.Entity;
+using System.Linq;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public class PackageVulnerabilitiesCacheService : IPackageVulnerabilitiesCacheService
```
joelverhagen, line 13: So the idea is to load all the data once at start-up and then incrementally keep individual package IDs up to date as users request for them? What about a background refresh of the data set so that "foreground" package details requests never pay the SQL cost for vulnerability info?

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
joelverhagen: Sorry, I missed the `ShouldCachedValueBeUpdated` check first. Ignore.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        public IReadOnlyDictionary<int, IReadOnlyList<PackageVulnerability>> GetVulnerabilitiesById(string id)
+        {
+            if (string.IsNullOrEmpty(id))
+            {
+                throw new ArgumentException("Must have a value.", nameof(id));
+            }
+
+            if (ShouldCachedValueBeUpdated(id))
```
joelverhagen, line 35: This dictionary read could operate on the cache dictionary while it is being modified on line 49. Better to use a `ConcurrentDictionary`. See https://stackoverflow.com/a/24586924.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,83 @@
...
+                {
+                    if (ShouldCachedValueBeUpdated(id))
+                    {
+                        var packageKeyAndVulnerability = _packageVulnerabilitiesManagementService
+                            .GetVulnerableRangesById(id)
+                            .Include(x => x.Vulnerability)
+                            .SelectMany(x => x.Packages.Select(p => new {PackageKey = p.Key, x.Vulnerability}))
+                            .GroupBy(pv => pv.PackageKey, pv => pv.Vulnerability)
+                            .ToDictionary(pv => pv.Key,
+                                pv => pv.ToList().AsReadOnly() as IReadOnlyList<PackageVulnerability>);
```
joelverhagen, line 47: You don't need `AsReadOnly`. You can just cast to `IReadOnlyList`.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
joelverhagen: Note the background refresh flow for data is exactly what we use for download counts. All of the downloads are loading into a dictionary from a downloads.v1.json blob (analogous to the `GetAllVulnerableRanges` method) and it's regularly reloaded in the background. Requests read from the dictionary but never force the blob to be reloaded.

drewgillies:
> background refresh
@joelverhagen I love this idea. That's the missing piece. Thanks for all of your granular feedback above as well.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
drewgillies: The separation of responsibilities was initially related to DI scopes. `PackageVulnerabilitiesManagementService` is `InstancePerLifetimeScope`, and this is a `SingleInstance` (in order for caching to work), initially with this service passed to methods rather than stored in a field. I expect I'll return to this model if I'm unable to run the `Initialize` method from the constructor, which in that sense I like better.

The reason we have a caching service class which is separate from `PackageVulnerabilitiesService` (rather than just adding caching capabilities to it) is the same --`PackageVulnerabilitiesService` is `InstancePerLifetimeScope`.

drewgillies: Revising this.

drewgillies:
@joelverhagen @skofman1 @zhhyu I've implemented the background refresh on the cache, and I'll have it report load time telemetry like the downloads refresh does--it would be good to track this over time as the cache grows in size.

drewgillies:
@joelverhagen @skofman1 @zhhyu I've implemented the background refresh on the cache, and I'll have it report load time telemetry like the downloads refresh does--it would be good to track this over time as the cache grows in size.

src/NuGetGallery/App_Start/AppActivator.cs
```diff
@@ -265,12 +265,28 @@ private static void BackgroundJobsPostStart(IAppConfiguration configuration)
...
+            var packageVulnerabilitiesCacheService =
+                    DependencyResolver.Current.GetService<IPackageVulnerabilitiesCacheService>() as
+                        PackageVulnerabilitiesCacheService;
+            if (packageVulnerabilitiesCacheService != null)
+            {
+                // Perform initial refresh + schedule new refreshes every 30 minutes
+                HostingEnvironment.QueueBackgroundWorkItem(_ => packageVulnerabilitiesCacheService.RefreshCache());
+                if (configuration.StorageType == StorageType.AzureStorage)
+                {
+                    jobs.Add(new PackageVulnerabilitiesCacheRefreshJob(TimeSpan.FromMinutes(30),
```
skofman1, line 288: >30 [](http://example.com/codeflow?start=92&length=2)

what's the frequency of the GitHub job? Why 30 minutes?

src/NuGetGallery.Services/Telemetry/TelemetryService.cs
```diff
@@ -91,6 +91,7 @@ public class Events
             public const string ABTestEvaluated = "ABTestEvaluated";
             public const string PackagePushDisconnect = "PackagePushDisconnect";
             public const string SymbolPackagePushDisconnect = "SymbolPackagePushDisconnect";
+            public const string VulnerabilitiesCacheRefreshDuration = "VulnerabilitiesCacheRefreshDuration";
```
skofman1, line 94: add units to the metric name for debuggability. i.e.  VulnerabilitiesCacheRefreshDurationMs

src/NuGetGallery/App_Start/AppActivator.cs
drewgillies: This was arbitrary - it could easily be a day and work quite well. I thought that if we were to extend this cache's use to the manage packages page, a customer would want to see more "live" updates, so 30 minutes seemed like the happy medium.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,94 @@
...
+            Dictionary<int, IReadOnlyList<PackageVulnerability>>> _vulnerabilitiesByIdCache
+            = new ConcurrentDictionary<string, Dictionary<int, IReadOnlyList<PackageVulnerability>>>();
+        private readonly object _refreshLock = new object();
+        private bool _isRefreshing;
+
+        private readonly IPackageVulnerabilitiesManagementService _packageVulnerabilitiesManagementService;
+        private readonly ITelemetryService _telemetryService;
+
+        public PackageVulnerabilitiesCacheService(
+            IPackageVulnerabilitiesManagementService packageVulnerabilitiesManagementService,
```
skofman1, line 26: >packageVulnerabilitiesManagementService [](http://example.com/codeflow?start=53&length=39)

if I understand correctly, the IPackageVulnerabilitiesManagementService will be created only once, which is not good because it has a dependency on EntityContext (DB connection), meaning we need a new one with every request.

src/NuGetGallery.Services/Telemetry/TelemetryService.cs
drewgillies: Sure thing--I was modelling it off `TrackDownloadJsonRefreshDuration` - I can change that too if you like. Also, do you want the string value changed or just the name of the const?

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
drewgillies: The lifecycle pieces of this are the real trick (and the reason the cache service exists at all)--I may have the Job pass the management service each time to get around this.

src/NuGetGallery.Services/Telemetry/TelemetryService.cs
skofman1: Add to both. When you look at telemetry it will be super useful to know the units without going to the code. Regarding TrackDownloadJsonRefreshDuration , it will be nice to change that, assuming no one is depending on the event name for dashboards/monitoring.

drewgillies: Will do! I might hang off on ...DownloadJson... then, but yes, will change mine.

src/NuGetGallery/App_Start/AppActivator.cs
```diff
@@ -265,12 +265,28 @@ private static void BackgroundJobsPostStart(IAppConfiguration configuration)
...
+                }
+            }
+
+            var packageVulnerabilitiesCacheService =
+                    DependencyResolver.Current.GetService<IPackageVulnerabilitiesCacheService>() as
+                        PackageVulnerabilitiesCacheService;
+            if (packageVulnerabilitiesCacheService != null)
+            {
+                // Perform initial refresh + schedule new refreshes every 30 minutes
+                HostingEnvironment.QueueBackgroundWorkItem(_ => packageVulnerabilitiesCacheService.RefreshCache());
```
zhhyu, line 285: I don't know too much about what happens for this behind the scene. This seems running as a background thread, and the query may take some time to run. Will it have a race condition between the application and the query? For example, the application starts accepting requests while the query has not finished running. That may not mean customer impacts because we have functional/e2e tests but if we have some tests for vulnerabilities, the test may fail because the result has not been loaded yet.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,94 @@
...
+                        .ToDictionary(ikv => ikv.Key,
+                            ikv =>
+                                ikv.GroupBy(kv => kv.PackageKey, kv => kv.Vulnerability)
+                                    // - build the inner dictionaries, all under the same <id>, each keyed by <package key>
+                                    .ToDictionary(kv => kv.Key,
+                                        kv => kv.ToList().AsReadOnly() as IReadOnlyList<PackageVulnerability>));
+                    
+                    stopwatch.Stop();
+                    
+                    _telemetryService.TrackVulnerabilitiesCacheRefreshDuration(stopwatch.ElapsedMilliseconds);
```
zhhyu, line 85: Besides logging the duration, I am thinking whether it will be better if we also log the status of the query, for example, possible exceptions thrown from the DB query (Timeout, etc..).

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            if (shouldRefresh)
+            {
+                try
+                {
+                    var stopwatch = Stopwatch.StartNew();
+
+                    // We need to build a dictionary of dictionaries. Breaking it down:
+                    // - this give us a list of all vulnerable package version ranges
+                    _vulnerabilitiesByIdCache = _packageVulnerabilitiesManagementService.GetAllVulnerableRanges()
```
zhhyu, line 68: This looks like a heavy query for DB. I wonder it may cause glitches for DB performance. We may need some stress tests to understand the behavior.

Besides, my understanding is that each time we reload everything from DB again. I am not sure whether we can only load the difference because it sounds like most of existing vulnerabilities should not need to be updated with a high frequency, but the effort may not be worth if the performance is ok.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Linq;
+using NuGet.Services.Entities;
+
+namespace NuGetGallery
+{
+    public class PackageVulnerabilitiesCacheService : IPackageVulnerabilitiesCacheService
+    {
+        private IDictionary<string,
+            Dictionary<int, IReadOnlyList<PackageVulnerability>>> _vulnerabilitiesByIdCache
+            = new ConcurrentDictionary<string, Dictionary<int, IReadOnlyList<PackageVulnerability>>>();
```
zhhyu, line 18: For discussions: I can see that the only write action now is to update the reference to another new dictionary when we refresh the cache. Do we still need to use "ConcurrenctDictionary"? We don't update the granularity such as some entries in the dictionary.

src/NuGetGallery/App_Start/AppActivator.cs
drewgillies: If you look in the `RefreshCache` you'll see that this is catered for in the `_isRefreshing` sync'd field. The first refresh is run in `AppActivator` so we should have data available (and I don't believe any E2E tests cover vulnerabilities data).

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
drewgillies: I'll trace exceptions in the query code. Good idea.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
drewgillies: I think it would be good to track telemetry on this and see how it performs generally, before introducing performance tweaks. It's only heavy in post-query transformation--the query itself loads records from the `VulnerablePackageVersionRange` table and related `Package` records--this should be relatively lightweight. But we should watch telemetry. If it turns out we need to find ways to speed up load, we could start looking at timestamping vulnerable range records, but of course, the table and all of the records would need to be loaded in order to test for changes (but it would reduce `Packages` reads.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
drewgillies: Good call--I'm reverting this to a `Dictionary`.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
drewgillies: Another option for reducing DTUs is, of course, increasing the refresh interval from 30 minutes to something far longer.

drewgillies:
@skofman1 I've worked with @joelverhagen and @agr to land on a more elegant solution here--we're caching a scope factory in the `PackageVulnerabilitiesCacheService` which will spin up an entirely new scope every time we refresh, and generate a new `EntitiesContext` instance each time we do. I believe I've addressed all feedback now (including @zhhyu's). Interested in how close you feel we are to an acceptable design. I'll be running some more tests on it before merging.

src/NuGetGallery/App_Start/AppActivator.cs
skofman1: if the GitHub job runs once a day to get updates, there's no point in having this cache updated every 30 mins. If the job runs frequently (say every 10 minutes), 30 minutes definetly makes sense.

src/NuGetGallery/App_Start/AppActivator.cs
```diff
@@ -265,12 +265,29 @@ private static void BackgroundJobsPostStart(IAppConfiguration configuration)
...
-                    jobs.Add(new CloudDownloadCountServiceRefreshJob(TimeSpan.FromMinutes(15), cloudDownloadCountService));
+                    jobs.Add(new CloudDownloadCountServiceRefreshJob(TimeSpan.FromMinutes(15),
+                        cloudDownloadCountService));
+                }
+            }
+
+            var packageVulnerabilitiesCacheService =
+                    DependencyResolver.Current.GetService<IPackageVulnerabilitiesCacheService>() as
+                        PackageVulnerabilitiesCacheService;
+            if (packageVulnerabilitiesCacheService != null)
```
skofman1, line 282: >packageVulnerabilitiesCacheService [](http://example.com/codeflow?start=16&length=34)

why would packageVulnerabilitiesCacheService ever be null? If it's null it means we have a bug and forgot to register it. In that case, it would be better to fail fast then to get into a state where the cache isn't refreshed.

src/NuGetGallery/Services/PackageVulnerabilitiesCacheService.cs
```diff
@@ -0,0 +1,107 @@
...
+            Dictionary<int, IReadOnlyList<PackageVulnerability>>> _vulnerabilitiesByIdCache
+            = new Dictionary<string, Dictionary<int, IReadOnlyList<PackageVulnerability>>>();
+        private readonly object _refreshLock = new object();
+        private bool _isRefreshing;
+
+        private readonly ITelemetryService _telemetryService;
+
+        public PackageVulnerabilitiesCacheService(ITelemetryService telemetryService)
+        {
+            _telemetryService = telemetryService;
```
skofman1, line 27: >telemetryService [](http://example.com/codeflow?start=32&length=16)

null check

skofman1 (approved):
Some comments. Otherwise looks good :)

src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
```diff
@@ -404,5 +404,11 @@ public interface ITelemetryService
             bool isAuthenticated,
             int testBucket,
             int testPercentage);
+
+        /// <summary>
+        /// Track how long it takes to populate the vulnerabilities cache
+        /// </summary>
+        /// <param name="milliseconds">Refresh duration for vulnerabilities cache</param>
+        void TrackVulnerabilitiesCacheRefreshDurationMs(long milliseconds);
```
joelverhagen, line 412: nit: pass `TimeSpan` here, don't mention "ms". The implementation below can make a "ms" specific metric emitted to AI but the interface can be cleaner than that. It's not a big deal though.

src/NuGetGallery/App_Start/AppActivator.cs
```diff
@@ -265,12 +265,29 @@ private static void BackgroundJobsPostStart(IAppConfiguration configuration)
 
             if (configuration.StorageType == StorageType.AzureStorage)
             {
-                var cloudDownloadCountService = DependencyResolver.Current.GetService<IDownloadCountService>() as CloudDownloadCountService;
+                var cloudDownloadCountService =
```
joelverhagen, line 268: nit: please minimize changes to unrelated lines.

src/NuGetGallery/App_Start/AppActivator.cs
joelverhagen: Why not make the necessary methods just available on the interface? I don't see why any cast is necessary.

src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
drewgillies: I'm so confused :) https://github.com/NuGet/NuGetGallery/pull/8580#discussion_r634786033

src/NuGetGallery/App_Start/AppActivator.cs
```diff
@@ -265,12 +265,29 @@ private static void BackgroundJobsPostStart(IAppConfiguration configuration)
...
+
+            var packageVulnerabilitiesCacheService =
+                    DependencyResolver.Current.GetService<IPackageVulnerabilitiesCacheService>() as
+                        PackageVulnerabilitiesCacheService;
+            if (packageVulnerabilitiesCacheService != null)
+            {
+                // Perform initial refresh + schedule new refreshes every 30 minutes
+                var serviceScopeFactory = DependencyResolver.Current.GetService<IServiceScopeFactory>();
+                HostingEnvironment.QueueBackgroundWorkItem(_ => packageVulnerabilitiesCacheService.RefreshCache(serviceScopeFactory));
+                if (configuration.StorageType == StorageType.AzureStorage)
```
joelverhagen, line 287: Why is this predicated on AzureStorage?

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
```diff
@@ -460,6 +460,11 @@ protected override void Load(ContainerBuilder builder)
                 .As<IPackageVulnerabilitiesManagementService>()
                 .InstancePerLifetimeScope();
 
+            builder.Register(c => new PackageVulnerabilitiesCacheService(c.Resolve<ITelemetryService>()))
```
joelverhagen, line 463: You can just registry the type, i.e. `RegistryType`. You don't need explicit ctor here right?

src/NuGetGallery/App_Start/AppActivator.cs
drewgillies: It didn't seem to make sense to have this running on local environments. If you think it should be all cases, I'm happy to remove it.

src/NuGetGallery/App_Start/AppActivator.cs
drewgillies: Wasn't sure of our policy on cleanup. Thanks.

src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
drewgillies: `TimeSpan` makes sense but is inconsistent with other telemetry methods. I'd like us to retrofit code if we pivot on practice, so other devs can follow the lead of what's there more efficiently. I'm happy to do this here but it may be good to consider this going forward.

src/NuGetGallery/App_Start/DefaultDependenciesModule.cs
drewgillies: Yes, I'm sure there was a reason I thought this was appropriate when it was more complex, but on reflection even then I was mistaken. Will clean up.

src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
drewgillies: Thinking further, `TimeSpan` here would mean removing `Ms` from the method name, which would conflict with @skofman1's comment as well. I think I'll leave it as-is for now? When we arrive at consensus on how we want our telemetry to function for duration tracking, perhaps we can go through and change them all as a separate change.

drewgillies: Just did a quick visual scan: `TrackDownloadJsonRefreshDuration` is the only outlier. Perhaps I could change both of them in this change?

drewgillies: (I would still leave `Ms` off the method name as it's no longer ms, and would make them both use `TimeSpan`)

src/NuGetGallery/App_Start/AppActivator.cs
drewgillies: Yep--good call. The same thing is being done for `cloudDownloadCountService` above, but it is redundant. I'll remove that.

src/NuGetGallery/Infrastructure/Jobs/PackageVulnerabilitiesCacheRefreshJob.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public class PackageVulnerabilitiesCacheRefreshJob : Job
+    {
+        private readonly PackageVulnerabilitiesCacheService _packageVulnerabilitiesCacheService;
+        private IServiceScopeFactory _serviceScopeFactory;
+
+        public PackageVulnerabilitiesCacheRefreshJob(TimeSpan interval, 
+            PackageVulnerabilitiesCacheService packageVulnerabilitiesCacheService,
+            IServiceScopeFactory serviceScopeFactory)
+            : base("", interval)
```
joelverhagen, line 19: please use explicit parameter name for ambiguous values like this empty string

src/NuGetGallery/Infrastructure/Jobs/PackageVulnerabilitiesCacheRefreshJob.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            _packageVulnerabilitiesCacheService = packageVulnerabilitiesCacheService;
+            _serviceScopeFactory = serviceScopeFactory;
+        }
+
+        public override Task Execute()
+        {
+            return new Task(() => _packageVulnerabilitiesCacheService.RefreshCache(_serviceScopeFactory));
```
joelverhagen, line 27: Does this need to be a new task? Who calls `Execute`? Is it already in a background task? If so, you can just synchronously execute `RefreshCache` and return `Task.CompletedTask`.

joelverhagen (approved):

Issue #8397, opened 2021-01-26
Reserved namespace error should mention a link to docs (aka.ms?)


joelverhagen:
### Describe the bug

When a user pushes a new package to NuGet.org and the ID is under  a reserved namespace, we show this error:
![image](https://user-images.githubusercontent.com/94054/105903905-572a6100-5fd5-11eb-8102-761e5ee19100.png)

This should mention a link to docs:
https://docs.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation

Requested by @anangaur

jcjiang:
For the text that shows in the banner -

This package ID has been reserved, per https://docs.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation. Please request access to upload from the prefix owner, or re-upload the package under a different prefix.

@skofman1 Let me know if there is anything else I can help with! :)

PR #8610, opened 2021-05-27
Add link for reserved namespace error during package upload 

lyndaidaii:

Add link to message when there is conflict of namespace during uploading package 

After changes: 
uploadWithName

Addresses https://github.com/NuGet/NuGetGallery/issues/8397

### DIFF
--- a/src/NuGetGallery/Controllers/PackagesController.cs
+++ b/src/NuGetGallery/Controllers/PackagesController.cs
@@ -518,7 +518,7 @@ namespace NuGetGallery
                 {
                     var version = nuspec.GetVersion().ToNormalizedString();
                     _telemetryService.TrackPackagePushNamespaceConflictEvent(id, version, currentUser, User.Identity);
-                    return Json(HttpStatusCode.Conflict, new[] { new JsonValidationMessage(Strings.UploadPackage_IdNamespaceConflict) });
+                    return Json(HttpStatusCode.Conflict, new[] { new JsonValidationMessage(new UploadPackageIdNamespaceConflict()) });
                 }
                 else if (result == PermissionsCheckResult.OwnerlessReservedNamespaceFailure)
                 {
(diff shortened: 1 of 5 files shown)


shishirx34:
I recommend including the link as part of the statement, something like:

... or re-upload the package with a different ID. Click [here](https://docs.microsoft.com/en-us/nuget/reference/id-prefix-reservation) to learn more about the Package ID prefix reservation.

tests/NuGetGallery.Facts/Controllers/PackagesControllerFacts.cs
```diff
@@ -6552,8 +6552,8 @@ public async Task WillShowTheViewWithErrorsWhenThePackageIdMatchesOwnedByOtherUs
                 var result = await controller.UploadPackage(fakeUploadedFile.Object) as JsonResult;
 
                 Assert.NotNull(result);
-                Assert.Equal(Strings.UploadPackage_IdNamespaceConflict, (result.Data as JsonValidationMessage[])[0].PlainTextMessage);
-                Assert.Null((result.Data as JsonValidationMessage[])[0].RawHtmlMessage);
+                Assert.Equal(Strings.UploadPackage_IdNamespaceConflictHtml, (result.Data as JsonValidationMessage[])[0].RawHtmlMessage);
+                Assert.Null((result.Data as JsonValidationMessage[])[0].PlainTextMessage) ;
```
joelverhagen, line 6556:
```suggestion
                Assert.Null((result.Data as JsonValidationMessage[])[0].PlainTextMessage);
```

joelverhagen (approved):

joelverhagen:
Here's an interesting perspective on using "here" for link text: https://www.smashingmagazine.com/2012/06/links-should-never-say-click-here/
I don't think it's a huge deal but perhaps we can consider a wording that avoids the focus on mouse mechanics. Users can navigate using keyboard -- no "click" per se or "tap" on their phone.

anangaur:
Can we say (with the full text as link):
"Learn more about Package Prefix Id reservation"

loic-sharma (approved):

PR #8612, opened 2021-05-28
Created gird for new page layout of Display Package Page for redesign 

sophiamfavro:
# Background
For the redesign of the display package page, the new layout follows this image 
image
# Solution
In order to achieve this I used Bootstraps grid feature to create this new layout. Here is the code needed to achieve this structure without any of the page content. Below it is the output that the code creates
image
image
I used this code as a guide for editing the DispayPackageV2.cshtml as to not impact the existing page. After completeing my changes, the page now looks like this 
image
The mobile view looks like this 
image
image
# Testing 
To check that my solution had worked I ran the page and 
- checked that there were no errors in the .js by looking at the console 
- collapsed the code so that only the row/column divs were shown and compared to the simplified structure code above 
- Inspected the page and hovered over the sections to see that all the content was in the correct column/row

# Recommend using hide white spaces changes when reviewing 
image

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
@@ -869,176 +869,193 @@
                         }
                     </div>
                 </div>
-                <div class="col-sm-3">
-                    <aside aria-label="Package details info" class="package-details-info">
-                        <h2>Info</h2>
-                        <ul class="list-unstyled ms-Icon-ul">
+                <aside aria-label="Package details info" class="col-sm-3 package-details-info">
+                    <h2>Info</h2>
+                    <ul class="list-unstyled ms-Icon-ul">
+                        <li>
+                            <i class="ms-Icon ms-Icon--History" aria-hidden="true"></i>
+                            last updated <span data-datetime="@Model.LastUpdated.ToString("O")">@Model.LastUpdated.ToNuGetShortDateString()</span>
+                        </li>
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
```diff
@@ -250,954 +250,960 @@
...
+                            <a href="#" role="button" data-toggle="collapse" class="icon-link" data-target="#hidden-versions"
+                               aria-expanded="@(versionsExpanded ? "true" : "false")" aria-controls="hidden-versions" id="show-hidden-versions">
+                                <i class="ms-Icon ms-Icon--Calculator@(versionsExpanded ? "Subtract" : "Addition")" aria-hidden="true"></i>
+                                <span>@(versionsExpanded ? "Show less" : "Show more")</span>
+                            </a>
+                        }
+                    </div>
+                </div>
+                <div class="col-sm-3">
+                    <aside aria-label="Package details info" class="package-details-info">
```
loic-sharma, line 873: I believe you should be able to merge lines 872 and 873 together (notice how the class contains both `package-details-info` and `col-sm-3`):

```suggestion
                <aside aria-label="Package details info" class="col-sm-3 package-details-info ">
```

loic-sharma: If this change does work, you will also need to update lines 1205 and 1206:

```diff
-            </div>
        </article>
```

loic-sharma (approved):
Looks great!

zhhyu (approved):

Issue #8618, opened 2021-06-04
Expand Package Deprecation Warnings by Default


fowl2:
I've used the the package page a bunch of times, but I just spent aaages trying to find the replacement for a package because I didn't notice the little disclosure triangle. 

Currently (collapsed by default):

![image](https://user-images.githubusercontent.com/12844147/120746638-fa225700-c542-11eb-89ad-ad270e89bd9f.png)

Proposal (expanded by default):

![image](https://user-images.githubusercontent.com/12844147/120746661-04445580-c543-11eb-914e-ae664a159334.png)

PR #8619, opened 2021-06-04
Expand deprecation warning by default

fowl2:
(untested)

Addresses https://github.com/NuGet/NuGetGallery/issues/8618

### DIFF
--- a/src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
+++ b/src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
@@ -2,10 +2,10 @@
 
 <div class="deprecation-container">
     <div class="icon-text alert alert-warning">
-        <div id="show-deprecation-content-container" class="deprecation-expander" tabindex="0" data-toggle="collapse" data-target="#deprecation-content-container" aria-expanded="false" aria-controls="deprecation-content-container" aria-labelledby="deprecation-container-label" role="button">
+        <div id="show-deprecation-content-container" class="deprecation-expander" tabindex="0" data-toggle="collapse" data-target="#deprecation-content-container" aria-expanded="true" aria-controls="deprecation-content-container" aria-labelledby="deprecation-container-label" role="button">
             <div class="deprecation-expander" role="button">
                 <div class="deprecation-expander-container">
-                    <i class="deprecation-expander-icon ms-Icon ms-Icon--Warning" aria-hidden="true"></i>
+                    <i class="deprecation-expander-icon ms-Icon ms-Icon--Warning" aria-hidden="false"></i>
 
                     <div id="deprecation-container-label" class="deprecation-expander-info-right">
                         @{
@@ -35,7 +35,7 @@
                 </div>
 
                 <div class="deprecation-expander-container">
-                    <i id="deprecation-expander-icon-right" class="deprecation-expander-icon deprecation-expander-info-right ms-Icon ms-Icon--ChevronDown" aria-hidden="true"></i>
+                    <i id="deprecation-expander-icon-right" class="deprecation-expander-icon deprecation-expander-info-right ms-Icon ms-Icon--ChevronDown" aria-hidden="false"></i>
                 </div>
             </div>
         </div>


joelverhagen:
Takes for taking the initiative to address a frustration you had! Let me loop in several folks who have been working on the package details page design recently.

@jcjiang, @sophiamfavro, @loic-sharma: what do you all think about this idea of making the deprecation details expanded by default? I think it makes sense since it's a relatively "strong" signal provided by the package owner. In this case, it's extra work the owner went through to urge folks to use an entirely different package, e.g. (from @fowl2's issue https://github.com/NuGet/NuGetGallery/issues/8618)
image

loic-sharma:
+1, I think this is a good idea! The Azure SDK team ran a user study and found that customers often miss the deprecation banner. I expect expanding by default would help make the banner more noticeable. /cc @chgill-MSFT

@anangaur Do you remember why we made this banner collapsed by default? Do you have any concerns? (I couldn't find any info in the [original spec](https://github.com/NuGet/Home/wiki/Deprecate-packages#nugetorg))

src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
```diff
@@ -2,10 +2,10 @@
 
 <div class="deprecation-container">
     <div class="icon-text alert alert-warning">
-        <div id="show-deprecation-content-container" class="deprecation-expander" tabindex="0" data-toggle="collapse" data-target="#deprecation-content-container" aria-expanded="false" aria-controls="deprecation-content-container" aria-labelledby="deprecation-container-label" role="button">
+        <div id="show-deprecation-content-container" class="deprecation-expander" tabindex="0" data-toggle="collapse" data-target="#deprecation-content-container" aria-expanded="true" aria-controls="deprecation-content-container" aria-labelledby="deprecation-container-label" role="button">
             <div class="deprecation-expander" role="button">
                 <div class="deprecation-expander-container">
-                    <i class="deprecation-expander-icon ms-Icon ms-Icon--Warning" aria-hidden="true"></i>
+                    <i class="deprecation-expander-icon ms-Icon ms-Icon--Warning" aria-hidden="false"></i>
```
loic-sharma, line 8: The `aria-hidden` attribute is a little confusing. It means "hey screen reader, don't mention this icon is here, it's just decoration". It does _not_ mean "hey screen reader this icon is currently not visible because the deprecation warning is collapsed". We should keep this `aria-hidden` attribute as `true` so that the screen reader properly ignores the icon:

```suggestion
                    <i class="deprecation-expander-icon ms-Icon ms-Icon--Warning" aria-hidden="true"></i>
```

We do this on all our icons. For example:

https://github.com/NuGet/NuGetGallery/blob/7c4d9b4716647c453cb6b9f0ffbeb98809e33ef0/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml#L204

loic-sharma: Same comment for line 38

fowl2: Done. Although, from your description I'm not sure modifying the `aria-expanded` is enough to expand by default. Perhaps someone more familiar with the codebase might be better author :)

skofman1:
@fowl2 , thanks for the contribution! Folks are OOF right now, so we will take a look at the proposal next week.

loic-sharma:
Hello, thank you for the contribution! It looks like this change needs a few minor tweaks:

1. On line 38, the `ms-Icon--ChevronDown` class should be replace with the `ms-Icon--ChevronUp` class
1. On line 38, the `aria-hidden` attribute should be set to `false`
1. On line 43, we need to add the `in` class to the div for the deprecation content container (see [Bootstrap 3's documentation](https://getbootstrap.com/docs/3.4/javascript/#collapse-usage))

We should be able to merge this pull request in once these items are updated!

Here's a patch file with the suggested changes:

```diff
diff --git a/src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml b/src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
index 33b4dc6a4..6e31409d7 100644
--- a/src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
+++ b/src/NuGetGallery/Views/Packages/_DisplayPackageDeprecation.cshtml
@@ -35,12 +35,12 @@
                 </div>
 
                 <div class="deprecation-expander-container">
-                    <i id="deprecation-expander-icon-right" class="deprecation-expander-icon deprecation-expander-info-right ms-Icon ms-Icon--ChevronDown" aria-hidden="true"></i>
+                    <i id="deprecation-expander-icon-right" class="deprecation-expander-icon deprecation-expander-info-right ms-Icon ms-Icon--ChevronUp" aria-hidden="false"></i>
                 </div>
             </div>
         </div>
 
-        <div class="deprecation-content-container collapse" id="deprecation-content-container">
+        <div class="deprecation-content-container collapse in" id="deprecation-content-container">
             @if (!string.IsNullOrEmpty(Model.AlternatePackageId))
             {
                 string alternateUrl;
```

PR #8620, opened 2021-06-07
Utilize vulnerabilities caching service for in manage packages page

drewgillies:
Addresses https://github.com/NuGet/NuGetGallery/issues/8361, with a far more performant approach (no direct database queries) than https://github.com/NuGet/NuGetGallery/pull/8435 which we've left behind a disabled feature flag for this reason. This change utilizes the cache set up here: https://github.com/NuGet/NuGetGallery/pull/8580.


skofman1:
How is perf looking?

drewgillies:
>How is perf looking?

@skofman1 I'd like to see it in INT before evaluating, but this change introduces no new queries, simple a dictionary lookup. Perf on the caching itself (a different change for Package Details) is looking fine in INT, with no discernable regression in the INT Package Details page as it has many queries already and is not under high load.

drewgillies:
@skofman1 I've done some preliminary perf investigations (comparing enabling/disabling the feature) in DEV and:
- for a large list (359 listed packages), I see a consistent regression of ~450ms, averaging over multiple attempts
- for a small list (5 packages), I actually saw small perf improvement of 35ms, which is unlikely to be related--so, a negligible perf difference.

This is different from the previous approach (direct SQL queries) where a consistent perf regression of ~1000ms applied to manage packages packages regardless of package list size.

tests/NuGetGallery.Facts/Services/PackageVulnerabilitiesServiceFacts.cs
```diff
@@ -1,20 +1,55 @@
...
+        private Package _packageVulnerable;
+        private Package _packageNotVulnerable;
+        private List<VulnerablePackageVersionRange> _versionRanges;
+
         [Fact]
         public void GetsVulnerableStatusOfPackage()
         {
             // Arrange
+            Setup();
+            var entitiesContext = new Mock<IEntitiesContext>();
```
zhhyu, line 27: For discussions: I am not very sure whether we need to set up and mock so many dependencies here. I can see that we only have one dependency: "IPackageVulnerabilitiesCacheService" for "PackageVulnerabilitiesService". Will it work if we just mock the method "_packageVulnerabilitiesCacheService.GetVulnerabilitiesById(id)"? This will help us keep the unit test efficient and clean.

drewgillies: Good call, Zhaohua--that simplifies the tests. I've included this change now.

tests/NuGetGallery.Facts/Services/PackageVulnerabilitiesServiceFacts.cs
```diff
@@ -26,34 +57,27 @@ public void GetsVulnerableStatusOfPackage()
...
 
-            var packageVulnerable = new Package
+            _packageVulnerable = new Package
             {
                 Key = 0,
                 PackageRegistration = registrationVulnerable,
                 Version = "1.0.0",
-                VulnerablePackageRanges = new List<VulnerablePackageVersionRange> {versionRangeModerate}
+                VulnerablePackageRanges = new List<VulnerablePackageVersionRange> { versionRangeModerate }
             };
```
zhhyu, line 72: It will be better if we have a test to cover the following case:
The package registration Id matches, but the package key can't be found in the dictionary retuned by "GetVulnerabilitiesById".

zhhyu:
Nice work! :+1:

zhhyu (approved):

Issue #8623, opened 2021-06-07
[a11y] Vulnerability alerts on package details page and manage packages page aren't accessible


drewgillies:
They need to be made into tab stops, and given a screen readable popover tooltip, like the reserved checkmarks changes here: https://github.com/NuGet/NuGetGallery/pull/8521

PR #8624, opened 2021-06-07
Add accessible popovers to vulnerability alerts

drewgillies:
Addresses https://github.com/NuGet/NuGetGallery/issues/8623.

image

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -171,17 +171,6 @@ $(function () {
         });
     }
 
-    // Add popover to warning icons
-    var popoverFunction = function () {
-        var checkmarkImage = $(this);
-        checkmarkImage.popover({ trigger: 'hover focus' });
-        checkmarkImage.click(function () {
-            checkmarkImage.popover('show');
-            setTimeout(function () {
-                    checkmarkImage.popover('destroy');
-                },
-                1000);
-        });
-    };
-    $(".package-warning-icon").each(popoverFunction);
+    $(".reserved-indicator").each(window.nuget.setPopovers);
+    $(".package-warning-icon").each(window.nuget.setPopovers);
 });
--- a/src/NuGetGallery/Scripts/gallery/page-display-package.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package.js
@@ -183,5 +183,5 @@ $(function () {
                 1000);
         });
     };
-    $(".ms-Icon--Warning").each(popoverFunction);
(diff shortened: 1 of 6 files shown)


src/NuGetGallery/Scripts/gallery/page-display-package.js
```diff
@@ -170,4 +170,18 @@ $(function () {
...
+    // Add popover to warning icons
+    var popoverFunction = function () {
+        var checkmarkImage = $(this);
+        checkmarkImage.popover({ trigger: 'hover focus' });
+        checkmarkImage.click(function () {
+            checkmarkImage.popover('show');
+            setTimeout(function () {
+                    checkmarkImage.popover('destroy');
+                },
+                1000);
```
shishirx34, line 183: Is 1 second enough for pop over dismissal? Is that a standard timeout value, do we use the same values across our site?

shishirx34 (approved):

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
```diff
@@ -413,6 +413,19 @@
...
+            var checkmarkImage = $(this);
+            checkmarkImage.popover({ trigger: 'hover focus' });
+            checkmarkImage.click(function () {
+                checkmarkImage.popover('show');
+                setTimeout(function () {
+                        checkmarkImage.popover('destroy');
+                    },
+                    1000);
+            });
+        };
```
shishirx34, line 427: Its been a while I've worked on UI changes, but don't we have a common.js or something that is loaded by all the scripts where you could add this function in and share it across different JS scripts? 

see `nuget.copyTextToClipboard` function that we use in similar way.

src/NuGetGallery/Scripts/gallery/page-display-package.js
drewgillies: We do--this is what we're using on the existing popover on reserved checkmarks on the package list.

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
drewgillies: I'll check it out--thanks!

src/NuGetGallery/Scripts/gallery/page-display-package.js
drewgillies: (note that the timer starts ticking on mouseleave)

src/NuGetGallery/Scripts/gallery/page-manage-packages.js
drewgillies: Yes, that worked well. I've included this change.

Issue #8590, opened 2021-05-20
[online safety] Extend the report form to comply with online safety standards


drewgillies:
Covered by spec developed here: https://github.com/NuGet/NuGetGallery/issues/8589

PR #8635, opened 2021-06-15
[online safety] Add safety categories to report abuse form

drewgillies:
Addresses https://github.com/NuGet/NuGetGallery/issues/8590
Spec: https://github.com/NuGet/Home/blob/dev/proposed/2021/NuGetOrgReportForm-SafetyChanges.md

- Adds online safety categories and associated text to form
- Adds tab indexes for a11y to other text (more a11y work is needed for this form, but this was a quick win for now)
- Removed contacted owner checkbox for safety categories - we don't want customers to feel they should do this
- All under a feature flag for now

[note that the gif below shows inconsistent "the" and "this" in dropdown. Since fixed.]

reportabuseform

### DIFF
--- a/src/NuGetGallery/Views/Packages/ReportAbuse.cshtml
+++ b/src/NuGetGallery/Views/Packages/ReportAbuse.cshtml
@@ -25,13 +25,13 @@
                 </text>
             )
 
-            <p>
+            <p tabindex="0">
                 Please provide a detailed abuse report with evidence to support your claim! We cannot delete packages without evidence that they exhibit malicious behavior.
             </p>
 
             @if (!Model.ConfirmedUser)
             {
-                <p>
+                <p tabindex="0">
                     Note: If this is your package and you would like to contact support, please
                     <a href="@Url.LogOn(returnUrl)">sign in.</a>
                 </p>
@@ -42,7 +42,7 @@
 
                 <div id="form-field-reason" class="form-group @Html.HasErrorFor(m => m.Reason)">
                     @Html.ShowLabelFor(m => m.Reason)
-                    <p>Please select the reason for contacting support about this package.</p>
+                    <p tabindex="0">Please select the reason for contacting support about this package.</p>
                     @Html.ShowEnumDropDownListFor(m => m.Reason, Model.ReasonChoices, "<Choose a Reason>")
                     @Html.ShowValidationMessagesFor(m => m.Reason)
                 </div>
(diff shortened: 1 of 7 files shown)


jcjiang:
I think these are great changes, especially conditionally rendering the description that contains details that complainants should include. Once we replace the security/DMCA forms with associated portals, we should have far less support requests sent in with insufficient information.

anangaur:
Should some of these be sub-topics i.e. another dropdown for specific selections? Seems a bit dense and overload in the single dropdown

src/NuGetGallery/Views/Packages/ReportAbuse.cshtml
```diff
@@ -112,6 +145,42 @@
                 $('#report-abuse-form').show();
             }
 
+            // We don't suggest the customer contact the owner in the case of safety violations
+            if (val === 'ChildSexualExploitationOrAbuse'
```
skofman1, line 149: If you move this logic to the Controller it can be unit tested.

drewgillies:
> Should some of these be sub-topics i.e. another dropdown for specific selections? Seems a bit dense and overload in the single dropdown
@anangaur I ran this idea by legal some time back, but our policy requires a more prominent position for the safety categories.

src/NuGetGallery/Views/Packages/ReportAbuse.cshtml
drewgillies: I'm not sure what we'd be testing. This is simply getting the value of the dropdown and turning messages on/off (and duplicates logic we're already using).

zhhyu:
How about a grouping of options:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup

drewgillies:
> How about a grouping of options:

@zhhyu That may be redundant as we have some other changes potentially coming down the pike, which may remove some of the pre-existing options (i.e. route customers to other locations for them and remove them from the dropdown). If this happens, most of what remains will be safety options. So, we could consider grouping options if these future changes determine that leaving the original ones untouched is better. For now I'd suggest leaving it as a simple list. I'm not sure if this function (https://github.com/NuGet/NuGetGallery/pull/8635/files#diff-aad522d53d7c37d2d65abfa950169d203d7eff0393d17330cda261623c1f9ee7R46) supports grouping, so it may not be a trivial change.

drewgillies:
@zhhyu @skofman1 I'm blocked on merging this--if my explanations above satisfy you, would you please approve?

zhhyu (approved):

zhhyu:
I suggest to get another approval from @jcjiang , but I see that this change is behind the feature flag, so it's fine to check in because it will not be alive until we are ready.

drewgillies:
@zhhyu I'm talking with @jcjiang about some tweaks, and I can float them as a new (and more trivial) PR. Thanks. Will merge this one.

PR #8656, opened 2021-06-23
Added warning banner for owners that description is in README tab

sophiamfavro:
I added a warning for packages without READMEs that only owners can see. It is meant to let them know that the package description is being shown in the README tab when there is no README. Along with telling them what is being shown it also links to docs on package READMEs on NuGet.org. 
Here is the link to the docs: https://docs.microsoft.com/nuget/nuget-org/package-readme-on-nuget-org

Here is what it looks like when logged in as the owner
image

Here is what it looks like when not an owner
image

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
@@ -545,7 +545,7 @@
                                 {
                                     if (Model.CanDisplayPrivateMetadata) 
                                     { 
-                                        @ViewHelpers.AlertWarning(@<text>The package description is shown below. To display a README instead, please follow these <a href='https://docs.microsoft.com/nuget/nuget-org/package-readme-on-nuget-org'>instructions</a>.</text>);
+                                        @ViewHelpers.AlertWarning(@<text>The package description is shown below. Please update your package to <a href='https://docs.microsoft.com/nuget/nuget-org/package-readme-on-nuget-org'>include a README</a>.</text>);
                                     }
 
                                     <p>@Html.PreFormattedText(Model.Description, Config)</p>
--- a/src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
@@ -545,7 +545,7 @@
                                 {
                                     if (Model.CanDisplayPrivateMetadata) 
                                     { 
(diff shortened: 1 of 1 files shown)


src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
```diff
@@ -543,6 +543,10 @@
                                 }
                                 else if (!Model.Deleted && !String.IsNullOrWhiteSpace(Model.Description))
                                 {
+                                    if (Model.CanDisplayPrivateMetadata) 
+                                    { 
+                                        @ViewHelpers.AlertWarning(@<text>The package description is shown below. To display a README instead, please follow instructions found <a href='https://docs.microsoft.com/nuget/nuget-org/package-readme-on-nuget-org'>here</a>.</text>);
+                                    }
```
loic-sharma, line 549: Don't forget to add a line after here

loic-sharma (approved):

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
sophiamfavro: done 👍

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
```diff
@@ -543,6 +543,11 @@
                                 }
                                 else if (!Model.Deleted && !String.IsNullOrWhiteSpace(Model.Description))
                                 {
+                                    if (Model.CanDisplayPrivateMetadata) 
+                                    { 
+                                        @ViewHelpers.AlertWarning(@<text>The package description is shown below. To display a README instead, please follow instructions found <a href='https://docs.microsoft.com/nuget/nuget-org/package-readme-on-nuget-org'>here</a>.</text>);
```
agr, line 548: How about changing the last sentence to
```
To display a README instead, please <a href="https://docs.microsoft.com/nuget/nuget-org/package-readme-on-nuget-org">modify</a> your package to include one.
```
? Or something similar?

drewgillies (approved):

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
sophiamfavro: Thank you for bringing this up! I talked to Jiachen and we came up with some new wording, so it should now make sense regardless of the presence of links.

lyndaidaii (approved):

PR #8673, opened 2021-07-09
Fix newline breaks in markdown

CyberAndrii:
Fixes https://github.com/NuGet/NuGetGallery/issues/8642

### DIFF
--- a/tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
+++ b/tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
@@ -79,6 +79,8 @@ namespace NuGetGallery
             [InlineData("- List", "<ul>\r\n<li>List</li>\r\n</ul>", false, false)]
             [InlineData("This is a paragraph\nwithout a break inside", "<p>This is a paragraph\nwithout a break inside</p>", false, true)]
             [InlineData("This is a paragraph\r\nwithout a break inside", "<p>This is a paragraph\r\nwithout a break inside</p>", false, false)]
+            [InlineData("Hello world\n\nI expect a break between these two lines", "<p>Hello world</p>\n<p>I expect a break between these two lines</p>", false, true)]
+            [InlineData("Hello world\n\nI expect a break between these two lines", "<p>Hello world</p>\r\n<p>I expect a break between these two lines</p>", false, false)]
             [InlineData("[text](http://www.test.com)", "<p><a href=\"http://www.test.com/\" rel=\"noopener noreferrer nofollow\">text</a></p>", false, true)]
             [InlineData("[text](http://www.test.com)", "<p><a href=\"http://www.test.com/\" rel=\"noopener noreferrer nofollow\">text</a></p>", false, false)]
             [InlineData("[text](javascript:alert('hi'))", "<p><a href=\"\" rel=\"noopener noreferrer nofollow\">text</a></p>", false, true)]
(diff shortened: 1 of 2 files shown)


joelverhagen:
Hey @CyberAndrii, thanks for taking the time to investigate this and provide a fix. @lyndaidaii will review this PR and support you in any PR comments that come up.

lyndaidaii:
Looks great to me. Already verified changes. Overall, it improved how existing readme display, made prettier. However, It might impact few existing readme. I think it's great contribution and encourage users to adapt best practice of markdown readme.  
Let me get another reviewer before we merge. @CyberAndrii  thanks for great work!

lyndaidaii (approved):

tests/NuGetGallery.Facts/Services/MarkdownServiceFacts.cs
```diff
@@ -77,6 +77,8 @@ public void EncodesHtmlInMarkdownWithAdaptiveHeader(string originalMd, string ex
             [InlineData("\ufeff# Heading with BOM", "<h2>Heading with BOM</h2>", false, false)]
             [InlineData("- List", "<ul>\n<li>List</li>\n</ul>", false, true)]
             [InlineData("- List", "<ul>\r\n<li>List</li>\r\n</ul>", false, false)]
+            [InlineData("This is a paragraph\nwithout a break inside", "<p>This is a paragraph\nwithout a break inside</p>", false, true)]
+            [InlineData("This is a paragraph\r\nwithout a break inside", "<p>This is a paragraph\r\nwithout a break inside</p>", false, false)]
```
loic-sharma, line 81: A break should still be inserted if I have a readme with multiple empty lines, right? For example:

```
Hello world

I expect a break between these two lines
```

Could you add a test for this case?

loic-sharma: @lyndaidaii Should all new markdown tests use markdig? The commonmark code is effectively dead now, right?

```suggestion
            [InlineData("This is a paragraph\r\nwithout a break inside", "<p>This is a paragraph\r\nwithout a break inside</p>", false, true)]
```

lyndaidaii: We are no longer use commonMark to render markdown file right now. this unit test case here, is because that in commonMark we encode first. that's why it appear \r\n instead for commonMark. it will nice to have a this case included as we did 
`            [InlineData("- List", "<ul>\n<li>List</li>\n</ul>", false, true)]
            [InlineData("- List", "<ul>\r\n<li>List</li>\r\n</ul>", false, false)]`

Also `This is a paragraph\r\nwithout a break inside", "<p>This is a paragraph\r\nwithout a break inside</p>` is same to 'This is a paragraph\nwithout a break inside', only \r\n is encode by windows.

lyndaidaii: @CyberAndrii Would you like to add unit test as Loic suggested above? thanks

CyberAndrii: Sorry, missed his message.

It converts to separate paragraphs: `<p>Hello world</p>\n<p>I expect a break between these two lines</p>`

Is this right?

lyndaidaii: Thanks for adding this unit test with multiple new line. we will treat as one line here. 

I added unit test for soft line break and hard line break. also include your multiple new line cases. 

`You can create line breaks in two ways.

Soft line break
Hardline break

Soft line breaks can be created by adding two spaces at the end of the line. This way markdown will render each line to be separate lines.
Hardline breaks can be created by inserting an empty line between each line.`

lyndaidaii: @loic-sharma, could you take a look at changes we added for unit test?

loic-sharma (approved):

PR #8677, opened 2021-07-12
[AccountDeleter]Adding stubbed definition for missing DI objects.

ryuyu:
Stubs out a minimal FeatureFlagService (used by PackageService) to allow DI to proceed successfully.

### DIFF
--- a/src/AccountDeleter/EmptyFeatureFlagService.cs
+++ b/src/AccountDeleter/EmptyFeatureFlagService.cs
@@ -1,0 +1,264 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.Entities;
+using System;
+using System.Collections.Generic;
+
+namespace NuGetGallery.AccountDeleter
+{
+    public class EmptyFeatureFlagService : IFeatureFlagService
+    {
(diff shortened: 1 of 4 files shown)


loic-sharma (approved):
Would it be possible to add a test that verifies the DI container doesn't have issues at runtime?

joelverhagen (approved):

src/AccountDeleter/EmptyFeatureFlagService.cs
```diff
@@ -0,0 +1,264 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.Entities;
+using System;
+using System.Collections.Generic;
```
zhhyu, line 6: nit: The order of dependencies?

zhhyu (approved):

ryuyu:
> 
> 
> Would it be possible to add a test that verifies the DI container doesn't have issues at runtime?

How would i go about doing this? Just have a test that attempts to hydrate the full real objects? That would require a full configuration for test, which I'm not against, but would be more work. Or if there is a simpler way, this sounds like a great idea.

dannyjdev (approved):

ryuyu:
I filed https://github.com/NuGet/NuGetGallery/issues/8700 for adding this test. I had a bit of difficulty immediately. I'll look into pulling it in when possible.

Issue #8602, opened 2021-05-24
[Redesign display package page] Implement new installation instructions in header


sophiamfavro:
Create the dropdown menu with the different package managers and the corresponding instructions.

loic-sharma:
@dannyjdev You can verify this by playing around with the display package page. Please try a regular .NET package, a .NET tool, and .NET template.

PR #8690, opened 2021-07-16
[Redesign] Update installation instructions

loic-sharma:
Addresses https://github.com/nuget/nugetgallery/issues/8602

## Screenshots

<details>
<summary>Default screenshot...</summary>

image

Package with long ID:

image

</details>

<details>
<summary>Installation instructions with banners below...</summary>

image

image

</details>

<details>
<summary>Multi-line installation instructions...</summary>

image

image

</details>

<details>
<summary>.NET Tools...</summary>

image

</details>

<details>
<summary> ⚠ .NET Template...</summary>

image

⚠ NOTE: There is only one package manager that supports .NET Templates today. The UI has a dropdown with only a single option.

</details>

<details>
<summary>Mobile...</summary>

FYI the installation instructions are sideways scrollable. You can swipe (or is it pan?) to see the text:

image

image

</details>

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-display-package-v2.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package-v2.js
@@ -48,51 +48,18 @@
         });
     }
 
-    // Configure package manager copy buttons
-    function configureCopyButton(id) {
-        var copyButton = $('#' + id + '-button');
-        copyButton.popover({ trigger: 'manual' });
-
-        copyButton.click(function () {
-            var text = $('#' + id + '-text').text().trim();
-            window.nuget.copyTextToClipboard(text, copyButton);
-            copyButton.popover('show');
-            //This is workaround for Narrator announce the status changes of copy button to achieve accessibility.
-            copyButton.attr('aria-pressed', 'true');
-            setTimeout(function () {
-                copyButton.popover('destroy');
-            }, 1000);
-            setTimeout(function () {
-                copyButton.attr('aria-pressed', 'false');
-            }, 1500);
-            window.nuget.sendMetric("CopyInstallCommand", 1, {
-                ButtonId: id,
-                PackageId: packageId,
-                PackageVersion: packageVersion
-            });
-        });
-    }
-
-    for (var i in packageManagers)
-    {
-        configureCopyButton(packageManagers[i]);
-    }
-
     var storage = window['localStorage'];
+    var installationKey = 'preferred_instruction';
+
+    var currentPackageManagerId = packageManagers[0];
+    var packageManagerSelector = $('.installation-instructions-dropdown');
+
     if (storage) {
-        // set preferred installation instruction tab
-        var installationKey = 'preferred_tab';
-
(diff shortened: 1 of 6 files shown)


joelverhagen:
Have you tested multi-line commands (`<PackageReference>` for dev dependency) or long commands (long ID/version)?

loic-sharma:
@joelverhagen Yup, I added a bunch of screenshots for a bunch of different scenarios including multi-line, long package IDs, mobile, etc... Let me know if I missed any fun cases!

joelverhagen:
These two lines look different length, i.e. it looks like the command text is not vertically centered. Intentional?
image

loic-sharma:
@joelverhagen Yup that's intentional. It is added when there is an alert below for spacing, but the spacing is equal when there is no alert below. /cc @jcjiang

joelverhagen (approved):

loic-sharma:
@jcjiang confirmed they approve of the extra padding on the installation instructions if they have a banner below.

zhhyu:
Looking now...

src/NuGetGallery/Scripts/gallery/page-display-package-v2.js
```diff
@@ -48,67 +47,90 @@
...
+
+        // Make sure we save the user's preferred package manager to localStorage.
+        if (storage) {
+            storage.setItem(packageManagerStorageKey, currentPackageManagerId);
+        }
+    });
+
+    // Used to switch installation instructions when a new package manager is selected 
+    function updatePackageManager(newPackageManagerId, updateSelector) {
+        var currentInstructions = $('#' + currentPackageManagerId + '-instructions');
```
zhhyu, line 92: I can see that "updateSelector" is used only when restoring the preferred package manager selection from the local Storage. So we can set "packageManagerSelector[0].value = preferredPackageManagerId;" after line 63 "updatePackageManager(preferredPackageManagerId, true);", and then there is no need to keep the argument "updateSelector".

I don't feel strong, but this may be more clean and easier to follow.

src/NuGetGallery/Scripts/gallery/page-display-package-v2.js
```diff
@@ -48,67 +47,90 @@
...
+        updatePackageManager(newPackageManagerId, false);
+
+        // Make sure we save the user's preferred package manager to localStorage.
+        if (storage) {
+            storage.setItem(packageManagerStorageKey, currentPackageManagerId);
+        }
+    });
+
+    // Used to switch installation instructions when a new package manager is selected 
+    function updatePackageManager(newPackageManagerId, updateSelector) {
```
zhhyu, line 91: I can see that updateSelector is mainly used when restoring the preferred package manager selection from the local storage. So we can set packageManagerSelector[0].value = preferredPackageManagerId; after line 63 updatePackageManager(preferredPackageManagerId, true);, and then there is no need to keep the argument updateSelector any more.

I don't feel strong, but this may be more clean and easier to follow.

PR #8691, opened 2021-07-17
Add admin panel for bulk unlisted or relisting packages

joelverhagen:
This uses the same logic as deprecation to allow many changes to packages to be done at once, efficiently.

image

- You can search for any number of IDs or versions, but upon submission they are processed in batches grouped by ID, where one "update is latest booleans" call is needed
- Packages that already have the matching listed status are still processed. There is no no-op. This is to allow a simple reflow of listed status.
- I tested on DEV with 1000, 2000, and 5000 versions of a single package ID. Worked great.


src/NuGetGallery.Services/Telemetry/ITelemetryService.cs
```diff
@@ -38,6 +38,8 @@ public interface ITelemetryService
 
         void TrackPackageListed(Package package);
 
+		void TrackPackagesUpdateListed(IReadOnlyList<Package> packages, bool listed);
```
loic-sharma, line 41: Looks like there's tabs here instead of spaces.

src/NuGetGallery/Areas/Admin/Controllers/AdminControllerBase.cs
```diff
@@ -1,12 +1,109 @@
...
 using NuGetGallery.Filters;
 
 namespace NuGetGallery.Areas.Admin.Controllers
 {
     [UIAuthorize(Roles="Admins")]
     public class AdminControllerBase : AppController
     {
+        internal List<Package> SearchForPackages(IPackageService packageService, string query)
+        {
+            // Search suports several options:
```
loic-sharma, line 19:
```suggestion
            // Search supports several options:
```

zhhyu:
Looking now...

src/NuGetGallery/Areas/Admin/Controllers/UpdateListedController.cs
```diff
@@ -0,0 +1,107 @@
...
+                    .Where(x => x.PackageStatusKey != PackageStatus.FailedValidation)
+                    .ToList();
+
+                packageCount += packages.Count;
+                noOpCount += normalizedVersions.Count - packages.Count;
+
+                if (packages.Any())
+                {
+                    packageRegistrationCount++;
+                    await _packageUpdateService.UpdateListedInBulkAsync(packages, updateListed.Listed);
```
shishirx34, line 96: Do you think if we should only update the unlisted in delayed batches to prevent the issue we have when some packages do not reflow correctly due to `Updated` time field being very close to the cursor? I remember this causing the differences caught by the v3 metadata monitoring we have.

shishirx34 (approved):
Lets get this merged in, so that we can use it to handle the spam packages.

src/NuGetGallery/Areas/Admin/Controllers/UpdateListedController.cs
joelverhagen: I tried a bunch of IDs in DEV and it didn't seem to repro the monitoring alert... I think the batch update and the raw SQL we generate to update the last edited and last updated values in a transaction mitigates this problem, since it makes timestamps, per ID group, match and each ID group to be in chronological order. 

Perhaps I am missing some context, but from my perspective I need more information about the problem before I try to write a fix.

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
```diff
@@ -32,6 +32,61 @@ public class PackageUpdateService : IPackageUpdateService
...
+
+                _telemetryService.TrackPackagesUpdateListed(packages, listed);
+
+                foreach (var package in packages)
+                {
+                    await _auditingService.SaveAuditRecordAsync(new PackageAuditRecord(
+                        package,
+                        listed ? AuditedPackageAction.List : AuditedPackageAction.Unlist));
+
+                    _indexingService.UpdatePackage(package);
```
zhhyu, line 85: I am not familiar with how it works here. But it seems that "await UpdatePackagesAsync(packages);" at line 73 has already updated the index somehow. Is it?

zhhyu:
> using System.Threading.Tasks;

nit: the order of dependencies.

---
Refers to: src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs:9 in 374a79e. [](commit_id = 374a79ec8ad6c79a2679ae3d5ec9c6a646abc900, deletion_comment = False)

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
```diff
@@ -32,6 +32,61 @@ public class PackageUpdateService : IPackageUpdateService
             _indexingService = indexingService ?? throw new ArgumentNullException(nameof(indexingService));
         }
 
+        public async Task UpdateListedInBulkAsync(IReadOnlyList<Package> packages, bool listed)
```
zhhyu, line 35: I see the interface comment that this method will ignore the package that already has a matching listed state. But it seems that there are no filters here for these cases. Is this expected? Will it be a better design if we move this filter to the layer ("UpdateListedController") above it? My apologies that maybe I miss something here.

src/NuGetGallery/Areas/Admin/Controllers/UpdateListedController.cs
```diff
@@ -0,0 +1,107 @@
...
+                            group.Key,
+                            PackageDeprecationFieldsToInclude.DeprecationAndRelationships)
+                        .Where(x => normalizedVersions.Contains(x.NormalizedVersion))
+                        .ToList();
+                }
+
+                packages = packages
+                    .Where(x => x.PackageStatusKey != PackageStatus.Deleted)
+                    .Where(x => x.PackageStatusKey != PackageStatus.FailedValidation)
+                    .ToList();
```
zhhyu, line 88: I don't know too much about the business logic around this. Is that a concern for packages in validating?

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
joelverhagen: Great catch! Yes, you are totally right. I will leave the `updateIndex` as true explicitly to make this clearer.

src/NuGetGallery/Areas/Admin/Controllers/AdminControllerBase.cs
```diff
@@ -1,12 +1,109 @@
...
+            //   3) Any of the above separated by comma
+            // We are not using Lucene index here as we want to have the database values.
+
+            var queryParts = query.Split(new[] { ',', '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+
+            var packages = new List<Package>();
+            var completedQueries = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (var queryPart in queryParts)
+            {
+                var spitQuery = queryPart.Split(new[] { ' ', '/' }, StringSplitOptions.RemoveEmptyEntries);
```
loic-sharma, line 31:
```suggestion
                var splitQuery = queryPart.Split(new[] { ' ', '/' }, StringSplitOptions.RemoveEmptyEntries);
```

(Though `spitQuery does sound very cool!)

src/NuGetGallery.Services/PackageManagement/PackageUpdateService.cs
joelverhagen: Good catch. Fixed. The intent is to no longer no-op.

src/NuGetGallery/Areas/Admin/Controllers/UpdateListedController.cs
joelverhagen: Currently users are able to unlisted/relist via API while their package is still validating. I mimicked this.

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
```diff
@@ -82,13 +93,13 @@
...
+                            <span>Clear Content Cache</span>
                     </text>,
-                    actionName: "ClearContentCache",
-                    controllerName: "Home",
-                    role: string.Empty,
-                    area: "Admin")
+             actionName: "ClearContentCache",
+             controllerName: "Home",
+             role: string.Empty,
+             area: "Admin")
```
loic-sharma, line 102: Looks like Razor's auto disformatter striked again 😱

joelverhagen: Yeah, I left this one since it seems to make it happy, to reach a steady state. I can undo if desired.

loic-sharma: Would it be possible to move the `htmlContent` parameter to be last? That way the parameters currently on lines 99-102 would be properly aligned and there'd be minimal weirdness after Razor formatting does its thing.

loic-sharma (approved):

src/NuGetGallery/Areas/Admin/Views/Home/Index.cshtml
joelverhagen: Done.

zhhyu (approved):

Issue #8693, opened 2021-07-20
Add imgur in trusted domains


Tagliatti:
Hi, I would like imgur.com to be included in trusted domains if possible. It is widely used in the README of various projects and is one of the top free image hosts.
Thanks.

Tagliatti:
PR #8694

lyndaidaii:
@Tagliatti, I will work with you to evaluate this domain. We will evaluate this domain from privacy and security perspective. 
I will let you know once we have a conclusion about this domain. thank you!

loic-sharma:
FYI the following popular .NET projects on GitHub use imgur as their image host:

GitHub project | Stars
-- | --
[Wox-launcher/Wox](https://www.github.com/Wox-launcher/Wox#readme) | 20493
[xupefei/Locale-Emulator](https://www.github.com/xupefei/Locale-Emulator#readme) | 6194
[HearthSim/Hearthstone-Deck-Tracker](https://www.github.com/HearthSim/Hearthstone-Deck-Tracker#readme) | 4411
[marian42/wavefunctioncollapse](https://www.github.com/marian42/wavefunctioncollapse#readme) | 3101
[MediaBrowser/Emby](https://www.github.com/MediaBrowser/Emby#readme) | 2882
[Ombi-app/Ombi](https://www.github.com/Ombi-app/Ombi#readme) | 2565
[keijiro/Skinner](https://www.github.com/keijiro/Skinner#readme) | 2474
[IgnaceMaes/MaterialSkin](https://www.github.com/IgnaceMaes/MaterialSkin#readme) | 2303
[sqlkata/querybuilder](https://www.github.com/sqlkata/querybuilder#readme) | 1964
[Codeusa/SteamCleaner](https://www.github.com/Codeusa/SteamCleaner#readme) | 1679
[mellinoe/veldrid](https://www.github.com/mellinoe/veldrid#readme) | 1644
[antonpup/Aurora](https://www.github.com/antonpup/Aurora#readme) | 1598
[kwsch/PKHeX](https://www.github.com/kwsch/PKHeX#readme) | 1593
[keijiro/Klak](https://www.github.com/keijiro/Klak#readme) | 1399
[BeardedManStudios/ForgeNetworkingRemastered](https://www.github.com/BeardedManStudios/ForgeNetworkingRemastered#readme) | 1387
[Unity-Technologies/VolumetricLighting](https://www.github.com/Unity-Technologies/VolumetricLighting#readme) | 1348
[...]

PR #8694, opened 2021-07-20
Added i.imgur.com in trusted domain

Tagliatti:
Issue #8693


lyndaidaii (changes requested):
@Tagliatti, could you add `i.imgur.com` to the list in alphabet order?  which add it after `        img.shields.io`, thank you

lyndaidaii (approved):

joelverhagen (approved):

Issue #8428, opened 2021-02-22
Launch NuGet Package Explorer from NuGet.org


Giorgi:
It would be very useful if there was a link which uses custom scheme to launch [NuGet Package Explorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer) with the current package to load.

skofman1:
@clairernovotny , what do you think?

clairernovotny:
We have something special coming soon. Will email offline.

clairernovotny:
It's live!

https://twitter.com/UnoPlatform/status/1407379209467617280
https://platform.uno/blog/nuget-package-explorer-is-now-a-web-app-powered-by-uno-platform/

Check out NuGet.info. You can also go to any NuGet.org package and change the .org to .info and it'll load up!

loic-sharma:
FYI we do have a link to NuGet Package Explorer already, but it is only shown on Internet Explorer:

https://github.com/NuGet/NuGetGallery/blob/908da7fb5c1ae6768fd04e5ee093bab6ef5c1dba/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml#L1002-L1005

Replacing that link with nuget.info would be great! @jcjiang and @anangaur what are your thoughts? :)

anangaur:
>FYI we do have a link to NuGet Package Explorer already, but it is only shown on Internet Explorer

Thats a revelation to me. Never seen it (no surprise here) 😄 . Replacing it makes sense to me as we enable for all browsers.

Pasting the current IE screenshot for others:
![image](https://user-images.githubusercontent.com/14800916/123006640-dba6c300-d36c-11eb-9497-b9298b83933d.png)

jcjiang:
Love the idea of replacing that link with nuget.info!

I think we should change 'Package Explorer' to 'NuGet Package Explorer' for clarity, but the icon and general placement of the IE link looks good.

anangaur:
I would rather call it nuget.info. Something like: `<icon> Explore package at nuget.info`. WYSIWYG

loic-sharma:
Or: `<icon> Open in nuget.info`

clairernovotny:
I don't like nuget.info. That's not the brand. Sorry.

loic-sharma:
@clairernovotny What would you prefer instead? `Open in NuGet Package Explorer`, `Open in Package Explorer`, or something else?

clairernovotny:
[...]

PR #8697, opened 2021-07-20
Display NuGet Package Explorer link on details page.

dannyjdev:
Summary of the changes:

* A link was added for both detail page versions (current, tabs).
* The link sends you to a nuget.info page where you can see the package assets.
* The link will only be displayed when the package is "Available", to avoid sending the user to a broken/non ready page.
* When you click the link it goes to a new page instead of using the current one.
* Minor refactor to the FuGet link to behave like this one.

### DisplayPackage page:
DisplayPackage

### DisplayPackageV2 page:
DisplayPackageV2

Addresses https://github.com/nuget/nugetgallery/issues/8428

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
@@ -999,10 +999,6 @@
                             &nbsp;(@Model.LatestSymbolsPackage.FileSize.ToUserFriendlyBytesLabel())
                         </li>
                     }
-                    <li class="no-clickonce">
-                        <i class="ms-Icon ms-Icon--OpenInNewWindow" aria-hidden="true"></i>
-                        <a href="@Url.ExplorerDeepLink(2, Model.Id, Model.Version)" title="Explore the nupkg with the NuGet Package Explorer (IE only)" rel="nofollow">Open in Package Explorer</a>
-                    </li>
                 }
 
                 @if (Model.CanEdit || Model.CanManageOwners || Model.CanUnlistOrRelist)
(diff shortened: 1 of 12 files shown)


src/NuGetGallery/Views/Packages/DisplayPackage.cshtml
```diff
@@ -1085,7 +1081,21 @@
...
-                @if (Model.IsFuGetLinksEnabled && !string.IsNullOrEmpty(Model.FuGetUrl))
+                @if (Model.CanDisplayNuGetPackageExplorerLink())
+                {
+                    var disclaimer = "nuget.info is a 3rd party website, not controlled by Microsoft. This link is made available to you per the NuGet Terms of Use.";
+
+                    <li>
+                        <i class="ms-Icon ms-Icon--FabricFolderSearch" aria-hidden="true" aria-label="@disclaimer" title="@disclaimer"></i>
+                        <a href="@Model.NuGetPackageExplorerUrl" data-track="outbound-nugetpackageexplorer-url"
+                           aria-label="open in NuGet Package Explorer"
+                           title="Explore additional package info on NuGet Package Explorer" target="_blank" rel="nofollow">
```
joelverhagen, line 1092: I don't think we should be using `target="_blank"`. This is different behavior than the project URL or repo URL which I don't see as categorically different. However I don't feel strongly.

joelverhagen: What was the motivation for this change on the FuGet link? Perhaps I am missing something

dannyjdev: With `target="_blank"` we open a new browser tab with the Fuget/NPE page while keeping the current NuGet display package page. Since those links sends you to a 3rd party site, I think it's better to have our NuGet display package page open while the user go and see those sites for more information. Basically to avoid the user to go backwards on the browser history to return to the same NuGet display package page.

joelverhagen:
Looks good to me! Thanks for cleaning up the old, IE-only thing 👏.

Wanna take a final look, @jcjiang + @clairernovotny?

joelverhagen (approved):

Issue #8698, opened 2021-07-22
[Redesign] Fix registration's link to README


loic-sharma:

loic-sharma:
@dannyjdev You can verify this by:

1. Open Package Manager UI in Visual Studio
2. Set your NuGet package source to the target environment
3. Use the `Browse` tab to look at a package with an embedded README. There should be a `README` link
4. Click the `README` link. It should open the display package page on the target environment and scroll down to the README

PR #8699, opened 2021-07-22
[Redesign] Fix registration's link to the README

loic-sharma:
## Background

For packages with an embedded readme, registration's `readmeUrl` links to the display package page using the `#show-readme-container` URL fragment. For example: https://api.nuget.org/v3/registration5-gz-semver2/knapcode.torsharp/index.json:

<details>
<summary>Example registration JSON for package with embedded readme...</summary>

```json
{
  ...

  "items": [
    {
      ...

      "items": [
        {
          ...

          "@id": "https://api.nuget.org/v3/registration5-gz-semver2/knapcode.torsharp/2.7.0.json",
          "catalogEntry": {
            ...
            "licenseUrl": "https://www.nuget.org/packages/Knapcode.TorSharp/2.7.0/license",
            "readmeUrl": "https://www.nuget.org/packages/Knapcode.TorSharp/2.7.0#show-readme-container",
            ...
            "version": "2.7.0"
          },
          ...
        },

        ...
      ],

      ...
    }
  ]

  ...
}
```

</details>

## Fix

In the current design, this is implemented using Bootstrap 3's collapse component. The redesign broke this feature as it uses Bootstrap 3's tab component instead, which does not support opening using fragments. This change re-adds support for the `#show-readme-container` fragment link using Javascript.

I also fixed some ARIA attributes to improve accessibility.

Addresses https://github.com/NuGet/NuGetGallery/issues/8698

### DIFF
--- a/src/NuGetGallery/Scripts/gallery/page-display-package-v2.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package-v2.js
@@ -56,6 +56,12 @@
     var packageManagerStorageKey = 'preferred_package_manager';
     var bodyStorageKey = 'preferred_body_tab';
 
+    var restorePreferredBodyTab = true;
+    if (window.location.hash === '#show-readme-container') {
+        $('#readme-body-tab').focus();
+        restorePreferredBodyTab = false;
+    }
+
     if (storage) {
         // Restore preferred package manager selection from localStorage.
         var preferredPackageManagerId = storage.getItem(packageManagerStorageKey);
@@ -64,9 +70,11 @@
         }
 
         // Restore preferred body tab selection from localStorage.
-        var preferredBodyTab = storage.getItem(bodyStorageKey);
-        if (preferredBodyTab) {
-            $('#' + preferredBodyTab).tab('show');
+        if (restorePreferredBodyTab) {
+            var preferredBodyTab = storage.getItem(bodyStorageKey);
+            if (preferredBodyTab) {
+                $('#' + preferredBodyTab).tab('show');
+            }
         }
 
         // Make sure we save the user's preferred body tab to localStorage.
--- a/src/NuGetGallery/Scripts/gallery/page-display-package-v2.js
+++ b/src/NuGetGallery/Scripts/gallery/page-display-package-v2.js
@@ -56,6 +56,8 @@
     var packageManagerStorageKey = 'preferred_package_manager';
     var bodyStorageKey = 'preferred_body_tab';
 
+    // The V3 registration API links to the display package page's README using
+    // the 'show-readme-container' URL fragment.
     var restorePreferredBodyTab = true;
     if (window.location.hash === '#show-readme-container') {
         $('#readme-body-tab').focus();
--- a/src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
@@ -494,14 +494,25 @@
                 <ul class="nav nav-tabs" role="tablist">
                     @if (!Model.Deleted)
                     {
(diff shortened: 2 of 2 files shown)


src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
```diff
@@ -494,14 +494,25 @@
                 <ul class="nav nav-tabs" role="tablist">
                     @if (!Model.Deleted)
                     {
-                        <li role="presentation" class="active"><a href="#readme-tab" aria-controls="ReadMe" role="tab" data-toggle="tab" id="readme-body-tab" class="body-tab">README</a></li>
-                        <li role="presentation"><a href="#dependencies-tab" aria-controls="Dependencies" role="tab" data-toggle="tab" id="dependencies-body-tab" class="body-tab">Dependencies</a></li>
+                        <li role="presentation" class="active" id="show-readme-container">
```
loic-sharma, line 497: I added the `id="show-readme-container"` attribute here

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
```diff
@@ -494,14 +494,25 @@
...
+                            <a href="#readme-tab" aria-controls="readme-tab" role="tab" data-toggle="tab" id="readme-body-tab" class="body-tab">README</a>
```
loic-sharma, line 498: I fixed the `aria-controls` attributes on these tab links.

src/NuGetGallery/Scripts/gallery/page-display-package-v2.js
```diff
@@ -56,6 +56,12 @@
     var packageManagerStorageKey = 'preferred_package_manager';
     var bodyStorageKey = 'preferred_body_tab';
 
+    var restorePreferredBodyTab = true;
+    if (window.location.hash === '#show-readme-container') {
+        $('#readme-body-tab').focus();
```
loic-sharma, line 61: The current design focuses on the collapsible section when you use the `#show-readme-container` URL fragment. For example: https://www.nuget.org/packages/Knapcode.TorSharp/2.7.0#show-readme-container

image


This line mimics that behavior.

dannyjdev (approved):

joelverhagen:
I think the expander section works generically with:
https://github.com/NuGet/NuGetGallery/blob/dab825905a5844e999408a2413d0b6b3417d672a/src/NuGetGallery/Scripts/gallery/common.js#L204-L234

Would it make sense to do something similar for tabs?

loic-sharma:
@joelverhagen Nice catch, I didn't realize we had implemented that URL functionality.

> Would it make sense to do something similar for tabs?

This would be useful, but this particular fix wouldn't use that general/shared logic. The URL fragment to open the README, `show-readme-container`, is specific to the naming scheme we use for collapsible sections (and changing this URL fragment would require rebuilding registration). Tabs use a different naming scheme: `readme-tab`, `dependencies-tab`, etc... If we add generic "open tab using URL fragment" logic, ideally it would work using this new naming scheme with a URL fragment like `#readme-tab`. So if we add generic logic, we would still need this "backwards compatibility" hack to support the existing registration API.

I'd suggest we punt the generic logic until we have a scenarios that would use this logic. Let me know if you feel otherwise @joelverhagen .

joelverhagen (approved):

lyndaidaii (approved):

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
zhhyu, line 497: I am not very sure. Do we need the id here?

zhhyu (approved):

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
loic-sharma: Yup, as the browser will move down the page to where the `show-readme-container` ID is when I have the URL fragment `#show-readme-container`.

Issue #8706, opened 2021-07-27
Serve static assets with efficient cache policy


loic-sharma:
### Related Problem

We can speed up page load time on repeat visits by enable HTTP caching on additional static assets.

See: https://web.dev/uses-long-cache-ttl/?utm_source=lighthouse&utm_medium=devtools

### The Elevator Pitch

Better performance! Improved Lighthouse score!

### Additional Context and Details

To reproduce:

1. Open Chrome or Edge
2. Navigate to: https://www.nuget.org/packages/Newtonsoft.Json/
3. Open Developer Tools (press `F12`)
4. Go to `Lighthouse` tab and run a report
5. Open the performance section:

![image](https://user-images.githubusercontent.com/737941/127074571-4e7fcfbc-7042-4d9f-b552-71551dbcaa23.png)

Issue #8605, opened 2021-05-24
[Redesign display package page] Use Google lighthouse on page 


sophiamfavro:
Run Google insights to do performance test on page and make improvements based on findings. Repeat until the page is as best as possible.

PR #8707, opened 2021-07-27
[Redesign] Improve Lighthouse score

loic-sharma:
Lighthouse is a tool to check a website follows best practices: https://developers.google.com/web/tools/lighthouse

Category | Score before | Score after | Change
-- | -- | -- | --
Performance | 100 | 100 | -
Accessibility | 96 | 100 | +4
Best practices | 80 | 87 | +7
SEO | 91 | 100 | +9

Remaining issues I did not address:
* Fabric font loading inefficiently. May be fixed by updating `fabric.css`.
* Gallery assets served without efficient cache policy. Tracked by: https://github.com/NuGet/NuGetGallery/issues/8706
* Known vulnerabilities in jQuery@3.4.1 and Moment.js@2.18.1. Tracked by https://github.com/NuGet/Engineering/issues/3983
* SameSite cookie use by WCP consent cookie. Owned by WCP team

Addresses https://github.com/nuget/nugetgallery/issues/8605

### DIFF
--- a/src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
+++ b/src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
@@ -137,6 +137,10 @@
     }
 }
 
+@section Meta {
+    <meta name="description" content="@Model.ShortDescription">
+}
+
 @section SocialMeta {
     @if (!String.IsNullOrWhiteSpace(ViewBag.FacebookAppID))
     {
@@ -1056,7 +1060,7 @@
                                     <i class="ms-Icon ms-Icon--FabricFolderSearch" aria-hidden="true" aria-label="@disclaimer" title="@disclaimer"></i>
                                     <a href="@Model.NuGetPackageExplorerUrl" data-track="outbound-nugetpackageexplorer-url"
                                        aria-label="open in NuGet Package Explorer"
-                                       title="Explore additional package info on NuGet Package Explorer" target="_blank" rel="nofollow">
+                                       title="Explore additional package info on NuGet Package Explorer" target="_blank" rel="nofollow noreferrer">
                                         Open in NuGet Package Explorer
                                     </a>
                                 </li>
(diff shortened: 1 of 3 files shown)


src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
```diff
@@ -1056,7 +1060,7 @@
                                     <i class="ms-Icon ms-Icon--FabricFolderSearch" aria-hidden="true" aria-label="@disclaimer" title="@disclaimer"></i>
                                     <a href="@Model.NuGetPackageExplorerUrl" data-track="outbound-nugetpackageexplorer-url"
                                        aria-label="open in NuGet Package Explorer"
-                                       title="Explore additional package info on NuGet Package Explorer" target="_blank" rel="nofollow">
+                                       title="Explore additional package info on NuGet Package Explorer" target="_blank" rel="nofollow noreferrer">
```
loic-sharma, line 1063: This improves privacy and security.

See: https://stackoverflow.com/questions/50773152/when-should-i-use-rel-noreferrer
See: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel

`rel` value | Description
-- | --
nofollow | Indicates that the current document's original author or publisher does not endorse the referenced document.
noreferrer | No Referer header will be included. Additionally, has the same effect as noopener.
noopener | Creates a top-level browsing context that is not an auxiliary browsing context if the hyperlink would create either of those, to begin with (i.e., has an appropriate target attribute value).

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
```diff
@@ -1113,17 +1115,17 @@
...
-                            {
-                                @ViewHelpers.AlertWarning(@<text>This package has no owners and is not being actively maintained.</text>)
-                            }
-                            else
-                            {
-                                foreach (var owner in Model.Owners)
+                        @if (!Model.Owners.Any())
+                        {
+                            @ViewHelpers.AlertWarning(@<text>This package has no owners and is not being actively maintained.</text>)
+                        }
```
loic-sharma, line 1121: Previously we would nest a `<div>` in a `<ul>` if the package had owners. That is not a best practice

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
```diff
@@ -137,6 +137,10 @@
     }
 }
 
+@section Meta {
+    <meta name="description" content="@Model.ShortDescription">
```
loic-sharma, line 141: This improves our SEO

zhhyu:
Looking now ...

dannyjdev (approved):

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
zhhyu: It should be "if the package had no owners", is it? Will the appearance be changed for packages without owners?

zhhyu (approved):

PR #8712, opened 2021-07-28
Adding an option to disable admin panel

agr:
Progress on https://github.com/NuGet/Engineering/issues/3418

Adds an `AdminPanelEnabled` option to the configuration (defaulted to `true`) that controls whether `Admin` area of the site gets registered as well as some admin routes not directly under `/Admin`.

`AdminHelper` was added for ease of checking and caching the results. `AdminAction` replaced the `UIAuthorize(Roles = "Admins")` to contain the check.

### DIFF
--- a/src/NuGetGallery/App_Start/Routes.cs
+++ b/src/NuGetGallery/App_Start/Routes.cs
@@ -271,20 +271,28 @@ namespace NuGetGallery
                 new { controller = "Packages", action = "SetLicenseReportVisibility", visible = false },
                 new { version = new VersionRouteConstraint() });
 
-            routes.MapRoute(
-                RouteName.PackageReflowAction,
-                "packages/manage/reflow",
-                new { controller = "Packages", action = nameof(PackagesController.Reflow) });
+            if (adminPanelEnabled)
+            {
+                routes.MapRoute(
+                    RouteName.PackageReflowAction,
+                    "packages/manage/reflow",
+                    new { controller = "Packages", action = nameof(PackagesController.Reflow) });
 
-            routes.MapRoute(
-                RouteName.PackageRevalidateAction,
-                "packages/manage/revalidate",
-                new { controller = "Packages", action = nameof(PackagesController.Revalidate) });
+                routes.MapRoute(
+                    RouteName.PackageRevalidateAction,
+                    "packages/manage/revalidate",
+                    new { controller = "Packages", action = nameof(PackagesController.Revalidate) });
 
-            routes.MapRoute(
-                RouteName.PackageRevalidateSymbolsAction,
-                "packages/manage/revalidate-symbols",
-                new { controller = "Packages", action = nameof(PackagesController.RevalidateSymbols) });
+                routes.MapRoute(
+                    RouteName.PackageRevalidateSymbolsAction,
+                    "packages/manage/revalidate-symbols",
+                    new { controller = "Packages", action = nameof(PackagesController.RevalidateSymbols) });
+
+                routes.MapRoute(
+                    RouteName.PackageDeleteAction,
+                    "packages/manage/delete",
+                    new { controller = "Packages", action = "Delete" });
+            }
 
             var packageVersionActionRoute = routes.MapRoute(
                 RouteName.PackageVersionAction,
@@ -296,11 +304,6 @@ namespace NuGetGallery
                 RouteName.PackageAction,
                 "packages/{id}/{action}",
                 new { controller = "Packages" });
-
-            var packageDeleteRoute = routes.MapRoute(
-                RouteName.PackageDeleteAction,
-                "packages/manage/delete",
-                new { controller = "Packages", action = "Delete" });
 
             var confirmationRequiredRoute = routes.MapRoute(
                 "ConfirmationRequired",
--- a/src/NuGetGallery/App_Start/Routes.cs
+++ b/src/NuGetGallery/App_Start/Routes.cs
@@ -10,11 +10,11 @@ namespace NuGetGallery
 {
     public static class Routes
     {
-        public static void RegisterRoutes(RouteCollection routes, bool feedOnlyMode = false)
+        public static void RegisterRoutes(RouteCollection routes, bool feedOnlyMode = false, bool adminPanelEnabled = false)
         {
             if (!feedOnlyMode)
             {
-                RegisterUIRoutes(routes);
+                RegisterUIRoutes(routes, adminPanelEnabled);
             }
             else
             {
@@ -29,7 +29,7 @@ namespace NuGetGallery
             RegisterApiV2Routes(routes);
         }
 
-        public static void RegisterUIRoutes(RouteCollection routes)
+        public static void RegisterUIRoutes(RouteCollection routes, bool adminPanelEnabled)
         {
             routes.MapRoute(
                 RouteName.Home,
@@ -449,11 +449,14 @@ namespace NuGetGallery
                 "account/changeMultiFactorAuthentication",
                 new { controller = "Users", action = "ChangeMultiFactorAuthentication" });
 
-            routes.MapRoute(
-                RouteName.AdminDeleteAccount,
-                "account/delete/{accountName}",
-                new { controller = "Users", action = "Delete" },
-                new RouteExtensions.ObfuscatedPathMetadata(2, Obfuscator.DefaultTelemetryUserName));
+            if (adminPanelEnabled)
+            {
+                routes.MapRoute(
+                    RouteName.AdminDeleteAccount,
+                    "account/delete/{accountName}",
+                    new { controller = "Users", action = "Delete" },
+                    new RouteExtensions.ObfuscatedPathMetadata(2, Obfuscator.DefaultTelemetryUserName));
+            }
 
             routes.MapRoute(
                 RouteName.UserDeleteAccount,
--- a/src/NuGetGallery/Controllers/AccountsController.cs
+++ b/src/NuGetGallery/Controllers/AccountsController.cs
@@ -338,7 +338,7 @@ namespace NuGetGallery
         }
 
         [HttpGet]
-        [UIAuthorize(Roles = "Admins")]
+        [AdminAction()]
         public virtual ActionResult Delete(string accountName)
         {
             var accountToDelete = UserService.FindByUsername(accountName) as TUser;
@@ -351,7 +351,7 @@ namespace NuGetGallery
         }
 
         [HttpDelete]
-        [UIAuthorize(Roles = "Admins")]
+        [AdminAction()]
         [RequiresAccountConfirmation("Delete account")]
         [ValidateAntiForgeryToken]
         public virtual async Task<ActionResult> Delete(DeleteAccountAsAdminViewModel model)
--- a/src/NuGetGallery/Filters/AdminActionAttribute.cs
+++ b/src/NuGetGallery/Filters/AdminActionAttribute.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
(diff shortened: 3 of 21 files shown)


agr:
Please check https://github.com/NuGet/NuGetGallery/pull/8713/files first to make `DisplayPackage.cshtml` diff nicer.

src/NuGetGallery/UrlHelperExtensions.cs
```diff
@@ -1387,6 +1387,7 @@ public static string About(this UrlHelper url, bool relativeUrl = true)
 
         public static string Admin(this UrlHelper url, bool relativeUrl = true)
         {
+            // TODO: change to something when admin panel is disabled?
```
dannyjdev, line 1390: Are we going to track this TODO on a GitHub issue? or is it not necessary?

agr: Added exception throwing there if admin panel is disabled. All the calls to this method should check it before calling (they do now).

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
```diff
@@ -990,7 +990,7 @@
                                 </li>
                             }
 
-                            @if (Model.Available && User.IsAdministrator())
+                            @if (AdminHelper.IsAdminPanelEnabled && Model.Available && User.IsAdministrator())
```
joelverhagen, line 993: For all of these `PostLink` admin cases... is it possible to replicate the behavior in a plain admin panel? Or will we need the whole site (e.g. package details page included) to be available from the secure location in addition to the admin panel pages? I was imagining the secure location only needing the `/admin/*` pages but perhaps it's simplest to allow the whole site...

src/NuGetGallery/Views/Users/Profiles.cshtml
```diff
@@ -28,7 +28,7 @@
                     <div class="description">Total @(Model.TotalPackageDownloadCount == 1 ? "download" : "downloads") of packages</div>
                 </div>
             </div>
-            @if (User.IsAdministrator())
+            @if (AdminHelper.IsAdminPanelEnabled && User.IsAdministrator())
```
joelverhagen, line 31: Are read-only flows okay outside of the secure location?

joelverhagen (changes requested):
I mainly just have questions.

src/NuGetGallery/Views/Users/Profiles.cshtml
agr: My goal was to completely cut down admin functionality (even read-only): if someone gains admin privileges, it shouldn't give the attacker any extra information. I wasn't explicitly told to do that by security folks, so I can revert those and ask about it explicitly during the security review.

src/NuGetGallery/Views/Packages/DisplayPackageV2.cshtml
agr: The idea was to run the complete web site in secure location. Carving out everything but admin panel sounds a bit too much.

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -449,11 +452,14 @@ public static void RegisterUIRoutes(RouteCollection routes)
...
-                "account/delete/{accountName}",
-                new { controller = "Users", action = "Delete" },
-                new RouteExtensions.ObfuscatedPathMetadata(2, Obfuscator.DefaultTelemetryUserName));
+            if (adminPanelEnabled)
+            {
+                routes.MapRoute(
+                    RouteName.AdminDeleteAccount,
+                    "account/delete/{accountName}",
+                    new { controller = "Users", action = "Delete" },
+                    new RouteExtensions.ObfuscatedPathMetadata(2, Obfuscator.DefaultTelemetryUserName));
```
loic-sharma, line 461: Should we move this up to be next to the other admin panel routes?

src/NuGetGallery/Filters/AdminActionAttribute.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
```
loic-sharma, line 1: Don't forget the file header :)

src/NuGetGallery/Filters/AdminActionAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace NuGetGallery.Filters
+{
+    public class AdminActionAttribute : UIAuthorizeAttribute
+    {
+        public override void OnAuthorization(AuthorizationContext filterContext)
+        {
+            if (!AdminHelper.IsAdminPanelEnabled)
+            {
+                filterContext.Result = new HttpStatusCodeResult(HttpStatusCode.Forbidden);
```
loic-sharma, line 16: Should this return immediately?

src/NuGetGallery/Controllers/AccountsController.cs
```diff
@@ -338,7 +338,7 @@ public virtual ActionResult DeleteRequest(string accountName = null)
         }
 
         [HttpGet]
-        [UIAuthorize(Roles = "Admins")]
+        [AdminAction()]
```
loic-sharma, line 341: We should be able to get rid of the `()`:

```suggestion
        [AdminAction]
```

tests/NuGetGallery.Facts/Helpers/ObfuscationHelperFacts.cs
```diff
@@ -22,7 +22,7 @@ public TheObfuscateCurrentRequestUrlFacts()
                 {
                     _currentRoutes = new RouteCollection();
                     Routes.RegisterApiV2Routes(_currentRoutes);
-                    Routes.RegisterUIRoutes(_currentRoutes);
+                    Routes.RegisterUIRoutes(_currentRoutes, true);
```
loic-sharma, line 25: Please add named parameter for the `true` to clarify

loic-sharma: Same comment for `ClientTelemetryPIIProcessorTests`

src/NuGetGallery/App_Start/Routes.cs
agr: I wanted to keep routes grouped by their prefix (i.e. this one sits to next to other `account/*` routes).

src/NuGetGallery/Filters/AdminActionAttribute.cs
agr: Mostly based it on a `UIAuthorizeAttribute` implementation that sets `Result` but still calls base:
https://github.com/NuGet/NuGetGallery/blob/cd383ecedad7dc4adeb09d16ccdc82491f0b1106/src/NuGetGallery/Filters/UiAuthorizeAttribute.cs#L20-L40

The behavior as it is now is: if admin panel is disabled, admins calling actions marked with this attribute will get 403 while non-admins will get 401. If I add the return statement, everybody will get 403. This probably makes a bit more sense (that is, getting proper authorization will not let you use the action), so I'll update the code.

src/NuGetGallery/App_Start/Routes.cs
loic-sharma: Gotcha, resolving this

loic-sharma (approved):
Nice tests

src/NuGetGallery/Helpers/AdminHelper.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+using System;
```
zhhyu, line 3: very nit: empty line

src/NuGetGallery/Helpers/AdminHelper.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Web.Mvc;
+using NuGetGallery.Configuration;
+
+namespace NuGetGallery
+{
+    public static class AdminHelper
+    {
+        private static Lazy<bool> AdminPanelEnabled = new Lazy<bool>(() =>  
+            DependencyResolver.Current.GetService<IGalleryConfigurationService>()
+                .Current.AdminPanelEnabled);
```
zhhyu, line 13: I am a little interested in why we use the lazy approach here. Is that because of DI container may be called to resolve before its initialization?

src/NuGetGallery/App_Start/Routes.cs
```diff
@@ -271,20 +271,28 @@ public static void RegisterUIRoutes(RouteCollection routes)
...
+
+                routes.MapRoute(
+                    RouteName.PackageRevalidateSymbolsAction,
+                    "packages/manage/revalidate-symbols",
+                    new { controller = "Packages", action = nameof(PackagesController.RevalidateSymbols) });
+
+                routes.MapRoute(
+                    RouteName.PackageDeleteAction,
+                    "packages/manage/delete",
+                    new { controller = "Packages", action = "Delete" });
```
zhhyu, line 294: I think that we allow customers to delete packages by themselves . Will the change break that path?

src/NuGetGallery/Helpers/AdminHelper.cs
agr: Without `Lazy` `AdminPanelEnabled` here will be initialized before the rest of the code have a chance to set up DI. The alternative would be to just set it in `DefaultDependenciesModule`, but I didn't want to couple those classes like that.

src/NuGetGallery/App_Start/Routes.cs
agr: Delete action is only available for admins (see the first attribute):
https://github.com/NuGet/NuGetGallery/blob/436398e79436b4d5fc89a74e45703872f4808cae/src/NuGetGallery/Controllers/PackagesController.cs#L2020-L2024

However, there is a legacy redirect for another case. It does not allow to delete, but redirects to a `Manage` action:
https://github.com/NuGet/NuGetGallery/blob/436398e79436b4d5fc89a74e45703872f4808cae/src/NuGetGallery/Controllers/PackagesController.cs#L2012-L2015

Need to look into this one.

tests/NuGetGallery.Facts/Services/ActionsRequiringPermissionsAdminFacts.cs
```diff
@@ -0,0 +1,134 @@
...
+
+namespace NuGetGallery.Services
+{
+    public class ActionsRequiringPermissionsAdminFacts
+    {
+        private int _key = 0;
+
+        public static IEnumerable<object[]> PackageActions = new object[][]
+        {
+            new object[] { (Func<ActionRequiringPackagePermissions>)(() => ActionsRequiringPermissions.DisplayPrivatePackageMetadata) },
```
zhhyu, line 20: I am a little curious about why we define an object array here. Will it also work like this?

src/NuGetGallery.Services/Permissions/ActionRequiringPackagePermissions.cs
```diff
@@ -49,6 +49,25 @@ public PermissionsCheckResult CheckPermissionsOnBehalfOfAnyAccount(User currentU
...
+            }
+            var other = obj as ActionRequiringPackagePermissions;
+            if (other == null)
+            {
+                return false;
+            }
+            return PackageRegistrationPermissionsRequirement == other.PackageRegistrationPermissionsRequirement;
+        }
+
+        public override int GetHashCode()
```
zhhyu, line 66: I am a little curious about the risk without overriding these "Equals" and "GetHashCode" methods. I see we mainly use bit operations for these permissions. My apologies that I may miss something here.

zhhyu (approved):

Issue #8702, opened 2021-07-26
Create an admin panel to revalidate all packages stuck in validation


loic-sharma:
### Problem

Currently admins need to restart validations one-by-one. This can be time intensive if there is an incident that affects many packages.

See: https://portal.microsofticm.com/imp/v3/incidents/postmortem/513953

### Solution

Create an admin panel that revalidates all packages stuck in validation.

PR #8738, opened 2021-08-06
Add bulk revalidate buttons to Validation admin panel

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/8702.

This PR makes three changes to the Validation admin panel

1. **Polish:** Redirect back to admin panel when clicking an individual revalidate (feedback in retrospective from @loic-sharma)
1. **Bug fix:** Fix the individual revalidate link. Previously, only the first form on the page worked due to duplicate IDs 😨
1. **Feature:** Adds two links to the Validation admin panel
    - Revalidate pending packages
    - Revalidate pending symbol packages

Each link queries for all entities of the respective type that are in the `Validating` state then enqueues a validation for each entity.

Each link has a confirmation dialog to avoid accidental clicks.

revall

### DIFF
--- a/src/NuGetGallery/Areas/Admin/Views/Validation/Index.cshtml
+++ b/src/NuGetGallery/Areas/Admin/Views/Validation/Index.cshtml
@@ -17,7 +17,37 @@
 <section role="main" class="container main-container">
     <h2>Package Validations</h2>
 
-    <p>Get <a href="@Url.Action(actionName:"Pending")">pending validations</a>.</p>
+    <div>
+        <a href="@Url.Action(actionName:"Pending")">Get pending validations</a>
+        |
+        @ViewHelpers.PostLink(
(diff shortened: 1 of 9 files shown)


zhhyu:
Looking now ...

src/NuGetGallery/Areas/Admin/Views/Validation/Index.cshtml
```diff
@@ -87,14 +117,15 @@
                             {
                                 { "id", package.Id },
                                 { "version", package.NormalizedVersion },
+                                { "returnUrl", Url.Current() },
```
loic-sharma, line 120: 🚀🚀

loic-sharma (approved):

drewgillies (approved):

